PR #486, opened 2018-01-02
Add metrics to persistence for visibility

vancexu:


common/persistence/persistenceMetricClients.go
```diff
@@ -541,3 +555,162 @@ func (p *metadataPersistenceClient) updateErrorMetric(scope int, err error) {
...
+	sw.Stop()
+
+	if err != nil {
+		p.updateErrorMetric(metrics.PersistenceGetClosedWorkflowExecutionScope, err)
+	}
+
+	return response, err
+}
+
+func (p *visibilityPersistenceClient) updateErrorMetric(scope int, err error) {
```
samarabbas: Need to handle 'EntityNotExistsError' separately.  We don't want to count this error against our SLA.

PR #487, opened 2018-01-02
Fix workflow timeout not created for new execution when ContinuedAsNew

vancexu:


common/persistence/cassandraPersistence.go
```diff
@@ -1075,6 +1075,8 @@ func (d *cassandraPersistence) UpdateWorkflowExecution(request *UpdateWorkflowEx
 		d.CreateWorkflowExecutionWithinBatch(startReq, batch, cqlNowTimestamp)
 		d.createTransferTasks(batch, startReq.TransferTasks, startReq.DomainID, *startReq.Execution.WorkflowId,
 			*startReq.Execution.RunId, cqlNowTimestamp)
+		d.createTimerTasks(batch, startReq.TimerTasks, nil, startReq.DomainID, *startReq.Execution.WorkflowId,
```
samarabbas: Can you use the generated Getter.

service/history/historyEngine.go
```diff
@@ -1140,6 +1149,8 @@ Update_History_Loop:
 
 		// Inform timer about the new ones.
 		e.timerProcessor.NotifyNewTimer(timerTasks)
+		e.timerProcessor.NotifyNewTimer(continueAsNewTimerTasks)
```
samarabbas: It seems weird we call NotifyNewTimer twice.  Can we merge it into a single call to Notify?

PR #489, opened 2018-01-03
implement filter which allow caller choose all events or only close e…

wxing1292:
…vent of when dumping history events

when long poll on end event, use history service for end event
  
#470 experienced some weird issue when merging to master, redo the commit


samarabbas:
This PR was reviewed earlier by me.

PR #497, opened 2018-01-04
Move history and matching failure metric to common

madhuravi:


common/metrics/defs.go
```diff
@@ -535,14 +535,15 @@ const (
 	PersistenceErrBusyCounter
 
 	NumCommonMetrics
+	HistoryFailures
```
samarabbas: NumCommonMetrics need to be the last enum.  Please declare the new ones before this.

common/metrics/defs.go
```diff
@@ -620,13 +620,14 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		PersistenceErrConditionFailedCounter:          {metricName: "persistence.errors.condition-failed", metricType: Counter},
 		PersistenceErrTimeoutCounter:                  {metricName: "persistence.errors.timeout", metricType: Counter},
 		PersistenceErrBusyCounter:                     {metricName: "persistence.errors.busy", metricType: Counter},
+		HistoryFailures:                               {metricName: "history.client.errors", metricType: Counter},
```
samarabbas: can you change the name to be client.history.errors and client.matching.errors

samarabbas: Also rename the scopes to be HistoryClientFailures and HistoryMatchingFailures.

common/metrics/defs.go
madhuravi: Why?

madhuravi: Nm, I see the subsequent constants.

PR #502, opened 2018-01-10
pprof should be initialized only once per process

wxing1292:
tested using 2 canary running concurrently

### DIFF
--- a/common/service/config/pprof.go
+++ b/common/service/config/pprof.go
@@ -23,6 +23,7 @@ package config
 import (
 	"fmt"
 	"net/http"
+	"sync/atomic"
 	// DO NOT REMOVE THE LINE BELOW
 	_ "net/http/pprof"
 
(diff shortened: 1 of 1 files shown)


common/service/config/pprof.go
```diff
@@ -53,10 +63,11 @@ func (initializer *PProfInitializerImpl) Start() error {
 		return nil
 	}
 
-	go func() {
-		initializer.Logger.Info("PProf listen on %d", port)
-		http.ListenAndServe(fmt.Sprintf("localhost:%d", port), nil)
-	}()
-
+	if atomic.CompareAndSwapInt32(&pprofStatus, pprofNotInitialized, pprofInitialized) {
```
samarabbas: What is the situation where Start can be called multiple times?

wxing1292: ./cadence start, i.e. local run

PR #508, opened 2018-01-13
Change schema version check to pick up package path

madhuravi:
Currently, the schema check is using the running directory which will not work if a different repo is using it. Change it to be the directory of this specific file so that even if run from another repo, it will pick the right schema file so it can be used internally.


madhuravi:
Note: Unclear why the idl gen is tagging along. Likely was missed in an earlier PR with a thrift update.

tools/cassandra/version.go
```diff
@@ -124,16 +127,45 @@ func getExpectedVersion(dir string) (string, error) {
...
+		return errors.New("unable to get the current function path")
+	}
+	// Traverse until project root i.e. "cadence" dir to navigate to the schema/ directory
+	projRoot := filename
+	if _, err := ioutil.ReadDir(path.Dir(projRoot)); err != nil {
+		return err
+	}
+	for path.Base(projRoot) != "cadence" {
+		projRoot = path.Dir(projRoot)
+		if projRoot == "." {
```
venkat1109: add a comment that - Dir will return "." if it cannot go one level up anymore

tools/cassandra/version.go
```diff
@@ -124,16 +127,45 @@ func getExpectedVersion(dir string) (string, error) {
...
 
-// CheckCompatibleVersion check the version compatibility
-func CheckCompatibleVersion(cfg config.Cassandra, keyspace string, dirPath string) error {
+// VerifyCompatibleVersion ensures that the installed version of cadence and visibility keyspaces
+// is greater than or equal to the expected version.
+// In most cases, the versions should match. However if after a schema upgrade there is a code
+// rollback, the code version (expected version) would fall lower than the actual version in
+// cassandra.
+func VerifyCompatibleVersion(cfg config.Cassandra) error {
+	_, filename, _, ok := runtime.Caller(1)
```
venkat1109: this doesn't look right, double check

madhuravi: Fixed

PR #516, opened 2018-01-18
Passing in root path to determine cassandra version

madhuravi:
runtime.Caller does not return absolute path in all cases. Changed to pass in the root path so dependent repos can send in appropriate roots.

### DIFF
--- a/cmd/server/cadence.go
+++ b/cmd/server/cadence.go
@@ -23,6 +23,7 @@ package main
 import (
 	"log"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/uber/cadence/common/service/config"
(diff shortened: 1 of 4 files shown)


cmd/server/cadence.go
```diff
@@ -53,7 +54,12 @@ func startHandler(c *cli.Context) {
 	log.Printf("config=\n%v\n", cfg.String())
 
 	cassCfg := cfg.Cassandra
-	if err := cassandra.VerifyCompatibleVersion(cassCfg); err != nil {
+	dir, err := os.Getwd() // dir = /Users/madhuravi/Uber/gocode/src/github.com/uber/cadence
```
venkat1109: how about we just change our schema update process to also update a constant in version.go ? that would be simple and less brittle going forward

madhuravi: I know I brought up  that discussion again but I'm not so convinced about it. It does not introduce much risk since it is the same path, we will know early in local testing etc. if this breaks. It's more risky if it somehow breaks in production without us knowing but I can't see that happening.

venkat1109: ok, will leave it to you - my preference is always for the const in code - its super straightforward - but this is something we can change it when we need it

madhuravi: Agreed. Will try this out now, open to changing it based on what works, can see the value of both.

PR #519, opened 2018-01-20
Unload tasklist when no poller recently

vancexu:
Related to https://github.com/uber/cadence/issues/514


service/matching/service.go
```diff
@@ -52,6 +53,7 @@ func NewConfig() *Config {
 		RangeSize:                       100000,
 		GetTasksBatchSize:               1000,
 		UpdateAckInterval:               10 * time.Second,
+		CheckPollerInterval:             5 * time.Minute,
```
samarabbas: rename this to IdleTasklistCheckInterval

PR #520, opened 2018-01-22
Use domain name in completeActivityByID instead of domainID

vancexu:
Related issue: https://github.com/uber/cadence/issues/515

Also changed existing methods that use domainCache.GetDomain() to use domainCache.GetDomainID()


vancexu:
Also changed existing methods that use `domainCache.GetDomain()` to use `domainCache.GetDomainID()`

PR #523, opened 2018-01-23
Remove throttling logs and set min burst size

madhuravi:
Warn logs are clogging up disk space in matching service.

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -629,7 +629,10 @@ deliverBufferTasksLoop:
 				c.logger.Info("Tasklist manager context is cancelled, shutting down")
 				break deliverBufferTasksLoop
 			}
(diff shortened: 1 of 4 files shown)


service/matching/taskListManager.go
```diff
@@ -81,9 +81,12 @@ func newRateLimiter(maxDispatchPerSecond *float64, ttl time.Duration) rateLimite
 		ttlTimer:             time.NewTimer(ttl),
 	}
 	// Note: Potentially expose burst config in future
-	limiter := rate.NewLimiter(
-		rate.Limit(*maxDispatchPerSecond), int(*maxDispatchPerSecond),
-	)
+	// Set burst to be a minimum of 5 when maxDispatch is set to low numbers
+	burst := int(*maxDispatchPerSecond)
+	if burst <= 5 {
```
samarabbas: Can you define minBurst as a config knob for matching service.

PR #527, opened 2018-01-25
refactor existing domain API for cross DC, refactor existing domain p…

wxing1292:
…ersistence layer, refactor existing domain cache

solve #493 #495


common/cache/domainCache.go
```diff
@@ -86,50 +87,54 @@ func newDomainCacheEntry() *domainCacheEntry {
...
 // store and writes it to the cache with an expiry before returning back
-func (c *domainCache) getDomain(key, id, name string, cache Cache) (*persistence.DomainInfo, *persistence.DomainConfig, error) {
+func (c *domainCache) getDomain(key, id, name string, cache Cache) (*domainCacheEntry, error) {
 	now := c.timeSource.Now().UnixNano()
-	refreshCache := false
-	var info *persistence.DomainInfo
-	var config *persistence.DomainConfig
+	var result *domainCacheEntry
+
+	copyEntry := func(input *domainCacheEntry) *domainCacheEntry {
```
vancexu: Would it be better if put `copyEntry` and `isCacheExpired` outside?  (to save some cycles on definition)

common/persistence/cassandraMetadataPersistence.go
```diff
@@ -176,89 +191,108 @@ func (m *cassandraMetadataPersistence) GetDomain(request *GetDomainRequest) (*Ge
...
-
-		return nil, &workflow.InternalServiceError{
+		return &workflow.InternalServiceError{
 			Message: fmt.Sprintf("GetDomain operation failed. Error %v", err),
 		}
 	}
 
+	domainName := request.Name
+	if len(request.ID) > 0 {
+		query = m.session.Query(templateGetDomainQuery, request.ID)
```
vancexu: I think this change will lead to an additional call to get domain info. Why we cannot get all info with domainID?

wxing1292: Cassandra do not support conditional update to multiple table.
Previously, we were using batch query to atomically update 2 tables, which (in my opinion) is an overkill, and cause a lot of unnecessary load on Cassandra.
This change will use the domains table just for domain uuid -> domain name lookup.
The actual content will be in domain_by_name table.

There will be additional call to Cassandra, but we have cache, and most of time, we do get, not update

common/persistence/dataInterfaces.go
```diff
@@ -1004,3 +1020,13 @@ func (h *SerializedHistoryEventBatch) String() string {
 	return fmt.Sprintf("[encodingType:%v,historyVersion:%v,history:%v]",
 		h.EncodingType, h.Version, string(h.Data))
 }
+
+func (config *ClusterReplicationConfig) serialize() map[string]interface{} {
+	output := make(map[string]interface{})
+	output["cluster_name"] = config.ClusterName
```
vancexu: make `cluster_name` a const probably

common/persistence/cassandraMetadataPersistence.go
```diff
@@ -139,23 +147,30 @@ func (m *cassandraMetadataPersistence) CreateDomain(request *CreateDomainRequest
...
+		request.ReplicationConfig.FailoverVersion,
+		clusterReplicationConfigs,
+	)
 
 	previous := make(map[string]interface{})
 	applied, err := query.MapScanCAS(previous)
+
+	if !applied || err != nil {
+		// Domain already exist.  Delete orphan domain record before returning back to user
+		if errDelete := m.session.Query(templateDeleteDomainQuery, domainUUID).Exec(); errDelete != nil {
```
samarabbas: It is dangerous to delete the domain on any errors.

common/persistence/cassandraMetadataPersistence.go
```diff
@@ -176,89 +191,108 @@ func (m *cassandraMetadataPersistence) GetDomain(request *GetDomainRequest) (*Ge
...
+		&replicationClusters,
+		&version,
+	)
+
+	if err != nil {
+		return nil, handleError(request.Name, request.ID, err)
+	}
+
+	replicationConfig.ActiveClusterName = m.clusterMetadata.GetOrUseDefaultActiveCluster(replicationConfig.ActiveClusterName)
+	for index := range replicationClusters {
```
samarabbas: Can you move this out to a helper method.

common/persistence/cassandraMetadataPersistence.go
```diff
@@ -176,89 +191,108 @@ func (m *cassandraMetadataPersistence) GetDomain(request *GetDomainRequest) (*Ge
...
+		Info:              info,
+		Config:            config,
+		ReplicationConfig: replicationConfig,
+		Version:           version,
 	}, nil
 }
 
 func (m *cassandraMetadataPersistence) UpdateDomain(request *UpdateDomainRequest) error {
-	batch := m.session.NewBatch(gocql.LoggedBatch)
+	clusterReplicationConfigs := []map[string]interface{}{}
```
samarabbas: It would be great to create some helpers around dealing with replicationConfigs to make the code easier to read.

common/persistence/clusterMetadata.go
```diff
@@ -0,0 +1,120 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package persistence
+
+type (
+	// ClusterMetadata provides information about clusters
+	ClusterMetadata interface {
```
samarabbas: Define the interface in dataInterfaces.go and use the same naming convention we use for other persistence API.

samarabbas: Actually I'm not sure if this even belongs in the persistence layer.  Let's talk about this.

common/persistence/cassandraMetadataPersistence.go
wxing1292: this is the create API, and existing logic does the same thing, except existing logic have bug dealing with second create call to Cassandra having errors.

common/persistence/cassandraMetadataPersistence.go
wxing1292: ok

common/persistence/cassandraMetadataPersistence.go
wxing1292: ok

common/persistence/clusterMetadata.go
wxing1292: the thing is, this originally belongs to common package, until go compiler complains about circular reference between its package and persistence package

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -749,17 +759,22 @@ struct DescribeDomainRequest {
 struct DescribeDomainResponse {
   10: optional DomainInfo domainInfo
   20: optional DomainConfiguration configuration
+  30: optional DomainReplicationConfiguration replicationConfiguration
+  40: optional i64 (js.type = "Long") failoverVersion
```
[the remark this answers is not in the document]
wxing1292: because the failover cannot be modified by customer and DomainReplicationConfiguration is used for update API

common/persistence/cassandraMetadataPersistence.go
wxing1292: fixed

PR #529, opened 2018-01-26
Fix throttling burst and add debug logs

madhuravi:


service/matching/matchingEngine_test.go
```diff
@@ -627,9 +627,10 @@ func (s *matchingEngineSuite) TestConcurrentPublishConsumeActivitiesWithZeroDisp
 	const taskCount = 100
 	s.matchingEngine.metricsClient = metrics.NewClient(tally.NewTestScope("test", nil), metrics.Matching)
 	throttleCt := s.concurrentPublishConsumeActivities(workerCount, taskCount, dispatchLimitFn)
+	s.logger.Infof("Number of tasks throttled: %d", throttleCt)
```
vancexu: What should `throttleCt` be?

service/matching/service.go
```diff
@@ -55,7 +55,7 @@ func NewConfig() *Config {
 		UpdateAckInterval:               10 * time.Second,
 		OutstandingTaskAppendsThreshold: 250,
 		MaxTaskBatchSize:                100,
-		MinTaskThrottlingBurstSize:      10000,
+		MinTaskThrottlingBurstSize:      1,
```
vancexu: test change or final value?

service/matching/matchingEngine_test.go
madhuravi: Right now, I test for anything non-zero since it's hard to determine the exact number. The log line is there more for the flaky test. I don't believe it should be flaky anymore but if it is, it's good to see this in the log.

service/matching/service.go
madhuravi: Final value. I did some more research to understand this and having a high burst size would allow a burst of say 10000 requests in an instant which is not what we want. If you're interested, I can explain more offline.

PR #539, opened 2018-01-30
Handle workflow signal itself

vancexu:
Related: https://github.com/uber/cadence/issues/531
This fix will handle case when workflow try to signal itself

### DIFF
--- a/service/history/transferQueueProcessor.go
+++ b/service/history/transferQueueProcessor.go
@@ -630,6 +630,28 @@ func (t *transferQueueProcessorImpl) processSignalExecution(task *persistence.Tr
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


service/history/transferQueueProcessor.go
```diff
@@ -630,6 +630,28 @@ func (t *transferQueueProcessorImpl) processSignalExecution(task *persistence.Tr
 		return nil
 	}
 
+	// handle workflow signal itself
```
samarabbas: Probably you want to do the check for IsWorkflowExecutionRunning before even hitting any processing for signal.  Also make sure to perform this check in childworkflow transfer task processing.

PR #543, opened 2018-02-02
Define dynamic config and integrate in service bootstrap

madhuravi:
Setting up dynamic config and enable for a few config parameters.


mfateev:
I think we agreed on the design that doesn't use lookups to config from the application code. Instead we only use structures to configure any application logic.

madhuravi:
@mfateev Similar to latest commit? Or ware you implying dependency injection?

common/service/dynamicconfig/config.go
```diff
@@ -0,0 +1,117 @@
...
+	return &Collection{client}
+}
+
+// Collection of values that are in dynamic config
+type Collection struct {
+	client Client
+}
+
+// GetIntPropertyWithTaskList gets property with taskList filter and asserts that it's an integer
+func (c *Collection) GetIntPropertyWithTaskList(key Key, defaultVal int) func(string) int {
```
mfateev: I would make this package non Cadence specific. So this method would be something like:

```
func (c *Collection) GetIntPropertyWithStringFilter(key Key, defaultValue int, filter Filter) func(string) int {
	return func(filterValue string) int {
 		filters := make(map[Filter]interface{})
 		filters[filter] = filterValue
 		val, err := c.client.GetValueWithFilters(key, filters)
 		if err != nil {
 			return defaultVal
 		}
 		return val.(int)
 	}
 }
```
Then the code that uses it would make it task list specific. If the task list appears as filter in a lot of places the 
then the following function (outside of this package) can be created:

```
func GetIntPropertyWithTaskListFilter(c *Collection, key Key, defaultValue int) func(string) int {
      return c.GetIntPropertyWithString(key, defaultValue, TaskListName)
}
```

common/service/dynamicconfig/config.go
```diff
@@ -0,0 +1,117 @@
...
+func (c *Collection) GetIntProperty(key Key, defaultVal int) func() int {
+	return func() int {
+		val, err := c.client.GetValue(key)
+		if err != nil {
+			return defaultVal
+		}
+		return val.(int)
+	}
+}
+
```
mfateev: Each of Get methods repeats the same code snippet. What about:

```
func (c *Collection) GetProperty(key Key, defaultVal interface{}) func() interface{} {
       return func() int {
 		val, err := c.client.GetValue(key)
 		if err != nil {
 			return defaultVal
 		}
 		return val
 	}
}

```
Then Get methods become like:

```
func (c *Collection) GetFloat64Property(key Key, defaultVal float64) func() float64 {
      return func() float64 { 
          return GetProperty(key, defaultVal)().(float64) 
      }
}
```

common/service/dynamicconfig/constants.go
```diff
@@ -0,0 +1,75 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dynamicconfig
```
mfateev: I would break this into two packages dynamicconfig and cadenceconfig. The content of this file doesn't belong to generic dynamicconfig one.

service/matching/matchingEngine_test.go
```diff
@@ -195,7 +196,7 @@ func (s *matchingEngineSuite) TestPollForDecisionTasksEmptyResult() {
 
 func (s *matchingEngineSuite) PollForTasksEmptyResultTest(taskType int) {
 	s.matchingEngine.config.RangeSize = 2 // to test that range is not updated without tasks
-	s.matchingEngine.config.LongPollExpirationInterval = 10 * time.Millisecond
+	s.matchingEngine.config.LongPollExpirationInterval = func() time.Duration { return 10 * time.Millisecond }
```
mfateev: May be introduce small helper method? 
     s.matchingEngine.config.LongPollExpirationInterval  = DurationConfigMillis(10)

common/service/dynamicconfig/config.go
```diff
@@ -0,0 +1,117 @@
...
+		val, err := c.client.GetValueWithFilters(key, filters)
+		if err != nil {
+			return defaultVal
+		}
+		return val.(int)
+	}
+}
+
+// GetIntProperty gets property and asserts that it's an integer
+func (c *Collection) GetIntProperty(key Key, defaultVal int) func() int {
```
mfateev: rename defaultVal to defaultValue in all methods.

common/service/dynamicconfig/config.go
madhuravi: What's the benefit of a public generic function? Makes sense as a private method if we have public methods for both task list and domain name.
Also all these methods are for Cadence, what is a non-Cadence use case for this?

common/service/dynamicconfig/config.go
madhuravi: sure but why? Go usually advocates for short variable names as opposed to Java eg. https://talks.golang.org/2014/names.slide#6

common/service/dynamicconfig/config.go
madhuravi: Good idea.

common/service/dynamicconfig/constants.go
madhuravi: Same as above, not sure what the non-cadence usage is. The dynamic config is solely for cadence.

common/service/dynamicconfig/config.go
madhuravi: With generic functions here and the previous filter one as well, there is an extra cast for converting defaultValue from empty interface to the corresponding type. I wrote some benchmark tests to compare the two and the previous non-generic approach is about 20% faster (260 versus 210 ns). I would favor performance here over reducing duplication since this function could be called quite frequently based on where it's being used. Let me know what you think. Will push my commit with benchmark test so you can take a look.

madhuravi: madhurav-C02RQ0RTG8WM:cadence madhuravi$ go test ./common/service/dynamicconfig/ -bench=.
goos: darwin
goarch: amd64
pkg: github.com/uber/cadence/common/service/dynamicconfig
BenchmarkGetIntProperty-8        5000000               279 ns/op
BenchmarkGetIntPropertyOld-8    10000000               219 ns/op
PASS
ok      github.com/uber/cadence/common/service/dynamicconfig    4.122s

common/service/dynamicconfig/constants.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Key represents a key/property stored in dynamic config
+type Key int
+
+func (k Key) String() string {
+	keys := []string{
```
yiminc: This will create an array variable every time this method is called. 
I think a simple switch case clause is easier to understand and more performant.
switch k {
  case unknownKey: 
    return "unknownKey"
  case MinTaskThrottlingBurstSize:
    return "minTaskThrottlingBurstSize"
 ...
}

service/matching/taskListManager.go
```diff
@@ -138,7 +138,9 @@ func newTaskListManager(
 	e *matchingEngineImpl, taskList *taskListID, taskListKind *s.TaskListKind, config *Config,
 ) taskListManager {
 	dPtr := _defaultTaskDispatchRPS
-	rl := newRateLimiter(&dPtr, _defaultTaskDispatchRPSTTL, config.MinTaskThrottlingBurstSize)
+	rl := newRateLimiter(
+		&dPtr, _defaultTaskDispatchRPSTTL, config.MinTaskThrottlingBurstSize(taskList.taskListName),
```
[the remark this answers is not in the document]
madhuravi: Done.

common/service/dynamicconfig/constants.go
madhuravi: Moved it out so array variable is not created everytime.

PR #548, opened 2018-02-09
Add log in frontend API for InternalServiceError

vancexu:


service/frontend/handler.go
```diff
@@ -1702,6 +1703,7 @@ func (wh *WorkflowHandler) error(err error, scope int) error {
 		wh.metricsClient.IncCounter(scope, metrics.CadenceErrQueryFailedCounter)
 		return err
 	default:
+		wh.Service.GetLogger().Errorf("Uncategorized error: %v", err)
```
samarabbas: Instead of Errorf, can you emit the error using a TagErr?

vancexu: Changed.

PR #553, opened 2018-02-09
Fix README for cassandra tool

vancexu:


tools/cassandra/README.md
```diff
@@ -6,19 +6,26 @@ How
 ---
 - Run make bins
 - You should see an executable `cadence-cassandra-tool`
+- Make sure you already created cadence and visibility keyspaces (see `keyspace.cql` in schema)
 
 Setting up initial cassandra schema on a new cluster
 ----------------------------------------------------
 ```
 ./cadence-cassandra-tool -ep 127.0.0.1 -k cadence setup-schema -v 0.0 -- this sets up just the schema version tables with initial version of 0.0
```
madhuravi: There is a make target for this. You can just replace these with
make install-schema

tools/cassandra/README.md
```diff
@@ -6,19 +6,26 @@ How
 ---
 - Run make bins
 - You should see an executable `cadence-cassandra-tool`
+- Make sure you already created cadence and visibility keyspaces (see `keyspace.cql` in schema)
```
madhuravi: This is a bit confusing. Aren't we creating this using the commands below? How will the user make sure they have already created the keyspaces.

tools/cassandra/README.md
vancexu: make sense, added

tools/cassandra/README.md
vancexu: removed

tools/cassandra/README.md
```diff
@@ -7,18 +7,30 @@ How
...
 - You should see an executable `cadence-cassandra-tool`
 
-Setting up initial cassandra schema on a new cluster
+Setting up cassandra schema on a new cluster using Makefile
+----------------------------------------------------
+``` 
+make install-schema
+```
+
+Setting up schema on a new cluster manually
```
madhuravi: I don't think this section is necessary If folks want to dig in, they can easily look up Makefile to see command breakdown.

tools/cassandra/README.md
```diff
@@ -7,18 +7,30 @@ How
 - Run make bins
 - You should see an executable `cadence-cassandra-tool`
 
-Setting up initial cassandra schema on a new cluster
+Setting up cassandra schema on a new cluster using Makefile
```
madhuravi: Can avoid "using Makefile". It's a bit confusing, Instead, you call it out as a shortcut.

PR #563, opened 2018-02-16
Cadence Worker service to host replicator

samarabbas:
New cadence-worker service and bootstrap code to bring up the new
service.  This service currently only host the replicator which consumes
replication tasks from kafka topic and applies to current Cadence
cluster.

MessagingClient interface abstract out interaction with Kafka.
KafkaClient provides implementation for the interface using kafka-client
library which consumes messages from Kafka broker.

Created separate Kafka configuration used for bootstrapping the
KafkaClient created during service startup and passed down to
replicator.

Created replicator.thrift which defines the payload which is sent over
to other clusters through KafkaClient.


service/worker/processor.go
```diff
@@ -0,0 +1,179 @@
...
+func (p *replicationTaskProcessor) processorPump() {
+	defer p.shutdownWG.Done()
+
+	var workerWG sync.WaitGroup
+	for i := 0; i < p.config.ReplicatorConcurrency; i++ {
+		workerWG.Add(1)
+		go p.worker(&workerWG)
+	}
+
+processorPumpLoop:
```
wxing1292: is this loop necessary?

service/worker/processor.go
```diff
@@ -0,0 +1,179 @@
...
+	for {
+		select {
+		case msg, ok := <-p.consumer.Messages():
+			if !ok {
+				p.logger.Info("Worker for replication task processor shutting down.")
+				return // channel closed
+			}
+
+			if task, err := deserialize(msg.Value()); err != nil {
+				p.logger.Errorf("Deserialize Error. Value: %v, Error: %v", string(msg.Value()), err)
```
wxing1292: if the DLQ part is working in progress, plz add a TODO here.
Since we should investigate any message which cannot be processed.

same applies to line: 151, 160

common/messaging/kafkaConfig.go
```diff
@@ -0,0 +1,77 @@
...
+		brokers[cluster] = cfg.Brokers
+	}
+
+	// mapping from topic name to cluster that has that topic
+	topicClusterAssignment := map[string][]string{}
+	for topic, cfg := range k.Topics {
+		topicClusterAssignment[topic] = []string{cfg.Cluster}
+	}
+
+	client := kafkaclient.New(kafka.NewStaticNameResolver(topicClusterAssignment, brokers), zap.NewNop(), tally.NoopScope)
```
venkat1109: suggest passing a real tally scope to emit metrics for debugging

service/worker/processor.go
```diff
@@ -0,0 +1,179 @@
...
+
+	for {
+		select {
+		case msg, ok := <-p.consumer.Messages():
+			if !ok {
+				p.logger.Info("Worker for replication task processor shutting down.")
+				return // channel closed
+			}
+
+			if task, err := deserialize(msg.Value()); err != nil {
```
venkat1109: also emit a metric for this

service/worker/processor.go
samarabbas: You are right, it is not necessary.  Previously I had more case statements which I removed later but forgot to remove the loop.

service/worker/processor.go
samarabbas: Added a TODO.  Adding a DLQ is pretty straightforward as the underlying kafka-client library already has support for this.  But currently the idea is to block the consumer when this happens until the fix for such issues are rolled out.

service/worker/README.md
```diff
@@ -0,0 +1,54 @@
+Cadence Worker (In Development)
```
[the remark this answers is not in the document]
samarabbas: We use the worker naming on the client-side, but not on the server.  Worker is well understood convention for hosting background processing.  Let's stick with worker for now.

service/worker/replicator.go
```diff
@@ -0,0 +1,88 @@
...
+			topicName := getTopicName(cluster)
+			consumerName := getConsumerName(currentClusterName, cluster)
+			r.processors = append(r.processors, newReplicationTaskProcessor(topicName, consumerName, r.client, r.config,
+				r.logger))
+		}
+	}
+
+	for _, processor := range r.processors {
+		if err := processor.Start(); err != nil {
+			return err
```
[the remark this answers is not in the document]
samarabbas: Good point.  I have changed service to call Stop on the replicator if Start returns an error.

service/worker/processor.go
samarabbas: Done.

PR #574, opened 2018-02-22
add domain replication logic

wxing1292:
pending unit test.

solve #569 

close in favor of #586


idl/github.com/uber/cadence/replicator.thrift
```diff
@@ -20,12 +20,25 @@
...
   History
 }
 
+enum DomainOperation {
+  Create
+  Update
+}
+
 struct DomainTaskAttributes {
+  05: optional DomainOperation domainOperation
```
samarabbas: How about we include the actual request payload for domain create/update operations.  Basically from the replicator perspective it receives the message, pull out the domain create/update request and calls the exact same API on the frontend to register/update domain.
The only thing you need to add extra to the request is domain uuid.

service/worker/domainReplicationTaskHandler.go
```diff
@@ -0,0 +1,204 @@
...
+			EmitMetric: task.Config.GetEmitMetric(),
+		},
+		ReplicationConfig: &persistence.DomainReplicationConfig{
+			ActiveClusterName: task.ReplicationConfig.GetActiveClusterName(),
+			FailoverVersion:   task.GetFailoverVersion(),
+			Clusters:          domainReplicator.convertClusterReplicationConfig(task.ReplicationConfig.Clusters),
+		},
+	}
+
+	_, err = domainReplicator.metadataManager.CreateDomain(request)
```
samarabbas: Can we go through the frontend instead of directly calling the metadata CRUD?

wxing1292: on the frontend, there are logics which deal with domain failover, and a lot of input validation.
on the replicator side, the replicator only does basic logic, i.e. save the domain object to DB, with correct DB record version, and comparison of failover version.

idl/github.com/uber/cadence/replicator.thrift
wxing1292: if using the exactly same request information sent from customer and using the same frontend API, this case can happen: 
There are 3 DC, DC1, DC2, DC3. DC1 is currently active.
update1 request only change attr1 by DC1, this update1 is delivered to DC3, but not DC2.

update2 request is a failover request, making DC3 active
update3 request only change attr3 by DC3

DC2 receive the update3 first, and then update2, since the update2 has a lower failover version, the update2 will be rejected.

Then the results are:
DC3 have all updates, DC2 only have 1

wxing1292:
close in favor of #586

PR #577, opened 2018-02-23
Cadence CLI

vancexu:
command line tool for users to perform common tasks with cadence, see readme for details. 
Tests will be done in separate PR

### DIFF
--- a/cmd/tools/cli/main.go
+++ b/cmd/tools/cli/main.go
@@ -1,0 +1,34 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package main
+
+import (
+	"os"
+
+	"github.com/uber/cadence/tools/cli"
+)
+
+// Start using this CLI tool with command
+// See cadence/tools/cli/README.md for usage
+func main() {
+	app := cli.NewCliApp()
+	app.Run(os.Args)
+}
--- a/tools/cli/README.md
+++ b/tools/cli/README.md
@@ -1,0 +1,114 @@
+## What
+Cadence CLI provide an command-line tool for users to perform various tasks on Cadence.  
+Users can perform operations like register, update and describe on domain;  
+also start workflow, show workflow history, signal workflow ... and many other tasks.    
+
+## How
+- Run `make bins`
+- You should see an executable `cadence`
+
+## Quick Start
+Run `./cadence` to view help message. There are some top level commands and global options.   
+Run `./cadence domain` to view help message about operations on domain  
+Run `./cadence workflow` to view help message about operations on workflow  
+(`./cadence help`, `./cadence help [domain|workflow]` will also print help messages)
+
+**Note:** make sure you have cadence server running before using CLI 
+
+### Domain operation examples 
+- Register a new domain named "samples-domain":  
+```
+./cadence --domain samples-domain domain register 
+# OR using short alias  
+./cadence --do samples-domain domain re
+```   
+- View "samples-domain" details:   
+```
+./cadence --domain samples-domain domain describe  
+```
+
+**Tips:**  
+to avoid repeated input global option **domain**, user can export domain-name in environment variable CADENCE_CLI_DOMAIN.
+```
+export CADENCE_CLI_DOMAIN=samples-domain
+
+# then just run commands without --domain flag, like
+./cadence domain desc
+```
+
+### Workflow operation examples
+(The following examples assume you already export CADENCE_CLI_DOMAIN environment variable as Tips above)
+
+- Run workflow: Start a workflow and see it's progress, this command doesn't finish until workflow completes
+```
+./cadence workflow run --tl helloWorldGroup --wt main.Workflow --et 60 -i '"cadence"'
+
+# view help messages for workflow run
+./cadence workflow run -h
+``` 
+Brief explanation:  
+To run a workflow, user must specify 
+1. Tasklist name (--tl), 
+2. Workflow type (--wt), 
+3. Execution start to close timeout in seconds (--et), 
+
+Example uses [this cadence-samples workflow](https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/recipes/helloworld/helloworld_workflow.go) 
+and it takes a string as input, so there is the `-i '"cadence"'`. Single quote `''` is used to wrap input as json. 
+
+**Note:** you need to start worker so that workflow can make progress.  
+(Run `make && ./bin/helloworld -m worker` in cadence-samples to start the worker)
+
+- Start workflow: 
+```
+./cadence workflow start --tl helloWorldGroup --wt main.Workflow --et 60 -i '"cadence"'
+
+# view help messages for workflow start
+./cadence workflow start -h
+
+# for workflow with multiple input, seperate each json with space/newline like
+./cadence workflow start --tl helloWorldGroup --wt main.WorkflowWith3Args --et 60 -i '"your_input_string" 123 {"Name":"my-string", "Age":12345}'
+```
+Workflow `start` command is similar to `run` command and takes same flag options. But it just start the workflow and immediately return workflow_id and run_id.  
+User need to run `show` to view workflow history/progress.  
+
+- Show workflow history
+```
+./cadence workflow show -w 3ea6b242-b23c-4279-bb13-f215661b4717 -r 866ae14c-88cf-4f1e-980f-571e031d71b0
+# a shortcut of this is (without -w -r flag)
+./cadence workflow showid 3ea6b242-b23c-4279-bb13-f215661b4717 866ae14c-88cf-4f1e-980f-571e031d71b0
+
+# if run_id is not provided, it will show the latest run history for that workflow_id
+./cadence workflow show -w 3ea6b242-b23c-4279-bb13-f215661b4717
+# a shortcut of this is
+./cadence workflow showid 3ea6b242-b23c-4279-bb13-f215661b4717
+```
+
+- List open or closed workflow executions
+```
+./cadence workflow list
+```
+
+- Query workflow execution
+```
+# use custom query type
+./cadence workflow query -w <wid> -r <rid> --qt <query-type>
+
+# use build-in query type "__stack_trace" which is supported by cadence client library
+./cadence workflow query -w <wid> -r <rid> --qt __stack_trace
+# a shortcut to query using __stack_trace is (without --qt flag)
+./cadence workflow stack -w <wid> -r <rid> 
+```
+
+- Signal, cancel, terminate workflow
+```
+# signal
+./cadence workflow signal -w <wid> -r <rid> -n <signal-name> -i '"signal-value"'
+
+# cancel
+./cadence workflow cancel -w <wid> -r <rid>
+
+# terminate
+./cadence workflow terminate -w <wid> -r <rid> --reason 
+```
+Terminate a running workflow execution will record WorkflowExecutionTerminated event as closing event in the history. No more decision task will be scheduled for terminated workflow execution.  
+Cancel a running workflow execution will record WorkflowExecutionCancelRequested event in the history, and a new decision task will be scheduled. Workflow has a chance to do some clean up work after cancellation.
--- a/tools/cli/app.go
+++ b/tools/cli/app.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cli
+
+import "github.com/urfave/cli"
+
+const (
+	// Version is the controlled version string. It should be updated every time
+	// before we release a new version.
+	Version = "0.5.0"
+)
+
+// NewCliApp instantiates a new instance of the CLI application.
+func NewCliApp() *cli.App {
+	app := cli.NewApp()
+	app.Name = "cadence"
+	app.Usage = "A command-line tool for cadence users"
+	app.Version = Version
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:   FlagAddressWithAlias,
+			Value:  "",
+			Usage:  "host:port for cadence frontend service",
+			EnvVar: "CADENCE_CLI_ADDRESS",
+		},
+		cli.StringFlag{
+			Name:   FlagDomainWithAlias,
+			Usage:  "cadence workflow domain",
+			EnvVar: "CADENCE_CLI_DOMAIN",
+		},
+	}
+	app.Commands = []cli.Command{
+		{
+			Name:        "domain",
+			Aliases:     []string{"d"},
+			Usage:       "Operate cadence domain",
+			Subcommands: newDomainCommands(),
+		},
+		{
+			Name:        "workflow",
+			Aliases:     []string{"wf"},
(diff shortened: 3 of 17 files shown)


common/cli/util.go
```diff
@@ -0,0 +1,193 @@
...
+		data = e.TimerCanceledEventAttributes
+
+	case s.EventTypeMarkerRecorded:
+		data = e.MarkerRecordedEventAttributes
+
+	case s.EventTypeWorkflowExecutionTerminated:
+		data = e.WorkflowExecutionTerminatedEventAttributes
+
+	default:
+		data = e
```
wxing1292: this list of events is not complete, at least missing the child workflow events

common/cli/factory.go
```diff
@@ -0,0 +1,188 @@
...
+		return Staging
+	}
+	if env == "production" {
+		return Production
+	}
+	return Development
+}
+
+// WorkflowClientBuilderInterface is an interface to build client to cadence service
+type WorkflowClientBuilderInterface interface {
```
madhuravi: Why is this file here? Where is it being used? Can't find any references in the diff.

common/cli/util.go
vancexu: Good catch, added all

common/cli/factory.go
vancexu: I previously thought common might be a good place, and later playground or other apps can use it. But now I think it might be better just put everything in cli folder for now, so changed. 
This interface is used in "commands.go", and I need it so that I can have a customized builder used internal. See update comments in the file.

tools/cli/app.go
```diff
@@ -0,0 +1,76 @@
...
+// NewCliApp instantiates a new instance of the CLI application.
+func NewCliApp() *cli.App {
+	app := cli.NewApp()
+	app.Name = "cadence"
+	app.Usage = "A command-line tool for cadence users"
+	app.Version = Version
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:   FlagAddressWithAlias,
+			Value:  "",
```
yiminc: I think we have default value of 127.0.0.1:7933, it is not the case?

tools/cli/app.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+		cli.StringFlag{
+			Name:   FlagDomainWithAlias,
+			Usage:  "cadence workflow domain",
+			EnvVar: "CADENCE_CLI_DOMAIN",
+		},
+		cli.BoolFlag{
+			Name:   FlagCustomizeBuilderWithAlias,
+			Usage:  "customize builder function or default",
+			EnvVar: "CADENCE_CUSTOMIZED_BUILDER",
```
yiminc: this flag is confusing. How is user suppose to use this flag?

tools/cli/commands.go
```diff
@@ -0,0 +1,871 @@
...
+	defaultDomainRetentionDays = 3
+)
+
+// For color output to terminal
+var colorRed func(i ...interface{}) string
+var colorMagenta func(i ...interface{}) string
+var colorGreen func(i ...interface{}) string
+
+func init() {
+	colorRed = color.New(color.FgRed).SprintFunc()
```
yiminc: could you assign the value when you declare the variables so we don't need this init()?

tools/cli/commands.go
```diff
@@ -0,0 +1,871 @@
...
+	colorMagenta = color.New(color.FgMagenta).SprintFunc()
+	colorGreen = color.New(color.FgGreen).SprintFunc()
+}
+
+// cBuilder is used to create cadence clients if user want to provide customized builder
+var cBuilder WorkflowClientBuilderInterface
```
yiminc: I don't understand why do we need this?

tools/cli/commands.go
```diff
@@ -0,0 +1,871 @@
...
+	}
+	return fmt.Sprintf("cadence-cli@%s", hostName)
+}
+
+func newContext() (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.Background(), time.Second*5)
+}
+
+func newContextForLongPoll() (context.Context, context.CancelFunc) {
+	return context.WithTimeout(context.Background(), time.Minute*2)
```
yiminc: Do we need to let user specify timeout for this? (Make it optional, and default to 2 minutes.) For example, what if the workflow takes longer than 2 minutes?

tools/cli/app.go
vancexu: Default `localHostPort` is 127.0.0.1:7933. I put empty here is easier for processing.

tools/cli/app.go
vancexu: Agree it is confusing, removed.

tools/cli/commands.go
vancexu: changed

tools/cli/commands.go
vancexu: This is necessary for customized builder. Internally, we have special tag like ENV and additional processing when creating builder. By implementing these interfaces and inject with SetBuilder, we can achieve customized builder. This is transparent to CLI user.

tools/cli/commands.go
vancexu: Yes, changed.

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+./cadence-cli --domain samples-domain workflow start --tl helloWorldGroup --wt main.Workflow --et 60 --dt 10 -i '"vancexu"'
+
+# see more options for workflow start
+./cadence-cli --domain samples-domain workflow start -h
+
+# for workflow with multiple input, seperate each json with space/newline like
+./cadence-cli --domain samples-domain workflow start --tl helloWorldGroup --wt main.WorkflowWith3Args --et 60 --dt 10 -i '"your_input_string" 123 {"Name":"my-string", "Age":12345}'
+```
+
+- Run workflow and see the progress
```
mfateev: I think this one should come before start as most developers should use it during development. I think you also want to expand on what "run" means in this context and what benefits over start it has.

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+*Tips:*  
+to avoid repeated input global option **domain**, user can export domain-name in environment variable CADENCE_CLI_DOMAIN.
+(There are also other environment variables like this, see help message for details)
+```
+export CADENCE_CLI_DOMAIN=samples-domain
+
+# then you can just
+./cadence-cli domain desc
+./cadence-cli workflow list
+./cadence-cli workflow show -w greetings_38a9e481-a00c-433a-b1ff-b6f67536efd8
```
mfateev: Can we make -w optional after show? Just workflow show UUID?

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+Cadence CLI provide an command-line tool for users to perform various tasks on Cadence.  
+Users can perform operations like register, update and describe on domain;  
+also start workflow, show workflow history, signal workflow ... and many other tasks.    
+
+## How
+- Run `make bins`
+- You should see an executable `cadence-cli`
+
+## Quick Start
+Run `./cadence-cli` to view help message. There are some top level commands and global options.
```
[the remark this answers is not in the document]
vancexu: Yes internally user use `cadence` through brew install, but the runnable we generate is `./cadence-cli`. In public server repo, `cadence` is used as the runnable to start cadence server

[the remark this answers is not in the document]
vancexu: Maybe I can combine currently `cadence` and `cadence-cli` to one, and create commands like `cadence service start`? What do you think?
Otherwise we have to use `cadence-cli ...`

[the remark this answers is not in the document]
mfateev: Let's rename cadence to cadence-server and cadence-cli to cadence then

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Quick Start
+Run `./cadence-cli` to view help message. There are some top level commands and global options.   
+Run `./cadence-cli domain` to view help message about operations on domain  
+Run `./cadence-cli workflow` to view help message about operations on workflow  
+
+*Note:* make sure you have cadence server running before using CLI 
+
+### Domain operation examples 
+- Register a new domain named "samples-domain":
```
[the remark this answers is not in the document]
vancexu: Good suggestion, I will add DOMAIN env here also. 
File support will be done as a feature, together with cache workflow_id and run_id

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+### Domain operation examples 
+- Register a new domain named "samples-domain":  
+```
+./cadence-cli --domain samples-domain domain register 
+# OR using short alias  
+./cadence-cli --do samples-domain domain re
+```   
+- View "samples-domain" details:   
+```
+./cadence-cli --domain samples-domain domain describe
```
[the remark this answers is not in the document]
vancexu: I think "workflow describe` is a little bit confusing which looks like describe detail info of workflow such as workflow type, arguments, wid, runid etc, but `workflow show` is showing history.

tools/cli/factory.go
```diff
@@ -0,0 +1,144 @@
...
+const (
+	_cadenceClientName      = "cadence-client"
+	_cadenceFrontendService = "cadence-frontend"
+)
+
+// WorkflowClientBuilderInterface is an interface to build client to cadence service
+// Usually default WorkflowClientBuilder is enough for development purpose, but
+// user can still customized builder by implementing this interface, and call SetBulder after initialize cli.
+// The customized builder may have more processing on Env, Address and other info.
+type WorkflowClientBuilderInterface interface {
```
yiminc: I thought we talked about this interface, and decided that it only need one method:
CreateServiceClient(ctx *cli.Context). Or am I missing something?

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+### Workflow operation examples
+
+- Start workflow: (for example [this workflow](https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/recipes/helloworld/helloworld_workflow.go))
+```
+./cadence-cli --domain samples-domain workflow start --tl helloWorldGroup --wt main.Workflow --et 60 --dt 10 -i '"vancexu"'
+
+# see more options for workflow start
+./cadence-cli --domain samples-domain workflow start -h
```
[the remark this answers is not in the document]
vancexu: No, `cadence workflow start help` won't work because there is no `help` subsubcommands, but only `--help` or `-h` command options. 

When view help of `cadence workflow` it will show below:
```
NAME:
   cadence workflow - Operate cadence workflow

USAGE:
   cadence workflow command [command options] [arguments...]

COMMANDS:
     show             show workflow history
     start            start a new workflow execution
     run              start a new workflow execution and get workflow progress
     cancel, c        cancel a workflow execution
     signal, s        signal a workflow execution
     terminate, term  terminate a new workflow execution
     list             list open or closed workflow executions
     query            query workflow execution

OPTIONS:
   --help, -h  show help

```
the usage will indicate how the help option can be used

tools/cli/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+# for workflow with multiple input, seperate each json with space/newline like
+./cadence-cli --domain samples-domain workflow start --tl helloWorldGroup --wt main.WorkflowWith3Args --et 60 --dt 10 -i '"your_input_string" 123 {"Name":"my-string", "Age":12345}'
+```
```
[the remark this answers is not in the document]
vancexu: Unfortunately we cannot. 
Because using your proposal there is no way to tell if `["your_input_string", 123, {"Name":"my-string", "Age":12345}]` input is 1 list argument or 3 arguments.

tools/cli/README.md
vancexu: changed

tools/cli/README.md
vancexu: With the cli framework we currently use, it would be a little bit tricky to achieve this requirement. Because `show` commands also accepts other optional arguments. 
Instead, I propose to add new command like `showid <wid>` as to separate command which can only take wid and have no other arguments to achieve this. what do you think?

tools/cli/README.md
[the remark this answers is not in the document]
vancexu: Yes, added

tools/cli/factory.go
vancexu: Changed, I missed one part previously. Thanks

tools/cli/README.md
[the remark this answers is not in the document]
vancexu: Renamed to cadence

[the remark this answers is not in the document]
vancexu: Done

tools/cli/README.md
```diff
@@ -0,0 +1,86 @@
...
+# then just run commands without --domain flag, like
+./cadence domain desc
+```
+
+### Workflow operation examples
+(The following examples assume you already export CADENCE_CLI_DOMAIN environment variable as Tips above)
+
+- Run workflow: Start a workflow and see it's progress 
+```
+./cadence workflow run --tl helloWorldGroup --wt main.Workflow --et 60 --dt 10 -i '"cadence"'
```
mfateev: Let's default decision task timeout to something like 5 seconds and do not require it.

tools/cli/README.md
```diff
@@ -0,0 +1,86 @@
...
+export CADENCE_CLI_DOMAIN=samples-domain
+
+# then just run commands without --domain flag, like
+./cadence domain desc
+```
+
+### Workflow operation examples
+(The following examples assume you already export CADENCE_CLI_DOMAIN environment variable as Tips above)
+
+- Run workflow: Start a workflow and see it's progress
```
mfateev: Add that this command doesn't finish until workflow completes.

tools/cli/README.md
```diff
@@ -0,0 +1,86 @@
...
+./cadence workflow show -w 3ea6b242-b23c-4279-bb13-f215661b4717 -r 866ae14c-88cf-4f1e-980f-571e031d71b0
+
+# if run_id is not provided, it will show the latest run history for that workflow_id
+./cadence workflow show -w 3ea6b242-b23c-4279-bb13-f215661b4717
+```
+
+- List open or closed workflow executions
+```
+./cadence workflow list
+```
```
mfateev: What about query?

tools/cli/README.md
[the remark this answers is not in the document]
mfateev: We could say that list always defaults to the list of arguments. If they indeed want to specify list as a single argument (which is very rare) then they have to pass list of lists:
[[...]].

cmd/tools/cli/main.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// Start using this CLI tool with command
+// `./cadence-cli` to show help message
+// `./cadence-cli domain` to show help message on domain operations
+// `./cadence-cli workflow` to show help message on workflow operations
+func main() {
+	app := cli.NewCliApp()
+
+	logger, err := zap.NewDevelopment()
```
yiminc: Can we not make this logger a required argument to NewBuilder()? It could create a default inside that builder. Just want to keep code here to be minimal.

cmd/tools/cli/main.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		panic(err)
+	}
+	cli.SetBuilder(cli.NewBuilder(logger))
```
yiminc: I would even make this call optional and use cli.NewBuilder() if it is not set.
Actually, I would check the builder inside the cli.NewCliApp() and set it to the default value if it is not set yet. So in internal version, you can call the cli.SetBuilder() before cli.NewCliApp().

tools/cli/README.md
vancexu: Done, default 10 sec

tools/cli/README.md
vancexu: Added

tools/cli/README.md
vancexu: Added. Need to add samples to make it looks better later

cmd/tools/cli/main.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"os"
+
+	"github.com/uber/cadence/tools/cli"
+	"go.uber.org/zap"
+)
+
+// Start using this CLI tool with command
+// `./cadence-cli` to show help message
```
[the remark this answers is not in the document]
vancexu: updated

tools/cli/factory.go
```diff
@@ -0,0 +1,107 @@
...
+	BuildServiceClient(c *cli.Context) (workflowserviceclient.Interface, error)
+}
+
+// WorkflowClientBuilder build client to cadence service
+type WorkflowClientBuilder struct {
+	hostPort   string
+	dispatcher *yarpc.Dispatcher
+	Logger     *zap.Logger
```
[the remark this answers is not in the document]
vancexu: Good catch, changed Logger to logger. 
WorkflowClientBuilder is needed, otherwise lint complain on NewBuilder function.

cmd/tools/cli/main.go
vancexu: Updated

cmd/tools/cli/main.go
vancexu: Make sense I like it. Updated.

tools/cli/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Query workflow execution
+```
+# use build-in query type "__stack_trace" which is supported by cadence client library
+./cadence workflow query -w <wid> -r <rid>
```
mfateev: What about 
cadence workflow stack -w ...

tools/cli/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+Example uses [this cadence-samples workflow](https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/recipes/helloworld/helloworld_workflow.go) 
+and it takes a string as input, so there is the `-i '"cadence"'`. Single quote `''` is used to wrap input as json. 
+
+**Note:** you need to start worker so that workflow can make progress.  
+(Run `make && ./bin/helloworld -m worker` in cadence-samples to start the worker)
+
+- Start workflow: 
+```
+./cadence workflow start --tl helloWorldGroup --wt main.Workflow --et 60 --dt 10 -i '"cadence"'
```
mfateev: I would change ./cadence to cadence in all these samples. In most cases it would be installed though brew or be somewhere in the PATH.

tools/cli/README.md
```diff
@@ -0,0 +1,108 @@
...
+- Query workflow execution
+```
+# use build-in query type "__stack_trace" which is supported by cadence client library
+./cadence workflow query -w <wid> -r <rid> 
+
+# use custom query type
+./cadence workflow query -w <wid> -r <rid> --qt <query-type>
+```
+
+- Signal, cancel, terminate workflow
```
mfateev: We have the open activities visibility query. Where is it?
Also we have list of workers connected to a task list. Where is it?

tools/cli/README.md
vancexu: I think query at least link me with QueryWorkflow of cadence-client, no sure why you propose stack?

tools/cli/README.md
vancexu: For this readme, I think `./cadence` is better because reader can easily understand it is the executable after running `make`, and "advanced player" can obviously use PATH and replace it to whatever they what. 
Directly put `cadence` instead of `./cadence` will confuse beginner, if I don't have instructions on how to set PATH or use brew. And for the simplicity of this readme, I don't want to include those.

tools/cli/README.md
vancexu: Create https://github.com/uber/cadence/issues/588 and will work on separate pull request.

tools/cli/README.md
mfateev: Because if you don't specify query type you default to stack which is confusing. I would require query type, but provide separate command "stack" which would execute query type "__stack_trace" which is hard to remember.

tools/cli/README.md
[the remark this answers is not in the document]
mfateev: Why is it resolved?

tools/cli/README.md
mfateev: If it is indeed that hard then showid is OK.

tools/cli/README.md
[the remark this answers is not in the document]
vancexu: After discussion, we decide to stick with using spaces/newlines separate json, instead of the proposed using list.

tools/cli/README.md
vancexu: Cool, added `showid <wid>` as a shortcut of `show -w <wid>`

tools/cli/README.md
vancexu: Added `stack` command

Issue #573, opened 2018-02-22
Query stops working after worker crash


samarabbas:
Query stops working after worker restart until new decision happens for execution.  Looks like query is continuously sent to sticky tasklist and after worker restart all queries will timeout until sticky tasklist is cleared.

Here is the example workflow definition:
```
func Workflow(ctx workflow.Context) error {
	queryResult := "started"
	logger := workflow.GetLogger(ctx)
	logger.Info("QueryWorkflow started")
	// setup query handler for query type "state"
	err := workflow.SetQueryHandler(ctx, "state", func(input []byte) (string, error) {
		return queryResult, nil
	})
	if err != nil {
		logger.Info("SetQueryHandler failed: " + err.Error())
		return err
	}

	queryResult = "waiting on timer"
	// to simulate workflow been blocked on something, in reality, workflow could wait on anything like activity, signal or timer
	workflow.NewTimer(ctx, time.Minute*10).Get(ctx, nil)
	logger.Info("Timer fired")

	queryResult = "done"
	logger.Info("QueryWorkflow completed")
	return nil
}
``` 

Once the workflow is started it immediately waits on the timer for 10 minutes.  Running the query returns the following result:
```
./cmd/cli/cadence-cli --domain samples-domain wf query -w query_752300dc-1569-43d4-b4a6-acff803a1b48 -qt state
Query result as JSON:
"waiting on timer"
```

Now crash the worker and restart it.  Now running the query returns the following result:
```
./cmd/cli/cadence-cli --domain samples-domain wf query -w query_752300dc-1569-43d4-b4a6-acff803a1b48 -qt state
Query failed: code:deadline-exceeded message:timeout
```

History of workflow at this point:
```
./cmd/cli/cadence-cli --domain samples-domain wf show -w query_752300dc-1569-43d4-b4a6-acff803a1b48
1, 2018-02-21T20:03:21-08:00, WorkflowExecutionStarted: (WorkflowType:(Name:code.uber.internal/devexp/cadence-samples.git/cmd/samples/recipes/query.Workflow), TaskList:(Name:query-workflow-sample), Input:[], ExecutionStartToCloseTimeoutSeconds:36000, TaskStartToCloseTimeoutSeconds:60, Identity:34439@samar-C02Q512VG8WM@)
2, 2018-02-21T20:03:21-08:00, DecisionTaskScheduled: (TaskList:(Name:query-workflow-sample), StartToCloseTimeoutSeconds:60, Attempt:0)
[...]

PR #579, opened 2018-02-23
bugfix: #573 sticky query should enforce sticky decision timeout

wxing1292:
feel free to ask me change the name of the new API on history side: ResetMutableState


idl/github.com/uber/cadence/history.thrift
```diff
@@ -208,6 +219,23 @@ service HistoryService {
...
+  /**
+  * Reset the volatile information in mutable state of a given workflow.
+  * Volatile information are the information related to client, such as:
+  * 1. StickyTaskList
+  * 2. StickyScheduleToStartTimeout
+  * 3. ClientLibraryVersion
+  * 4. ClientFeatureVersion
+  * 5. ClientImpl
+  **/
+  ResetMutableStateResponse ResetMutableState(1: ResetMutableStateRequest resetRequest)
```
yiminc: ResetMutableState sounds like it is going to reset the mutable state. It might be better to just call it ResetStickyTaskList.

service/frontend/handler.go
```diff
@@ -1511,15 +1513,38 @@ func (wh *WorkflowHandler) QueryWorkflow(ctx context.Context,
...
+		stickyDecisionTimeout := response.GetStickyTaskListScheduleToStartTimeout()
+		// using a clean new context in case customer provide a context which has
+		// a really short deadline, causing we clear the stickyness
+		stickyContext, cancel := context.WithTimeout(context.Background(), time.Duration(stickyDecisionTimeout)*time.Second)
+		matchingResp, err := wh.matching.QueryWorkflow(stickyContext, matchingRequest)
+		cancel()
+		if err == nil {
+			return matchingResp, nil
+		}
+		if yarpcError, ok := err.(*yarpcerrors.Status); !ok || yarpcError.Code() != yarpcerrors.CodeDeadlineExceeded {
```
yiminc: should it be !ok? Could you test this code path?

service/history/historyEngine.go
```diff
@@ -444,21 +444,51 @@ func (e *historyEngineImpl) getMutableState(
...
+		return nil, err
+	}
+	defer release()
+
+	msBuilder, err := context.loadWorkflowExecution()
+	if err != nil {
+		return nil, err
+	}
+
+	msBuilder.clearStickyness()
```
yiminc: where do we actually write to db? Could you verify that the mutable state is updated in db.

idl/github.com/uber/cadence/history.thrift
wxing1292: sure

service/frontend/handler.go
wxing1292: this is tested by TestQueryWorkflow_StickyTimeout in integ test

service/history/historyEngine.go
wxing1292: ideally, clearing the stickness need to be updated in DB side.
however, clearing the stickness is not an important operation, since the update does not contains workflow related information.

idl/github.com/uber/cadence/history.thrift
wxing1292: changed to ResetMutableStateStickyness

let me know if this name works

PR #582, opened 2018-02-26
rename domain version to db_version, move failover_version to top level

wxing1292:


schema/cadence/schema.cql
```diff
@@ -248,7 +247,8 @@ CREATE TABLE domains_by_name (
   domain             frozen<domain>,
   config             frozen<domain_config>,
   replication_config frozen<domain_replication_config>, -- indicating active cluster and standby cluster used for replication
-  version            bigint, --indicate the version of the record, used for update
+  failover_version   bigint,
```
madhuravi: Add a comment on what this means as well

PR #585, opened 2018-02-27
Kafka based publisher for replication tasks

samarabbas:


common/messaging/kafkaClient.go
```diff
@@ -65,3 +68,16 @@ func (c *kafkaClient) NewConsumer(topicName, consumerName string, concurrency in
 	consumer, err := c.client.NewConsumer(consumerConfig)
```
madhuravi: Fixing the interface issue with the latest kafka client as part of this?

common/messaging/kafkaClient.go
```diff
@@ -30,6 +32,7 @@ type (
 	kafkaClient struct {
 		config *KafkaConfig
 		client *kafkaclient.Client
```
madhuravi: I think all you need is to glide up and make this a non-pointer

common/messaging/kafkaClient.go
samarabbas: I will take care of it in a separate PR.

common/messaging/kafkaClient.go
samarabbas: I will take care of it in a separate PR.

PR #586, opened 2018-02-28
Replicate domain

wxing1292:
solve #569


service/frontend/handler.go
```diff
@@ -273,15 +300,69 @@ func (wh *WorkflowHandler) UpdateDomain(ctx context.Context,
...
+	configVersion := getResponse.ConfigVersion
 	failoverVersion := getResponse.FailoverVersion
+	currentCluster := clusterMetadata.GetCurrentClusterName()
 
 	// whether active cluster is changed
 	activeClusterChanged := false
 	// whether anything other than active cluster is changed
 	configurationChanged := false
+
+	validateReplicationConfig := func(existingDomain *persistence.GetDomainResponse,
```
samarabbas: nit: Why this validation is defined as a closure? Can we just have a validation function defined on the handler like we do for all different kind of validation?

service/frontend/handler.go
```diff
@@ -273,15 +300,69 @@ func (wh *WorkflowHandler) UpdateDomain(ctx context.Context,
...
+			}
+			replicationConfig.ActiveClusterName = currentCluster
+		}
+
+		// validate active cluster is also specified in all clusters
+		activeClusterInClusters := false
+		for _, cluster := range replicationConfig.Clusters {
+			if cluster.ClusterName == replicationConfig.ActiveClusterName {
+				activeClusterInClusters = true
+				break
```
samarabbas: one coding practice we follow is to always use label when breaking out of loops.

schema/cadence/versioned/v0.6/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
wxing1292: we have not release that yet, however, 0.5 is rolled out to staging.
we can definitely do some manual work on staging.
let me know your preference

service/frontend/handler.go
wxing1292: the reason i put this validation as a closure: this function is not reusable, by other logics.

but sure, i will make the change

service/frontend/handler.go
wxing1292: ok sure

schema/cadence/versioned/v0.6/manifest.json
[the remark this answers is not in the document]
wxing1292: fixed, changed to 0.5

PR #587, opened 2018-03-01
Create type functions and filter options for dynamic config

madhuravi:
1. Type conversions are complicated. Each type requires handling different scenarios and different ways of casting. Internal dynamic config client handles these already. Created type specific functions as part of the client interface so we can make use of this instead if re-implementing type conversions.

2. Earlier, I had functions specfic to the task list name. This would require a full set of type functions for task list and again for domain, and for the combination. Options made more sense in keeping this simple and flexible.

### DIFF
--- a/common/service/dynamicconfig/config.go
+++ b/common/service/dynamicconfig/config.go
@@ -43,8 +43,23 @@ func (c *Collection) logNoValue(key Key, err error) {
 	c.logger.Debugf("Failed to fetch key: %s from dynamic config with err: %s", key.String(), err.Error())
 }
 
+// PropertyFn is a wrapper to get property from dynamic config
+type PropertyFn func() interface{}
+
+// IntPropertyFn is a wrapper to get int property from dynamic config
+type IntPropertyFn func(opts ...FilterOption) int
+
+// FloatPropertyFn is a wrapper to get float property from dynamic config
+type FloatPropertyFn func(opts ...FilterOption) float64
+
+// DurationPropertyFn is a wrapper to get duration property from dynamic config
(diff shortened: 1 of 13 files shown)


common/service/dynamicconfig/config.go
```diff
@@ -21,111 +21,88 @@
...
+		}
+		return val
 	}
 }
 
 // GetBoolProperty gets property and asserts that it's an bool
-func (c *Collection) GetBoolProperty(key Key, defaultVal bool) func() bool {
-	return func() bool {
-		return c.GetProperty(key, defaultVal)().(bool)
+func (c *Collection) GetBoolProperty(key Key, defaultValue bool) func(opts ...FilterOption) bool {
```
samarabbas: Can we define types for return type of each kind of PropertyValue?  Like BoolPropertyValue in this case?

madhuravi: Done.

common/service/dynamicconfig/constants.go
```diff
@@ -38,10 +38,13 @@ const (
 
 var keys = []string{
 	"unknownKey",
-	_matchingTaskListRoot + "MinTaskThrottlingBurstSize",
-	_matchingTaskListRoot + "MaxTaskBatchSize",
-	_matchingTaskListRoot + "LongPollExpirationInterval",
-	_historyRoot + "LongPollExpirationInterval",
+	_matchingTaskListRoot + "minTaskThrottlingBurstSize",
```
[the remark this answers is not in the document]
madhuravi: I was thinking about it more in the concept of system components i.e. matching service -> task list manager. 
I see why a domain makes sense though, I can add that.

PR #590, opened 2018-03-02
wire replicator transmission to domain APIs

wxing1292:
solve #569


service/frontend/handler.go
```diff
@@ -117,6 +124,8 @@ func NewWorkflowHandler(
 		hSerializerFactory: persistence.NewHistorySerializerFactory(),
 		domainCache:        cache.NewDomainCache(metadataMgr, sVice.GetLogger()),
 		rateLimiter:        common.NewTokenBucket(config.RPS, common.NewRealTimeSource()),
+		kafkaProducer:      kafkaProducer,
```
samarabbas: Let's not use any Kafka naming at this layer.

service/worker/domainReplicationTaskHandler.go
```diff
@@ -65,8 +67,45 @@ func NewDomainReplicator(metadataManager persistence.MetadataManager, logger bar
 	}
 }
 
-// handleDomainReplicationTask handle the domain replication task
-func (domainReplicator *domainReplicatorImpl) HandleReceiveTask(task *replicator.DomainTaskAttributes) error {
+// HandleTransmissionTask handle transmission of the domain replication task
+func (domainReplicator *domainReplicatorImpl) HandleTransmissionTask(kafka messaging.Producer, domainOperation replicator.DomainOperation,
```
samarabbas: Let's move this logic into a separate component which lives within frontend.

wxing1292: done

service/frontend/handler.go
wxing1292: reduced to just the domainReplicator dependency

PR #598, opened 2018-03-07
Add RecordActivityTaskHeartbeatByID

vancexu:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -871,6 +871,15 @@ struct RecordActivityTaskHeartbeatRequest {
   30: optional string identity
 }
 
+struct RecordActivityTaskHeartbeatByIDRequest {
+  10: optional string domain
+  20: optional string workflowID
```
wxing1292: do we conventionally use gen.WorkflowExecution?

vancexu: you are right, I will follow other XXByID for now, and change everywhere not use WorkflowExecution later

PR #600, opened 2018-03-08
history service should do event reordering making sure corresponding …

wxing1292:
…events for decision will have exactly the same order and no irrelevant event will be inserted in between, so client can predict the event ID of a corresponding decision.

tested using canary

counterpart: https://github.com/uber/cadence/pull/601


wxing1292:
close in favor of #601

PR #602, opened 2018-03-09
1. separate timer in timer queue processor into dedicated file, add UT

wxing1292:
2. send visibility time through channel instead setting minPendingTimer
3. increase history cache size from 256 to 512, note, this value was previously 1024


samarabbas:
Let's hold onto this change until we cut a new release.

service/history/timerQueueProcessor.go
```diff
@@ -171,7 +87,7 @@ func newTimerQueueProcessor(shard ShardContext, historyService *historyEngineImp
 		cache:            historyService.historyCache,
 		executionManager: executionManager,
 		shutdownCh:       make(chan struct{}),
-		newTimerCh:       make(chan struct{}, 1),
+		newTimerCh:       make(chan time.Time, shard.GetConfig().TimerProcessorTimerChanSize),
```
samarabbas: we always do non blocking write to the channel.  Why do you need a buffer?

service/history/timerQueueProcessor.go
```diff
@@ -238,16 +151,12 @@ func (t *timerQueueProcessorImpl) NotifyNewTimer(timerTasks []persistence.Task)
...
-	if updatedMinTimer {
-		select {
-		case t.newTimerCh <- struct{}{}:
-			// Notified about new timer.
-
-		default:
-			// Channel "full" -> drop and move on, since we are using it as an event.
-		}
+	select {
+	case t.newTimerCh <- newEarlistTime:
```
samarabbas: What happens if the earliest time is dropped due to non blocking write?

service/history/timer.go
```diff
@@ -0,0 +1,122 @@
...
+package history
+
+import (
+	"sync"
+	"time"
+)
+
+type (
+	// Timer interface
+	Timer interface {
```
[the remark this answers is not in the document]
wxing1292: ok

service/history/timer.go
```diff
@@ -0,0 +1,122 @@
...
+		// GetTimerChan return the channel which will be fired when time is up
+		GetTimerChan() <-chan bool
+		// WillFireAfter check will the timer get fired after a certain time
+		WillFireAfter(now time.Time) bool
```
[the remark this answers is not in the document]
wxing1292: changed to fireAfter. 
is after seems like comparing 2 time, and this function actually do "will this timer fire after time"

service/history/timer.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"sync"
+	"time"
+)
+
+type (
+	// Timer interface
+	Timer interface {
+		// GetTimerChan return the channel which will be fired when time is up
+		GetTimerChan() <-chan bool
```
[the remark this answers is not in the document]
wxing1292: sure

service/history/timerQueueProcessor.go
```diff
@@ -308,36 +217,31 @@ continueProcessor:
 				t.logger.Debug("Timer queue processor pump shutting down.")
 				return nil
 
-			case <-gateC:
+			case <-timerChan:
 				// Timer Fired.
 
-			case <-t.newTimerCh:
+			case newEarlistTime = <-t.newTimerCh:
```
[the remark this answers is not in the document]

service/history/timerQueueProcessor.go
wxing1292: changed back to use lock + variable for tracking min pending time

service/history/timerQueueProcessor.go
```diff
@@ -217,32 +221,33 @@ continueProcessor:
...
-			}
-		}
 
-		if !newEarlistTime.IsZero() {
-			t.metricsClient.IncCounter(metrics.TimerQueueProcessorScope, metrics.NewTimerNotifyCounter)
-			t.logger.Debugf("Woke up by the timer")
+			case <-t.newTimerCh:
+				t.newTimeLock.Lock()
+				newTime := t.newTime
+				t.newTime = time.Time{}
```
samarabbas: can we define a null time const and set it to null time instead of creating a new struct each time.

wxing1292: const can only be character, string, boolean, or numeric values.

i can change the variable type to *time.Time

PR #605, opened 2018-03-12
bugfix: timer processing logic

wxing1292:
@samarabbas maybe we can include this patch in the release

### DIFF
--- a/service/history/shardContext.go
+++ b/service/history/shardContext.go
@@ -425,15 +425,34 @@ func (s *shardContextImpl) updateShardInfoLocked() error {
 
 func (s *shardContextImpl) allocateTimerIDsLocked(timerTasks []persistence.Task) error {
 	// assign IDs for the timer tasks. They need to be assigned under shard lock.
+
+	// set a time limit, any timer to be fired before this time limit
+	// will be reset to this time limit, for more detaile, see below.
+	timeLimit := time.Now()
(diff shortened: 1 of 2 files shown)


service/history/shardContext.go
```diff
@@ -425,15 +425,34 @@ func (s *shardContextImpl) updateShardInfoLocked() error {
...
 func (s *shardContextImpl) allocateTimerIDsLocked(timerTasks []persistence.Task) error {
 	// assign IDs for the timer tasks. They need to be assigned under shard lock.
+
+	// set a time limit, any timer to be fired before this time limit
+	// will be reset to this time limit, for more detaile, see below.
+	timeLimit := time.Now()
+	if timeLimit.Before(s.shardInfo.TimerAckLevel) {
+		timeLimit = s.shardInfo.TimerAckLevel
+	}
+	timeLimit.Add(s.config.TimerCreationTolerance)
```
samarabbas: Looks like we will artificially add a delay of 10 seconds in that case.  This may cause us to miss the SLA on firing timers.  Let's think more how we can fix this without causing this artificial delay.

wxing1292: ok, this is a short time fix, which can help us reducing the chance when we miss timers.
the long term solution which i can come up with also will introduce some artificially delay.

wxing1292:
keep this PR open, until we finish the cross DC work and then refactor the timer logic.

PR #606, opened 2018-03-12
CLI: make show,list workflow looks better

vancexu:
1. Add color 2. remove datetime by default, can add using -pdt (print date time) flag

screen shot 2018-03-12 at 11 45 29 am
screen shot 2018-03-12 at 11 42 38 am

### DIFF
--- a/tools/cli/README.md
+++ b/tools/cli/README.md
@@ -90,9 +90,12 @@ User need to run `show` to view workflow history/progress.
 ./cadence workflow showid 3ea6b242-b23c-4279-bb13-f215661b4717
 ```
 
-- List open or closed workflow executions
+- List closed or open workflow executions
 ```
 ./cadence workflow list
+
+# default will only show one page, to view more items, use --more flag
+./cadence workflow list -m
 ```
 
 - Query workflow execution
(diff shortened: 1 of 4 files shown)


vancexu:
Change list cmd default to print one page (default 10 items)
Output in table and color
screen shot 2018-03-12 at 4 27 22 pm
screen shot 2018-03-12 at 4 27 37 pm

mfateev:
I think for the list we want to print one page per default. But if flag is specified just stream all the pages (with appropriate throttling). But if someone does

cadence workflow list -m|more

then block pagination when buffer is full. See http://www.pixelbeat.org/programming/stdio_buffering/

mfateev:
Can we use shorter timestamps by default?

vancexu:
@mfateev I see what you mean. Note currently open/close worklfows can not be grabbed as stream, but as page. 
The major difference between what you describe and my currently implementation is:  you have a stream somewhere and then a "big" buffer so user can see a lot items each time they press button; but my is some number of items in one page (default 10), and user may need to press many times.
I can change the -m mode default using large page size 100, but then user cannot use their own page size (with -pagesize flag) for more mode. How does this sound to you? I personally prefer the current way

For shorter timestamps, what's your proposal? only time without date? I agree them take spaces but have not better way to deal with it.

vancexu:
Add command `listall`, so that user can use pipe or redirect like`cadence workflow listall | more`

PR #607, opened 2018-03-12
bugfix: parent workflow, when signaling child workflow, can experienc…

wxing1292:
…e deadlock, if child workflow, at the same time, is completed.

### DIFF
--- a/service/history/transferQueueProcessor.go
+++ b/service/history/transferQueueProcessor.go
@@ -328,12 +328,12 @@ func (t *transferQueueProcessorImpl) processActivityTask(task *persistence.Trans
 	if err != nil {
 		return err
 	}
+	defer release()
 
-	var mb *mutableStateBuilder
-	mb, err = context.loadWorkflowExecution()
+	var msBuilder *mutableStateBuilder
+	msBuilder, err = context.loadWorkflowExecution()
 	timeout := int32(0)
 	if err != nil {
-		release()
 		if _, ok := err.(*workflow.EntityNotExistsError); ok {
 			// this could happen if this is a duplicate processing of the task, and the execution has already completed.
 			return nil
@@ -341,13 +341,13 @@ func (t *transferQueueProcessorImpl) processActivityTask(task *persistence.Trans
 		return err
 	}
 
-	if ai, found := mb.GetActivityInfo(task.ScheduleID); found {
+	if ai, found := msBuilder.GetActivityInfo(task.ScheduleID); found {
 		timeout = ai.ScheduleToStartTimeout
 	} else {
 		logging.LogDuplicateTransferTaskEvent(t.logger, persistence.TransferTaskTypeActivityTask, task.TaskID, task.ScheduleID)
 	}
-	release()
 
+	release()
 	if timeout != 0 {
 		err = t.matchingClient.AddActivityTask(nil, &m.AddActivityTaskRequest{
 			DomainUUID:                    common.StringPtr(targetDomainID),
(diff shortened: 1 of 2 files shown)


service/history/transferQueueProcessor.go
```diff
@@ -328,26 +328,26 @@ func (t *transferQueueProcessorImpl) processActivityTask(task *persistence.Trans
 	if err != nil {
 		return err
 	}
+	defer release()
```
samarabbas: Can you immediate call 'defer release()' after acquiring the release object?

service/history/transferQueueProcessor.go
```diff
@@ -328,26 +328,26 @@ func (t *transferQueueProcessorImpl) processActivityTask(task *persistence.Trans
...
 
-	if ai, found := mb.GetActivityInfo(task.ScheduleID); found {
+	if ai, found := msBuilder.GetActivityInfo(task.ScheduleID); found {
 		timeout = ai.ScheduleToStartTimeout
 	} else {
 		logging.LogDuplicateTransferTaskEvent(t.logger, persistence.TransferTaskTypeActivityTask, task.TaskID, task.ScheduleID)
 	}
-	release()
 
+	release()
```
samarabbas: Can you add a comment on why calling the release here is important.

service/history/transferQueueProcessor.go
```diff
@@ -455,16 +456,30 @@ func (t *transferQueueProcessorImpl) processCloseExecution(task *persistence.Tra
...
+	completionEvent, _ := msBuilder.GetCompletionEvent()
+	parentDomainID := msBuilder.executionInfo.ParentDomainID
+	parentWorkflowID := msBuilder.executionInfo.ParentWorkflowID
+	parentRunID := msBuilder.executionInfo.ParentRunID
+	initiatedID := msBuilder.executionInfo.InitiatedID
+
+	workflowTypeName := msBuilder.executionInfo.WorkflowTypeName
+	workflowStartTimestamp := msBuilder.executionInfo.StartTimestamp.UnixNano()
+	workflowCloseTimestamp := msBuilder.getLastUpdatedTimestamp()
+	workflowStatue := getWorkflowExecutionCloseStatus(msBuilder.executionInfo.CloseStatus)
```
samarabbas: nit: typo

service/history/transferQueueProcessor.go
```diff
@@ -455,16 +456,30 @@ func (t *transferQueueProcessorImpl) processCloseExecution(task *persistence.Tra
 		return err
 	}
 
+	isChildWorkflow := msBuilder.hasParentExecution()
+	workflowCloseStatus := msBuilder.executionInfo.CloseStatus
+	completionEvent, _ := msBuilder.GetCompletionEvent()
```
samarabbas: We will unnecessarily deserialize the completionEvent even for execution which does not have a Parent.  Can you bring this back with the If check?

service/history/transferQueueProcessor.go
wxing1292: there can be cases when cache is full and context, release being nil, while err not being nil, doing defer release() before error checking can lead to null pointer issue

service/history/transferQueueProcessor.go
wxing1292: sure

service/history/transferQueueProcessor.go
wxing1292: fixed

service/history/transferQueueProcessor.go
wxing1292: sure

PR #615, opened 2018-03-14
bugfix: visibility consistency is not using local qurom

wxing1292:


common/persistence/cassandraVisibilityPersistence.go
```diff
@@ -230,7 +229,7 @@ func (v *cassandraVisibilityPersistence) ListOpenWorkflowExecutions(
 		request.DomainUUID,
 		domainPartition,
 		common.UnixNanoToCQLTimestamp(request.EarliestStartTime),
-		common.UnixNanoToCQLTimestamp(request.LatestStartTime)).Consistency(v.lowConslevel)
+		common.UnixNanoToCQLTimestamp(request.LatestStartTime))
```
samarabbas: Actually thinking about this, we should keep low consistency level for read queries on visibility for performance reasons.  We provide eventual consistency for visibility store.  So we should keep the low consistency read for performance reasons.

wxing1292: agree that we should use less Cassandra capacity, however, the canary test for visibility can (and it actually did) experience issue, i.e. workflow started but list open workflow return 0.

wxing1292: So, overall, it depends on what kind of SLA that we guarantee.

wxing1292:
The conclusion is visibility part only provide eventually consistency, close PR

PR #618, opened 2018-03-19
add helper functions in domain cache for events replication

wxing1292:


service/history/shardController.go
```diff
@@ -106,15 +109,15 @@ func newShardController(host *membership.HostInfo, resolver membership.ServiceRe
...
+	metadataMgr persistence.MetadataManager, executionMgrFactory persistence.ExecutionManagerFactory, clusterMetadata cluster.Metadata,
+	factory EngineFactory, host *membership.HostInfo, config *Config, logger bark.Logger, metricsClient metrics.Client) (*historyShardsItem, error) {
 
 	executionMgr, err := executionMgrFactory.CreateExecutionManager(shardID)
 	if err != nil {
 		return nil, err
 	}
 
-	domainCache := cache.NewDomainCache(metadataMgr, logger)
+	domainCache := cache.NewDomainCache(metadataMgr, clusterMetadata, logger)
```
samarabbas: Why do we have a cache per shard?  I think domain cache should be created for the history host rather than instantiated per shard.

common/cache/domainCache.go
```diff
@@ -50,34 +52,46 @@ type (
...
+		config            *persistence.DomainConfig
+		replicationConfig *persistence.DomainReplicationConfig
+		configVersion     int64
+		failoverVersion   int64
+		isGlobalDomain    bool
 		expiry            time.Time
 		sync.RWMutex
 	}
+
+	domainNotActiveErr struct {
```
samarabbas: Can we define this error on the thrift IDL for the service?

common/cache/domainCache.go
```diff
@@ -162,12 +179,83 @@ func (c *domainCache) getDomain(key, id, name string, cache Cache) (*domainCache
...
+}
+
+func (entry *domainCacheEntry) ShouldReplicateEvent() bool {
+	// frontend guarantee that the clusters always contains the active domain, so if the # of clusters is 1
+	// then we do not need to send out any events for replication
+	return entry.clusterMetadata.GetCurrentClusterName() == entry.replicationConfig.ActiveClusterName &&
+		entry.isGlobalDomain && len(entry.replicationConfig.Clusters) > 1
+}
+
+func (entry *domainCacheEntry) GetDomainNotActiveErr() *domainNotActiveErr {
```
samarabbas: Let's define this error as an helper somewhere in common.

service/history/shardController.go
wxing1292: the cache is implemented using just one mutex, as shown below, so if the lock is not a concern, i can do the update:
```
lru struct {
	mut      sync.Mutex
	byAccess *list.List
	byKey    map[interface{}]*list.Element
	maxSize  int
	ttl      time.Duration
	pin      bool
	rmFunc   RemovedFunc
}
```

common/cache/domainCache.go
wxing1292: the idea here is to generate the error if and only if the domain is not active; 
i can separate the generation of error, see below, in common package. 
```
newDomainNotActiveErr(entry.info.Name, entry.clusterMetadata.GetCurrentClusterName(), entry.replicationConfig.ActiveClusterName)
```

PR #621, opened 2018-03-20
Add SignalWithStart

vancexu:
Currently, Cadence support sending signals to running workflows using Client SingalWorkflow APIs. However, if workflow is not running, either not started or already closed, SignalWorkflow will return error. So if user wants to make sure a signal is delivered, user needs to catch the error, start the workflow and then send signal again. It’s not convenient for user, also it requires additional calls to Cadence.   
This CR create new API SignalWithStart to ensure signal delivery. 

- if workflow is running, it will signal success (same as existing SignalWorkflow behavior);
- If workflow is not running, it will restart that workflow and then signal; 
- If workflow is not found, it will start workflow using input args and then signal 

In this way, user can perform all these signal-start-signal in one call.


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -962,6 +962,22 @@ struct SignalWorkflowExecutionRequest {
   70: optional binary control
 }
 
+struct SignalWithStartWorkflowExecutionRequest {
```
wxing1292: out of curiosity, why not just add the signal part to existing start workflow API?

service/history/handler.go
```diff
@@ -470,8 +470,9 @@ func (h *Handler) StartWorkflowExecution(ctx context.Context,
 
 	response, err2 := engine.StartWorkflowExecution(wrappedRequest)
 	if err2 != nil {
-		h.updateErrorMetric(metrics.HistoryStartWorkflowExecutionScope, h.convertError(err2))
-		return nil, h.convertError(err2)
+		tmpErr := h.convertError(err2)
```
wxing1292: convertedErr?

service/history/historyEngine.go
```diff
@@ -1593,6 +1582,208 @@ func (e *historyEngineImpl) SignalWorkflowExecution(signalRequest *h.SignalWorkf
...
+
+	isBrandNew := false
+	prevRunID := ""
+	attempt := 0
+
+	context, release, err0 := e.historyCache.getOrCreateWorkflowExecution(domainID, execution)
+	if err0 != nil {
+		if _, ok := err0.(*workflow.EntityNotExistsError); ok {
+			isBrandNew = true
+			goto Start_And_Signal
```
wxing1292: why do we use goto?

service/history/historyEngine.go
```diff
@@ -1593,6 +1582,208 @@ func (e *historyEngineImpl) SignalWorkflowExecution(signalRequest *h.SignalWorkf
...
+	if request.TaskStartToCloseTimeoutSeconds == nil || *request.TaskStartToCloseTimeoutSeconds <= 0 {
+		return nil, &workflow.BadRequestError{Message: "Missing or invalid TaskStartToCloseTimeoutSeconds."}
+	}
+	execution = workflow.WorkflowExecution{
+		WorkflowId: request.WorkflowId,
+		RunId:      common.StringPtr(uuid.New()),
+	}
+
+	// Generate first decision task event.
+	taskList := *request.TaskList.Name
```
wxing1292: GetName() if possible

service/history/historyEngine.go
```diff
@@ -1593,6 +1582,208 @@ func (e *historyEngineImpl) SignalWorkflowExecution(signalRequest *h.SignalWorkf
...
+		TransactionID: 0,
+		FirstEventID:  startedEvent.GetEventId(),
+		Events:        serializedHistory,
+	})
+	if err != nil {
+		return nil, err
+	}
+	msBuilder.executionInfo.LastFirstEventID = startedEvent.GetEventId()
+
+	createWorkflow := func(isBrandNew bool, prevRunID string) (string, error) {
```
wxing1292: this func does not take the workflow ID reuse policy in consideration.
moreover, it is hard to actually maintain a second set of start workflow logic.
so could we actually merge the signal & start to start workflow API?

idl/github.com/uber/cadence/shared.thrift
vancexu: That's a decision after design meeting. We prefer creating new API (and related requests) for such new cases. e.g. completeActivityByID is new instead of reuse. 
But we still try best to reuse as much as possible.

service/history/historyEngine.go
mfateev: I think merging the implementation does make sense. But merging the APIs is not.

wxing1292: i believe, history side start && start & signal APIs can be merged since it is not public.

service/history/historyEngine.go
vancexu: Changed.

service/history/historyEngine.go
vancexu: Changed

service/history/historyEngine.go
vancexu: agree it might feel uncomfortable, removed.

service/frontend/handler.go
```diff
@@ -1362,6 +1362,70 @@ func (wh *WorkflowHandler) SignalWorkflowExecution(ctx context.Context,
...
+	signalWithStartRequest *gen.SignalWithStartWorkflowExecutionRequest) (*gen.StartWorkflowExecutionResponse, error) {
+	scope := metrics.FrontendSignalWithStartWorkflowExecutionScope
+	sw := wh.startRequestProfile(scope)
+	defer sw.Stop()
+
+	if ok, _ := wh.rateLimiter.TryConsume(1); !ok {
+		return nil, wh.error(createServiceBusyError(), scope)
+	}
+
+	if signalWithStartRequest.Domain == nil {
```
yiminc: do you need to check if Domain == ""?

service/frontend/handler.go
```diff
@@ -1362,6 +1362,70 @@ func (wh *WorkflowHandler) SignalWorkflowExecution(ctx context.Context,
...
+		return nil, wh.error(errDomainNotSet, scope)
+	}
+
+	if signalWithStartRequest.WorkflowId == nil || *signalWithStartRequest.WorkflowId == "" {
+		return nil, wh.error(&gen.BadRequestError{Message: "WorkflowId is not set on request."}, scope)
+	}
+
+	if signalWithStartRequest.SignalName == nil {
```
yiminc: check SignalName == ""

service/history/historyEngine.go
```diff
@@ -1593,6 +1582,207 @@ func (e *historyEngineImpl) SignalWorkflowExecution(signalRequest *h.SignalWorkf
...
+		if _, ok := err0.(*workflow.EntityNotExistsError); ok {
+			isBrandNew = true
+		} else {
+			return nil, err0
+		}
+	}
+
+	// Start workflow and signal
+	request := getStartRequest(sRequest)
+	if request.ExecutionStartToCloseTimeoutSeconds == nil || *request.ExecutionStartToCloseTimeoutSeconds <= 0 {
```
yiminc: maybe worth to reuse those parameter check with StartWorkflow()

service/frontend/handler.go
vancexu: Added. Will check and change all APIs in another request. https://github.com/uber/cadence/issues/581

service/frontend/handler.go
vancexu: added.

service/history/historyEngine.go
vancexu: good suggestion, changed.

PR #623, opened 2018-03-20
implement standby timer processor, which keeps standby timers which are fired

wxing1292:
implement standby timer processor, which keeps standby timers which are fired.
rename SequenceID -> TimerSequenceID

Note:
Handling current domain become active, i.e. redirect current timers to timer queue processor is in next PR.
Handling of other domain become active is TBD, probably need to rebuild timers for workflow being active in other cluster


wxing1292:
close this PR since this solution cannot handle host restart, should rely on mutable state instead

PR #624, opened 2018-03-21
Support for generating replication task on workflow execution updates

samarabbas:
History engine changes to generate replication task when IsGlobalDomain
is enabled.  MutableStateBuilder is also updated to load replication
state for the domain and have support for creating replication task on
updates along with relevant updates to mutable state.

Replication tasks are written to a separate queue for the shard.  This
requires persistence changes to write replication tasks on updates in
the same transaction and new API for reading and completing the task.
New AckLevel is needed for processing of replication tasks.  This
requires changes to shard crud on the persistence layer for reading and
writing the new ack level for replication task processing.

ClusterMetadata and MessagingClient are already exposed on the Service.
So passing through Service to all layers in History service which needs
access to both ClusterMetadata and MessagingClient.


common/mocks/ClusterMetadata.go
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: since metadataTestBase.go is provided, this mock does not do much, is it?

wxing1292:
looks like we have a dedicated replication task queue, does that means each host will start additional # of shard of go routine for acking & processing replication task?

common/persistence/dataInterfaces.go
```diff
@@ -73,7 +73,12 @@ const (
 	TransferTaskTypeCancelExecution
 	TransferTaskTypeStartChildExecution
 	TransferTaskTypeSignalExecution
-	TransferTaskTypeReplicationTask
+)
+
+// Types of replication tasks
+const (
+	ReplicationTaskTypeHistory = iota
+	ReplicationTaskTypeHeartbeat
```
wxing1292: is ReplicationTaskTypeHeartbeat in the next PR?

common/mocks/ClusterMetadata.go
samarabbas: I'm using it in some unit test to assert IsGlobalDomain API.

common/persistence/dataInterfaces.go
samarabbas: Yes.  This is for future purpose.

PR #626, opened 2018-03-21
bugfix: ringpop should not contain worker service before cross dc is …

wxing1292:
…public


wxing1292:
travis still has trouble with ringpop, close since this potential bugfix does not work

PR #630, opened 2018-03-23
separate timer queue ack manager in separate file, add functionality …

wxing1292:
…to timer queue processor to be cluster aware

rename SequenceID -> TimerSequenceID

Standby timer processing logic && domain failover timer ack manager scan DB for domain specific timer in next PR.

### DIFF
--- a/service/history/shardContext.go
+++ b/service/history/shardContext.go
@@ -115,13 +115,14 @@ func (s *shardContextImpl) GetTransferAckLevel() int64 {
 	s.RLock()
 	defer s.RUnlock()
 
-	// TODO change make this cluster input parameter
+	// TODO cluster should be an input parameter
 	cluster := s.GetService().GetClusterMetadata().GetCurrentClusterName()
-	// if can find corresponding ack level in the cluster to timer ack level map
+	// if we can find corresponding ack level
 	if ackLevel, ok := s.shardInfo.ClusterTransferAckLevel[cluster]; ok {
 		return ackLevel
 	}
 	// otherwise, default to existing ack level, which belongs to local cluster
+	// this can happen if you add more cluster
 	return s.shardInfo.TransferAckLevel
 }
 
@@ -135,7 +136,7 @@ func (s *shardContextImpl) UpdateTransferAckLevel(ackLevel int64) error {
 	s.Lock()
 	defer s.Unlock()
 
-	// TODO change make this cluster input parameter
+	// TODO cluster should be an input parameter
 	cluster := s.GetService().GetClusterMetadata().GetCurrentClusterName()
 	if cluster == s.GetService().GetClusterMetadata().GetCurrentClusterName() {
 		s.shardInfo.TransferAckLevel = ackLevel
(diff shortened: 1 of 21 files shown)


service/history/timerQueueAckMgr.go
```diff
@@ -0,0 +1,306 @@
...
+		shard         ShardContext
+		executionMgr  persistence.ExecutionManager
+		logger        bark.Logger
+		metricsClient metrics.Client
+		lastUpdated   time.Time
+		config        *Config
+
+		sync.Mutex
+		// outstanding tasks, key cluster name, value map of outstanding timer task -> finished (true)
+		clusterOutstandingTasks map[string]map[TimerSequenceID]bool
```
samarabbas: I think we should keep this component simple by not making it aware of all clusters.  We should have a separate instance of TimerQueueProcessor with its own AckManager per cluster.

service/history/shardContext.go
```diff
@@ -58,8 +58,8 @@ type (
 		GetConfig() *Config
 		GetLogger() bark.Logger
 		GetMetricsClient() metrics.Client
-		GetTimerAckLevel() time.Time
-		UpdateTimerAckLevel(ackLevel time.Time) error
+		GetTimerAckLevel(cluster string) time.Time
```
samarabbas: Now we are going to have too many components managing ack levels separately.  Currently each update is a write to cassandra, this is not going to scale once cross dc support is in production.  Let's put a TODO to accept updates from individual components and have a separate job which updates all the ack levels in a single write for the entire shard.

service/history/shardContext.go
```diff
@@ -121,6 +121,13 @@ func (s *shardContextImpl) GetTransferAckLevel() int64 {
 	s.RLock()
 	defer s.RUnlock()
 
+	// TODO change make this cluster input parameter
+	cluster := s.GetService().GetClusterMetadata().GetCurrentClusterName()
+	// if can find corresponding ack level in the cluster to timer ack level map
```
samarabbas: nit: fix comment

service/history/timerQueueProcessor.go
```diff
@@ -169,6 +156,12 @@ func (t *timerQueueProcessorImpl) NotifyNewTimers(timerTasks []persistence.Task)
 	}
 }
 
+// NotifyNewTimerEvents - Notify the processor about the new standby history events arrival.
+// This should be called each time new timer events arrives, otherwise timers maybe fired unexpected.
+func (t *timerQueueProcessorImpl) NotifyNewTimerEvents() {
```
samarabbas: I think we should have the same API for notifying new timers whether it is for active processor or standby processor.  Basically we should have a baseTimerQueueProcessor which has common logic and then activeTimerQueueProcessor and passiveTimerQueueProcessor with their own handling of TimerTasks.  But the way you notify both should be exactly the same.

service/history/shardContext.go
wxing1292: track in: https://github.com/uber/cadence/issues/634

service/history/shardContext.go
wxing1292: done

service/history/timerQueueProcessor.go
wxing1292: this will be done in the next PR

PR #631, opened 2018-03-24
Add retry in some frontend API

vancexu:
Oncall observed multiple pollForActivity/Decision error in short period which affects our SLA.

Error log shows as:
handler.go:385 PollForDecisionTask failed. TaskList: adjust_ledgerv2, Error: code:unknown message:received unknown error calling service: "cadence-matching", procedure: "MatchingService::PollForDecisionTask", err: dial tcp 10.189.33.24:31849: getsockopt: connection refused

We need to add retry logic in Frontend to mitigate such error affecting our SLA to client.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -42,6 +42,10 @@ const (
 	historyServiceOperationInitialInterval    = 50 * time.Millisecond
 	historyServiceOperationMaxInterval        = 10 * time.Second
 	historyServiceOperationExpirationInterval = 30 * time.Second
+
+	frontendServiceOperationInitialInterval    = 200 * time.Millisecond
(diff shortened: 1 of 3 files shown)


common/util.go
```diff
@@ -42,6 +42,10 @@ const (
 	historyServiceOperationInitialInterval    = 50 * time.Millisecond
 	historyServiceOperationMaxInterval        = 10 * time.Second
 	historyServiceOperationExpirationInterval = 30 * time.Second
+
+	frontendServiceOperationInitialInterval    = 50 * time.Millisecond
```
wxing1292: personal preference: 50 is too low, 200 maybe a good idea

PR #636, opened 2018-03-27
CLI: add descirbe workflow execution

vancexu:
image
image


mfateev:
Could you show a screenshot of a result that includes activities with a progress? The progress part of activities is the most important part of the result as it cannot be extracted from the history.

vancexu:
image

mfateev:
Nice!

PR #639, opened 2018-03-29
add standby timer processing logic, separate existing timer processing logic into active & standby

wxing1292:
* add standby timer processing logic, separate existing timer processing logic into active & standby
* rename timer gate to local timer gate, add remote timer gate
* add local time / remote time to shard context

Note: 
UT for standby timer processor in #648
DB change adding failover version in timer info in next PR
Actual failover for timer in #648


common/persistence/persistenceTestBase.go
```diff
@@ -168,10 +168,13 @@ func (s *TestBase) SetupWorkflowStoreWithOptions(options TestBaseOptions) {
...
 	s.ShardInfo = &ShardInfo{
-		ShardID:             shardID,
-		RangeID:             0,
-		TransferAckLevel:    0,
-		ReplicationAckLevel: 0,
+		ShardID:                 shardID,
+		RangeID:                 0,
+		TransferAckLevel:        0,
+		ReplicationAckLevel:     0,
+		TimerAckLevel:           time.Time{},
```
samarabbas: why do we need a separate TimerAckLevel.  Can we use the cluster map for current cluster too?

service/history/historyEngine.go
```diff
@@ -380,7 +380,7 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 	}
 
 	if err == nil {
-		e.timerProcessor.NotifyNewTimers(timerTasks)
+		e.timerProcessor.NotifyNewTimers(e.shard.GetService().GetClusterMetadata().GetCurrentClusterName(), timerTasks)
```
samarabbas: Can we just store the current cluster name on the historyEngine instead of getting each time from Cluster Metadata.

service/history/shardContext.go
```diff
@@ -522,6 +525,24 @@ func (s *shardContextImpl) GetTimeSource() common.TimeSource {
 	return common.NewRealTimeSource()
 }
 
+func (s *shardContextImpl) SetCurrentTime(cluster string, currentTime time.Time) {
```
samarabbas: Is this called under a lock?

service/history/timerGate.go
```diff
@@ -86,23 +115,25 @@ func NewTimerGate() TimerGate {
...
 // FireChan return the channel which will be fired when time is up
-func (timerGate *TimerGateImpl) FireChan() <-chan struct{} {
+func (timerGate *LocalTimerGateImpl) FireChan() <-chan struct{} {
 	return timerGate.fireChan
 }
 
 // FireAfter check will the timer get fired after a certain time
-func (timerGate *TimerGateImpl) FireAfter(now time.Time) bool {
+func (timerGate *LocalTimerGateImpl) FireAfter(now time.Time) bool {
+	timerGate.Lock()
```
samarabbas: Previously this was not under a lock.  What has changed?

common/persistence/persistenceTestBase.go
wxing1292: we can, however, when loading a ack level for a new cluster, say "cluster foo", we will use the 
ClusterTimerAckLeve["current cluster"] to initialize which seems to be weird.

service/history/historyEngine.go
wxing1292: sure

service/history/shardContext.go
wxing1292: nice catch

PR #640, opened 2018-03-29
Add unit tests for CLI commands

vancexu:


tools/cli/app_test.go
```diff
@@ -0,0 +1,388 @@
...
+func TestCLIAppSuite(t *testing.T) {
+	s := new(cliAppSuite)
+	suite.Run(t, s)
+}
+
+func (s *cliAppSuite) SetupSuite() {
+	s.app = NewCliApp()
+}
+
+func (s *cliAppSuite) SetupTest() {
```
yiminc: add TearDownTest() and verify mock calls are expected.

tools/cli/app_test.go
```diff
@@ -0,0 +1,388 @@
...
+func (s *cliAppSuite) TestDomainRegister() {
+	s.service.EXPECT().RegisterDomain(gomock.Any(), gomock.Any(), callOptions...).Return(nil)
+	err := s.app.Run([]string{"", "--do", domainName, "domain", "register"})
+	s.Nil(err)
+}
+
+func (s *cliAppSuite) TestDomainRegister_DomainExist() {
+	s.service.EXPECT().RegisterDomain(gomock.Any(), gomock.Any(), callOptions...).Return(&shared.DomainAlreadyExistsError{})
+	err := s.app.Run([]string{"", "--do", domainName, "domain", "register"})
+	s.Nil(err)
```
yiminc: this does not test anything, because func RegisterDomain(c *cli.Context) return nothing so it always will be nil.

tools/cli/app_test.go
vancexu: Thanks for point it out. Changed

tools/cli/app_test.go
vancexu: Understand, currently these unit tests do not test the "real output", will probably revisit in future.

PR #643, opened 2018-03-30
Apply replication history events to passive cluster

samarabbas:
Created history replicator which is invoked by the replicator for
processing of history replication tasks.  It processes the history
events from the replication task and make mutable state updates for each
event.  Once all events are processed it commits the entire update using
the workflowContext API used by rest of the stack.

Also mutableStateBuilder changes to apply mutable state changes using
the actual event.  This required some refactoring of mutableStateBuilder
to reuse as much code as possible between replicator and rest of the
stack.

History service has new API ReplicateEvents which is called by the
replicator to apply history events.

Current this change only works for the happy case and is not guarded by
version updates on the domain.
Replicator does not support out of order processing of history
replication tasks.


service/history/mutableStateBuilder.go
```diff
@@ -230,14 +229,14 @@ func (e *mutableStateBuilder) FlushBufferedEvents() error {
 	return nil
 }
 
-func (e *mutableStateBuilder) ApplyReplicationStateUpdates(failoverVersion int64) {
+func (e *mutableStateBuilder) ApplyReplicationStateUpdates(failoverVersion, lastEventID int64) {
 	e.replicationState.CurrentVersion = failoverVersion
```
yiminc: where is the logic that you prevent failover from version that is lower than current version?

service/history/mutableStateBuilder.go
```diff
@@ -1076,7 +1112,7 @@ func (e *mutableStateBuilder) AddDecisionTaskScheduleToStartTimeoutEvent(schedul
 
 	event := e.hBuilder.AddDecisionTaskTimedOutEvent(scheduleEventID, 0, workflow.TimeoutTypeScheduleToStart)
 
-	e.DeleteDecision()
```
yiminc: this means the client might never see the decision task with attempt==0. Maybe we need server to give additional explicit flag for client to decide if it should fail the decision with error message or it should silently not-response-and-let-it-timeout.

service/history/replicatorQueueProcessor.go
```diff
@@ -156,6 +156,7 @@ func (p *replicatorQueueProcessorImpl) CompleteTask(taskID int64) error {
 
 func (p *replicatorQueueProcessorImpl) getHistory(task *persistence.ReplicationTaskInfo) (*shared.History, error) {
 
+	p.logger.Warnf("Received replication task: %v", task)
```
yiminc: should not be warning

service/worker/processor.go
```diff
@@ -178,7 +184,19 @@ func (p *replicationTaskProcessor) worker(workerWG *sync.WaitGroup) {
 						p.logger.Debugf("Recieved domain replication task %v.", task.DomainTaskAttributes)
 						err = p.domainReplicator.HandleReceivingTask(task.DomainTaskAttributes)
 					case replicator.ReplicationTaskTypeHistory:
-						p.logger.Debugf("Recieved history replication task %v.", task.HistoryTaskAttributes)
+						p.logger.Warn("Recieved history replication task %v.", task.HistoryTaskAttributes)
```
yiminc: should not be warning

service/history/historyReplicator.go
```diff
@@ -0,0 +1,261 @@
...
+	if request == nil || request.History == nil || len(request.History.Events) == 0 {
+		return nil
+	}
+
+	domainID, err := getDomainUUID(request.DomainUUID)
+	if err != nil {
+		return err
+	}
+	execution := shared.WorkflowExecution{
+		WorkflowId: request.WorkflowExecution.WorkflowId,
```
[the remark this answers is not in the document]
yiminc: but you could use *request.WorkflowExecution right?

service/history/mutableStateBuilder.go
yiminc: Could you help answer this?

samarabbas: This change just implements the happy path.  I will be implementing that part of the replication protocol in a separate PR.

service/history/historyReplicator.go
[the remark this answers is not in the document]
samarabbas: You are right.  Done.

PR #648, opened 2018-04-02
implemented functionality in timer ack manager & timer processor base to allow failover

wxing1292:
Partially solve #565


service/history/historyEngineInterfaces.go
```diff
@@ -106,15 +106,23 @@ type (
 
 	timerQueueProcessor interface {
 		common.Daemon
-		NotifyNewTimers(timerTask []persistence.Task)
+		NotifyNewTimers(clusterName string, timerTask []persistence.Task)
+		SetCurrentTime(clusterName string, currentTime time.Time)
+	}
+
+	timerProcessor interface {
```
samarabbas: Naming is quite confusing between timerQueueProcessor and timerProcessor.

service/history/shardContext.go
```diff
@@ -536,6 +557,18 @@ func acquireShard(shardID int, svc service.Service, shardManager persistence.Sha
...
 	updatedShardInfo.Owner = owner
 
+	// initialize the cluster current time to be the same as ack level
+	standbyClusterCurrentTime := make(map[string]time.Time)
+	for clusterName := range svc.GetClusterMetadata().GetAllClusterFailoverVersions() {
+		if clusterName != svc.GetClusterMetadata().GetCurrentClusterName() {
+			if currentTime, ok := shardInfo.ClusterTimerAckLevel[clusterName]; ok {
+				standbyClusterCurrentTime[clusterName] = currentTime
+			} else {
+				standbyClusterCurrentTime[clusterName] = shardInfo.TimerAckLevel
```
samarabbas: Why are we using the current cluster ack level to initialize the standby cluster?

service/history/timerGate.go
```diff
@@ -35,14 +35,19 @@ type (
 		// Update update the timer gate, return true if update is a success
 		// success means timer is idle or timer is set with a sooner time to fire
 		Update(nextTime time.Time) bool
+	}
+
+	// LocalTimerGate interface
+	LocalTimerGate interface {
```
samarabbas: Any reason Close is not needed by remote timer gate?  Both local and remote timers should have the same life-cycle.

service/history/timerGate.go
```diff
@@ -117,6 +148,65 @@ func (timerGate *TimerGateImpl) Update(nextTime time.Time) bool {
 }
 
 // Close shutdown the timer
-func (timerGate *TimerGateImpl) Close() {
+func (timerGate *LocalTimerGateImpl) Close() {
 	close(timerGate.closeChan)
 }
+
+// NewRemoteTimerGate create a new timer gate instance
+func NewRemoteTimerGate() RemoteTimerGate {
```
samarabbas: Lot of implementation between Local and Remote looks the same.  Can we refactor it in a way to reuse code?

I'm OK with copy/pasting code if this is temporary.  But make sure create an issue to revisit this later.

service/history/shardContext.go
wxing1292: there will be case that when a new cluster is added, there is no existing ack level for it, so we have to use the current cluster's ack level

service/history/historyEngineInterfaces.go
wxing1292: any suggestion?

service/history/timerGate.go
wxing1292: local is implemented using a go timer, which can be closed.

remote is purely driven by events, and there is no background coroutine  involved.

service/history/timerGate.go
wxing1292: this is actually not a copy paste, although in general they looks the same, however they are backed by different thing, one with go's timer, one with comparison of new time.

service/history/timerQueueProcessor.go
```diff
@@ -93,6 +104,13 @@ func (t *timerQueueProcessorImpl) SetCurrentTime(clusterName string, currentTime
 	standbyTimerProcessor.setCurrentTime(currentTime)
 }
 
+func (t *timerQueueProcessorImpl) FailoverDomain(domainID string, standbyClusterName string) {
+	// we should consider make the failover idempotent
+	failoverTimerProcessor := newTimerQueueFailoverProcessor(t.shard, t.historyService, domainID, standbyClusterName, t.logger)
```
samarabbas: how is the life-cycle of this cursor managed?  Someone needs to call stop on it? or it calls stop by itself?

service/history/timerQueueProcessorBase.go
```diff
@@ -224,7 +223,7 @@ func (t *timerQueueProcessorBase) internalProcessor() error {
 
 continueProcessor:
 	for {
-		now := t.shard.GetCurrentTime(t.clusterName)
+		now := t.now()
```
samarabbas: why?

service/history/timerQueueStandbyProcessor.go
```diff
@@ -40,32 +38,27 @@ type (
...
 	}
 )
 
-func newTimerQueueStandbyProcessor(shard ShardContext, historyService *historyEngineImpl, executionManager persistence.ExecutionManager, clusterName string, logger bark.Logger) *timerQueueStandbyProcessorImpl {
-	log := logger.WithFields(bark.Fields{
-		logging.TagWorkflowComponent: logging.TagValueTimerQueueComponent,
-	})
-	timerQueueAckMgr := newTimerQueueAckMgr(shard, historyService.metricsClient, executionManager, clusterName, log)
+func newTimerQueueStandbyProcessor(shard ShardContext, historyService *historyEngineImpl, clusterName string, logger bark.Logger) *timerQueueStandbyProcessorImpl {
+	timeNow := func() time.Time {
```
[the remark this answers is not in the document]
wxing1292: i know, right now, there is no time delay in both timer processor and transfer processor, will be added (change is fairly small) when we are actually doing failover on both timer / transfer

service/history/timerQueueAckMgr.go
```diff
@@ -287,6 +355,33 @@ func (t *timerQueueAckMgrImpl) getTimerTasks(minTimestamp time.Time, maxTimestam
...
+
 func (t *timerQueueAckMgrImpl) isProcessNow(expiryTime time.Time) bool {
-	return !expiryTime.IsZero() && expiryTime.UnixNano() <= t.shard.GetCurrentTime(t.clusterName).UnixNano()
+	var now time.Time
+	if !t.isFailover {
+		// not failover, use the cluster's local time
+		now = t.shard.GetCurrentTime(t.clusterName)
+	} else {
+		// if ack manager is a failover manager, we need to use the current local time
+		now = t.shard.GetCurrentTime(t.shard.GetService().GetClusterMetadata().GetCurrentClusterName())
```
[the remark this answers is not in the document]
wxing1292: not really, when doing failover, timer ack manager has 
cluster : standby cluster name
isFailover: true
and ack manager need to use the the above 2 to "update the shard timer ack level for failover process"

and here, when doing failover, we need to get the current active cluster's time to determine whether to process a timer

service/history/timerQueueProcessor.go
wxing1292: timer ack manager will have a finished chan which used by timer active processor, when chan fired, active processor will stop it self

service/history/timerQueueProcessorBase.go
wxing1292: i basically change the "t.shard.GetCurrentTime(t.clusterName)" to a closure

service/history/timerQueueProcessor.go
```diff
@@ -104,3 +146,85 @@ func (t *timerQueueProcessorImpl) getTimerFiredCount(clusterName string) uint64
 	}
 	return standbyTimerProcessor.getTimerFiredCount()
 }
+
+func (t *timerQueueProcessorImpl) completeTimersLoop() {
```
samarabbas: Let's create a separate task to add emit some metric from this new logic.  I think we are missing key metric which would help us investigate issues in production.

service/history/timerQueueStandbyProcessor.go
```diff
@@ -169,12 +197,13 @@ func (t *timerQueueStandbyProcessorImpl) processExpiredUserTimer(timerTask *pers
 				//
 				// we do not need to notity new timer to base, since if there is no new event being replicated
 				// checking again if the timer can be completed is meaningless
-				t.timerQueueAckMgr.retryTimerTask(timerTask)
-				return nil
+				return newTaskRetryError()
```
samarabbas: instead of creating a new error each time lets just define this as a global variable and reuse the same error.

common/persistence/cassandraPersistence.go
```diff
@@ -2037,9 +2037,6 @@ func (d *cassandraPersistence) GetTimerIndexTasks(request *GetTimerIndexTasksReq
 
 		response.Timers = append(response.Timers, t)
 	}
-	nextPageToken := iter.PageState()
```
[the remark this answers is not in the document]
wxing1292: because the existing query is using `limit`
https://github.com/uber/cadence/blob/master/common/persistence/cassandraPersistence.go#L546

we can change to use the pagination later, this is not a blocking issue

service/history/historyEngineInterfaces.go
```diff
@@ -44,6 +44,11 @@ type (
 		isWorkflowRunning bool
 		timestamp         time.Time
 	}
+
+	// error which will be thrown if the timer / transfer task should be
+	// retries due to various of reasons
+	taskRetryError struct{}
```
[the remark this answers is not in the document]
wxing1292: nop, i need to use a dedicated error so the worker can do retry accordingly.

service/history/shardContext.go
```diff
@@ -51,6 +51,10 @@ type (
 		UpdateTransferAckLevel(ackLevel int64) error
 		GetReplicatorAckLevel() int64
 		UpdateReplicatorAckLevel(ackLevel int64) error
+		GetTimerAckLevel() time.Time
```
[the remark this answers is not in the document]
wxing1292: when we do scan from DB, we always scan from the current ack level (inclusive) (time.Time) to int64max.
https://github.com/uber/cadence/blob/master/common/persistence/cassandraPersistence.go#L577
so, each time, when we do a scan from DB, we will not lose timers.

service/history/timerQueueAckMgr.go
```diff
@@ -39,23 +39,31 @@ type (
...
 		executionMgr  persistence.ExecutionManager
 		logger        bark.Logger
 		metricsClient metrics.Client
 		lastUpdated   time.Time
 		config        *Config
+		// immutable max possible timer level
+		maxAckLevel time.Time
+		// isReadFinished indicate timer queue ack manager
+		// have no more task to send out
+		isReadFinished bool
```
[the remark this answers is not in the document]
wxing1292: the finishedChan is used by the workers to see whether to stop.
when the read is finished, the worker still have to process the existing tasks in read, which can take time.

service/history/timerQueueAckMgr.go
```diff
@@ -80,43 +90,73 @@ func (t timerSequenceIDs) Less(i, j int) bool {
 	return compareTimerIDLess(&t[i], &t[j])
 }
 
-func newTimerQueueAckMgr(shard ShardContext, metricsClient metrics.Client, executionMgr persistence.ExecutionManager, clusterName string, logger bark.Logger) *timerQueueAckMgrImpl {
-	config := shard.GetConfig()
-	ackLevel := shard.GetTimerAckLevel(clusterName)
+func newTimerQueueAckMgr(shard ShardContext, metricsClient metrics.Client, clusterName string, logger bark.Logger) *timerQueueAckMgrImpl {
+	ackLevel := TimerSequenceID{VisibilityTimestamp: shard.GetTimerClusterAckLevel(clusterName)}
```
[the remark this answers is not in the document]
wxing1292: TaskID in by default 0, which is the minimum possible.
Moreover, we do not store the TaskID in DB.

service/history/timerQueueAckMgr.go
```diff
@@ -126,9 +166,12 @@ func (t *timerQueueAckMgrImpl) readTimerTasks() ([]*persistence.TimerTaskInfo, *
 	var lookAheadTask *persistence.TimerTaskInfo
 	t.Lock()
 	defer t.Unlock()
+	if t.isFailover && !morePage {
+		t.isReadFinished = true
+	}
+
 	// fillin the retry task
```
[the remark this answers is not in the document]
wxing1292: good catch

service/history/timerQueueProcessor.go
```diff
@@ -22,49 +22,84 @@ package history
...
+	}
 	t.activeTimerProcessor.Start()
-	for _, standbyTimerProcessor := range t.standbyTimerProcessors {
-		standbyTimerProcessor.Start()
+	if t.isGlobalDomainEnabled {
+		for _, standbyTimerProcessor := range t.standbyTimerProcessors {
+			standbyTimerProcessor.Start()
+		}
 	}
+	go t.completeTimersLoop()
```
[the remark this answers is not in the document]
wxing1292: well, the whole flow is changed.
if you really want, i can certainly do something, like adding a lot of if else.

service/history/timerQueueProcessor.go
```diff
@@ -104,3 +146,85 @@ func (t *timerQueueProcessorImpl) getTimerFiredCount(clusterName string) uint64
...
+		return nil
+	}
+
+	executionMgr := t.shard.GetExecutionManager()
+	minTimestamp := lowerAckLevel.VisibilityTimestamp
+	batchSize := t.config.TimerTaskBatchSize
+
+LoadCompleteLoop:
+	for {
+		request := &persistence.GetTimerIndexTasksRequest{
```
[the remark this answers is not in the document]
wxing1292: nop, that will not happen, let us sync.

service/history/timerQueueProcessor.go
```diff
@@ -104,3 +146,85 @@ func (t *timerQueueProcessorImpl) getTimerFiredCount(clusterName string) uint64
...
+func (t *timerQueueProcessorImpl) completeTimersLoop() {
+	timer := time.NewTimer(t.config.TimerProcessorCompleteTimerInterval)
+	defer timer.Stop()
+	for {
+		select {
+		case <-t.shutdownChan:
+			// before shutdown, make sure the ack level is up to date
+			t.completeTimers()
+			return
+		case <-timer.C:
```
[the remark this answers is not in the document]
wxing1292: this will not be a top priority issue, let us solve this later

service/history/timerQueueProcessor.go

service/history/timerQueueProcessor.go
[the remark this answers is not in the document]
wxing1292: https://github.com/uber/cadence/issues/679

service/history/timerQueueProcessor.go
wxing1292: https://github.com/uber/cadence/issues/680

service/history/timerQueueStandbyProcessor.go
wxing1292: ok

PR #651, opened 2018-04-03
create separate ack manager for transfer queue, for failover, as well as ack manager base

wxing1292:
UT in next PR

### DIFF
--- a/service/history/queueAckMgrBase.go
+++ b/service/history/queueAckMgrBase.go
@@ -1,0 +1,81 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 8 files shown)


service/history/queueProcessor.go
```diff
@@ -355,7 +355,8 @@ MoveAckLevelLoop:
 	}
 
 	// Always update ackLevel to detect if the shared is stolen
-	if err := a.shard.UpdateTransferAckLevel(updatedAckLevel); err != nil {
+	clusterName := a.shard.GetService().GetClusterMetadata().GetCurrentClusterName()
```
samarabbas: Can we store the clusterName when queueProcessor is created?

service/history/queueAckMgrBase.go
```diff
@@ -0,0 +1,81 @@
...
+		t.outstandingTasks[queueTask.GetTaskID()] = false
+	}
+}
+
+func (t *queueAckMgrBase) completeTask(queueTask queueTaskInfo) {
+	t.outstandingTasks[queueTask.GetTaskID()] = true
+}
+
+func (t *queueAckMgrBase) updateAckLevel(maxAckLevel int64) int64 {
+	t.metricsClient.IncCounter(metrics.TimerQueueProcessorScope, metrics.AckLevelUpdateCounter)
```
samarabbas: TimerQueueProcessorScope?

service/history/queueAckMgrBase.go
```diff
@@ -0,0 +1,81 @@
...
+
+	initialAckLevel := t.ackLevel
+	updatedAckLevel := t.ackLevel
+
+MoveAckLevelLoop:
```
samarabbas: original ackLevel does this within a lock?

service/history/queueAckMgrBase.go
```diff
@@ -0,0 +1,81 @@
...
+	for current := initialAckLevel + 1; current <= maxAckLevel; current++ {
+		acked := t.outstandingTasks[current]
```
samarabbas: ackLevels can have holes.  This will not work if we have holes in the task ids.

wxing1292: good catch

service/history/queueAckMgrBase.go
wxing1292: everything which use this base will have its dedicated lock.

service/history/queueAckMgrBase.go
wxing1292: ok

PR #657, opened 2018-04-05
Add CLI show workflow progress

vancexu:


tools/cli/commands.go
```diff
@@ -451,17 +453,31 @@ func RunWorkflow(c *cli.Context) {
 	table.AppendBulk(executionData) // Add Bulk Data
 	table.Render()
 
-	// print workflow progress
+	printWorkflowProgress(c, wid, resp.GetRunId())
+}
+
+// helper function to print workflow progress with time refresh every second
```
mfateev: I'm confused about the refresh every second. Why do we need to set timeout to one second for long poll? Wouldn't it return immediately if there is a new information even with a large timeout?

vancexu: This is just for UI purpose. We want to give user a feeling something is still running even when there is no new history comming. So currently UI dynamically show elapsed time in seconds and the number change every second.

PR #658, opened 2018-04-05
Bugfix: Recreate activity heartbeat timeout after first timer fire

samarabbas:
When the heartbeat timeout is created for the very first time it 
uses StartEventID as the dedupe event ID.  But it is possible that
it is a buffered event and no ID is specified when the timer task
is created.  This trips the recreation logic from recreating the 
timer. 
This change instead relies on the schedule ID for the event
for dedupe event ID of the heartbeat timer.

### DIFF
--- a/service/history/timerBuilder.go
+++ b/service/history/timerBuilder.go
@@ -289,7 +289,7 @@ func (tb *timerBuilder) loadActivityTimers(msBuilder *mutableStateBuilder) {
 				td := &timerDetails{
 					TimerSequenceID: TimerSequenceID{VisibilityTimestamp: startToCloseExpiry},
 					ActivityID:      v.ScheduleID,
-					EventID:         v.StartedID,
+					EventID:         v.ScheduleID,
 					TimeoutType:     w.TimeoutTypeStartToClose,
 					TimeoutSec:      v.StartToCloseTimeout,
 					TaskCreated:     (v.TimerTaskStatus & TimerTaskStatusCreatedStartToClose) != 0}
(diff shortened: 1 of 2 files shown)


service/history/timerBuilder.go
```diff
@@ -303,7 +303,7 @@ func (tb *timerBuilder) loadActivityTimers(msBuilder *mutableStateBuilder) {
 					td := &timerDetails{
 						TimerSequenceID: TimerSequenceID{VisibilityTimestamp: heartBeatExpiry},
 						ActivityID:      v.ScheduleID,
-						EventID:         v.StartedID,
+						EventID:         v.ScheduleID,
 						TimeoutType:     w.TimeoutTypeHeartbeat,
 						TimeoutSec:      v.HeartbeatTimeout,
 						TaskCreated:     (v.TimerTaskStatus & TimerTaskStatusCreatedHeartbeat) != 0}
```
yiminc: The problem is after first heartbeat timer is created, the bit for heartbeat timer will be set, and after that this TaskCreated will always be true so firstActivityTimerTask() will not return the timer task and it won't be created.

wxing1292: @yiminc 
if my understanding is correct, https://github.com/uber/cadence/blob/master/service/history/timerQueueActiveProcessor.go#L307 will take care of that?

wxing1292: i actually have some doubt:
here: https://github.com/uber/cadence/pull/658/files#diff-b3d25d1c01ad6e7393fd3958caea1333R287
the start event ID is actually checked
so when the timer detail is created (by loading activity into memory), we are sure that the start event ID is a valid event ID

yiminc: Yes, I think this fix works.

yiminc: The problem is the startedID is buffered event id which changed to real start event after it is flushed.

PR #659, opened 2018-04-06
Replication support for all history events

samarabbas:
IDL changes to replicator and history to also include history of next
run with the same replication task.
Mutable state changes to have replication api directly invoked by
replicator when it is applying history events on the passive
cluster.
Fixed ContinuedAsNew processing to lookup parentDomainName from cache
and correctly create history events using ParentDomainInfo.


service/history/historyReplicator.go
```diff
@@ -185,8 +154,160 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) error
...
+			msBuilder.ReplicateActivityTaskTimedOutEvent(event)
+
+		case shared.EventTypeActivityTaskCancelRequested:
+			msBuilder.ReplicateActivityTaskCancelRequestedEvent(event)
+
+		case shared.EventTypeActivityTaskCanceled:
+			msBuilder.ReplicateActivityTaskCanceledEvent(event)
+
+		case shared.EventTypeRequestCancelActivityTaskFailed:
+			// No mutable state action is needed
```
yiminc: we should also replicate that event, no?

service/history/historyReplicator.go
```diff
@@ -185,8 +154,160 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) error
...
+			msBuilder.ReplicateTimerStartedEvent(event)
+
+		case shared.EventTypeTimerFired:
+			msBuilder.ReplicateTimerFiredEvent(event)
+
+		case shared.EventTypeTimerCanceled:
+			msBuilder.ReplicateTimerCanceledEvent(event)
+
+		case shared.EventTypeCancelTimerFailed:
+			// No mutable state action is needed
```
yiminc: not sure I understand it. But I thought we need to replicate every events, is that not the case?

service/history/mutableStateBuilder.go
```diff
@@ -572,48 +572,21 @@ func (e *mutableStateBuilder) GetCompletionEvent() (*workflow.HistoryEvent, bool
 }
 
 // DeletePendingChildExecution deletes details about a ChildExecutionInfo.
-func (e *mutableStateBuilder) DeletePendingChildExecution(initiatedEventID int64) error {
-	_, ok := e.pendingChildExecutionInfoIDs[initiatedEventID]
```
yiminc: why is it safe to remove those checks?

service/history/historyReplicator.go
```diff
@@ -185,8 +154,160 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) error
...
+
+			msBuilder.ReplicateWorkflowExecutionContinuedAsNewEvent(domainID, domainName, event, startedEvent, di,
+				newStateBuilder)
+
+			// Generate a transaction ID for appending events to history
+			transactionID, err := r.shard.GetNextTransferTaskID()
+			if err != nil {
+				return err
+			}
+			err = context.replicateContinueAsNewWorkflowExecution(newStateBuilder, nil, nil, transactionID)
```
yiminc: for the new run, you don't need timer tasks and transfer tasks?

service/history/mutableStateBuilder.go
samarabbas: This was dead code even before my change.  All the places it is called from, already checks existence of ChildInfo before calling this API.

service/history/historyReplicator.go
samarabbas: Support for generating Timer and Transfer task for replicated event will come later.  This change is only about generating replication tasks.

PR #665, opened 2018-04-13
bugfix: cache get function, when in pin mode, should not increase the…

wxing1292:
… counter before return

### DIFF
--- a/common/cache/lru.go
+++ b/common/cache/lru.go
@@ -167,16 +167,15 @@ func (c *lru) Get(key interface{}) interface{} {
 
 	entry := element.Value.(*entryImpl)
 
-	if c.pin {
-		entry.refCount++
-	}
-
 	if c.isEntryExpired(entry, time.Now()) {
 		// Entry has expired
 		c.deleteInternal(element)
 		return nil
 	}
 
+	if c.pin {
+		entry.refCount++
+	}
 	c.byAccess.MoveToFront(element)
 	return entry.value
 }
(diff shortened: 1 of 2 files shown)


common/cache/lru.go
```diff
@@ -167,16 +167,15 @@ func (c *lru) Get(key interface{}) interface{} {
 
 	entry := element.Value.(*entryImpl)
 
-	if c.pin {
-		entry.refCount++
-	}
-
 	if c.isEntryExpired(entry, time.Now()) {
 		// Entry has expired
```
mfateev: I'm confused. `c.deleteInternal(element) `apparently deletes the entry with associated `refCount`, so it doesn't matter if it is incremented or not.

wxing1292: isEntryExpired actually check whether the ref count is 0 or not:
https://github.com/uber/cadence/pull/665/files/5f8ad70e36c2d500163634b7c3c6199638c38877#diff-15d63b195284f9a26bdbeb5e40398165R303

mfateev: I see. Then the question is why it wasn't found before? It looks like we don't have an expiration test on staging.

wxing1292: https://github.com/uber/cadence/pull/665/files#diff-15d63b195284f9a26bdbeb5e40398165R279
on the put side, the logic is sound.
hence, when the cache is full, the least used entry is deleted.

wxing1292: moreover, there was test conducted in staging, testing the history service memory usage, since the put side logic is fine, this issue is not detected.

PR #667, opened 2018-04-13
bugfix: mutable state should be reset if the operation is not success…

wxing1292:
…fully

### DIFF
--- a/service/history/historyCache.go
+++ b/service/history/historyCache.go
@@ -123,14 +123,14 @@ func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
 		if atomic.CompareAndSwapInt32(&status, cacheNotReleased, cacheReleased) {
 			if err != nil {
 				switch err.(type) {
-				case *workflow.EntityNotExistsError:
-					// do not clear cache for this type: EntityNotExistsError
 				case *workflow.CancellationAlreadyRequestedError:
 					// do not clear cache for this type: CancellationAlreadyRequestedError
 				case *workflow.WorkflowExecutionAlreadyStartedError:
 					// do not clear cache for this type: WorkflowExecutionAlreadyStartedError
 				case *h.EventAlreadyStartedError:
 					// do not clear cache for this type: EventAlreadyStartedError
+				// we cannot skip the clear of mutable state if there is entity not exist error
+				// since sometimes the domain cache is used and there can be error thrown by it.
 				// we cannot skip the clear of mutable state if there is a bad request,
 				// since the decision is translated to history events in a loop and we do not do the check
 				// at the very begining.
--- a/service/history/historyCache.go
+++ b/service/history/historyCache.go
@@ -23,6 +23,7 @@ package history
 import (
 	"sync/atomic"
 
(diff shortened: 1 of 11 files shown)


service/history/historyCache_test.go
```diff
@@ -76,37 +88,72 @@ func (s *historyCacheSuite) SetupTest() {
...
+func (s *historyCacheSuite) TestHistoryCacheClear() {
+	s.mockShard.GetConfig().HistoryCacheMaxSize = 20
+	domainID := "test_domain_id"
+	s.cache = newHistoryCache(s.mockShard, s.logger)
+	we := workflow.WorkflowExecution{
+		WorkflowId: common.StringPtr("wf-cache-test-clear"),
+		RunId:      common.StringPtr(uuid.New()),
+	}
+
+	context, release, err := s.cache.getOrCreateWorkflowExecution(domainID, we)
```
venkat1109: can you spin up 2 or more go-routines that do the same set of things you do below ? that way, if there is any race condition, there is a good chance of finding it at unit test time with race detector

service/history/historyCache_test.go
```diff
@@ -76,37 +88,72 @@ func (s *historyCacheSuite) SetupTest() {
...
 	// Old context should be evicted.
-	newContext, release, err4 := s.cache.getOrCreateWorkflowExecution(domain, we)
+	newContext, release, err4 := s.cache.getOrCreateWorkflowExecution(domainID, we)
 	s.Nil(err4)
 	s.False(context == newContext)
-	release()
+	release(err4)
+}
+
+func (s *historyCacheSuite) TestHistoryCacheClear() {
```
venkat1109: [NOT A BLOCKER for this CR ] can you also add a test that exercises mutable state reload code path (if it doesn't exist already ?) i.e. cache hit followed by clear() due to error followed by mutable state reload

service/history/historyCache.go
```diff
@@ -118,8 +118,11 @@ func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
 	// This will create a closure on every request.
 	// Consider revisiting this if it causes too much GC activity
 	status := cacheNotReleased
-	releaseFunc := func() {
+	releaseFunc := func(err error) {
```
venkat1109: Like the fact that you changed the signature.. that's the best way to find out all places this is being used at compile time 👍

service/history/historyCache_test.go
wxing1292: tested in the update

service/history/historyCache_test.go
wxing1292: done in the update

service/history/historyCache.go
```diff
@@ -118,8 +118,11 @@ func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
...
 		if atomic.CompareAndSwapInt32(&status, cacheNotReleased, cacheReleased) {
+			if err != nil {
```
samarabbas: Let's create a workitem to review if we can skip clearing the cache for certain errors.  Like EntityNotFound.

PR #669, opened 2018-04-13
server side retry

yiminc:
This diff added RetryPolicy and also implemented server side retry for activity. (Will have follow up diff to add the same behavior for child workflow soon, but want to keep this diff short.)

The behavior is, when a retry policy is supplied in ScheduleActivityTaskDecision, server will put that retry policy into activity_info saved into mutable state. 
When record activity started, we only record it if there is no retry policy. If there is retry policy, we update activity_info to remember the started event, but not write to history.
When activity task failed or timed out, we check if a retry is needed. If a retry is needed, we schedule a backoff timer, otherwise we close the activity task. When closing activity task, we added the transient started event before we write the closing event.
When the retry backoff timer fires, we directly generate activity task from timer queue process and send it to matching.
Also added attempt to ActivityTaskStarted, and PollForActivityTaskResponse. Expose the attempt in client side ActivityInfo as well.


service/history/historyEngine.go
```diff
@@ -752,12 +753,26 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(
...
+				}
+				fmt.Printf("^^^ does not have retry policy, adding started event now, :%v\n", ai.StartedID)
+			} else {
+				// we might need to retry, so do not append started event just yet,
+				// instead update mutable state and will record started event when activity task is closed
+				ai.StartedID = transientEventID
+				ai.RequestID = requestID
+				ai.StartedTime = time.Now()
+				ai.StartedIdentity = request.PollRequest.GetIdentity()
+				fmt.Printf("^^^ having retry policy, adding transient started event now: %v\n", ai.StartedID)
```
vancexu: remove

service/history/timerQueueActiveProcessor.go
```diff
@@ -259,6 +265,27 @@ Update_History_Loop:
...
+					continue
+				}
+
+				var retryTask persistence.Task
+				if timeoutType != workflow.TimeoutTypeScheduleToStart {
+					// ScheduleToStart (queue timeout) is not retriable. Instead of retry, user should set larger ScheduleToStart timeout.
+					retryTask = prepareNextRetry(ai, getTimeoutErrorReason(timeoutType))
+				}
+				if retryTask != nil {
+					msBuilder.UpdateActivity(ai)
```
samarabbas: I think we should use the existing pattern to have this logic within mutableState.

service/history/timerQueueActiveProcessor.go
```diff
@@ -422,6 +449,79 @@ Update_History_Loop:
...
+	t.metricsClient.IncCounter(metrics.TimerTaskRetryTimerScope, metrics.TaskRequests)
+	sw := t.metricsClient.StartTimer(metrics.TimerTaskRetryTimerScope, metrics.TaskLatency)
+	defer sw.Stop()
+
+	processFn := func() error {
+		context, release, err0 := t.cache.getOrCreateWorkflowExecution(t.timerQueueProcessorBase.getDomainIDAndWorkflowExecution(task))
+		released := false
+		defer func() {
+			if !released {
+				release()
```
samarabbas: You want to merge the change which takes release with an error, and update this PR with the correct release API.

service/history/timerQueueActiveProcessor.go
```diff
@@ -422,6 +449,79 @@ Update_History_Loop:
...
+		scheduleToStartTimeout := ai.ScheduleToStartTimeout
+
+		release() // release earlier as we don't need the lock anymore
+		released = true
+		err = t.historyService.matchingClient.AddActivityTask(nil, &m.AddActivityTaskRequest{
+			DomainUUID:                    common.StringPtr(targetDomainID),
+			SourceDomainUUID:              common.StringPtr(domainID),
+			Execution:                     &execution,
+			TaskList:                      taskList,
+			ScheduleId:                    &scheduledID,
```
samarabbas: If you don't include the attempt then how do you dedupe started calls?

service/history/timerQueueStandbyProcessor.go
```diff
@@ -237,6 +241,25 @@ func (t *timerQueueStandbyProcessorImpl) processDecisionTimeout(timerTask *persi
 	})
 }
 
+func (t *timerQueueStandbyProcessorImpl) processRetryTimer(timerTask *persistence.TimerTaskInfo) error {
```
samarabbas: @wxing1292 can you make sure this looks good from passive processing of retry tasks?

idl/github.com/uber/cadence/history.thrift
```diff
@@ -114,8 +114,13 @@ struct RecordActivityTaskStartedRequest {
 }
 
 struct RecordActivityTaskStartedResponse {
-  10: optional shared.HistoryEvent startedEvent
```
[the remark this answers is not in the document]
yiminc: If there is retry policy specified, we don't actually create start event, so we won't be able to supply a startedEvent in that case.

idl/github.com/uber/cadence/history.thrift
```diff
@@ -114,8 +114,13 @@ struct RecordActivityTaskStartedRequest {
...
   20: optional shared.HistoryEvent scheduledEvent
```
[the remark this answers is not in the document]
yiminc: We still need the scheduledEvent for the activityType, activityId, and most importantly the input.

service/history/historyBuilder.go
```diff
@@ -30,9 +30,10 @@ import (
 )
 
 const (
-	firstEventID    int64 = 1
-	emptyEventID    int64 = -23
-	bufferedEventID int64 = -123
+	firstEventID     int64 = 1
+	emptyEventID     int64 = -23
+	bufferedEventID  int64 = -123
+	transientEventID int64 = -124
```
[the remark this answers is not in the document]
yiminc: These events are used in many places. I would send a separate diff to remove these ids from here.

schema/cadence/versioned/v0.6/manifest.json
```diff
@@ -1,9 +1,10 @@
 {
   "CurrVersion": "0.6",
   "MinCompatibleVersion": "0.6",
-  "Description": "Persistence support for replication of workflow execution, add cross DC shard attr cluster_transfer_ack_level and cluster_timer_ack_level",
+  "Description": "Persistence support for replication of workflow execution, add cross DC shard attr cluster_transfer_ack_level and cluster_timer_ack_level, add retry policy",
   "SchemaUpdateCqlFiles": [
     "history_replication_task.cql",
-    "add_shard_cluster_ack_level.cql"
+    "add_shard_cluster_ack_level.cql",
+    "retry_policy.cql"
```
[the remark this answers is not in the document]
yiminc: done

schema/cadence/schema.cql
```diff
@@ -126,7 +126,18 @@ CREATE TYPE activity_info (
...
+  domain_id                 text,   -- targetDomainID for this activity
+  task_list                 text,
+  started_identity          text,   -- last started poller's identity
+  has_retry_policy          boolean,-- If there is a retry policy
+  init_interval             int,    -- initial retry interval, in seconds
+  backoff_coefficient       double,
+  max_interval              int,    -- max retry interval in seconds
+  expiration_time           timestamp, -- retry expiration time
+  max_attempts              int,    -- max number of attempts including initial non-retry attempt
+	non_retriable_errors      list<text>,
```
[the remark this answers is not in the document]
yiminc: done

service/history/historyEngine.go
```diff
@@ -1348,12 +1364,12 @@ func (e *historyEngineImpl) RespondActivityTaskCompleted(req *h.RespondActivityT
 				return nil, ErrStaleState
 			}
 
-			if !isRunning || ai.StartedID == emptyEventID {
+			if !isRunning || ai.StartedID == emptyEventID ||
```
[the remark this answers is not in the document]
yiminc: It is still valid, be cause when startedID is set to TransientID it means it is started.

service/history/historyEngine.go
```diff
@@ -752,12 +753,25 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(
...
 			}
 
-			startedEvent := msBuilder.AddActivityTaskStartedEvent(ai, scheduleID, requestID, request.PollRequest)
-			if startedEvent == nil {
-				// Unable to add ActivityTaskStarted event to history
-				return nil, &workflow.InternalServiceError{Message: "Unable to add ActivityTaskStarted event to history."}
-			}
-			response.StartedEvent = startedEvent
+			if !ai.HasRetryPolicy {
+				startedEvent := msBuilder.AddActivityTaskStartedEvent(ai, scheduleID, requestID, request.PollRequest.GetIdentity())
```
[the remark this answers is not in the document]
yiminc: Correct, we don't need that started event in mutable state anymore. I'm not storing it here. I think I can remove it from mutable state.

service/history/mutableStateBuilder.go
```diff
@@ -1246,15 +1256,26 @@ func (e *mutableStateBuilder) ReplicateActivityTaskScheduledEvent(
 	return ai
 }
 
+func (e *mutableStateBuilder) addTransientActivityStartedEvent(scheduleEventID int64) {
+	if ai, ok := e.GetActivityInfo(scheduleEventID); ok && ai.StartedID == transientEventID {
```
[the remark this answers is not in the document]
yiminc: I think it should not. In the case of the transient started event, we don't even create the event itself, so there is no event to be buffered.

service/history/timerQueueActiveProcessor.go
yiminc: sure, I need to merge that change, as well as your change to the mutable state, there might be other conflicts, I want to hold on rebasing until this diff is good.

service/history/timerQueueActiveProcessor.go
```diff
@@ -422,6 +449,79 @@ Update_History_Loop:
...
+		targetDomainID := ai.DomainID
+		execution := workflow.WorkflowExecution{
+			WorkflowId: common.StringPtr(task.WorkflowID),
+			RunId:      common.StringPtr(task.RunID)}
+		taskList := &workflow.TaskList{
+			Name: &ai.TaskList,
+		}
+		scheduleToStartTimeout := ai.ScheduleToStartTimeout
+
+		release() // release earlier as we don't need the lock anymore
```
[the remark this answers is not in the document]
yiminc: cool, will remove the flag.

service/history/timerQueueActiveProcessor.go
yiminc: good catch, will upate

yiminc: as discussed, will leave it as is now. The impact is that if the task delivered multiple times and the previous one already dispatched to an activity worker, and that task already failed, and a new retry is needed, it would schedule a new backoff timer, but the new task delivery will be dispatched with that new attempt.

service/history/timerQueueStandbyProcessor.go
```diff
@@ -124,6 +124,10 @@ func (t *timerQueueStandbyProcessorImpl) process(timerTask *persistence.TimerTas
 		scope = metrics.TimerTaskWorkflowTimeoutScope
 		err = t.processWorkflowTimeout(timerTask)
 
+	case persistence.TaskTypeRetryTimer:
+		scope = metrics.TimerTaskRetryTimerScope
+		err = t.processRetryTimer(timerTask)
```
wxing1292: the retry timer task, should not be processed by the standby timer processor, so this should be
```
scope = metrics.TimerTaskRetryTimerScope
err = nil
```

service/history/timerQueueStandbyProcessor.go
wxing1292: @samarabbas 
the retry timer task, should not be processed by the standby timer processor, since only the active cluster get to process the activity retry. standby cluster will only deal with "existing" history events, so, the "phantom" events created by active cluster will not be replicated to standby cluster, making retry timer meaningless to standby processor.

service/history/timerBuilder.go
```diff
@@ -315,6 +323,7 @@ func (tb *timerBuilder) loadActivityTimers(msBuilder *mutableStateBuilder) {
 					TimerSequenceID: TimerSequenceID{VisibilityTimestamp: scheduleToStartExpiry},
```
wxing1292: on line 321:
```
scheduleToStartExpiry := v.ExpirationTime
if scheduleToCloseExpiry.IsZero() {
	// v.ExpirationTime could be zero for old activity_infos before this code change (for retry)
	scheduleToStartExpiry = v.ScheduledTime.Add(time.Duration(v.ScheduleToStartTimeout) * time.Second)
}
```

PR #682, opened 2018-04-18
 adding version to history events

wxing1292:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -364,6 +364,7 @@ struct Decision {
 }
 
 struct WorkflowExecutionStartedEventAttributes {
+  01: optional i64 (js.type = "Long") version
```
samarabbas: Let's add version to HistoryEvent as it needs to be on every type of event.  Similar to EventID, EventType and Timestamp.

PR #683, opened 2018-04-18
adding version to history events

wxing1292:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -676,6 +676,7 @@ struct ChildWorkflowExecutionTerminatedEventAttributes {
 }
 
 struct HistoryEvent {
+  01:  optional i64 (js.type = "Long") version
```
samarabbas: Can you add this after EventType as ID 35?

wxing1292: sure

PR #684, opened 2018-04-18
persist event version as part of history, add DomainNotActive error to proper APIs

wxing1292:
solve #622 
partially solve #678


service/history/mutableStateBuilder.go
```diff
@@ -455,6 +455,9 @@ func (e *mutableStateBuilder) createNewHistoryEventWithTimestamp(eventID int64,
 	historyEvent.EventId = common.Int64Ptr(eventID)
 	historyEvent.Timestamp = ts
 	historyEvent.EventType = common.EventTypePtr(eventType)
+	if e.replicationState != nil {
```
samarabbas: This could be problematic.  I think we update the replicationState at the very end of transaction, so you could write old state version on the history event.

wxing1292: ok, i see your point, but should we do the update of the failover version at the very beginning?
history events need to use the failover version, activity / user timer / child workflow, etc, all use that.

service/history/historyEngine.go
```diff
@@ -214,6 +214,8 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 
 	// Generate first decision task event.
 	taskList := request.TaskList.GetName()
+	// TODO when the workflow is going to be replicated, use the
+	// newMutableStateBuilderWithReplicationState
```
samarabbas: We need to hookup this too, otherwise we loose the capability to test cross dc.

service/history/historyEngine.go
```diff
@@ -1750,6 +1752,8 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(signalWithStartRequ
 
 	// Generate first decision task event.
 	taskList := request.TaskList.GetName()
+	// TODO when the workflow is going to be replicated, use the
```
samarabbas: We need to hookup this too, otherwise we loose the capability to test cross dc.

service/history/workflowExecutionContext.go
```diff
@@ -120,26 +126,32 @@ func (c *workflowExecutionContext) replicateWorkflowExecution(request *h.Replica
...
+		c.msBuilder.updateReplicationStateVersion(domainEntry.GetFailoverVersion())
+	}
+	return nil
+}
 
+func (c *workflowExecutionContext) updateWorkflowExecution(transferTasks []persistence.Task,
+	timerTasks []persistence.Task, transactionID int64) error {
+
+	crossDCEnabled := c.msBuilder.replicationState != nil
+	if crossDCEnabled {
```
samarabbas: You probably need to make the same changes to continueAsNew code path.

service/history/mutableStateBuilder.go
```diff
@@ -1851,7 +1867,16 @@ func (e *mutableStateBuilder) AddContinueAsNewEvent(decisionCompletedEventID int
 
 	continueAsNewEvent := e.hBuilder.AddContinuedAsNewEvent(decisionCompletedEventID, newRunID, attributes)
 
-	newStateBuilder := newMutableStateBuilder(e.config, e.logger)
+	var newStateBuilder *mutableStateBuilder
+	if domainEntry.IsGlobalDomain() {
```
wxing1292: continue as new handled over here

service/history/historyEngine.go
```diff
@@ -214,7 +217,15 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 
 	// Generate first decision task event.
 	taskList := request.TaskList.GetName()
-	msBuilder := newMutableStateBuilder(e.shard.GetConfig(), e.logger)
+	// TODO when the workflow is going to be replicated, use the
+	var msBuilder *mutableStateBuilder
+	if domainEntry.IsGlobalDomain() {
```
samarabbas: Let's wrap the new logic behind 'IsGlobalDomainEnabled' check.

service/history/historyEngine.go
```diff
@@ -1750,7 +1784,15 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(signalWithStartRequ
...
```
samarabbas: Let's wrap the new logic behind 'IsGlobalDomainEnabled' check.

service/history/mutableStateBuilder.go
samarabbas: Sounds good.

service/history/workflowExecutionContext.go
```diff
@@ -120,26 +126,32 @@ func (c *workflowExecutionContext) replicateWorkflowExecution(request *h.Replica
...
 	return c.updateHelper(builder, nil, nil, false, true, transactionID)
 }
 
-func (c *workflowExecutionContext) updateWorkflowExecution(transferTasks []persistence.Task,
-	timerTasks []persistence.Task, transactionID int64) error {
-
-	crossDCEnabled := c.shard.GetService().GetClusterMetadata().IsGlobalDomainEnabled()
-	if crossDCEnabled {
+func (c *workflowExecutionContext) updateVersion() error {
+	if c.msBuilder.replicationState != nil {
```
samarabbas: Same thing here.  Just to be sure wrap this check with  'IsGlobalDomainEnabled'.

samarabbas:
WorkflowExecutionContext.UpdateHelper has one more location where it updates replication state for new run under the flag: 'updateReplicationState'.
Make sure to update that code path.

PR #686, opened 2018-04-19
persist version (failover version) as part of timer / transfer task s…

wxing1292:
…o as to deal with case when events are overwritten.

partially solve #678


service/history/workflowExecutionContext.go
```diff
@@ -227,6 +227,18 @@ func (c *workflowExecutionContext) updateHelper(builder *historyBuilder, transfe
 		replicationTasks = append(replicationTasks, c.msBuilder.createReplicationTask())
 	}
 
+	// this is the current failover version
+	var version int64
```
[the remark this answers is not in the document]
wxing1292: i believe those IDs are different.
one is the version, which is only related to domain, task IDs are only related to shard.
Moreover, in the future, we should move all transfer tasks / timer tasks generation into workflow execution contest / ms builder, so, i am going to keep it here.

service/history/historyBuilder.go
```diff
@@ -30,6 +30,8 @@ import (
 )
 
 const (
+	emptyVersion int64 = 0
```
samarabbas: Please give it a non-zero value.  It makes debugging easier.

service/history/workflowExecutionContext.go
```diff
@@ -239,6 +239,15 @@ func (c *workflowExecutionContext) updateHelper(builder *historyBuilder, transfe
 		replicationTasks = append(replicationTasks, c.msBuilder.createReplicationTask())
 	}
 
+	// this is the current failover version
```
samarabbas: ShardContext.Update already has the loop to go over all tasks and assign TaskIDs.  Can we use the same loop to assign version?

PR #691, opened 2018-04-20
Fix development instruction

longquanzheng:
Need to run "glide up" before "make bins".


venkat1109:
```make bins``` is supposed to run glide install before running go-build.
https://github.com/uber/cadence/blob/master/Makefile#L65

And as Maxim already said, glide up is neither necessary nor is it guaranteed to succeed.

PR #693, opened 2018-04-24
 Support for out of order processing of replication tasks

### LINKED ISSUE #672: Support for out of order processing of replication events

samarabbas:
Schema changes to buffer replication tasks as part of mutable state.
Persistence layer changes to provide support for adding new replication
tasks, and deleting replication tasks after they are processed and
applied to workflow execution.

MutableState change to support CRUD of buffered replication tasks.  And
HistoryReplicator changes to detect out-of-order events and instead add
replication task to the buffer.  Also loop through all replication task
and flush any buffered tasks after processing of an update.

fixes #672


common/persistence/cassandraPersistence.go
```diff
@@ -2804,6 +2928,28 @@ func createSignalInfo(result map[string]interface{}) *SignalInfo {
...
+		case "first_event_id":
+			info.FirstEventID = v.(int64)
+		case "next_event_id":
+			info.NextEventID = v.(int64)
+		case "version":
+			info.Version = v.(int64)
+		case "history":
+			h := v.(map[string]interface{})
+			info.History = createSerializedHistoryEventBatch(h)
+		case "new_run_history":
```
wxing1292: what is this new run history?

common/persistence/cassandraPersistence_test.go
```diff
@@ -1754,3 +1943,21 @@ func copyReplicationInfo(sourceInfo *ReplicationInfo) *ReplicationInfo {
 		LastEventID: sourceInfo.LastEventID,
 	}
 }
+
+func serializeHistoryEvents(events []*gen.HistoryEvent) *SerializedHistoryEventBatch {
```
wxing1292: can you add (s *cassandraPersistenceSuite)  so this function will not be used by src code?

common/persistence/cassandraPersistence_test.go
```diff
@@ -1754,3 +1943,21 @@ func copyReplicationInfo(sourceInfo *ReplicationInfo) *ReplicationInfo {
...
+	historySerializer := NewJSONHistorySerializer()
+	bufferedBatch := NewHistoryEventBatch(GetDefaultHistoryVersion(), events)
+	serializedEvents, _ := historySerializer.Serialize(bufferedBatch)
+
+	return serializedEvents
+}
+
+func deserializedHistoryEvents(batch *SerializedHistoryEventBatch) *HistoryEventBatch {
```
wxing1292: can you add (s *cassandraPersistenceSuite)  so this function will not be used by src code?

common/persistence/dataInterfaces.go
```diff
@@ -511,20 +521,22 @@ type (
...
+		DeleteChildExecutionInfo      *int64
+		UpsertRequestCancelInfos      []*RequestCancelInfo
+		DeleteRequestCancelInfo       *int64
+		UpsertSignalInfos             []*SignalInfo
+		DeleteSignalInfo              *int64
+		UpsertSignalRequestedIDs      []string
+		DeleteSignalRequestedID       string
+		NewBufferedEvents             *SerializedHistoryEventBatch
+		ClearBufferedEvents           bool
+		NewBufferedReplicationTask    *BufferedReplicationTask
```
wxing1292: should we use a slice? or is NewBufferedReplicationTask and DeleteBufferedReplicationTask guaranteed to have at most one at a time?

service/history/historyReplicator.go
```diff
@@ -88,7 +89,42 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
...
+			if t := msBuilder.BufferReplicationTask(request); t == nil {
+				return errors.New("failed to add buffered replication task")
+			}
+
+			return nil
+		}
+	}
+
+	// First check if there are events which needs to be flushed before applying the update
+	err = r.FlushBuffer(context, msBuilder, request)
```
wxing1292: the flush buffer function will internally call ApplyReplicationTask, so is there any redundant logic here?

service/history/mutableStateBuilder.go
```diff
@@ -291,10 +300,86 @@ func (e *mutableStateBuilder) CloseUpdateSession() (*mutableStateSessionUpdates,
...
 	}
+	e.updateBufferedReplicationTasks = nil
+	e.deleteBufferedReplicationEvent = nil
 
 	return updates, nil
 }
 
+func (e *mutableStateBuilder) BufferReplicationTask(
+	request *h.ReplicateEventsRequest) *persistence.BufferedReplicationTask {
+	if _, ok := e.GetBufferedReplicationTask(request.GetFirstEventId()); ok {
```
wxing1292: what if there are 2 active cluster?
2 can sent out a batch of events which have the same first event ID, i believe a comparison of failover version is needed?

wxing1292:
if travis could not build, just change this line:
https://github.com/uber/cadence/blob/master/.travis.yml#L26

common/persistence/cassandraPersistence.go
samarabbas: Replication Task for the ContinueAsNew has the history for both current run and new run as the same transaction.

common/persistence/cassandraPersistence_test.go
samarabbas: Done.

common/persistence/cassandraPersistence_test.go
samarabbas: Done.

common/persistence/dataInterfaces.go
samarabbas: Currently we are applying replication task as a single transaction, so there is no need currently.  We will cross that bridge when there is a need for batching them together.

service/history/historyReplicator.go
samarabbas: Flush calls it for buffered replication tasks, not for the current one being processed.

service/history/mutableStateBuilder.go
samarabbas: Yes.  It will be handled once I start implementing the merge protocol.

PR #695, opened 2018-04-25
Use pagination, instead of limit keyword

wxing1292:
deprecate in favor of #829


common/persistence/cassandraPersistence.go
```diff
@@ -1617,7 +1617,7 @@ func (d *cassandraPersistence) GetTransferTasks(request *GetTransferTasksRequest
 		defaultVisibilityTimestamp,
 		request.ReadLevel,
 		request.MaxReadLevel,
-		request.BatchSize)
+	).PageSize(request.BatchSize).PageState(request.NextPageToken)
```
samarabbas: How do you guarantee parameters for the query are not changed between multiple calls for different pages?

common/persistence/cassandraPersistence.go
```diff
@@ -2011,7 +2017,7 @@ func (d *cassandraPersistence) GetTasks(request *GetTasksRequest) (*GetTasksResp
 		rowTypeTask,
 		request.ReadLevel,
 		request.MaxReadLevel,
-		request.BatchSize)
+	).PageSize(request.BatchSize)
```
samarabbas: Let's leave this one to how it was before using the LIMIT.  Matching engine already protects range query by limiting the range of task_id.

service/history/replicatorQueueProcessor.go
```diff
@@ -173,7 +173,7 @@ func (p *replicatorQueueProcessorImpl) readTasks(readLevel int64) ([]queueTaskIn
 		tasks[i] = response.Tasks[i]
 	}
 
-	return tasks, len(tasks) >= batchSize, nil
+	return tasks, len(response.NextPageToken) != 0, nil
```
samarabbas: Looks like you will reissue a new query instead of draining the pagination token.  Can we change that?

service/history/timerQueueAckMgr.go
```diff
@@ -286,7 +286,7 @@ func (t *timerQueueAckMgrImpl) getTimerTasks(minTimestamp time.Time, maxTimestam
 	for attempt := 0; attempt < retryCount; attempt++ {
 		response, err := t.executionMgr.GetTimerIndexTasks(request)
 		if err == nil {
-			return response.Timers, len(response.Timers) >= batchSize, nil
+			return response.Timers, len(response.NextPageToken) != 0, nil
```
samarabbas: Same as above.

wxing1292:
deprecate in favor of #829

PR #702, opened 2018-04-27
Update replication state correctly on standby cluster

samarabbas:
Passive cluster need to update replication state map per cluster within
mutable state.  Also simplified update of mutableState for new run when
continued as new.


service/history/mutableStateBuilder.go
```diff
@@ -269,10 +269,29 @@ func (e *mutableStateBuilder) updateReplicationStateVersion(version int64) {
...
+// mutableState is loaded for this workflow execution.
+func (e *mutableStateBuilder) updateReplicationStateLastEventID(clusterName string, lastEventID int64) {
+	if clusterName == "" {
+		// ReplicationState update for active cluster
+		e.replicationState.LastWriteVersion = e.replicationState.CurrentVersion
+		// TODO: Rename this to NextEventID to stay consistent naming convention with rest of code base
+		e.replicationState.LastWriteEventID = lastEventID
+	} else {
+		// ReplicationState update for passive cluster
+		if e.replicationState.LastReplicationInfo == nil {
```
wxing1292: can we initialize this at the very beginning? this is a lot of if else

PR #707, opened 2018-04-28
Make replicator generate transfer / timer tasks generation, add handling of failover version

wxing1292:
* Wire replicator to transfer / timer task generation
* Add failover version checking on transfer / timer task
* Deprecate existing transfer queue test in favor of new tests powered by mocks.

After this PR, we need to throughly test the domain failover.

We probably should also rewrite the timer active processor test, using mocks

solves #567 #568


service/history/transferQueueStandbyProcessor.go
```diff
@@ -222,6 +252,13 @@ func (t *transferQueueStandbyProcessorImpl) processCloseExecution(transferTask *
 	processTaskIfClosed := true
 	return t.processTransfer(processTaskIfClosed, transferTask, func(msBuilder *mutableStateBuilder) error {
 
+		ok, err := t.verifyVersion(transferTask.DomainID, msBuilder.GetCurrentVersion(), transferTask)
```
samarabbas: Can this logic be put within base processTransfer function which already takes care of loading the mutable state?

wxing1292: nop, some task requires checking whether a certain task, i.e. activity task is finished or not before checking the version, sometimes, i.e. close execution, do not require that.
By moving the checking within the function which load the mutable state, the logic will be nested

PR #713, opened 2018-05-03
bugfix: next token is always nil when using limit keyword against cas…

wxing1292:
…sandra


samarabbas:
Can you abandon this request as it is already landed in the patch branch.

PR #715, opened 2018-05-04
Move event constants from historyBuilder to constants 

### LINKED ISSUE #699: [code cleanup]: emptyEventId/bufferedEventId defined in 2 places.
one in common/constants.go, the other one history/historyBuilder.go
need to remove one of them.

mudit3774:
This PR fixes #699


service/history/historyBuilder.go
```diff
@@ -29,15 +29,6 @@ import (
 	"github.com/uber/cadence/common/persistence"
 )
 
-const (
-	emptyVersion int64 = -2234
```
yiminc: looks like we have 2 different values here, @wxing1292 , which value should we use? -2234 or -24.

wxing1292: does not matter, since this value is only used for cross dc, and cross dc is not enabled yet

PR #716, opened 2018-05-04
Mutiple bugfixes for cross DC

wxing1292:
* bugfix: replicator should call notify when error is nil
* bugfix: wire missing transfer task retry

### DIFF
--- a/service/history/historyEngineInterfaces.go
+++ b/service/history/historyEngineInterfaces.go
@@ -112,8 +112,7 @@ type (
 
 	transferQueueProcessor interface {
 		common.Daemon
-		NotifyNewTask(clusterName string, currentTime time.Time)
-		RetryTask(clusterName string)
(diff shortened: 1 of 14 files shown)


service/history/historyEngineInterfaces.go
```diff
@@ -113,6 +113,7 @@ type (
 	transferQueueProcessor interface {
 		common.Daemon
 		NotifyNewTask(clusterName string, currentTime time.Time)
+		RetryTask(clusterName string)
```
samarabbas: Why do we need another API for RetryTask?  Why Notify is not enough?

wxing1292: notify, internally have a check to intentionally drop of the notification if the delay is not met. (for transfer queue side)
retry is just unblocking the standby tasks if there are any. those are total different.

PR #717, opened 2018-05-04
bugfix: history replicator should use domain ID instead of domain

wxing1292:


service/history/historyReplicator.go
```diff
@@ -284,10 +284,13 @@ func (r *historyReplicator) ApplyReplicationTask(context *workflowExecutionConte
 			// Create a new request ID which is used by transfer queue processor if domain is failed over at this point
 			cancelRequestID := uuid.New()
 			rci := msBuilder.ReplicateRequestCancelExternalWorkflowExecutionInitiatedEvent(event, cancelRequestID)
-
 			attributes := event.RequestCancelExternalWorkflowExecutionInitiatedEventAttributes
+			targetDomainEntry, err := r.shard.GetDomainCache().GetDomain(attributes.GetDomain())
```
vancexu: Domain Cache has a `GetDomainID` which might be handy here.

PR #722, opened 2018-05-08
 Conflict resolution of history events after domain failover

samarabbas:
History replicator refactoring to support ApplyHistory events to 
mutable state from replication task or by resetting the mutable
state by replaying events from events table.

Persistence layer support for resetting mutable state

IDL changes to include ReplicationInfo on call to history service for
ReplicateEventsRequest.

IDL changes to include ReplicationInfo map per cluster on each
repliation task generated when workflow execution is updated.

ReplicationQueueProcessor to include ReplicationInfo when pushing
message to kafka to be replicated to other clusters.

ReplicationQueueProcessor to pass through ReplicationInfo from
replication task to ReplicateEvents call to history service on standby
cluster.


service/history/historyReplicator.go
```diff
@@ -64,13 +67,18 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
 	if request == nil || request.History == nil || len(request.History.Events) == 0 {
+		r.logger.Warn("Dropping empty replication task")
 		return nil
 	}
 
 	domainID, err := validateDomainUUID(request.DomainUUID)
 	if err != nil {
 		return err
 	}
+	if len(request.History.Events) == 0 {
```
wxing1292: line 78 is duplication of line 69

service/history/historyReplicator.go
```diff
@@ -499,6 +251,18 @@ func (r *historyReplicator) ApplyReplicationTask(context *workflowExecutionConte
 			LastWriteEventID: lastEvent.GetEventId(),
 		}
 
+		// Set decision attributes after replication of history events
```
wxing1292: why reset the decision?

service/history/historyReplicator.go
```diff
@@ -93,10 +101,43 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
...
+			// Replication state is already on a higher version, we can drop this event
+			// TODO: We need to replay external events like signal to the new version
+			r.logger.Warnf("Dropping stale replication task.  Current Version: %v, Task Version: %v", rState.CurrentVersion,
+				request.GetVersion())
+			return nil
+		}
+
+		// Check if this is the first event after failover
+		if rState.CurrentVersion < request.GetVersion() {
+			previousActiveCluster := r.metadataMgr.ClusterNameForFailoverVersion(rState.CurrentVersion)
```
wxing1292: i think you should use LastWriteVersion instead of CurrentVersion

service/history/conflictResolver.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &conflictResolver{
+		shard:      shard,
+		context:    context,
+		historyMgr: historyMgr,
+		logger:     logger,
+	}
+}
+
+func (r *conflictResolver) reset(replayEventID int64) (*mutableStateBuilder, error) {
```
wxing1292: just want to confirm, after the conflict resolution, the transfer / timer task will be created and notified?

samarabbas: Yes.

service/history/historyReplicator.go
samarabbas: removed.

service/history/historyReplicator.go
samarabbas: Done.

Issue #592, opened 2018-03-02
XDC - Wire config for Cross DC to dynamic config


wxing1292:

vancexu:
Resolved along with https://github.com/uber/cadence/issues/741

PR #725, opened 2018-05-09
reduce number of calls to update shard

wxing1292:
solve #634

### DIFF
--- a/service/history/service.go
+++ b/service/history/service.go
@@ -54,8 +54,8 @@ type Config struct {
 	TimerTaskMaxRetryCount                       int
 	TimerProcessorGetFailureRetryCount           int
 	TimerProcessorCompleteTimerFailureRetryCount int
+	TimerProcessorUpdateShardTaskCount           int
 	TimerProcessorUpdateAckInterval              time.Duration
-	TimerProcessorForceUpdateInterval            time.Duration
 	TimerProcessorCompleteTimerInterval          time.Duration
 	TimerProcessorMaxPollInterval                time.Duration
 	TimerProcessorStandbyTaskDelay               time.Duration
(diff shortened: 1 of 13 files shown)


service/history/service.go
```diff
@@ -107,30 +110,31 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
 		TimerTaskMaxRetryCount:                             5,
 		TimerProcessorGetFailureRetryCount:                 5,
 		TimerProcessorCompleteTimerFailureRetryCount:       10,
-		TimerProcessorUpdateAckInterval:                    10 * time.Second,
-		TimerProcessorForceUpdateInterval:                  10 * time.Minute,
+		TimerProcessorUpdateShardTaskCount:                 100,
```
samarabbas: Let's file a task to make these dynamic config knobs.

service/history/service.go
```diff
@@ -107,30 +110,31 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
...
-		TransferProcessorForceUpdateInterval:               10 * time.Minute,
-		TransferProcessorCompleteTransferInterval:          1 * time.Second,
-		TransferProcessorStandbyTaskDelay:                  0 * time.Minute,
 		TransferTaskWorkerCount:                            10,
 		TransferTaskMaxRetryCount:                          100,
 		TransferProcessorCompleteTransferFailureRetryCount: 10,
+		TransferProcessorUpdateShardTaskCount:              100,
+		TransferProcessorMaxPollInterval:                   60 * time.Second,
+		TransferProcessorUpdateAckInterval:                 1 * time.Minute,
+		TransferProcessorCompleteTransferInterval:          1 * time.Second,
```
samarabbas: 1 second for complete sounds pretty low, considering we have 16K shards.  I think it should also be notification based when there is something to cleanup.  
Let's file a task to review all service configs.

service/history/service.go
wxing1292: see #592

service/history/service.go
wxing1292: #741

PR #726, opened 2018-05-10
bugfix: related to failover version

wxing1292:
* bugfix: when creating workflow, the timer / transfer tasks should be updated with corresponding version
* bugfix: when continue as new a workflow, the timer / transfer tasks should be updated with corresponding version

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -318,9 +318,9 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 		}
 		replicationTasks = append(replicationTasks, replicationTask)
 	}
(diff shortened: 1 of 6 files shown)


service/history/historyEngine.go
```diff
@@ -318,9 +318,9 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 		}
 		replicationTasks = append(replicationTasks, replicationTask)
 	}
+	setTaskVersion(msBuilder.GetCurrentVersion(), transferTasks, timerTasks)
```
samarabbas: here it is outside the create workflow handler and down below it is inside the handler.  Why the difference?

wxing1292: good catch, copy paste error

PR #732, opened 2018-05-13
Fix missing heartbeat timeout issue

samarabbas:
Heartbeat timeout logic expects the same timer is recreated
during the processing of activity heartbeat timeout.  This 
assumption is incorrect when the workflow as many activities
heartbeating at the same time and heartbeating at their own
rate.
Updated the heartbeat timer processing logic to just clear
the timer created flag when heartbeat timer fires and let 
the timer creation logic deal with creation of timer like for 
other timer use cases.


host/integration_test.go
```diff
@@ -1591,25 +1606,25 @@ func (s *integrationSuite) TestActivityHeartbeatTimeouts() {
...
+
+				secondsToSleep := rand.Intn(3)
+				s.logger.Infof("Activity ID '%v' sleeping for: %v seconds", activityID, secondsToSleep)
+				time.Sleep(time.Duration(secondsToSleep) * time.Second)
+			}
 		}
+		s.logger.Infof("End Heartbeating. ID: %v", activityID)
+
+		s.logger.Infof("Sleeping activity before completion. ID: %v", activityID)
+		time.Sleep(5 * time.Second)
```
yiminc: with the sleep in the loop, this test would run for up to 35s? that seems very long for one test.

samarabbas: It will be roughly 25 seconds for this test.  Can't think of another way to test this for now.  Let's work on running the test in parallel to reduce the time.

PR #755, opened 2018-05-18
enable lock with timeout for history

yiminc:
lock with timeout

### DIFF
--- a/service/history/historyCache.go
+++ b/service/history/historyCache.go
@@ -21,6 +21,7 @@
 package history
 
 import (
+	"context"
 	"sync/atomic"
 
 	workflow "github.com/uber/cadence/.gen/go/shared"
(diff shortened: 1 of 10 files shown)


service/history/historyCache.go
```diff
@@ -124,12 +130,15 @@ func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
...
 				context.clear()
 			}
-			context.Unlock()
+			context.locker.Unlock()
 			c.Release(key)
 		}
 	}
 
-	context.Lock()
+	if err := context.locker.Lock(ctx); err != nil {
```
samarabbas: I think we also need to decrement the cache refcount here, as the caller will never call release.
@wxing1292 can you also take a look at this part of the change.

service/history/historyCache.go
```diff
@@ -124,12 +130,15 @@ func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
...
+		// ctx is done before lock can be acquired
```
wxing1292: missing "c.Release(key)" ?
when the context is acquired after line 124, there is a internal counter in the caching layer which will be ++ed.

PR #759, opened 2018-05-19
Limit maximum number of buffered events to protect state

samarabbas:
We store all events within mutable state when the decision is
outstanding.  This could be a problem when decision timeout is very
large and lots of new events comes in during the timeout interval.
This change adds protection to start rejecting new events when there are
already 100 batches in the mutable state.


service/history/handler.go
```diff
@@ -891,6 +892,12 @@ func (h *Handler) updateErrorMetric(scope int, err error) {
 		h.metricsClient.IncCounter(scope, metrics.CadenceErrEntityNotExistsCounter)
 	case *gen.CancellationAlreadyRequestedError:
 		h.metricsClient.IncCounter(scope, metrics.CadenceErrCancellationAlreadyRequestedCounter)
+	case *gen.LimitExceededError:
+		h.metricsClient.IncCounter(scope, metrics.CadenceErrLimitExceededCounter)
+	case *yarpcerrors.Status:
+		if err.Code() == yarpcerrors.CodeDeadlineExceeded {
+			h.metricsClient.IncCounter(scope, metrics.CadenceErrContextTimeoutCounter)
+		}
```
yiminc: what about other yarpc codes?

PR #762, opened 2018-05-19
Add first admin API: InquiryWorkflowExecution

longquanzheng:


idl/github.com/uber/cadence/admin.thrift
```diff
@@ -0,0 +1,48 @@
...
+service AdminService {
+  /**
+  * InquiryWorkflowExecution returns information about the internal states of workflow execution.
+  **/
+  InquiryWorkflowExecutionResponse InquiryWorkflowExecution(1: shared.DescribeWorkflowExecutionRequest inquiryRequest)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.InternalServiceError internalServiceError,
+      3: shared.EntityNotExistsError entityNotExistError,
+      4: AccessDeniedError accessDeniedError,
```
wxing1292: make this shared? access denied can be used for all other APIs in the future

longquanzheng: I feel like it will only be used in admin service in the near future, and our shared.thrift is already too big, so I put it here. I think it is never hard to move it to shard.thrift.

wxing1292: i think we will have to eventually add access control, i.e. eventually throw AccessDeniedError to caller. @samarabbas see what is his preference

longquanzheng: yeah, I am open for it. Let's see our plan. If we are going to do some privilege checking for all frontend APIs, putting it in shared will make good sense. Let me ping him offline.

longquanzheng:
Tested in my laptop:

```
longer@~/gocode/src/github.com/uber/cadence:(admin-dev)$ ./cadence -do test admin wf inq  -w 123
{
  "shardId": "\u0003",
  "historyAddr": "127.0.0.1:7934"
}
```

tools/cli/adminCommands.go
```diff
@@ -0,0 +1,62 @@
...
+	wid := getRequiredOption(c, FlagWorkflowID)
+	rid := c.String(FlagRunID)
+
+	ctx, cancel := newContext()
+	defer cancel()
+
+	resp, err := serviceClient.InquiryWorkflowExecution(ctx, &s.DescribeWorkflowExecutionRequest{
+		Domain: &domain,
+		Execution: &s.WorkflowExecution{
+			WorkflowId: &wid,
```
vancexu: Please use common.StringPtr here to be consistent with existing style

service/frontend/adminHandler.go
```diff
@@ -0,0 +1,121 @@
...
+	if request == nil {
+		return nil, adh.error(errRequestNotSet)
+	}
+
+	if err := adh.validateExecution(request.Execution); err != nil {
+		return nil, err
+	}
+
+	shardID := common.WorkflowIDToHistoryShard(*request.Execution.WorkflowId, adh.numberOfHistoryShards)
+	shardIDstr := string(shardID)
```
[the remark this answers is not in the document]
longquanzheng: I noticed that. But that is the acutal shardID we are using. Our shardID is a string instead of 
a number. Using strconv will make it looks nice however, is incorrect.

service/frontend/adminHandler.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	case *gen.EntityNotExistsError:
+		return err
+	default:
+		logging.LogUncategorizedError(adh.Service.GetLogger(), err)
+		return &gen.InternalServiceError{Message: err.Error()}
+	}
+}
+
+func (adh *AdminHandler) validateExecution(w *gen.WorkflowExecution) error {
```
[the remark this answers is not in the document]
longquanzheng: I wanted to do that at the beginning. It is going to refactor error, and also need to define all metrics(about 10) for a new scope for admin service. But I felt like these will be too heavy for admin API at this point. It will introduce more code, and emit lots of metrics that we are not going to use.

tools/cli/adminCommands.go
longquanzheng: OK

service/frontend/adminHandler.go
```diff
@@ -0,0 +1,121 @@
...
+
+	if err := adh.validateExecution(request.Execution); err != nil {
+		return nil, err
+	}
+
+	shardID := common.WorkflowIDToHistoryShard(*request.Execution.WorkflowId, adh.numberOfHistoryShards)
+	shardIDstr := string(shardID)
+
+	historyHost, err := adh.GetMembershipMonitor().Lookup(common.HistoryServiceName, shardIDstr)
+	historyAddr := historyHost.GetAddress()
```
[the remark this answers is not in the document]
longquanzheng: Sure

service/frontend/adminHandler.go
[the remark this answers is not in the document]
vancexu: got it, then how about just convert back to nice number when output so that it's not "\u0003" but "3"

service/frontend/adminHandler.go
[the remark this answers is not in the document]
vancexu: something from `func (wh *WorkflowHandler) validateExecution(w *gen.WorkflowExecution, scope int) error ` to `func validateExecution(w *gen.WorkflowExecution, scope int) error '; from `func (wh *WorkflowHandler) error(err error, scope int) error` to `func error(err error, scope int, logger, metric...) error`

service/frontend/workflowHandler.go
```diff
@@ -2076,15 +2076,23 @@ func (wh *WorkflowHandler) validateTaskList(t *gen.TaskList, scope int) error {
 	return nil
 }
 
-func (wh *WorkflowHandler) validateExecution(w *gen.WorkflowExecution, scope int) error {
+func (wh *WorkflowHandler) validateExecutionAndEmitMetrics(w *gen.WorkflowExecution, scope int) error {
+	err := validateExecution(w)
+	if err != nil {
+		wh.error(err, scope)
```
vancexu: did you forget `return wh.error`

service/frontend/adminHandler.go
```diff
@@ -0,0 +1,108 @@
...
+
+	historyHost, err := adh.GetMembershipMonitor().Lookup(common.HistoryServiceName, shardIDstr)
+	if err != nil {
+		return nil, adh.error(err)
+	}
+
+	historyAddr := historyHost.GetAddress()
+
+	return &admin.InquiryWorkflowExecutionResponse{
+		ShardId:     &shardIDstr,
```
vancexu: could you use common.StringPtr to follow the existing style?

PR #765, opened 2018-05-20
bugfix: add missing handling of child workflow in history replicator

wxing1292:


service/history/historyReplicator.go
```diff
@@ -205,10 +205,17 @@ func (r *historyReplicator) ApplyReplicationTask(context *workflowExecutionConte
 	firstEvent := request.History.Events[0]
 	switch firstEvent.GetEventType() {
 	case shared.EventTypeWorkflowExecutionStarted:
-		// TODO: Support for child execution
 		var parentExecution *shared.WorkflowExecution
 		initiatedID := common.EmptyEventID
 		parentDomainID := ""
+		if msBuilder.executionInfo.ParentDomainID != "" {
```
samarabbas: Have you tested this change?  I don't think it can work.
In the case of StartWorkflowExecution event, msBuilder is pretty much empty and we need to read the parent execution information from the history event instead.

samarabbas: Actually nevermind.  Just realized the stateBuilder actually applies the event before that check.  Looks good.

PR #774, opened 2018-05-25
Handling of duplicate StartWorkflowExecution event processing

samarabbas:
Replicator could potentially see duplicate StartWorkflowExecution
replication events.  Adding logic to replicator to dedupe such events.

### DIFF
--- a/service/history/historyReplicator.go
+++ b/service/history/historyReplicator.go
@@ -78,21 +78,26 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
 
 	execution := *request.WorkflowExecution
 
-	var context *workflowExecutionContext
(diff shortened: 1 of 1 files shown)


service/history/historyReplicator.go
```diff
@@ -78,21 +78,26 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
...
+		return err
+	}
+	defer func() { release(retError) }()
+
 	var msBuilder *mutableStateBuilder
 	firstEvent := request.History.Events[0]
 	switch firstEvent.GetEventType() {
 	case shared.EventTypeWorkflowExecutionStarted:
+		msBuilder, err = context.loadWorkflowExecution()
+		if err == nil {
```
wxing1292: maybe need a type assertion here, i.e. entity not exist

samarabbas: Good catch.  Let me add that.

PR #775, opened 2018-05-25
Expose ResetStickyTaskList API in frontend

arthurguber:
This commit exposes the ResetStickyTaskList API in frontend service, which up until now
has been internal only and resides in history service.

We expose this API by adding a simple forwarder of the same name in frontend service that client side can call. This forwarder is almost entirely just a pass-through.

The motivation for exposing this previously internal API is to clear the way
for future client side changes that would involve client side workers explicit
clearing the stickiness of their workflows, for example, due to potentially newly
introduced logic that evicts workflow contexts based on a timer.


idl/github.com/uber/cadence/cadence.thrift
```diff
@@ -393,6 +393,23 @@ service WorkflowService {
...
+  /**
+  * Reset the sticky tasklist related information in mutable state of a given workflow.
+  * Things cleared are:
+  * 1. StickyTaskList
+  * 2. StickyScheduleToStartTimeout
+  * 3. ClientLibraryVersion
+  * 4. ClientFeatureVersion
+  * 5. ClientImpl
+  **/
+  shared.ResetStickyTaskListResponse ResetStickyTaskList(1: shared.ResetStickyTaskListRequest resetRequest)
```
vancexu: Why we need to return empty response here? Is this Reset API going to expect anything in future?
Otherwise shouldn't we just return void instead of an empty response?

arthurguber: Yes, it seems in Wenquan's original code it is for future return values.

struct ResetStickyTaskListResponse {
  // The reason to keep this response is to allow returning
  // information in the future.
}

Issue #734, opened 2018-05-14
Cross DC - Better Story to Domain Failover Event.


wxing1292:
Right now, the actual failover event is delivered to history service by a trigger registered in the domain cache, which requires all domain to be present in the cache (which is less ideal in few years later, when we will have millions of domains) and can potentially suffer from loss of domain failover event when there is a shard movement.

PR #777, opened 2018-05-25
Reliable Domain Change Notification

wxing1292:
* add new domain v2 table, with additional 2 parameter: 1. domain change notification version, 2 failover change notification version
* add domain notification version to shard to track the domain change
* modify the domain cache and shard for reliable notification

solve #734


common/persistence/persistenceTestBase.go
```diff
@@ -77,6 +77,7 @@ type (
 		TaskMgr              TaskManager
 		HistoryMgr           HistoryManager
 		MetadataManager      MetadataManager
+		MetadataManagerV2    MetadataManager
```
samarabbas: Can we just have these 2 versions only at the persistence layer?  I'm a little worried bubbling it up across all layers of application might cause problems and bugs.  There are just too many changes because of that.

schema/cadence/schema.cql
```diff
@@ -317,3 +318,19 @@ CREATE TABLE domains_by_name (
...
+CREATE TABLE domains_by_name_v2 (
+  domains_partition             int,
+  name                          text,
+  domain                        frozen<domain>,
+  config                        frozen<domain_config>,
+  replication_config            frozen<domain_replication_config>, -- indicating active cluster and standby cluster used for replication
+  is_global_domain              boolean, -- indicating whether a domain is a global domain
+  config_version                bigint, -- indicating the version of domain config, excluding the failover / change of active cluster name
+  failover_version              bigint, -- indicating the version of active domain only, used for domain failover
+  failover_notification_version bigint, -- indicating the last change related to domain failover
```
samarabbas: why do we need 2 separate notification versions?
Previously we used to have db_version too, why we don't need that anymore?

service/frontend/service.go
```diff
@@ -88,6 +88,20 @@ func (s *Service) Start() {
 	}
 	metadata = persistence.NewMetadataPersistenceClient(metadata, base.GetMetricsClient())
 
+	metadataV2, err := persistence.NewCassandraMetadataPersistenceV2(p.CassandraConfig.Hosts,
```
samarabbas: This is exactly I want to avoid having 2 versions of metadata persistence passed through all layers of the code.

service/history/historyEngine.go
```diff
@@ -167,27 +167,13 @@ func (e *historyEngineImpl) Start() {
 	logging.LogHistoryEngineStartingEvent(e.logger)
 	defer logging.LogHistoryEngineStartedEvent(e.logger)
 
+	e.registerDomainFailoverCallback()
```
samarabbas: You are setting up the callback before starting the processors.  This could be problematic if the callback is received before the processors are even started.

service/history/historyEngine.go
```diff
@@ -205,6 +191,49 @@ func (e *historyEngineImpl) Stop() {
...
+				e.shard.UpdateDomainNotificationVersion(nextDomain.GetNotificationVersion())
+			}
+		},
+	)
+
+	// second check whether this shard should do catch up
+	domainNotificationVersion := e.shard.GetDomainCache().GetDomainNotificationVersion()
+	shardDomainNotificationVersion := e.shard.GetDomainNotificationVersion()
+	if domainNotificationVersion > shardDomainNotificationVersion {
+		for _, domain := range e.shard.GetDomainCache().GetAllDomain() {
```
samarabbas: I think this logic should live within domainCache.  RegisterDomainCallback should take the last notification version and have logic to fire notifications since then for catch up.

common/persistence/dataInterfaces.go
```diff
@@ -912,14 +927,16 @@ type (
...
+	// MetadataManager is used to manage metadata CRUD for domain entities
 	MetadataManager interface {
 		Closeable
 		CreateDomain(request *CreateDomainRequest) (*CreateDomainResponse, error)
 		GetDomain(request *GetDomainRequest) (*GetDomainResponse, error)
 		UpdateDomain(request *UpdateDomainRequest) error
 		DeleteDomain(request *DeleteDomainRequest) error
 		DeleteDomainByName(request *DeleteDomainByNameRequest) error
+		ListDomain(request *ListDomainRequest) (*ListDomainResponse, error)
+		GetMetadata() (int64, error)
```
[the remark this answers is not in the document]
wxing1292: not entirely sure what do you mean.

the list api will list all domains in the new table.
get metadata method will return the metadate (notification version) from the new table

common/persistence/persistenceTestBase.go
wxing1292: you see, i am using the same interface.
so, anywhere other than persistence (and domain frontend API and the domain cache) will not use the new metadata manager

schema/cadence/schema.cql
wxing1292: the failover_notification_version is actually your idea, tracking the last change which related to domain failover

the previous db_version is now the notification_version, managed by the metadata record

service/frontend/service.go
wxing1292: maybe we should talk

service/history/historyEngine.go
wxing1292: the failover processor is independent of the normal processor

service/history/historyEngine.go
wxing1292: i thought the domain cache notification aims to be generic?
if so, then we should not put this domain failover specific logic in the domain cache.

common/persistence/dataInterfaces.go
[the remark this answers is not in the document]
wxing1292: ok

Issue #779, opened 2018-05-25
Standby processor logic needs to account for conflict resolution


samarabbas:
For example transferQueueStandbyProcessor has code which checks if the transfer task can be processed or not:
```
// First check to see if cache needs to be refreshed as we could potentially have stale workflow execution in
		// some extreme cassandra failure cases.
		if transferTask.ScheduleID >= msBuilder.GetNextEventID() {
			t.metricsClient.IncCounter(metrics.TimerQueueProcessorScope, metrics.StaleMutableStateCounter)
			t.logger.Debugf("processExpiredUserTimer: timer event ID: %v >= MS NextEventID: %v.", transferTask.ScheduleID, msBuilder.GetNextEventID())
			// Reload workflow execution history
			context.clear()
			continue Process_Loop
		}
```

This code will have a problem if history is re-written during replay.  This also needs to check the version on task and workflow execution to detect that and either process or throw away the task accordingly.

samarabbas:
This actually causes transfer queue to get into stuck state.

Issue #770, opened 2018-05-23
Flaky test TestActivityHeartbeatTimeouts


vancexu:
Saw on traivs build once:
--- FAIL: TestIntegrationSuite/TestActivityHeartbeatTimeouts (16.98s)
        testing.go:730: race detected during execution of test

error details: https://api.travis-ci.org/v3/job/382345863/log.txt
``` 
==================
WARNING: DATA RACE
Write at 0x00c4260f42f0 by goroutine 713:
  github.com/uber/cadence/host.(*integrationSuite).TestActivityHeartbeatTimeouts.func3()
      /home/travis/gopath/src/github.com/uber/cadence/host/integration_test.go:1716 +0x63

Previous write at 0x00c4260f42f0 by goroutine 198:
  github.com/uber/cadence/host.(*integrationSuite).TestActivityHeartbeatTimeouts.func3()
      /home/travis/gopath/src/github.com/uber/cadence/host/integration_test.go:1716 +0x63

Goroutine 713 (running) created at:
  github.com/uber/cadence/host.(*integrationSuite).TestActivityHeartbeatTimeouts()
      /home/travis/gopath/src/github.com/uber/cadence/host/integration_test.go:1715 +0xf52
  runtime.call32()
      /home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/asm_amd64.s:573 +0x3a
  reflect.Value.Call()
      /home/travis/.gimme/versions/go1.10.1.linux.amd64/src/reflect/value.go:308 +0xc0
  github.com/uber/cadence/vendor/github.com/stretchr/testify/suite.Run.func2()
      /home/travis/gopath/src/github.com/uber/cadence/vendor/github.com/stretchr/testify/suite/suite.go:102 +0x2ff
  testing.tRunner()
      /home/travis/.gimme/versions/go1.10.1.linux.amd64/src/testing/testing.go:777 +0x16d

Goroutine 198 (finished) created at:
  github.com/uber/cadence/host.(*integrationSuite).TestActivityHeartbeatTimeouts()
      /home/travis/gopath/src/github.com/uber/cadence/host/integration_test.go:1715 +0xf52
  runtime.call32()
      /home/travis/.gimme/versions/go1.10.1.linux.amd64/src/runtime/asm_amd64.s:573 +0x3a
  reflect.Value.Call()
      /home/travis/.gimme/versions/go1.10.1.linux.amd64/src/reflect/value.go:308 +0xc0
[...]

PR #794, opened 2018-05-30
Mutiple Bugfixes

wxing1292:
* fix workflow timeout version check bug
* fix task event ID check bug
* fix integtest race condition

solve #779 #770

### DIFF
--- a/service/history/conflictResolver.go
+++ b/service/history/conflictResolver.go
@@ -21,8 +21,6 @@
 package history
 
 import (
-	"time"
-
 	"github.com/pborman/uuid"
 	"github.com/uber-common/bark"
 	"github.com/uber/cadence/.gen/go/shared"
@@ -52,7 +50,7 @@ func newConflictResolver(shard ShardContext, context *workflowExecutionContext, 
 	}
 }
 
-func (r *conflictResolver) reset(replayEventID int64, startTime time.Time) (*mutableStateBuilder, error) {
+func (r *conflictResolver) reset(replayEventID int64) (*mutableStateBuilder, error) {
 	domainID := r.context.domainID
 	execution := r.context.workflowExecution
 	replayNextEventID := replayEventID + 1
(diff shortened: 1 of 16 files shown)


service/history/conflictResolver.go
```diff
@@ -60,33 +62,44 @@ func (r *conflictResolver) reset(replayEventID int64) (*mutableStateBuilder, err
 	var resetMutableStateBuilder *mutableStateBuilder
 	var sBuilder *stateBuilder
 	requestID := uuid.New()
-	for hasMore := true; hasMore; hasMore = len(nextPageToken) > 0 {
-		history, nextPageToken, err = r.getHistory(domainID, execution, common.FirstEventID, replayNextEventID,
+
+	var lastFirstEventID int64
+	for remainingHistorySize := replayNextEventID - common.FirstEventID; remainingHistorySize > 0; {
```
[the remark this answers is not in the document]
wxing1292: 1. i did not change the query
2. the change here make sure we are actually applying events up to the next event id. if the next event id < the highest event id in a batch of history

say in a batch, the first event id is 10, next event id is 15, and we only want events up to 13 (exclusive).
then this change can make it work, while before, we are actually applying all events from 10 -> 15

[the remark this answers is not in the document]
samarabbas: My question is why can't we rely on nextPageToken for the looping condition.  I agree with you that last batch could have more events within the batch after nextEventID (although it is not possible unless we have a bug somewhere else).  But then it guarantees that there won't be any more batches after that.  So it is still ok to loop on the nextPageToken logic.

service/history/failoverCheck.go
```diff
@@ -61,3 +64,77 @@ func verifyTimerTaskVersion(shard ShardContext, domainID string, version int64,
 	}
 	return true, nil
 }
+
+// load mutable state, if mutable state's next event ID <= task ID, will attempt to refresh
+// if still mutable state's next event ID <= task ID, will return nil, nil
+func loadMutableStateForTransferTask(context *workflowExecutionContext, transferTask *persistence.TransferTaskInfo, metricsClient metrics.Client, logger bark.Logger) (*mutableStateBuilder, error) {
```
samarabbas: Why only transfer task?  Won't TimerTask have a similar issue?

service/history/failoverCheck.go
```diff
@@ -61,3 +64,77 @@ func verifyTimerTaskVersion(shard ShardContext, domainID string, version int64,
...
+
+	// check to see if cache needs to be refreshed as we could potentially have stale workflow execution
+	// the exception is decision consistently fail
+	// there will be no event generated, thus making the decision schedule ID == next event ID
+	isDecisionRetry := transferTask.TaskType == persistence.TransferTaskTypeDecisionTask &&
+		msBuilder.executionInfo.DecisionScheduleID == transferTask.ScheduleID &&
+		msBuilder.executionInfo.DecisionAttempt > 0
+
+	if transferTask.ScheduleID >= msBuilder.GetNextEventID() && !isDecisionRetry {
+		metricsClient.IncCounter(metrics.TimerQueueProcessorScope, metrics.StaleMutableStateCounter)
```
samarabbas: Probably you want to use a different metric scope here.

service/history/failoverCheck.go
samarabbas: Is there a way to generalize the check for both timer and transfer tasks?  Can we instead rely on version on the task and version on the mutable state?

wxing1292: the main difference is here
```
isDecisionRetry := transferTask.TaskType == persistence.TransferTaskTypeDecisionTask &&
```
vs 
```
isDecisionRetry := timerTask.TaskType == persistence.TaskTypeDecisionTimeout &&
```

service/history/failoverCheck.go
wxing1292: fixed

PR #799, opened 2018-05-30
CrossDC bugfixes to replication task generation and conflict resolver

### LINKED ISSUE #792: Cross DC - Investigate MS Builder Being Reset After Conflict Resolution

samarabbas:
Replication information captured for replication task has the wrong
last event id, as it is captured before CloseSession on mutable state.

Update stale replication task detection to use LastWriteEventID instead
of Current version.  Otherwise this results in throwing away tasks when
they could potentially be applied.

Fix broken conflict resolution logic to set mutable state fields like
next event id, lastFirstEventID, etc which were not set as a result of
applying history events and need to be set explicitly.

fix #792 #796

### DIFF
--- a/service/history/conflictResolver.go
+++ b/service/history/conflictResolver.go
@@ -80,6 +80,8 @@ func (r *conflictResolver) reset(replayEventID int64, startTime time.Time) (*mut
 			history.Events = history.Events[0:eventsToApply]
 		}
 
+		eventsToApply -= batchSize
+
 		if len(history.Events) == 0 {
 			break
 		}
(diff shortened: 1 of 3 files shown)


samarabbas:
fixes #792

service/history/conflictResolver.go
```diff
@@ -59,34 +61,58 @@ func (r *conflictResolver) reset(replayEventID int64) (*mutableStateBuilder, err
...
-			if event.GetEventId() == common.FirstEventID {
-				resetMutableStateBuilder = newMutableStateBuilderWithReplicationState(r.shard.GetConfig(), r.logger,
-					event.GetVersion())
+		batchSize := int64(len(history.Events))
+		// NextEventID could be in the middle of the batch.  Trim the history events to not have more events then what
+		// need to be applied
+		if batchSize > eventsToApply {
+			history.Events = history.Events[0:eventsToApply]
+		}
+
```
wxing1292: missing a 
```
eventsToApply -= int64(len(history.Events))
```
here

PR #804, opened 2018-05-31
bugfix: reset mutable state

wxing1292:


service/history/conflictResolver.go
```diff
@@ -106,6 +106,7 @@ func (r *conflictResolver) reset(replayEventID int64, startTime time.Time) (*mut
 	resetMutableStateBuilder.executionInfo.StartTimestamp = startTime
 	// the last updated time is not important here, since this should be updated with event time afterwards
 	resetMutableStateBuilder.executionInfo.LastUpdatedTimestamp = startTime
+	resetMutableStateBuilder.updateReplicationStateLastEventID(sourceCluster, replayEventID)
```
samarabbas: I think just setting the lastWriteEventID won't be enough.  We also need to set the lastWriteVersion correctly too.

service/history/historyReplicator.go
```diff
@@ -142,7 +142,7 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
 					request.GetSourceCluster(), request.GetVersion(), request.GetFirstEventId(), request.GetNextEventId())
 
 				resolver := newConflictResolver(r.shard, context, r.historyMgr, r.logger)
-				msBuilder, err = resolver.reset(ri.GetLastEventId(), msBuilder.executionInfo.StartTimestamp)
+				msBuilder, err = resolver.reset(request.GetSourceCluster(), ri.GetLastEventId(), msBuilder.executionInfo.StartTimestamp)
```
samarabbas: I think it is incorrect to pass in the source cluster on the replication task.  It needs to be inferred from the lastWriteVersion.

service/history/conflictResolver.go
wxing1292: those are updated in #806

service/history/historyReplicator.go
wxing1292: those are updated in #806

wxing1292:
close in favor of #806

PR #805, opened 2018-05-31
Implement inquireMutableState

longquanzheng:
Initial PR, I will implement unit test for HistoryCache if there is no major issue in the code change.


client/history/client.go
```diff
@@ -39,10 +39,10 @@ type clientImpl struct {
 	resolver        membership.ServiceResolver
 	tokenSerializer common.TaskTokenSerializer
 	numberOfShards  int
-	// TODO: consider refactor thriftCache into a separate struct
-	thriftCacheLock sync.RWMutex
-	thriftCache     map[string]historyserviceclient.Interface
-	rpcFactory      common.RPCFactory
+	// TODO: consider refactor InquireCache into a separate struct
```
yiminc: why this change?

client/history/client.go
```diff
@@ -39,10 +39,10 @@ type clientImpl struct {
...
+	InquireCacheLock sync.RWMutex
```
arthurguber: Are these intentionally Upper case?

yiminc: why this change?

client/history/client.go
```diff
@@ -517,24 +541,24 @@ func (c *clientImpl) createContext(parent context.Context) (context.Context, con
 }
 
 func (c *clientImpl) getThriftClient(hostPort string) historyserviceclient.Interface {
-	c.thriftCacheLock.RLock()
-	client, ok := c.thriftCache[hostPort]
-	c.thriftCacheLock.RUnlock()
+	c.InquireCacheLock.RLock()
```
arthurguber: If you just do:
c.InquireCacheLock.RLock()
defer c.InquireCacheLock.Unlock()
here, you won't have to acquire the lock again and the lock is cleaner?

idl/github.com/uber/cadence/admin.thrift
```diff
@@ -38,7 +38,15 @@ service AdminService {
...
-struct InquiryWorkflowExecutionResponse{
+struct InquireWorkflowExecutionRequest {
+  10: optional string domain
+  20: optional shared.WorkflowExecution execution
+}
+
+struct InquireWorkflowExecutionResponse{
   10: optional string shardId
   20: optional string historyAddr
+  30: optional string otherInfo
```
yiminc: what is this?

service/history/historyCache.go
```diff
@@ -96,12 +95,45 @@ func (c *historyCache) getOrCreateWorkflowExecutionWithTimeout(ctx context.Conte
...
+	execution workflow.WorkflowExecution) (*workflowExecutionContext, *workflowExecutionContext, releaseWorkflowExecutionFunc, bool, error) {
+	if err := c.validateWorkflowExecutionInfo(domainID, &execution); err != nil {
+		return nil, nil, nil, false, err
+	}
+
+	key := execution.GetRunId()
+	contextFromCache, cacheHit := c.Get(key).(*workflowExecutionContext)
+	releaseFunc := func(error) {}
+	// If cache hit, we need to lock the cache to prevent race condition
+	if cacheHit {
```
yiminc: this feels complicated, it would be simpler if we just return from whatever in the cache, and provide another API to force unload the cache, which is also needed as an admin command.

service/history/historyEngine.go
```diff
@@ -527,7 +529,56 @@ func (e *historyEngineImpl) getMutableState(ctx context.Context,
...
+	}
+	defer func() { release(retError) }()
+	retResp = &h.InquireMutableStateResponse{}
+
+	if cacheHit && cacheCtx.msBuilder != nil {
+		msb := cacheCtx.msBuilder
+		retResp.MutableStateInCache, retError = e.toMutableStateJSON(msb)
+		otherInfo := ""
+		if len(msb.updateActivityInfos) > 0 {
+			otherInfo += "updateActivityInfos is not empty;"
```
yiminc: why do we need this otherInfo? The toMutableStateJSON should already include everything, right?

service/history/mutableStateBuilder.go
```diff
@@ -169,6 +169,22 @@ func newMutableStateBuilderWithReplicationState(config *Config, logger bark.Logg
 	return s
 }
 
+func (e *mutableStateBuilder) Unload() *persistence.WorkflowMutableState {
```
yiminc: The function name is unclear. Maybe name it as CopyTo...

client/history/client.go
longquanzheng: This is a mistake when I refactor the naming. I will change it back. Thanks for reminding.

client/history/client.go
longquanzheng: As above, I mistakenly changed the naming and this is not part of my scope. But we can open another issue to optimize the locking mechanism.
One reason I can see that why they did this, is they want to release the lock earlier? But I can see that they did that on purpose to improve the performance, as the trade off, it will sacrify some readability

idl/github.com/uber/cadence/admin.thrift
longquanzheng: @wxing1292 said he need more information in mutableStateBuilder like update/deleteActivityInfos, making sure that they are empty.

service/history/historyCache.go
longquanzheng: This s about requirement of the task. I will talk to you in uChat

service/history/historyEngine.go
longquanzheng: Like above. there are still someother information like updateActivityInfos. @wxing1292 you guys can discuss if they are really needed. I don't have too much context of how we are going to use it.

service/history/mutableStateBuilder.go
longquanzheng: I felt so too. did that because we already have 
```
func (e *mutableStateBuilder) Load(state *persistence.WorkflowMutableState)" doing the opposite. 
```
CopyTo might not be a better name because mutableStateBuilder and mutabeState struct are not exactly the same.
To be more clear, we probably can call it loadFromMutableState and unloadToMutableState.

client/history/client.go
arthurguber: ok got it.

idl/github.com/uber/cadence/admin.thrift
```diff
@@ -27,9 +27,9 @@ include "shared.thrift"
 **/
 service AdminService {
   /**
-  * InquiryWorkflowExecution returns information about the internal states of workflow execution.
+  * InquireWorkflowExecution returns information about the internal states of workflow execution.
   **/
-  InquiryWorkflowExecutionResponse InquiryWorkflowExecution(1: shared.DescribeWorkflowExecutionRequest inquiryRequest)
+  InquireWorkflowExecutionResponse InquireWorkflowExecution(1: InquireWorkflowExecutionRequest inquireRequest)
```
wxing1292: let us use simple words here, try describe or get.

idl/github.com/uber/cadence/admin.thrift
wxing1292: but not called otherInfo, as a string.

moreover, shardId is a int64, i believe.

service/history/historyEngine.go
wxing1292: well, i still prefer dumping everything as original type, not string

idl/github.com/uber/cadence/admin.thrift
longquanzheng: Talked offline, I will get rid of otherInfo

idl/github.com/uber/cadence/admin.thrift
yiminc: +1 for Describe

service/history/mutableStateBuilder.go
yiminc: Unload give false impression that you are resetting the mutable state.

longquanzheng: Make sense. How about rename it to "copyTo"?

idl/github.com/uber/cadence/admin.thrift
longquanzheng: OK, I will rename them to DescribeXX

Issue #791, opened 2018-05-29
Cross DC - Better UT Coverage


wxing1292:
We should separate msbuilder, into interface and implementation, so we can mock and have better UT coverage.

samarabbas:
Closing this one as we have already added UT coverage for core components.  Will file explicit workitems for things which need more coverage.

PR #806, opened 2018-05-31
Ut conflict resolver

wxing1292:
*bugfix: getHistory should return output next token, not input token
*bugfix: state builder should set the version on mutable state before apply events, especially when ms builder is reset. 

partially solve #791


service/history/conflictResolver.go
```diff
@@ -107,6 +108,9 @@ func (r *conflictResolver) reset(replayEventID int64, startTime time.Time) (*mut
 	// the last updated time is not important here, since this should be updated with event time afterwards
 	resetMutableStateBuilder.executionInfo.LastUpdatedTimestamp = startTime
 
+	sourceCluster := r.clusterMetadata.ClusterNameForFailoverVersion(resetMutableStateBuilder.GetCurrentVersion())
```
samarabbas: I think you should read the version from last event rather than rely on CurrentVersion.

service/history/stateBuilder.go
```diff
@@ -48,21 +50,24 @@ type (
...
+func (b *stateBuilder) applyEvents(domainID, requestID string, execution shared.WorkflowExecution,
+	history *shared.History, newRunHistory *shared.History) (*shared.HistoryEvent,
 	*decisionInfo, *mutableStateBuilder, error) {
 	var lastEvent *shared.HistoryEvent
 	var lastDecision *decisionInfo
 	var newRunStateBuilder *mutableStateBuilder
 	for _, event := range history.Events {
 		lastEvent = event
+		// must set the current version, since this is standby here, not active
+		b.msBuilder.updateReplicationStateVersion(event.GetVersion())
```
samarabbas: I'm not sure this is the right place to update CurrentStateVersion.  It is currently updated on workflow execution load.  I think we should just update the API on 'updateReplicationStateLastEventID' to also take in clusterName and lastWriteVersion and update both lastWriteVersion and lastWriteEventID as part of the same API.  Then we can call this explicitly from conflict resolver at the end like we update other mutableState fields.

samarabbas: Let's hold onto this till Monday.  I think we should have a quick discussion before we land this.

service/history/mutableStateBuilder.go
```diff
@@ -324,7 +324,9 @@ func (e *mutableStateBuilder) updateReplicationStateVersion(version int64) {
 
 // Assumption: It is expected CurrentVersion on replication state is updated at the start of transaction when
 // mutableState is loaded for this workflow execution.
-func (e *mutableStateBuilder) updateReplicationStateLastEventID(clusterName string, lastEventID int64) {
+func (e *mutableStateBuilder) updateReplicationStateLastEventID(clusterName string, lastWriteVersion,
+	lastEventID int64) {
+	e.replicationState.LastWriteVersion = lastWriteVersion
 	e.replicationState.LastWriteVersion = e.replicationState.CurrentVersion
```
wxing1292: this line should be deleted.

wxing1292: and the active workflow, when calling this function, should use the current version

samarabbas: fixed.

PR #809, opened 2018-06-04
Add client integration test for data converter

vancexu:
The new added testSuite can be used to write tests using cadence-client.
This pull request contains a test for client feature DataConverter. 

~~On mac, running this test always cause a popup dialog asking you to "accept incoming connections".
Currently there is no good solution to get rid of this, you can press ESC each time or just turn off firewall.~~  This Yarpc related issue is fixed.

### DIFF
--- a/host/client_integration_test.go
+++ b/host/client_integration_test.go
@@ -196,9 +196,13 @@ func (s *clientIntegrationSuite) setupSuite(enableGlobalDomain bool, isMasterClu
 }
 
 // testDataConverter implements encoded.DataConverter using gob
-type testDataConverter struct{}
+type testDataConverter struct {
+	NumOfCallToData   int // for testing to know testDataConverter is called as expected
+	NumOfCallFromData int
+}
 
 func (tdc *testDataConverter) ToData(value ...interface{}) ([]byte, error) {
+	tdc.NumOfCallToData++
 	var buf bytes.Buffer
 	enc := gob.NewEncoder(&buf)
 	for i, obj := range value {
@@ -211,6 +215,7 @@ func (tdc *testDataConverter) ToData(value ...interface{}) ([]byte, error) {
 }
 
 func (tdc *testDataConverter) FromData(input []byte, valuePtr ...interface{}) error {
+	tdc.NumOfCallFromData++
 	dec := gob.NewDecoder(bytes.NewBuffer(input))
 	for i, obj := range valuePtr {
 		if err := dec.Decode(obj); err != nil {
@@ -254,10 +259,10 @@ func testDataConverterWorkflow(ctx workflow.Context, tl string) (string, error) 
 	return result + "," + result1, nil
 }
 
-func (s *clientIntegrationSuite) startWorkerWithDataConverter(tl string, useTestConverter bool) cworker.Worker {
+func (s *clientIntegrationSuite) startWorkerWithDataConverter(tl string, dataConverter encoded.DataConverter) cworker.Worker {
 	opts := cworker.Options{}
-	if useTestConverter {
-		opts.DataConverter = newTestDataConverter()
+	if dataConverter != nil {
+		opts.DataConverter = dataConverter
 	}
 	worker := cworker.New(s.wfService, s.domainName, tl, opts)
 	if err := worker.Start(); err != nil {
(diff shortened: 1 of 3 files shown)


host/client_integration_test.go
```diff
@@ -0,0 +1,372 @@
...
+		wfService  workflowserviceclient.Interface
+		wfClient   client.Client
+		worker     cworker.Worker
+		taskList   string
+	}
+)
+
+func TestClientIntegrationSuite(t *testing.T) {
+	flag.Parse()
+	if *integration {
```
yiminc: where is this defined?

host/client_integration_test.go
```diff
@@ -0,0 +1,372 @@
...
+	return "hello_" + msg, nil
+}
+
+var activityTaskList = "client-integration-data-converter-activity-tasklist"
+
+func testDataConverterWorkflow(ctx workflow.Context) (string, error) {
+	ao := workflow.ActivityOptions{
+		ScheduleToStartTimeout: time.Minute,
+		StartToCloseTimeout:    time.Minute,
+		HeartbeatTimeout:       20 * time.Second,
```
yiminc: remove HeartbeatTimeout as we dont need it here

host/client_integration_test.go
```diff
@@ -0,0 +1,372 @@
...
+	ctx1 := workflow.WithChildOptions(ctx, cwo1)
+	ctx1 = workflow.WithDataConverter(ctx1, newTestDataConverter())
+	var result1 string
+	err1 := workflow.ExecuteChildWorkflow(ctx1, testChildWorkflow, 0, 3).Get(ctx1, &result1)
+	if err1 != nil {
+		logger.Error("Parent execution received child execution 1 failure.", zap.Error(err1))
+		return "", err1
+	}
+
+	logger.Info("Parent execution completed.", zap.String("Result", result+result1))
```
yiminc: could you assert the result and result1 here?

host/client_integration_test.go
```diff
@@ -0,0 +1,372 @@
...
+		Config: &persistence.DomainConfig{
+			Retention:  1,
+			EmitMetric: false,
+		},
+		ReplicationConfig: &persistence.DomainReplicationConfig{},
+	})
+}
+
+// testDataConverter implements encoded.DataConverter using gob
+type testDataConverter struct{}
```
yiminc: could you add 2 counter field to testDataConverter and increase them when ToData() and FromData() is called. We need to assert at the end of the test that they are called expected number of times.

host/client_integration_test.go
vancexu: in onebox.go. This provide a flag to control whether to run integration or not, default to run integration test. It exists in all integration test suites.

host/client_integration_test.go
vancexu: removed

host/client_integration_test.go
vancexu: added

host/client_integration_test.go
vancexu: This doesn't look like integration test, because from this test level you don't care/know what is the correct call times. Also those counters will change if client change somewhere, which confuse people.  
I added a TestClientDataConverter_Failed case to cover that different data converter is actually used.

yiminc: I think it is valuable to verify that count number, and it is also good to verify that we don't call this encoding/decoding more than needed. For example, if there is a regression happen on client code that do this encoding

host/client_integration_test.go
```diff
@@ -0,0 +1,416 @@
...
+	iter := s.wfClient.GetWorkflowHistory(ctx, id, we.GetRunID(), false, 0)
+	for iter.HasNext() {
+		event, _ := iter.Next()
+		if event.GetEventId() == 7 {
+			s.Equal(shared.EventTypeActivityTaskCompleted, event.GetEventType())
+		}
+		if event.GetEventId() == 13 {
+			s.Equal(shared.EventTypeActivityTaskFailed, event.GetEventType())
+		}
+		if event.GetEventId() == 17 {
```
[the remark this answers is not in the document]
vancexu: Good point. Add assertion.

host/client_integration_test.go
vancexu: Ok, I am fine with that. Added.

PR #818, opened 2018-06-06
bugfix: should use shard' domain notification version

wxing1292:


service/history/timerQueueProcessor.go
```diff
@@ -135,9 +135,9 @@ func (t *timerQueueProcessorImpl) FailoverDomain(domainID string) {
 			standbyClusterName = cluster
 		}
 	}
-
 	// the ack manager is exclusive, so just add a cassandra min presicition
 	maxLevel := t.activeTimerProcessor.timerQueueAckMgr.getReadLevel().VisibilityTimestamp.Add(1 * time.Millisecond)
+	t.logger.Infof("Timer Failover Triggered: %v, min level: %v, max level: %v.\n", domainID, minLevel, maxLevel)
```
samarabbas: Can we have domainID as a tag?

service/history/transferQueueProcessor.go
```diff
@@ -147,8 +147,10 @@ func (t *transferQueueProcessorImpl) FailoverDomain(domainID string) {
 			standbyClusterName = cluster
 		}
 	}
+
 	// the ack manager is exclusive, so add 1
 	maxLevel := t.activeTaskProcessor.getReadLevel() + 1
+	t.logger.Infof("Transfer Failover Triggered: %v, min level: %v, max level: %v.\n", domainID, minLevel, maxLevel)
```
samarabbas: Let's have domainID as a tag.

service/history/historyEngine.go
```diff
@@ -199,9 +199,16 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
 		e.shard.GetShardID(),
 		e.shard.GetDomainCache().GetDomainNotificationVersion(),
 		func(prevDomain *cache.DomainCacheEntry, nextDomain *cache.DomainCacheEntry) {
+			domainFailoverNotificationVersion := nextDomain.GetFailoverNotificationVersion()
+			shardNotificationVersion := e.shard.GetDomainNotificationVersion()
+			domainActiveCluster := nextDomain.GetReplicationConfig().ActiveClusterName
+
+			e.logger.Infof("Domain Failover Event: Shard: %v, Domain: %v, ID: %v, Failover Notification Version: %v, Active Cluster: %v, Shard Domain Notification Version: %v\n",
```
samarabbas: Can you convert these parameters into tags?

service/history/timerQueueProcessor.go

service/history/historyEngine.go
wxing1292: this log will only appears once, per shard.
the domain id tagged logger is in each processor

PR #822, opened 2018-06-06
Add ActivityScheduleTimeout deduction logic

### LINKED ISSUE #589: Do not require heartbeat timeout
It is surprising that heartbeat timeout is required:
https://github.com/uber/cadence/blob/master/service/history/historyEngine.go#L1903

Also I would not require other timeouts. My proposal is to require either:

ScheduleToClose or both ScheduleToStart and StartToClose.

arthurguber:
This commit relaxes validateActivityScheduleAttributes by having it fill in some
of the fields related to schedule timeouts if they are empty.

The logic used is identical to the logic used in AWS SWF's java client.

The logic is as follows:
If ScheduleToClose is present and ScheduleToStart or StartToClose is missing,
we will set what is missing to ScheduleToClose.

If ScheduleToClose is missing but both ScheduleToStart and StartToClose are present,
we will set ScheduleToClose to the sum of ScheduleToStart and StartToClose.

resolves #589

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2373,18 +2373,6 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
 		return &workflow.BadRequestError{Message: "ActivityType is not set on decision."}
 	}
 
-	if attributes.StartToCloseTimeoutSeconds == nil || *attributes.StartToCloseTimeoutSeconds <= 0 {
-		return &workflow.BadRequestError{Message: "A valid StartToCloseTimeoutSeconds is not set on decision."}
-	}
-	if attributes.ScheduleToStartTimeoutSeconds == nil || *attributes.ScheduleToStartTimeoutSeconds <= 0 {
-		return &workflow.BadRequestError{Message: "A valid ScheduleToStartTimeoutSeconds is not set on decision."}
-	}
-	if attributes.ScheduleToCloseTimeoutSeconds == nil || *attributes.ScheduleToCloseTimeoutSeconds <= 0 {
-		return &workflow.BadRequestError{Message: "A valid ScheduleToCloseTimeoutSeconds is not set on decision."}
-	}
-	if attributes.HeartbeatTimeoutSeconds == nil || *attributes.HeartbeatTimeoutSeconds < 0 {
-		return &workflow.BadRequestError{Message: "A valid HeartbeatTimeoutSeconds is not set on decision."}
-	}
 	if policy := attributes.RetryPolicy; policy != nil {
 		if policy.GetInitialIntervalInSeconds() <= 0 {
 			return &workflow.BadRequestError{Message: "A valid InitialIntervalInSeconds is not set on retry policy."}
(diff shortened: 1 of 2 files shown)


service/history/historyEngine.go
```diff
@@ -2397,6 +2388,23 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
 		}
 	}
 
+	validScheduleToClose := attributes.GetScheduleToCloseTimeoutSeconds() > 0
```
yiminc: we need to error out when value < 0

service/history/historyEngine.go
```diff
@@ -2397,6 +2388,23 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
...
+	validScheduleToStart := attributes.GetScheduleToStartTimeoutSeconds() > 0
+	validStartToClose := attributes.GetStartToCloseTimeoutSeconds() > 0
+
+	if validScheduleToClose {
+		if !validScheduleToStart {
```
yiminc: we should do the overwrite only when the ScheduleToStartTimeoutSeconds is 0

service/history/historyEngine.go
```diff
@@ -2397,6 +2388,23 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
...
+			attributes.ScheduleToStartTimeoutSeconds = common.Int32Ptr(attributes.GetScheduleToCloseTimeoutSeconds())
+		}
+		if !validStartToClose {
+			attributes.StartToCloseTimeoutSeconds = common.Int32Ptr(attributes.GetScheduleToCloseTimeoutSeconds())
```
yiminc: same as above

service/history/historyEngine.go
```diff
@@ -2397,6 +2388,23 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
...
+		}
+	} else if validScheduleToStart && validStartToClose {
+		attributes.ScheduleToCloseTimeoutSeconds = common.Int32Ptr(attributes.GetScheduleToStartTimeoutSeconds() + attributes.GetStartToCloseTimeoutSeconds())
```
yiminc: and same here, only overwrite when ScheduleToCloseTimeoutSeconds is 0

service/history/historyEngine.go
arthurguber: resolved

service/history/historyEngine.go
arthurguber: resolved

service/history/historyEngine.go
arthurguber: resolved

service/history/historyEngine.go
```diff
@@ -2373,15 +2373,6 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
...
-		return &workflow.BadRequestError{Message: "A valid StartToCloseTimeoutSeconds is not set on decision."}
-	}
-	if attributes.ScheduleToStartTimeoutSeconds == nil || *attributes.ScheduleToStartTimeoutSeconds <= 0 {
-		return &workflow.BadRequestError{Message: "A valid ScheduleToStartTimeoutSeconds is not set on decision."}
-	}
-	if attributes.ScheduleToCloseTimeoutSeconds == nil || *attributes.ScheduleToCloseTimeoutSeconds <= 0 {
-		return &workflow.BadRequestError{Message: "A valid ScheduleToCloseTimeoutSeconds is not set on decision."}
-	}
 	if attributes.HeartbeatTimeoutSeconds == nil || *attributes.HeartbeatTimeoutSeconds < 0 {
 		return &workflow.BadRequestError{Message: "A valid HeartbeatTimeoutSeconds is not set on decision."}
```
mfateev: Why heartbeat is required?

arthurguber: resolved

PR #823, opened 2018-06-07
Multiple bugfixes

### LINKED ISSUE #810: Active cluster check should also account for LastWriteVersion before allowing updates
It is possible for a cluster to think it is active cluster but still get replication task from another cluster with higher failover version.  We should prevent updates for all workflow executions with higher LastWriteVersion then the CurrentVersion for the domain.

wxing1292:
bugfix: schedule delete history timer
bugfix: handle workflow current version < last write version, when domain cache is not updated
bugfix: handle workflow domain cache got updated, after domain active check, before mutable state version update

fix #810


service/history/workflowExecutionContext.go
```diff
@@ -168,8 +172,19 @@ func (c *workflowExecutionContext) updateWorkflowExecution(transferTasks []persi
...
 	createReplicationTask := c.msBuilder.replicationState != nil
 
-	lastWriteVersion := c.msBuilder.GetCurrentVersion()
-	return c.updateHelper(nil, transferTasks, timerTasks, createReplicationTask, "", lastWriteVersion, transactionID)
+	currentVersion := c.msBuilder.GetCurrentVersion()
+	if createReplicationTask {
+		lastWriteVersion := c.msBuilder.GetLastWriteVersion()
+		activeCluster := c.clusterMetadata.ClusterNameForFailoverVersion(currentVersion)
+		currentCluster := c.clusterMetadata.GetCurrentClusterName()
+		if currentVersion < lastWriteVersion {
```
samarabbas: I think we should make sure to now let CurrentVersion go back to previous value once it detects a higher version.
Once we have this guarantee, then this check is no longer needed.

PR #827, opened 2018-06-07
Support for retrying messages within replicator processor

samarabbas:
Kafka consumer for replication of events now only relies on actual topic
it is consuming from and DLQ for messages which cannot be processed due
to bugs in the replication stack.  We no longer uses retry queue for the
configured consumer.

Replication message processing will now infinitely sit in the loop of
transient errors in processing the message.  In the event where
processing logic returns RetryTaskError, it will retry that error few
times before moving the message to DLQ.

Also added a bunch of new metric to help with debugging replication
related issues.


service/worker/processor.go
```diff
@@ -170,78 +177,166 @@ func (p *replicationTaskProcessor) worker(workerWG *sync.WaitGroup) {
...
+	case *shared.WorkflowExecutionAlreadyStartedError:
+		p.metricsClient.IncCounter(scope, metrics.CadenceErrExecutionAlreadyStartedCounter)
+	case *shared.EntityNotExistsError:
+		p.metricsClient.IncCounter(scope, metrics.CadenceErrEntityNotExistsCounter)
+	case *shared.LimitExceededError:
+		p.metricsClient.IncCounter(scope, metrics.CadenceErrLimitExceededCounter)
+	case *shared.RetryTaskError:
+		p.metricsClient.IncCounter(scope, metrics.CadenceErrRetryTaskCounter)
+	case *yarpcerrors.Status:
+		if err.Code() == yarpcerrors.CodeDeadlineExceeded {
```
wxing1292: context deadline?

wxing1292: ok that is actually correct

PR #828, opened 2018-06-08
V0.3.12 patch

wxing1292:


common/persistence/cassandraPersistence.go
```diff
@@ -2272,7 +2278,7 @@ func (d *cassandraPersistence) GetTasks(request *GetTasksRequest) (*GetTasksResp
 		rowTypeTask,
 		request.ReadLevel,
 		request.MaxReadLevel,
-		request.BatchSize)
+	).PageSize(request.BatchSize)
```
samarabbas: You are using paginated query but not returning pageToken back to the caller.

service/history/queueAckMgr.go
```diff
@@ -133,15 +132,21 @@ TaskFilterLoop:
 	return tasks, morePage, nil
 }
 
-func (a *queueAckMgrImpl) completeTask(taskID int64) {
+func (a *queueAckMgrImpl) completeQueueTask(taskID int64) error {
+	err := a.processor.completeTask(taskID)
```
samarabbas: I'm a little confused here.  I thought we are going to cleanup the task as a separate background go-routine.  Each individual processor only updates the acklevel for the cluster.

service/history/replicatorQueueProcessor.go
```diff
@@ -175,7 +175,7 @@ func (p *replicatorQueueProcessorImpl) readTasks(readLevel int64) ([]queueTaskIn
 		tasks[i] = response.Tasks[i]
 	}
 
-	return tasks, len(tasks) >= batchSize, nil
+	return tasks, len(response.NextPageToken) != 0, nil
```
samarabbas: Looks like we don't use the nextPageToken for the next query.

service/history/service.go
```diff
@@ -114,8 +114,8 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
 		TimerProcessorGetFailureRetryCount:                 5,
 		TimerProcessorCompleteTimerFailureRetryCount:       10,
 		TimerProcessorUpdateShardTaskCount:                 100,
-		TimerProcessorUpdateAckInterval:                    1 * time.Minute,
-		TimerProcessorCompleteTimerInterval:                1 * time.Second,
+		TimerProcessorUpdateAckInterval:                    5 * time.Second,
```
samarabbas: Updating shard is a very expensive query.  I'm not sure changing update interval for all 16K shards is a good idea.

service/history/service.go
```diff
@@ -125,16 +125,16 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
 		TransferProcessorCompleteTransferFailureRetryCount: 10,
 		TransferProcessorUpdateShardTaskCount:              100,
 		TransferProcessorMaxPollInterval:                   60 * time.Second,
-		TransferProcessorUpdateAckInterval:                 1 * time.Minute,
-		TransferProcessorCompleteTransferInterval:          1 * time.Second,
+		TransferProcessorUpdateAckInterval:                 5 * time.Second,
```
samarabbas: Same comment as timer update interval.

service/history/service.go
```diff
@@ -125,16 +125,16 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
...
+		TransferProcessorCompleteTransferInterval:          3 * time.Second,
 		TransferProcessorStandbyTaskDelay:                  0 * time.Minute,
 		ReplicatorTaskBatchSize:                            10,
 		ReplicatorTaskWorkerCount:                          10,
 		ReplicatorTaskMaxRetryCount:                        100,
 		ReplicatorProcessorMaxPollRPS:                      100,
 		ReplicatorProcessorUpdateShardTaskCount:            100,
 		ReplicatorProcessorMaxPollInterval:                 60 * time.Second,
-		ReplicatorProcessorUpdateAckInterval:               1 * time.Minute,
+		ReplicatorProcessorUpdateAckInterval:               5 * time.Second,
```
samarabbas: Same as above.

service/history/timerQueueProcessor.go
```diff
@@ -182,6 +182,7 @@ func (t *timerQueueProcessorImpl) completeTimers() error {
 		}
 	}
 
+	t.logger.Infof("Start completing timer task from: %v, to %v.", lowerAckLevel, upperAckLevel)
```
samarabbas: This log will be very noisy.

service/history/transferQueueProcessor.go
```diff
@@ -172,38 +172,39 @@ func (t *transferQueueProcessorImpl) completeTransferLoop() {
...
 		for _, standbyTaskProcessor := range t.standbyTaskProcessors {
-			ackLevel := standbyTaskProcessor.queueAckMgr.getAckLevel()
+			ackLevel := standbyTaskProcessor.queueAckMgr.getQueueAckLevel()
 			if upperAckLevel > ackLevel {
 				upperAckLevel = ackLevel
 			}
 		}
 	}
 
+	t.logger.Infof("Start completing transfer task from: %v, to %v.", lowerAckLevel, upperAckLevel)
```
samarabbas: This is going to be very noisy log.

service/history/transferQueueProcessorBase.go
```diff
@@ -0,0 +1,77 @@
...
+		options:               options,
+		executionManager:      shard.GetExecutionManager(),
+		maxReadAckLevel:       maxReadAckLevel,
+		updateClusterAckLevel: updateClusterAckLevel,
+	}
+}
+
+func (t *transferQueueProcessorBase) readTasks(readLevel int64) ([]queueTaskInfo, bool, error) {
+	batchSize := t.options.BatchSize
+	response, err := t.executionManager.GetTransferTasks(&persistence.GetTransferTasksRequest{
```
samarabbas: It is weird that we use paginationToken in one place and does not use it another place.  I think we should make both paths consistent.

common/persistence/cassandraPersistence.go
wxing1292: this is for the task, by the matching side.
for anything other than timer, we use the task id for querying

service/history/queueAckMgr.go
wxing1292: this will be actually called by each processor.
the replicator queue processor will actually complete the task
the transfer queue processor will do nothing

service/history/replicatorQueueProcessor.go
wxing1292: for anything other than timer, we use the task id for querying

service/history/service.go
wxing1292: we have the shard level min update interval to guarantee shard is not updated frequently (on the database level), however, on the application level (in memory), as frequent as possible

service/history/timerQueueProcessor.go
wxing1292: i can remove this

service/history/transferQueueProcessor.go
wxing1292: i can remove this

service/history/transferQueueProcessorBase.go
wxing1292: that means using the complex logic, over simple existing logic.
the reason that we have to use the pagination token is timer task.

PR #832, opened 2018-06-09
Prevent duplicate user timer creation

samarabbas:
TimerQueueProcesor logic is not setting the correct value on timer
information on mutable state to detect if the timer is already created
or not.  This fix creates the value to TimerTaskStatusCreated to prevent
creation of same user timer again.


service/history/timerBuilder.go
```diff
@@ -372,16 +372,6 @@ func (tb *timerBuilder) createActivityTimeoutTask(fireTimeOut int32, timeoutType
 	}
 }
 
-func (tb *timerBuilder) loadUserTimer(expires time.Time, timerID string, taskCreated bool) (*timerDetails, bool) {
```
wxing1292: not used anywhere?

Issue #814, opened 2018-06-05
XDC - Corner cases for conflict resolution around ContinueAsNew event


samarabbas:
Conflict resolution for ContinueAsNew replication task need to be think through.  
If we reset the ContinueAsNew decision we may have a zombie execution which needs to be terminated.  
On the flip side if we discard ContinueAsNew task we somehow need to figure out a way to disregard later events for discarded new run.

PR #836, opened 2018-06-11
Handle more edge case when apply events

wxing1292:
* UT for history replicator
* Remove deadcode
* Bugfix start & signal workflow

solve #675, #814


service/history/historyReplicator.go
```diff
@@ -37,6 +38,9 @@ import (
 	"github.com/uber/cadence/common/persistence"
 )
 
+var errRetry = &shared.RetryTaskError{Message: "History replicator retry error"}
+var errDLQ = errors.New("History replicator DQL error")
```
samarabbas: nit: Fix "DQL" on error message.

service/history/historyReplicator.go
```diff
@@ -37,6 +38,9 @@ import (
 	"github.com/uber/cadence/common/persistence"
 )
 
+var errRetry = &shared.RetryTaskError{Message: "History replicator retry error"}
```
samarabbas: Can you define all errors within the same var section.  There is already one error for RetryTaskError.

service/history/historyReplicator.go
```diff
@@ -77,32 +81,39 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
 }
 
 func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retError error) {
+	defer func() {
+		if _, ok := retError.(*shared.EntityNotExistsError); ok {
+			retError = errRetry
```
samarabbas: This catch all error could make debugging little harder.  I think we should also include the actual message here.  Or atleast log a warning before returning the catch all error.

service/history/historyReplicator.go
```diff
@@ -112,113 +123,165 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
 			// Workflow execution already exist, looks like a duplicate start event, it is safe to ignore it
 			logger.Info("Dropping stale replication task for start event.")
 			return nil
-		}
-
-		// GetWorkflowExecution failed with some transient error.  Return err so we can retry the task later
-		if _, ok := err.(*shared.EntityNotExistsError); !ok {
+		} else if _, ok := err.(*shared.EntityNotExistsError); !ok {
```
samarabbas: why converting to else-if?

service/history/historyReplicator.go
```diff
@@ -112,113 +123,165 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
...
+		// the continue as new + start workflow execution combination will also be processed here
 		msBuilder, err = context.loadWorkflowExecution()
 		if err != nil {
-			// Check to see if this is workflow execution not exist error.  This means task is delivered out of order
-			// and since we cannot process any other task for workflow execution before processing workflow execution
-			// started event, so let's return RetryTaskErr to let replicator know to retry the task after sometime
-			if _, ok := err.(*shared.EntityNotExistsError); ok {
-				return ErrRetryEntityNotExists
+			if _, ok := err.(*shared.EntityNotExistsError); !ok {
+				return err
```
samarabbas: ErrRetryEntityNotExists error may not be needed if you remove this.

service/history/historyReplicator.go
```diff
@@ -440,14 +375,184 @@ func (r *historyReplicator) FlushBuffer(context *workflowExecutionContext, msBui
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// TODO this pile of logic should be merge into workflow execution context / mutable state
+	msBuilder.executionInfo.LastFirstEventID = firstEvent.GetEventId()
+	msBuilder.executionInfo.NextEventID = lastEvent.GetEventId() + 1
+	incomingVersion := firstEvent.GetVersion()
+	replicationState := &persistence.ReplicationState{
```
samarabbas: Probably we should get rid of this and just call:
msBuilder.updateReplicationStateLastEventID

service/history/historyReplicator.go
```diff
@@ -440,14 +392,197 @@ func (r *historyReplicator) FlushBuffer(context *workflowExecutionContext, msBui
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// TODO this pile of logic should be merge into workflow execution context / mutable state
+	executionInfo.LastFirstEventID = firstEvent.GetEventId()
+	executionInfo.NextEventID = lastEvent.GetEventId() + 1
+	incomingVersion := firstEvent.GetVersion()
+	msBuilder.UpdateReplicationStateLastEventID("", incomingVersion, lastEvent.GetEventId())
```
samarabbas: I think we should use the sourceCluster from replication task instead of empty string.

service/history/historyReplicator.go
```diff
@@ -265,155 +344,28 @@ func (r *historyReplicator) ApplyReplicationTask(context *workflowExecutionConte
...
-			}
-
-			// if the existing workflow is completed, ignore the worklow ID reuse policy
-			// since the policy should be applied by the active cluster,
-			// standby cluster should apply this event without question.
-			isBrandNew = false
-			_, err = createWorkflow(isBrandNew, errExist.RunID)
-		}
-
+		requestID := uuid.New()
```
samarabbas: Can we move all this logic into ApplyStartEvent?  We can simplify this logic to deal with only other events and have an assertion that it should not be called for WorkflowExecutionStartedEvent.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+		release(err)
+		return err
+	}
+	currentLastWriteVersion := currentMsBuilder.GetLastWriteVersion()
+	currentRunID := currentMsBuilder.GetExecutionInfo().RunID
+	release(nil)
+
+	// we can also use the start version
+	if currentLastWriteVersion > incomingVersion {
+		logger.Infof("Dropping replication task. Current RunID: %v, Current LastWriteVersion: %v, Incoming Version: %v.",
```
samarabbas: increment the dropped counter.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+	request *h.ReplicateEventsRequest, logger bark.Logger) (mutableState, error) {
+	var err error
+	// check if to buffer / drop / conflict resolution
+	incomingVersion := request.GetVersion()
+	replicationInfo := request.ReplicationInfo
+	rState := msBuilder.GetReplicationState()
+	if rState.LastWriteVersion > incomingVersion {
+		// Replication state is already on a higher version, we can drop this event
+		// TODO: We need to replay external events like signal to the new version
+		logger.Warnf("Dropping stale replication task. CurrentV: %v, LastWriteV: %v, LastWriteEvent: %v, IncomingV: %v.",
```
samarabbas: increment the 'StaleReplicationEventsCounter' counter.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+			rState.CurrentVersion, rState.LastWriteVersion, rState.LastWriteEventID, incomingVersion)
+		return nil, nil
+	} else if rState.LastWriteVersion < incomingVersion {
```
samarabbas: Can we change 'else if' to 'if'

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+		// Check if this is the first event after failover
+		logger.Infof("First Event after replication. CurrentV: %v, LastWriteV: %v, LastWriteEvent: %v, IncomingV: %v.",
+			rState.CurrentVersion, rState.LastWriteVersion, rState.LastWriteEventID, incomingVersion)
+		previousActiveCluster := r.clusterMetadata.ClusterNameForFailoverVersion(rState.LastWriteVersion)
+		ri, ok := replicationInfo[previousActiveCluster]
+		if !ok {
+			logger.Errorf("No ReplicationInfo Found For Previous Active Cluster. Previous Active Cluster: %v, Request Source Cluster: %v, Request ReplicationInfo: %v.",
+				previousActiveCluster, request.GetSourceCluster(), request.ReplicationInfo)
+			// TODO: Handle missing replication information, #840
+			return nil, ErrMissingReplicationInfo
```
samarabbas: Can you also include the existing comment for this part of the code.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
 		}
 
-		// Check for out of order replication task and store it in the buffer
-		if firstEvent.GetEventId() > msBuilder.GetNextEventID() {
-			logger.Debugf("Buffer out of order replication task.  NextEvent: %v, FirstEvent: %v",
-				msBuilder.GetNextEventID(), firstEvent.GetEventId())
-			r.metricsClient.IncCounter(metrics.ReplicateHistoryEventsScope, metrics.BufferedReplicationTaskCounter)
+		// Detect conflict
+		if ri.GetLastEventId() < rState.LastWriteEventID {
+			logger.Infof("Conflict detected.  State: {CurrentV: %v, LastWriteV: %v, LastWriteEvent: %v}, ReplicationInfo: {PrevActiveCluster: %v, V: %v, LastEventID: %v}",
```
samarabbas: increment 'HistoryConflictsCounter' counter.

service/history/historyReplicator.go
wxing1292: added all 3 types:
HistoryConflictsCounter
StaleReplicationEventsCounter
BufferedReplicationTaskCounter

service/history/historyReplicator.go
wxing1292: fine

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+			return nil, ErrMissingReplicationInfo
 		}
 
-		// Check for out of order replication task and store it in the buffer
-		if firstEvent.GetEventId() > msBuilder.GetNextEventID() {
-			logger.Debugf("Buffer out of order replication task.  NextEvent: %v, FirstEvent: %v",
-				msBuilder.GetNextEventID(), firstEvent.GetEventId())
-			r.metricsClient.IncCounter(metrics.ReplicateHistoryEventsScope, metrics.BufferedReplicationTaskCounter)
+		// Detect conflict
+		if ri.GetLastEventId() < rState.LastWriteEventID {
```
samarabbas: you changed '!=' to '<'.  Whats the reasoning?  Can we capture that as a comment.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
-				return errors.New("failed to add buffered replication task")
+			resolver := r.getNewConflictResolver(context, logger)
+			msBuilder, err = resolver.reset(uuid.New(), ri.GetLastEventId(), msBuilder.GetExecutionInfo().StartTimestamp)
+			logger.Infof("Completed Resetting of workflow execution.  NextEventID: %v. Err: %v", msBuilder.GetNextEventID(), err)
+			if err != nil {
+				return nil, err
 			}
+		} else if ri.GetLastEventId() > rState.LastWriteEventID {
+			// TODO handle this case
+			logger.Errorf("Conflict detected.  State: {CurrentV: %v, LastWriteV: %v, LastWriteEvent: %v}, ReplicationInfo: {PrevActiveCluster: %v, V: %v, LastEventID: %v}",
```
samarabbas: I don't think I get this.  Can you explain?

wxing1292: say if we have a bug, this case can happen, we should return some unique error and do logging

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+				rState.CurrentVersion, rState.LastWriteVersion, rState.LastWriteEventID,
+				previousActiveCluster, ri.GetVersion(), ri.GetLastEventId())
+			return nil, ErrMissingReplicationInfo
```
samarabbas: Let's use a different error here.

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+}
 
-			return nil
+func (r *historyReplicator) ApplyOtherEvents(context *workflowExecutionContext, msBuilder mutableState,
+	request *h.ReplicateEventsRequest, logger bark.Logger) error {
+	var err error
+	firstEventID := request.GetFirstEventId()
+	if firstEventID < msBuilder.GetNextEventID() {
+		// duplicate replication task
+		replicationState := msBuilder.GetReplicationState()
```
samarabbas: increment 'StaleReplicationEventsCounter' metric counter

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+		logger.Debugf("Dropping replication task.  State: {NextEvent: %v, Version: %v, LastWriteV: %v, LastWriteEvent: %v}",
+			msBuilder.GetNextEventID(), replicationState.CurrentVersion, replicationState.LastWriteVersion, replicationState.LastWriteEventID)
+		return nil
+	} else if firstEventID > msBuilder.GetNextEventID() {
```
samarabbas: change 'if' to 'else if'

service/history/historyReplicator.go
wxing1292: all done

service/history/historyReplicator.go
wxing1292: done
// ErrCorruptedReplicationInfo is returned when replication task has corrupted replication information from source cluster
ErrCorruptedReplicationInfo = &shared.BadRequestError{Message: "replication task is has corrupted cluster replication info"}

service/history/historyReplicator.go
```diff
@@ -68,167 +77,238 @@ func newHistoryReplicator(shard ShardContext, historyEngine *historyEngineImpl,
...
+		// out of order replication task and store it in the buffer
+		logger.Debugf("Buffer out of order replication task.  NextEvent: %v, FirstEvent: %v",
+			msBuilder.GetNextEventID(), firstEventID)
+
+		err = msBuilder.BufferReplicationTask(request)
```
samarabbas: increment 'BufferedReplicationTaskCounter' metric counter

service/history/historyReplicator.go
wxing1292: done

service/history/historyReplicator.go
wxing1292: done

service/history/historyReplicator.go
wxing1292: good catch

service/history/historyReplicator.go
wxing1292: copy pasted the existing one: "// Returning BadRequestError to force the message to land into DLQ"
is this the comment you are talking about?

service/history/historyReplicator.go
wxing1292: done

service/history/historyReplicator.go
wxing1292: actually, i think we should keep this logic like this.

all other functions, other than ApplyReplicationTask should prepare the mutable state.
ApplyReplicationTask should apply events to mutable state. and accordingly, create workflow or update workflow.
this can be useful if say, we want to apply events from a remote cluster to recover from a disaster, or simple reply a workflow using existing events for debugging reason

PR #847, opened 2018-06-13
some behavior change on worker

wxing1292:


service/worker/processor.go
```diff
@@ -188,20 +188,10 @@ func (p *replicationTaskProcessor) worker(workerWG *sync.WaitGroup) {
...
-							time.Sleep(p.config.ReplicatorRetryDelay)
-							continue ProcessRetryLoop
-						}
-
+					if err != nil && p.isTransientRetryableError(err) {
 						// Keep on retrying transient errors for ever
-						if p.isTransientRetryableError(err) {
-							continue ProcessRetryLoop
-						}
+						time.Sleep(p.config.ReplicatorRetryDelay)
```
samarabbas: Do we really need this extra delay now?  We already have a backoff on transient errors, and now RetryTaskError is also categorized as a transient error so I think we can get rid of this sleep interval.

wxing1292: sure

PR #851, opened 2018-06-13
Change cadence config to dynamic

vancexu:


common/service/dynamicconfig/constants.go
```diff
@@ -31,31 +31,97 @@ func (k Key) String() string {
...
+// These are keys name dynamic config source use.
 var keys = []string{
 	"unknownKey",
-	_matchingDomainTaskListRoot + "minTaskThrottlingBurstSize",
-	_matchingDomainTaskListRoot + "maxTaskBatchSize",
-	_matchingDomainTaskListRoot + "longPollExpirationInterval",
-	_matchingDomainTaskListRoot + "enableSyncMatch",
-	_matchingDomainTaskListRoot + "updateAckInterval",
-	_matchingDomainTaskListRoot + "idleTasklistCheckInterval",
+	"testGetPropertyKey",
```
samarabbas: Instead of having keys as a flat array, can we make it a map of service to list of configs?  
I think it will be much cleaner if we separate them out per service.  
Another thing is I don't like we rely on the ordering of how keys are defined.  Can we convert them into a map too, otherwise it will be very error prone.  
Basically the keys definition would look something like:
map[serviceIdx]map[int]string

service/frontend/workflowHandler.go
```diff
@@ -1337,8 +1338,9 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		return nil, err
 	}
 
-	if getRequest.GetMaximumPageSize() == 0 {
-		getRequest.MaximumPageSize = common.Int32Ptr(wh.config.DefaultHistoryMaxPageSize)
+	if getRequest.GetMaximumPageSize() <= 0 {
+		domainFilter := dynamicconfig.DomainFilter(getRequest.GetDomain())
```
samarabbas: Can we define a strongly typed helper function on config which takes the domain when reading any config which supports filter on domain.  Then we can have the code to create filter in one location and have a straightforward API for rest of the code to load the config.

common/service/dynamicconfig/constants.go
vancexu: changed

service/frontend/workflowHandler.go
vancexu: add GetIntPropertyWithDomainFilter

common/service/dynamicconfig/constants.go
```diff
@@ -24,38 +24,106 @@ package dynamicconfig
...
+	ReplicatorProcessorMaxPollInterval:                 "history.replicatorProcessorMaxPollInterval",
+	ReplicatorProcessorUpdateAckInterval:               "history.replicatorProcessorUpdateAckInterval",
+	ExecutionMgrNumConns:                               "history.executionMgrNumConns",
+	HistoryMgrNumConns:                                 "history.historyMgrNumConns",
+	MaximumBufferedEventsBatch:                         "history.maximumBufferedEventsBatch",
+	ShardUpdateMinInterval:                             "history.shardUpdateMinInterval",
 }
 
 const (
 	// The order of constants is important. It should match the order in the keys array above.
```
samarabbas: Probably this comment is not relevant anymore.  Delete this.

service/frontend/workflowHandler.go
```diff
@@ -1337,8 +1338,9 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		return nil, err
 	}
 
-	if getRequest.GetMaximumPageSize() == 0 {
-		getRequest.MaximumPageSize = common.Int32Ptr(wh.config.DefaultHistoryMaxPageSize)
+	if getRequest.GetMaximumPageSize() <= 0 {
+		getRequest.MaximumPageSize = common.Int32Ptr(int32(
```
samarabbas: This is still too verbose.  Can we have a strongly typed API on config like:
GetHistoryMaxPageSize(domain)?

common/service/dynamicconfig/constants.go
vancexu: Good catch done

service/frontend/workflowHandler.go
vancexu: Refactored

PR #857, opened 2018-06-15
Refactor and fix matching dynamic configs

vancexu:
Previous matching configs are only filtered by tasklist, which is not correct because taskList might be same for different domain. 
This change all matching configs to dynamic configs, and filtered by domain, taskListName, taskType.


common/service/dynamicconfig/constants.go
```diff
@@ -236,6 +239,8 @@ const (
 	DomainName
 	// TaskListName is the tasklist name
 	TaskListName
+	// TaskType is the task type (0:Decision, 1:Activity)
+	TaskType
```
samarabbas: Probably you need to have a filters item for TaskType.

service/matching/service.go
```diff
@@ -31,34 +31,34 @@ import (
...
 		RangeSize:                       100000,
-		GetTasksBatchSize:               dc.GetIntProperty(dynamicconfig.MatchingGetTasksBatchSize, 1000),
-		UpdateAckInterval:               dc.GetDurationProperty(dynamicconfig.MatchingUpdateAckInterval, 10*time.Second),
-		IdleTasklistCheckInterval:       dc.GetDurationProperty(dynamicconfig.MatchingIdleTasklistCheckInterval, 5*time.Minute),
-		LongPollExpirationInterval:      dc.GetDurationProperty(dynamicconfig.MatchingLongPollExpirationInterval, time.Minute),
-		MinTaskThrottlingBurstSize:      dc.GetIntProperty(dynamicconfig.MatchingMinTaskThrottlingBurstSize, 1),
-		OutstandingTaskAppendsThreshold: dc.GetIntProperty(dynamicconfig.MatchingOutstandingTaskAppendsThreshold, 250),
-		MaxTaskBatchSize:                dc.GetIntProperty(dynamicconfig.MatchingMaxTaskBatchSize, 100),
+		GetTasksBatchSize:               dc.GetIntPropertyFilteredByTaskListInfo(dynamicconfig.MatchingGetTasksBatchSize, 1000),
+		UpdateAckInterval:               dc.GetDurationPropertyFilteredByTaskListInfo(dynamicconfig.MatchingUpdateAckInterval, 10*time.Second),
```
samarabbas: Can you change this to 1 Minute.  It could become a problem if number of tasklist grows in the system.

service/matching/taskListManager.go
```diff
@@ -77,36 +76,46 @@ type taskListConfig struct {
...
 	// taskWriter configuration
-	OutstandingTaskAppendsThreshold int
-	MaxTaskBatchSize                int
+	OutstandingTaskAppendsThreshold func() int
+	MaxTaskBatchSize                func() int
 }
 
-func newTaskListConfig(id *taskListID, config *Config) *taskListConfig {
+func newTaskListConfig(id *taskListID, config *Config, domainCache cache.DomainCache) (*taskListConfig, error) {
+	domainEntry, err := domainCache.GetDomainByID(id.domainID)
```
samarabbas: @wxing1292 Can you review this to make sure whether it is OK to host DomainCache within Matching Engine?  Another option is to pass in DomainName along with DomainID on all Matching Engine API calls.

common/service/dynamicconfig/constants.go
vancexu: Good catch, addressed. Will also change and strengthen unit test for these later

service/matching/matchingEngine.go
```diff
@@ -230,7 +237,7 @@ func (e *matchingEngineImpl) AddActivityTask(addRequest *m.AddActivityTaskReques
 	sourceDomainID := addRequest.GetSourceDomainUUID()
 	taskListName := addRequest.TaskList.GetName()
 	taskListKind := common.TaskListKindPtr(addRequest.TaskList.GetKind())
-	e.logger.Debugf("Received AddActivityTask for taskList=%v WorkflowID=%v, RunID=%v",
+	e.logger.Infof("Received AddActivityTask for taskList=%v WorkflowID=%v, RunID=%v",
```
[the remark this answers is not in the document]
vancexu: changed

service/matching/service.go
vancexu: Changed

service/matching/taskListManager.go
wxing1292: @samarabbas  the initialization in matching side looks fine

PR #858, opened 2018-06-15
Add EnableGlobalDomain as dynamic config

vancexu:
This only changed EnableGlobalDomain knob. 

For replicator queue, we need re-visit what to do


common/service/dynamicconfig/constants.go
```diff
@@ -35,11 +35,15 @@ func (k Key) String() string {
...
-	testGetBoolPropertyKey:     "testGetBoolPropertyKey",
+	testGetPropertyKey:                         "testGetPropertyKey",
+	testGetIntPropertyKey:                      "testGetIntPropertyKey",
+	testGetFloat64PropertyKey:                  "testGetFloat64PropertyKey",
+	testGetDurationPropertyKey:                 "testGetDurationPropertyKey",
+	testGetBoolPropertyKey:                     "testGetBoolPropertyKey",
+	testGetIntPropertyFilteredByDomainKey:      "testGetIntPropertyFilteredByDomainKey",
+	testGetDurationPropertyFilteredByDomainKey: "testGetDurationPropertyFilteredByDomainKey",
+
+	EnableGlobalDomainKey: "enableGlobalDomainKey",
```
samarabbas: May want to use prefix of "system."

Either remove Key suffix or use Key for all other properties.

PR #859, opened 2018-06-18
CLI: add AuthorizationManager layer

longquanzheng:


tools/cli/factory.go
```diff
@@ -45,6 +45,15 @@ type WorkflowClientBuilderInterface interface {
 	BuildAdminServiceClient(c *cli.Context) (adminserviceclient.Interface, error)
 }
 
+// AuthorizationManagerInterface is an interface to build permission checking layer to cadence service.
+// User can customize builder by implementing this interface, and call SetAuthorizationManager after initialize cli.
+// (See cadence/cmd/tools/cli/main.go for example)
+// The customized AuthorizationManager may have more processing on Env, Address and other info.
+type AuthorizationManagerInterface interface {
```
mfateev: I don't think anywhere in our code base we use ...Interface for interface names. I would call it AuthorizationManager

tools/cli/factory.go
```diff
@@ -45,6 +45,15 @@ type WorkflowClientBuilderInterface interface {
...
+	//FilterUnauthorizedOperations decides which operation will be available in this CLI
+	FilterUnauthorizedOperations(*cli.App) *cli.App
```
mfateev: How does it do it? From the comment it is not clear at all. What about when an operation is allowed, but only for objects (like domain) that user owns?

tools/cli/commands.go
```diff
@@ -151,6 +151,14 @@ func SetBuilder(builder WorkflowClientBuilderInterface) {
 	cBuilder = builder
 }
 
+// cAuthorizationManager for manging permission
+// To provide customized builder, call SetAuthorizationManager() before call NewCliApp()
+var cAuthorizationManager AuthorizationManagerInterface
+
+func SetAuthorizationManager(m AuthorizationManagerInterface) {
+	cAuthorizationManager = m
```
mfateev: Is the lack of synchronization because the app is single threaded?

PR #864, opened 2018-06-19
add jitter to timer / transfer / replication queue sanity scan

wxing1292:


service/history/queueAckMgr_test.go
```diff
@@ -140,7 +140,10 @@ func (s *queueAckMgrSuite) SetupTest() {
 	}
 	s.mockShard.config.ShardUpdateMinInterval = dynamicconfig.GetDurationPropertyFn(0 * time.Second)
 
-	s.queueAckMgr = newQueueAckMgr(s.mockShard, &QueueProcessorOptions{UpdateShardTaskCount: 1, MetricScope: metrics.ReplicatorQueueProcessorScope}, s.mockProcessor, 0, s.logger)
+	s.queueAckMgr = newQueueAckMgr(s.mockShard, &QueueProcessorOptions{
+		UpdateShardTaskCount: func(opts ...dynamicconfig.FilterOption) int { return 1 },
```
vancexu: Could use dynamicconfig.GetIntPropertyFn() which is mock for tests.

service/history/queueAckMgr_test.go
```diff
@@ -346,7 +349,10 @@ func (s *queueFailoverAckMgrSuite) SetupTest() {
 	}
 	s.mockShard.config.ShardUpdateMinInterval = dynamicconfig.GetDurationPropertyFn(0 * time.Second)
 
-	s.queueFailoverAckMgr = newQueueFailoverAckMgr(s.mockShard, &QueueProcessorOptions{UpdateShardTaskCount: 1, MetricScope: metrics.ReplicatorQueueProcessorScope}, s.mockProcessor, 0, s.logger)
+	s.queueFailoverAckMgr = newQueueFailoverAckMgr(s.mockShard, &QueueProcessorOptions{
+		UpdateShardTaskCount: func(opts ...dynamicconfig.FilterOption) int { return 1 },
```
vancexu: Could use dynamicconfig.GetIntPropertyFn()

common/service/dynamicconfig/constants.go
```diff
@@ -177,6 +177,8 @@ const (
 	TimerProcessorCompleteTimerInterval
 	// TimerProcessorMaxPollInterval is max poll interval for timer processor
 	TimerProcessorMaxPollInterval
+	// TimerProcessorMaxPollIntervalJitterCoefficient is the max poll interval jitter coefficient
+	TimerProcessorMaxPollIntervalJitterCoefficient
```
vancexu: There new added keys also need to updated in `keys` map in this file.

service/history/queueProcessor.go
```diff
@@ -156,10 +160,10 @@ processorPumpLoop:
 			p.processBatch(tasksCh)
 		case <-pollTimer.C:
 			p.processBatch(tasksCh)
-			pollTimer.Reset(p.options.MaxPollInterval)
+			pollTimer.Reset(jitter.JitDuration(p.options.MaxPollInterval(), p.options.MaxPollIntervalJitterCoefficient()))
```
samarabbas: I think we want same logic to skip the forced scan if already performed recently.

PR #878, opened 2018-06-21
handle DomainNotActive error in timer / transfer queue

wxing1292:
solve  #843


service/history/queueProcessor.go
```diff
@@ -265,9 +268,20 @@ ProcessRetryLoop:
...
 					logging.LogTaskProcessingFailedEvent(p.logger, task.GetTaskID(), task.GetTaskType(), err)
-					backoff := time.Duration(retryCount * 100)
-					time.Sleep(backoff * time.Millisecond)
-					retryCount++
+
+					// it is possible that DomainNotActiveError is thrown
+					// just keep try for cache.DomainCacheRefreshInterval
+					// and giveup
+					switch err.(type) {
+					case *workflow.DomainNotActiveError:
```
samarabbas: This may cause retries in tight loop.

wxing1292: you are right

PR #879, opened 2018-06-22
Implement ListDomain API

longquanzheng:


service/frontend/workflowHandler.go
```diff
@@ -268,6 +268,50 @@ func (wh *WorkflowHandler) RegisterDomain(ctx context.Context, registerRequest *
...
+	listRequest *gen.ListDomainRequest) (*gen.ListDomainResponse, error) {
+	scope := metrics.FrontendListDomainScope
+	sw := wh.startRequestProfile(scope)
+	defer sw.Stop()
+
+	if listRequest == nil {
+		return nil, wh.error(errRequestNotSet, scope)
+	}
+
+	pageSize := 100
```
wxing1292: care to make this configurable? in the service.go file

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -802,8 +802,18 @@ struct RegisterDomainRequest {
   80: optional map<string,string> data
 }
 
+struct ListDomainRequest {
+  10: optional i32 pageSize
+  20: optional string nextPageToken
```
samarabbas: Can you binary blob for nextPageToken.

service/frontend/workflowHandler.go
longquanzheng: I would suggest not to use config file for this number unless you really need it.
1. User can always pass in a pageSize in the API. That would be the most use cases.
2. It is not necessary that we have to use different default page size in different environment(staging/sjc/dca/autobots/public repo/etc)
3. Having config will add more code and more configs, which would introduce lots of unnecessary code/config(staging/sjc/dca/autobots/public repo/etc). 
4. If having it in config, we still need to verify if the config value is set or not. If not, do you want to exit the process? Or use a another default value from code? Both look unnecessary. 

I am trying to understand why you want this. If this hardcoding number bothers you, I think we can use a constant number defined on the top of the file or somewhere else.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: Done

service/frontend/workflowHandler.go
wxing1292: if you take a look at the service.go file or at least the top of the file you are changing, you can see a lot of const, which makes it easy to manage.

defining a arbitrary number everywhere in the code is not a good practice.

PR #882, opened 2018-06-22
Task list query server implementation

meiliang86:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1083,6 +1084,22 @@ struct WorkflowQuery {
   20: optional binary queryArgs
 }
 
+struct Query {
```
mfateev: Either call it TaskListQuery or use existing WorkflowQuery struct.

PR #883, opened 2018-06-22
dynamic set the retry for retryable error on worker processor

wxing1292:


service/worker/processor.go
```diff
@@ -65,6 +94,12 @@ type (
 		metricsClient    metrics.Client
 		domainReplicator DomainReplicator
 		historyClient    history.Client
+
+		sync.Mutex
+		// worker in retry is used by underlying processor when doing retry on a task
+		// this help the replicator / underlying processor understanding the overall
+		// situation and giveup retrying
+		workerInRetry map[int]struct{}
```
samarabbas: Can we use atomic increment/decrement for keeping the retry count?  Then we don't need  lock

service/worker/processor.go
```diff
@@ -176,44 +260,62 @@ func (p *replicationTaskProcessor) worker(workerWG *sync.WaitGroup) {
...
+			if err != nil && p.isTransientRetryableError(err) {
+				// Keep on retrying transient errors for ever
+				continue ProcessRetryLoop
 			}
-		case <-p.consumer.Closed():
-			p.logger.Info("Consumer closed. Processor shutting down.")
-			return
 		}
+
+		p.updateWorkerStatus(workerID, workerStatusRunning)
```
samarabbas: I think it is much safer to do this as a defer to make sure all code paths which exit set the status back to running.

wxing1292: sure

service/worker/processor.go
wxing1292: done

PR #885, opened 2018-06-23
workflow/child_workflow retry

yiminc:
server side workflow/child_workflow retry


common/persistence/dataInterfaces.go
```diff
@@ -97,6 +97,7 @@ const (
 	TaskTypeWorkflowTimeout
 	TaskTypeDeleteHistoryEvent
 	TaskTypeRetryTimer
```
samarabbas: Can we rename this to TaskTypeActivityRetryTimer

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -417,6 +423,7 @@ struct WorkflowExecutionContinuedAsNewEventAttributes {
   50: optional i32 executionStartToCloseTimeoutSeconds
   60: optional i32 taskStartToCloseTimeoutSeconds
   70: optional i64 (js.type = "Long") decisionTaskCompletedEventId
+  80: optional i32 backoffStartIntervalInSeconds
```
mfateev: How is it different from retryPolicy.initialIntervalSeconds?

service/history/historyBuilder.go
```diff
@@ -477,6 +479,18 @@ func (b *historyBuilder) newWorkflowExecutionStartedEvent(
 	attributes.ChildPolicy = request.ChildPolicy
 	attributes.ContinuedExecutionRunId = previousRunID
 	attributes.Identity = common.StringPtr(common.StringDefault(request.Identity))
+	attributes.RetryPolicy = request.RetryPolicy
+	attributes.Attempt = common.Int32Ptr(startRequest.GetAttempt())
```
samarabbas: We should allow to have Attempt and expiration being nil.

service/history/historyBuilder.go
```diff
@@ -477,6 +479,18 @@ func (b *historyBuilder) newWorkflowExecutionStartedEvent(
...
+	if startRequest.GetAttempt() == 0 {
+		if request.RetryPolicy != nil && request.RetryPolicy.GetExpirationIntervalInSeconds() > 0 {
+			expirationInSeconds := request.RetryPolicy.GetExpirationIntervalInSeconds()
+			deadline := time.Unix(0, historyEvent.GetTimestamp()).Add(time.Second * time.Duration(expirationInSeconds))
+			attributes.ExpirationTimestamp = common.Int64Ptr(deadline.Round(time.Millisecond).UnixNano())
```
samarabbas: It doesn't feel right to have this logic at the history builder layer.  I think Expiration should either always be set by the client or history engine and passed in as the start request to history builder.

service/history/historyBuilder.go
```diff
@@ -477,6 +479,18 @@ func (b *historyBuilder) newWorkflowExecutionStartedEvent(
...
+		}
+	} else {
+		attributes.ExpirationTimestamp = common.Int64Ptr(startRequest.GetExpirationTimestamp())
```
samarabbas: just directly set attributes.ExpirationTimestamp = startRequest.ExpirationTimestamp

I think we need to support this field not being set at all.

service/history/mutableStateBuilder.go
```diff
@@ -2318,42 +2343,80 @@ func (e *mutableStateBuilder) ReplicateWorkflowExecutionContinuedAsNewEvent(sour
...
+	}}
+
+	if di != nil {
+		continueAsNew.DecisionVersion = di.Version
+		continueAsNew.DecisionScheduleID = di.ScheduleID
+		continueAsNew.DecisionStartedID = di.StartedID
+		continueAsNew.DecisionStartToCloseTimeout = di.DecisionTimeout
+
+		if newStateBuilder.GetReplicationState() != nil {
+			newStateBuilder.UpdateReplicationStateLastEventID(sourceClusterName, startedEvent.GetVersion(), di.ScheduleID)
```
samarabbas: We need to update replication state in both cases.  Regular case should use the di.ScheduleID and retry case should still set it to startedEventID.

service/history/timerQueueActiveProcessor.go
```diff
@@ -604,15 +647,70 @@ Update_History_Loop:
...
+			return err
+		}
+		_, continueAsNewBuilder, err := msBuilder.AddContinueAsNewEvent(common.EmptyEventID, domainEntry, startAttributes.GetParentWorkflowDomain(), continueAsnewAttributes)
+		if err != nil {
+			return err
+		}
+
+		tBuilder := t.historyService.getTimerBuilder(&context.workflowExecution)
+		var transferTasks, timerTasks []persistence.Task
+		tranT, timerT, err := getDeleteWorkflowTasksFromShard(t.shard, domainID, tBuilder)
```
samarabbas: Can we test to make sure we are not creating 2 separate delete tasks on workflow timeout.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -349,6 +349,8 @@ struct ContinueAsNewWorkflowExecutionDecisionAttributes {
   30: optional binary input
   40: optional i32 executionStartToCloseTimeoutSeconds
   50: optional i32 taskStartToCloseTimeoutSeconds
+  60: optional i32 backoffStartIntervalInSeconds
```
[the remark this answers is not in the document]
yiminc: task created to track this https://github.com/uber/cadence/issues/902

idl/github.com/uber/cadence/shared.thrift
yiminc: This is the backoff interval to create the first schedule event for next iteration (which may not be the first time of retry). The initialInterval is the backoff interval for first retry.

service/history/historyBuilder.go
yiminc: yes, if RetryPolicy is nil, it will just set the nil to attributes. The GetAttempt() will return 0 for nil case.

service/history/historyBuilder.go
yiminc: will do

service/history/mutableStateBuilder.go
yiminc: added

idl/github.com/uber/cadence/shared.thrift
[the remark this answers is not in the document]
yiminc: It is persist in the WorkflowExecutionStartedEvent of next iteration.

service/history/timerQueueActiveProcessor.go
yiminc: workflowTimeout does not go through this path.

PR #895, opened 2018-06-26
Fixes of tasklist frequently being loaded and unloaded

vancexu:


common/persistence/cassandraPersistence.go
```diff
@@ -780,15 +780,15 @@ const (
...
-	templateUpdateTaskListQueryWithTTL = `INSERT INTO tasks (` +
-		`domain_id, ` +
-		`task_list_name, ` +
-		`task_list_type, ` +
-		`type, ` +
-		`task_id, ` +
-		`range_id, ` +
-		`task_list ` +
-		`) VALUES (?, ?, ?, ?, ?, ?, ` + templateTaskListType + `) USING TTL ?`
+	templateUpdateTaskListQueryWithTTL = `UPDATE tasks USING TTL ? SET ` +
```
samarabbas: Changing INSERT into UPDATE may cause tasklist metadata entries to leak in DB.

Issue #697, opened 2018-04-25
[retry]: server side retry for child workflow


yiminc:

yiminc:
Implemented https://github.com/uber/cadence/commit/c18338477f1b56f87569cc239b10febbd69f682e

PR #907, opened 2018-06-28
retry expiration for activity

yiminc:
Properly set retry expiration for activity

Previously, we decided to use activity's Schedule_To_Close as expiration for entire retry process.  However, when we work on workflow/child_workflow retry, we realize we needed a retry policy expiration time. So we want to make it consistent across workflow and activity retry. Thus, we would apply this retryPolicy's expiration time to activity retry. If the expiration is not supplied, there will


dmetzgar:
Waiting on #697

PR #911, opened 2018-06-28
Fix leaking goroutine in matching

vancexu:


service/matching/taskListManager.go
```diff
@@ -349,12 +349,24 @@ func (c *taskListManagerImpl) Stop() {
...
 	close(c.deliverBufferShutdownCh)
 	c.cancelFunc()
+	c.drainTaskList()
 	close(c.shutdownCh)
 	c.taskWriter.Stop()
 	c.engine.removeTaskListManager(c.taskListID)
 	logging.LogTaskListUnloadedEvent(c.logger)
 }
 
+func (c *taskListManagerImpl) drainTaskList() {
```
samarabbas: Instead of having a separate drain pump can we have a sent to buffer in a select with another case to unblock when shutdown is triggered.

Issue #912, opened 2018-06-29
Buffered replication task is not correctly deleted


wxing1292:
for loop delete replication tasks: https://github.com/uber/cadence/blob/master/service/history/historyReplicator.go#L426

mutable state only record the last one to be deleted:
https://github.com/uber/cadence/blob/master/service/history/mutableStateBuilder.go#L646

PR #913, opened 2018-06-29
Bugfix: buffered replication task is not correctly deleted

wxing1292:
solve #912


service/history/mutableStateBuilder.go
```diff
@@ -255,7 +255,7 @@ type (
 		newBufferedEvents                *persistence.SerializedHistoryEventBatch
 		clearBufferedEvents              bool
 		newBufferedReplicationEventsInfo *persistence.BufferedReplicationTask
-		deleteBufferedReplicationEvent   *int64
+		deleteBufferedReplicationEvent   []int64
```
samarabbas: deleteBufferedReplicationEvents?

common/persistence/dataInterfaces.go
```diff
@@ -574,22 +574,22 @@ type (
...
+		UpsertRequestCancelInfos       []*RequestCancelInfo
+		DeleteRequestCancelInfo        *int64
+		UpsertSignalInfos              []*SignalInfo
+		DeleteSignalInfo               *int64
+		UpsertSignalRequestedIDs       []string
+		DeleteSignalRequestedID        string
+		NewBufferedEvents              *SerializedHistoryEventBatch
+		ClearBufferedEvents            bool
+		NewBufferedReplicationTask     *BufferedReplicationTask
+		DeleteBufferedReplicationTasks []int64
```
samarabbas: I thought we only process single buffered event in each transaction.  What is the use case for processing more than one buffered replication task?

wxing1292: https://github.com/uber/cadence/blob/master/service/history/historyReplicator.go#L419

service/history/mutableStateBuilder.go

wxing1292:
looks like for each buffered event, we do transaction to delete it and apply it, close this PR

PR #915, opened 2018-06-29
Fix ContinueAsNew event replication

yiminc:
The workflow retry change break the ContinueAsNew event replication.
The reason is the mutableState.ReplicateWorkflowExecutionContinuedAsNewEvent() was changed to create the DecisionTaskScheduled event within it. This break the replicator's assumption that replication event only update mutable state and history event is handled by directly use the replication task's payload.

The fix is simply move back the decisionInfo creation from within ReplicateWorkflowExecutionContinuedAsNewEvent back to outside of it.


service/history/stateBuilder.go
```diff
@@ -350,39 +349,44 @@ func (b *stateBuilderImpl) applyEvents(domainID, requestID string, execution sha
...
 			newRunExecutionInfo.NextEventID = nextEventID
 			newRunExecutionInfo.LastFirstEventID = startedEvent.GetEventId()
 			// Set the history from replication task on the newStateBuilder
 			newRunStateBuilder.SetHistoryBuilder(newHistoryBuilderFromEvents(newRunHistory.Events, b.logger))
 			sourceClusterName := b.clusterMetadata.ClusterNameForFailoverVersion(startedEvent.GetVersion())
 			newRunStateBuilder.UpdateReplicationStateLastEventID(sourceClusterName, startedEvent.GetVersion(), nextEventID-1)
 
-			b.newRunTransferTasks = append(b.newRunTransferTasks, b.scheduleDecisionTransferTask(domainID,
-				b.getTaskList(newRunStateBuilder), di.ScheduleID))
+			if startedAttributes.GetAttempt() == 0 {
```
wxing1292: if attempt is !=0, then no decision is scheduled?
this can be fine if the workflow is being replicated, however, when doing a failover and the new attempt workflow is picked up by the previous standby cluster (now active), the no worker can pick this workflow up. 
this workflow will be invisible until the decision task scheduled event is replicated from active cluster (now standby after the failover). however, there is no guarantee that this decision task scheduled event will be replicated.

service/history/historyEngine.go
```diff
@@ -1392,6 +1423,9 @@ Update_History_Loop:
 		// the history and try the operation again.
 		var updateErr error
 		if continueAsNewBuilder != nil {
+			if msBuilder.GetContinueAsNew() != nil {
```
wxing1292: i think this will always be non nil after checking the continueAsNewBuilder != nil
ref: https://github.com/uber/cadence/pull/915/files#diff-01045b0895719962d1f13440e8795d1bR2416

service/history/stateBuilder.go
yiminc: the first scheduled event will be created by a backoff timer, the timer should fire and because it is now become active, it would create the scheduled event locally.

yiminc: I think the stand by timer queue process will need to be updated to handle that case. will update it.

wxing1292: @yiminc the standby processor only does verification, as long as the timer task is created, the failover will handle the task processing

wxing1292: the failover processing logic is the same as the active processing logic

PR #916, opened 2018-06-30
Fix check idle tasklist to include active task writers

vancexu:
Currently in matching getTaskPump there is a timer to check whether tasklist is "idle" and need to recycle. But it only treat no poller within 5 min as idle to recycle. 

This PR add check idle to include if there are tasks added recently (5 min).

The default live time it currently hardcoded as 5 min for both poller and writer. Which can be make configurable if necessary.

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -51,6 +51,8 @@ const (
 const (
 	_defaultTaskDispatchRPS    = 100000.0
 	_defaultTaskDispatchRPSTTL = 60 * time.Second
+
+	maxAddTasksIdleTime = 5 * time.Minute
 )
 
 var errAddTasklistThrottled = errors.New("cannot add to tasklist, limit exceeded")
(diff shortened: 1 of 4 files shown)


service/matching/taskListManager.go
```diff
@@ -338,6 +343,7 @@ func (c *taskListManagerImpl) Start() error {
 	c.taskWriter.Start()
 	c.signalNewTask()
 	go c.getTasksPump()
+	go c.checkIdle()
```
samarabbas: Why do we need a separate goroutine for idle check?  why can't be done in the same pump like we do for zero pollers check?

service/matching/taskListManager.go
```diff
@@ -949,3 +969,16 @@ func (c *taskContext) completeTask(err error) {
 func createServiceBusyError(msg string) *s.ServiceBusyError {
 	return &s.ServiceBusyError{Message: msg}
 }
+
+func (c *taskListManagerImpl) updateLastAddTaskTime() {
+	c.lastAddTaskTimeLock.Lock()
```
samarabbas: I think we should do this without adding another lock.

service/matching/taskListManager.go
vancexu: Changed, previously because I try to avoid this check potentially block getTaskPump

service/matching/taskListManager.go
vancexu: Update with better solution in getTaskPump

service/matching/taskListManager.go
```diff
@@ -51,6 +51,8 @@ const (
 const (
 	_defaultTaskDispatchRPS    = 100000.0
 	_defaultTaskDispatchRPSTTL = 60 * time.Second
+
+	maxAddTasksIdleTime = 5 * time.Minute
```
samarabbas: Can you convert this to a config knob.

vancexu: Done

PR #917, opened 2018-07-02
Fix noisy log

wxing1292:


service/history/historyReplicator.go
```diff
@@ -125,6 +125,9 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
 			case *shared.WorkflowExecutionAlreadyStartedError:
 				logger.Debugf("Encounter WorkflowExecutionAlreadyStartedError: %v", retError)
 				retError = ErrRetryExecutionAlreadyStarted
+			case *persistence.WorkflowExecutionAlreadyStartedError:
```
samarabbas: I saw a lot of errors leaked on frontend and that too on the active cluster.  So my guess is there is another place we are leaking this error.

PR #919, opened 2018-07-03
Bugfix: deadlock when domain failover

wxing1292:
* fix deadlock issue in timer queue processor failover
* fix domain cache copy without locking issue

### DIFF
--- a/common/cache/domainCache.go
+++ b/common/cache/domainCache.go
@@ -173,8 +173,11 @@ func (c *domainCache) GetAllDomain() map[string]*DomainCacheEntry {
 	for ite.HasNext() {
 		entry := ite.Next()
 		id := entry.Key().(string)
-		domainCacheEntry := entry.Value().(*DomainCacheEntry).duplicate()
-		result[id] = domainCacheEntry
+		domainCacheEntry := entry.Value().(*DomainCacheEntry)
+		domainCacheEntry.RLock()
+		dup := domainCacheEntry.duplicate()
+		domainCacheEntry.RUnlock()
+		result[id] = dup
 	}
 	return result
 }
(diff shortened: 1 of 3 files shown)


service/history/timerQueueProcessor.go
```diff
@@ -143,7 +143,9 @@ func (t *timerQueueProcessorImpl) FailoverDomain(domainID string) {
 	failoverTimerProcessor := newTimerQueueFailoverProcessor(t.shard, t.historyService, domainID,
 		standbyClusterName, minLevel, maxLevel, t.matchingClient, t.logger)
 	failoverTimerProcessor.Start()
-	failoverTimerProcessor.timerQueueProcessorBase.readAndFanoutTimerTasks()
+	// use a fake timer to trigger db scan
```
samarabbas: 'readAndFanoutTimerTasks' is still called from other places and I thought that is the one which could block.

common/cache/domainCache.go
```diff
@@ -185,13 +188,13 @@ func (c *domainCache) GetAllDomain() map[string]*DomainCacheEntry {
 // afterCallback will be invoked when NOT holding the domain cache lock.
 func (c *domainCache) RegisterDomainChangeCallback(shard int, initialNotificationVersion int64, beforeCallback CallbackFn, afterCallback CallbackFn) {
 	c.Lock()
-	defer c.Unlock()
-
 	c.beforeCallbacks[shard] = beforeCallback
 	c.afterCallbacks[shard] = afterCallback
+	domainNotificationVersion := c.domainNotificationVersion
+	c.Unlock()
```
samarabbas: Looks like both before and after callbacks were called within the lock before this change and now they are called outside of the lock.  Do we even need 2 separate callbacks?

service/history/timerQueueProcessor.go
wxing1292: readAndFanoutTimerTasks will be called by an separate go routine, so we should be fine.

common/cache/domainCache.go
wxing1292: we do not really need a 2 type of callbacks, one before one after, i thought providing 2 type of callbacks can be useful

PR #920, opened 2018-07-03
Apply 5 min delay for standby task

### LINKED ISSUE #914: Standby timer / transfer task are not correctly delayed.

wxing1292:
fix #914 

transfer task will be delayed by at most 5 min (meaning the delay can be less than 5 min, i.e. if the standby task can be discarded, or wait until 5 min and re-check)
timer task will be delayed by 5 min.


service/history/queueProcessor.go
```diff
@@ -283,7 +283,13 @@ ProcessRetryLoop:
 			if err != nil {
 				if err == ErrTaskRetry {
 					p.metricsClient.IncCounter(p.options.MetricScope, metrics.HistoryTaskStandbyRetryCounter)
-					<-notificationChan
+				DelayLoop:
```
samarabbas: Looks like we will delay the task after processing for first time.  I thought the idea is to even delay on first processing.

service/history/historyEngine.go
```diff
@@ -237,7 +237,7 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
 				// its length > 0 and has correct timestamp, to trkgger a db scan
 				fakeDecisionTask := []persistence.Task{&persistence.DecisionTask{}}
 				fakeDecisionTimeoutTask := []persistence.Task{&persistence.DecisionTimeoutTask{VisibilityTimestamp: now}}
-				e.txProcessor.NotifyNewTask(e.currentClusterName, now, fakeDecisionTask)
+				e.txProcessor.NotifyNewTask(e.currentClusterName, fakeDecisionTask)
```
samarabbas: I think we should have the exact same contract on moving the clock on both timer and transfer processors.  Timer notification takes in the currentTime as notification while transfer does not.  I think we should unify these.

service/history/queueProcessor.go
wxing1292: for transfer task, there is no timestamp associated with it.
we can tag the queue processor, however, that require more code.

service/history/historyEngine.go
wxing1292: transfer task do not have timestamp associated with it

common/persistence/cassandraPersistence.go
```diff
@@ -136,6 +136,7 @@ const (
 		`create_request_id: ?, ` +
 		`decision_version: ?, ` +
 		`decision_schedule_id: ?, ` +
+		`decision_schedule_time: ?, ` +
```
samarabbas: Probably we should just have a visibility_ts on transfer task for we have the same mechanism for each kind of task for delaying them.

service/history/historyEngine.go
```diff
@@ -357,6 +357,7 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 		}
 	}
 	setTaskVersion(msBuilder.GetCurrentVersion(), transferTasks, timerTasks)
+	setTransferTaskTimestamp(common.NewRealTimeSource().Now(), transferTasks)
```
samarabbas: Can we just have a single API for setting task information, instead of making 2 separate API calls.

service/history/historyEventNotifier.go
```diff
@@ -177,7 +177,7 @@ func (notifier *historyEventNotifierImpl) dispatchHistoryEventNotification(event
 
 func (notifier *historyEventNotifierImpl) enqueueHistoryEventNotification(event *historyEventNotification) {
 	// set the timestamp just before enqueuing the event
-	event.timestamp = time.Now()
+	event.timestamp = common.NewRealTimeSource().Now()
```
samarabbas: Can you have time source as a field on historyEventNotifierImpl so another implementation could be injected.

service/history/workflowExecutionContext.go
```diff
@@ -187,12 +187,13 @@ func (c *workflowExecutionContext) updateWorkflowExecution(transferTasks []persi
 			c.msBuilder.GetExecutionInfo().LastFirstEventID, c.msBuilder.GetExecutionInfo().NextEventID)
 	}
 
-	return c.updateHelper(nil, transferTasks, timerTasks, c.createReplicationTask, "", currentVersion, transactionID)
+	now := common.NewRealTimeSource().Now()
```
samarabbas: Lets call the API UtcNow and have a timesource defined on workflowExecutionContext

service/history/MockMutableState.go
```diff
@@ -1347,6 +1347,28 @@ func (_m *mockMutableState) GetHistoryEvent(serializedEvent []byte) (*shared.His
 	return r0, r1
 }
 
+func (_m *mockMutableState) GetInFlightDecisionTask() (*decisionInfo, bool) {
```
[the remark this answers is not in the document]
wxing1292: i did

PR #927, opened 2018-07-05
Add protection to standby cluster against randomized remote cluster time.

wxing1292:
* fix noisy log
* fix remote timer gate issue, making remote timer resilient to new timer in the past
add more unit test
* refactor existing time processing event loop to make it less confusing
* add rate limiter to timer queue processor, set all max tps from 100 to 20

### DIFF
--- a/service/history/queueProcessor.go
+++ b/service/history/queueProcessor.go
@@ -63,6 +63,8 @@ type (
 		// worker coroutines notification
 		workerNotificationChans []chan struct{}
 
+		lastPollTime time.Time
+
 		notifyCh   chan struct{}
 		status     int32
 		shutdownWG sync.WaitGroup
@@ -92,6 +94,7 @@ func newQueueProcessorBase(shard ShardContext, options *QueueProcessorOptions, p
 		metricsClient:           shard.GetMetricsClient(),
 		logger:                  logger,
 		ackMgr:                  queueAckMgr,
+		lastPollTime:            time.Time{},
 	}
 
 	return p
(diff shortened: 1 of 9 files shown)


service/history/queueProcessor.go
```diff
@@ -92,6 +94,7 @@ func newQueueProcessorBase(shard ShardContext, options *QueueProcessorOptions, p
 		metricsClient:           shard.GetMetricsClient(),
 		logger:                  logger,
 		ackMgr:                  queueAckMgr,
+		lastPollTime:            time.Time{},
```
samarabbas: Let's keep this within pump rather than exposing it as a field.  
You can return a boolean from processBatch to indicate if it actually polled or not.

service/history/timerQueueProcessorBase.go
```diff
@@ -319,105 +330,90 @@ func (t *timerQueueProcessorBase) notifyNewTimers(timerTasks []persistence.Task)
...
 }
 
 func (t *timerQueueProcessorBase) readAndFanoutTimerTasks() (*persistence.TimerTaskInfo, error) {
-	for {
-		// Get next set of timer tasks.
-		timerTasks, lookAheadTask, moreTasks, err := t.timerQueueAckMgr.readTimerTasks()
-		if err != nil {
-			return nil, err
-		}
+	if !t.rateLimiter.Consume(1, t.shard.GetConfig().TransferProcessorMaxPollInterval()) {
```
samarabbas: shouldn't this be TimerProcessorMaxPollInterval?

PR #930, opened 2018-07-05
CLI: support domain data operation for xdc

longquanzheng:


tools/cli/app.go
```diff
@@ -25,7 +25,7 @@ import "github.com/urfave/cli"
 const (
 	// Version is the controlled version string. It should be updated every time
 	// before we release a new version.
-	Version = "0.5.3"
+	Version = "0.5.4"
```
vancexu: Would be nicer to make a separate pull request before release

PR #936, opened 2018-07-06
Worker should retry on network error

wxing1292:


service/worker/processor.go
```diff
@@ -443,18 +443,10 @@ func (p *replicationTaskProcessor) updateFailureMetric(scope int, err error) {
...
-	case *shared.ServiceBusyError:
-		return true
-	case *shared.LimitExceededError:
-		return true
-	case *shared.InternalServiceError:
-		return true
-	case *shared.RetryTaskError:
-		return true
-	default:
+	case *shared.BadRequestError:
```
samarabbas: process also deserializes the payload.  Any deserialization error should also cause the message to immediately land in DLQ.

PR #940, opened 2018-07-06
Handle service deployment causing buffer events not flushed

wxing1292:


service/history/historyReplicator.go
```diff
@@ -182,6 +182,11 @@ func (r *historyReplicator) ApplyEvents(request *h.ReplicateEventsRequest) (retE
 			return r.ApplyOtherEventsMissingMutableState(domainID, request.WorkflowExecution.GetWorkflowId(), firstEvent.GetVersion(), logger)
 		}
 
+		err = r.FlushBuffer(context, msBuilder, logger)
+		if err != nil {
+			logger.Errorf("Fail to pre-flush buffer, Err: %v", err)
```
samarabbas: Can you emit a TagErr rather than using doing an Errorf.

PR #945, opened 2018-07-09
Simplify replication worker retry logic

samarabbas:
Current retry logic can result in some replication tasks to be retried
indifinitely if we have bugs in our history replication code.
Updated the logic to make it much simpler.  It now uses a retryable
history client to retry any history service transient errors before
returning the error back to processWithRetry pump.  The pump will keep
on retrying whiltelisted transient errors for ever as it makes no sense
to move those messages to DLQ, otherwise it decrements the remaining
retry count and continue applying the message until fixed value of max
retry count is hit for that task.

### DIFF
--- a/service/worker/processor.go
+++ b/service/worker/processor.go
@@ -21,14 +21,11 @@
 package worker
 
 import (
-	"math"
+	"context"
+	"encoding/json"
 	"sync"
 	"sync/atomic"
 	"time"
-
-	"encoding/json"
-
-	"context"
 
 	"github.com/uber-common/bark"
 	"github.com/uber-go/kafka-client/kafka"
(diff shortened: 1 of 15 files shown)


service/worker/processor.go
```diff
@@ -267,54 +196,55 @@ func (p *replicationTaskProcessor) messageProcessLoop(workerWG *sync.WaitGroup,
...
+	remainingRetryCount := p.config.ReplicationTaskMaxRetry
 
 ProcessRetryLoop:
-	for {
+	for attempt := 0; ; attempt++ {
 		select {
 		case <-p.shutdownCh:
 			return
 		default:
 			op := func() error {
```
wxing1292: op should be move outside of the for loop

service/worker/processor.go
```diff
@@ -456,6 +393,14 @@ func (p *replicationTaskProcessor) updateFailureMetric(scope int, err error) {
 	}
 }
 
+func (p *replicationTaskProcessor) isRetryTaskError(err error) bool {
+	if _, ok := err.(*shared.RetryTaskError); ok {
```
wxing1292: maybe this should also include the timeout error

samarabbas: Timeout is already covered by transient retry policy.

service/worker/processor.go
samarabbas: Good point.  Moved outside.

PR #946, opened 2018-07-09
Add maximum timeout protection

vancexu:
Currently we do not limit workflow timeout for customers. But because Cassandra TTL maximum is 630720000 (20 years) and if customer put a number larger than that, it will cause errors like "RecordWorkflowExecutionStarted operation failed" , and CreateTask failure. 

This PR protects:
1. Task ScheduleToClose
2. Visibility TTL

### DIFF
--- a/common/logging/helpers.go
+++ b/common/logging/helpers.go
@@ -432,3 +432,12 @@ func LogCriticalErrorEvent(logger bark.Logger, msg string, err error) {
 		TagWorkflowErr:     err,
 	}).Error(msg)
 }
+
+// LogDecisionTimeoutTooLarge is used to log warning msg for workflow that contains large decision timeout
+func LogDecisionTimeoutTooLarge(logger bark.Logger, t int32, domain, wid, wfType string) {
+	logger.WithFields(bark.Fields{
+		"Domain":       domain,
+		"WorkflowID":   wid,
+		"WorkflowType": wfType,
+	}).Warnf("Decision timeout %d is too large", t)
+}
--- a/common/logging/helpers.go
+++ b/common/logging/helpers.go
@@ -436,8 +436,9 @@ func LogCriticalErrorEvent(logger bark.Logger, msg string, err error) {
 // LogDecisionTimeoutTooLarge is used to log warning msg for workflow that contains large decision timeout
 func LogDecisionTimeoutTooLarge(logger bark.Logger, t int32, domain, wid, wfType string) {
 	logger.WithFields(bark.Fields{
-		"Domain":       domain,
-		"WorkflowID":   wid,
-		"WorkflowType": wfType,
-	}).Warnf("Decision timeout %d is too large", t)
+		"Domain":          domain,
+		"WorkflowID":      wid,
+		"WorkflowType":    wfType,
+		"DecisionTimeout": t,
+	}).Warn("Decision timeout is too large")
 }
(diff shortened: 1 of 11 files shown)


service/frontend/workflowHandler.go
```diff
@@ -1349,11 +1349,21 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 			Message: "A valid ExecutionStartToCloseTimeoutSeconds is not set on request."}, scope)
 	}
 
+	if startRequest.GetExecutionStartToCloseTimeoutSeconds() > common.MaxWorkflowTimeout {
```
samarabbas: I think we decided to not restrict workflow timeout and instead have a max limit on decision task timeout of 1 year.

vancexu: Updated.

service/history/historyEngine.go
```diff
@@ -2449,6 +2449,13 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
 		return &workflow.BadRequestError{Message: "A valid timeout may not be negative."}
 	}
 
+	if attributes.GetScheduleToCloseTimeoutSeconds() > common.MaxWorkflowTimeout ||
```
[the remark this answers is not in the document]
vancexu: Make sense, changed

common/logging/helpers.go
```diff
@@ -432,3 +432,12 @@ func LogCriticalErrorEvent(logger bark.Logger, msg string, err error) {
...
 	}).Error(msg)
 }
+
+// LogDecisionTimeoutTooLarge is used to log warning msg for workflow that contains large decision timeout
+func LogDecisionTimeoutTooLarge(logger bark.Logger, t int32, domain, wid, wfType string) {
+	logger.WithFields(bark.Fields{
+		"Domain":       domain,
+		"WorkflowID":   wid,
+		"WorkflowType": wfType,
+	}).Warnf("Decision timeout %d is too large", t)
```
samarabbas: Either use an eventID for this log or don't use Warnf and add timeout as a tag.

common/persistence/cassandraVisibilityPersistence.go
```diff
@@ -150,7 +152,7 @@ func (v *cassandraVisibilityPersistence) Close() {
 
 func (v *cassandraVisibilityPersistence) RecordWorkflowExecutionStarted(
 	request *RecordWorkflowExecutionStartedRequest) error {
-	ttl := request.WorkflowTimeout + openExecutionTTLBuffer
+	ttl := common.MinInt64(request.WorkflowTimeout+openExecutionTTLBuffer, maxCassandraTTL)
```
samarabbas: Open visibility record will be deleted after this TTL.  It will be very confusing if we allow WorkflowTimeout of greater than 20 years but only allow visibility record for upto 20 years max.  I think we should just drop the TTL part from the query if Workflow Timeout is bigger than the supported TTL.

service/history/historyEngine.go
```diff
@@ -2463,6 +2463,13 @@ func validateActivityScheduleAttributes(attributes *workflow.ScheduleActivityTas
 		return &workflow.BadRequestError{Message: "A valid timeout may not be negative."}
 	}
 
+	if attributes.GetScheduleToCloseTimeoutSeconds() > wfTimeout ||
```
samarabbas: Rather than failing decision tasks can we reduce the timeout to be the same as workflow timeout?
I'm worried this might cause issues in production.

PR #949, opened 2018-07-10
Kafka dlq tool

longquanzheng:
Same example of my local tests:
```
./cadence-kafka-tool --local mergeDLQ --dlqTopic active-dlq --destTopic active
```
```
./cadence-kafka-tool --local reset --topic active  --consumerGroup DLQConsumerGroup --offsets 0:1,1:1
```
```
./cadence-kafka-tool -f tools/kafka/kafka_cluster.yaml purge --topic active  --consumerGroup DLQConsumerGroup --offsets 0:8,1:8 --cluster test
partition 0 setting to offset 8
partition 1 setting to offset 8
mark offset for topic active group DLQConsumerGroup is completed
```

```
./kafka-consumer-groups.sh  --bootstrap-server localhost:9092  --group DLQConsumerGroup --describe
Note: This will not show information about old Zookeeper-based consumers.

Consumer group 'DLQConsumerGroup' has no active members.

TOPIC                          PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG        CONSUMER-ID                                       HOST                           CLIENT-ID
active                         0          9               24              15         -                                                 -                              -
active-dlq                     0          6               31              25         -                                                 -                              -
active                         1          9               3               -6         -                                                 -                              -
```

### DIFF
--- a/tools/kafka/commands.go
+++ b/tools/kafka/commands.go
@@ -208,7 +208,6 @@ outloop:
 
 	for p, off := range offsetPerPartition {
 		fmt.Printf("partition %v setting to offset %v \n", p, off)
-		consumer.MarkPartitionOffset(topic, p, off, "")
 		if isForward {
 			consumer.MarkPartitionOffset(topic, p, off, "")
 		} else {
--- a/tools/kafka/commands.go
+++ b/tools/kafka/commands.go
@@ -6,16 +6,14 @@ import (
 
 	"io/ioutil"
 
-	"github.com/uber-go/kafka-client"
-	"github.com/uber-go/kafka-client/kafka"
-	"github.com/uber-go/tally"
-	"go.uber.org/zap"
-
 	"strings"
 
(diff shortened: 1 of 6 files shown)


tools/kafka/commands.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	offsetStr := c.String(FlagOffsets)
+	offsets, err := parseOffsetStr(offsetStr)
+	if err != nil {
+		ErrorAndExit("input offset is not valid", err)
+	}
+	setTopicOffsets(c, offsets, false)
+}
+
+func parseOffsetStr(offsetStr string) (map[int32]int64, error) {
```
wxing1292: we should be able to get the current water mark and use it for merge and purge instead of manually providing the offset

longquanzheng:
Close this request since there is already the same tool in Uber

Issue #951, opened 2018-07-11
Heartbeat ReplicationTask from each shard 


samarabbas:
Each shard needs to generate an heartbeat replication task to move the clock forward for standby cursor on other clusters.

Issue #786, opened 2018-05-26
Cross DC - Add new heartbeat replication event 


wxing1292:
Add new heartbeat event so the time among clusters can by synced and timer / transfer standby cursor moved.

PR #952, opened 2018-07-11
Add time sync functionality, to move the view of time of a remote cluster

wxing1292:
* Add time sync functionality, to move the view of time of a remote cluster

solve #951 #786


service/history/handler.go
```diff
@@ -924,6 +924,30 @@ func (h *Handler) ReplicateEvents(ctx context.Context, replicateRequest *hist.Re
...
+// SyncShardStatus is called by processor to sync history shrad information from another cluster
+func (h *Handler) SyncShardStatus(ctx context.Context, syncShardStatusRequest *hist.SyncShardStatusRequest) error {
+	h.startWG.Wait()
+
+	h.metricsClient.IncCounter(metrics.HistorySyncShardStatusScope, metrics.CadenceRequests)
+	sw := h.metricsClient.StartTimer(metrics.HistorySyncShardStatusScope, metrics.CadenceLatency)
+	defer sw.Stop()
+
+	// shard ID is already provided in the request
+	engine, err := h.controller.getEngineForShard(int(syncShardStatusRequest.GetShardId()))
```
samarabbas: Can you validate if the request has shardID and return BadRequest if it does not.

service/history/historyEngine.go
```diff
@@ -2194,6 +2194,16 @@ func (e *historyEngineImpl) ReplicateEvents(replicateRequest *h.ReplicateEventsR
 	return e.replicator.ApplyEvents(replicateRequest)
 }
 
+func (e *historyEngineImpl) SyncShardStatus(ctx context.Context, request *h.SyncShardStatusRequest) error {
+	clusterName := request.GetSourceCluster()
+	now := time.Unix(0, request.GetTimestamp())
+
+	e.shard.SetCurrentTime(clusterName, now)
+	e.txProcessor.NotifyNewTask(clusterName, now, []persistence.Task{})
```
samarabbas: Do we really need this notification here?  I think we already have protection to trigger scan at regular bases.  Firing a notification to force a scan sounds redundant.

service/history/shardContext.go
```diff
@@ -585,6 +589,19 @@ func (s *shardContextImpl) updateShardInfoLocked() error {
 	}
 	updatedShardInfo := copyShardInfo(s.shardInfo)
 
+	if s.messageProducer != nil {
```
samarabbas: Can we only create sync task on idle scenario?

service/history/handler.go
wxing1292: sure

service/history/queueAckMgr.go
```diff
@@ -188,7 +188,6 @@ MoveAckLevelLoop:
 			break MoveAckLevelLoop
 		}
 	}
-	updateShard := a.ackLevel != ackLevel
```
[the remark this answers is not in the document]
wxing1292: we have a check on the actual update shard API call, which has min update interval = 5min

service/history/shardContext.go
wxing1292: #956

service/history/historyEngine.go
wxing1292: talked offline, this should be fine

PR #960, opened 2018-07-12
Few performance optimization.

wxing1292:
* Use exponential backoff for task processing
* Do not send shard time sync if has sent out replication task


service/history/shardContext.go
```diff
@@ -60,6 +60,8 @@ type (
 		UpdateTimerAckLevel(ackLevel time.Time) error
 		GetTimerClusterAckLevel(cluster string) time.Time
 		UpdateTimerClusterAckLevel(cluster string, ackLevel time.Time) error
+		GetLastShardSyncTimestamp() time.Time
```
samarabbas: I think we should simplify the logic and have generation of sync task limited to replicator processor.  This way we don't need all these API changes to ShardContext.

PR #961, opened 2018-07-12
Pass through request timeout to replicator code path

samarabbas:


service/history/historyReplicator.go
```diff
@@ -783,7 +794,7 @@ func (r *historyReplicator) terminateWorkflow(domainID string, workflowID string
 		return err
 	}
 	// same workflow ID, same shard
-	return r.historyEngine.TerminateWorkflowExecution(ctx.Background(), &h.TerminateWorkflowExecutionRequest{
+	return r.historyEngine.TerminateWorkflowExecution(context.Background(), &h.TerminateWorkflowExecutionRequest{
```
wxing1292: i think this should use the ctx passed in

PR #965, opened 2018-07-14
Adding more metrics, fix some bugs

wxing1292:
* Increase the throttle limit on history host to 9K
* Adding more metrics
* Fix bugs in retry policy

### DIFF
--- a/service/history/historyReplicator.go
+++ b/service/history/historyReplicator.go
@@ -23,7 +23,6 @@ package history
 import (
 	"context"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/pborman/uuid"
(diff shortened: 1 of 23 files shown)


service/history/historyReplicator.go
```diff
@@ -111,11 +114,17 @@ func (r *historyReplicator) ApplyEvents(ctx context.Context, request *h.Replicat
 		logging.TagWorkflowExecutionID: request.WorkflowExecution.GetWorkflowId(),
 		logging.TagWorkflowRunID:       request.WorkflowExecution.GetRunId(),
 		logging.TagSourceCluster:       request.GetSourceCluster(),
-		logging.TagVersion:             request.GetVersion(),
+		logging.TagIncomingVersion:     request.GetVersion(),
 		logging.TagFirstEventID:        request.GetFirstEventId(),
 		logging.TagNextEventID:         request.GetNextEventId(),
 	})
 
+	r.metricsClient.RecordTimer(
```
samarabbas: It seems strange to report replication events size as a timer which is generally used for computing latencies.  Is the intention here to compute various percentiles for this counter?

wxing1292: yes, i would like to get the distribution of the number of events per replication task, this can be useful for debugging or explaining the potential high latency issue

PR #970, opened 2018-07-17
Change timer processor logic to Fix timer skipped issue

vancexu:
### Problem
Cadence timer queue processor sometime skip processing timers, for example https://github.com/uber/cadence/issues/967 
The reason is history engine write tasks before read level, so timer queue will never pick those tasks. We need a way to guarantee that all tasks can be read at least once. 

#### Proposal 1: 
Add a timerLevel in shard. 
Every write is guarantee to be larger than timerLevel, every read use timerLevel as maxReadLevel, and will move timerLevel during each read. 
timerLevel is init with shard acklevel. 
When history write timers, it check if timer is larger than timerLevel, if not adjust it to timerLevel.
When timerQueue reads timers, it first update timerLevel to now + offset (offset is for safety). read till new timerLevel, and update minReadLevel to timerLevel after read. 
#### Proposal 2: serialize timer read and write
Add a readLevel in shard.
When history write timers, getReadLevel from shard and guarantees timer timestamp is larger than readLevel. 
When timerQueue reads timers, it getReadLevel from shard, reads from this value to upper (which is a big number), this guarantee created tasks never been skipped. After it process tasks, it updateReadLevel. The whole read process is locked by shard. 
This easy to implement but not efficient.

This pull request implement proposal 1. 
And it will address issue: https://github.com/uber/cadence/issues/967


service/history/shardContext.go
```diff
@@ -242,6 +244,13 @@ func (s *shardContextImpl) UpdateDomainNotificationVersion(domainNotificationVer
 	return s.updateShardInfoLocked()
 }
 
+func (s *shardContextImpl) UpdateActiveTimerReadLevel(readLevel time.Time) {
+	s.Lock()
```
vancexu: This might has performance impact on timerqueue read tasks, but if changed with property lock, there still might be race condition which requires more changes. I am looking into other possible solution

service/history/timerQueueAckMgr.go
```diff
@@ -205,6 +205,8 @@ TaskFilterLoop:
 		filteredTasks = append(filteredTasks, task)
 	}
 
+	t.shard.UpdateActiveTimerReadLevel(t.readLevel.VisibilityTimestamp)
```
wxing1292: first this does not deal with standby timers.
btw why we cannot use now?

service/history/timerQueueAckMgr.go
```diff
@@ -143,10 +139,9 @@ func newTimerQueueFailoverAckMgr(shard ShardContext, metricsClient metrics.Clien
 		timerQueueShutdown:  timerQueueShutdown,
 		outstandingTasks:    make(map[TimerSequenceID]bool),
 		ackLevel:            ackLevel,
-		readLevel:           ackLevel,
 		minQueryLevel:       ackLevel.VisibilityTimestamp,
 		pageToken:           nil,
-		maxQueryLevel:       maxLevel,
+		maxQueryLevel:       ackLevel.VisibilityTimestamp,
```
wxing1292: the max level for failover must be passed in

service/history/timerQueueAckMgr.go
```diff
@@ -64,8 +64,6 @@ type (
 		outstandingTasks map[TimerSequenceID]bool
 		// timer task ack level
 		ackLevel TimerSequenceID
-		// timer task read level
-		readLevel TimerSequenceID
```
wxing1292: the read level is used by failover, as far as i remember

service/history/timerQueueAckMgr.go
```diff
@@ -203,22 +201,22 @@ TaskFilterLoop:
 		}
 
 		if !t.isProcessNow(task.VisibilityTimestamp) {
-			lookAheadTask = task
+			lookAheadTask = task                       // this means there is task in the time range (now, now + offset)
+			t.maxQueryLevel = task.VisibilityTimestamp // adjust maxQueryLevel so that this task will be read next time
```
wxing1292: should this be the min query level?

service/history/timerQueueProcessorBase.go
```diff
@@ -360,6 +360,12 @@ func (t *timerQueueProcessorBase) readAndFanoutTimerTasks() (*persistence.TimerT
 	}
 
 	if !moreTasks {
+		if lookAheadTask == nil { // try to get lookAheadTask
+			lookAheadTask, err = t.timerQueueAckMgr.readLookAheadTask()
```
wxing1292: i believe you should combine this readLookAheadTask  function and readTimerTasks function.

service/history/timerQueueAckMgr.go
vancexu: changed

service/history/timerQueueAckMgr.go
vancexu: added back

service/history/timerQueueAckMgr.go
```diff
@@ -229,6 +227,23 @@ TaskFilterLoop:
 	return filteredTasks, lookAheadTask, moreTasks, nil
 }
 
+// read lookAheadTask from s.GetTimerMaxReadLevel to poll interval from there.
+func (t *timerQueueAckMgrImpl) readLookAheadTask() (*persistence.TimerTaskInfo, error) {
+	minQueryLevel := t.shard.GetTimerMaxReadLevel()
```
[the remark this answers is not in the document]
vancexu: changed

service/history/timerQueueActiveProcessor.go
```diff
@@ -104,7 +104,7 @@ func newTimerQueueActiveProcessor(shard ShardContext, historyService *historyEng
 }
 
 func newTimerQueueFailoverProcessor(shard ShardContext, historyService *historyEngineImpl, domainID string, standbyClusterName string,
-	minLevel time.Time, maxLevel time.Time, matchingClient matching.Client, logger bark.Logger) *timerQueueActiveProcessorImpl {
```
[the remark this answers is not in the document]
vancexu: added back

service/history/timerQueueProcessorBase.go
vancexu: updated

service/history/timerQueueAckMgr.go
```diff
@@ -113,10 +111,9 @@ func newTimerQueueAckMgr(scope int, shard ShardContext, metricsClient metrics.Cl
 		timerQueueShutdown:  func() error { return nil },
 		outstandingTasks:    make(map[TimerSequenceID]bool),
 		ackLevel:            ackLevel,
-		readLevel:           ackLevel,
```
wxing1292: should not delete this line

service/history/timerQueueAckMgr.go
```diff
@@ -143,7 +140,6 @@ func newTimerQueueFailoverAckMgr(shard ShardContext, metricsClient metrics.Clien
 		timerQueueShutdown:  timerQueueShutdown,
 		outstandingTasks:    make(map[TimerSequenceID]bool),
 		ackLevel:            ackLevel,
-		readLevel:           ackLevel,
```
wxing1292: same here

service/history/timerQueueAckMgr.go
```diff
@@ -174,11 +171,10 @@ func (t *timerQueueAckMgrImpl) readTimerTasks() ([]*persistence.TimerTaskInfo, *
 			return nil, nil, false, err
 		}
 		morePage = len(pageToken) != 0
-		t.logger.Debugf("readTimerTasks: ReadLevel: (%s) count: %v, more timer: %v", minQueryLevel, len(tasks), morePage)
+		t.logger.Debugf("readTimerTasks: minQueryLevel: (%s) count: %v, more timer: %v", minQueryLevel, len(tasks), morePage)
```
wxing1292: plz also include max level

PR #979, opened 2018-07-18
Bugfix replication protocol

wxing1292:
* Bugfix: when workflow is idle during the failover, no event will be generated, thus not replication info will be added; if a double failover is done for this workflow, the source cluster is not changed at all, so no replication info is needed.

* Add more metrics on worker

### DIFF
--- a/service/worker/processor.go
+++ b/service/worker/processor.go
@@ -199,7 +199,10 @@ func (p *replicationTaskProcessor) processWithRetry(msg kafka.Message, workerID 
 	forceBuffer := false
 	remainingRetryCount := p.config.ReplicationTaskMaxRetry
 
+	attempt := 0
+	startTime := time.Now()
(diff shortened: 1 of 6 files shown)


service/worker/processor.go
```diff
@@ -199,7 +199,10 @@ func (p *replicationTaskProcessor) processWithRetry(msg kafka.Message, workerID
 	forceBuffer := false
 	remainingRetryCount := p.config.ReplicationTaskMaxRetry
 
+	attempt := 0
```
samarabbas: We already have an attempt variable as part of for loop.  Can we just have a one instead of 2?

samarabbas: Probably you also want to include attempt on warn logs for every 100 failures.

wxing1292: all done

PR #983, opened 2018-07-19
When failover happen, and if there is an in flight decision, that decision should be invalidated.

wxing1292:


service/history/mutableStateBuilder.go
```diff
@@ -853,6 +853,18 @@ func (e *mutableStateBuilder) CreateNewHistoryEventWithTimestamp(eventType workf
...
 func (e *mutableStateBuilder) shouldBufferEvent(eventType workflow.EventType) bool {
+	// we need to make sure that if there is a decision in flight
+	// that decision has the same version as the current version
+	// otherwise, the decision is in flight in a different datacenter
+	if di, ok := e.GetInFlightDecisionTask(); ok {
+		if di.Version < e.GetCurrentVersion() {
+			// decision is in flight in a different data center
+			e.AddDecisionTaskFailedEvent(di.ScheduleID, di.StartedID,
+				workflow.DecisionTaskFailedCauseFailoverCloseDecision, nil, identityHistoryService)
+
```
wxing1292: should also add a new decision?

Issue #929, opened 2018-07-05
Cross DC - Reset should clear all irrelevant history


wxing1292:
```
~|⇒ cadence --address 127.0.0.1:8933 --do canary-test workflow show -w cron.workflow.sanity/workflow.sanity-4/workflow.concurrent-execution
    1  WorkflowExecutionStarted   (WorkflowType:(Name:workflow.concurrent-execution), TaskList:(Name:canary-task-queue), Input:[1530808803614241262      
                                  {"Env":"development","Deployment":"test","Domain":"canary-test","ServerHostPort":"127.0.0.1:7933","HTTPListenPort":0}  
                                  ], ExecutionStartToCloseTimeoutSeconds:360, TaskStartToCloseTimeoutSeconds:180)                                        
    2  DecisionTaskScheduled      (TaskList:(Name:canary-task-queue),                                                                                    
                                  StartToCloseTimeoutSeconds:180,                                                                                        
                                  Attempt:0)                                                                                                             
    3  DecisionTaskStarted        (ScheduledEventId:2,                                                                                                   
                                  Identity:14571@wenquanx-C02V91NKHTDG@canary-task-queue,                                                                
                                  RequestId:1ab79389-2ffa-4e40-9f1a-bc8d54ee07b1)                                                                        
    4  DecisionTaskCompleted      (ExecutionContext:[],                                                                                                  
                                  ScheduledEventId:2, StartedEventId:3,                                                                                  
                                  Identity:14571@wenquanx-C02V91NKHTDG@canary-task-queue)                                                                
    5  MarkerRecorded             (MarkerName:Version,                                                                                                   
                                  Details:["initial                                                                                                      
                                  version" 3 ],                                                                                                          
                                  DecisionTaskCompletedEventId:4)                                                                                        
    6  ActivityTaskScheduled      (ActivityId:0,                                                                                                         
                                  ActivityType:(Name:activity.concurrent-execution),                                                                     
                                  TaskList:(Name:canary-task-queue),                                                                                     
                                  Input:[1530808803827488913 ],                                                                                          
                                  ScheduleToCloseTimeoutSeconds:360,                                                                                     
                                  ScheduleToStartTimeoutSeconds:180,                                                                                     
                                  StartToCloseTimeoutSeconds:180,                                                                                        
                                  HeartbeatTimeoutSeconds:0,                                                                                             
                                  DecisionTaskCompletedEventId:4)                                                                                        
[...]

PR #988, opened 2018-07-20
Bugfix: reset workflow history should also clear stale history events

wxing1292:


common/persistence/cassandraHistoryPersistence.go
```diff
@@ -51,6 +51,13 @@ const (
 		`WHERE domain_id = ? ` +
 		`AND workflow_id = ? ` +
 		`AND run_id = ? `
+
+	templateDeleteWorkflowExecutionPartialHistory = `DELETE FROM events ` +
```
samarabbas: We need to protect partial delete using rangeID and txID.

common/persistence/cassandraHistoryPersistence.go
```diff
@@ -51,6 +51,13 @@ const (
...
+		`WHERE domain_id = ? ` +
+		`AND workflow_id = ? ` +
+		`AND run_id = ? ` +
+		`AND first_event_id >= ? ` +
+		`AND first_event_id < ?`
```
samarabbas: Lets not allow the user to pass upper bound for delete.  Current use case is to delete the entire tail, so we should only allow callers to pass in the start index.

common/persistence/dataInterfaces.go
```diff
@@ -800,6 +800,16 @@ type (
...
 	}
 
+	// DeleteWorkflowExecutionPartialHistoryRequest is used to delete partial workflow execution history
+	DeleteWorkflowExecutionPartialHistoryRequest struct {
+		DomainID  string
+		Execution workflow.WorkflowExecution
+		// Delete the history events from FirstEventID. Inclusive.
+		StartEventID int64
+		// Delete the history events upto NextEventID.  Not Inclusive.
+		EndEventID int64
```
samarabbas: Lets not allow the user to pass upper bound for delete. Current use case is to delete the entire tail, so we should only allow callers to pass in the StartEventID.

service/history/conflictResolver.go
```diff
@@ -85,9 +91,9 @@ func (r *conflictResolverImpl) reset(requestID string, replayEventID int64, star
 		// NextEventID could be in the middle of the batch.  Trim the history events to not have more events then what
 		// need to be applied
 		if batchSize > eventsToApply {
-			history.Events = history.Events[0:eventsToApply]
+			r.logError("Conflict resolution err mis-allignment history.", ErrCorruptedHistory)
```
samarabbas: Probably this error is now redundant.

common/persistence/dataInterfaces.go
wxing1292: sure

service/history/conflictResolver.go
wxing1292: ok removed

wxing1292: actually, i would like to keep this one, since when using ELK to go through error logs in history service, this can be helpful.

wxing1292:
close in favor of history version filtering. see #929

PR #994, opened 2018-07-23
Fix CLI show history to better display long text

vancexu:
Currently issue is when history has long text in argument such wf input or signal input, the output looks bad with line breaks. for example:
image

The reason is, table lib we use enable autowrap only when text are separate by spaces and newlines. 
This PR provides 3 fixes:
1. Add spaces to long text so that table can autowrap properly.
2. Limit max length 
3. Provide fully print option through -pf / -eid flag (also -maxl flag to custom max length).

Results as follows:
screen shot 2018-07-24 at 6 25 39 pm
screen shot 2018-07-24 at 6 26 31 pm
screen shot 2018-07-24 at 6 27 15 pm
screen shot 2018-07-24 at 6 43 12 pm

### DIFF
--- a/tools/cli/util.go
+++ b/tools/cli/util.go
@@ -196,14 +196,14 @@ func HistoryEventToString(e *s.HistoryEvent) string {
 		data = e
 	}
 
-	return anyToString(data)
+	return anyToString(data, false)
 }
 
-func anyToString(d interface{}) string {
+func anyToString(d interface{}, printFully bool) string {
 	v := reflect.ValueOf(d)
 	switch v.Kind() {
 	case reflect.Ptr:
-		return anyToString(v.Elem().Interface())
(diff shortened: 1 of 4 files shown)


tools/cli/util.go
```diff
@@ -235,24 +237,58 @@ func anyToString(d interface{}) string {
...
+			return fmt.Sprintf("[%v]", n)
 		}
 		return fmt.Sprintf("[len=%d]", v.Len())
 	default:
 		return fmt.Sprint(v.Interface())
 	}
 }
 
+// limit the maximum length for each field
+func trimInput(input string) string {
```
longquanzheng: Can you pick a better name than "input". This confuses me a while. Looks like input is one of the field value that may need to be trimmed. But it can be generic to any filed value.

vancexu: Changed

PR #1007, opened 2018-07-27
expose public tasklist in PollForDecisionTask

halakaraki:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -879,6 +879,7 @@ struct PollForDecisionTaskResponse {
   60: optional History history
   70: optional binary nextPageToken
   80: optional WorkflowQuery query
+  90: optional string publicTaskList
```
yiminc: use TaskList type here

idl/github.com/uber/cadence/matching.thrift
```diff
@@ -40,6 +40,7 @@ struct PollForDecisionTaskResponse {
   70: optional bool stickyExecutionEnabled
   80: optional shared.WorkflowQuery query
   90: optional shared.TransientDecisionInfo decisionInfo
+  100: optional string publicTaskList
```
yiminc: use TaskList type

idl/github.com/uber/cadence/shared.thrift
halakaraki: changing to TaskList

idl/github.com/uber/cadence/matching.thrift
halakaraki: changing to TaskList

PR #1008, opened 2018-07-27
Optimze matching getTasks with frequent break

vancexu:
When tasklist bounce between 2 hosts, it may wait efforts read tasks from readlevel to a big number. 
This PR change getTaskBatch to break big for loop in this case,  and let the caller (getTaskPump) to check whether tasklist is down or not.


service/matching/taskListManager.go
```diff
@@ -562,26 +562,35 @@ func (c *taskListManagerImpl) CancelPoller(pollerID string) {
 
 // Returns a batch of tasks from persistence starting form current read level.
 // Also return a number that can be used to update readLevel
-func (c *taskListManagerImpl) getTaskBatch() ([]*persistence.TaskInfo, int64, error) {
+func (c *taskListManagerImpl) getTaskBatch() ([]*persistence.TaskInfo, int64, bool, error) {
 	var tasks []*persistence.TaskInfo
 	readLevel := c.taskAckManager.getReadLevel()
 	maxReadLevel := c.taskWriter.GetMaxReadLevel()
+	times := 0
+	isReadBatchDone := true
```
samarabbas: Let's simplify the loop logic as discussed.

service/matching/taskListManager.go
```diff
@@ -562,26 +562,35 @@ func (c *taskListManagerImpl) CancelPoller(pollerID string) {
...
 		}
 		// return as long as it grabs any tasks
 		if len(tasks) > 0 {
-			return tasks, upper, nil
+			return tasks, upper, isReadBatchDone, nil
 		}
 		readLevel = upper
+
+		times++
+		if times >= 100 {
```
samarabbas: Let's add code to return back every 10 iteration.

service/matching/taskListManager.go
```diff
@@ -562,26 +562,35 @@ func (c *taskListManagerImpl) CancelPoller(pollerID string) {
 
 // Returns a batch of tasks from persistence starting form current read level.
 // Also return a number that can be used to update readLevel
-func (c *taskListManagerImpl) getTaskBatch() ([]*persistence.TaskInfo, int64, error) {
+func (c *taskListManagerImpl) getTaskBatch() ([]*persistence.TaskInfo, int64, bool, error) {
 	var tasks []*persistence.TaskInfo
 	readLevel := c.taskAckManager.getReadLevel()
```
samarabbas: I think we need to make sure read level is initialized from last checkpointed ack level from task list metadata.

service/matching/taskListManager.go
vancexu: simplified

service/matching/taskListManager.go
vancexu: Changed

service/matching/taskListManager.go
vancexu: Yes we are sure read level will be init with ack level when start

PR #1015, opened 2018-07-29
Add service protection to signal input size

wxing1292:
solve #1013


wxing1292:
validated everywhere except the place which really requires

vancexu:
We probably need to update go docs to reflect this limit, since it a new added hard limit.

PR #1025, opened 2018-08-02
Handle history replication race case

wxing1292:
* When all event batches generated fo a given version (e.g. generated by local cluster) are rejected by remote cluster when remote cluster become active, local cluster should properly reset local history.


service/history/historyReplicator.go
```diff
@@ -868,6 +894,51 @@ func (r *historyReplicator) terminateWorkflow(ctx context.Context, domainID stri
 	})
 }
 
+func (r *historyReplicator) getLatestCheckpoint(replicationInfo1 map[string]*h.ReplicationInfo,
```
samarabbas: Can you rename replicationInfo1 and replicationInfo2 to replicationInfoLocal and replicationInfoRemote?

service/history/historyReplicator.go
```diff
@@ -868,6 +894,51 @@ func (r *historyReplicator) terminateWorkflow(ctx context.Context, domainID stri
...
+
+	for _, ri := range replicationInfo1 {
+		if lastValidVersion == common.EmptyVersion || ri.GetVersion() > lastValidVersion {
+			lastValidVersion = ri.GetVersion()
+			lastValidEventID = ri.GetLastEventId()
+		}
+	}
+
+	for _, ri := range replicationInfo2 {
+		if lastValidVersion == common.EmptyVersion || ri.Version > lastValidVersion {
```
samarabbas: ri.GetVersion?

service/history/historyReplicator.go
wxing1292: sure

PR #1031, opened 2018-08-05
Auto fix #1030

wxing1292:


common/persistence/cassandraPersistence.go
```diff
@@ -1162,7 +1167,7 @@ func (d *cassandraPersistence) CreateWorkflowExecutionWithinBatch(request *Creat
 			permanentRunID,
 			defaultVisibilityTimestamp,
 			rowTypeExecutionTaskID,
-			request.PreviousRunID,
+			prevRunID,
```
yiminc: data type changed from string to *string, is this expected? why do we make this change?

service/history/historyEngine.go
```diff
@@ -426,6 +426,10 @@ func (e *historyEngineImpl) StartWorkflowExecution(startRequest *h.StartWorkflow
 		prevState := err.State
 		prevCloseState := err.CloseStatus
 
+		if prevRunID == "" {
```
yiminc: why is this change needed?

PR #1033, opened 2018-08-08
Handle potential coroutine leak issue

wxing1292:


service/history/timerGate.go
```diff
@@ -104,8 +104,11 @@ func NewLocalTimerGate() LocalTimerGate {
 		for {
 			select {
 			case <-timer.timer.C:
+				select {
 				// re-transmit on gateC
-				timer.fireChan <- struct{}{}
+				case timer.fireChan <- struct{}{}:
+				default:
```
dmetzgar: Still new to go. Will this prevent a send to the `fireChan` channel if there is already something in the channel?

service/history/timerGate.go
```diff
@@ -88,7 +88,7 @@ func NewLocalTimerGate() LocalTimerGate {
 	timer := &LocalTimerGateImpl{
 		timer:          time.NewTimer(0),
 		nextWakeupTime: time.Time{},
-		fireChan:       make(chan struct{}),
+		fireChan:       make(chan struct{}, 1),
```
vancexu: Will this cause skipping timer in fireChan?

wxing1292: nop, the code, after the change looks like this
```
				select {
				// re-transmit on gateC				
				case timer.fireChan <- struct{}{}:
				default:
				}
```

wxing1292: so if there is a signal in the channel already, just skip

service/history/timerGate.go
dmetzgar: Looks like you already answered this.

wxing1292: yes, if the channel is full, then it means there is already a signal inside.

this class aims to send out signal after given time, (like a timer). if the receiving side of the channel is not polling a message in the channel, adding more does not make anything better.

PR #1035, opened 2018-08-09
Return fmt.Errorf for internal error instead of thrift error

longquanzheng:
As we decided for T1878655
Note that changing here should change any retry logic behavior in cadence-client as we discussed.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -2238,7 +2238,9 @@ func (wh *WorkflowHandler) error(err error, scope int) error {
 	case *gen.InternalServiceError:
 		logging.LogInternalServiceError(wh.Service.GetLogger(), err)
 		wh.metricsClient.IncCounter(scope, metrics.CadenceFailures)
-		return err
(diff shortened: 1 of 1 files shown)


service/frontend/workflowHandler.go
```diff
@@ -2238,7 +2238,9 @@ func (wh *WorkflowHandler) error(err error, scope int) error {
 	case *gen.InternalServiceError:
 		logging.LogInternalServiceError(wh.Service.GetLogger(), err)
 		wh.metricsClient.IncCounter(scope, metrics.CadenceFailures)
-		return err
+		// NOTE: For internal error, we won't return thrift error from cadence-frontend.
```
yiminc: what user see on client side for this error?

longquanzheng: as discussed, client is not using it for determining retry: https://github.com/uber-go/cadence-client/blob/master/internal/internal_task_pollers.go#L150

wxing1292: so we just let the client deal with unknown error?

PR #1042, opened 2018-08-13
Skip staled batch of events

yiminc:
When server complete transient decision, it write transient decision in its own batch of events and then followed by a new batch of events containing DecisionTaskCompleted and decision events. The problem comes if the 2 batches (let's say batch A and B) were successfully append but the update mutable state failed. In that case, the NextEventID won't move forward. GetHistory will not look beyond NextEventID so batch A and B just sitting in events table doing no harm. Later, when new events comes in (let's say batch C and D). We know batch C will overwrite batch A (because they have same first_event_id) but D may not overwrite B. Because batch C and batch A may have different number of events so the first_event_id for batch B and D could be different. This would result in staled batch of events to be left in events table. When NextEventID go beyond first_event_id of batch B, then that batch B become problematic. We need to skip that batch.

### DIFF
--- a/common/persistence/cassandraHistoryPersistence.go
+++ b/common/persistence/cassandraHistoryPersistence.go
@@ -156,12 +156,9 @@ func (h *cassandraHistoryPersistence) AppendHistoryEvents(request *AppendHistory
 func (h *cassandraHistoryPersistence) GetWorkflowExecutionHistory(request *GetWorkflowExecutionHistoryRequest) (
 	*GetWorkflowExecutionHistoryResponse, error) {
 	execution := request.Execution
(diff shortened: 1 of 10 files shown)


common/persistence/cassandraHistoryPersistence.go
```diff
@@ -157,6 +160,9 @@ func (h *cassandraHistoryPersistence) GetWorkflowExecutionHistory(request *GetWo
 	if err != nil {
 		return nil, err
 	}
+	if len(request.NextPageToken) == 0 {
```
wxing1292: can you put this 3 lines in the deserializeToken function?

Issue #49, opened 2017-03-06
Range Delete for Transfer Tasks/Timer Tasks/Task Lists


samarabbas:

wxing1292:
close in favor of #1272

PR #1043, opened 2018-08-14
Enable range delete on timer / transfer tasks

wxing1292:
solve #49


common/metrics/defs.go
```diff
@@ -136,12 +136,16 @@ const (
 	PersistenceGetReplicationTasksScope
 	// PersistenceCompleteTransferTaskScope tracks CompleteTransferTasks calls made by service to persistence layer
 	PersistenceCompleteTransferTaskScope
```
vancexu: We may removed these related metrics and methods if not used anymore.

common/persistence/cassandraPersistence.go
```diff
@@ -719,6 +719,16 @@ const (
...
 		`and task_id = ?`
 
+	templateRangeCompleteTransferTaskQuery = `DELETE FROM executions ` +
+		`WHERE shard_id = ? ` +
+		`and type = ? ` +
+		`and domain_id = ? ` +
+		`and workflow_id = ? ` +
+		`and run_id = ? ` +
+		`and visibility_ts = ? ` +
+		`and task_id > ? ` +
```
vancexu: Is that possible to make it [inclusive, exclusive) same as timer processor?

wxing1292:
let us keep this open, and verify by load tests

Issue #1022, opened 2018-07-31
Remove shard creation logic from history role startup


samarabbas:
History roles run the logic to create all shards in the DB when the role starts.  This works nicely with dev setup and running cadence locally.  But it could cause issues in production as few cassandra nodes being down can prevent the node to startup.
We need to remove that logic on role start up to setup of keyspace.

PR #1047, opened 2018-08-15
Do not acquire all shard when doing start up.

wxing1292:
solve #1022


service/history/shardContext.go
```diff
@@ -785,12 +786,52 @@ func acquireShard(shardID int, svc service.Service, shardManager persistence.Sha
...
+
+	retryPolicy := backoff.NewExponentialRetryPolicy(50 * time.Millisecond)
+	retryPolicy.SetMaximumInterval(time.Second)
+	retryPolicy.SetExpirationInterval(5 * time.Second)
+
+	retryPredicate := func(err error) bool {
+		if common.IsPersistenceTransientError(err) {
+			return true
+		}
+		_, ok := err.(*persistence.ShardAlreadyExistError)
```
yiminc: what is this? why is it not handled previously?

wxing1292: previously the shard creation / initialization is handled at the very beginning of the history service start, and if there is any error encounter, the start will be aborted.

this PR aim to increase the resilience when one shard is not accessible

PR #1048, opened 2018-08-15
Add PriorityTokenBucket

vancexu:


common/tb.go
```diff
@@ -185,3 +211,79 @@ func (tb *tokenBucketImpl) isRefillDue(now int64) bool {
...
+	tb.refill(time.Now().UnixNano())
+	return tb
+}
+
+func (tb *priorityTokenBucketImpl) GetToken(priority, count int) (bool, time.Duration) {
+	now := tb.timeSource.Now().UnixNano()
+	tb.Lock()
+	defer tb.Unlock()
+	tb.refill(now)
+	nextRefillTime := time.Duration(tb.nextRefillTime - now)
```
wxing1292: this line should be inside the refill?

common/tb.go
```diff
@@ -185,3 +211,79 @@ func (tb *tokenBucketImpl) isRefillDue(now int64) bool {
...
+		return false, nextRefillTime
+	}
+	tb.tokens[priority] -= count
+	return true, nextRefillTime
+}
+
+func (tb *priorityTokenBucketImpl) refill(now int64) {
+	tb.refillOverFlow(now)
+	if tb.isRefillDue(now) {
+		more := tb.fillRate
```
wxing1292: this `more` is the total tokens to be refilled for all level of priority.
so it would be good to rename this to remainingTokens or totalTokens
and add a comment here explaining the behavior

vancexu: It's the additional tokens for each buckets, none of these names looks perfect so far, so will keep it simple as more for now.

common/tb.go
```diff
@@ -44,6 +44,15 @@ type (
 		Consume(count int, timeout time.Duration) bool
 	}
 
+	// PriorityTokenBucket is the interface for rate limiter with priority
+	PriorityTokenBucket interface {
+		// GetToken attempts to take count tokens from the
+		// bucket with that priority. Priority 0 is highest.
+		// Returns true on success, false
+		// otherwise along with the duration for the next refill
+		GetToken(priority, count int) (bool, time.Duration)
```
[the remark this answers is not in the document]
vancexu: I don't think TryConsume and Consume are good names, IMO they should be GetToken and GetTokenWithTimeout like other quota services. Also we don't need Consume.

common/tb.go
```diff
@@ -62,6 +71,23 @@ type (
 		nextOverflowRefillTime int64
 		timeSource             TimeSource
 	}
+
+	priorityTokenBucketImpl struct {
+		sync.Mutex
+		tokens         []int
+		fillRate       int
```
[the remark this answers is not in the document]
vancexu: I know in practice fillRate could be float, even tokens could be float, but our tokenBucketImpl choose a simple way to allow only integer rate, so does priorityTokenBucketImpl.

PR #1049, opened 2018-08-15
Add metrics for SyncMatch

longquanzheng:


common/metrics/defs.go
```diff
@@ -452,6 +452,10 @@ const (
 	MatchingAddActivityTaskScope
 	// MatchingAddDecisionTaskScope tracks AddDecisionTask API calls received by service
 	MatchingAddDecisionTaskScope
+	// MatchingAddActivityTaskScope tracks AddActivityTask API calls received by service
```
vancexu: It doesn't look right to use scope in this case, we may need new metrics not scopes

service/matching/handler.go
```diff
@@ -115,28 +117,38 @@ func (h *Handler) startRequestProfile(api string, scope int) tally.Stopwatch {
...
 
 // AddDecisionTask - adds a decision task.
 func (h *Handler) AddDecisionTask(ctx context.Context, addRequest *m.AddDecisionTaskRequest) error {
+	startT := time.Now()
 	scope := metrics.MatchingAddDecisionTaskScope
-	sw := h.startRequestProfile("AddDecisionTask", scope)
-	defer sw.Stop()
 
 	if ok, _ := h.rateLimiter.TryConsume(1); !ok {
 		return h.handleErr(errMatchingHostThrottle, scope)
```
vancexu: missing metric recored here with existing implementation

service/matching/handler.go
```diff
@@ -115,28 +117,38 @@ func (h *Handler) startRequestProfile(api string, scope int) tally.Stopwatch {
...
+	h.metricsClient.RecordTimer(scope, metrics.CadenceLatency, time.Since(startT))
+	h.metricsClient.IncCounter(scope, metrics.CadenceRequests)
+	return h.handleErr(err, scope)
 }
 
 // AddDecisionTask - adds a decision task.
 func (h *Handler) AddDecisionTask(ctx context.Context, addRequest *m.AddDecisionTaskRequest) error {
+	startT := time.Now()
 	scope := metrics.MatchingAddDecisionTaskScope
-	sw := h.startRequestProfile("AddDecisionTask", scope)
```
vancexu: I would keep this, but add a new metric for sync match

service/matching/handler.go
```diff
@@ -115,28 +117,38 @@ func (h *Handler) startRequestProfile(api string, scope int) tally.Stopwatch {
...
 	scope := metrics.MatchingAddActivityTaskScope
-	sw := h.startRequestProfile("AddActivityTask", scope)
-	defer sw.Stop()
 
 	if ok, _ := h.rateLimiter.TryConsume(1); !ok {
 		return h.handleErr(errMatchingHostThrottle, scope)
 	}
 
-	return h.handleErr(h.engine.AddActivityTask(addRequest), scope)
+	err, syncMatch := h.engine.AddActivityTask(addRequest)
```
yiminc: we can use the same scope, but create new metric name for this dispatch latency.

service/matching/matchingEngineInterfaces.go
```diff
@@ -31,8 +31,8 @@ type (
 	// Engine exposes interfaces for clients to poll for activity and decision tasks.
 	Engine interface {
 		Stop()
-		AddDecisionTask(addRequest *m.AddDecisionTaskRequest) error
-		AddActivityTask(addRequest *m.AddActivityTaskRequest) error
+		AddDecisionTask(addRequest *m.AddDecisionTaskRequest) (err error, syncMatch bool)
+		AddActivityTask(addRequest *m.AddActivityTaskRequest) (err error, syncMatch bool)
```
yiminc: put error as last return value.

service/matching/taskListManager.go
```diff
@@ -58,7 +58,7 @@ var errAddTasklistThrottled = errors.New("cannot add to tasklist, limit exceeded
 type taskListManager interface {
 	Start() error
 	Stop()
-	AddTask(execution *s.WorkflowExecution, taskInfo *persistence.TaskInfo) error
+	AddTask(execution *s.WorkflowExecution, taskInfo *persistence.TaskInfo) (err error, syncMatch bool)
 	GetTaskContext(ctx context.Context, maxDispatchPerSecond *float64) (*taskContext, error)
```
yiminc: put error as last return value

service/matching/handler.go
```diff
@@ -24,6 +24,8 @@ import (
 	"context"
 	"sync"
```
vancexu: may remove additional line

service/matching/taskListManager.go
```diff
@@ -385,7 +385,10 @@ func (c *taskListManagerImpl) AddTask(execution *s.WorkflowExecution, taskInfo *
 	if err == nil {
 		c.signalNewTask()
 	}
-	return err
+	if r != nil && err == nil {
```
vancexu: This is incorrect because non-sync match can also go into this. We may need a closure instead of using `r`

vancexu: Can we also add a test to make sure it return Ture when expected?

yiminc: I think this won't happen for non-syncmatch. But if you want to be sure, we can add a field to the response struct and check that field here.

vancexu: Correct me if I am wrong: 
executeWithRetry return its operation's result and err, and it's operation could be either `c.trySyncMatch(taskInfo)` or `c.taskWriter.appendTask(execution, taskInfo, rangeID)`. 
The latter `appendTask` can return `r != nil && err == nil`, which will lead to unwanted `syncMatch` be true. 
Not sure why you think this won't happen.

vancexu: Correct me if I am wrong:
executeWithRetry return its operation's result and err, and it's operation could be either c.trySyncMatch(taskInfo) or c.taskWriter.appendTask(execution, taskInfo, rangeID).
The latter appendTask can return r != nil && err == nil, which will lead to unwanted syncMatch be true.
Not sure why you think this won't happen.

yiminc: Actually Bowei is correct. @longquanzheng , could you add a field to the CreateTasksResponse to indicate if it is sync match or not.

vancexu: Maybe not need a new field in response, just use closure. Inside trySyncMatch if `if (err != nil && err != errAddTasklistThrottled) || r != nil {` add `syncMatch = true`
after `r, err = c.taskWriter.appendTask(execution, taskInfo, rangeID)` add `syncMatch = false`. This may work

PR #1050, opened 2018-08-16
Bugfix: should notify new timer if current timer scan is a failure

wxing1292:


service/history/timerQueueProcessorBase.go
```diff
@@ -352,6 +352,7 @@ func (t *timerQueueProcessorBase) readAndFanoutTimerTasks() (*persistence.TimerT
 	t.lastPollTime = time.Now()
 	timerTasks, lookAheadTask, moreTasks, err := t.timerQueueAckMgr.readTimerTasks()
 	if err != nil {
+		t.notifyNewTimer(time.Time{}) // re-enqueue the event
```
samarabbas: Can we end up in a tight loop here?  
Probably it is safer to do a backoff here.  Or do we have backoff at the persistence level?

wxing1292: just few line above, there is a rate limiter, which will protect the DB

PR #1052, opened 2018-08-16
Change behavior from panic to log error when task cannot be processed.

wxing1292:


service/history/queueProcessor.go
```diff
@@ -330,7 +330,8 @@ ProcessRetryLoop:
 			return
 		}
 
-		logging.LogOperationPanicEvent(logger, "Retry count exceeded for transfer task", err)
+		p.metricsClient.IncCounter(p.options.MetricScope, metrics.TaskStuck)
+		logging.LogOperationStuckEvent(logger, "Retry count exceeded for transfer task", err)
```
samarabbas: Can we loose task by converting fatal into error?

wxing1292: what do you mean by lose?

PR #1057, opened 2018-08-18
Bugfix: reset workflow should also reset the current execution record

wxing1292:


common/persistence/cassandraPersistence_test.go
```diff
@@ -2468,6 +2468,78 @@ func (s *cassandraPersistenceSuite) TestResetMutableState() {
...
+		StartVersion:   int64(8780),
+	}
+
+	err = s.ResetMutableState(workflowExecutionCurrent.GetRunId(), resetExecutionInfo, rState, int64(5), resetActivityInfos, resetTimerInfos,
+		resetChildExecutionInfos, resetRequestCancelInfos, resetSignalInfos, nil)
+	s.Nil(err, "No error expected.")
+
+	// this test only assert whether the current workflow execution record is reseted
+	runID, err = s.GetCurrentWorkflowRunID(domainID, workflowID)
+	s.Equal(workflowExecutionReset.GetRunId(), runID)
```
samarabbas: Can you modify this test to do a reset after terminating the continuedAsNew execution.

samarabbas:
Looks good after the update.  Let's modify the unit test as I suggested in the comment and land this PR.

common/persistence/cassandraPersistence_test.go
wxing1292: done

PR #1059, opened 2018-08-20
Add visibility sampling

vancexu:
Visibility table uses domain_id + domain_partition as partition key. Currently, we are hardcode 0 as domain_partition for all domains. This will not scale in performance, and also may hit  cassandra limit of 2 billion columns per partition, if we have large number of workflows in a domain.

But instead of working on tuning partition (which will introduce some limitation when read), we want to avoid having large volume of visibility at the beginning.


common/metrics/defs.go
```diff
@@ -824,6 +825,7 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		PersistenceErrConditionFailedCounter:          {metricName: "persistence.errors.condition-failed", metricType: Counter},
 		PersistenceErrTimeoutCounter:                  {metricName: "persistence.errors.timeout", metricType: Counter},
 		PersistenceErrBusyCounter:                     {metricName: "persistence.errors.busy", metricType: Counter},
+		PersistenceErrSampledCounter:                  {metricName: "persistence.errors.sampled", metricType: Counter},
```
samarabbas: I don't think we should use "errors" in the name of the counter.  Let's just call the counter persistence.sampled.

service/history/service.go
```diff
@@ -259,10 +265,13 @@ func (s *Service) Start() {
 		p.Logger)
 
 	if err != nil {
-		log.Fatalf("failed to create visiblity manager: %v", err)
+		log.Fatalf("failed to create visibility manager: %v", err)
 	}
 	visibility = persistence.NewVisibilityPersistenceRateLimitedClient(visibility, persistenceRateLimiter, log)
 	visibility = persistence.NewVisibilityPersistenceMetricsClient(visibility, base.GetMetricsClient(), log)
+	if s.config.EnableVisibilitySampling() {
```
samarabbas: probably we should instantiate the metric client at the very end.

service/history/visibilitySamplingClient.go
```diff
@@ -0,0 +1,167 @@
...
+
+	m.Lock()
+	if rateLimiter, ok := m.mappings[domain]; ok { // read again to ensure no duplicate create
+		m.Unlock()
+		return rateLimiter
+	}
+	if numOfPriority == 1 {
+		rateLimiter = common.NewPriorityTokenBucket(numOfPriority, qps, common.NewRealTimeSource())
+	} else {
+		rateLimiter = common.NewFullPriorityTokenBucket(numOfPriority, qps, common.NewRealTimeSource())
```
samarabbas: I think we can just simplify this code to not have if-else branch and always use FullPriorityTokenBucket

PR #1088, opened 2018-09-05
Seperate metrics of client API from service API

vancexu:


common/metrics/defs.go
```diff
@@ -709,8 +709,9 @@ const (
 	PersistenceErrBusyCounter
 	PersistenceSampledCounter
 
-	HistoryClientFailures
-	MatchingClientFailures
+	CadenceClientRequests
```
samarabbas: Can you also add another tag for client type.  Like 'type=history' for all history client operations and 'type=matching' for all matching client operations?

common/metrics/defs.go
```diff
@@ -66,14 +66,18 @@ const (
...
+	CadenceRoleTagName = "cadence-role"
 )
 
 // This package should hold all the metrics and tags for cadence
 const (
 	UnknownDirectoryTagValue = "Unknown"
 	AllShardsTagValue        = "ALL"
 	NoneShardsTagValue       = "NONE"
+
+	HistoryRoleTagName  = "history"
```
samarabbas: Should be value suffix

PR #1097, opened 2018-09-10
Fix execution table update condition for current_run_id checking and fix error msg

longquanzheng:
Still working on unit tests

### DIFF
--- a/common/persistence/cassandraPersistence.go
+++ b/common/persistence/cassandraPersistence.go
@@ -1916,7 +1916,14 @@ GetFailureReasonLoop:
 		}
 	}
 
-	if nextEventIDUnmatch || runIDUnmatch {
+	if runIDUnmatch {
+		return &WorkflowConditionFailedError{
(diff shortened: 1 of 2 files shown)


common/persistence/cassandraPersistence.go
```diff
@@ -1902,15 +1911,15 @@ GetFailureReasonLoop:
 	if rangeIDUnmatch {
 		return &ShardOwnershipLostError{
 			ShardID: d.shardID,
-			Msg: fmt.Sprintf("Failed to reset mutable state.  Request RangeID: %v, Actual RangeID: %v",
+			Msg: fmt.Sprintf("Failed to update mutable state.  Request RangeID: %v, Actual RangeID: %v",
 				requestRangeID, actualRangeID),
 		}
 	}
```
wxing1292: plz separate the runIDUnmatch to here, and if not match, throw a new error such as currentRunIDMismatchError

wxing1292:
close in favor of #1098

PR #1101, opened 2018-09-10
Add handling of CurrentWorkflowConditionFailedError

wxing1292:


common/metrics/defs.go
```diff
@@ -789,6 +789,7 @@ const (
 	StandbyTransferTaskQueueLatency
 	StandbyTimerTaskQueueLatency
 	CompleteTaskFailedCounter
+	WorkflowConcurrentExecutionCounter
```
samarabbas: Can we define this as a PersistenceErr Counter similar to 'PersistenceErrConditionFailedCounter'

service/history/handler.go
```diff
@@ -1140,6 +1140,8 @@ func (h *Handler) updateErrorMetric(scope int, err error) {
 		h.metricsClient.IncCounter(scope, metrics.CadenceErrRetryTaskCounter)
 	case *gen.ServiceBusyError:
 		h.metricsClient.IncCounter(scope, metrics.CadenceErrServiceBusyCounter)
+	case *persistence.CurrentWorkflowConditionFailedError:
```
samarabbas: Persistence errors are not defined on the service API.  These needs to be wrapped with an InternalServiceError or something.

PR #1107, opened 2018-09-12
Refactor task processing logic

wxing1292:


service/history/queueProcessor.go
```diff
@@ -267,93 +267,116 @@ func (p *queueProcessorBase) retryTasks() {
...
-	logger = p.logger.WithFields(bark.Fields{
-		logging.TagTaskID:   task.GetTaskID(),
-		logging.TagTaskType: task.GetTaskType(),
-		logging.TagVersion:  task.GetVersion(),
+	logging.LogTaskProcessingFailedEvent(logger, "Fail to process task", err)
+	return err
+}
+
+func (p *queueProcessorBase) ackTaskOnce(task queueTaskInfo, scope int) {
+	p.ackMgr.completeQueueTask(task.GetTaskID())
```
vancexu: Looks like missing err handling for completeQueueTask

service/history/queueProcessor.go
```diff
@@ -267,93 +267,116 @@ func (p *queueProcessorBase) retryTasks() {
...
+		time.Since(task.GetVisibilityTimestamp()),
+	)
+}
+
+func (p *queueProcessorBase) initializeLoggerForTask(task queueTaskInfo) bark.Logger {
+	logger := p.logger.WithFields(bark.Fields{
+		logging.TagHistoryShardID: p.shard.GetShardID(),
+		logging.TagTaskID:         task.GetTaskID(),
+		logging.TagTaskType:       task.GetTaskType(),
+		logging.TagVersion:        task.GetVersion(),
```
vancexu: It will be better to add TagDomainID, TagWorkflowExecutionID and TagWorkflowRunID here also to save one WithFields call

service/history/replicatorQueueProcessor.go
```diff
@@ -104,27 +104,18 @@ func newReplicatorQueueProcessor(shard ShardContext, replicator messaging.Produc
...
-	var err error
+
 	switch task.TaskType {
 	case persistence.ReplicationTaskTypeHistory:
-		scope = metrics.ReplicatorTaskHistoryScope
-		err = p.processHistoryReplicationTask(task)
+		return metrics.ReplicatorTaskHistoryScope, p.processHistoryReplicationTask(task)
 	default:
-		err = errUnknownReplicationTask
+		return 0, errUnknownReplicationTask
```
vancexu: maybe a dummy scope instead of 0?

service/history/timerQueueActiveProcessor.go
```diff
@@ -194,80 +194,45 @@ func (t *timerQueueActiveProcessorImpl) notifyNewTimers(timerTasks []persistence
...
+		return metrics.TimerActiveTaskWorkflowRetryTimerScope, t.processWorkflowRetryTimer(timerTask)
 
 	case persistence.TaskTypeDeleteHistoryEvent:
-		scope = metrics.TimerActiveTaskDeleteHistoryEvent
-		err = t.timerQueueProcessorBase.processDeleteHistoryEvent(timerTask)
-	}
-
-	t.logger.Debugf("Processing timer: (%v, %v), for WorkflowID: %v, RunID: %v, Type: %v, EventID: %v, Error: %v",
-		timerTask.TaskID, timerTask.VisibilityTimestamp, timerTask.WorkflowID, timerTask.RunID, timerTask.TaskType, timerTask.EventID, err)
+		return metrics.TimerActiveTaskDeleteHistoryEvent, t.timerQueueProcessorBase.processDeleteHistoryEvent(timerTask)
```
vancexu: can we rename TimerActiveTaskDeleteHistoryEvent to TimerActiveTaskDeleteHistoryEventScope

service/history/queueProcessor.go
```diff
@@ -267,93 +267,116 @@ func (p *queueProcessorBase) retryTasks() {
...
+	p.metricsClient.RecordTimer(scope, metrics.TaskProcessingLatency, time.Since(startTime))
+
+	return scope, err
+}
+
+func (p *queueProcessorBase) handleTaskError(scope int, startTime time.Time,
+	notificationChan <-chan struct{}, err error, logger bark.Logger) error {
+
+	if _, ok := err.(*workflow.EntityNotExistsError); ok {
+		return nil
```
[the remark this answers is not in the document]
wxing1292: why do you need this metrics? this error can be returned if task is a dup task, or varies of reasons, like workflow has already completed

service/history/queueProcessor.go
```diff
@@ -267,93 +267,116 @@ func (p *queueProcessorBase) retryTasks() {
...
 		}
 
-		logging.LogOperationPanicEvent(logger, "Retry count exceeded for replication task", err)
+		return err
+	}
+
+	p.metricsClient.IncCounter(scope, metrics.TaskFailures)
+
+	if _, ok := err.(*persistence.CurrentWorkflowConditionFailedError); ok {
+		logging.LogTaskProcessingFailedEvent(logger, "More than 2 workflow is running.", err)
```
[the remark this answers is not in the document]
wxing1292: ok

service/history/queueProcessor.go
wxing1292: this is intentional, will apply range delete replication task. after that, this issue is gone

service/history/queueProcessor.go
wxing1292: just look down a few lines and you will find 
```
 	switch task := task.(type) {
 	case *persistence.TransferTaskInfo:
 		logger = logger.WithFields(bark.Fields{
 			logging.TagDomainID:            task.DomainID,
 			logging.TagWorkflowExecutionID: task.WorkflowID,
 			logging.TagWorkflowRunID:       task.RunID,
 		})
 
 		logger.Debug("Processing transfer task")
 	case *persistence.ReplicationTaskInfo:
 		logger = logger.WithFields(bark.Fields{
 			logging.TagDomainID:            task.DomainID,
 			logging.TagWorkflowExecutionID: task.WorkflowID,
 			logging.TagWorkflowRunID:       task.RunID,
 		})
 
 		logger.Debug("Processing replication task")
 	}
```

service/history/replicatorQueueProcessor.go
wxing1292: sure

service/history/timerQueueActiveProcessor.go
wxing1292: ok

service/history/queueProcessor.go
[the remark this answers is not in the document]
vancexu: ok then make sense not emit here.

service/history/queueProcessor.go
vancexu: I saw that, that's why I think we could add them to one place instead of two. code below just looks like unnecessary duplicates.

service/history/MockProcessor.go
```diff
@@ -30,16 +30,24 @@ type MockProcessor struct {
 }
 
 // process is mock implementation for process of Processor
-func (_m *MockProcessor) process(task queueTaskInfo) error {
+func (_m *MockProcessor) process(task queueTaskInfo) (int, error) {
```
halakaraki: What does the int return value signify? What outcome does process other than it worked or error?

service/history/queueProcessor.go
```diff
@@ -267,93 +267,116 @@ func (p *queueProcessorBase) retryTasks() {
...
-				retryCount++
-				logger = p.initializeLoggerForTask(task, logger)
-				logging.LogTaskProcessingFailedEvent(logger, err)
-			}
-		}
-		return err
+		scope, err = p.processTaskOnce(notificationChan, task, logger)
+		return p.handleTaskError(scope, startTime, notificationChan, err, logger)
 	}
+	retryCondition := func(err error) bool { return true }
```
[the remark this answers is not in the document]
halakaraki: why not drop the attempt and just log after retrying a certain number of times?

common/logging/helpers.go
```diff
@@ -56,12 +56,12 @@ func LogOperationFailedEvent(logger bark.Logger, msg string, err error) {
 	}).Warnf("%v.  Error: %v", msg, err)
 }
 
-// LogOperationPanicEvent is used to log fatal errors by application to cause panic
-func LogOperationPanicEvent(logger bark.Logger, msg string, err error) {
+// LogOperationStuckEvent is used to log fatal errors by application to cause panic
```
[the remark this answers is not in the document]
wxing1292: ok

service/history/queueProcessor.go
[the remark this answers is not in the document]
wxing1292: > why not drop the attempt and just log after retrying a certain number of times?

i need to have a counter for "certain number of times"

service/history/MockProcessor.go
wxing1292: this is to some degree an anti pattern, but can be useful for this use case.

basically the scope is unknown, before processing the task. we can either do a switch case at the top layer to get the scope then internal implementation do an switch case again to redirect the task to target processing logic, or just let the target processing logic return the scope, which eliminate unnecessary logic.

PR #1109, opened 2018-09-13
Should check whether workflow has closed before processing decision

wxing1292:


service/history/historyEngine.go
```diff
@@ -781,7 +785,7 @@ Update_History_Loop:
 
 		// Check execution state to make sure task is in the list of outstanding tasks and it is not yet started.  If
 		// task is not outstanding than it is most probably a duplicate and complete the task.
-		if !msBuilder.IsWorkflowExecutionRunning() || !isRunning {
+		if !isRunning {
```
halakaraki: What's the difference between isRunning and IsWorkflowExecutionRunning()?

wxing1292: "is running" is whether the decision is running

PR #1112, opened 2018-09-13
History Events V2: Schema and Interface

longquanzheng:


longquanzheng:
Too many conflicts, use this PR instead: https://github.com/uber/cadence/pull/1156

PR #1114, opened 2018-09-14
Added metric for history cache component

samarabbas:
Added the following metric for history cache:
HistoryCacheRequests
HistoryCacheFailures
HistoryCacheLatency
CacheMissCounter
AcquireLockFailedCounter
WorkflowContextCleared


common/metrics/defs.go
```diff
@@ -450,6 +450,14 @@ const (
 	ReplicateHistoryEventsScope
 	// ShardInfoScope is the scope used when updating shard info
 	ShardInfoScope
+	// WorkflowContextScope is the scope used when by WorkflowContext component
```
vancexu: typo "when by"

common/metrics/defs.go
```diff
@@ -794,6 +806,12 @@ const (
 	StandbyTransferTaskQueueLatency
 	StandbyTimerTaskQueueLatency
 	CompleteTaskFailedCounter
+	HistoryCacheRequests
+	HistoryCacheFailures
+	HistoryCacheLatency
+	CacheMissCounter
```
vancexu: Will it be better named as HistoryCacheMissCounter and "history-cache.miss"?

samarabbas: Leaving it as-is.  Will address this once we have use case for cache-miss counter which needs to be used by other components.

PR #1122, opened 2018-09-19
Bugfix error is not handled for replication task ack

wxing1292:
Adding more metrics tracking replication / trensfer / timer level lag


service/history/shardContext.go
```diff
@@ -722,23 +722,32 @@ func (s *shardContextImpl) emitShardInfoMetricsLogsLocked() {
...
 	diffTimerLevel := maxTimerLevel.Sub(minTimerLevel)
 
-	if logWarnTransferLevelDiff < diffTransferLevel || logWarnTimerLevelDiff < diffTimerLevel {
+	replicationLag := s.transferMaxReadLevel - s.shardInfo.ReplicationAckLevel
+	transferLag := s.transferMaxReadLevel - s.shardInfo.TransferAckLevel
+	timerLag := time.Since(s.shardInfo.TimerAckLevel)
+
+	if logWarnTransferLevelDiff < diffTransferLevel ||
+		logWarnTimerLevelDiff < diffTimerLevel ||
+		logWarnTransferLevelDiff < replicationLag ||
```
vancexu: Should we have a separate threshold for replicationLag?

wxing1292: if there is no replication generated on this shard, then the replicationLag will increase since the replication ack level is unchanged.

emitting metrics here helps us understand the shard status, but we should not use this as an indicator whether to alarm.

i am dropping emit warn log on replicationLag to reduce unnecessary warn log

PR #1125, opened 2018-09-19
Thrift encoding for history/mutableState events 

longquanzheng:
I want to get feedbacks before I proceed. 

Still need to do:
1. Replace SerializedHistoryEventBatch with datablob in other components like historyReplicator/ShardContext/...
2. Decide which encoding based on domain dynamic config
3. Make the same change for MySQL persistence
4. Fix unit test


schema/cassandra/cadence/versioned/v0.12/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
```
samarabbas: Can we do this in v0.11?  So we can roll this out in next release.

common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -34,15 +34,15 @@ import (
...
+		`VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?) IF NOT EXISTS`
 
 	templateOverwriteHistoryEvents = `UPDATE events ` +
-		`SET event_batch_version = ?, range_id = ?, tx_id = ?, data = ?, data_encoding = ?, data_version = ? ` +
+		`SET event_batch_version = ?, range_id = ?, tx_id = ?, data = ?, data_headers = ? ` +
 		`WHERE domain_id = ? AND workflow_id = ? AND run_id = ? AND first_event_id = ? ` +
 		`IF range_id <= ? AND tx_id < ?`
 
-	templateGetWorkflowExecutionHistory = `SELECT first_event_id, event_batch_version, data, data_encoding, data_version FROM events ` +
+	templateGetWorkflowExecutionHistory = `SELECT first_event_id, event_batch_version, data, data_headers FROM events ` +
```
samarabbas: This could might break existing events written to database as the encoding and version is stored as separate columns rather than this new map.

common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -247,10 +244,9 @@ func (h *cassandraHistoryPersistence) GetWorkflowExecutionHistory(request *p.Get
 	return response, nil
 }
 
-func (h *cassandraHistoryPersistence) deserializeEvents(e *p.SerializedHistoryEventBatch) (*p.HistoryEventBatch, error) {
-	p.SetSerializedHistoryDefaults(e)
-	s, _ := h.serializerFactory.Get(e.EncodingType)
-	return s.Deserialize(e)
+func (h *cassandraHistoryPersistence) deserializeEvents(blob *p.DataBlob) (*p.HistoryEventBatch, error) {
+	s, _ := h.serializerFactory.Get(blob.GetEncoding())
```
samarabbas: Why did you remove setting of serialized history defaults?  Can this cause problem with existing data?

schema/cassandra/cadence/versioned/v0.12/manifest.json
longquanzheng: OK will do that.

common/persistence/cassandra/cassandraHistoryPersistence.go
longquanzheng: Replied offline. GetEncoding already return Json by default when headers is empty for backward compatibility.

common/persistence/cassandra/cassandraHistoryPersistence.go
longquanzheng: Replied offline.

service/history/transferQueueActiveProcessor.go
```diff
@@ -649,10 +649,10 @@ func (t *transferQueueActiveProcessorImpl) processSignalExecution(task *persiste
 			},
 			Identity:   common.StringPtr(identityHistoryService),
 			SignalName: common.StringPtr(si.SignalName),
-			Input:      si.Input,
+			Input:      si.Input.Data,
```
longquanzheng: NOTE: we will pass both Data and Headers for compression. That will be along with step 4 changes(API changes for compression)

common/persistence/dataInterfaces.go
```diff
@@ -112,6 +114,13 @@ const (
 	TransferTaskTransferTargetRunID = "30000000-0000-f000-f000-000000000002"
 )
 
+// Keys of DataBlob headers
+const (
+	DataBlobHeaderKeyEncoding            = "$Cadence.Blob.Encoding"
+	DataBlobHeaderKeyVersion             = "$Cadence.Blob.Version"
+	DataBlobHeaderKeyFrontendCompression = "$Cadence.Frontend.Compression"
```
samarabbas: Let's not create any compression related changes in the PR.

common/persistence/dataInterfaces.go
```diff
@@ -461,18 +470,18 @@ type (
 		SignalRequestedIDs       map[string]struct{}
 		ExecutionInfo            *WorkflowExecutionInfo
 		ReplicationState         *ReplicationState
-		BufferedEvents           []*SerializedHistoryEventBatch
+		BufferedEvents           []*workflow.History
```
samarabbas: Can we just represent this as a flat array of workflow.HistoryEvents?

common/persistence/dataInterfaces.go
```diff
@@ -659,10 +668,11 @@ type (
 		DeleteSignalInfo              *int64
 		UpsertSignalRequestedIDs      []string
 		DeleteSignalRequestedID       string
-		NewBufferedEvents             *SerializedHistoryEventBatch
+		NewBufferedEvents             *workflow.History
```
samarabbas: I think we should not use workflow.History at this layer and directory expose collection of events as []*workflow.HistoryEvents

common/persistence/dataInterfaces.go
```diff
@@ -821,11 +832,11 @@ type (
...
 
-	// SerializedHistoryEventBatch represents a serialized batch of history events
-	SerializedHistoryEventBatch struct {
-		EncodingType common.EncodingType
-		Version      int
-		Data         []byte
+	// DataBlob represents a blob for any binary data.
+	// The raw data is in Data field, and metadata(encoding/version/compression/etc) is in headers field
+	DataBlob struct {
+		Headers map[string]string
```
samarabbas: These are always going to be Cadence specific headers so I think we should just stick to strongly typed interface here.

common/persistence/dataInterfaces.go
```diff
@@ -842,8 +853,9 @@ type (
 		EventBatchVersion int64
 		RangeID           int64
 		TransactionID     int64
-		Events            *SerializedHistoryEventBatch
+		Events            *workflow.History
```
samarabbas: Let's keep it consistent and use []*workflow.HistoryEvent

longquanzheng:
Use another PR since there were too many conflicts that I have to start over from latest master

PR #1129, opened 2018-09-20
Review IDL: fix throw error list to correct SLO calculation

longquanzheng:


service/frontend/workflowHandler.go
```diff
@@ -1942,7 +1942,7 @@ func (wh *WorkflowHandler) ListClosedWorkflowExecutions(ctx context.Context,
 	} else if listRequest.StatusFilter != nil {
 		persistenceResp, err = wh.visibitiltyMgr.ListClosedWorkflowExecutionsByStatus(&persistence.ListClosedWorkflowExecutionsByStatusRequest{
 			ListWorkflowExecutionsRequest: baseReq,
-			Status:                        listRequest.GetStatusFilter(),
+			Status: listRequest.GetStatusFilter(),
```
vancexu: need `make fmt`? Or mismatch go tool version lead to inconsistent fmt issue?

longquanzheng: yea I am figuring it out

PR #1133, opened 2018-09-21
Fixed SQL metadata persistence 

mfateev:
Removed legacy SQL metadata and kept only V2 version.
Fixed SQL metadata implementation to pass all relevant unit tests.


mfateev:
https://github.com/uber/cadence/issues/854

common/persistence/sql/sqlMetadataPersistenceV2.go
```diff
@@ -201,99 +218,84 @@ func lockMetadata(tx *sqlx.Tx) error {
...
-		tx, err := m.db.Beginx()
-		if err != nil {
-			return nil, err
-		}
-		defer tx.Rollback()
+	tx, err := m.db.Beginx()
+	if err != nil {
+		return nil, err
+	}
+	defer tx.Rollback()
```
venkat1109: [optional] I believe rollback is un-necessary when err is nil i.e. we successfully committed this transaction. Don't think it affects correctness - its might result in a additional call to the dbserver that results in a noop

common/persistence/sql/sqlMetadataPersistenceV2.go
```diff
@@ -34,8 +35,10 @@ import (
 
 type (
 	// Implements MetadataManager
-	sqlMetadataManager struct {
-		db *sqlx.DB
+	sqlMetadataManagerV2 struct {
+		db                 *sqlx.DB
+		currentClusterName string
```
venkat1109: can we call this 'activeClusterName' just to be consistent with the db field ?

PR #1145, opened 2018-09-26
Event thrift encoding and refactoring dataInterfaces

longquanzheng:
We want to support binary encoding for our history events. While doing that, we decided to refactor our dataInterface so that our application layer doesn't need to know about encoding/decoding. Also our persistence layer is broken down into two, the middle layer will be responsible for decoding/encoding and calculating size information.


common/persistence/dataStoreImpl.go
```diff
@@ -0,0 +1,717 @@
...
+	"github.com/uber-common/bark"
+	workflow "github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/logging"
+)
+
+type (
+
+	// executionManagerImpl implements ExecutionManager based on PersistenceExecutionManager, statsComputer and HistorySerializer
+	executionManagerImpl struct {
```
samarabbas: Can you separate out each one of manager implementations in its own file.

common/persistence/statsComputer.go
```diff
@@ -0,0 +1,251 @@
...
+	// statsComputer is to computing struct sizes after serialization
+	statsComputer struct{}
+)
+
+func (sc *statsComputer) computeMutableStateStats(req *PersistenceUpdateWorkflowExecutionRequest) *MutableStateStats {
+	executionInfoSize := computeExecutionInfoSize(req.ExecutionInfo)
+
+	activityInfoCount := 0
+	activityInfoSize := 0
+	for _, ai := range req.UpsertActivityInfos {
```
samarabbas: This is not going to work.  Mutable state stats cannot be computed using update request.  It only has the upsert activity/timer/child/signal infos, not the entire size of mutable state.

common/persistence/dataStoreImpl.go
longquanzheng: Got it

common/persistence/historySerializer.go
```diff
@@ -23,25 +23,23 @@ package persistence
...
 type (
-	// HistorySerializer is used to serialize/deserialize history
+	// HistorySerializer is used by persistence to serialize/deserialize history event(s)
+	// It will only be used inside persistence, so that serialize/deserialize is transparent for application
 	HistorySerializer interface {
-		Serialize(batch *HistoryEventBatch) (*SerializedHistoryEventBatch, error)
-		Deserialize(batch *SerializedHistoryEventBatch) (*HistoryEventBatch, error)
-	}
+		// serialize/deserialize history events
+		SerializeBatchEvents(batch *workflow.History, encodingType common.EncodingType) (*DataBlob, error)
```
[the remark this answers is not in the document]
longquanzheng: We can do that. I thought you only want that flatening in DataInterface.

common/persistence/persistenceInterface.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// PersistenceShardManager is a lower level of ShardManager
+	PersistenceShardManager = ShardManager
+	// PersistenceTaskManager is a lower level of TaskManager
+	PersistenceTaskManager = TaskManager
+	// PersistenceMetadataManager is a lower level of MetadataManager
+	PersistenceMetadataManager = MetadataManager
+
+	// PersistenceExecutionManager is used to manage workflow executions for Persistence layer
+	PersistenceExecutionManager interface {
```
samarabbas: Let's use Store naming convention here.
ExecutionStore
EventsStore
etc

longquanzheng: OK sounds good to me.
BTW EventsStore I think you meant HistoryStore

common/persistence/historySerializer.go
```diff
@@ -63,79 +61,120 @@ type (
...
-// of HistorySerializerFactory
-func NewHistorySerializerFactory() HistorySerializerFactory {
-	return &serializerFactoryImpl{
-		jsonSerializer: NewJSONHistorySerializer(),
+func (t *serializerImpl) SerializeEvent(event *workflow.HistoryEvent, encodingType common.EncodingType) (*DataBlob, error) {
+	if event == nil {
+		event = &workflow.HistoryEvent{}
+	}
+	switch encodingType {
+	case common.EncodingTypeGob:
```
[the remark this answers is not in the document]
longquanzheng: We need to return error here to make unit test happy. The default case is Json which shouldn't return error.

PR #1153, opened 2018-09-28
Add dynamic config for encoding type

longquanzheng:
This is to allow each domain to customize event encoding type.

PS: still fixing unit tests

### DIFF
--- a/common/service/dynamicconfig/constants.go
+++ b/common/service/dynamicconfig/constants.go
@@ -129,6 +129,7 @@ var keys = map[Key]string{
 	MaximumBufferedEventsBatch:                            "history.maximumBufferedEventsBatch",
 	ShardUpdateMinInterval:                                "history.shardUpdateMinInterval",
 	ShardSyncMinInterval:                                  "history.shardSyncMinInterval",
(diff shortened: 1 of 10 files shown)


common/service/dynamicconfig/constants.go
```diff
@@ -308,7 +309,8 @@ const (
 	ShardUpdateMinInterval
 	// ShardSyncMinInterval is the minimal time interval which the shard info should be sync to remote
 	ShardSyncMinInterval
-
+	// EventEncodingPropertyFilteredByDomainKey is the encoding type for history events
+	EventEncodingPropertyFilteredByDomainKey
```
samarabbas: Let's stick to the convention used by other properties and call it DefaultEventEncoding or something.

longquanzheng: I will talk to Bowei to make sure this sticks to our convention

vancexu: We don't need surfix `PropertyFilteredByDomainKey` in config name. Agree with samar just name it `EventEncoding` or `DefaultEncoding` etc.

longquanzheng: I see. Suffix is only a convention in config function.

PR #1158, opened 2018-10-02
 Add sampling for long rentention in visibility

vancexu:


common/cache/domainCache_test.go
```diff
@@ -601,4 +601,52 @@ func Test_GetRetentionDays(t *testing.T) {
...
+			Data: make(map[string]string),
+		},
+		config: &persistence.DomainConfig{
+			Retention: 7,
+		},
+	}
+	wid := uuid.New()
+	require.False(t, d.IsSampledForLongerRetentionEnabled(wid))
+	d.info.Data[SampleRetentionKey] = "30"
+	d.info.Data[SampleRateKey] = "0"
```
yiminc: It is weird that we treat 0 as enabled.

PR #1165, opened 2018-10-08
Bugfix: domain replication should not replicate non global domain

wxing1292:


service/frontend/workflowHandler.go
```diff
@@ -257,7 +257,8 @@ func (wh *WorkflowHandler) RegisterDomain(ctx context.Context, registerRequest *
 	// TODO remove the IsGlobalDomainEnabled check once cross DC is public
 	if clusterMetadata.IsGlobalDomainEnabled() {
```
samarabbas: Should we also check IsGlobalDomain here before creating the replication task?

PR #1167, opened 2018-10-09
Fix race issue in signalwithstart

vancexu:
Related issue https://github.com/uber/cadence/issues/1162

In this fix:
concurrent SignalWithStart with same closed workflowID will caused Internal CurrentWorkflowConditionFailedError
retry in frontend to avoid affecting server side SLA

### DIFF
--- a/common/persistence/cassandra/cassandraPersistence.go
+++ b/common/persistence/cassandra/cassandraPersistence.go
@@ -1135,7 +1135,7 @@ func (d *cassandraPersistence) CreateWorkflowExecution(request *p.CreateWorkflow
 					}
 				}
 
-				if prevRunID, ok := previous["current_run_id"]; ok && prevRunID != request.Execution.GetRunId() {
+				if prevRunID := previous["current_run_id"].(gocql.UUID).String(); prevRunID != request.Execution.GetRunId() {
 					// currentRunID on previous run has been changed, return to caller to handle
 					msg := fmt.Sprintf("Workflow execution creation condition failed by mismatch runID. WorkflowId: %v, CurrentRunID: %v, columns: (%v)",
 						request.Execution.GetWorkflowId(), request.Execution.GetRunId(), strings.Join(columns, ","))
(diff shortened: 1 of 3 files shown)


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1135,8 +1135,15 @@ func (d *cassandraPersistence) CreateWorkflowExecution(request *p.CreateWorkflow
 					}
 				}
 
-				msg := fmt.Sprintf("Workflow execution creation condition failed. WorkflowId: %v, columns: (%v)",
-					request.Execution.GetWorkflowId(), strings.Join(columns, ","))
+				if prevRunID, ok := previous["current_run_id"]; ok && prevRunID != request.Execution.GetRunId() {
```
longquanzheng: I believe you need to use
```
previous["current_run_id"].(gocql.UUID).String()
```
Otherwise it always compare gocql.UUID with string.

vancexu: Changed, Thanks for catching this

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1135,8 +1135,15 @@ func (d *cassandraPersistence) CreateWorkflowExecution(request *p.CreateWorkflow
 					}
 				}
 
-				msg := fmt.Sprintf("Workflow execution creation condition failed. WorkflowId: %v, columns: (%v)",
-					request.Execution.GetWorkflowId(), strings.Join(columns, ","))
+				if prevRunID := previous["current_run_id"].(gocql.UUID).String(); prevRunID != request.Execution.GetRunId() {
```
yiminc: will this throw nil pointer error in case previous does not have "current_run_id"?

vancexu: There should be no case `current_run_id` is not in previous. But for resiliency, I can add a check which might be dead code.

PR #1172, opened 2018-10-10
Add exe checksum to DecisionTaskCompletedEvent

yiminc:
Add exe checksum to DecisionTaskCompletedEvent


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -446,6 +446,7 @@ struct DecisionTaskCompletedEventAttributes {
   20: optional i64 (js.type = "Long") scheduledEventId
   30: optional i64 (js.type = "Long") startedEventId
   40: optional string identity
+  50: optional string exeChecksum
```
mfateev: I think name is too specific. Not all languages use exe and checksum. May be something like buildTag?

PR #1176, opened 2018-10-12
Next batch of SQL persistence fixes

mfateev:
Most of the unit tests are passing now. The only disabled ones are related to cross DC.
There are multiple cleanups that are pending the initial POC release.


common/persistence/sql/sqlExecutionManager.go
```diff
@@ -469,42 +501,86 @@ func (m *sqlExecutionManager) Close() {
...
+}
+
+func (m *sqlExecutionManager) createWorkflowExecutionTx(tx *sqlx.Tx, request *p.CreateWorkflowExecutionRequest) (*p.CreateWorkflowExecutionResponse, error) {
+	if request.CreateWorkflowMode == p.CreateWorkflowModeContinueAsNew {
+		return nil, &workflow.InternalServiceError{
+			Message: "CreateWorkflowExecution operation failed. Invalid CreateWorkflowModeContinueAsNew is used",
+		}
+	}
 	/*
 		(x) make a transaction
```
venkat1109: this comment is out of date - we are already inside of a transaction

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1325,23 +1455,39 @@ func (m *sqlExecutionManager) RangeCompleteTimerTask(request *p.RangeCompleteTim
...
+		sqlManager: sqlManager{
+			db:     db,
+			logger: logger,
+		},
+	}, nil
 }
 
+// getCurrentExecutionIfExists returns current execution or nil if none is found for the workflowID
+// locking it in the DB
 func getCurrentExecutionIfExists(tx *sqlx.Tx, shardID int64, domainID string, workflowID string) (*currentExecutionRow, error) {
```
venkat1109: consider renaming to lockCurrentExecutionIfExist - this method no longer does a SELECT, it does a SELECT FOR UPDATE

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -469,42 +501,86 @@ func (m *sqlExecutionManager) Close() {
...
-
-	tx, err := m.db.Beginx()
-	if err != nil {
-		return nil, &workflow.InternalServiceError{
-			Message: fmt.Sprintf("CreateWorkflowExecution failed. Failed to start transaction. Error: %v", err),
-		}
+	var row *currentExecutionRow
+	var err error
+	workflowID := *request.Execution.WorkflowId
+	if row, err = getCurrentExecutionIfExists(tx, int64(m.shardID), request.DomainID, workflowID); err != nil {
```
venkat1109: Given that this is a locking read (and we do rangeID check after this), I am not sure if you still need the lockShard() towards the end of this method. I am not even sure if the locks acquired during a tx are re-entrant. So, this double acquisition of lock may not even work.

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1417,25 +1563,22 @@ func createCurrentExecution(tx *sqlx.Tx, request *p.CreateWorkflowExecutionReque
...
+		arg.LastWriteVersion = &replicationState.LastWriteVersion
 	}
 	if request.ParentExecution != nil {
 		arg.State = p.WorkflowStateCreated
 	}
 
 	switch request.CreateWorkflowMode {
 	case p.CreateWorkflowModeContinueAsNew:
-		if err := continueAsNew(tx,
+		if err := conditionalUpdateCurrentExecution(tx,
```
venkat1109: I don't think the method "conditionalUpdateCurrentExecution" actually does a conditional update. Name just adds confusion.

common/persistence/cassandra/cassandraMetadataPersistenceProxy.go
```diff
@@ -53,6 +53,10 @@ func newMetadataManagerProxy(cfg config.Cassandra,
 	return &metadataManagerProxy{metadataMgr: metadataMgr, metadataMgrV2: metadataMgrV2, logger: logger}, nil
 }
 
+func (m *metadataManagerProxy) GetName() string {
```
venkat1109: consider creating a baseClass that has just this method. Then all the implementations can simply inherit from this.

common/persistence/sql/common.go
```diff
@@ -22,14 +22,67 @@ package sql
...
+type sqlManager struct {
+	db     *sqlx.DB
+	logger bark.Logger
+}
+
+func (m *sqlManager) GetName() string {
+	return m.db.DriverName()
+}
+
+func (m *sqlManager) txExecute(operation string, f func(tx *sqlx.Tx) error) error {
```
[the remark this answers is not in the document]
mfateev: I'm going to merge them in a separate PR. This one is already too big.

common/persistence/sql/common.go
```diff
@@ -22,14 +22,67 @@ package sql
...
-
+	"github.com/go-sql-driver/mysql"
 	"github.com/jmoiron/sqlx"
+	"github.com/uber-common/bark"
 	workflow "github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common/persistence"
 	p "github.com/uber/cadence/common/persistence"
 )
 
+type sqlManager struct {
```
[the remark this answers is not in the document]
mfateev: Renamed. I"m going to rename all other managers to stores in a separate PR.

common/persistence/sql/sqlExecutionManager.go
mfateev: These are for different purpose. Current execution is locks only a current_execution row for the workflowID. The shard is locked to make sure that the whole shard lease wasn't lost. This record is shared by all workflows in the shard and that's why it is locked at the end of the transaction to minimize contention.

common/persistence/cassandra/cassandraMetadataPersistenceProxy.go
mfateev: Yes, I"m planning to do it in the follow up PRs for both stores and proxies.

common/persistence/sql/sqlExecutionManager.go
venkat1109: **> Current execution is locks only a current_execution row for the workflowID**
This is no longer true after your code change. lockCurrentExecutionIfExists does an INNER JOIN on shards table and so, it implicitly also locks the shard / rangeID. However, I verified that the locks acquired within a transaction are re-entrant and that's probably why the unit test succeed. Either ways, the lockShard() is no longer necessary AFAIK

PR #1186, opened 2018-10-16
Fix query task before first decision task completed

yiminc:
There are 2 specific problems that this diff is trying to solve.

1) If the query request comes in while the first decision task is not started yet. The existing implementation would dispatch a query task. That query task will contain WorkflowExecutionTaskStarted and DecisionTaskScheduled, but no DecisionTaskStarted. So the user's decider code has no chance to be executed, thus no query handler will be registered, so the query will fail because the query type is unknown.

2) If the query request comes in while the first decision was started but not completed yet. The existing implementation would dispatch a query task. That query task will contains the first DecisionTaskStarted event. However, current implementation would put previousStartedEventID to be the NextEventID so all event will be treated as replay. This will result in non-deterministic error because there is no corresponding history events. (https://github.com/uber/cadence/issues/1094)

This diff try to fix these 2 problems by blocking the query if the first decision task is not completed yet.


service/matching/matchingEngine.go
```diff
@@ -294,14 +298,15 @@ pollLoop:
...
-				completeType := workflow.QueryTaskCompletedTypeFailed
-				e.RespondQueryTaskCompleted(ctx, &m.RespondQueryTaskCompletedRequest{
-					TaskID: common.StringPtr(tCtx.queryTaskInfo.taskID),
-					CompletedRequest: &workflow.RespondQueryTaskCompletedRequest{
-						CompletedType: &completeType,
-						ErrorMessage:  common.StringPtr("server internal error: failed to get nextID " + err.Error()),
-					},
-				})
+				e.deliverQueryResult(tCtx.queryTaskInfo.taskID,
+					&queryResult{err: err, waitNextEventID: mutableStateResp.GetNextEventId()})
```
samarabbas: No need to set waitNextEventID on the result

service/matching/matchingEngine.go
```diff
@@ -256,6 +256,8 @@ func (e *matchingEngineImpl) AddActivityTask(addRequest *m.AddActivityTaskReques
 	return tlMgr.AddTask(addRequest.Execution, taskInfo)
 }
 
+var errQueryBeforeFirstDecisionCompleted = errors.New("query cannot be handled before first decision task is processed, please retry later")
```
samarabbas: We should use QueryFailedError and let user retry.

service/matching/matchingEngine.go
```diff
@@ -422,50 +426,94 @@ func (e *matchingEngineImpl) QueryWorkflow(ctx context.Context, queryRequest *m.
...
+
+		select {
+		case result := <-queryResultCh:
+			if result.err == nil {
+				return &workflow.QueryWorkflowResponse{QueryResult: result.result}, nil
+			}
+
+			if result.err == errQueryBeforeFirstDecisionCompleted {
+				// query before first decision completed, so wait for first decision task to complete
+				expectedNextEventID := result.waitNextEventID
```
samarabbas: Can we initial expectedNextEventID to some value like 'firstEventID' or something before we call GetMutableState first time.  Then we don't need to return it from QueryResult.  Then once wait_loop calls GetMutableState we set it for future calls.

yiminc: That would require an additional round trip from matching to history. We do GetMutableState() to block and wait for the "expectedNextEventID". If we initialize it to firstEventID, the first call to GetMutableState() will not block and is a waste.

service/matching/matchingEngine.go
yiminc: will update

service/matching/matchingEngine.go
yiminc: as discussed offline, will keep this behavior to wait/retry on server.

PR #1187, opened 2018-10-17
Fix sticky query timeout

yiminc:
The problem is the default timeout of our CLI is 5s which is the same as default sticky timeout. So when the sticky timeout happen, the ctx is also expire, so the call to ResetSticky always fail on timeout error. This fix is to make sure we at least has 5s to do the ResetSticky.


service/frontend/workflowHandler.go
```diff
@@ -2072,7 +2072,8 @@ func (wh *WorkflowHandler) QueryWorkflow(ctx context.Context,
 		}
 		// this means sticky timeout, should try using the normal tasklist
 		// we should clear the stickyness of this workflow
-		_, err = wh.history.ResetStickyTaskList(ctx, &h.ResetStickyTaskListRequest{
+		resetContext, _ := context.WithTimeout(context.Background(), 5*time.Second)
+		_, err = wh.history.ResetStickyTaskList(resetContext, &h.ResetStickyTaskListRequest{
```
wxing1292: i think this error should be ignored

```suggestion
		wh.history.ResetStickyTaskList(resetContext, &h.ResetStickyTaskListRequest{
```

PR #1191, opened 2018-10-18
Out of order history events

samarabbas:
Sometimes we see buffered events could be out of order when loaded from
database.  This is not an issue on most cases but becomes a problem when
activity completion events is processed before activity started event.
This change reorders the processing of buffered events to always process
activity completion events at the end.


service/history/mutableStateBuilder.go
```diff
@@ -268,14 +268,39 @@ func (e *mutableStateBuilder) FlushBufferedEvents() error {
...
+	// except in the Activity case where ActivityStarted and ActivityCompleted gets out of order.  The following code
+	// is added to reorder buffered events to guarantee all activity completion events will always be processed at the end.
+	var reorderedEvents []*workflow.HistoryEvent
+	reorderFunc := func(bufferedEvents []*workflow.HistoryEvent) {
+		for _, e := range bufferedEvents {
+			switch e.GetEventType() {
+			case workflow.EventTypeActivityTaskCompleted,
+				workflow.EventTypeActivityTaskFailed,
+				workflow.EventTypeActivityTaskCanceled,
+				workflow.EventTypeActivityTaskCancelRequested,
```
yiminc: this one is not close event, and it does not have StartedID

PR #1193, opened 2018-10-18
mysql: convert sqlx batch deletes to normal deletes

venkat1109:
Looks like there are couple of places in code [ timer_info_maps, activity_info_maps ] where we simply pass a slice of keys to a single delete query hoping that the sqlx library will convert them into a batch delete. This doesn't really work.

This patch converts these *batch deletes* into simple deletes. These deletes happen within a transaction, so there is no risk of partial deletes.

### DIFF
--- a/common/persistence/sql/workflowStateMaps.go
+++ b/common/persistence/sql/workflowStateMaps.go
@@ -309,38 +309,37 @@ func updateActivityInfos(tx *sqlx.Tx,
 	}
 
 	if len(deleteInfos) > 0 {
-		activityInfoMapsPrimaryKeys := make([]*activityInfoMapsPrimaryKey, len(deleteInfos))
-		for i, v := range deleteInfos {
-			activityInfoMapsPrimaryKeys[i] = &activityInfoMapsPrimaryKey{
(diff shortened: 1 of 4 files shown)


common/persistence/sql/workflowStateMaps.go
```diff
@@ -309,38 +309,37 @@ func updateActivityInfos(tx *sqlx.Tx,
...
-		}
 
-		query, args, err := tx.BindNamed(deleteKeyInActivityInfoMapSQLQuery, activityInfoMapsPrimaryKeys)
-		if err != nil {
-			return &workflow.InternalServiceError{
-				Message: fmt.Sprintf("Failed to update activity info. Failed to bind query. Error: %v", err),
+			query, args, err := tx.BindNamed(deleteKeyInActivityInfoMapSQLQuery, deleteKeys)
+			if err != nil {
+				return &workflow.InternalServiceError{
+					Message: fmt.Sprintf("Failed to update activity info. Failed to bind query. Error: %v", err),
```
mfateev: May be more specific error message?

PR #1211, opened 2018-10-30
Emit metric on history sizes by frontend

samarabbas:
Emit history size metric from frontend API which loads history
from persistence.  Also log a warning with domainID, WorkflowID,
and RunID when history returned is larger than 500KB.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -2269,7 +2269,7 @@ func (wh *WorkflowHandler) getHistory(scope int, domainID string, execution gen.
 				logging.TagWorkflowRunID:       execution.GetRunId(),
 				logging.TagDomainID:            domainID,
 				logging.TagSize:                response.Size,
-			})
+			}).Warn("GetHistory size threshold breached")
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


service/frontend/workflowHandler.go
```diff
@@ -2256,6 +2260,19 @@ func (wh *WorkflowHandler) getHistory(domainID string, execution gen.WorkflowExe
 		return nil, nil, err
 	}
 
+	if response != nil {
+		wh.metricsClient.RecordTimer(scope, metrics.HistorySize, time.Duration(response.Size))
+
+		if response.Size > getHistoryWarnSizeLimit {
+			wh.GetLogger().WithFields(bark.Fields{
```
longquanzheng: Looks like you forget to call .Infof() or Warnf()

service/frontend/workflowHandler.go
```diff
@@ -2256,6 +2260,19 @@ func (wh *WorkflowHandler) getHistory(domainID string, execution gen.WorkflowExe
 		return nil, nil, err
 	}
 
+	if response != nil {
+		wh.metricsClient.RecordTimer(scope, metrics.HistorySize, time.Duration(response.Size))
```
longquanzheng: Is it possible to also include domain here?

samarabbas: Currently we can't emit domain as a tag.  Added a log instead.

PR #1213, opened 2018-10-30
Reject terminting a WF that just get started

longquanzheng:


service/history/historyEngine.go
```diff
@@ -2173,6 +2175,23 @@ func (e *historyEngineImpl) RemoveSignalMutableState(ctx context.Context, reques
...
+	if err0 != nil {
+		return err0
+	}
+	defer func() { release(retError) }()
+	msBuilder, retError := context.loadWorkflowExecution()
+	if retError != nil {
+		return retError
+	}
+	// terminating a WF that just get started would be wasting resources, and causing problem without eventsV2
+	if time.Now().Before(msBuilder.GetExecutionInfo().StartTimestamp.Add(time.Second * 3)) {
```
wxing1292: make this a constant or a dynamic config

wxing1292: one more thing, this patch is "treating the symptom not the disease", the correct way is to include the previous run id in the start event, so when the start event is applied, order between run ID can be maintained.

longquanzheng: This will also allow customer to start-terminate-start within very short time of period which is wasting resources. Or another way I am thinking is to reject starting the same wfID within short time.

service/history/historyEngine.go
```diff
@@ -2173,6 +2176,23 @@ func (e *historyEngineImpl) RemoveSignalMutableState(ctx context.Context, reques
...
+	if err0 != nil {
+		return err0
+	}
+	defer func() { release(retError) }()
+	msBuilder, retError := context.loadWorkflowExecution()
+	if retError != nil {
+		return retError
+	}
+	// terminating a WF that just get started would be wasting resources, and causing problem without eventsV2
+	if time.Now().Before(msBuilder.GetExecutionInfo().StartTimestamp.Add(minDurationSinceStarted)) {
```
samarabbas: Can you do this check within the update handler for Terminate? Then you don't have to duplicate the logic for loading workflow context.

PR #1215, opened 2018-10-31
Add log for more visibility on timer tasks

vancexu:


service/history/shardContext.go
```diff
@@ -818,7 +818,11 @@ func (s *shardContextImpl) allocateTimerIDsLocked(timerTasks []persistence.Task)
 		if ts.Before(s.timerMaxReadLevelMap[cluster]) {
 			// This can happen if shard move and new host have a time SKU, or there is db write delay.
 			// We generate a new timer ID using timerMaxReadLevel.
-			s.logger.Warnf("%v: New timer generated is less than read level. timestamp: %v, timerMaxReadLevel: %v",
+			s.logger.WithFields(bark.Fields{
+				logging.TagWorkflowEventID: logging.ShardAllocateTimerBeforeRead,
+				"Domain":                   domainID,
```
samarabbas: There is a tag for DomainID and WorkflowID, can we use the defined constants under logging.

PR #1217, opened 2018-10-31
Add metrics and logs to visibility

vancexu:
Metrics are already there in persistence, so just add log


service/frontend/workflowHandler.go
```diff
@@ -114,6 +114,10 @@ var (
 
 const (
 	getHistoryWarnSizeLimit = 500 * 1024 // Warn when size goes over 500KB
+
+	listWorkflowFilterByID     = "WID"
```
samarabbas: Can we define these under logging constants?  Basically these are well defined values for FilterType tag.

PR #1218, opened 2018-10-31
Support events v2 for workflow history

longquanzheng:
I will add tests and code for CassandraPersistence after the main logic is reviewed


common/persistence/dataInterfaces.go
```diff
@@ -249,6 +251,18 @@ type (
 		ExpirationTime     time.Time
 		MaximumAttempts    int32
 		NonRetriableErrors []string
+		// events V2 related
+		EventStoreVersion   int32
+		CurrentResetVersion int32
+		HistoryTreeID       string
+		HistoryBranches     map[int32]*HistoryBranch // map from each resetVersion to the associated branch, resetVersion increase from 0
```
samarabbas: Is this cached in-memory or written to the database?

longquanzheng: HistoryBranches will be written into database as part of mutableState. 
For now this map will only have one item.
Actually it is not required to have a map (resetVersion->branch) if we don't implement reset. We can just use a field of branch. But I am going to implement reset very soon, it is not good to have a field that will be deprecated.

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
```diff
@@ -25,4 +25,17 @@ CREATE TABLE history_tree (
   PRIMARY KEY ((tree_id), row_type, branch_id )
 ) WITH COMPACTION = {
 'class': 'org.apache.cassandra.db.compaction.LeveledCompactionStrategy'
-};
\ No newline at end of file
+};
+
+CREATE TYPE history_branch_info (
+  branch_token                     blob,
```
samarabbas: Can you put a comment on what's encoded in this branch token?

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
```diff
@@ -25,4 +25,17 @@ CREATE TABLE history_tree (
...
+  next_event_id                    bigint,
+  last_first_event_id              bigint,
+  history_size                     bigint,
+);
+
+--- using eventsV2
+ALTER TYPE workflow_execution ADD  event_store_version             int;
```
samarabbas: Do we really need this?  Can we rely on history_tree_id being set as a mechanism to check if history is stored as v1 vs v2?

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
```diff
@@ -25,4 +25,17 @@ CREATE TABLE history_tree (
...
 'class': 'org.apache.cassandra.db.compaction.LeveledCompactionStrategy'
-};
\ No newline at end of file
+};
+
+CREATE TYPE history_branch_info (
+  branch_token                     blob,
+  next_event_id                    bigint,
+  last_first_event_id              bigint,
+  history_size                     bigint,
```
samarabbas: This is for history size for single branch?  Or it also accumulates all ancestor branches?  Please also include a comment.

common/persistence/dataInterfaces.go
```diff
@@ -249,6 +251,18 @@ type (
 		ExpirationTime     time.Time
 		MaximumAttempts    int32
 		NonRetriableErrors []string
+		// events V2 related
+		EventStoreVersion   int32
+		CurrentResetVersion int32
```
samarabbas: Do we need to include this reset version on all transfer/timer tasks? Also I think we this on history events too for cross dc replication.

common/persistence/executionStore.go
```diff
@@ -21,6 +21,7 @@
 package persistence
 
 import (
+	"code.uber.internal/devexp/cadence-iot-demo/.tmp/.go/goroot/src/fmt"
```
samarabbas: remove this.

idl/github.com/uber/cadence/history.thrift
```diff
@@ -74,6 +74,8 @@ struct GetMutableStateResponse {
   90: optional string clientImpl
   100: optional bool isWorkflowRunning
   110: optional i32 stickyTaskListScheduleToStartTimeout
+  120: optional i32 eventStoreVersion
+  130: optional binary branchToken
```
samarabbas: What about treeID?

service/frontend/workflowHandler.go
```diff
@@ -788,7 +790,11 @@ func (wh *WorkflowHandler) PollForDecisionTask(
 		return nil, nil
 	}
 
-	resp, err := wh.createPollForDecisionTaskResponse(ctx, domainID, matchingResp)
+	useEventsV2 := false
+	if matchingResp.GetEventStoreVersion() == 2 {
```
samarabbas: Let's define a constant for history events v2 in common.utils.

common/service/dynamicconfig/constants.go
```diff
@@ -321,6 +322,8 @@ const (
 	// AdminOperationToken is the token to pass admin checking
 	AdminOperationToken
 
+	// EnableEventsV2 is whether to use eventsV2
+	EnableEventsV2
```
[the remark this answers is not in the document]
longquanzheng: OK, sure

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -401,6 +401,7 @@ struct WorkflowExecutionStartedEventAttributes {
   70: optional RetryPolicy retryPolicy
   80: optional i32 attempt
   90: optional i64 (js.type = "Long") expirationTimestamp
+  100: optional i32 eventStoreVersion
```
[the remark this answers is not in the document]
longquanzheng: yes, I will move it to replication task instead

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
longquanzheng: Yes sure

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
longquanzheng: OK sure

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
longquanzheng: I am going to remove that tree_id as we discussed. But even with that tree_id, I personally prefer to make it more straightforward if it doesn't introduce any issue(storage/perf) than using nil/not nil. Also checking nil with gocql is not always safe.

common/persistence/dataInterfaces.go
longquanzheng: Probably yes. Thanks for reminding. I will think about it in next PR for reset

common/persistence/executionStore.go
longquanzheng: Yes sure

idl/github.com/uber/cadence/history.thrift
longquanzheng: branchToken is enough for querying historyV2 because it includes treeID internally

schema/cassandra/cadence/versioned/v0.12/events_v2.cql
```diff
@@ -25,4 +25,17 @@ CREATE TABLE history_tree (
...
+);
+
+--- using eventsV2
+ALTER TYPE workflow_execution ADD  event_store_version             int;
+ALTER TYPE workflow_execution ADD  current_reset_version            int;
+ALTER TYPE workflow_execution ADD  history_tree_id                  uuid;
```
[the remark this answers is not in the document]
longquanzheng: Yes I will remove it.

PR #1223, opened 2018-11-05
Starts cadence client workers as part of worker role

andrewjdawson2016:
- Hosts cadence client workers inside worker role
- Adds client and retryable client for frontend service
- Follow up diff will use to this to support system workflows


service/worker/service.go
```diff
@@ -101,6 +112,14 @@ func (s *Service) Start() {
 		log.Fatalf("Fail to start replicator: %v", err)
 	}
 
+	client, err := base.GetClientFactory().NewFrontendClient()
+	if err != nil {
+		log.Fatalf("failed to create frontend client: %v", err)
+	}
+	retryableClient := frontend.NewRetryableClient(client, common.CreateFrontendServiceRetryPolicy(),
+		common.IsWhitelistServiceTransientError)
+	go s.startWorkers(retryableClient, log)
```
yiminc: startWorkers is not blocking call, we don't need a new goroutine.

client/frontend/client.go
```diff
@@ -21,17 +21,561 @@
...
+	ctx, cancel := c.createContext(ctx)
+	defer cancel()
+	return client.UpdateDomain(ctx, request, opts...)
 }
 
-// New creates a client to cadence frontend
-func New(d *yarpc.Dispatcher) Client {
-	return workflowserviceclient.New(d.ClientConfig(common.FrontendServiceName))
+func (c *clientImpl) createContext(parent context.Context) (context.Context, context.CancelFunc) {
+	// TODO: make timeout configurable
```
yiminc: what is the current timeout before this change?

client/frontend/client.go
```diff
@@ -21,17 +21,561 @@
...
+
+type clientImpl struct {
+	resolver        membership.ServiceResolver
+	thriftCacheLock sync.RWMutex
+	thriftCache     map[string]workflowserviceclient.Interface
+	rpcFactory      common.RPCFactory
+}
+
+// NewClient creates a new frontend service TChannel client
+func NewClient(d common.RPCFactory, monitor membership.Monitor) (Client, error) {
```
samarabbas: Can you also wrap this with a metricClient to emit client-side metric on the caller.  Similar pattern we do for other clients.

PR #1224, opened 2018-11-05
sql: support for getTimerIndexTasks pagination

venkat1109:
This patch adds pagination support for MySQL GetTimerIndexTasks API.

### DIFF
--- a/common/persistence/sql/sqlExecutionManager.go
+++ b/common/persistence/sql/sqlExecutionManager.go
@@ -21,9 +21,12 @@
 package sql
 
 import (
-	"database/sql"
 	"fmt"
+	"math"
 	"time"
+
+	"database/sql"
+	"encoding/json"
 
 	"github.com/jmoiron/sqlx"
 	"github.com/uber-common/bark"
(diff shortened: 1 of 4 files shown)


common/persistence/sql/sqlExecutionManager.go
```diff
@@ -476,8 +479,9 @@ LIMIT ?`
 		`
 FROM timer_tasks WHERE
 shard_id = ? AND
-visibility_timestamp >= ? AND
-visibility_timestamp < ?`
+((visibility_timestamp >= ? AND task_id >= ?) OR visibility_timestamp > ?) AND
+visibility_timestamp < ?
```
mfateev: This condition looks different from the one i'm using. Which would be:
visibility_timestamp < ? and visibility_timestamp >=? and (visibility_timestamp > ? OR task_id >= ?)

venkat1109: its different because - timer tasks are stored / queried in natural order (asc) whereas with your other case - data is queried in reverse sorted order

PR #1241, opened 2018-11-12
Handle mixed short & long workflow replication case

lihannan99:
All cases below have the precondition that incoming version >= current local version

* when sending the first event, include the previous run ID as well as corresponding version
* when replication, if start event does not have prev run ID & prev version, than just apply (existing behavior)
* when replication, if start event has incoming prev run ID & incoming prev version
** if incoming prev version == incoming version and incoming prev run ID == current local run ID, apply the event
** if incoming prev version == incoming version but incoming prev run ID != current local run ID, then retry (this is out or order case)
** if incoming version > incoming prev version, then just apply


common/persistence/persistence-tests/persistenceTestBase.go
```diff
@@ -197,9 +196,10 @@ func (s *TestBase) Setup() {
 		RangeID:                 0,
 		TransferAckLevel:        0,
 		ReplicationAckLevel:     0,
-		TimerAckLevel:           time.Time{},
-		ClusterTimerAckLevel:    map[string]time.Time{clusterName: time.Time{}},
+		TimerAckLevel:           time.Now(),
+		ClusterTimerAckLevel:    map[string]time.Time{clusterName: time.Now()},
```
vancexu: Should this time be same as `TimerAckLevel` for testClusterName?

schema/cassandra/cadence/versioned/v0.12/manifest.json
```diff
@@ -3,6 +3,7 @@
   "MinCompatibleVersion": "0.12",
   "Description": "Support history events v2",
   "SchemaUpdateCqlFiles": [
+    "workflow_order.cql",
```
vancexu: I think 0.12 is already released in Cadence v0.4.0, so you need to create a new versioned change v0.13, please also update description.

service/history/historyEngine.go
```diff
@@ -43,6 +40,7 @@ import (
 	"github.com/uber/cadence/common/messaging"
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
+	"go.uber.org/yarpc"
```
vancexu: where is this import used?

schema/cassandra/cadence/versioned/v0.12/manifest.json
vancexu: I think 0.12 is already released in Cadence v0.4.0, so you need to create a new versioned change v0.13, please also update description.

PR #1254, opened 2018-11-15
Do not start client worker as part of worker role

andrewjdawson2016:


andrewjdawson2016:
When the cadence client worker is started and sits idol, we see error logs from worker's activity poller. These logs are not a correctness concern but they are confusing to teammates running locally. Until I submit a follow up diff to fix this, I will simply not start cadence worker as part of worker role. I will however keep all the plumbing from the last diff that makes it possible to run the client worker.

PR #1256, opened 2018-11-15
Add publish visibility for open workflow to Kafka

vancexu:
Refactor messaging client for publish visibility data to kafka,
Add publish open workflow data to kafka in transferQueue.
This doesn't include consumer yet.


config/development.yaml
```diff
@@ -73,3 +73,12 @@ clustersInfo:
   clusterInitialFailoverVersion:
     active: 0
     standby: 1
+
+kafkaCommon:
```
longquanzheng: Just curious is there any topics using it other than visibility-topic?

longquanzheng: Can we combine this config with kafka ? https://github.com/uber/cadence/blob/master/config/development_active.yaml#L77

It will be confusing to see two configs: kafka and kafkaCommon

vancexu: 1. Yes, there will be at least topic for workflow insight. 
2. That could be done, but the difference is Kafka config for replication is cadence cluster specific, but for visibility is not. Also, there might be case people want to enable visibility to kafka but not enable cross dc. That's why I currently make it like this.

service/history/historyEngine.go
```diff
@@ -3040,3 +3041,22 @@ func getWorkflowAlreadyStartedError(errMsg string, createRequestID string, workf
 		RunId:          common.StringPtr(fmt.Sprintf("%v", runID)),
 	}
 }
+
+// change default policy to "AllowDuplicate" for signalWithStart if not set in request.
+func getWorkflowIDReusePolicyForSigStart(policy *workflow.WorkflowIdReusePolicy) workflow.WorkflowIdReusePolicy {
```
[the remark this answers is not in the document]
vancexu: Got it, removed.

config/development.yaml
longquanzheng: Enabling cross DC is another config: https://github.com/uber/cadence/blob/master/config/development_active.yaml#L69
So it is fine to add other kafka config there

common/service/dynamicconfig/constants.go
```diff
@@ -51,6 +51,7 @@ var keys = map[Key]string{
 	EnableGlobalDomain:       "system.enableGlobalDomain",
 	EnableNewKafkaClient:     "system.enableNewKafkaClient",
 	EnableVisibilitySampling: "system.enableVisibilitySampling",
+	EnableKafkaCommon:        "system.enableKafkaCommon",
```
longquanzheng: I would recommend to make it more specific for each feature: enableKafkaVisibility / enableKafkaInsight. Then we will have better control over each feature. Also EnableKafkaCommon is too generic as feature flag

PR #1259, opened 2018-11-16
Add all plumbing for system workflows

andrewjdawson2016:
- Sending signal for archival to system workflows
- Sharding system workflows
- Added needed config
- Stubbed out archival activity with logger
- Refactored worker to separate sysworker and replicator


service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Archive starts an archival task
+func (i *initiator) Archive(request *ArchiveRequest) error {
+	workflowID := PickRandom(i.sysWorkflowIDs)
+	workflowOptions := client.StartWorkflowOptions{
+		ID: workflowID,
+		// TODO: once we have higher load, this should select one random of X task lists to do load balancing
+		TaskList:                        DecisionTaskList,
+		ExecutionStartToCloseTimeout:    time.Hour * 48,
```
samarabbas: This will cause workflow to timeout after 48 hours.  How do you guarantee new run before the timeout?  Or prevent existing archival will not be lost.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,87 @@
...
+		DecisionTaskStartToCloseTimeout: time.Minute * 10,
```
samarabbas: 10 minute for decision timeout is too long.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,87 @@
...
+		RequestType:    ArchivalRequest,
+		ArchiveRequest: request,
+	}
+	_, err := i.cadenceClient.SignalWithStartWorkflow(
+		context.Background(),
+		workflowID,
+		SignalName,
+		signal,
+		workflowOptions,
+		SystemWorkflow,
```
samarabbas: Let's invoke using string name for the workflow.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,87 @@
...
+func NewInitiator(frontendClient frontend.Client, numSysWorkflows int) Initiator {
+	return &initiator{
+		cadenceClient:  client.NewClient(frontendClient, Domain, &client.Options{}),
+		sysWorkflowIDs: RandomSlice(numSysWorkflows),
+	}
+}
+
+// Archive starts an archival task
+func (i *initiator) Archive(request *ArchiveRequest) error {
+	workflowID := PickRandom(i.sysWorkflowIDs)
```
samarabbas: I'm a little confused by this pattern.  Won't this create new workflow IDs for system workflow each time you restart the server?  How do you control the number of concurrent system workflow execution?
It would be good to put a comment around expectation here.

service/history/historyEngine.go
```diff
@@ -152,6 +163,7 @@ func NewEngineWithShardContext(shard ShardContext, visibilityMgr persistence.Vis
 		}),
 		metricsClient:        shard.GetMetricsClient(),
 		historyEventNotifier: historyEventNotifier,
+		initiator:            sysworkflow.NewInitiator(frontendClient, shard.GetConfig().NumSysWorkflows()),
```
samarabbas: It would be good to make this config truly dynamic which does not need a restart of the roles.

service/worker/sysworkflow/initiator.go
andrewjdawson2016: What do you think about just making this duration very long?

service/worker/sysworkflow/initiator.go
andrewjdawson2016: noted

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,87 @@
...
+		ExecutionStartToCloseTimeout:    time.Hour * 48,
+		DecisionTaskStartToCloseTimeout: time.Minute * 10,
+		WorkflowIDReusePolicy:           client.WorkflowIDReusePolicyAllowDuplicate,
+	}
+	signal := Signal{
+		RequestType:    ArchivalRequest,
+		ArchiveRequest: request,
+	}
+	_, err := i.cadenceClient.SignalWithStartWorkflow(
+		context.Background(),
```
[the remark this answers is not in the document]
andrewjdawson2016: The frontend client client that is being used has a timeout of 10 seconds https://github.com/uber/cadence/blob/master/client/frontend/client.go#L41

service/worker/sysworkflow/constants.go
```diff
@@ -0,0 +1,59 @@
...
+	DecisionTaskList = "cadsys-decision-tl"
+
+	// SignalName is the name of the cadence signal that system tasks are sent on
+	SignalName = "cadsys-signal-sig"
+
+	// WorkflowIDPrefix is the prefix of all system workflow ids
+	WorkflowIDPrefix = "cadsys-wf"
+
+	// SignalsUntilContinueAsNew is the number of signals system workflow must receive before continuing as new
+	SignalsUntilContinueAsNew = 1000
```
andrewjdawson2016: This feels somewhat arbitrary. I think we need to do some load testing to figure out a reasonable value here.

service/worker/sysworkflow/constants.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// SystemWorkflowScope scope for all metrics emitted by system workflow
+	SystemWorkflowScope = "system-workflow"
```
andrewjdawson2016: I do not like that we have these metric constants here. I could not figure out how to make our existing metrics defs.go work directly with tally without making changes that impact how we do metrics system wide.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,92 @@
...
+func NewInitiator(frontendClient frontend.Client, numSWFn dynamicconfig.IntPropertyFn) Initiator {
+	return &initiator{
+		cadenceClient: client.NewClient(frontendClient, Domain, &client.Options{}),
+		numSWFn:       numSWFn,
+	}
+}
+
+// Archive starts an archival task
+func (i *initiator) Archive(request *ArchiveRequest) error {
+	workflowID := fmt.Sprintf("%v-%v", WorkflowIDPrefix, rand.Intn(i.numSWFn()))
```
andrewjdawson2016: I think this pattern is okay. The basic idea is 

(1) We might have more system workflows then last time we read numSWFn. In this case we just start spreading new load evenly among the added system workflows
(2) If we reduced the number of system workflows, then the system workflows now out of range will sit ideal not doing anything until they expire.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,92 @@
...
+		ArchiveRequest: request,
+	}
+
+	_, err := i.cadenceClient.SignalWithStartWorkflow(
+		context.Background(),
+		workflowID,
+		SignalName,
+		signal,
+		workflowOptions,
+		"SystemWorkflow",
```
andrewjdawson2016: Yes I like this pattern by naming by string more

service/worker/sysworkflow/system_workflow.go
```diff
@@ -0,0 +1,106 @@
...
+
+		ao := workflow.ActivityOptions{
+			ScheduleToStartTimeout: time.Minute,
+			StartToCloseTimeout:    time.Minute,
+			HeartbeatTimeout:       time.Second * 10,
+			RetryPolicy: &cadence.RetryPolicy{
+				InitialInterval:          time.Second,
+				BackoffCoefficient:       2.0,
+				MaximumInterval:          time.Minute,
+				ExpirationInterval:       time.Hour * 24 * 30,
```
andrewjdawson2016: Basically my reasoning here is we set this value large enough that we can be certain we will get 1000 signals before this interval is up, causing us to continue as new.

service/worker/sysworkflow/initiator.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Archive starts an archival task
+func (i *initiator) Archive(request *ArchiveRequest) error {
+	workflowID := fmt.Sprintf("%v-%v", WorkflowIDPrefix, rand.Intn(i.numSWFn()))
+	workflowOptions := client.StartWorkflowOptions{
+		ID: workflowID,
+		// TODO: once we have higher load, this should select one random of X task lists to do load balancing
+		TaskList:                        DecisionTaskList,
+		ExecutionStartToCloseTimeout:    time.Hour * 24 * 30,
```
andrewjdawson2016: Basically my reasoning here is we set this value large enough that we can be certain we will get 1000 signals before this interval is up, causing us to continue as new.

service/worker/sysworkflow/initiator.go
andrewjdawson2016: Updated from my last diff is that these workflowIDs are now stable, so the same workflows will be used upon role restart.

service/worker/sysworkflow/sysworker.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return &SysWorker{
+		// TODO: after we do task list fan out workers should listen on all task lists
+		worker: worker.New(frontendClient, Domain, DecisionTaskList, wo),
+	}
+}
+
+// Start the SysWorker
+func (w *SysWorker) Start() error {
+	workflow.RegisterWithOptions(SystemWorkflow, workflow.RegisterOptions{Name: "SystemWorkflow"})
```
samarabbas: The usual pattern is to register workflows and activities on "init".  Is there a specific reason to register them here on start?

service/worker/sysworkflow/system_workflow.go
```diff
@@ -0,0 +1,106 @@
...
+	for i := 0; i < SignalsUntilContinueAsNew; i++ {
+		var signal Signal
+		if more := ch.Receive(ctx, &signal); !more {
+			scope.Counter(ChannelClosedUnexpectedlyError).Inc(1)
+			logger.Error("cadence channel was unexpectedly closed")
+			break
+		}
+		scope.Counter(HandledSignalCount).Inc(1)
+
+		ao := workflow.ActivityOptions{
```
samarabbas: Probably it's better to do this once outside of the loop.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -0,0 +1,106 @@
...
+			)
+		default:
+			scope.Counter(UnknownSignalTypeErr).Inc(1)
+			logger.Error("received unknown request type")
+		}
+	}
+
+	logger.Info("completed current set of iterations, continuing as new",
+		zap.Int(logging.TagIterationsUntilContinueAsNew, SignalsUntilContinueAsNew))
+	return workflow.NewContinueAsNewError(ctx, "SystemWorkflow")
```
samarabbas: Make sure to drain all signals using ReceiveAsync before doing ContinueAsNew.

service/worker/sysworkflow/initiator.go
samarabbas: Can you define constants for system workflow and activity names.

service/worker/sysworkflow/initiator.go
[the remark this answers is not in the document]
samarabbas: Got it.  Looks good then.

service/worker/sysworkflow/constants.go
samarabbas: All metrics should go under https://github.com/uber/cadence/blob/master/common/metrics/defs.go#L581

PR #1276, opened 2018-11-26
Add archival interface

andrewjdawson2016:
- Adds archival interface, which is a simplified version of terrablob/s3 interface
- Adds nop impl of the interface
- Plumbs through archival client into system workflow


common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+	Owner     *string
+	Size      *int64
+	Tags      map[string]string
+}
+
+// UploadBlobRequest is request for UploadBlob
+type UploadBlobRequest struct {
+	Path       *string
+	Body       io.Reader
+	Size       *int64
```
venkat1109: Size may or may not be known at the time of upload i.e. document that this is an optional param

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+	CreateDirs *bool
```
venkat1109: do we need this ?

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+	Tags       map[string]string
+	ChunkSize  *int64
```
venkat1109: what does this represent ?

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+// UploadBlobResponse is response from UploadBlob
+type UploadBlobResponse struct {
+	Status WriteStatus
+}
+
+// DownloadBlobRequest is request for DownloadBlob
+type DownloadBlobRequest struct {
+	Path        *string
+	Writer      io.WriterAt
+	IsStreaming *bool
```
venkat1109: see comment above - for the caller, there should be no difference between streaming / non-streaming

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// MakeDirectoryRequest is request for MakeDirectory
+type MakeDirectoryRequest struct {
+	Path  *string
+	Owner *string
+}
+
+// ListDirectoryRequest is request for ListDirectory
+type ListDirectoryRequest struct {
```
venkat1109: can we rename this to ListByPrefix - assuming this API is also capable of listing by Prefix

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+	// ItemTypeBlob indicates that the object is a blob
+	ItemTypeBlob
+)
+
+// ListItem is a single listing in ListDirectoryResponse
+type ListItem struct {
+	Type      ItemType
+	Name      *string
+	CreatedAt *time.Time
+	Owner     *string
```
venkat1109: some of this information is duplicated in MetadataResponse as well - consider creating a separate type called - ObjectMetadata or BlobMetadata for this info

common/archival/client.go
```diff
@@ -0,0 +1,166 @@
...
+func (mc *nopClient) MakeDirectory(ctx context.Context, request *MakeDirectoryRequest) error {
+	return errors.New("method not supported")
+}
+
+func (mc *nopClient) ListDirectory(ctx context.Context, request *ListDirectoryRequest) (*ListDirectoryResponse, error) {
+	return nil, errors.New("method not supported")
+}
+
+// NewNopClient creates a nop client
+func NewNopClient() Client {
```
venkat1109: Where are the compression, security settings to be specified ?

Issue #800, opened 2018-05-30
XDC - Consolidate the Next Event ID & Failover Version Check


wxing1292:

wxing1292:
close due to no longer relevant

PR #1280, opened 2018-11-26
Optimize failover performance by batching domain failovers

wxing1292:
* domain cache will periodically scan domain v2 table for domain updates
* domain updates will be sent in batch to history engine and every domain failover in this period will be handled by one failover processor

solve #989, #800


common/cache/domainCache.go
```diff
@@ -199,39 +191,43 @@ func (c *domainCache) GetAllDomain() map[string]*DomainCacheEntry {
...
+	prevEntries := []*DomainCacheEntry{}
+	nextEntries := []*DomainCacheEntry{}
+	for _, domain := range domains {
+		if domain.notificationVersion >= initialNotificationVersion {
+			prevEntries = append(prevEntries, nil)
+			nextEntries = append(nextEntries, domain)
 		}
 	}
+	if len(prevEntries) > 0 {
+		prepareCallback()
```
samarabbas: Can we simplify this have just single callback?  We seems to be calling them immediately one after another.  I don't see why we need to register 2 separate callbacks.

common/cache/domainCache.go
```diff
@@ -199,39 +191,43 @@ func (c *domainCache) GetAllDomain() map[string]*DomainCacheEntry {
...
 }
 
 // UnregisterDomainChangeCallback delete a domain failover callback
 func (c *domainCache) UnregisterDomainChangeCallback(shard int) {
 	c.Lock()
 	defer c.Unlock()
 
-	delete(c.beforeCallbacks, shard)
-	delete(c.afterCallbacks, shard)
+	delete(c.callbacks, shard)
```
samarabbas: what about the second map 'prepareCallbacks'?

service/history/historyEngine.go
```diff
@@ -236,27 +234,37 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
...
+			e.timerProcessor.LockTaskPrrocessing()
 		},
-		// after the domain change, this will be invoked when domain cache is NOT locked
-		func(prevDomain *cache.DomainCacheEntry, nextDomain *cache.DomainCacheEntry) {
-			failoverPredicate(nextDomain, func() {
-				e.logger.Infof("Domain Failover Notify Active: Shard: %v, Domain: %v, ID: %v\n",
-					e.shard.GetShardID(), nextDomain.GetInfo().Name, nextDomain.GetInfo().ID)
+		func(prevDomains []*cache.DomainCacheEntry, nextDomains []*cache.DomainCacheEntry) {
+			defer func() {
+				e.txProcessor.UnlockTaskPrrocessing()
```
samarabbas: It will be good to reduce the lock on processors, for the duration of taking the snapshot and creation of failover cursors.

service/history/historyEngine.go
```diff
@@ -236,27 +234,37 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
...
-				e.logger.Infof("Domain Failover Start: Shard: %v, Domain: %v, ID: %v\n",
-					e.shard.GetShardID(), nextDomain.GetInfo().Name, nextDomain.GetInfo().ID)
-
-				domainID := nextDomain.GetInfo().ID
-				e.txProcessor.FailoverDomain(domainID)
-				e.timerProcessor.FailoverDomain(domainID)
-			})
+		e.shard.GetDomainNotificationVersion(),
+		func() {
+			e.txProcessor.LockTaskPrrocessing()
```
samarabbas: Can you evaluate if 2 callback handlers are still needed.  Looks like this logic can easily be moved to callback handler.

common/cache/domainCache.go
```diff
@@ -199,39 +191,43 @@ func (c *domainCache) GetAllDomain() map[string]*DomainCacheEntry {
...
+	// we mush notify the change in a ordered fashion
+	// since history shard have to update the shard info
+	// with domain change version.
+	sort.Sort(domains)
+
+	prevEntries := []*DomainCacheEntry{}
+	nextEntries := []*DomainCacheEntry{}
+	for _, domain := range domains {
+		if domain.notificationVersion >= initialNotificationVersion {
+			prevEntries = append(prevEntries, nil)
```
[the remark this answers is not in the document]
wxing1292: the idea is to match the size of prev entries and next entries.
for catch up case (shard got loaded and try to catch up with domain change), since we do not store a prev version of the domain, we put nil here.

common/cache/domainCache.go
wxing1292: good catch

service/history/historyEngine.go
wxing1292: checked the start of the failover logic, the logic is pretty light, so it should not take much time

common/cache/domainCache.go
wxing1292: this one is the catch up logic, meaning that there is no domain change involved when shard register the callback.

the actual logic is shown below, basically pause all shards, then do domain updates, then trigger a snapshot of shard ack levels from min to max and start failover.
this will guarantee that all task after the min ack level will be progressed as active task at least once.

from:
https://github.com/uber/cadence/pull/1280/files/886b3449137fd6db32f87827cb716f2a534cb503#diff-f8e74f79013d8fa5cd8f4fa32c12d21fR316
to:
https://github.com/uber/cadence/pull/1280/files/886b3449137fd6db32f87827cb716f2a534cb503#diff-f8e74f79013d8fa5cd8f4fa32c12d21fR337

Issue #1229, opened 2018-11-08
Starting and Terminating workflows in quick sequence cause messages in DLQ


samarabbas:
If user repeatedly starts and terminates workflow very fast then it could cause messages to go into DLQ for the previous terminated run if the new starts get replicated before the previous start for terminated run.

PR #1281, opened 2018-11-26
Handle mixed short & long workflow replication case

wxing1292:
solve #1229


wxing1292:
close in favor of #1512

PR #1285, opened 2018-11-28
Teach cadence server to bind on any IP specified in configuration

artms:
In some environments (to satisfy difference between local laptop environment and CI) we need to make sure cadence binds on all interfaces


artms:
Our internal integration framework creates isolated docker network and during local development it is exporting ports back to host system - here we need to make sure cadence server binds on external IP of docker container (docker is not exporting binds on localhost of container), when we run same test inside CI environment - we're not exporting anything to host machine and run test in that isolated network and here we need cadence to bind on `127.0.0.1`. Binding on all interfaces will satisfy both environments and it is defaul behavior of yarpc apps (afaik).

common/service/config/config.go
```diff
@@ -69,6 +69,8 @@ type (
 		Port int `yaml:"port"`
 		// BindOnLocalHost is true if localhost is the bind address
 		BindOnLocalHost bool `yaml:"bindOnLocalHost"`
+		// BindOnLocalHost is true if localhost is the bind address
+		BindOnAllInterfaces bool `yaml:"bindOnAllInterfaces"`
```
mindaugasrukas: Instead, I would suggest adding a parameter `BindOnIP string`.
And then parse it with `net.ParseIP(BindOnIP)`.
And yes we need to leave `BindOnLocalHost` for backwards compatibility.

artms: True, waiting for cadence team to suggest what they think is best for them

venkat1109: artms - I prefer BindOnIP to BindOnAllInterfaces. It gives maximum flexibility i.e. we no longer need special options for BindOnAll or BindOnLocalHost

artms: Fair enough, what about backwards compatibility, should I fully eradicated BindOnLocalHost or make something like:
1. if `BindOnLocalhost` is set and BindOnIP is set at the same time - "crash" with error that values are mutually exclusive
2. if `BindOnIP` is not set - old behavior where we try listening on external machine IP
3. `BindOnIP` is set - do as setting specifies

venkat1109: We have a few internal development configs that depend on BindOnLocaHost. So, lets leave it aside for backwards compatibility for now. I created a tracking task to get to this later
https://github.com/uber/cadence/issues/1300

To sumarize, lets go with the following behavior
(1) If both BindOnIP and BindOnLocalHost are set - panic
(2) If BindOnLocalHost or BindOnIP is set, use it
(3) Else use default behavior

venkat1109:
@artms -  Just a ping - Where are you with this PR ? Do you still need the changes in this PR ?

artms:
> @artms - Just a ping - Where are you with this PR ? Do you still need the changes in this PR ?
Sorry, last week was very busy didn't have time to sit and do it, I'm keeping this change in mind :)

artms:
@venkat1109 - change is ready

artms:
@venkat1109 - gentle nudge.

common/service/config/rpc.go
```diff
@@ -86,9 +86,21 @@ func (d *RPCFactory) CreateDispatcherForOutbound(
 }
 
 func (d *RPCFactory) getListenIP() net.IP {
+	if d.config.BindOnLocalHost && len(d.config.BindOnIP) > 0 {
+		d.logger.Fatalf("ListenIP failed, bindOnLocalHost and bindOnIP are mutually exclussive")
```
venkat1109: typo - s/exclussive/exclusive

docker/config_template.yaml
```diff
@@ -29,6 +29,7 @@ services:
     rpc:
       port: 7933
       bindOnLocalHost: ${BIND_ON_LOCALHOST}
+      bindOnIP: ${BIND_ON_IP}
```
venkat1109: I believe this docker startup script also needs to be modified to support BIND_ON_IP correctly
https://github.com/uber/cadence/blob/master/docker/start.sh#L56
It defaults to the address returned by `hostname --ip-address` when BIND_ON_LOCALHOST is false - which may be wrong when BIND_ON_IP is explicitly set.

artms: When `BIND_ON_IP` is set (assuming != 127.0.0.1) - `hostname --ip-address` might be a more correct approach, you can bind on `0.0.0.0` but connecting to 0.0.0.0 is not very defined...
I propose changing docker `start.sh`:
```
    export HOST_IP=`hostname --ip-address`

    if [ "$BIND_ON_LOCALHOST" == true ] || [ "$BIND_ON_IP" == "127.0.0.1" ]; then
            export BIND_ON_IP="127.0.0.1"
            export HOST_IP="127.0.0.1"
    fi

    # deprecating this env key
    export BIND_ON_LOCALHOST=false
```

Essentially I deprecate `BIND_ON_LOCALHOST` and replace it with BIND_ON_IP, leaving HOST_IP which is used by cassandra and ringpop

venkat1109: makes sense to me.

venkat1109:
Closing this PR in favor of https://github.com/uber/cadence/pull/1471

PR #1288, opened 2018-11-28
 Move history events out of mutableState to eventsCache

samarabbas:
Storing history events within mutableState limits scalability of the
system.  So moving all history events out of mutableState to
separate eventsCache.  This makes the mutable state size to be quite
small and allows us to store very large number of activities.
EventsCache is a LRU cache which stores the history event.  Any
processing logic which needs the history event like
RecordActivityStarted can get it from the cache without reading
it from the database.  Incase event is not found in cache it
uses the HistoryManager to load it from database and return it
to caller.

### DIFF
--- a/service/history/eventsCache.go
+++ b/service/history/eventsCache.go
@@ -145,8 +145,8 @@ func (e *eventsCacheImpl) getHistoryEventFromStore(domainID, workflowID, runID s
 			return nil, err
 		}
 
-		if len(response.History) > 0 {
-			historyEvent = response.History[0]
+		if len(response.HistoryEvents) > 0 {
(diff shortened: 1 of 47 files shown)


service/history/eventsCache.go
```diff
@@ -0,0 +1,182 @@
...
+	// find history event from batch and return back single event to caller
+	var historyEvent *shared.HistoryEvent
+	for _, e := range historyEvents {
+		if e.GetEventId() == eventID {
+			historyEvent = e
+			break
+		}
+	}
+
+	return historyEvent, nil
```
yiminc: should it return error when it cannot find the event? I would just return in the for loop when event was found, and return error here when event cannot be found.

PR #1291, opened 2018-11-29
Refactor client lib separating client retrieval logic

wxing1292:


wxing1292:
we need to implement the DispatcherProvider interface and pass it in as an parameter, like this:
https://github.com/uber/cadence/pull/1291/files#diff-4c51d95b26718af4a2ac3bec33e54e49R126

client/clientBean.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/uber/cadence/common/cluster"
+)
+
+const (
+	ipPortRegex = `\b(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9]):[1-9]\d*\b`
+)
+
+type (
+	// Bean in an collection of clients
+	Bean interface {
```
yiminc: Would name it ClientFactory or something similar.

wxing1292: the name client factory is already taken by an actual factory which generate new client.
cannot really name it "Client*" since the package's name is client, or it would be "client.ClientFoobar" when being used.

will address this issue in the next PR when applying the re-replication when evens are lost by kafka since this is a small change

PR #1299, opened 2018-12-02
Bugfix: leaking goroutine

wxing1292:
* Timer gate is not properly closed
* Task processing retry policy need to check shutdown chan

### DIFF
--- a/service/history/queueProcessor.go
+++ b/service/history/queueProcessor.go
@@ -279,7 +279,14 @@ func (p *queueProcessorBase) processTaskAndAck(notificationChan <-chan struct{},
 		scope, err = p.processTaskOnce(notificationChan, task, logger)
 		return p.handleTaskError(scope, startTime, notificationChan, err, logger)
 	}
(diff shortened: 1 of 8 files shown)


service/history/queueProcessor.go
```diff
@@ -279,7 +279,14 @@ func (p *queueProcessorBase) processTaskAndAck(notificationChan <-chan struct{},
 		scope, err = p.processTaskOnce(notificationChan, task, logger)
 		return p.handleTaskError(scope, startTime, notificationChan, err, logger)
 	}
-	retryCondition := func(err error) bool { return true }
+	retryCondition := func(err error) bool {
+		select {
+		case <-p.shutdownCh:
```
samarabbas: I thought we retry for a maximum of 10 seconds.  How can this cause go-routine leak? The pump retrying tasks should check for shutdownCh and bail out of the pump.

wxing1292: the retry can possibly cause blocking on standby task, so, should check whether to continue to do an retry

PR #1303, opened 2018-12-04
Change CLI to exit on any errors with exit code 1

yux0:
Change CLI tools to throw exit code 1 on errors.
Issue: https://github.com/uber/cadence/issues/706

### DIFF
--- a/tools/cli/adminCommands.go
+++ b/tools/cli/adminCommands.go
@@ -43,7 +43,7 @@ const (
 func getAdminServiceClient(c *cli.Context) adminserviceclient.Interface {
 	client, err := cBuilder.BuildAdminServiceClient(c)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tools/cli/adminCommands.go
```diff
@@ -43,7 +43,7 @@ const (
 func getAdminServiceClient(c *cli.Context) adminserviceclient.Interface {
 	client, err := cBuilder.BuildAdminServiceClient(c)
 	if err != nil {
-		ExitIfError(err)
```
longquanzheng: Are we deprecating this method? ExitIfError
Then probably we should delete it if not being used anywhere.

yux0:
Github identity problem. I am closing this PR and re-send a new one.

PR #1304, opened 2018-12-04
Change CLI to exit on any errors with exit code 1

yux0:
Change CLI tools to throw exit code 1 on errors.
Issue: https://github.com/uber/cadence/issues/706

### DIFF
--- a/tools/cli/commands.go
+++ b/tools/cli/commands.go
@@ -304,7 +304,7 @@ func RegisterDomain(c *cli.Context) {
 			ErrorAndExit(fmt.Sprintf("Domain %s already registered.\n", domain), err)
 		}
 	} else {
-		fmt.Printf("Domain %s succeesfully registered.\n", domain)
+		fmt.Printf("Domain %s successfully registered.\n", domain)
 	}
 }
 
@@ -409,7 +409,7 @@ func UpdateDomain(c *cli.Context) {
 			ErrorAndExit(fmt.Sprintf("Domain %s does not exist.\n", domain), err)
 		}
 	} else {
-		fmt.Printf("Domain %s succeesfully updated.\n", domain)
+		fmt.Printf("Domain %s successfully updated.\n", domain)
 	}
 }
 
@@ -718,7 +718,7 @@ func TerminateWorkflow(c *cli.Context) {
 	if err != nil {
 		ErrorAndExit("Terminate workflow failed.", err)
 	} else {
-		fmt.Println("Terminate workflow succeed.")
+		fmt.Println("Terminate workflow succeeded.")
 	}
 }
 
@@ -736,7 +736,7 @@ func CancelWorkflow(c *cli.Context) {
 	if err != nil {
 		ErrorAndExit("Cancel workflow failed.", err)
 	} else {
-		fmt.Println("Cancel workflow succeed.")
+		fmt.Println("Cancel workflow succeeded.")
 	}
 }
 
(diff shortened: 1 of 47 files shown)


tools/cli/commands.go
```diff
@@ -770,7 +765,7 @@ func SignalWorkflow(c *cli.Context) {
 	})
 
 	if err != nil {
-		ErrorAndExit("Signal workflow failed", err)
+		ErrorAndExit("Signal workflow failed.", err)
 	} else {
 		fmt.Println("Signal workflow succeed.")
```
dmetzgar: Can you put succeeded instead of succeed for these cases?

yux0: Updated

tools/cli/commands.go
```diff
@@ -205,29 +205,27 @@ var workflowClosedStatusMap = map[string]s.WorkflowExecutionCloseStatus{
...
-}
-
-// ExitIfError exit while err is not nil and print the calling stack also
-func ExitIfError(err error) {
-	const stacksEnv = `CADENCE_CLI_SHOW_STACKS`
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err.Error())
-		if os.Getenv(stacksEnv) != `` {
-			debug.PrintStack()
+		if os.Getenv(showErrorStackEnv) != `` {
```
vancexu: I am confused, how did you print stack trace?

tools/cli/commands.go
```diff
@@ -369,7 +364,7 @@ func UpdateDomain(c *cli.Context) {
 		if c.IsSet(FlagEmitMetric) {
 			emitMetric, err = strconv.ParseBool(c.String(FlagEmitMetric))
 			if err != nil {
-				ErrorAndExit("Update Domain failed", err)
+				ErrorAndExit(FlagEmitMetric+"format is invalid.", err)
```
vancexu: missing space

tools/cli/commands.go
```diff
@@ -886,7 +881,7 @@ func DescribeWorkflow(c *cli.Context) {
 // DescribeWorkflowWithID show information about the specified workflow execution
 func DescribeWorkflowWithID(c *cli.Context) {
 	if !c.Args().Present() {
-		ExitIfError(errors.New("workflow_id is required"))
+		ErrorAndExit("Parameter workflow_id is required.", nil)
```
vancexu: Better to consist with Argument instead of Parameter

tools/cli/commands.go
```diff
@@ -1155,7 +1149,7 @@ func ObserveHistory(c *cli.Context) {
 // ObserveHistoryWithID show the process of running workflow
 func ObserveHistoryWithID(c *cli.Context) {
 	if !c.Args().Present() {
-		ExitIfError(errors.New("workflow_id is required"))
+		ErrorAndExit("workflow_id is required.", nil)
```
vancexu: Would be better to start all Error msg with Upper case, or add "Argument workflow_id ..."

tools/cli/commands.go
yux0: Updated. Will send out a new PR about the change.

tools/cli/commands.go
yux0: Updated. Will send out a new PR about the change.

tools/cli/commands.go
yux0: Updated. Will send out a new PR about the change.

tools/cli/commands.go
yux0: Faultily thought the error should has stack error. Adding printStack in the new PR.

PR #1309, opened 2018-12-05
Size limit: blob size limit

yiminc:


common/metrics/client.go
```diff
@@ -61,15 +59,15 @@ func NewClient(scope tally.Scope, serviceIdx ServiceIdx) Client {
 		scopeTags := map[string]string{
 			OperationTagName: def.operation,
 		}
-		common.MergeDictoRight(def.tags, scopeTags)
+		mergeDictoRight(def.tags, scopeTags)
```
yiminc: This change is to break circle dependency because common.CheckEventBlobSizeLimit() in common package depends metrics package.

common/metrics/client.go
```diff
@@ -129,3 +127,9 @@ func getMetricDefs(serviceIdx ServiceIdx) map[int]metricDefinition {
 
 	return defs
 }
+
+func mergeDictoRight(src map[string]string, dest map[string]string) {
```
wxing1292: maybe name it mergeMapToRight? at least mergeDicToRight

yiminc: will do

service/history/historyEngine.go
```diff
@@ -1151,9 +1156,12 @@ Update_History_Loop:
 					continue Process_Decision_Loop
 				}
 				attributes := d.CompleteWorkflowExecutionDecisionAttributes
-				if err = validateCompleteWorkflowExecutionAttributes(attributes); err != nil {
+				if err = validateCompleteWorkflowExecutionAttributes(attributes, sizeLimitWarn, sizeLimitError,
+					domainID, workflowExecution.GetWorkflowId(), workflowExecution.GetRunId(),
+					e.metricsClient, metrics.HistoryRespondDecisionTaskCompletedScope, e.logger); err != nil {
```
samarabbas: I think it is better to fail the workflow instead of failing decision.  Otherwise client would just keep on retrying with the same payload.  Once we have restart user can always restart failed workflows automatically.

service/history/historyEngine.go
```diff
@@ -1178,10 +1186,19 @@ Update_History_Loop:
 				}
 
 				failedAttributes := d.FailWorkflowExecutionDecisionAttributes
-				if err = validateFailWorkflowExecutionAttributes(failedAttributes); err != nil {
-					failDecision = true
-					failCause = workflow.DecisionTaskFailedCauseBadFailWorkflowExecutionAttributes
-					break Process_Decision_Loop
+				if err = validateFailWorkflowExecutionAttributes(failedAttributes, sizeLimitWarn, sizeLimitError,
```
samarabbas: Same as size limit error on complete.  Fail the workflow.

service/history/historyEngine.go
yiminc: only for this decision DecisionTypeCompleteWorkflowExecution? Or for all other decisions?

service/history/historyEngine.go
```diff
@@ -1012,6 +1010,39 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(ctx context.Context,
...
+	domainID       string
+	workflowID     string
+	runID          string
+	metricsClient  metrics.Client
+	logger         bark.Logger
+	msBuilder      mutableState
+	completedID    int64
+}
+
+func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(blob []byte) (bool, error) {
```
samarabbas: Can you also pass in a reason along with blob, so instead of having a generic 'FailureReasonDecisionBlobSizeExceedsLimit' we can have something finer grained like you had in your previous diff.

service/history/historyEngine.go
```diff
@@ -1012,6 +1010,39 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(ctx context.Context,
...
+
+	err := common.CheckEventBlobSizeLimit(len(blob), c.sizeLimitWarn, c.sizeLimitError,
+		c.domainID, c.workflowID, c.runID, c.metricsClient, metrics.HistoryRespondDecisionTaskCompletedScope, c.logger)
+
+	if err == nil {
+		return false, nil
+	}
+
+	attributes := &workflow.FailWorkflowExecutionDecisionAttributes{
```
samarabbas: Should we also disable workflow retries when it fails with this error?

PR #1311, opened 2018-12-05
Add archival interface and plumb into sysworkflow

andrewjdawson2016:
- Adds archival interface
- Adds NOP impl
- Does plumbing to call archival interface


common/archival/client.go
```diff
@@ -0,0 +1,80 @@
...
+	RunID        string
+	WorkflowType string
+	Status       string
+	CloseTime    time.Time
+	History      *shared.History
+}
+
+// Client is used to store and retrieve blobs
+type Client interface {
+	Archive(ctx context.Context, request *ArchiveRequest) error
```
venkat1109: naming: how about we call this ArchiveWorkflow() or ArchiveWorkflowHistory() - just so that - in future if we archive something else (like domain metadata change history) - the APIs become less confusing ?

common/archival/client.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+	"errors"
+	"go.uber.org/cadence/.gen/go/shared"
+	"time"
+)
+
+// ArchiveRequest is request for Archive
+type ArchiveRequest struct {
```
venkat1109: [optional] archival.ArchiveRequest stutters. Consider renaming this PutRequest. So that, the api then becomes 

archival.PutWorkflow(request *PutWorkflowRequest)

Its also consistent with your other APIs - Get / List

common/archival/client.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"context"
+	"errors"
+	"github.com/uber/cadence/.gen/go/shared"
+	"time"
+)
+
+// PutRequest is request for Archive
+type PutRequest struct {
```
venkat1109: PutWorkflowRequest

PR #1312, opened 2018-12-05
CLI improvement on ErrorAndExit

yux0:
Follow up on the previous PR: https://github.com/uber/cadence/pull/1304

1. Words/format updates.
2. print stack trace.


tools/cli/commands.go
```diff
@@ -217,10 +218,12 @@ func SetBuilder(builder WorkflowClientBuilderInterface) {
 // ErrorAndExit print easy to understand error msg first then error detail in a new line
 func ErrorAndExit(msg string, err error) {
 	if err != nil {
+		fmt.Printf("%s %s\n%s %+v\n", colorRed("Error:"), msg, colorMagenta("Error Details:"), err)
 		if os.Getenv(showErrorStackEnv) != `` {
-			fmt.Printf("%s %s\n%s %+v\n", colorRed("Error:"), msg, colorMagenta("Error Details:"), err)
+			fmt.Print(colorMagenta("Stack trace:"))
```
vancexu: How about we just use default for stack trace? or use another easily differenciate color. I just thought it might be better to let it easier to sperate stack trace with err details.

vancexu:
BTW thanks for fixing this, and good job!

tools/cli/commands.go
yux0: I think we have enough color for text. I will just use the default style in term of the stack trace is for debugging.

PR #1315, opened 2018-12-05
 Add page size limit for GetWorkflowExecutionHistory

vancexu:
Related issue https://github.com/uber/cadence/issues/1302


service/frontend/workflowHandler.go
```diff
@@ -1495,6 +1495,18 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
 	}
 
+	// force limit page size if exceed
+	if getRequest.GetMaximumPageSize() > common.GetHistoryMaxPageSize {
+		wh.GetLogger().WithFields(bark.Fields{
+			logging.TagWorkflowExecutionID: getRequest.Execution.GetWorkflowId(),
+			logging.TagWorkflowRunID:       getRequest.Execution.GetRunId(),
+			logging.TagDomainID:            domainID,
+			logging.TagSize:                getRequest.GetMaximumPageSize(),
+		}).Warn(fmt.Sprintf("GetHistory page size is larger than threshold %d", common.GetHistoryMaxPageSize))
```
yiminc: it is better to use static text as log message so it could be easy searchable.

PR #1321, opened 2018-12-07
Refactor cli and add server side admin clients

andrewjdawson2016:
- Refactor builder 
- Include methods to get server side clients
- Remove client side admin client
- Consume server admin client in admin cli commands

### DIFF
--- a/tools/cli/factory.go
+++ b/tools/cli/factory.go
@@ -21,112 +21,96 @@
 package cli
 
 import (
-	"errors"
-
+	serverAdmin "github.com/uber/cadence/.gen/go/admin/adminserviceclient"
+	serverFrontend "github.com/uber/cadence/.gen/go/cadence/workflowserviceclient"
 	"github.com/urfave/cli"
-	"go.uber.org/cadence/.gen/go/admin/adminserviceclient"
-	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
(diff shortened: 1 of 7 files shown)


tools/cli/factory.go
```diff
@@ -21,112 +21,87 @@
...
 
-	return adminserviceclient.New(b.dispatcher.ClientConfig(_cadenceFrontendService)), nil
+// ServerAdminClient builds an admin client (based on server side thrift interface)
+func (b *clientFactory) ServerAdminClient(c *cli.Context) serverAdmin.Interface {
+	b.ensureDispatcher(c)
+	return serverAdmin.New(b.dispatcher.ClientConfig(cadenceFrontendService))
 }
 
-func (b *WorkflowClientBuilder) build() error {
+func (b *clientFactory) ensureDispatcher(c *cli.Context) {
```
vancexu: I think we still need to return err and handle in caller. Otherwise these Client may be created with dispatch not running

tools/cli/factory.go
```diff
@@ -21,112 +21,87 @@
 package cli
 
 import (
-	"errors"
-
+	serverAdmin "github.com/uber/cadence/.gen/go/admin/adminserviceclient"
```
longquanzheng: Any reason to change it from "go.uber.org/cadence/.gen/go/admin/adminserviceclient" to "github.com/uber/cadence/.gen/go/admin/adminserviceclient"? Looks like the cadence-tools will break and I don't see an easy way to fix.

PR #1322, opened 2018-12-08
New functionality for cross DC handling case if Kafka is not lossless

wxing1292:
2 New APIs:
* GetWorkflowExecutionRawHistory return list of serialized history batch events
* ReplicateRawEvents apply serialized history batch events

Handling Kafka being not lossless:
* History re-replicator handling Kafka being not lossless situation

Future work:
1. Expose getting raw history directly from DB layer: #1323
2. Serialize Request To Getting Missing History In Replicator: #1324

### DIFF
--- a/service/history/timerQueueActiveProcessor.go
+++ b/service/history/timerQueueActiveProcessor.go
@@ -281,7 +281,6 @@ Update_History_Loop:
 					// Update the task ID tracking the corresponding timer task.
 					ti.TaskID = TimerTaskStatusCreated
 					msBuilder.UpdateUserTimer(ti.TimerID, ti)
-					defer t.notifyNewTimers(timerTasks)
 				}
 
 				// Done!
@@ -457,7 +456,6 @@ Update_History_Loop:
 				return err
 			}
 
-			t.notifyNewTimers(timerTasks)
 			return nil
 		}
 
@@ -754,12 +752,13 @@ Update_History_Loop:
 
 		timersToNotify := append(timerTasks, msBuilder.GetContinueAsNew().TimerTasks...)
 		err = context.continueAsNewWorkflowExecution(nil, continueAsNewBuilder, transferTasks, timerTasks, transactionID)
-		t.notifyNewTimers(timersToNotify)
 
 		if err != nil {
 			if err == ErrConflict {
 				continue Update_History_Loop
 			}
+		} else {
+			t.historyService.timerProcessor.NotifyNewTimers(t.currentClusterName, t.shard.GetCurrentTime(t.currentClusterName), timersToNotify)
 		}
 		return err
 	}
@@ -789,7 +788,7 @@ func (t *timerQueueActiveProcessorImpl) updateWorkflowExecution(
 		tBuilder := t.historyService.getTimerBuilder(context.getExecution())
 		tranT, timerT, err := t.historyService.getDeleteWorkflowTasks(executionInfo.DomainID, executionInfo.WorkflowID, tBuilder)
 		if err != nil {
-			return nil
+			return err
 		}
 		transferTasks = append(transferTasks, tranT)
 		timerTasks = append(timerTasks, timerT)
@@ -811,6 +810,6 @@ func (t *timerQueueActiveProcessorImpl) updateWorkflowExecution(
 		return err
 	}
 
-	t.notifyNewTimers(timerTasks)
-	return err
+	t.historyService.timerProcessor.NotifyNewTimers(t.currentClusterName, t.shard.GetCurrentTime(t.currentClusterName), timerTasks)
+	return nil
 }
--- a/service/history/timerQueueProcessor_test.go
+++ b/service/history/timerQueueProcessor_test.go
@@ -52,6 +52,7 @@ type (
 
 		mockMetadataMgr     *mocks.MetadataManager
 		mockVisibilityMgr   *mocks.VisibilityManager
+		mockMatchingClient  *mocks.MatchingClient
 		mockClusterMetadata *mocks.ClusterMetadata
 	}
 )
@@ -80,6 +81,7 @@ func (s *timerQueueProcessorSuite) SetupTest() {
 	s.ShardContext.config.TransferProcessorUpdateAckInterval = dynamicconfig.GetDurationPropertyFn(100 * time.Millisecond)
 	s.ShardContext.config.TimerProcessorUpdateAckInterval = dynamicconfig.GetDurationPropertyFn(100 * time.Millisecond)
 
+	s.mockMatchingClient = &mocks.MatchingClient{}
 	s.mockClusterMetadata = &mocks.ClusterMetadata{}
 	s.mockClusterMetadata.On("GetCurrentClusterName").Return(cluster.TestCurrentClusterName)
 
@@ -101,6 +103,7 @@ func (s *timerQueueProcessorSuite) SetupTest() {
 	s.engineImpl.txProcessor = newTransferQueueProcessor(
 		s.ShardContext, s.engineImpl, s.mockVisibilityMgr, &mocks.KafkaProducer{}, &mocks.MatchingClient{}, &mocks.HistoryClient{}, s.logger,
 	)
+	s.engineImpl.timerProcessor = newTimerQueueProcessor(s.ShardContext, s.engineImpl, s.mockMatchingClient, s.logger)
 }
 
 func (s *timerQueueProcessorSuite) TearDownTest() {
@@ -280,7 +283,7 @@ func (s *timerQueueProcessorSuite) TestSingleTimerTask() {
 	s.NotEmpty(timerInfo, "Expected non empty timers list")
 	s.Equal(1, len(timerInfo))
 
-	processor := newTimerQueueProcessor(s.ShardContext, s.engineImpl, s.matchingClient, s.logger).(*timerQueueProcessorImpl)
+	processor := s.engineImpl.timerProcessor.(*timerQueueProcessorImpl)
 	processor.Start()
 	processor.NotifyNewTimers(cluster.TestCurrentClusterName, s.ShardContext.GetCurrentTime(cluster.TestCurrentClusterName), tt)
 
@@ -303,7 +306,7 @@ func (s *timerQueueProcessorSuite) TestManyTimerTasks() {
 	s.NotEmpty(timerInfo, "Expected non empty timers list")
 	s.Equal(1, len(timerInfo))
 
-	processor := newTimerQueueProcessor(s.ShardContext, s.engineImpl, s.matchingClient, s.logger).(*timerQueueProcessorImpl)
+	processor := s.engineImpl.timerProcessor.(*timerQueueProcessorImpl)
 	processor.Start()
 	processor.NotifyNewTimers(cluster.TestCurrentClusterName, s.ShardContext.GetCurrentTime(cluster.TestCurrentClusterName), tt)
 
@@ -326,7 +329,7 @@ func (s *timerQueueProcessorSuite) TestTimerTaskAfterProcessorStart() {
 	s.NoError(err, "No error expected.")
 	s.Empty(timerInfo, "Expected empty timers list")
 
-	processor := newTimerQueueProcessor(s.ShardContext, s.engineImpl, s.matchingClient, s.logger).(*timerQueueProcessorImpl)
+	processor := s.engineImpl.timerProcessor.(*timerQueueProcessorImpl)
 	processor.Start()
 
 	tBuilder := newTimerBuilder(s.ShardContext.GetConfig(), s.logger, common.NewRealTimeSource())
@@ -401,7 +404,7 @@ func (s *timerQueueProcessorSuite) TestTimerActivityTaskScheduleToStart_WithOutS
 	s.createExecutionWithTimers(domainID, workflowExecution, taskList, "identity", []int32{})
 
 	// TimeoutTypeScheduleToStart - Without Start
-	processor := newTimerQueueProcessor(s.ShardContext, s.engineImpl, s.matchingClient, s.logger).(*timerQueueProcessorImpl)
+	processor := s.engineImpl.timerProcessor.(*timerQueueProcessorImpl)
 	processor.Start()
 
 	state, err := s.GetWorkflowExecutionInfo(domainID, workflowExecution)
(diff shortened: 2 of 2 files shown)


service/worker/replicator/historyRereplicator.go
```diff
@@ -0,0 +1,438 @@
...
+			return "", err
+		}
+
+		branchToken = response.BranchToken
+		eventStoreVersion = response.GetEventStoreVersion()
+		replicationInfo = h.replicationInfoFromPublic(response.ReplicationInfo)
+		token = response.NextPageToken
+
+		for _, externalBatch := range response.HistoryBatches {
+			err := h.sendReplicationRawRequest(request)
```
yiminc: the first request is nil, yes it is handled in the function, but it is very confusing to read this code. is it possible to re-organize the code so it prepare the request first and then send it after, or at least adding some more comments.

service/worker/replicator/historyRereplicator.go
```diff
@@ -0,0 +1,438 @@
...
+			if err != nil {
+				return "", err
+			}
+			internalBatch, err := h.dataBlobFromPublic(externalBatch)
+			if err != nil {
+				return "", err
+			}
+			request = h.createReplicationRawRequest(domainID, workflowID, runID, internalBatch, eventStoreVersion, replicationInfo)
```
yiminc: would rename this request to nextRequest

service/worker/replicator/historyRereplicator.go
```diff
@@ -0,0 +1,438 @@
...
+			logging.TagDomainID:            request.GetDomainUUID(),
+			logging.TagWorkflowExecutionID: request.WorkflowExecution.GetWorkflowId(),
+			logging.TagWorkflowRunID:       request.WorkflowExecution.GetRunId(),
+			logging.TagErr:                 err,
+		}).Error("error sending history")
+	}
+	return err
+}
+
+func (h *HistoryRereplicatorImpl) getHistory(domainID string, workflowID string, runID string, branchToken []byte,
```
yiminc: maybe rename this to getHistoryFromRemote or FromActiveCluster.

service/worker/replicator/historyRereplicator.go
```diff
@@ -0,0 +1,438 @@
...
+	// this means that beginingRunID to endingRunID does not have a continue as new relation ship
+	// such as beginingRunID -> runID1, then runID2 (no continued as new), then runID3 -> endingRunID
+	// need to work backwords using endingRunID,
+	// moreover, in the above case, runID2 cannot be replicated since no one points to it
+
+	// runIDs to be use to resend history, in reverse order
+	runIDs := []string{endingRunID}
+	runID = endingRunID
+	for len(runID) != 0 {
+		runID, err = h.getPrevRunID(domainID, workflowID, runID)
```
yiminc: need to handle the edge case where the initial runID is already deleted from active cluster.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -254,6 +258,20 @@ struct TaskList {
   20: optional TaskListKind kind
 }
 
+enum EncodingType {
```
samarabbas: Considering we are not exposing GetWorkflowExecutionRawHistory api to clients in this PR, I would recommend to not expose this on public thrift interfaces.  Especially how we expose headers like encoding back to clients needs some discussion.

service/history/historyReplicator.go
```diff
@@ -1085,8 +1138,36 @@ func (r *historyReplicator) notify(clusterName string, now time.Time, transferTa
 	r.historyEngine.timerProcessor.NotifyNewTimers(clusterName, now, timerTasks)
 }
 
+func (r *historyReplicator) deserializeBlob(blob *workflow.DataBlob) ([]*workflow.HistoryEvent, error) {
+
+	if blob.GetEncodingType() != workflow.EncodingTypeThriftRW {
```
samarabbas: This looks like a layering violation.  We should just use the same history serializers used by persistence layer for deserializing the payloads.  We should be able to deserialize all supported payloads like json and thrift.

service/worker/replicator/replicator.go
```diff
@@ -55,21 +63,30 @@ type (
...
 
+const (
+	replicationTimeout = 30 * time.Second
+)
+
 // NewReplicator creates a new replicator for processing replication tasks
 func NewReplicator(clusterMetadata cluster.Metadata, metadataManagerV2 persistence.MetadataManager,
-	historyClient history.Client, config *Config, client messaging.Client, logger bark.Logger,
-	metricsClient metrics.Client) *Replicator {
+	domainCache cache.DomainCache, clientBean client.Bean, historyClient history.Client,
```
samarabbas: looks like client.Bean interface already supports creating all clients including client for remote cadence cluster.  Can we just use this to initialize other clients like historyClient instead of explicitly passing in.

service/worker/replicator/processor.go
```diff
@@ -437,6 +435,22 @@ RetryLoop:
 		}
 		break RetryLoop
 	}
+
+	if retryErr, ok := err.(*shared.RetryTaskError); ok {
```
samarabbas: Let's emit a log here with all relevant tags.  This could potentially mean Kafka lost messages.  We should emit a metric on all re-replicate calls initiated by the cluster so we can set up alarms if this starts happening very frequently.

service/worker/replicator/processor.go
```diff
@@ -437,6 +435,22 @@ RetryLoop:
...
+		beginRunID := retryErr.GetRunId()
+		beginEventID := retryErr.GetNextEventId()
+		endRunID := attr.GetRunId()
+		endEventID := attr.GetFirstEventId()
+
+		// history service will return RetryTaskError, we can use it as an hint
+		errResend := p.historyRereplicator.SendMultiWorkflowHistory(
```
samarabbas: This feature needs to be rolled out to all clusters before enabling it.  Here are the 2 options I see:
1) If target cluster is running old version I expect something like handler not found error.  We should disable reties on this error and proceed with old style processing.
2) Separate config knob to enable/disable this feature.  Config knob is enabled once feature is rolled out to all clusters.  Config knob will be useful in other situations too, like disabling this feature due to unforeseen issue.

idl/github.com/uber/cadence/replicator.thrift
```diff
@@ -53,7 +53,7 @@ struct HistoryTaskAttributes {
   40: optional i64 (js.type = "Long") firstEventId
   50: optional i64 (js.type = "Long") nextEventId
   60: optional i64 (js.type = "Long") version
-  70: optional map<string, history.ReplicationInfo> replicationInfo
```
samarabbas: Is this a backward compatible change?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -254,6 +258,20 @@ struct TaskList {
...
+  ThriftRW,
+}
+
+struct DataBlob {
+  10: optional EncodingType EncodingType
+	20: optional binary Data
```
samarabbas: fix indenting

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1036,6 +1054,24 @@ struct GetWorkflowExecutionHistoryResponse {
   20: optional binary nextPageToken
 }
 
+struct GetWorkflowExecutionRawHistoryRequest {
+  10: optional string domain
+  20: optional WorkflowExecution execution
+  30: optional binary branchToken
```
samarabbas: I don't think these branch tokens are transferable across clusters.  How does the remote cluster gets branchToken for the very first call when requesting history for an execution?  Is the expectation here is to call GetMutableState first before calling GetWorkflowExecutionRawHistory?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1036,6 +1054,24 @@ struct GetWorkflowExecutionHistoryResponse {
...
+  40: optional i64 (js.type = "Long") firstEventId
+  50: optional i64 (js.type = "Long") nextEventId
+  60: optional i32 maximumPageSize
+  70: optional binary nextPageToken
+}
+
+struct GetWorkflowExecutionRawHistoryResponse {
+  10: optional binary branchToken
```
samarabbas: Can we just return single token back?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1036,6 +1054,24 @@ struct GetWorkflowExecutionHistoryResponse {
...
+  20: optional binary nextPageToken
+  30: optional list<DataBlob> historyBatches
+  40: optional map<string, ReplicationInfo> replicationInfo
```
samarabbas: I don't like returning part of mutable state along with this call.  ReplicationInfo and EventStoreVersion are part of mutable state and should only be exposed through GetMutableState API.

service/frontend/workflowHandler.go
```diff
@@ -1620,6 +1623,166 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+			Execution:           execution,
+			ExpectedNextEventId: common.Int64Ptr(common.FirstEventID), // common.FirstEventID means no long poll
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		// branch token is provided as part of request, makeing sure that no branching has happened yet
+		// otherwise, caller need to do a reload of history from the begining
+		if request.BranchToken != nil && bytes.Compare(request.BranchToken, response.BranchToken) != 0 {
```
samarabbas: I don't think this is necessary.  We always create a new run when branching workflow executions.  So we don't expect multiple branches with the same run id anymore.

service/frontend/workflowHandler.go
```diff
@@ -1620,6 +1623,166 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+			HistoryBatches:    []*gen.DataBlob{},
+			ReplicationInfo:   token.ReplicationInfo,
+			EventStoreVersion: common.Int32Ptr(token.EventStoreVersion),
+			NextPageToken:     nil, // no further pagination
+		}, nil
+	}
+
+	// TODO need to deal with transient decision if to be used by client getting history
+	var historyBatches []*gen.History
+	_, historyBatches, token.PersistenceToken, _, err = historyService.PaginateHistory(
```
samarabbas: This defeats the purpose of RawHistory API.  If we are already deserializing history before sending it back to caller then I don't see a point in exposing this GetRawHistory API.

service/frontend/workflowHandler.go
wxing1292: this is true except that conflict resolution, when that is enabled, a new branch will be created and run ID remains the same.

anyway, removing the branch token for now, since conflict resolution & branching is not implemented.

service/frontend/workflowHandler.go
wxing1292: the refactoring of the persistent layer providing raw & valid data is an separate task, and there is already an issue for it: https://github.com/uber/cadence/issues/1323

for your question, this does not defeat the purpose of sending raw history, since worker in most cases, do not need to do serialization & deserialization.

moreover, our persistent layer does not guarantee returning the correct raw data now, since there is fair amount of logic filtering out the stale data, plus, event store v1 and v2 do not have any unit test.

idl/github.com/uber/cadence/shared.thrift
wxing1292: GetMutableState API is not exposed to public side, while the new API providing history events for re-sending message when Kafka fail to deliver message should also include necessary information.

anyway, the new API is moved to admin handler.

idl/github.com/uber/cadence/shared.thrift
wxing1292: the purpose of branch token is to guarantee that the returned history will not come from more than one branch, when conflict resolution & history branching is done.

since the feature mentioned above is not implemented, this branchToken will be removed for now.

idl/github.com/uber/cadence/shared.thrift
wxing1292: done

idl/github.com/uber/cadence/replicator.thrift
wxing1292: only the name is changed, as long as thrift's concrete implementation remains the same, we should be fine

service/worker/replicator/processor.go
wxing1292: we can use the new get raw history API metrics

service/history/historyReplicator.go
wxing1292: currently the new API only use ThriftRW.

service/worker/replicator/historyRereplicator.go
wxing1292: if the beginning run ID is already deleted from remote, then there is nothing this logic can do, since current cluster needs those events. manual operation must be done.

the only edge case this logic can handle is when tracing back using the ending run ID, and encounter EntityNotExistsError. this means that a workflow is not replicated to the currently cluster and the same workflow is also missing in remote.

Gopkg.toml
```diff
@@ -108,7 +108,8 @@ ignored = ["github.com/uber/cadence/.gen"]
 
 [[constraint]]
   name = "go.uber.org/cadence"
-  version = "0.7.6"
+  # version = "0.7.6"
+  branch = "idl-update"
```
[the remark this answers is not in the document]
wxing1292: done

service/worker/replicator/historyRereplicator.go
```diff
@@ -0,0 +1,438 @@
...
+}
+
+// SendMultiWorkflowHistory sends multiple run IDs's history events to remote
+func (h *HistoryRereplicatorImpl) SendMultiWorkflowHistory(domainID string, workflowID string,
+	beginingRunID string, beginingFirstEventID int64, endingRunID string, endingNextEventID int64) (err error) {
+	// NOTE: begining run ID and ending run ID can be different
+	// the logic will try to grab events from begining run ID, first event ID to ending run ID, ending next event ID
+
+	// beginingRunID can be empty, if there is no workflow in DB
+	// endingRunID must not be empty, since this function is trigger messing events of endingRunID
```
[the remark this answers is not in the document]
wxing1292: done

service/worker/replicator/historyRereplicator.go
wxing1292: comment added

service/worker/replicator/historyRereplicator.go
wxing1292: renamed to pending request

wxing1292:
will add an config to turn on the re replication if necessary in the next PR

this PR is so large

PR #1325, opened 2018-12-08
SizeLimit: enforce hitory size/count limit

yiminc:
force terminate workflow if history grow over limit

### DIFF
--- a/service/history/service.go
+++ b/service/history/service.go
@@ -123,8 +123,12 @@ type Config struct {
 	EnableArchival  dynamicconfig.BoolPropertyFnWithDomainFilter
 	NumSysWorkflows dynamicconfig.IntPropertyFn
 
-	BlobSizeLimitError dynamicconfig.IntPropertyFnWithDomainFilter
-	BlobSizeLimitWarn  dynamicconfig.IntPropertyFnWithDomainFilter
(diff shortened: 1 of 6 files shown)


service/history/service.go
```diff
@@ -196,8 +200,12 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int) *Config {
 		EnableArchival:  dc.GetBoolPropertyFnWithDomainFilter(dynamicconfig.EnableArchival, false),
 		NumSysWorkflows: dc.GetIntProperty(dynamicconfig.NumSystemWorkflows, 1000),
 
-		BlobSizeLimitError: dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 2*1024*1024),
-		BlobSizeLimitWarn:  dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitWarn, 256*1024),
+		BlobSizeLimitError:     dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 2*1024*1024),
```
wxing1292: this should be slightly smaller than 2mb, since there will be some metadata included in the message to kafka.

PR #1326, opened 2018-12-08
Fail inflight decision when buffered events exceeds limit

yiminc:
fail in-flight decision if buffered events exceeds limit. By fail the decision, it automatically flush buffered events.


host/integration_test.go
```diff
@@ -72,7 +71,7 @@ func (s *IntegrationBase) setupShards() {
 	}
 }
 
-func TestIntegrationSuite(t *testing.T) {
+func TestRateLimitBufferedEventsTestIntegrationSuite(t *testing.T) {
```
vancexu: Why this name is so long? Can we make it shorter?

PR #1327, opened 2018-12-10
Bugfix: GetActivityTimerTaskIfNeeded did not set the LastTimeoutVisib…

wxing1292:
…ility


service/history/timerQueueActiveProcessor.go
```diff
@@ -350,7 +350,9 @@ Update_History_Loop:
 				continue ExpireActivityTimers
 			}
 
-			if isExpired := tBuilder.IsTimerExpired(td, referenceTime); isExpired {
+			if isExpired := tBuilder.IsTimerExpired(td, referenceTime); !isExpired {
+				break ExpireActivityTimers
+			} else {
```
yiminc: the if will break from the loop, so no need for an else.

PR #1328, opened 2018-12-11
Add new option WorkflowIdReusePolicy on creating new workflow in cli

yux0:
Add new option WorkflowIdReusePolicy on creating new workflow in cli

Issue: https://github.com/uber/cadence/issues/704

### DIFF
--- a/tools/cli/commands.go
+++ b/tools/cli/commands.go
@@ -167,6 +167,7 @@ const (
 	defaultContextTimeoutForLongPoll = 2 * time.Minute
 	defaultDecisionTimeoutInSeconds  = 10
 	defaultPageSizeForList           = 500
+	defaultWorkflowIdReusePolicy     = s.WorkflowIdReusePolicyAllowDuplicateFailedOnly
 
 	workflowStatusNotSet = -1
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
@@ -535,7 +536,11 @@ func StartWorkflow(c *cli.Context) {
 	if len(wid) == 0 {
 		wid = uuid.New()
 	}
-	reusePolicy := getWorkflowIdReusePolicy(c.String(FlagWorkflowIdReusePolicy))
+	reusePolicy := defaultWorkflowIdReusePolicy.Ptr()
+	if c.IsSet(FlagWorkflowIdReusePolicy) {
+		reusePolicy = getWorkflowIdReusePolicy(c.Int(FlagWorkflowIdReusePolicy))
+	}
+
 	input := processJSONInput(c)
 
 	tcCtx, cancel := newContext()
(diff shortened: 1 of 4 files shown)


tools/cli/commands.go
```diff
@@ -1336,3 +1340,23 @@ func getWorkflowStatus(statusStr string) s.WorkflowExecutionCloseStatus {
...
+
+func getWorkflowIdReusePolicy(definedPolicy string) *s.WorkflowIdReusePolicy {
+	if definedPolicy != "" {
+		value, err := strconv.Atoi(definedPolicy)
+		if err != nil {
+			ErrorAndExit(fmt.Sprintf("Option %v format is not supported.", FlagWorkflowIdReusePolicy), err)
+			return nil
+		}
+		policy := s.WorkflowIdReusePolicy(value)
+		for _, p := range s.WorkflowIdReusePolicy_Values() {
```
yux0: Not sure if there is better way to check value in array

tools/cli/workflow.go
```diff
@@ -143,6 +143,11 @@ func newWorkflowCommands() []cli.Command {
 					Usage: "Optional input for the workflow from JSON file. If there are multiple JSON, concatenate them and separate by space or newline. " +
 						"Input from file will be overwrite by input from command line",
 				},
+				cli.StringFlag{
+					Name: FlagWorkflowIdReusePolicyAlias,
+					Usage: "Optional input to configure if the same workflow ID is allow to use for new workflow execution (choose from 0 to 2). " +
```
vancexu: Could we explicit present all options like `0: AllowDuplicateFailedOnly, 1: AllowDuplicate, 2: RejectDuplicate` to save user another link click?

tools/cli/workflow.go
```diff
@@ -143,6 +143,11 @@ func newWorkflowCommands() []cli.Command {
 					Usage: "Optional input for the workflow from JSON file. If there are multiple JSON, concatenate them and separate by space or newline. " +
 						"Input from file will be overwrite by input from command line",
 				},
+				cli.StringFlag{
```
vancexu: Maybe worth using IntFlag, and use IsSet() to tell difference between 0 and isSet.

tools/cli/commands.go
vancexu: To validate input, use `value >= 0 && value < len(s.WorkflowIdReusePolicy_Values())` then you can directly use policy. 
Other way is to use `switch policy ... `.

tools/cli/workflow.go
yux0: Thanks. This is what I am looking for. Updated

tools/cli/workflow.go
yux0: Done.

tools/cli/commands.go
yux0: Good point

PR #1329, opened 2018-12-11
size limit: max ID length

yiminc:
limit string length on various ID fields.
I did not make this dynamic configurable because 1K limit should be enough for all use cases. So this is more to protect some crazy long random string like 1MB of data rather than a control knob that we would want to tweak further.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -71,10 +71,6 @@ const (
 	FailureReasonDecisionBlobSizeExceedsLimit = "DECISION_BLOB_SIZE_EXCEEDS_LIMIT"
 	// TerminateReasonSizeExceedsLimit is reason to terminate workflow when history size exceed limit
 	TerminateReasonSizeExceedsLimit = "HISTORY_SIZE_EXCEEDS_LIMIT"
-
-	// MaxIDLengthLimit is the length limit for various IDs, including: Domain, TaskList, WorkflowID, ActivityID, TimerID,
-	// WorkflowType, ActivityType, SignalName, MarkerName, ErrorReason/FailureReason/CancelCause, Identity, RequestID
-	MaxIDLengthLimit = 1000 // max ID length limit
 )
 
 var (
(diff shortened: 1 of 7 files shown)


common/util.go
```diff
@@ -70,6 +70,10 @@ const (
 	FailureReasonDecisionBlobSizeExceedsLimit = "DECISION_BLOB_SIZE_EXCEEDS_LIMIT"
 	// TerminateReasonSizeExceedsLimit is reason to terminate workflow when history size exceed limit
 	TerminateReasonSizeExceedsLimit = "HISTORY_SIZE_EXCEEDS_LIMIT"
+
+	// MaxIDLengthLimit is the length limit for various IDs, including: Domain, TaskList, WorkflowID, ActivityID, TimerID,
+	// WorkflowType, ActivityType, SignalName, MarkerName, ErrorReason/FailureReason/CancelCause, Identity, RequestID
+	MaxIDLengthLimit = 1000 // max ID length limit
```
samarabbas: I feel this is similar to 'RPS' config knob and should be converted to a system level (no domain override) dynamic config.

service/frontend/workflowHandler.go
```diff
@@ -701,13 +710,20 @@ func (wh *WorkflowHandler) PollForActivityTask(
...
 
+	if len(pollRequest.GetDomain()) > common.MaxIDLengthLimit {
+		return nil, wh.error(errDomainTooLong, scope)
+	}
+
 	if err := wh.validateTaskList(pollRequest.TaskList, scope); err != nil {
 		return nil, err
 	}
+	if len(pollRequest.GetIdentity()) > common.MaxIDLengthLimit {
+		return nil, wh.error(errIdentityTooLong, scope)
```
samarabbas: Let's make sure we log enough information like DomainName, WorkflowID, RunID, etc when logging these errors.

common/util.go
yiminc: changed to dynamic config

service/frontend/workflowHandler.go
yiminc: https://github.com/uber/cadence/issues/1364

PR #1330, opened 2018-12-12
Record failure reason for retry

yiminc:
Record failure reason for retry when server side do ContinueAsNew for retry policy. Will add CronSchedule as another ContinueAsNewInitiator later.


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -354,6 +354,9 @@ struct ContinueAsNewWorkflowExecutionDecisionAttributes {
   50: optional i32 taskStartToCloseTimeoutSeconds
   60: optional i32 backoffStartIntervalInSeconds
   70: optional RetryPolicy retryPolicy
+  80: optional ContinueAsNewInitiator initiator
+  90: optional string reason
```
samarabbas: Let's rename these to failureReason and failureDetails.  Add a separate filed for lastCompletionResult.

PR #1336, opened 2018-12-14
Added Prometheus metrics

### LINKED ISSUE #769: Feature request: Expose prometheus metrics at a /metrics endpoint
We primarily use prometheus for service monitoring.  It would be great if Cadence could expose prometheus formatted metrics at a /metrics endpoint.

h7kanna:
Hi team,

I made changes to cadence to expose prometheus metrics 
using https://github.com/uber-go/tally/tree/master/prometheus

This resolves #769 

But it does not work as the metric names in the cadence are not compatible with Prometheus query language. https://github.com/uber/cadence/blob/master/common/metrics/defs.go#L1021

Basically Prometheus does not like '-' and '.' in the names.
https://prometheus.io/docs/practices/naming/

There are two options,

1) Rename the metrics.
But this will be disruptive to the existing monitoring rules setup of Cadence within Uber. (I want to know how disruptive it will be?)

2) Rename the metrics on the fly when Prometheus monitoring is on.

Please suggest which one is acceptable.

Thanks,
Harsha

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -299,6 +299,17 @@
   version = "v1.0.2"
 
 [[projects]]
+  digest = "1:4142d94383572e74b42352273652c62afec5b23f325222ed09198f46009022d1"
+  name = "github.com/m3db/prometheus_client_golang"
+  packages = [
+    "prometheus",
+    "prometheus/promhttp",
+  ]
+  pruneopts = ""
+  revision = "c5b7fccd204277076155f10851dad72b76a49317"
+  version = "v0.8.0"
+
+[[projects]]
   branch = "master"
   digest = "1:b7ac1ad7d98781c0168803a258d7bf8b3544fd59f356a0e2dd65c43312cf6aac"
   name = "github.com/mailru/easyjson"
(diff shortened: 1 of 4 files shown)


samarabbas:
@h7kanna option 1) is not going to work as it will break all Cadence dashboards for us.  Let's go with option 2) for now.  I'll try to look into PR this coming week to see if we can get prometheus support landed to master without breaking existing deployments.

sagikazarmark:
I would love to see this feature in Cadence. 😉

h7kanna:
When I tried, I couldn't make it completely work because of Prometheus client panicking on naming conflicts with uber-tally scoped metric names.
I will find some time this week to rebase this

shreyassrivatsan:
With #1575, we started dual emitting metrics which should be compatible with prometheus. So should be able to scrape prometheus metrics with the appropriate tally config now.

h7kanna:
oh nice. I will verify after rebasing. Thanks

shreyassrivatsan:
Closing as #1770 supersedes and adds prometheus metrics support.

PR #1341, opened 2018-12-18
Cadence Indexer

vancexu:
Cadence indexer is a new component for consuming visibility data from Kafka and writing to ElasticSearch. It used [BulkProcessor](https://github.com/olivere/elastic/wiki/BulkProcessor) 

Todo:
~~- Clean logging~~
~~- Add metrics~~
~~- Make some arguments configurable~~
~~- Add tests ~~
- Add delete logic (in separate PR)


vancexu:
Hi @huawangpp thanks for bring this up! I considered it during my design but I was told by ESaaS team there is potential data loss, so we decide to customize. If that's not the case, I'd like to have a meeting with you to see if Cadence can use it. Thanks again.

huawangpp:
@vancexu, got it. We indeed doesn't support loss less at this moment. It is best effort right now. It is on our roadmap though. I also notice bulkprocessor has similar SLA? It will be good to meet and go over the details.

vancexu:
@huawangpp Not sure what do you mean by similar SLA. But feel free to schedule meeting with me (@boweixu) when you have time. Thanks

huawangpp:
@vancexu sure, let's sync up offline.

config/development.yaml
```diff
@@ -81,5 +81,20 @@ kafka:
...
+    cadence-visibility:
       cluster: test
+    cadence-visibility-dlq:
+      cluster: test
+  applications:
+    visibility:
+      topic: cadence-visibility
+      dlq-topic: cadence-visibility-dlq
+
+elasticsearch:
```
samarabbas: Can we leave development config to use Cassandra as visibility store?

idl/github.com/uber/cadence/indexer.thrift
```diff
@@ -0,0 +1,41 @@
...
+
+include "shared.thrift"
+
+enum VisibilityMsgType {
+  Open
+  Closed
+  Delete
+}
+
+struct VisibilityMsg {
```
samarabbas: Let's use similar technique we use in other places.  Having separate visibilityMessageAttributes and a wrapper VisibilityMessage where you can switch between various using a messageType field.

service/history/transferQueueProcessorBase.go
```diff
@@ -212,6 +215,26 @@ func (t *transferQueueProcessorBase) recordWorkflowClosed(
...
+			DomainID:      common.StringPtr(domainID),
+			WorkflowID:    common.StringPtr(wid),
+			RunID:         common.StringPtr(execution.GetRunId()),
+			WorkflowType:  common.StringPtr(workflowTypeName),
+			StartTime:     common.Int64Ptr(startTimeUnixNano),
+			CloseTime:     common.Int64Ptr(endTimeUnixNano),
+			CloseStatus:   &closeStatus,
+			HistoryLength: common.Int64Ptr(historyLength),
+		}
+		err := t.visibilityProducer.Publish(msg)
```
samarabbas: Can you make sure we have request/failure/latency metric emitted for visibilityProducer.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		kafkaMsg.Ack() // duplicate message
+		return
+	}
+	p.mapToKafkaMsg.Put(key, kafkaMsg)
+	p.processor.Add(request)
+}
+
+// bulkAfterAction is triggered after bulk processor commit
+func (p *esProcessorImpl) bulkAfterAction(id int64, requests []elastic.BulkableRequest, response *elastic.BulkResponse, err error) {
+	if err != nil {
```
samarabbas: What is the impact if we start failing these messages? Will this cause kafka ingestor to stop pumping messages to indexer as you accumulate lots of unacked message?  What is the value at which it stops? Is it configurable.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		// This happens after configured retry, which means something bad happens on cluster or index
+		// When cluster back to live, processor will re-commit those failure requests
+		p.logger.WithFields(bark.Fields{
+			logging.TagErr: err,
+		}).Error("Error commit bulk request.")
+
+		for _, request := range requests {
+			p.logger.WithFields(bark.Fields{
+				logging.TagESRequest: request.String(),
```
samarabbas: Will this be too noisy?

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (p *esProcessorImpl) Stop() {
+	p.processor.Stop()
+	p.mapToKafkaMsg = nil
+}
+
+// Add an ES request, and an map item for kafka message
+func (p *esProcessorImpl) Add(request elastic.BulkableRequest, key string, kafkaMsg messaging.Message) {
+	if p.mapToKafkaMsg.Contains(key) {
+		kafkaMsg.Ack() // duplicate message
```
samarabbas: This looks dangerous, could this result in ack'ing unprocessed message? Can we also store the offset as part of the key to make sure we don't ack unprocessed message.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		return
+	}
+	p.mapToKafkaMsg.Put(key, kafkaMsg)
```
samarabbas: Why do we need to store the entire kafkaMsg in the map? Who does the retry? If retry is taken care of by bulk processor then let's just keep the key in the map.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		if visType == "" {
+			continue
+		}
+		responseItem := responseItems[i]
+		for _, resp := range responseItem {
+			key := resp.Id + visType
+			switch {
+			case isResponseSuccess(resp.Status):
+				p.ackKafkaMsg(key)
+			case !isResponseRetriable(resp.Status):
```
samarabbas: I'm not sure how the action is getting retried.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+
+	p.logger.Errorf("Unknown version %v in ES request. ", version)
+	p.metricsClient.IncCounter(metrics.ESProcessorScope, metrics.ESProcessorCorruptedData)
+	return ""
+}
+
+// 409 - Version Conflict
+// 404 - Not Found
+func isResponseSuccess(status int) bool {
+	if status >= 200 && status < 300 || status == 409 || status == 404 {
```
samarabbas: Why '404' is treated as success? Won't this result is loosing messages?

config/development.yaml
vancexu: Cassandra is always used. ES will become an additional store only when this config switched to enabled.

idl/github.com/uber/cadence/indexer.thrift
vancexu: I updated data interface to be more generic.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+		FlushInterval(config.ESProcessorFlushInterval()).
+		Backoff(elastic.NewExponentialBackoff(esProcessorInitialRetryInterval, esProcessorMaxRetryInterval)).
+		After(p.bulkAfterAction).
+		Do(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	p.processor = processor
+	p.mapToKafkaMsg = collection.NewShardedConcurrentTxMap(1024, p.hashFn)
```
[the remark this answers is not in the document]
vancexu: I think it is fine to init with 1024 (each map is init with 32 since we use fixed 32 shards). Because writer to this map have `config.IndexerConcurrency` (default 1000) concurrency, while reader of this map have `config.ESProcessorNumOfWorker` (default 1) concurrency.

service/worker/indexer/esProcessor.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	p.mapToKafkaMsg.Remove(key)
+}
+
+func (p *esProcessorImpl) hashFn(key interface{}) uint32 {
+	id, ok := key.(string)
+	if !ok {
+		return 0
+	}
+	numOfShards := p.config.IndexerConcurrency()
```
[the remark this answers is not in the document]
vancexu: "IndexerConcurrency" is number of goroutines consume message from kafka (and add to esProcessor). "ESProcessorNumOfWorkers" is number of goroutines commit batch to ES.

service/history/transferQueueProcessorBase.go
vancexu: Added. Changed to use metricsProducer

service/worker/indexer/esProcessor.go
vancexu: It shouldn't be that noisy. 
This err section happens when the bulk request returned with err, which means all requests in this bulk failed. This may be caused by ES cluster issue, in which case we log all requests in bulk (only one time) so that later when ES is back online, we can check if these error requests been resent successfully.

service/worker/indexer/esProcessor.go
vancexu: I am not sure how this can result in acking unprocessed message, but I removed this optimization since there is no correctness impact.

service/worker/indexer/esProcessor.go
vancexu: Because the reader of this map (bulk processor) need to ack the kafkaMsg, there has to be a place I buffered these kafka messages. It is bulk processor do the retry, but I don't understand how could we ack with only key?

service/worker/indexer/esProcessor.go
vancexu: Every requests added to ElasticBulkProcessor will be removed from queue after successfully committed, or failed with non-retriable status code (configurable). All the other requests not removed will stay in queue and retry forever.

service/worker/indexer/esProcessor.go
vancexu: 404 can be returned by delete requests (not yet added). It's treated as success, since the doc to delete not found.

service/history/historyEngine.go
```diff
@@ -167,6 +167,9 @@ func NewEngineWithShardContext(
 	var visibilityProducer messaging.Producer
 	if config.EnableVisibilityToKafka() {
 		visibilityProducer = getVisibilityProducer(messagingClient)
+		if shard.GetMetricsClient() != nil {
```
samarabbas: move this logic inside getVisibilityProducer.

idl/github.com/uber/cadence/indexer.thrift
```diff
@@ -0,0 +1,50 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+namespace java com.uber.cadence.indexer
+
+include "shared.thrift"
+
+enum MessageType {
```
samarabbas: Can we create a separate enum to select which database to update and include it as part of each Message.
```
enum IndexType {
  Visibility
}
```

idl/github.com/uber/cadence/indexer.thrift
```diff
@@ -0,0 +1,50 @@
...
+  40: optional bool boolData
+}
+
+struct Message {
+  10: optional MessageType messageType
+  20: optional string domainID
+  30: optional string workflowID
+  40: optional string runID
+  50: optional i64 (js.type = "Long") version
+  60: optional map<string,Field> fields
```
samarabbas: Can you move this into its own attributes for Index type.  IndexAttributes.

service/history/transferQueueProcessorBase.go
```diff
@@ -162,11 +163,18 @@ func (t *transferQueueProcessorBase) recordWorkflowStarted(
 
 	// publish to kafka
 	if t.visibilityProducer != nil {
-		msg := &messaging.OpenWorkflowMsg{
-			Domain:     domain,
-			WorkflowID: wid,
-			RunID:      rid,
-			StartTime:  startTimeUnixNano,
+		msgType := indexer.MessageTypeIndex
```
samarabbas: Can you move this logic into separate type like VisibilityIndex.

idl/github.com/uber/cadence/indexer.thrift
vancexu: I won't include this for now, and if I do need anything for insight, I will create in this way.

PR #1352, opened 2018-12-21
Implement Workflow Reset on Active Side

longquanzheng:


service/history/stateBuilder.go
```diff
@@ -94,8 +99,10 @@ func (b *stateBuilderImpl) applyEvents(domainID, requestID string, execution sha
 	var newRunStateBuilder mutableState
 	for _, event := range history {
 		lastEvent = event
-		// must set the current version, since this is standby here, not active
-		b.msBuilder.UpdateReplicationStateVersion(event.GetVersion(), true)
+		if b.msBuilder.GetReplicationState() != nil {
```
wxing1292: for cross dc workflow, this check will always be true

service/history/historyReplicator.go
```diff
@@ -348,6 +353,123 @@ func (r *historyReplicator) ApplyEvents(ctx context.Context, request *h.Replicat
 	}
 }
 
+func (r *historyReplicator) checkDuplicatedReplicationEvent(context *workflowExecutionContext, logger bark.Logger) error {
```
wxing1292: is this function used anywhere?

service/history/historyReplicator.go
```diff
@@ -348,6 +353,123 @@ func (r *historyReplicator) ApplyEvents(ctx context.Context, request *h.Replicat
...
+		return nil
+	}
+	if _, ok := err.(*shared.EntityNotExistsError); !ok {
+		// GetWorkflowExecution failed with some transient error. Return err so we can retry the task later
+		return err
+	}
+	return nil
+}
+
+func (r *historyReplicator) ApplyResetEvent(ctx context.Context, request *h.ReplicateEventsRequest, logger bark.Logger) (retError error) {
```
wxing1292: the logic added can work, only for the simple case, for complex case, such as conflict resolution (which requires version comparison), as well as return retry error with hints for re-sending missing kafka message, the new logic will not work.

moreover, the application of events to start the reseted workflow, as well as termination of the current running workflow, should be handled inside the state builder. 

history replicator should only handle the start event (because this is special, require create workflow & conflict resolution), common non start workflow event conflict resolution, update workflow execution and force termination of current running workflow for conflict resolution.

service/history/historyReplicator.go
longquanzheng: No, it's already abondoned. I will remove it.

service/history/stateBuilder.go
longquanzheng: I will fix the comment, it is also being used in the active side.

service/history/historyReplicator.go
longquanzheng: I see, I also realized that yesterday.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -425,8 +429,8 @@ workflow_state = ? ` +
 		`and visibility_ts = ? ` +
 		`and task_id = ?`
 
-	templateUpdateWorkflowExecutionQuery = `UPDATE executions ` +
-		`SET execution = ` + templateWorkflowExecutionType + `, next_event_id = ? ` +
+	templateCheckWorkflowExecutionQuery = `UPDATE executions ` +
```
samarabbas: I think we should let the reset go through even if next_event_id changes.  All we need to make sure is current execution record does not change during this operation.

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1076,6 +1076,11 @@ func updateBufferedEvents(tx *sqlx.Tx, batch *p.DataBlob, clear bool, shardID in
 	return nil
 }
 
+func (m *sqlExecutionManager) ResetWorkflowExecution(request *p.InternalResetWorkflowExecutionRequest) error {
+	// TODO need to implement with history eventsV2 API
+	return fmt.Errorf("Not implemented")
```
samarabbas: Can you file a task to implement this later?

service/history/historyBuilder.go
```diff
@@ -119,10 +119,8 @@ func (b *historyBuilder) AddDecisionTaskTimedOutEvent(scheduleEventID int64,
 	return b.addEventToHistory(event)
 }
 
-func (b *historyBuilder) AddDecisionTaskFailedEvent(scheduleEventID int64, startedEventID int64,
-	cause workflow.DecisionTaskFailedCause, details []byte, identity string) *workflow.HistoryEvent {
-	event := b.newDecisionTaskFailedEvent(scheduleEventID, startedEventID, cause, details, identity)
-
+func (b *historyBuilder) AddDecisionTaskFailedEvent(attr workflow.DecisionTaskFailedEventAttributes) *workflow.HistoryEvent {
```
samarabbas: It kind of breaks the existing pattern.  This API takes the necessary fields as input parameters and then build the corresponding *EventAttributes within this implementation.  Can you just add the new fields as input parameters for now.

service/history/stateBuilder.go
```diff
@@ -569,3 +579,64 @@ func (b *stateBuilderImpl) getTimerBuilder(event *shared.HistoryEvent) *timerBui
 	}
 	return newTimerBuilder(b.shard.GetConfig(), b.logger, timeSource)
 }
+
+func (b *stateBuilderImpl) filterTransferTasksForReset() ([]int64, []int64, []persistence.Task, error) {
```
samarabbas: I think it is much safer to generate transferTasks from given mutable state rather than filtering them out from the list given by stateBuilder.

service/history/historyEngine.go
```diff
@@ -2557,6 +2560,526 @@ func (e *historyEngineImpl) SyncActivity(ctx context.Context, request *h.SyncAct
...
+		return
+	}
+	defer func() { forkRelease(retError) }()
+	forkMutableState, retError := forkContext.loadWorkflowExecution()
+	if retError != nil {
+		return
+	}
+
+	// also load the current run of the workflow, it can be different from the forking runID
+	resp, retError := e.shard.GetExecutionManager().GetCurrentExecution(&persistence.GetCurrentExecutionRequest{
```
samarabbas: This file is already getting large.  Can you move reset logic to its own component, so it is easy to maintain and test.

common/persistence/cassandra/cassandraPersistence.go
longquanzheng: Resolved offline. We still need to check next_event_id to prevent corrupting the workflow, there shouldn't be any cost or bad customer experience because of that

common/persistence/sql/sqlExecutionManager.go
longquanzheng: OK https://github.com/uber/cadence/issues/1368

service/history/historyBuilder.go
longquanzheng: OK

service/history/stateBuilder.go
longquanzheng: OK

service/history/historyEngine.go
longquanzheng: OK

PR #1355, opened 2018-12-22
Update cli to be able to set client factory from outside module

andrewjdawson2016:


tools/cli/commands.go
```diff
@@ -180,8 +180,12 @@ const (
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
 )
 
+func SetFactory(factory ClientFactory) {
+	CFactory = factory
+}
+
 var (
-	cFactory ClientFactory
+	CFactory ClientFactory
```
longquanzheng: probably change back to module private if you decided to use Set method

andrewjdawson2016: Oh yes you are right. Thank you.

PR #1357, opened 2018-12-26
Fix stale cache issue when mutableState is dirty

longquanzheng:


service/history/replicatorQueueProcessor.go
```diff
@@ -193,7 +193,8 @@ func (p *replicatorQueueProcessorImpl) processSyncActivityTask(task *persistence
 		},
 	}
 
-	return p.replicator.Publish(replicationTask)
+	retError = p.replicator.Publish(replicationTask)
```
wxing1292: this change does nothing

service/history/timerQueueActiveProcessor.go
```diff
@@ -262,13 +262,15 @@ Update_History_Loop:
 			hasTimer, ti := tBuilder.GetUserTimer(td.TimerID)
 			if !hasTimer {
 				t.logger.Debugf("Failed to find in memory user timer: %s", td.TimerID)
-				return fmt.Errorf("Failed to find in memory user timer: %s", td.TimerID)
+				retError = fmt.Errorf("Failed to find in memory user timer: %s", td.TimerID)
```
samarabbas: Another place on line 251

service/history/replicatorQueueProcessor.go
longquanzheng: Actually you are right. Returning error will also update it for defer func(){}() 
The only problem can cause it is in processActivityRetryTimer

PR #1361, opened 2018-12-27
Fix incorrect history length in visibility

vancexu:
Related issue https://github.com/uber/cadence/issues/1359

### DIFF
--- a/host/integration_test.go
+++ b/host/integration_test.go
@@ -3452,7 +3452,7 @@ func (s *integrationSuite) TestVisibility() {
 	s.Nil(err1)
 
 	// wait until the start workflow is done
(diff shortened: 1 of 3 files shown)


host/integration_test.go
```diff
@@ -3504,6 +3504,7 @@ func (s *integrationSuite) TestVisibility() {
 		s.Nil(err3)
 		closedCount = len(resp.Executions)
 		if closedCount == 1 {
+			s.Equal(int64(5), *(resp.Executions[0].HistoryLength))
```
dmetzgar: Could also be next to the `s.Equal(1, closedCount)` below

PR #1362, opened 2018-12-30
Add archival config and support file based blobstore

andrewjdawson2016:
**What does this diff contain?**
- An implementation of blobstore interface which is backed by local file system (this is most of the diff)
- Static config which is used to construct a file based blobstore at cadence boot-up time.
- A static config to enable/disable archival for all of cadence. If this config is set to false then cadence will startup without archival.
- A dynamic config override to enable/disable archival for all of cadence
- Plumbing needed to send signals to archival system workflows
- Dummy implementation of archival activity which trivially uses blobstore
- Method constructing archival record name based on hash of domainID, workflowID and runID
- Updated unit tests to assert archival was invoked

**What is not done as part of this diff that will be done in follow up diffs?**
- Implementation of archival activity (this includes visibility metrics and scheduling a timer)
- Implementation of backfill activity (current diff just stubs out backfill activity)
- Integration tests (current diff updates unit tests but archival is currently not enabled in integration tests). This was left as a follow up item because archival activity is not finished.
- Playing nicely with cross DC. Must make sure we do not do double uploads for all archivals.
- Gracefully stopping archival when we encounter too many errors

**What does running locally with archival enabled look like?**
On cadence boot-up time a file based blobstore client will be created, based on static yaml config. The client will write archival records to /tmp/development/blobstore/[buckets]. On startup there is a default bucket and two custom buckets created. File based blobstore implementation does not create buckets dynamically, this is because an implementation based on S3 or other blobstore would not do dynamic bucket creation. 

A new domain can be registered with archival enabled e.g. `cadence --do samples-domain admin domain register --archival_enabled true`

An existing domain can enable archival e.g. `cadence --do samples-domain admin domain update --archival_status enabled` 

When a domain first enables archival (either on update or register) a custom bucket name can be optionally provided. If no custom bucket name is provided the default bucket will be used. 

When a domain which has archival enabled has a workflow which completes the following will happen
- History engine will use archival client to send a signal to archival system workflow
- Archival client will send a signal with start to a system workflow which can handle the archival
- The signal contains information about the user workflow which finished
- The system workflow listens for signals and dispatches signals to archival activities
- Archival activities use a blobstore client to archive history (currently they are uploading a dummy blob not real history)
- Archival activities retry forever.


cmd/server/server.go
```diff
@@ -143,6 +142,13 @@ func (s *server) startService() common.Daemon {
 		params.MessagingClient = nil
 	}
 
+	if params.ClusterMetadata.IsArchivalEnabled() {
```
venkat1109: any reason this lives inside cluster metadata instead of being a flag like s.cfg.Archival.Enabled ? My understanding is cluster metadata is reserved for things like cluster name / addr / peers etc and not for other server side feature flags

common/blobstore/filestore/client.go
```diff
@@ -0,0 +1,179 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package filestore
+
+import (
+	"context"
+	"fmt"
+	"github.com/uber/cadence/common/blobstore"
```
venkat1109: gofmt

common/blobstore/filestore/util.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return nil
+}
+
+func readFile(filepath string) ([]byte, error) {
+	return ioutil.ReadFile(filepath)
+}
+
+func serializeBucketConfig(bucketCfg *BucketConfig) ([]byte, error) {
+	return yaml.Marshal(bucketCfg)
```
venkat1109: any reason you use yaml for bucket config and gob for blobs ? can you use gob for both ?

common/cluster/metadata.go
```diff
@@ -49,8 +49,11 @@ type (
 		ClusterNameForFailoverVersion(failoverVersion int64) string
 		// GetAllClientAddress return the frontend address for each cluster name
 		GetAllClientAddress() map[string]config.Address
-		// GetDeploymentGroup returns the deployment group of cluster
-		GetDeploymentGroup() string
+
+		// IsArchivalEnabled whether archival is enabled
+		IsArchivalEnabled() bool
```
venkat1109: see comment about this feature flag above

common/service/config/config.go
```diff
@@ -237,6 +237,14 @@ type (
 		FlushBytes int `yaml:"flushBytes"`
 	}
 
+	// Archival contains the config for archival
+	Archival struct {
+		// EnableArchival whether archival is enabled
+		EnableArchival bool `yaml:"enableArchival"`
```
venkat1109: s/EnableArchival/Enabled - Archical.EnableArchival is redundant

common/service/config/config.go
```diff
@@ -237,6 +237,14 @@ type (
...
+		// Blobstore the configuration for blobstore
+		Blobstore filestore.Config `yaml:"blobstore"`
```
venkat1109: - can we call this filestore instead of blobstore so that we can add additional config item for each store type (s3, gs, azureblobstore etc) in future ? I don' imagine specifying storeDirectory or bucket retention / ownership for non-filestore based options
- optionally, you can also consider having a storeType enum {filestore, s3, gs... }

cmd/server/server.go
andrewjdawson2016: Talked offline about this one.

common/blobstore/filestore/util.go
andrewjdawson2016: I could use gob for both of them but I wanted bucketConfig to be easily human readable.

Imagine the case where for some reason someone wants to test creating a custom bucket that was not created during startup. In order to do this they would create a new directory and create a bucketConfig file. I want that bucket config file to be easy to create.

common/cluster/metadata.go
andrewjdawson2016: spoke offline about this one.

common/blobstore/filestore/util.go
venkat1109: Fair enough. I know we use yaml all over the place - but consider using json to yaml for things like this - https://noyaml.com/  -

andrewjdawson2016:
rerunning tests

PR #1366, opened 2019-01-04
sql: make it easy to support heterogeneous SQL databases

venkat1109:
This patch contains a major refactoring of the sql persistence layer (with no behavior change). The goal of the refactoring is to make it easy to add support for other kinds of SQL DBs like SQLServer, SQLLite etc (today, the only supported DB is mysql). MySQL DBs differ in syntax and the supported types (see http://troels.arvin.dk/db/rdbms/). Today, adding support for a new DB involves re-implementing the entire persistence API interface. When the DBs differ only in syntax / types and more importantly, when the way the data is structured (schema) can remain the same, reimplementing the entire persistence api involves tons of code duplication. 

This patch makes it easy to deal with these syntactic differences by separating the persistence api layer from the sql client / query invocation layer. This way, DBs that only have syntax / type differences can simply implement the sql client / query invocation layer without having to touch the persistence api at all. DBs where the current schema cannot work (for efficiency reasons) - will need to re-implement the persistence api interface.


common/persistence/sql/storage/mysql/execution.go
```diff
@@ -0,0 +1,510 @@
...
+
+// InsertIntoTimerTasks inserts one or more rows into timer_tasks table
+func (mdb *DB) InsertIntoTimerTasks(rows []sqldb.TimerTasksRow) (sql.Result, error) {
+	return mdb.conn.NamedExec(createTimerTasksQry, rows)
+}
+
+// SelectFromTimerTasks reads one or more rows from timer_tasks table
+func (mdb *DB) SelectFromTimerTasks(filter *sqldb.TimerTasksFilter) ([]sqldb.TimerTasksRow, error) {
+	var rows []sqldb.TimerTasksRow
+	err := mdb.conn.Select(&rows, getTimerTasksQry, filter.ShardID, *filter.MinVisibilityTimestamp,
```
mfateev: Could we document in the filter structs or in DB interface methods when pointer fields are required?
Looking at the API it is not clear that it is going to dereference these pointers without checking.

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -589,51 +596,116 @@ type (
 		LockCurrentExecutions(filter *CurrentExecutionsFilter) (string, error)
 
 		InsertIntoTransferTasks(rows []TransferTasksRow) (sql.Result, error)
+		// SelectFromTransferTasks returns rows that match filter criteria from transfer_tasks table.
+		// Required filter params - {shardID, minTaskID, maxTaskID}
 		SelectFromTransferTasks(filter *TransferTasksFilter) ([]TransferTasksRow, error)
+		// DeleteFromTransferTasks deletes one or more rows from transfer_tasks table.
+		// Filter params - shardID is required. If TaskID is not nil, a single row is deleted.
+		// When MinTaskID and MaxTaskID are not-nil, a range of rows are deleted.
 		DeleteFromTransferTasks(filter *TransferTasksFilter) (sql.Result, error)
```
mfateev: An alternative would be to have two methods: DeleteTransferTask and DeleteTransferTaskRange with different filters.

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -589,51 +596,116 @@ type (
...
 
 		InsertIntoTimerTasks(rows []TimerTasksRow) (sql.Result, error)
+		// SelectFromTimerTasks returns one or more rows from timer_tasks table
+		// Required filter Params - {shardID, taskID, minVisibilityTimestamp, maxVisibilityTimestamp, pageSize}
 		SelectFromTimerTasks(filter *TimerTasksFilter) ([]TimerTasksRow, error)
+		// DeleteFromTimerTasks deletes one or more rows from timer_tasks table
+		// Required filter Params:
+		//  - to delete one row - {shardID, visibilityTimestamp, taskID}
+		//  - to delete multiple rows - {shardID, minVisibilityTimestamp, maxVisibilityTimestamp}
 		DeleteFromTimerTasks(filter *TimerTasksFilter) (sql.Result, error)
```
mfateev: Consider changing to two methods, single and range deletes.

PR #1376, opened 2019-01-07
Move historyV2 delete into util helper and add comment

andrewjdawson2016:


longquanzheng:
thanks for adding the comments

PR #1386, opened 2019-01-09
Add cron support to CLI

yiminc:
I have to use client's master branch as the new release is not cut yet.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -526,7 +526,8 @@
   version = "v1.3.2"
 
 [[projects]]
-  digest = "1:fd58696ccbfea94bb5773204f42c11416197587a1d22da198c251bcee859e572"
+  branch = "master"
+  digest = "1:1b1b609e54de5a35b56de2a4c0896a9bf56cc5d7c5438cead1bf6a68f5b627c0"
   name = "go.uber.org/cadence"
   packages = [
     ".",
@@ -547,8 +548,7 @@
     "workflow",
   ]
   pruneopts = ""
-  revision = "88b9aab2063a542ee2f6c0c78694d2c3bee6dd0f"
-  version = "v0.7.6"
+  revision = "4d59d362aae4fecd28c803dec4ce05c6bfc90382"
 
 [[projects]]
   digest = "1:22c7effcb4da0eacb2bb1940ee173fac010e9ef3c691f5de4b524d538bd980f5"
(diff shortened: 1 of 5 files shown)


tools/cli/commands.go
```diff
@@ -559,7 +560,7 @@ func StartWorkflow(c *cli.Context) {
 	tcCtx, cancel := newContext()
 	defer cancel()
 
-	resp, err := serviceClient.StartWorkflowExecution(tcCtx, &s.StartWorkflowExecutionRequest{
+	startRequest := &s.StartWorkflowExecutionRequest{
```
longquanzheng: If you use the serverFrontendClient, then you don't have to update the client repo IDLs to have CLI. Refering this: https://github.com/uber/cadence/blob/master/tools/cli/commands.go#L1176

Gopkg.lock
```diff
@@ -526,7 +526,7 @@
   version = "v1.3.2"
 
 [[projects]]
-  digest = "1:fd58696ccbfea94bb5773204f42c11416197587a1d22da198c251bcee859e572"
+  digest = "1:5e70aa639271a54e9ed87875b9f7e495255df54b1a64c7d4a338e33b44ee33b7"
```
longquanzheng: You also need to stash this change since you don't change Gopkg.toml anymore

yiminc: done

tools/cli/workflow.go
```diff
@@ -134,6 +134,17 @@ func newWorkflowCommands() []cli.Command {
 					Value: defaultDecisionTimeoutInSeconds,
 					Usage: "Decision task start to close timeout in seconds",
 				},
+				cli.StringFlag{
```
vancexu: Could you also add this to `RunWorkflow`? There are duplicate code from RunWorkflow and `StartWorkflow` which would be good to refactor.

tools/cli/commands.go
```diff
@@ -613,22 +620,27 @@ func RunWorkflow(c *cli.Context) {
...
 			Name: common.StringPtr(taskList),
 		},
 		Input:                               []byte(input),
 		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(int32(et)),
 		TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(int32(dt)),
 		Identity:                            common.StringPtr(getCliIdentity()),
 		WorkflowIdReusePolicy:               reusePolicy,
-	})
+	}
+	if c.IsSet(FlagCronSchedule) {
```
vancexu: You also need to add this flag to to Run command, like the one you already added for Start command.

PR #1404, opened 2019-01-18
Fix visibility producer initialization 

vancexu:
Currently each shard will create a visibility producer, which will lead to kafka "peer reset" error when load is large. 
This PR fix this by moving visibility producer initialization process to history handler, which is host level. 
This PR also include:
1. a fix on metrics client being wrapped twice. 
2. a dep update


vancexu:
This is closed because unexpected gen files changed.

PR #1407, opened 2019-01-23
Add esVisibiltyManager for list workflow APIs on ElasticSearch

vancexu:
This PR add an implementation for VisibilityManager on ES, so that users can make use of visibility data on ElasticSearch using existing list APIs.


service/worker/indexer/esProcessor.go
```diff
@@ -85,7 +85,7 @@ func NewESProcessorAndStart(config *Config, client *elastic.Client, processorNam
 		metricsClient: metricsClient,
 	}
 
-	processor, err := client.BulkProcessor().
+	processor, err := client.GetRawClient().BulkProcessor().
```
samarabbas: Do not expose the entire raw client.  You can just expose BulkProcessor.

service/frontend/workflowHandler.go
```diff
@@ -3112,3 +3119,13 @@ func (wh *WorkflowHandler) bucketName(customBucketName *string) string {
 func (wh *WorkflowHandler) customBucketNameProvided(customBucketName *string) bool {
 	return customBucketName != nil && len(*customBucketName) != 0
 }
+
+func (wh *WorkflowHandler) getVisibilityManager(domain string) persistence.VisibilityManager {
```
samarabbas: Can we wrap this under a wrapper implementation which also implements VisibilityMgr interface.

common/persistence/elasticsearchVisibility.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+func (v *esVisibilityManager) Close() {}
+
+func (v *esVisibilityManager) GetName() string {
+	return esPersistenceName
+}
+
+func (v *esVisibilityManager) RecordWorkflowExecutionStarted(request *RecordWorkflowExecutionStartedRequest) error {
+	return errOperationNotSupported
```
samarabbas: Can we have Kafka publisher code moved here.

service/worker/indexer/esProcessor.go
vancexu: updated

service/frontend/workflowHandler.go
vancexu: good suggestion, updated.

common/persistence/elasticsearchVisibility.go
vancexu: ok, will do in separate PR.

PR #1408, opened 2019-01-23
Fix visibility producer init process 

vancexu:
Currently each shard will create a visibility producer, which will lead to kafka "peer reset" error when load is large.
This PR fix this by moving visibility producer initialization process to history handler, which is host level.
This PR also include:

a fix on metrics client being wrapped twice.
a fix on consumer name.

### DIFF
--- a/config/development.yaml
+++ b/config/development.yaml
@@ -105,8 +105,8 @@ kafka:
       cluster: test
   applications:
     visibility:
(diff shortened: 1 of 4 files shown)


vancexu:
This is a duplicate RP to avoid wield godep update. 
https://github.com/uber/cadence/pull/1404

PR #1409, opened 2019-01-23
Add frontend DC redirection functionality and policy

wxing1292:


client/clientBean.go
```diff
@@ -45,6 +46,7 @@ type (
 		GetHistoryClient() history.Client
 		GetMatchingClient() matching.Client
 		GetFrontendClient() frontend.Client
+		GetPublicClient() public.Client
```
samarabbas: FrontendClient and PublicClient are same thing.  Why do we need separate definitions?

service/frontend/workflowHandler.go
```diff
@@ -155,13 +158,22 @@ func NewWorkflowHandler(sVice service.Service, config *Config, metadataMgr persi
...
 // Start starts the handler
 func (wh *WorkflowHandler) Start() error {
-	wh.Service.GetDispatcher().Register(workflowserviceserver.New(wh))
+	currentClusteName := wh.GetClusterMetadata().GetCurrentClusterName()
+	dcRedirectionPolicy := RedirectionPolicyGenerator(
+		wh.GetClusterMetadata(),
+		wh.domainCache,
+		wh.dcRedirectionPolicy,
+	)
+	dcRediectionHandle := NewDCRedirectionHandler(
```
samarabbas: nit: typo 'dcRediectionHandle'

service/frontend/workflowHandler.go
```diff
@@ -155,13 +158,22 @@ func NewWorkflowHandler(sVice service.Service, config *Config, metadataMgr persi
...
+		currentClusteName, dcRedirectionPolicy, wh.Service, wh,
+	)
+	wh.Service.GetDispatcher().Register(workflowserviceserver.New(dcRediectionHandle))
```
samarabbas: I'm a little worried that we always register the redirect handler.  I think RedirectHandler needs to wrap WorkflowHandler and only registered when we need forwarding.

client/clientBean.go
wxing1292: because each time the idl on the frontend is updated, developer will have to update the public client first, then get it approved, then update the public sever repo pointing to public client, even for developing features.

service/frontend/workflowHandler.go
wxing1292: fixed

service/frontend/workflowHandler.go
wxing1292: take a look at the noop redirection policy, which is only 6 lines of code (see below).
if we trust that the redirection handler works, we should also trust the code below.

```
// GetTargetDatacenterByName get target cluster name by domain Name
func (policy *NoopRedirectionPolicy) GetTargetDatacenterByName(domainName string) (string, error) {
	return policy.currentClusteName, nil
}

// GetTargetDatacenterByID get target cluster name by domain ID
func (policy *NoopRedirectionPolicy) GetTargetDatacenterByID(domainID string) (string, error) {
	return policy.currentClusteName, nil
}
```

client/clientBean.go
wxing1292: tried to use client IDL (revert back to prev impl: frontend depends on public client idl)
this will break the redirection handler: although the definition contents are exactly the same, the package name are different. this leads to only 2 solution: either change the idl definition dependency of frontend handler, which will break a lot of code, or use the solution in this PR

created a issue #1414, so in the future, client idl and service idl will be in the same repo

PR #1410, opened 2019-01-24
Add metric and retryable blobstore clients

andrewjdawson2016:
This diff does the following:
- Adds metrics client wrapper around blobstore
- Add retryable client wrapper around blobstore
- Introduces retryable vs non-retryable errors for blobstore
- Update file based blobstore to use new error types


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -24,6 +24,14 @@ exception BadRequestError {
   1: required string message
 }
 
+exception BlobEncodingError {
```
samarabbas: Can we just use BadRequestError instead of adding these new error types?  BadRequestError is already treated as non-retryable throughout the codebase.

common/blobstore/filestore/client.go
```diff
@@ -35,121 +37,179 @@ const (
...
 	defer c.Unlock()
 
-	exists, err := directoryExists(bucketDirectory(c.storeDirectory, bucket))
-	if err != nil {
-		return err
-	}
-	if !exists {
+	bd := bucketDirectory(c.storeDirectory, bucket)
+	if exists, err := directoryExists(bd); err != nil {
+		c.logger.Errorf("Upload failed. Error: %v. Bucket: %v", err, bucket)
```
samarabbas: Do not use 'Errorf', this makes log searching very hard.  Use tags for dynamic fields like Error and Bucket.

common/blobstore/interface.go
```diff
@@ -22,15 +22,19 @@ package blobstore
...
+	"github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common/blobstore/blob"
 )
 
 var (
-	// ErrBlobNotExists indicates that requested blob does not exist
-	ErrBlobNotExists = errors.New("requested blob does not exist")
 	// ErrBucketNotExists indicates that requested bucket does not exist
-	ErrBucketNotExists = errors.New("requested bucket does not exist")
+	ErrBucketNotExists = &shared.BadRequestError{Message: "requested bucket does not exist"}
```
samarabbas: We usually use EntityNotExistError for these kind of errors?

common/blobstore/interface.go
```diff
@@ -22,15 +22,19 @@ package blobstore
...
+	// ErrBlobNotExists indicates that requested blob does not exist
+	ErrBlobNotExists = &shared.BadRequestError{Message: "requested blob does not exist"}
```
samarabbas: Can we use EntityNotExistError here too?

common/blobstore/interface.go
```diff
@@ -22,15 +22,19 @@ package blobstore
...
+	// ErrBlobSerialization indicates that a failure occurred in serializing blob
+	ErrBlobSerialization = &shared.BlobEncodingError{Message: "failed to serialize blob"}
```
samarabbas: Do you need any special handling of BlobEncodingError?  If the intent is only to not treat it is Non-Retryable error then I would recommend to use BadRequestError here.

common/metrics/defs.go
```diff
@@ -713,6 +726,13 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		PersistenceCompleteForkBranchScope:                       {operation: "CompleteForkBranch", tags: map[string]string{ShardTagName: NoneShardsTagValue}},
 		PersistenceGetHistoryTreeScope:                           {operation: "GetHistoryTree", tags: map[string]string{ShardTagName: NoneShardsTagValue}},
 
+		BlobstoreClientUploadScope:         {operation: "Upload"},
```
samarabbas: Can you also include CadenceRoleTag for all blob client operations.  This makes it easier to just create a dashboard for all blob store operations.

common/blobstore/metricClient.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (c *metricClient) Upload(ctx context.Context, bucket string, key blob.Key, blob *blob.Blob) error {
+	c.metricsClient.IncCounter(metrics.BlobstoreClientUploadScope, metrics.CadenceClientRequests)
+
+	sw := c.metricsClient.StartTimer(metrics.BlobstoreClientUploadScope, metrics.CadenceClientLatency)
+	err := c.client.Upload(ctx, bucket, key, blob)
+	sw.Stop()
+
+	if err != nil {
+		c.metricsClient.IncCounter(metrics.BlobstoreClientUploadScope, metrics.CadenceClientFailures)
```
samarabbas: Can you also emit a metric per type of error so we can also get a breakdown of different kind of errors happening for various BlobStoreClient api calls.

andrewjdawson2016: I am unsure about this one for a few reasons let me know what you think 

- We do not do this in our other clients: frontend, admin, history. I like following the same pattern in blobstore.
- Some of the error types will be private repo types (e.g. lets say we cannot turn a key into a terrablob path because it breaks some constraint of terrablob). Those error types cannot be included here.
- In the happy path very few errors should be occurring if we start to see lots of errors here we will go to the logs and dive deeper into the specific errors we are getting.

samarabbas: We are actually planning to emit error type metric for other clients too.  Absence of this metric caused a servicing headache recently as it could have made things simpler to debug.  
Having said that, I'm also ok if you decide to do this later once we convert all metric clients to emit error type metric.

andrewjdawson2016: Okay for now I will leave it as is. But let's chat about how to do this in the future once some of the errors are private error types surfaced from Terrablob.

PR #1416, opened 2019-01-28
Schedule timer task to send signal to archival system workflow

andrewjdawson2016:


common/persistence/dataInterfaces.go
```diff
@@ -396,6 +397,16 @@ type (
 		Version             int64
 	}
 
+	// ArchiveHistoryEventTask identifies a timer task for archival of history events of completed execution.
+	ArchiveHistoryEventTask struct {
+		VisibilityTimestamp time.Time
+		TaskID              int64
+		Version             int64
+		TargetDomainID      string
```
samarabbas: You do not need to explicitly set DomainID, WorkflowID and RunID on timer tasks if they are created for same workflow execution.  TargetDomainID, TargetWorkflowID and TargetRunID are needed only for scenario where task is suppose to perform an action on another execution like transfer task for ChildWorkflowExecution processing.

service/history/timerQueueProcessorBase.go
```diff
@@ -578,6 +585,35 @@ func (t *timerQueueProcessorBase) processDeleteHistoryEvent(task *persistence.Ti
...
+	if err != nil {
+		return err
+	}
+	defer func() { release(retError) }()
+
+	msBuilder, err := loadMutableStateForTimerTask(context, task, t.metricsClient, t.logger)
+	if err != nil {
+		return err
+	} else if msBuilder == nil || msBuilder.IsWorkflowExecutionRunning() {
+		// this can happen if workflow is reset.
```
samarabbas: I don't believe the comment is correct.  Reset always create a new run, we never reuse the run when doing reset.

service/history/timerQueueProcessorBase.go
```diff
@@ -578,6 +585,35 @@ func (t *timerQueueProcessorBase) processDeleteHistoryEvent(task *persistence.Ti
...
+		return nil
+	}
+	ok, err := verifyTaskVersion(t.shard, t.logger, task.DomainID, msBuilder.GetLastWriteVersion(), task.Version, task)
+	if err != nil {
+		return err
+	} else if !ok {
+		return nil
+	}
+	return t.historyService.archivalClient.Archive(&sysworkflow.ArchiveRequest{
+		DomainID:         task.DomainID,
```
samarabbas: Looks like you are not even using TargetDomainID, TargetWorkflowID and TargetRunID during processing of the task.  You can safely remove them.

PR #1441, opened 2019-02-06
Archival workflow implementation 

andrewjdawson2016:
This diff does the following things

- Adds a public client metrics wrapper. This is somewhat overkill because our client already emits metrics but it is nice to have this symmetry with our other clients.
- Deletes mutable state and sends signal to archival system workflow in timerQueueProcessorBase.go
- Refactor worker/service.go (it had gotten messy with doing three different things).
- Add a history_blob_iterator which knows how to read history and deterministically construct blobs.
- Implement system workflow with retries, logging and metrics. 

There are two things that are not done yet but will be done as part of this diff
- In the header of the blob the algorithm used to construct the blob should be included
- Unit tests for system_workflow and history_blob_iterator

Something to pay attention to is, I implemented upload and delete as separate activities. I think logically this is the correct thing to do. But I am not sure if it will create too much performance overhead.


service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+			metricsClient.IncCounter(metrics.ArchivalUploadActivityScope, metrics.SysWorkerNextBlobNonRetryableFailures)
+			return errArchivalUploadActivityNextBlob
+		}
+		key, err := NewHistoryBlobKey(request.DomainID, request.WorkflowID, request.RunID, *historyBlob.Header.CurrentPageToken)
+		if err != nil {
+			logger.WithError(err).Error("failed to construct blob key, stopping archival upload")
+			metricsClient.IncCounter(metrics.ArchivalUploadActivityScope, metrics.SysWorkerKeyConstructionFailures)
+			return errArchivalUploadActivityConstructKey
+		}
+		if exists, err := blobExistsRetryForever(blobstoreClient, bucket, key); err != nil {
```
venkat1109: If this retries forever, you want to make sure you heartbeat periodically to avoid an activity heartbeat timeout

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+		if container.Config.EnableArchivalCompression(domainCacheEntry.GetInfo().Name) {
+			wrapFunctions = append(wrapFunctions, blob.GzipCompressed())
+		}
+		currBlob, err := blob.Wrap(blob.NewBlob(body, tags), wrapFunctions...)
+		if err != nil {
+			logger.WithError(err).Error("failed to wrap blob, stopping archival upload")
+			metricsClient.IncCounter(metrics.ArchivalUploadActivityScope, metrics.SysWorkerWrapBlobFailures)
+			return errArchivalUploadActivityWrapBlob
+		}
+		if err := blobUploadRetryForever(blobstoreClient, bucket, key, currBlob); err != nil {
```
venkat1109: same comment as above on retrying forever

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+		return err
+	}
+	for err != nil && common.IsPersistenceTransientError(err) {
+		err = backoff.Retry(op, common.CreatePersistanceRetryPolicy(), common.IsPersistenceTransientError)
+	}
+	return result, err
+}
+
+func blobExistsRetryForever(blobstoreClient blobstore.Client, bucket string, key blob.Key) (bool, error) {
+	exists, err := blobstoreClient.Exists(context.Background(), bucket, key)
```
venkat1109: you want to have some reasonable timeout set on the context instead of passing a block forever context

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+	for err != nil && common.IsBlobstoreTransientError(err) {
+		// blobstoreClient is already retryable so no extra retry/backoff logic is needed here
+		exists, err = blobstoreClient.Exists(context.Background(), bucket, key)
+	}
+	return exists, err
+}
+
+func blobUploadRetryForever(blobstoreClient blobstore.Client, bucket string, key blob.Key, blob *blob.Blob) error {
+	err := blobstoreClient.Upload(context.Background(), bucket, key, blob)
```
venkat1109: same comment as above on context timeout for blobstore calls

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+		}
+		op := func() error {
+			return persistence.DeleteWorkflowExecutionHistoryV2(container.HistoryV2Manager, request.BranchToken, container.Logger)
+		}
+		for err != nil && common.IsPersistenceTransientError(err) {
+			err = backoff.Retry(op, common.CreatePersistanceRetryPolicy(), common.IsPersistenceTransientError)
+		}
+		logger.WithError(err).Error("failed to delete history from events v2")
+		metricsClient.IncCounter(metrics.ArchivalDeleteHistoryActivityScope, metrics.SysWorkerDeleteHistoryV2NonRetryableFailures)
+		return errDeleteHistoryActivityDeleteFromV2
```
venkat1109: - i believe this error is declared as non-retryable in the activity retry policy options. However, your backoff.Retry() will give up all retries in about 30s. Isn't this the only place where history gets deleted ? If yes, you should retry infinitely

- activity options has heartbeat timeout set to 10s - if you are midst of a retry loop, you want to avoid timing out on heartbeat

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+	}
+	op := func() error {
+		return container.HistoryManager.DeleteWorkflowExecutionHistory(deleteHistoryReq)
+	}
+	for err != nil && common.IsPersistenceTransientError(err) {
+		err = backoff.Retry(op, common.CreatePersistanceRetryPolicy(), common.IsPersistenceTransientError)
+	}
+	logger.WithError(err).Error("failed to delete history from events v1")
+	metricsClient.IncCounter(metrics.ArchivalDeleteHistoryActivityScope, metrics.SysWorkerDeleteHistoryV1NonRetryableFailures)
+	return errDeleteHistoryActivityDeleteFromV1
```
venkat1109: same comment as above

client/public/metricClient.go
```diff
@@ -0,0 +1,583 @@
...
+
+var _ Client = (*metricClient)(nil)
+
+type metricClient struct {
+	client        Client
+	metricsClient metrics.Client
+}
+
+// NewMetricClient creates a new instance of Client that emits metrics
+func NewMetricClient(client Client, metricsClient metrics.Client) Client {
```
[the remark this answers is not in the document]
andrewjdawson2016: I agree, however, the other clients: frontend, history, and matching all follow this naming pattern. I wanted to follow the existing pattern.

service/history/timerQueueProcessorBase.go
```diff
@@ -604,12 +604,34 @@ func (t *timerQueueProcessorBase) processArchiveHistoryEvent(task *persistence.T
...
+	if err := t.historyService.archivalClient.Archive(req); err != nil {
+		t.logger.WithFields(bark.Fields{
+			logging.TagHistoryShardID:      t.shard.GetShardID(),
+			logging.TagTaskID:              task.GetTaskID(),
+			logging.TagTaskType:            task.GetTaskType(),
+			logging.TagDomainID:            task.DomainID,
+			logging.TagWorkflowExecutionID: task.WorkflowID,
+			logging.TagWorkflowRunID:       task.RunID,
+			logging.TagErr:                 err,
+		}).Error("failed to initiate archival")
```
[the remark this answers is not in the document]
andrewjdawson2016: Can you elaborate on what you mean by this?

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
+		logging.TagArchiveRequestLastFirstEventID:  request.LastFirstEventID,
+	})
+	metricsClient := container.MetricsClient
+	domainCache := container.DomainCache
+	clusterMetadata := container.ClusterMetadata
+	domainCacheEntry, err := domainCache.GetDomainByID(request.DomainID)
+	if err != nil {
+		logger.WithError(err).Error("failed to get domain from domain cache")
+		metricsClient.IncCounter(metrics.ArchivalUploadActivityScope, metrics.SysWorkerGetDomainFailures)
+		return errArchivalUploadActivityGetDomain
```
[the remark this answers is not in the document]
andrewjdawson2016: Let's chat in person about this comment. What is needed to handle this properly is for there to be a retryable domainCache client, that way handling these transient errors can be done the same way as all other transient errors.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,341 @@ package sysworkflow
...
 }
 
-func selectSystemTask(scope tally.Scope, signal signal, ctx workflow.Context, logger *zap.Logger) {
-	scope.Counter(HandledSignalCount).Inc(1)
-
+func selectSystemTask(signal signal, ctx workflow.Context, logger bark.Logger, metricsClient metrics.Client) {
 	ao := workflow.ActivityOptions{
 		ScheduleToStartTimeout: time.Minute,
 		StartToCloseTimeout:    time.Minute,
 		HeartbeatTimeout:       time.Second * 10,
```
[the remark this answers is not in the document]
andrewjdawson2016: I am not sure I understand. If you do not heartbeat for 10 seconds the activity will time out and be retried. When you try to upload a blob you will always retry forever as long as you are getting transient errors. So if you get transient errors for too long then this heartbeat will kick in and kill your activity, at which point the activity will be retried from the start again.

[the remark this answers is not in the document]
andrewjdawson2016: Basically the way I thought about it is, the activity retries everything it can forever, and heartbeat acts as a catch all to say "if you have not made enough forward progress in last 10 seconds, then start all the way over"

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Same response as above. It seems like we are thinking about heartbeating differently. Let's chat in person about the right model here.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: yup

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: - I am retrying indefinitely. The backoff.Retry() is inside of a loop that continues as long as error type is transient. 
- On the heartbeat time out issue. I see what you mean. I think it's actually related to how we decide to handle heartbeats more generally.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: same response.

service/worker/sysworkflow/history_blob_iterator.go
```diff
@@ -0,0 +1,218 @@
...
+		metricsClient metrics.Client
+
+		// the following defines the state of the iterator
+		blobPageToken        int
+		persistencePageToken []byte
+		finishedIteration    bool
+
+		// the following are only used to read history and dynamic config
+		historyManager       persistence.HistoryManager
+		historyV2Manager     persistence.HistoryV2Manager
```
samarabbas: Can we have 2 separate implementations of historyBlobIterator?  This will make things easier to remove any code related to historyV1Manager.

service/worker/sysworkflow/constants.go
```diff
@@ -24,62 +24,45 @@ import (
...
+	signalsUntilContinueAsNew           = 1000
+	systemWorkflowFnName                = "SystemWorkflow"
+	archivalUploadActivityFnName        = "ArchivalUploadActivity"
+	archivalDeleteHistoryActivityFnName = "ArchivalDeleteHistoryActivity"
+	backfillActivityFnName              = "BackfillActivity"
+	historyBlobKeyExt                   = "history"
+	blobstoreOperationsDefaultTimeout   = 5 * time.Second
+
+	// the following are all non-retryable error strings
+	errArchivalUploadActivityGetDomainStr           = "failed to get domain from domain cache"
```
samarabbas: Instead of strings we usually define actual errors which can be used in the code.

service/worker/sysworkflow/history_blob_iterator.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Next returns historyBlob and advances iterator.
+// Returns error if iterator is empty, or if history could not be read.
+// If error is returned then no iterator is state is advanced.
+func (i *historyBlobIterator) Next() (*HistoryBlob, error) {
+	if !i.HasNext() {
+		i.logger.Error("called Next on depleted iterator")
+		i.metricsClient.IncCounter(metrics.HistoryBlobIteratorScope, metrics.SysWorkerNextOnDepletedIterator)
+		return nil, errors.New("iterator is depleted")
```
samarabbas: Can you return a well defined error instead.

service/worker/sysworkflow/history_blob_iterator.go
```diff
@@ -0,0 +1,218 @@
...
+func NewHistoryBlobIterator(
+	logger bark.Logger,
+	metricsClient metrics.Client,
+	request ArchiveRequest,
+	container *SysWorkerContainer,
+	domainName string,
+	clusterName string,
+) HistoryBlobIterator {
+	return &historyBlobIterator{
+		logger:        logger,
```
samarabbas: Can you make sure the logger is tagged with useful tags like domain, workflowId, runId, etc

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
 // SystemWorkflow is the system workflow code
 func SystemWorkflow(ctx workflow.Context) error {
-	id := workflow.GetInfo(ctx).WorkflowExecution.ID
-	logger := workflow.GetLogger(ctx)
-	scope := workflow.GetMetricsScope(ctx).Tagged(map[string]string{SystemWorkflowIDTag: id})
-	ch := workflow.GetSignalChannel(ctx, SignalName)
-
-	logger.Info("started new system workflow")
+	sysWorkflowInfo := workflow.GetInfo(ctx)
+	logger := globalLogger.WithFields(bark.Fields{
```
samarabbas: Is this logger replay aware?

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
+		logging.TagWorkflowExecutionID: sysWorkflowInfo.WorkflowExecution.ID,
+		logging.TagWorkflowRunID:       sysWorkflowInfo.WorkflowExecution.RunID,
+	})
+	logger.Info("started system workflow")
+	globalMetricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerWorkflowStarted)
```
samarabbas: Is this metricClient replay aware? I would recommend both logger and metric client in a separate type which is replay aware.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
+	}).Info("system workflow is continuing as new")
+	return workflow.NewContinueAsNewError(ctx, systemWorkflowFnName)
 }
 
-func selectSystemTask(scope tally.Scope, signal signal, ctx workflow.Context, logger *zap.Logger) {
-	scope.Counter(HandledSignalCount).Inc(1)
-
+func selectSystemTask(signal signal, ctx workflow.Context, logger bark.Logger, metricsClient metrics.Client) {
 	ao := workflow.ActivityOptions{
 		ScheduleToStartTimeout: time.Minute,
```
samarabbas: Is this the timeout for archival of entire workflow execution history? Do you think 1 minute is good enough for all use cases?  I think currently we are allowing histories to grow as big as 200MB.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
+				logging.TagArchiveRequestWorkflowID:        request.WorkflowID,
+				logging.TagArchiveRequestRunID:             request.RunID,
+				logging.TagArchiveRequestEventStoreVersion: request.EventStoreVersion,
+				logging.TagArchiveRequestLastFirstEventID:  request.LastFirstEventID,
+			}).Error("ArchivalUploadActivity encountered non-retryable error")
+			metricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerArchivalUploadActivityNonRetryableFailures)
+		} else {
+			metricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerArchivalUploadSuccessful)
+		}
+		if err := workflow.ExecuteActivity(
```
samarabbas: Delete history is usually one API call to persistence layer.  Can this be converted to Local Activity.

samarabbas: If the upload keeps on failing repeatedly and exhaust retries, do we still delete workflow execution history?

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
+	for ; signalsHandled < signalsUntilContinueAsNew; signalsHandled++ {
 		var signal signal
 		if more := ch.Receive(ctx, &signal); !more {
-			scope.Counter(ChannelClosedUnexpectedlyError).Inc(1)
-			logger.Error("cadence channel was unexpectedly closed")
 			break
 		}
-		selectSystemTask(scope, signal, ctx, logger)
+		globalMetricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerReceivedSignal)
+		selectSystemTask(signal, ctx, logger, globalMetricsClient)
```
samarabbas: Can we do this in its own goroutine? Otherwise each archival will be serialized on after another.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,101 +22,337 @@ package sysworkflow
...
 	}
 
 	for {
 		var signal signal
 		if ok := ch.ReceiveAsync(&signal); !ok {
 			break
 		}
-		selectSystemTask(scope, signal, ctx, logger)
+		globalMetricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerReceivedSignal)
+		selectSystemTask(signal, ctx, logger, globalMetricsClient)
```
samarabbas: This could potentially cause you to miss signals.  It is unsafe to do any processing of signals after draining the signal pump.  I think the safest thing to do is pass all those signals to new run for processing.

service/worker/sysworkflow/history_blob_iterator.go
andrewjdawson2016: We can. There were a couple reasons I did not do this in the the first version let me know what you think:

- The complexity of handling events_v1 and events_v2 is all isolated to one method in this module, so the transition to events_v2 should not be hard. 
- If we want to have two separate iterators the code is going to be more complex, we need to factor common code out into a base class and have two subclasses each of which basically only reads history and then call into the base class.

service/worker/sysworkflow/constants.go
andrewjdawson2016: In this case I define both strings and errors that way I can reference these strings from the activity non-retryable errors config.

service/worker/sysworkflow/history_blob_iterator.go
andrewjdawson2016: Yes it is, the logger we use here is the same logger that we tag in the activity which has these useful tags defined.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: It was not, but I fixed this. Thanks.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Fixed this, let me know if it was fixed in a way you are okay with.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Good catch. I increased this to 5 minutes to be on the safe side.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Converted it to local activity. I assume that as long as I do retries of transient errors in the activity then this will be guaranteed to delete history still. Lets chat in person about this assumption.

andrewjdawson2016: The behavior is that that all transient errors will be retried forever. If a non-transisent error is received that the upload activity is aborted and history is still deleted. History is never deleted until a non-transisent error occurs.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Let's also chat about this one over zoom.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Can we chat about this one over zoom, I want to understand this better.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,103 +22,414 @@ package sysworkflow
...
+	errArchivalUploadActivityWrapBlob            = cadence.NewCustomError(errArchivalUploadActivityWrapBlobStr)
+	errArchivalUploadActivityUploadBlob          = cadence.NewCustomError(errArchivalUploadActivityUploadBlobStr)
+	errDeleteHistoryActivityDeleteFromV2         = cadence.NewCustomError(errDeleteHistoryActivityDeleteFromV2Str)
+	errDeleteHistoryActivityDeleteFromV1         = cadence.NewCustomError(errDeleteHistoryActivityDeleteFromV1Str)
+)
+
+// ArchiveSystemWorkflow is the system workflow which archives and deletes history
+func ArchiveSystemWorkflow(ctx workflow.Context, carryoverRequests []ArchiveRequest) error {
+	sysWorkflowInfo := workflow.GetInfo(ctx)
+	isReplay := workflow.IsReplaying(ctx)
```
samarabbas: This won't work.  You need to pass workflow context to the logger and metric wrapper client.  This need to specifically call workflow.IsReplaying(ctx) each time before emitting the log or metric.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,103 +22,414 @@ package sysworkflow
...
-		var signal signal
-		if ok := ch.ReceiveAsync(&signal); !ok {
+		var request ArchiveRequest
+		if ok := ch.ReceiveAsync(&request); !ok {
 			break
 		}
-		selectSystemTask(scope, signal, ctx, logger)
-		signalsHandled++
+		metricsClient.IncCounter(metrics.SystemWorkflowScope, metrics.SysWorkerReceivedSignal)
+		co = append(co, request)
```
samarabbas: drain needs to happen after waiting for all progress work to finish.

service/worker/sysworkflow/system_workflow.go
```diff
@@ -22,103 +22,414 @@ package sysworkflow
...
+				errArchivalUploadActivityConstructKeyStr,
+				errArchivalUploadActivityBlobExistsStr,
+				errArchivalUploadActivityMarshalBlobStr,
+				errArchivalUploadActivityConvertHeaderToTagsStr,
+				errArchivalUploadActivityWrapBlobStr,
+				errArchivalUploadActivityUploadBlobStr,
+			},
+		},
+	}
+	if err := workflow.ExecuteActivity(
```
samarabbas: Is there a way to interrupt long running activities if it is time to continueAsNew and one or more activities are continuously in retry.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Thanks for catching this.

service/worker/sysworkflow/system_workflow.go
andrewjdawson2016: Will address in text CR.

PR #1446, opened 2019-02-08
sql: schema perf optimizations

venkat1109:
This patch contains schema changes along with the corresponding code changes to improve performance of SQL persistence. Changes to schema are:
 - Store primary keys which are uuids as BINARY(16) instead of char or varchar
 - Enforce consistency across tables on usage of char/varchar(len) for same type 

The proposed changes are based on results from a benchmark performed against MySQL 5.7 in GTID mode.


common/persistence/sql/storage/sqldb/uuid.go
```diff
@@ -0,0 +1,93 @@
...
+// UUID represents a 16-byte universally unique identifier
+// this type is a wrapper around google/uuid with the following differences
+//  - type is a byte slice instead of [16]byte
+//  - db serialization converts uuid to bytes as opposed to string
+type UUID []byte
+
+// NewUUIDFromString returns a new UUID parsed from the given string
+// returns nil if the input is empty string
+// panics if the given input is malformed
+func NewUUIDFromString(s string) UUID {
```
mfateev: I don't like this name as it looks like it generates UUID from any string. 
I would just use the same name as uuid package: MustParse.

schema/mysql/v56/cadence/schema.sql
```diff
@@ -64,17 +64,17 @@ CREATE TABLE transfer_tasks(
 
 CREATE TABLE executions(
   shard_id INT NOT NULL,
-	domain_id CHAR(64) NOT NULL,
-	workflow_id VARCHAR(255) NOT NULL,
-	run_id CHAR(64) NOT NULL,
+	domain_id BINARY(16) NOT NULL,
+	workflow_id VARCHAR(64) NOT NULL,
```
mfateev: I don't think we can limit workflow_id to 64 bytes. This is too low for some use cases.

schema/mysql/v56/cadence/schema.sql
```diff
@@ -319,27 +319,26 @@ PRIMARY KEY (shard_id, domain_id, workflow_id, run_id, initiated_id)
...
-	run_id VARCHAR(64) NOT NULL,
+	domain_id BINARY(16) NOT NULL,
+	workflow_id VARCHAR(64) NOT NULL,
+	run_id BINARY(16) NOT NULL,
 	signal_id VARCHAR(64) NOT NULL,
 	--
 	PRIMARY KEY (shard_id, domain_id, workflow_id, run_id, signal_id)
-);
-
+);
```
mfateev: new line missing

common/persistence/sql/storage/sqldb/uuid.go
venkat1109: 3 options:
 (1) Either I create a sqldb/uuid subpackage and name the method MustParse - so for caller, it will look like uuid.MustParse() - same as uuid package
(2) Or name this method sqldb.ParseUUID(s string) - sqldb.MustParse() is weird
(3) Or name this method sqldb.StringToUUID(s string)

I am inclined towards option (2) - because it avoids import namespace collisions in other parts of code

schema/mysql/v56/cadence/schema.sql
venkat1109: ok. changed it back to 255

common/persistence/sql/storage/sqldb/uuid.go
venkat1109: renamed to MustParseUUID()

PR #1460, opened 2019-02-13
sql: add support for workflow retry params

### LINKED ISSUE #1339: fix sql to support workflow retry
looks like it is missing some meta data from persistence layer for sql to support workflow retry.

Reference: server side workflow retry implementation
https://github.com/uber/cadence/commit/c18338477f1b56f87569cc239b10febbd69f682e

Below are missing fields for sql (copy from Cassandra schema update):
ALTER TYPE workflow_execution ADD attempt              int;
ALTER TYPE workflow_execution ADD has_retry_policy     boolean;
ALTER TYPE workflow_execution ADD init_interval        int;
ALTER TYPE workflow_execution ADD backoff_coefficient  double;
ALTER TYPE workflow_execution ADD max_interval         int;
ALTER TYPE workflow_execution ADD expiration_time      timestamp;
ALTER TYPE workflow_execution ADD max_attempts         int;
ALTER TYPE workflow_execution ADD non_retriable_errors list<text>;

venkat1109:
- Fixes #1339
- Also contains improvements to test coverage and related fixes

Note to Reviewer:
The 'executions' table at this point has close to 50 columns. We could consider storing the mutable state as a blob in MySQL - however the pros/cons of doing so are not very clear at this time. So, I have decided to tackle this problem independently after getting some data on the performance implications of such a change.

Pros/Cons to having too many columns are below. The I/O cost is likely to be the same with both blob and wide rows

- Cons
  - CPU cost on the MySQL server to do conversions to/from blobs obtained from InnoDB
  - Mutable state (executions) is something we expect to add/deprecate columns frequently. In a sharded mysql cluster,  schema changes may be hard i.e. all storage shards must agree on the new schema despite operating independently - this may or may not add operational complexity

- Pros
  - Allows for joins on non-primary key fields - but we likely don't need this flexibility

### DIFF
--- a/common/persistence/persistence-tests/executionManagerTest.go
+++ b/common/persistence/persistence-tests/executionManagerTest.go
@@ -32,6 +32,10 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 
+	"math"
+
+	"math/rand"
+
 	gen "github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/cluster"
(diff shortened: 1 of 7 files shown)


common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -32,6 +32,10 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 
+	"math"
```
mfateev: Move above

venkat1109: done

Issue #1436, opened 2019-02-06
sql: fix gaps in persistence API wrt deletion of state


venkat1109:
Historically, cadence relied on cassandra's TTL mechanism to delete state in persistence for the following:
 - Workflow Execution
 - Tasks - activity_tasks, decision_tasks, sticky_tasks etc

Since MySQL does not have an equivalent of cassandra's TTL, these deletes need to be supported at the application layer. Following is a proposed approach for dealing with this:

**Workflow Execution Deletion**
Today, we have a timer_task associated with every completed workflow to cleanup state after retention. This timer_task already calls DeleteWFExecution API on the persistence layer. For SQL, this API also needs to delete the row from current_executions table.

**Task Deletion**
Today, there is a timeout associated with every task created. However, on a timeout, the timer_task handler does not delete the task entry from persistence today. There is assumption that cassandra's TTL mechanism will delete the task automatically after the timeout. For SQL, following changes needs to happen:
 - timer_task handler should call DeleteTask API to delete the task entry
 - ListTasks API must filter out expired tasks so that application only gets valid tasks

Cassandra will continue to use the TTL mechanism for backwards compatibility and noop the DeleteTask API.

PR #1461, opened 2019-02-13
sql: delete current_executions row as part of workflow deletion

venkat1109:
Problem:
As of today, we don't delete rows from current_executions table even after workflow retention expires. This is because, historically, we relied on cassandra's TTL mechanism to delete this record. Since SQL does not have a TTL equivalent, this patch contains a change to cleanup current_executions state after workflow retention. 

Fix: We currently have a timer_task that fires after retention to delete history/mutable state. This handler calls the persistence DeleteWFExecution API. This patch changes the behavior of this API (for SQL) to delete both mutable state and current_executions entry in a single transaction. The current_executions delete is a conditional delete because its possible for a new run to have started.

Fixes parts of #1436. 

Note to Reviewer:
One thing I don't like about this change is that - the behavior of DeleteWFExecution is different for SQL and Cassandra. This is because TTL is still the best way to delete the current_executions row for cassandra and it makes sense to leave the existing behavior as such. This behavior difference is transparent to rest of cadence. However, persistence layer implementations need to be aware that - 
  (1) if store supports TTL, its best for delete to happen in UpdateWorkflowExecution using TTL
  (2) if store doesn't support TTL, the delete needs to happen in DeleteWFExecution

I added a unit test that creates a workflow, updates it (completion) and then calls deleteWFExecution. After this, the test verifies all state is deleted.

### DIFF
--- a/common/persistence/persistence-tests/executionManagerTest.go
+++ b/common/persistence/persistence-tests/executionManagerTest.go
@@ -821,7 +821,7 @@ func (s *ExecutionManagerSuite) TestDeleteWorkflow() {
 // TestDeleteCurrentWorkflow test
 func (s *ExecutionManagerSuite) TestDeleteCurrentWorkflow() {
 	if s.ExecutionManager.GetName() != "cassandra" {
-		s.T().Skip("SQL doesn't support retention yet")
+		s.T().Skip("this test is only applicable for cassandra (uses TTL based deletes)")
 	}
 	finishedCurrentExecutionRetentionTTL := int32(3) // 3 seconds
 	domainID := "54d15308-e20e-4b91-a00f-a518a3892790"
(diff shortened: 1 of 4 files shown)


common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -897,14 +951,14 @@ func (s *ExecutionManagerSuite) TestTransferTasksThroughUpdate() {
 	s.Equal(p.TransferTaskTypeCloseExecution, task3.TaskType)
 	s.Equal("", task3.TargetRunID)
 
-	err8 := s.DeleteWorkflowExecution(info1)
+	err8 := s.CompleteTransferTask(task3.TaskID)
 	s.NoError(err8)
 
-	err9 := s.CompleteTransferTask(task3.TaskID)
-	s.NoError(err9)
+	_, err9 := s.CreateWorkflowExecution(domainID, newExecution, "queue1", "wType", 20, 13, nil, 3, 0, 2, nil)
```
mfateev: Add comment explaining why error is expected. Or change string on the assertion to something more meaningful.

common/persistence/sql/storage/mysql/execution.go
```diff
@@ -422,7 +423,7 @@ func (mdb *DB) SelectFromCurrentExecutions(filter *sqldb.CurrentExecutionsFilter
 
 // DeleteFromCurrentExecutions deletes a single row in current_executions table
```
mfateev: Add comment about meaning of the runID

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -596,7 +597,11 @@ type (
 
 		InsertIntoCurrentExecutions(row *CurrentExecutionsRow) (sql.Result, error)
 		UpdateCurrentExecutions(row *CurrentExecutionsRow) (sql.Result, error)
+		// SelectFromCurrentExecutions returns one or more rows from current_executions table
+		// Required params - {shardID, domainID, workflowID}
 		SelectFromCurrentExecutions(filter *CurrentExecutionsFilter) (*CurrentExecutionsRow, error)
+		// DeleteFromCurrentExecutions performa a conditinal delete of a row in current_executions table
+		// Required params - {shardID, domainID, workflowID, runID}
```
mfateev: From the doc it is not clear what happens when condition doesn't hold. Does it return an error or doesn't delete and returns OK.

venkat1109: you are right - let me update the doc - if the condition doesn't hold - it simply returns ok

PR #1468, opened 2019-02-14
Refactor duplicate code in CLI

vancexu:
There are some duplicate code in CLI which will increase difficulty when adding features, such as people who add Cron feature to `start` command might forget to add same thing to `run`
This PR includes:
1. Remove duplicates in flags definitions
2. Remove duplicates in command start and run


tools/cli/flags.go
```diff
@@ -0,0 +1,381 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yux0: Should we update header to the correct year?

PR #1474, opened 2019-02-17
Support GetWorkflowExecutionHistory API from archival

andrewjdawson2016:
This diff does the following:

- Adds support to frontend to get archived histories & unit tests.
- Changed name of from SystemWorkflow to ArchiveSystemWorkflow and change all related naming.
- Fix bug in archival signal: use NextEventID instead of LastFirstEventID. 
- Make history blob metadata for page tokens be simple ints instead of strings.

This diff does not support getting history from archival unless domainID, workflowID AND runID are given. Supporting the API without giving runID will be left as a task until after indices are being archived.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -25,7 +25,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"go.uber.org/cadence/.gen/go/shared"
 	"sync"
 	"time"
 
@@ -38,6 +37,7 @@ import (
 	h "github.com/uber/cadence/.gen/go/history"
 	m "github.com/uber/cadence/.gen/go/matching"
 	"github.com/uber/cadence/.gen/go/replicator"
+	"github.com/uber/cadence/.gen/go/shared"
 	gen "github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/client/history"
 	"github.com/uber/cadence/client/matching"
@@ -3132,7 +3132,7 @@ func (wh *WorkflowHandler) historyArchived(ctx context.Context, request *gen.Get
 	}
 	getMutableStateRequest := &h.GetMutableStateRequest{
 		DomainUUID: common.StringPtr(domainID),
-		Execution: request.Execution,
+		Execution:  request.Execution,
 	}
 	_, err := wh.history.GetMutableState(ctx, getMutableStateRequest)
 	if err == nil {
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -25,6 +25,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/uber/cadence/common/blobstore/blob"
+	"github.com/uber/cadence/service/worker/sysworkflow"
 	"sync"
 	"time"
 
@@ -87,6 +89,10 @@ type (
 		BranchToken       []byte
 		ReplicationInfo   map[string]*gen.ReplicationInfo
 	}
+
+	getHistoryContinuationTokenArchival struct {
+		BlobstorePageToken int
+	}
 )
 
 var (
@@ -111,6 +117,7 @@ var (
 	errWorkflowTypeNotSet                         = &gen.BadRequestError{Message: "WorkflowType is not set on request."}
 	errInvalidExecutionStartToCloseTimeoutSeconds = &gen.BadRequestError{Message: "A valid ExecutionStartToCloseTimeoutSeconds is not set on request."}
 	errInvalidTaskStartToCloseTimeoutSeconds      = &gen.BadRequestError{Message: "A valid TaskStartToCloseTimeoutSeconds is not set on request."}
+	errDomainArchivalBucketNotSet                 = &gen.BadRequestError{Message: "Domain config does not have archival bucket set."}
 
 	// err for string too long
 	errDomainTooLong       = &gen.BadRequestError{Message: "Domain length exceeds limit."}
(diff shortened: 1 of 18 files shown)


service/frontend/workflowHandler.go
```diff
@@ -1860,6 +1867,10 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		getRequest.MaximumPageSize = common.Int32Ptr(common.GetHistoryMaxPageSize)
 	}
 
+	if wh.historyArchived(ctx, getRequest, domainID) {
```
samarabbas: This results into extra call to history service for every GetWorkflowExecutionHistory invocation.  It would be good to consolidate this behind under single GetMutableState call.

service/frontend/workflowHandler.go
```diff
@@ -3098,3 +3124,73 @@ func (wh *WorkflowHandler) bucketName(customBucketName *string) string {
...
+func (wh *WorkflowHandler) historyArchived(ctx context.Context, request *gen.GetWorkflowExecutionHistoryRequest, domainID string) bool {
+	if request.GetExecution() == nil || request.GetExecution().GetRunId() == "" {
+		return false
+	}
+	describeMutableStateRequest := &h.DescribeMutableStateRequest{
+		DomainUUID: common.StringPtr(domainID),
+		Execution:  request.Execution,
+	}
+	resp, err := wh.history.DescribeMutableState(ctx, describeMutableStateRequest)
+	return err != nil || resp.GetMutableStateInDatabase() == ""
```
samarabbas: Current mental model is DescribeMutableState (or GetMutableState) should fail with EntityNotFoundError.  Current implementation seems to make some weird checks to handle mutableState not deleted from cache scenario.  I would rather fix the actual issue to remove mutable state from cache too during processing of retention timer instead of explicitly checking 'GetMutableStateInDatabase'.

service/frontend/workflowHandler.go
andrewjdawson2016: Did not address will explain offline why.

service/frontend/workflowHandler.go
andrewjdawson2016: Talked in person about this. Got on the same page. I am now deleting the entry from the cache and fixed a bug in Release method of the lru cache. 

The system invariant is whenever we get EntityNotExistsError from GetMutableState we will assume that means history has been archived.

service/history/timerQueueProcessorBase.go
```diff
@@ -623,21 +624,21 @@ func (t *timerQueueProcessorBase) processArchiveHistoryEvent(task *persistence.T
...
 		WorkflowID:           task.WorkflowID,
 		RunID:                task.RunID,
 		EventStoreVersion:    msBuilder.GetEventStoreVersion(),
 		BranchToken:          msBuilder.GetCurrentBranch(),
-		LastFirstEventID:     msBuilder.GetLastFirstEventID(),
+		NextEventID:          msBuilder.GetNextEventID(),
 		CloseFailoverVersion: msBuilder.GetLastWriteVersion(),
 	}
+	key := definition.NewWorkflowIdentifier(task.DomainID, task.WorkflowID, task.RunID)
+	t.cache.Delete(key)
```
samarabbas: I think a safer approach is to remove execution from cache when we call release.  How about we create a new overload for 'getOrCreateWorkflowExecution' which returns a release function that takes in removeFromCache flag.

andrewjdawson2016: We can do that but it will result in somewhat messy code. There are currently a lot of places that call getOrCreateWorkflowExecutionWithTimeout. That is the method that actually creates the release function. To make this work you need to do one of the following not so nice things:

1. Duplicate all the logic that method into a new method with a different name which returns a release function that deletes from the cache when released.
2. Pass through a boolean everywhere that invokes this method then branch on this boolean to determine if you should delete from cache in release function.

I do not like either of these options from a code clarity perspective. Can you elaborate on the safety concern you have with the current approach.

andrewjdawson2016: I made the change but I do not like how the code looks. There are basically three ways I could think of to implement this

1. Add new versions of the existing methods that know how to delete when released. I don't like this because it creates some really weird method names (this is the option I coded).
2. Add parameter to existing methods and check this parameter to know if the release function should delete from cache once invoked. This is not great because these methods are called in 43 places and only in once case do we actually want to delete from cache. So passing false in 42 places does not feel right.
3. Use the former approach of just deleting directly from the cache while holding the lock. You mentioned a safety concern around this but I am not sure I see the safety concern. Can we chat about this more because I like the former approach the best.

samarabbas: I'm getting a little worried as I think about this more.  Our lru cache also has refCounting semantics a mutableStateCache (historyCache) heavily relies on making sure we use the same workflowExecutionContext for various correctness reasons.  For pinned caches, deleting items from the cache is actually very risky as it could lead to multiple objects representing the same key pinned by various api calls.  
Now thinking about this more, maybe we could just call workflowExecutionContext.clear() to remove mutable state rather than deleting item from the cache.  Sorry for not mentioning this earlier.  Let's discuss this more in person in the morning.

service/history/timerQueueProcessorBase.go
```diff
@@ -633,12 +632,11 @@ func (t *timerQueueProcessorBase) processArchiveHistoryEvent(task *persistence.T
 		NextEventID:          msBuilder.GetNextEventID(),
 		CloseFailoverVersion: msBuilder.GetLastWriteVersion(),
 	}
-	key := definition.NewWorkflowIdentifier(task.DomainID, task.WorkflowID, task.RunID)
-	t.cache.Delete(key)
 	err = t.deleteWorkflowExecution(task)
 	if err != nil {
 		return err
 	}
+	context.clear()
```
samarabbas: Can you put a comment why clear is called explicitly here.

PR #1475, opened 2019-02-19
Fix race for handling duplicate kafka message

vancexu:
Currently, when handle duplicate delete message in kafka, there is a race condition which will cause message being acked twice. 
This fix just ack the new message instead of old message.

### DIFF
--- a/service/worker/indexer/esProcessor.go
+++ b/service/worker/indexer/esProcessor.go
@@ -112,7 +112,8 @@ func (p *esProcessorImpl) Stop() {
 // Add an ES request, and an map item for kafka message
 func (p *esProcessorImpl) Add(request elastic.BulkableRequest, key string, kafkaMsg messaging.Message) {
 	if p.mapToKafkaMsg.Contains(key) { // handle duplicate delete message
-		p.ackKafkaMsg(key)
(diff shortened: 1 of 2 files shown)


service/worker/indexer/esProcessor.go
```diff
@@ -112,7 +112,8 @@ func (p *esProcessorImpl) Stop() {
 // Add an ES request, and an map item for kafka message
 func (p *esProcessorImpl) Add(request elastic.BulkableRequest, key string, kafkaMsg messaging.Message) {
 	if p.mapToKafkaMsg.Contains(key) { // handle duplicate delete message
-		p.ackKafkaMsg(key)
+		kafkaMsg.Ack()
```
dmetzgar: One thing that `ackKafkaMsg` does is it removes the key from `mapToKafkaMsg`. Is that intentionally left out?

vancexu: Yes. Say handle duplicate event e1,e2 in kafka of same key, when e2 comes and it found map contains key: Old version is ack and remove the e1, which has race condition that can cause e1 being acked twice; New version just ack e2.

PR #1489, opened 2019-02-24
Update archival config

andrewjdawson2016:
This diff does the following things:
- Updates cluster level archival config. Before archival was either enabled or disabled at the cluster level. Now archival can be enabled, paused or disabled. 
- Introduce an ArchivalConfig concept which enables invariants about the relationship between cluster's archival status and if default bucket is set. 
- Update domain level archival config such that there is no more never enabled state. Instead if a domain has never been enabled for archival it will be represented as disabled state with an empty bucket.
- Update the domainConfigStateMachine to accept more inputs: all requests are now considered valid and empty string for bucket name is treated the same way as nil.
- Combine archival timer task and delete timer task into a single timer task.
- Update delete logic to handle all possible archival states.


common/cluster/archival_status.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	return ac
+}
+
+// GetArchivalStatus converts input string to ArchivalStatus.
+// Valid input strings are "disabled", "paused", and "enabled".
+// If non-valid string is given status of Disabled is returned.
+func GetArchivalStatus(str string) ArchivalStatus {
+	switch str {
```
venkat1109: remove whitespaces and convert to lower case before doing comparison. This is input that comes from a human and its possible for typos, extra whitespaces etc

common/cluster/archival_status.go
```diff
@@ -0,0 +1,98 @@
...
+	case "disabled":
+		return ArchivalDisabled
+	case "paused":
+		return ArchivalPaused
+	case "enabled":
+		return ArchivalEnabled
+	default:
```
venkat1109: can you add a case str > 0 before default ? idea is to detect typos and error about it.

common/cluster/metadata.go
```diff
@@ -219,12 +215,14 @@ func (metadata *metadataImpl) GetAllClientAddress() map[string]config.Address {
...
-// GetDefaultArchivalBucket returns the default archival bucket name
-func (metadata *metadataImpl) GetDefaultArchivalBucket() string {
-	return metadata.defaultArchivalBucket
+// ArchivalConfig returns the archival config of the cluster.
+// The cluster is considered disabled for archival if either there is no default bucket or if config indicates disabled.
+// If status is paused or enabled then default bucket is set.
+func (metadata *metadataImpl) ArchivalConfig() (retCfg *ArchivalConfig) {
+	bucketSet := len(metadata.defaultBucket) != 0
+	status := GetArchivalStatus(metadata.archivalStatus())
+	if !bucketSet || status == ArchivalDisabled {
```
venkat1109: I believe this validation already lives inside your ArchivalConfig class - so this code is redundant. And I like this code living inside the ArchivalConfig class

service/frontend/domainArchivalConfigStateMachine.go
```diff
@@ -21,216 +21,234 @@
...
 var (
-	errCannotProvideBucket       = &shared.BadRequestError{Message: "Request can only provide bucket name if request also gives status of enabled or disabled."}
-	errDisallowedBucketMetadata  = &shared.BadRequestError{Message: "Cannot set bucket owner or bucket retention (must update bucket manually)."}
-	errStateChangeToNeverEnabled = &shared.BadRequestError{Message: "Cannot change state to never enabled."}
-	errBucketNameUpdate          = &shared.BadRequestError{Message: "Cannot update bucket name after archival status has been moved out of never_enabled for the first time."}
-	errCannotProvideEmptyBucket  = &shared.BadRequestError{Message: "Cannot provide empty bucket name."}
-	errNoDefaultBucket           = &shared.BadRequestError{Message: "No default bucket provided, is cluster correctly configured for archival?"}
-	errCannotHandleStateChange   = &shared.BadRequestError{Message: "Encountered current state and update that cannot be handled (this should be impossible)"}
+	errInvalidState            = &shared.BadRequestError{Message: "Encountered illegal state: archival is enabled but bucket is not set (should be impossible)"}
```
venkat1109: It's weird for this class to be returning errors of type: BadRequestErrors. Those errors are for the top level rpc request handlers. This class ideally should be returning just errors.New("..") - and let the upper layer encapsulate into whatever error that makes sense according to caller context. 

On the other hand, if there is an assumption here that the methods in this file are only called within rpc request handlers - just add a comment here

service/frontend/domainArchivalConfigStateMachine.go
```diff
@@ -21,216 +21,234 @@
...
+		status: shared.ArchivalStatusDisabled,
 	}
 }
 
-func registerRequestToArchivalConfig(request *shared.RegisterDomainRequest, defaultBucket string) (*archivalConfigUpdate, error) {
-	requestArchivalConfig := &archivalConfigUpdate{
-		defaultBucket:   defaultBucket,
-		requestedBucket: request.ArchivalBucketName,
-		requestedStatus: request.ArchivalStatus,
+func registerToEvent(request *shared.RegisterDomainRequest, defaultBucket string) (*archivalEvent, error) {
```
venkat1109: I like having a constructor like newArchivalRegisterEvent(param1, param2...) to this approach - (1) less coupling between this and the rpc handler (2) registerToEvent() as a top level function doesn't make a lot of sense for the reader

Or alternatively, rename this method to toArchivalRegisterEvent() and make this a class method inside your workflowHandler class.

service/frontend/domainArchivalConfigStateMachine.go
```diff
@@ -21,216 +21,234 @@
...
+	if err := event.validate(); err != nil {
 		return nil, err
 	}
-	return requestArchivalConfig, nil
+	return event, nil
 }
 
-func updateRequestToArchivalConfig(request *shared.UpdateDomainRequest, defaultBucket string) (*archivalConfigUpdate, error) {
-	requestArchivalConfig := &archivalConfigUpdate{
+func updateToEvent(request *shared.UpdateDomainRequest, defaultBucket string) (*archivalEvent, error) {
```
venkat1109: same comment as above - I prefer newArchivalUpdateEvent(config, defaultBucket)

service/history/timerQueueProcessorBase.go
```diff
@@ -590,40 +585,45 @@ func (t *timerQueueProcessorBase) processDeleteHistoryEvent(task *persistence.Ti
...
+	domainCacheEntry, err := t.historyService.getActiveDomainEntry(common.StringPtr(task.DomainID))
 	if err != nil {
 		return err
 	}
+	domainArchivalStatus := domainCacheEntry.GetConfig().ArchivalStatus
 
-	err = t.deleteWorkflowHistory(task, msBuilder)
-	if err != nil {
-		return err
+	switch clusterArchivalStatus {
```
venkat1109: can you emit metrics to get an idea of how many times we (1) delete_history here (2) send signal to archival and delete history later (3) do nothing ?

service/frontend/domainArchivalConfigStateMachine.go
andrewjdawson2016: Yes logically this class is just a module used by the RPC layer. I will comment it as such.

service/frontend/domainArchivalConfigStateMachine.go
andrewjdawson2016: I picked the second option because logically I am thinking of this file as just an extension workflowHandler.go. Despite making these member methods on workflowHandler, I still left them in this file rather than moving them to workflowHandler.go file.

service/history/historyEngine.go
```diff
@@ -26,9 +26,8 @@ import (
 	"errors"
 	"fmt"
 	"github.com/uber/cadence/client/public"
-	"time"
-
 	"github.com/uber/cadence/service/worker/sysworkflow"
+	"time"
```
venkat1109: gofmt

PR #1494, opened 2019-02-25
DC migration temporary solution 

wxing1292:
* Add config to ignore disabled clusters
* Disable all history re-replication functionality, except the one triggered by retry task error when replicating history
* Frontend will allow change of target cluster in domain replication config


dmetzgar:
Since there is a bunch of code that will be changed/removed post DC migration, should we have this in master or in its own release branch?

PR #1495, opened 2019-02-26
docker/README.md: CLI: how to register new domain

vkuzmin-uber:
Specify "host" as the network when start Cadence-ClI with docker.
`docker run --network=host --rm ubercadence/cli:master`


vancexu:
Thanks for contribution @vkuzmin-uber !

PR #1501, opened 2019-03-02
Change visibility closed_time to use last event timestamp

vancexu:
Currently closed_time in visibility is equal to mutableState lastUpdateTime. For some cases lastUpdateTime might be changed when processing close task, which will cause one workflow produce duplicate visibility records with only timestamp is slightly different. 

This PR fixes this issue by using complete event timestamp as visibility closed_time. Now all workflow complete events will go to eventCache, so that when process close task we can make use of this cache.


service/history/mutableStateBuilder.go
```diff
@@ -2302,6 +2297,8 @@ func (e *mutableStateBuilder) AddContinueAsNewEvent(decisionCompletedEventID int
 	}
 
 	continueAsNewEvent := e.hBuilder.AddContinuedAsNewEvent(decisionCompletedEventID, newRunID, attributes)
+	// Write the event to cache only on active cluster
+	e.writeCompletionEventToCache(continueAsNewEvent)
```
samarabbas: I think we should move writing event to cache logic to Replicate*** API within mutable state for all closed events.
Then you can remove this comment too.

service/history/transferQueueActiveProcessor.go
```diff
@@ -403,7 +400,7 @@ func (t *transferQueueActiveProcessorImpl) processCloseExecution(task *persisten
 
 	workflowTypeName := executionInfo.WorkflowTypeName
 	workflowStartTimestamp := executionInfo.StartTimestamp.UnixNano()
-	workflowCloseTimestamp := msBuilder.GetLastUpdatedTimestamp()
```
samarabbas: Can you remove GetLastUpdatedTimestamp API from mutableState.  Looks like this is used also on DescribeWorkflowExecution and will now give incorrect information.

service/history/mutableStateBuilder.go
vancexu: Updated

service/history/transferQueueActiveProcessor.go
vancexu: Removed

PR #1502, opened 2019-03-04
Handle out of order workflow replication

wxing1292:
* Attach unique ID generated by shard to each workflow execution during creation
* All replication message will include the above ID
* When workflow with the same ID are replicated out of order, comparison of the above ID will drop the replication message with smaller ID


service/frontend/workflowHandler.go
```diff
@@ -87,6 +87,7 @@ type (
 		PersistenceToken  []byte
 		TransientDecision *gen.TransientDecisionInfo
 		EventStoreVersion int32
+		CreateTaskID      int64
```
samarabbas: Is this backward compatible?

wxing1292: public API does not use this field,
only re-send stack use it

service/history/historyReplicator.go
```diff
@@ -983,22 +997,28 @@ func (r *historyReplicator) replicateWorkflowStarted(ctx context.Context, contex
...
 		logError(logger, "Fail to flush buffer for current workflow.", err)
-		return "", 0, false, err
+		return "", 0, 0, false, err
 	}
-	return currentContext.getExecution().GetRunId(), currentMutableState.GetNextEventID(), currentMutableState.IsWorkflowExecutionRunning(), nil
+	runID = currentContext.getExecution().GetRunId()
+	nextEventID = currentMutableState.GetNextEventID()
+	createTaskID = currentMutableState.GetExecutionInfo().CreateTaskID
+	isRunning = currentMutableState.IsWorkflowExecutionRunning()
+	currentRelease(err)
```
samarabbas: You already have a release in a defer function.  Is that really needed?

wxing1292: last time there were panic within the lock, using defer right after getting the lock, will at least make the deadlock go away

PR #1509, opened 2019-03-05
Bugfix: there shall be no events in history after workflow finish event.

wxing1292:


service/history/workflowExecutionContext.go
```diff
@@ -927,3 +932,35 @@ func (c *workflowExecutionContextImpl) emitSessionUpdateStats(stats *persistence
...
 		time.Duration(stats.DeleteRequestCancelInfoCount))
 }
+
+func (c *workflowExecutionContextImpl) validateNoEventsAfterWorkflowFinish(input []*workflow.HistoryEvent) error {
+	if len(input) == 0 {
+		return nil
+	}
+
+	seenWorkflowCompletionEvent := false
+	for _, event := range input {
```
samarabbas: instead of scan through entire list of events, how about we validate using the workflow state in mutableState.  Basically if workflow state is CLOSED and last event in the update is not one of 6 workflow close events then fail the transaction.

service/history/workflowExecutionContext.go
```diff
@@ -927,3 +932,35 @@ func (c *workflowExecutionContextImpl) emitSessionUpdateStats(stats *persistence
...
+		if seenWorkflowCompletionEvent {
+			c.logger.WithFields(bark.Fields{
+				logging.TagDomainID:            c.domainID,
+				logging.TagWorkflowExecutionID: c.workflowExecution.GetWorkflowId(),
+				logging.TagWorkflowRunID:       c.workflowExecution.GetRunId(),
+			}).Warn("encount case where events appears after workflow finish.")
```
samarabbas: This should be error log.  Also fix spelling for "encounter".

service/history/workflowExecutionContext.go
```diff
@@ -927,3 +932,35 @@ func (c *workflowExecutionContextImpl) emitSessionUpdateStats(stats *persistence
...
+
+			return &shared.InternalServiceError{Message: "error validating last event being workflow finish event."}
```
samarabbas: Can you define an error in the beginning of file.  Let's use the same pattern we follow in other parts of the code.

service/history/workflowExecutionContext.go
wxing1292: i think there is an edge case about transient decision.
transient decision -> workflow completion will result to writing 2 batch of events to DB.
the first batch should be decision task scheduled & started (ms builder at this time indicate workflow closed), this batch will fail.

PR #1511, opened 2019-03-06
tasklistManager: support for range deletes / skipping over expired tasks

venkat1109:
This patch contains following changes to the task list manager implementation in matching:

__Range deletion of completed tasks__
Today, every time a task is dispatched to a poller, it is immediately deleted from the database. There is one database call for every task dispatch. This patch gets rid of this behavior and replaces it with range deletion of completed tasks. The deletion code is still invoked in the dispatch path, but it only issues a range delete when either of size_threshold or time_threshold conditions are met.

__Skipping over expired tasks__
This is applicable only for SQL based persistence implementations. Every task created has an associated schedule_to_start_timeout after which the task can be safely dropped. Today, this is achieved by using column TTLs in cassandra. With SQL based persistence, TTL is not an option. So, we explicitly store the task_expiry as part of the task data. This patch adds code to skip over expired tasks when reading tasks from persistence.

In addition to the changes above, this patch also contains minor refactoring for clarity / readability:
- getTasksPump and friends are moved to taskReader.go
- rateLimiter is moved to its own file
- a new file taskCleaner.go contains the code for range deletion of tasks


service/matching/taskGC.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return backlog > 0 && (ignoreTimeCond || time.Now().Sub(tgc.lastDeleteTime) > maxTimeBetweenTaskDeletes)
+}
+
+func (tgc *taskGC) tryLock() bool {
+	return atomic.CompareAndSwapInt64(&tgc.lock, 0, 1)
+}
+
+func (tgc *taskGC) unlock() {
+	atomic.StoreInt64(&tgc.lock, 0)
```
samarabbas: unlock is not idempotent, this could lead to interesting concurrency issues if caller invokes unlock more than once.

venkat1109: Given that the lock/unlock is an internal implementation detail and not exposed to callers, I don't feel strongly about providing this guarantee. Discussed offline and decided to punt on this

PR #1520, opened 2019-03-09
Add unit tests for signalWithStart

vancexu:
~~Currently to prevent history event leak, there is a `defer deleteEvents()` in start/signalWithStart workflow APIs. It will wrongly delete events in duplicate requestID cases; also using bool `needDeleteHistory` is little bit error prone.~~

~~This PR addressed above issues.~~

This PR add some comments and tests for start and signalwithstart.

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -366,7 +366,7 @@ func (e *historyEngineImpl) appendFirstBatchHistoryEvents(msBuilder mutableState
 	return
 }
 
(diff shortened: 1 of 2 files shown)


vancexu:
> That doesn't seem like a bug for me.
> Dupplicated request also generate randomly new runIDs, so it won't delete the previous request's history events.
> Changing this will leak data when startWF with duplicated requests. Becuase retError is nil, the defer function cannot clean up the unused duplicated history events.
> 
> I agree that using that boolean is not easy to understand. I am open to make it more readable.

Thanks long for pointing out this. You are right, it is not a bug. 
I add comments on the places confused me, and add some tests.

PR #1521, opened 2019-03-11
Bugfix: visibility RecordWorkflowExecutionClosed API should use query…

wxing1292:
… timestamp > RecordWorkflowExecutionStarted API query timestamp for the same workflow ID & run ID combination


vancexu:
Or we could just use default Cassandra WriteTimeStamp instead of explicit setting it. What do you think @samarabbas ?

wxing1292:
> Or we could just use default Cassandra WriteTimeStamp instead of explicit setting it. What do you think @samarabbas ?

probably due to #155

PR #1539, opened 2019-03-15
logging: use throttled logger for limits / size violations

venkat1109:
Patch contains the following changes
- New implementation of bark.Logger that throttles the emitted logs
- Noisy logs in frontend now use throttled logger instead of regular logger
- Added new log in shardContext for history-size violations
- To avoid circular dependency, I had to move tokenbucket and timesource into their separate packages. This part of the change is just mechanical

### DIFF
--- a/common/logging/throttle.go
+++ b/common/logging/throttle.go
@@ -38,6 +38,8 @@ type throttledLogger struct {
 	}
 }
 
+var _ bark.Logger = (*throttledLogger)(nil)
+
 // NewThrottledLogger returns an implementation of bark logger that throttles the
 // log messages being emitted. The underlying implementation uses a token bucket
 // ratelimiter and stops emitting logs once the bucket runs out of tokens
--- a/common/logging/throttle.go
+++ b/common/logging/throttle.go
@@ -21,26 +21,38 @@
 package logging
 
 import (
+	"sync/atomic"
+
 	"github.com/uber-common/bark"
 	"github.com/uber/cadence/common/clock"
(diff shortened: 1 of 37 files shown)


common/logging/throttle.go
```diff
@@ -0,0 +1,184 @@
...
+type throttledLogger struct {
+	rps int32
+	tb  tokenbucket.TokenBucket
+	log bark.Logger
+	cfg struct {
+		rps dynamicconfig.IntPropertyFn
+	}
+}
+
+// NewThrottledLogger returns an implementation of bark logger that throttles the
```
longquanzheng: Can we add a type assertion here to be more explicit for our IDE?

common/service/service.go
```diff
@@ -124,6 +127,9 @@ func New(params *BootstrapParams) Service {
 		dispatcherProvider:    params.DispatcherProvider,
 		dynamicCollection:     dynamicconfig.NewCollection(params.DynamicConfig, params.Logger),
 	}
+
+	tlRPS := sVice.dynamicCollection.GetIntProperty(dynamicconfig.ThrottledLogRPS, defaultThrottledLogRPS)
```
longquanzheng: It would be more consistent if we can put that dynamic config in each service config, initialize in "NewConfig" and put the default value there. 
A good example is that we have history/matching/frontendRPS. 
That will also give us flexibility to give different service different quotas.

common/logging/throttle.go
venkat1109: done

PR #1542, opened 2019-03-15
Add DescribeTaskList command to admin client

yycptt:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1230,10 +1230,20 @@ struct DescribeTaskListRequest {
   10: optional string domain
   20: optional TaskList taskList
   30: optional TaskListType taskListType
+  40: optional bool isAdmin
```
andrewjdawson2016: Rather than isAdmin lets called this something like: detailed. The reason here is I want to keep all admin related terminology as isolated from the rest of the codebase as possible. 

Another name that I like is includeTaskListStatus. You are already referring to this collection of fields as status in your comment "status of tasklist's ackManager (readLevel, ackLevel, backlogCountHint and taskIDBlock)."

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1230,10 +1230,20 @@ struct DescribeTaskListRequest {
...
 }
 
 struct DescribeTaskListResponse {
   10: optional list<PollerInfo> pollers
+  20: optional i64 (js.type = "Long") backlogCountHint
```
andrewjdawson2016: Let's create a type that holds all the detailed information. That way we have a clear distinction between what is set when detailed=true and what is not set when detailed=false. This also lets us add fields directly to DescribeTaskListResponse and have it still be clear which fields are for detailed and which fields are not for detailed.

service/matching/matchingEngine_test.go
```diff
@@ -330,12 +330,18 @@ func (s *matchingEngineSuite) PollForTasksEmptyResultTest(taskType int) {
...
 				TaskList:     taskList,
 				TaskListType: &taskListType,
+				IsAdmin:      common.BoolPtr(true),
 			},
 		})
 		s.NoError(err)
 		s.Equal(1, len(descResp.Pollers))
 		s.Equal(identity, descResp.Pollers[0].GetIdentity())
 		s.NotEmpty(descResp.Pollers[0].GetLastAccessTime())
+		s.Zero(descResp.GetBacklogCountHint())
```
andrewjdawson2016: There are two more tests I can see as useful here
1. A test where IsAdmin = false and you verify that the expected fields are not set (you actually want them to be null not just defaults)
2. A test where you get a response back that contains some non-default values. Right now this test will pass even if the response is not set because GetBacklogCountHint will return the default value of zero if backlogHint is not set.

tools/cli/admin.go
```diff
@@ -615,3 +615,27 @@ func newAdminElasticSearchCommands() []cli.Command {
 		},
 	}
 }
+
+func newAdminTaskListCommands() []cli.Command {
+	return []cli.Command{
+		{
+			Name:    "describe",
+			Aliases: []string{"desc"},
+			Usage:   "Describe pollers and ack manager related info of tasklist",
```
andrewjdawson2016: Its some what of an implementation detail that the contents of the stuff we are returning come from ack manager. A new cadence team member would not be super clear on what this is expected to return. What about something like "Describe pollers and task list status information"

tools/cli/adminTaskListCommands.go
```diff
@@ -0,0 +1,101 @@
...
+	fmt.Printf("\n")
+
+	pollers := response.Pollers
+	if len(pollers) == 0 {
+		ErrorAndExit(colorMagenta("No poller for tasklist: "+taskList), nil)
+	}
+	printPollerInfo(pollers, taskListType)
+}
+
+func printAckMgrInfo(response *s.DescribeTaskListResponse) {
```
andrewjdawson2016: Same comment here on the method name. Lets call this printTaskListStatus.

tools/cli/adminTaskListCommands.go
```diff
@@ -0,0 +1,101 @@
...
+	table.SetHeaderColor(tableHeaderBlue, tableHeaderBlue, tableHeaderBlue, tableHeaderBlue, tableHeaderBlue)
+	table.Append([]string{strconv.FormatInt(response.GetReadLevel(), 10),
+		strconv.FormatInt(response.GetAckLevel(), 10),
+		strconv.FormatInt(response.GetBacklogCountHint(), 10),
+		strconv.FormatInt(taskIDBlock.GetStartID(), 10),
+		strconv.FormatInt(taskIDBlock.GetEndID(), 10)})
+	table.Render()
+}
+
+func printPollerInfo(pollers []*s.PollerInfo, taskListType s.TaskListType) {
```
andrewjdawson2016: Now that we are using the same types can we use this code to generate a table both in the admin cli and in the public cli?

service/matching/matchingEngine.go
```diff
@@ -570,14 +570,7 @@ func (e *matchingEngineImpl) DescribeTaskList(ctx context.Context, request *m.De
...
 
-	pollers := []*workflow.PollerInfo{}
-	for _, poller := range tlMgr.GetAllPollerInfo() {
-		pollers = append(pollers, &workflow.PollerInfo{
-			Identity:       common.StringPtr(poller.identity),
-			LastAccessTime: common.Int64Ptr(poller.lastAccessTime.UnixNano()),
-		})
-	}
-	return &workflow.DescribeTaskListResponse{Pollers: pollers}, nil
+	return tlMgr.DescribeTaskList(request.DescRequest.GetIsAdmin()), nil
```
andrewjdawson2016: Nice, I like that this logic belongs to the taskListManager!

service/matching/taskListManager.go
```diff
@@ -542,6 +543,34 @@ func (c *taskListManagerImpl) GetAllPollerInfo() []*pollerInfo {
...
+	pollers := []*s.PollerInfo{}
+	for _, poller := range c.GetAllPollerInfo() {
+		pollers = append(pollers, &s.PollerInfo{
+			Identity:       common.StringPtr(poller.identity),
+			LastAccessTime: common.Int64Ptr(poller.lastAccessTime.UnixNano()),
+		})
+	}
+
+	response := &s.DescribeTaskListResponse{Pollers: pollers}
+	if !isAdminRequest {
```
andrewjdawson2016: Rename isAdmin here as well.

tools/cli/adminTaskListCommands.go
yycptt: I don't think they have the same type, in admin cli PollerInfo is defined in "github.com/uber/cadence/.gen/go/shared", while in the public cli it's defined in "go.uber.org/cadence/.gen/go/shared". The reason is that, when calling admin DescribeTaskList, we need to use ServerFrontendClient(public cli is using ClientFrontendClient); otherwise, the old(client side) shared.thrift file will be used.

service/matching/taskListManager_test.go
```diff
@@ -116,6 +116,53 @@ func TestIsTaskAddedRecently(t *testing.T) {
...
+	require.NotNil(t, taskListStatus)
+	require.Zero(t, taskListStatus.GetAckLevel())
+	require.Equal(t, taskCount, taskListStatus.GetReadLevel())
+	require.Equal(t, taskCount, taskListStatus.GetBacklogCountHint())
+	taskIDBlock := taskListStatus.GetTaskIDBlock()
+	require.Equal(t, int64(1), taskIDBlock.GetStartID())
+	require.Equal(t, tlm.config.RangeSize, taskIDBlock.GetEndID())
+
+	// Add a poller and complete all tasks
+	tlm.updatePollerInfo(pollerIdentity{identity: PollerIdentity})
```
andrewjdawson2016: Nice!

tools/cli/adminTaskListCommands.go
andrewjdawson2016: Oh yes.

PR #1548, opened 2019-03-19
Shorten context timeout for pollForDecision/ActivityTask at matching side

yycptt:


service/matching/taskListManager.go
```diff
@@ -410,12 +414,23 @@ func (c *taskListManagerImpl) getTask(ctx context.Context) (*getTaskResult, erro
...
+
 	childCtx := ctx
+	// We need to set a shorter timeout than the original ctx; otherwise, by the time ctx deadline is
+	// reached, instead of emptyTask, context timeout error is returned to the frontend by the rpc stack,
+	// which counts against our SLO. By shortening the timeout by a very small amount, the emptyTask can be
+	// returned to the handler before a context timeout error is generated.
+	if deadline, ok := childCtx.Deadline(); ok {
+		var cancel context.CancelFunc
+		childCtx, cancel = context.WithTimeout(ctx, deadline.Sub(time.Now())-returnEmptyTaskTimeBudget)
+		defer cancel()
```
andrewjdawson2016: How/When this context gets canceled seems weird to me. There is a defer cancel here and there is also one below. I think we want to make this more simple. How about the child context should always have a deadline and a cancel function. If the parent has a deadline then do what you are doing now. If the parent does not have a deadline then you can use `c.config.LongPollExpirationInterval()`

That way you can just have the function do `defer cancel()` in one place.

andrewjdawson2016: In fact think through if we even need the timer with this approach at all?

service/matching/matchingEngine_test.go
```diff
@@ -194,11 +194,23 @@ func (s *matchingEngineSuite) TestAckManager() {
...
+	s.PollForTasksEmptyResultTest(s.callContext, persistence.TaskListTypeActivity)
 }
 
 func (s *matchingEngineSuite) TestPollForDecisionTasksEmptyResult() {
-	s.PollForTasksEmptyResultTest(persistence.TaskListTypeDecision)
+	s.PollForTasksEmptyResultTest(s.callContext, persistence.TaskListTypeDecision)
+}
+
+func (s *matchingEngineSuite) TestPollForActivityTasksEmptyResultWithShortContext() {
+	callContext, cancel := context.WithTimeout(s.callContext, returnEmptyTaskTimeBudget+10*time.Millisecond)
```
andrewjdawson2016: So right now if somehow matching gets a context with a really short timeout we will still handle the request. But in frontend we say if we ever get a short timeout it's an error case right off the bat. I think it makes sense to also have these super short context timeout checks in matching. 

It may actually make sense to only have the checks exists in matching and make frontend dumber about the context timeout checks?

service/frontend/workflowHandler_test.go
```diff
@@ -254,6 +254,24 @@ func (s *workflowHandlerSuite) TestDisableListVisibilityByFilter() {
...
+	config := s.newConfig()
+	wh := s.getWorkflowHandler(config)
+	wh.metricsClient = wh.Service.GetMetricsClient()
+	wh.startWG.Done()
+
+	shortCtx, cancel := context.WithTimeout(context.Background(), minLongPollTimeout-time.Millisecond)
+	defer cancel()
+
+	_, err := wh.PollForDecisionTask(shortCtx, &shared.PollForDecisionTaskRequest{})
+	assert.Error(s.T(), err)
```
andrewjdawson2016: This should be written as `s.Error(err)`

service/frontend/workflowHandler_test.go
```diff
@@ -254,6 +254,24 @@ func (s *workflowHandlerSuite) TestDisableListVisibilityByFilter() {
...
+	assert.Equal(s.T(), errContextTimeoutTooShort, err)
```
andrewjdawson2016: This should be written as `s.Equal(errContextTimeoutTooShort, err)`

service/matching/taskListManager.go
```diff
@@ -410,12 +414,23 @@ func (c *taskListManagerImpl) getTask(ctx context.Context) (*getTaskResult, erro
 	defer timer.Stop()
 
 	pollerID, ok := ctx.Value(pollerIDKey).(string)
+
 	childCtx := ctx
+	// We need to set a shorter timeout than the original ctx; otherwise, by the time ctx deadline is
```
andrewjdawson2016: Nice comment.

service/frontend/workflowHandler_test.go
yycptt: Yes, this is my original way of writing this. But it seems like other tests in this file are using assert, so I followed their pattern.

service/frontend/workflowHandler_test.go
yycptt: Same here

service/matching/matchingEngine_test.go
andrewjdawson2016: Lets put `returnEmptyTaskTimeBudget+10*time.Millisecond` into some meaning variable name. Its hard to just read that and know what is being tested.

service/matching/matchingEngine_test.go
```diff
@@ -194,11 +194,23 @@ func (s *matchingEngineSuite) TestAckManager() {
...
+	defer cancel()
+	s.PollForTasksEmptyResultTest(callContext, persistence.TaskListTypeActivity)
+}
+
+func (s *matchingEngineSuite) TestPollForDecisionTasksEmptyResultWithShortContext() {
+	callContext, cancel := context.WithTimeout(s.callContext, returnEmptyTaskTimeBudget+10*time.Millisecond)
```
andrewjdawson2016: Same comment here.

service/matching/taskListManager.go
```diff
@@ -406,24 +410,34 @@ func (c *taskListManagerImpl) completeTaskPoll(taskID int64) int64 {
...
+		// which counts against our SLO. By shortening the timeout by a very small amount, the emptyTask can be
+		// returned to the handler before a context timeout error is generated.
+		shortenedCtxTimeout := deadline.Sub(time.Now()) - returnEmptyTaskTimeBudget
+		if shortenedCtxTimeout < childCtxTimeout {
+			childCtxTimeout = shortenedCtxTimeout
+		}
+	}
+	// ChildCtx timeout will be the shorter of longPollExpirationInterval and shortened parent context timeout.
+	childCtx, cancel := context.WithTimeout(ctx, childCtxTimeout)
+	defer cancel()
```
andrewjdawson2016: Yes, this is now much easier to reason about when the context gets canceled.

PR #1549, opened 2019-03-19
Fix losing signal when closing workflow with in-flight decision

longquanzheng:
Initially I wanted to centralize it in update() in WFContext. However it won't work for XDC because XDC requires no more events after WF closed.

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -1126,6 +1126,8 @@ func failInFlightDecisionToClearBufferedEvents(msBuilder mutableState) error {
 		if event == nil {
 			return &workflow.InternalServiceError{Message: "Failed to add decision failed event."}
 		}
+
+		return msBuilder.FlushBufferedEvents()
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


service/history/historyEngine.go
```diff
@@ -1107,6 +1111,20 @@ func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(blob []byte
...
+func failInFlightDecisionToClearBufferedEvents(msBuilder mutableState) error {
+	if msBuilder.HasInFlightDecisionTask() {
+		di, _ := msBuilder.GetInFlightDecisionTask()
+
+		event := msBuilder.AddDecisionTaskFailedEvent(di.ScheduleID, di.StartedID, workflow.DecisionTaskFailedCauseResetWorkflow, nil,
+			identityHistoryService, decisionFailureForBuffered, "", "", 0)
+		if event == nil {
+			return &workflow.InternalServiceError{Message: "Failed to add decision failed event."}
+		}
+	}
```
wxing1292: should also flush the buffer here, or otherwise, buffered events are still in the buffer

PR #1568, opened 2019-03-20
Use global optional argument for RPC timeout

longquanzheng:
https://github.com/uber/cadence/issues/880

### DIFF
--- a/tools/cli/commands.go
+++ b/tools/cli/commands.go
@@ -56,10 +56,13 @@ const (
 	defaultTimeFormat                = "15:04:05"   // used for converting UnixNano to string like 16:16:36 (only time)
 	defaultDateTimeFormat            = time.RFC3339 // used for converting UnixNano to string like 2018-02-15T16:16:36-08:00
 	defaultDomainRetentionDays       = 3
+	defaultContextTimeoutInSeconds   = 10
+	defaultContextTimeout            = defaultContextTimeoutInSeconds * time.Second
 	defaultContextTimeoutForLongPoll = 2 * time.Minute
-	defaultDecisionTimeoutInSeconds  = 10
-	defaultPageSizeForList           = 500
-	defaultWorkflowIDReusePolicy     = s.WorkflowIdReusePolicyAllowDuplicateFailedOnly
+
+	defaultDecisionTimeoutInSeconds = 10
+	defaultPageSizeForList          = 500
+	defaultWorkflowIDReusePolicy    = s.WorkflowIdReusePolicyAllowDuplicateFailedOnly
 
 	workflowStatusNotSet = -1
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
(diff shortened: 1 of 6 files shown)


tools/cli/commands.go
```diff
@@ -56,10 +56,13 @@ const (
 	defaultTimeFormat                = "15:04:05"   // used for converting UnixNano to string like 16:16:36 (only time)
 	defaultDateTimeFormat            = time.RFC3339 // used for converting UnixNano to string like 2018-02-15T16:16:36-08:00
 	defaultDomainRetentionDays       = 3
+	defaultContextTimeoutInSeconds   = 10
```
vancexu: let's keep default 5 sec, which is actually already very long for most api.

tools/cli/commands.go
```diff
@@ -942,7 +936,7 @@ func listOpenWorkflow(client client.Client, pageSize int, earliestTime, latestTi
 		request.TypeFilter = &s.WorkflowTypeFilter{Name: common.StringPtr(workflowType)}
 	}
 
-	ctx, cancel := newContextForLongPoll(timeout)
+	ctx, cancel := newContext(c)
```
vancexu: let's keep using longer default for list

tools/cli/commands.go
```diff
@@ -1148,12 +1142,20 @@ func getCliIdentity() string {
...
+		contextTimeout = time.Duration(c.GlobalInt(FlagContextTimeout)) * time.Second
+	}
+	return context.WithTimeout(context.Background(), contextTimeout)
 }
 
-func newContextForLongPoll(timeout time.Duration) (context.Context, context.CancelFunc) {
-	return context.WithTimeout(context.Background(), timeout)
+func newContextForLongPoll(c *cli.Context) (context.Context, context.CancelFunc) {
+	contextTimeout := defaultContextTimeoutForLongPoll
+	if c.GlobalInt(FlagContextTimeout) > 0 {
```
vancexu: I think because you set default value for FlagContextTimeout, this condition will always be true and lead to unexpected timeout for longpoll. Maybe use c.GlobalIsSet or remove default flag value.

longquanzheng: You are right, I missed that. I will use GlobalIsSet

tools/cli/commands.go
longquanzheng: OK

tools/cli/commands.go
longquanzheng: OK

tools/cli/commands.go
```diff
@@ -1148,12 +1142,20 @@ func getCliIdentity() string {
 	return fmt.Sprintf("cadence-cli@%s", hostName)
 }
 
-func newContext() (context.Context, context.CancelFunc) {
-	return context.WithTimeout(context.Background(), time.Second*5)
+func newContext(c *cli.Context) (context.Context, context.CancelFunc) {
+	contextTimeout := defaultContextTimeout
+	if c.GlobalInt(FlagContextTimeout) > 0 {
```
vancexu: would be better to also use GlobalIsSet here.

PR #1575, opened 2019-03-21
Rename metrics to be prometheus compatible

shreyassrivatsan:
This change dual emits our current metrics with prometheus compatible names. Once we start switch over our internal dashboards/alerts we will get rid of the old style metrics. In a subsequent change, we will also start adding the domain tag to the new style metrics.


common/metrics/client.go
```diff
@@ -26,6 +26,29 @@ import (
...
 )
 
+// N.B - this is a temporary Stopwatch struct to allow us to migrate to the
+// new style metric names
+
+// Stopwatch is a helper for simpler tracking of elapsed time, use the
+// Stop() method to report time elapsed since its created back to the
+// timer or histogram.
+type Stopwatch struct {
+	recorder1 tally.Stopwatch
```
venkat1109: since we have to create our own temporary StopWatch anyways - consider using scope.RecordTimer(latency) instead of creating two tally stopwatches internally. That would require passing the (tallyScope, oldName, newName) from the caller.

shreyassrivatsan: Makes sense. I changed it to be like that and added a `NewTestStopwatch` method that can be used for tests.

PR #1578, opened 2019-03-21
Add support for domain tags in metrics

shreyassrivatsan:
Creating an internal scope structure that allows adding pre-defined tags to metrics. Presently Domain is the one allowed tag.

### DIFF
--- a/common/metrics/interfaces.go
+++ b/common/metrics/interfaces.go
@@ -48,9 +48,20 @@ type (
 
 	// Scope is an interface for metrics
 	Scope interface {
-		Counter(id int) tally.Counter
-		Gauge(id int) tally.Gauge
-		Timer(id int) tally.Timer
+		// IncCounter increments a counter metric
+		IncCounter(counter int)
+		// AddCounter adds delta to the counter metric
+		AddCounter(counter int, delta int64)
+		// StartTimer starts a timer for the given
+		// metric name. Time will be recorded when stopwatch is stopped.
+		StartTimer(timer int) tally.Stopwatch
+		// RecordTimer starts a timer for the given
+		// metric name
+		RecordTimer(timer int, d time.Duration)
+		// UpdateGauge reports Gauge type absolute value metric
+		UpdateGauge(gauge int, value float64)
+		// Tagged return an internal scope that can be used to add additional
+		// information to metrics
 		Tagged(tags ...Tag) Scope
 	}
 )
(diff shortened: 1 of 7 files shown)


common/metrics/interfaces.go
```diff
@@ -43,5 +43,8 @@ type (
 		UpdateGauge(scope int, gauge int, value float64)
 		// Tagged returns a client that adds the given tags to all metrics
 		Tagged(tags map[string]string) Client
+		// Scope return an internal scope that can be used to add additional
+		// information to metrics
+		Scope(scope int) Scope
```
venkat1109: (1) can you get rid of the free style Tagged(tags map[string]string) method in favor of the strongly typed tags ? I don't expect a lot of places we use this today
(2) How about making the signature of Scope as  Scope(scope int, tags ..Tag) ? Any caller that calls Scope() almost always will need a sub-scope with some tags. So, taking the tags as input here saves additional api call / data copy later

shreyassrivatsan: Got rid of tagged. And change the Scope method to take tags. The actual Tagged support is still in the underlying scope returned though.

venkat1109: 👍

common/metrics/interfaces.go
```diff
@@ -39,7 +41,16 @@ type (
...
-		// Tagged returns a client that adds the given tags to all metrics
-		Tagged(tags map[string]string) Client
+		// Scope return an internal scope that can be used to add additional
+		// information to metrics
+		Scope(scope int, tags ...Tag) Scope
+	}
+
+	// Scope is an interface for metrics
+	Scope interface {
+		Counter(id int) tally.Counter
```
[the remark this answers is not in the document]
shreyassrivatsan: Done

common/metrics/tags.go
```diff
@@ -0,0 +1,29 @@
...
+	Value() string
+}
+
+// DomainTag wraps a domain tag
+type DomainTag struct {
+	value string
+}
+
+// NewDomainTag returns a new domain tag
+func NewDomainTag(value string) DomainTag {
```
venkat1109: Just a style preference. If you make DomainTag struct private (domainTag struct) - you could do something like following:

```
client.Scope(metrics.StartWFExecutionScope, 
metrics.DomainTag("domain-name"),
metrics.ShardIDTag(shardID)) 

reads better than 

client.Scope(metrics.StartWFExecutionScope, 
metrics.NewDomainTag("domain-name"),
metrics.NewShardIDTag(shardID))
```

Zap logger uses similar style for typed k/v tags

shreyassrivatsan: Done.

PR #1584, opened 2019-03-22
Fix gocql dependency breaking change on replication_factor check

vancexu:


longquanzheng:
Usually we want to avoid pinning to commit. Can you reach out to gocql team to make sure this is the right fix?

vancexu:
> Usually we want to avoid pinning to commit. Can you reach out to gocql team to make sure this is the right fix?

You are right, I reach out to gocql team and they recommend to pin for now. The long term fix is currently not sure yet, Samar is helping out and will talk with caas team on that

PR #1589, opened 2019-03-22
Fix startWorkflowExecution request with cron schedule

yycptt:
When starting a workflow, it will run immediately even when a cron schedule is specified. This PR fixes this bug. Related bug includes sending a signal to a cron workflow starts the workflow immediately and SignalWithStartWorkflowExecution. These are left as future work and separate issues will be created.


common/util.go
```diff
@@ -407,6 +408,24 @@ func ValidateCronSchedule(cronSchedule string) error {
...
+// GetBackoffForNextCronSchedule calculates the backoff time for the next run given
+// a cronSchedule and current time
+func GetBackoffForNextCronSchedule(cronSchedule string, nowTime time.Time) time.Duration {
+	if len(cronSchedule) == 0 {
+		return NoRetryBackoff
+	}
+
+	schedule, err := cron.ParseStandard(cronSchedule)
+	if err != nil {
+		return NoRetryBackoff
```
andrewjdawson2016: It seems weird to me that we don't surface the error if cron schedule is invalid. What about changing this method's API to return (time.Duration, error). The only real error case I see is cron schedule is not valid, all other cases we can return nil for error.

service/history/historyEngine.go
```diff
@@ -535,16 +536,26 @@ func (e *historyEngineImpl) StartWorkflowExecution(ctx context.Context, startReq
...
 	// Generate first timer task : WF timeout task
-	duration := time.Duration(*request.ExecutionStartToCloseTimeoutSeconds) * time.Second
+	cronBackoffDuration := time.Duration(cronBackoffSeconds) * time.Second
+	timeoutDuration := time.Duration(*request.ExecutionStartToCloseTimeoutSeconds)*time.Second + cronBackoffDuration
 	timerTasks := []persistence.Task{&persistence.WorkflowTimeoutTask{
-		VisibilityTimestamp: e.shard.GetTimeSource().Now().Add(duration),
+		VisibilityTimestamp: e.shard.GetTimeSource().Now().Add(timeoutDuration),
 	}}
+	if cronBackoffSeconds != 0 {
+		timerTasks = append(timerTasks, &persistence.WorkflowBackoffTimerTask{
```
samarabbas: Can you also validate this change to see if it works with cross-dc?  Basically we want to make sure WorkflowBackoffTimerTask is created on passive cluster so we can create the first decision task incase failover happens before first decision task getting created.

service/history/stateBuilder_test.go
```diff
@@ -188,10 +190,29 @@ func (s *stateBuilderSuite) TestApplyEvents_EventTypeWorkflowExecutionStarted()
 
 	s.mockMutableState.On("ClearStickyness").Once()
 	s.stateBuilder.applyEvents(domainID, requestID, execution, s.toHistory(event), nil, 0, 0)
-	s.Equal(1, len(s.stateBuilder.timerTasks))
-	timerTask, ok := s.stateBuilder.timerTasks[0].(*persistence.WorkflowTimeoutTask)
-	s.True(ok)
-	s.True(timerTask.VisibilityTimestamp.Equal(now.Add(time.Duration(executionInfo.WorkflowTimeout) * time.Second)))
+
+	backoffDuration := common.GetBackoffForNextCronSchedule(cronSchedule, now)
+	expectedTimerTasksLength := 1
```
wxing1292: plz write 2 test for the cron

common/util.go
```diff
@@ -407,6 +408,24 @@ func ValidateCronSchedule(cronSchedule string) error {
 	return nil
 }
 
+// GetBackoffForNextCronSchedule calculates the backoff time for the next run given
+// a cronSchedule and current time
+func GetBackoffForNextCronSchedule(cronSchedule string, nowTime time.Time) time.Duration {
```
wxing1292: plz do not put logic in the util file / package.

util does not provide good indication about anything
try time or cron package / file name

PR #1591, opened 2019-03-23
Integration test refactoring

meiliang86:
- Split large file into pieces
- Create common structure that can be shared between tests


host/testcluster.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+type (
+	// TestCluster is a base struct for integration tests
+	TestCluster struct {
+		persistencetests.TestBase
+		BlobstoreBase
+		sync.Mutex // This lock is to protect the race in ringpop initialization.
+
+		Host   Cadence
```
samarabbas: We should make the entire integration test completely agnostic to how Cadence cluster is setup.  All integration test cares about is cadence client to talk to cluster.  We need to store following use cases:

1. Startup Cadence host locally with desired config
2. Create a Cadence client for already running Cadence server, like docker.  (Through some environment variable)
3. Startup multiple cadence clusters for Cadence XDC testing

host/testcluster.go
```diff
@@ -0,0 +1,140 @@
...
+		engine wsc.Interface
+		Logger bark.Logger
+	}
+
+	// TestClusterOptions are options for test cluster
+	TestClusterOptions struct {
+		PersistOptions   *persistencetests.TestBaseOptions
```
samarabbas: I think we should get rid of these and completely rely on dynamic config when starting up the host.  We should be able to create multiple configurations by just providing different configs for the server.

Issue #1202, opened 2018-10-24
Expose server side retry information through describe API


samarabbas:
Describe workflow execution behavior is broken when server side retries kicks in for activities.  Currently we expose the following pending activity information:
```
{
      "ActivityID": "57",
      "ActivityType": {
        "name": "hearbeating_activity"
      },
      "State": "STARTED",
      "HeartbeatDetails": "162125\n",
      "LastHeartbeatTimestamp": "2018-10-24T20:15:57Z"
    }
```

If this activity timeouts due to heartbeat timeout it shows the following:
```
{
      "ActivityID": "57",
      "ActivityType": {
        "name": "hearbeating_activity"
      },
      "State": "SCHEDULED",
      "HeartbeatDetails": "",
      "LastHeartbeatTimestamp": "1970-01-01T00:00:00Z"
    }
```

Details and LastHeartbeatTimestamps are cleared.  We need to surface all retryable information in mutable state including previous errors, lastheartbeat, etc.

shail:
Any update on this? Would be helpful for debugging

andrewjdawson2016:
Let's start this task by showing what we want the response to look like. There are at least three interesting points in time to consider: (1) When the activity is running but retries are not exhausted. (2) Retries are exhausted for activity but workflow is not closed. (3) Workflow is closed. We need to make sure the describe workflow API surfaces correct and informative responses in all these cases.

Lets start by:
1. Figuring out what information we want to surface.
2. Write out sample responses in this ticket.

yycptt:
For case 2 and 3, the activity is either succeed or failed, and the information can be accessed by showing history of the workflow. For case 1, I am unable to get a local repro. I think it has already been fixed by [#1204](https://github.com/uber/cadence/pull/1204). See also [#1219](https://github.com/uber/cadence/pull/1219).

yycptt:
need to create an integration test to ensure the behavior is correct.

yycptt:
[PR](https://github.com/uber/cadence/pull/1593) for the integration test.

yycptt:
Just found that the issue still exists for cross-DC. Working on a fix.

yycptt:
Merged. Closing this issue.

PR #1593, opened 2019-03-25
Intergration test to ensure hearbeat details are not lost during retry and HB timeout

yycptt:
[#1202](https://github.com/uber/cadence/issues/1202) shows that heartbeat details are lost when there are retry policy and heartbeat timeout. Though this bug has already been fixed, there's no test to ensure the correct behavior. This PR creates a integration test to make sure heartbeat details are not lost.


samarabbas:
Can you also have a similar test as a cross-dc integration test.  I'm ok if you want to do it as a separate PR.

service/history/replicatorQueueProcessor.go
```diff
@@ -179,19 +179,11 @@ func (p *replicatorQueueProcessorImpl) processSyncActivityTask(task *persistence
...
-		// when activity is started, the hearbeat timestamp will be empty
-		// but due the in64 limitation, the actual timestamp got is
-		// roughly 17xx year.
-		// set the heartbeat timestamp to started time if empty
-		heartbeatTime = common.Int64Ptr(activityInfo.LastHeartBeatUpdatedTime.UnixNano())
-		if *heartbeatTime < *startedTime {
-			heartbeatTime = startedTime
-		}
 	}
+	heartbeatTime = common.Int64Ptr(activityInfo.LastHeartBeatUpdatedTime.UnixNano())
```
wxing1292: plz add a comment that LastHeartBeatUpdatedTime must be valid when seeing the sync activity replication task

PR #1597, opened 2019-03-25
Enable events v2 integration test

meiliang86:


longquanzheng:
Test code are fixed and landed in master. You should be able to land it after rebase, without changing any test code.

PR #1598, opened 2019-03-26
Schedule decision task when cancelled activity has not started

### LINKED ISSUE #1513: ActivityCancellation does not result in new DecisionTask
When decider makes a activity cancellation decision and activity is not started, we write 2 events to history 'ActivityTaskCancelRequested' and 'ActivityTaskCanceled'.  'ActivityTaskCanceled' event should result in new DecisionTask being created as the decider may be clocked on activity completion before it makes workflow close decision.
This could result into workflow getting stuck.

shreyassrivatsan:
Fixes #1513

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -1509,7 +1509,6 @@ Update_History_Loop:
 					// schedule a decision task to ensure the workflow makes progress.
 					msBuilder.AddActivityTaskCanceledEvent(ai.ScheduleID, ai.StartedID, *actCancelReqEvent.EventId,
 						[]byte(activityCancelationMsgActivityNotStarted), common.StringDefault(request.Identity))
-					isComplete = false
 					hasUnhandledEvents = true
 				}
 
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -1234,6 +1234,7 @@ Update_History_Loop:
 		completedID := *completedEvent.EventId
 		hasUnhandledEvents := msBuilder.HasBufferedEvents()
 		isComplete := false
+		activityNotStartedCancelled := false
 		transferTasks := []persistence.Task{}
 		timerTasks := []persistence.Task{}
 		var continueAsNewBuilder mutableState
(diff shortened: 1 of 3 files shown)


service/history/historyEngine.go
```diff
@@ -1495,9 +1498,12 @@ Update_History_Loop:
 				}
 
 				if ai.StartedID == common.EmptyEventID {
-					// We haven't started the activity yet, we can cancel the activity right away.
+					// We haven't started the activity yet, we can cancel the activity right away and
+					// schedule a decision task to ensure the workflow makes progress.
 					msBuilder.AddActivityTaskCanceledEvent(ai.ScheduleID, ai.StartedID, *actCancelReqEvent.EventId,
 						[]byte(activityCancelationMsgActivityNotStarted), common.StringDefault(request.Identity))
+					isComplete = false
```
samarabbas: I'm not sure overriding isComplete to false is the right thing to do over here.  If a decider makes request_cancel and complete decision in a single transaction we would still want to honor completion.

service/history/historyEngine_test.go
```diff
@@ -3740,10 +3740,10 @@ func (s *engineSuite) TestRequestCancel_RespondDecisionTaskCompleted_Scheduled()
 	s.Nil(err)
 
 	executionBuilder := s.getBuilder(domainID, we)
-	s.Equal(int64(11), executionBuilder.GetExecutionInfo().NextEventID)
+	s.Equal(int64(12), executionBuilder.GetExecutionInfo().NextEventID)
 	s.Equal(int64(7), executionBuilder.GetExecutionInfo().LastProcessedEvent)
 	s.Equal(persistence.WorkflowStateRunning, executionBuilder.GetExecutionInfo().State)
-	s.False(executionBuilder.HasPendingDecisionTask())
+	s.True(executionBuilder.HasPendingDecisionTask())
```
samarabbas: Can we also assert on expected history after processing of decision?
Also can you add a test case when completion and request_cancel happens in the same transaction.  We want to make sure workflow completion is the last event in this situation.

service/history/historyEngine.go
```diff
@@ -1502,9 +1506,11 @@ Update_History_Loop:
 				}
 
 				if ai.StartedID == common.EmptyEventID {
-					// We haven't started the activity yet, we can cancel the activity right away.
+					// We haven't started the activity yet, we can cancel the activity right away and
+					// schedule a decision task to ensure the workflow makes progress.
 					msBuilder.AddActivityTaskCanceledEvent(ai.ScheduleID, ai.StartedID, *actCancelReqEvent.EventId,
 						[]byte(activityCancelationMsgActivityNotStarted), common.StringDefault(request.Identity))
+					activityNotStartedCancelled = true
```
shreyassrivatsan: have to use a separate flag here, reusing `hasUnhandledEvents` causes issues if we send also have a complete in our set of decisions.

PR #1599, opened 2019-03-26
Add domain tagged metrics for history sizes

### LINKED ISSUE #1383: Emit domain specific metric for large workflow executions
This feature is allow customers to setup alerts when their workflow executions exceed a certain threshold.
We already have workflow history size stored in mutable state.  We just need to emit a domain specific metric on each update when new events are written to workflow execution.

shreyassrivatsan:
Fixes #1383


service/frontend/workflowHandler.go
```diff
@@ -2846,6 +2877,7 @@ func (wh *WorkflowHandler) getHistory(scope int, domainID string, execution gen.
 
 	if len(historyEvents) > 0 {
 		wh.metricsClient.RecordTimer(scope, metrics.HistorySize, time.Duration(size))
+		domainScope.RecordTimer(metrics.HistorySize, time.Duration(size))
```
samarabbas: Is there a reason we need to dual emit?  Can we just have single metric emitted with domainScope?

shreyassrivatsan: Yes, dual emission is actually necessary in the longer term too, specifically for timer style metrics - as they cannot be aggregated between series. If we want to see a distribution of history size across all domains we will need to view the metrics with no domain tag. Alternately we could emit to a domain tag all too.

shreyassrivatsan:
> There are other places in history service where we also emit size related metric. Can you also include domain tag for all those places?
Done.

PR #1602, opened 2019-03-26
scanner: first pass at background job that cleans up tasks/tasklists

### LINKED ISSUE #1436: sql: fix gaps in persistence API wrt deletion of state

venkat1109:
This patch contains an implementation of background scanner workflow that periodically deletes expired tasks / task lists from persistence. Task cleanup is primarily needed for SQL based stores, which don't have support for task TTLs. Tasklist cleanup is needed for both SQL / Cassandra, however, due to existing cassandra schema limitations, this patch only solves the problem for SQL based stores.

There are two parts to the changes in this patch

**Worker**
A new scanner component is added to the current cadence-worker role. This scanner component in turn will spin up a Scanner System Workflow on a cron schedule (daily). The Scanner system workflow is designed to be general purpose i.e. there are other tables we would want to scan in future for detecting anamolies, cleaning up resources etc. All of these will be part of this scanner workflow.

Within the scanner component, there is a task list cleanup module called "TaskListScavenger". This is a daemon which will go over all the task lists in the system and analyze them for expired tasks. The scavenger is also responsible for deleting idle / un-used task lists. The scavenger follows the worker pool pattern - where each task list is a job and a bunch of go-routines process them concurrently. The persistence calls made by scavenger are rate limited to avoid these low priority background jobs from overloading persistence 

Each run of the scanner workflow will invoke an activity which in turn will start the taskListScavenger daemon today. In future, there may be multiple of these activities executing in parallel.

**Persistence**
There is a minor API change to TaskManager.GetTasks() API. This is needed to allow the scavenger to read the first N tasks from a task list sorted by task_id. The change is totally backwards compatible. 

Fixes #1436


common/metrics/defs.go
```diff
@@ -1111,6 +1113,7 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		ArchiverPumpScope:                  {operation: "ArchiverPump"},
 		ArchiverArchivalWorkflowScope:      {operation: "ArchiverArchivalWorkflow"},
 		ArchiverClientScope:                {operation: "ArchiverClient"},
+		TaskListScavengerScope:             {operation: "tasklistscavenger"},
```
shreyassrivatsan: nit: capitalization

common/metrics/defs.go
```diff
@@ -1546,6 +1554,12 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		ArchiverHandleAllRequestsLatency:         {metricName: "archiver_handle_all_requests_latency", oldMetricName: "archiver.handle-all-requests-latency"},
 		ArchiverWorkflowStoppingCount:            {metricName: "archiver_workflow_stopping", oldMetricName: "archiver.workflow-stopping"},
 		ArchiverClientSendSignalFailureCount:     {metricName: "archiver_client_send_signal_error", oldMetricName: "archiver.client-send-signal-error"},
+		TaskProcessedCount:                       {metricName: "task.processed", metricType: Gauge},
```
shreyassrivatsan: names should have `_` in them . not allowed.. 

Adding a test that will catch this. https://github.com/uber/cadence/pull/1603

venkat1109: Fixed !

common/persistence/persistence-tests/matchingPersistenceTest.go
```diff
@@ -128,6 +128,53 @@ func (s *MatchingPersistenceSuite) TestGetDecisionTasks() {
...
+		readLevel int64
+		taskIDs   []int64
+	}{
+		{1, -1, []int64{firstTaskID}},
+		{2, firstTaskID, []int64{firstTaskID + 1, firstTaskID + 2}},
+		{5, firstTaskID + 2, []int64{firstTaskID + 3, firstTaskID + 4}},
+	}
+
+	for i, tc := range testCases {
+		response, err := s.TaskMgr.GetTasks(&p.GetTasksRequest{
```
shreyassrivatsan: you can use substests to iterate through test cases - https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks - provides better readability on test output so you dont need "wrong set of tasks for test case %v"

service/worker/scanner/scanner.go
```diff
@@ -0,0 +1,178 @@
...
+		CronSchedule:                 scannerWFCronSchedule,
+		RetryPolicy:                  &scannerWFRetryPolicy,
+	}
+	_, err := client.StartWorkflow(ctx, opts, scannerWFName)
+	cancel()
+	if err != nil {
+		if _, ok := err.(*shared.WorkflowExecutionAlreadyStartedError); ok {
+			return nil
+		}
+		s.context.logger.Errorf("error starting scanner workflow:%v", err)
```
shreyassrivatsan: use `logger.WithFields("err", err.Error()).Error(..)` - better indexable.

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,140 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package tasklist
+
+import (
+	"sync/atomic"
+	"time"
+
```
shreyassrivatsan: nit: no newline

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,140 @@
...
+		for _, task := range resp.Tasks {
+			nProcessed++
+			if !s.isTaskExpired(task) {
+				return handlerStatusDone, nil
+			}
+		}
+
+		taskID := resp.Tasks[nTasks-1].TaskID
+		_, err = s.completeTasks(key, taskID, nTasks)
```
shreyassrivatsan: combine in a line.

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	delta := time.Now().Sub(state.lastUpdated)
+	if delta < taskListGracePeriod {
+		return
+	}
+	// usually, matching engine is the authoritative owner of a tasklist
+	// and its incorrect for any other entity to mutate task lists (including deleting it)
+	// the delete here is safe because of two reasons:
+	//   - we delete the task list only if the lastUpdated is > 48H. If a task list is idle for
+	//     this amount of time, it will no longer be owned by any host in matching enginer (because
```
shreyassrivatsan: typo: `enginer`  -> `engine`

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,140 @@
...
+	//     of idle timeout). If any new host has to take ownership of this at this time, it can only
+	//     do so by updating the rangeID
+	//   - deleteTaskList is a conditional delete where condition is the rangeID
+	err := s.deleteTaskList(key, state.rangeID)
```
shreyassrivatsan: combine err checking lines?

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		s.logger.Errorf("deleteTaskList error: %v", err)
```
shreyassrivatsan: use `logger.WithFields.Error(...)`

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -0,0 +1,238 @@
...
+	}()
+
+	var wg sync.WaitGroup
+	var pageToken []byte
+
+	// Step #1: process each task list once
+	for {
+		resp, err := s.listTaskList(taskListBatchSize, pageToken)
+		if err != nil {
+			s.logger.Errorf("listTaskList error:%v", err)
```
shreyassrivatsan: log.WithFields

service/worker/scanner/workflow.go
```diff
@@ -0,0 +1,96 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package scanner
+
+import (
+	"context"
+
```
shreyassrivatsan: imports formatting.

common/persistence/persistence-tests/matchingPersistenceTest.go
venkat1109: Good idea. Done

service/worker/scanner/scanner.go
venkat1109: Done

service/worker/scanner/tasklist/handler.go
venkat1109: Done

service/worker/scanner/tasklist/handler.go
venkat1109: Done

service/worker/scanner/tasklist/handler.go
venkat1109: Done

service/worker/scanner/tasklist/handler.go
venkat1109: Done

service/worker/scanner/tasklist/handler.go
venkat1109: Done

service/worker/scanner/tasklist/scavenger.go
venkat1109: Done

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -0,0 +1,222 @@
...
+		logger:   logger,
+		stopC:    stopC,
+		executor: executor.NewFixedSizePoolExecutor(taskListBatchSize),
+	}
+}
+
+// Start starts the scavenger
+func (s *Scavenger) Start() {
+	s.startOnce.Do(func() {
+		s.logger.Info("Tasklist scavenger starting")
```
shreyassrivatsan: nit: maybe some newlines here to differentiate the logic from the metrics etc..
```
		s.logger.Info("Tasklist scavenger starting")

		atomic.StoreInt32(&s.running, 1)
		s.stopWG.Add(1)
		s.executor.Start()
		go s.run()

		s.metrics.IncCounter(metrics.TaskListScavengerScope, metrics.StartedCount)
		s.logger.Info("Tasklist scavenger started")
```

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -0,0 +1,222 @@
...
+		s.metrics.IncCounter(metrics.TaskListScavengerScope, metrics.StartedCount)
+		s.logger.Info("Tasklist scavenger started")
+	})
+}
+
+// Stop stops the scavenger
+func (s *Scavenger) Stop() {
+	if s.Alive() {
+		s.stopOnce.Do(func() {
+			s.metrics.IncCounter(metrics.TaskListScavengerScope, metrics.StoppedCount)
```
shreyassrivatsan: nit: newlines - same as above

service/worker/scanner/scanner.go
```diff
@@ -0,0 +1,179 @@
...
+// New returns a new instance of scanner daemon
+// Scanner is the background sub-system that does full
+// scans of database tables in an attempt to cleanup
+// resources, monitor system anamolies and emit stats
+// for analysis and alerting
+func New(params *BootstrapParams) *Scanner {
+	cfg := params.Config
+	cfg.Persistence.SetMaxQPS(cfg.Persistence.DefaultStore, cfg.PersistenceMaxQPS())
+	zapLogger, err := zap.NewProduction()
+	if err != nil {
```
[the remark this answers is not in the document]
venkat1109: it is a fatal condition, I will fix this.

service/worker/scanner/tasklist/db.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (s *Scavenger) retryForever(op func() error) error {
+	return backoff.Retry(op, retryForeverPolicy, s.isRetryable)
+}
+
+func newRetryForeverPolicy() backoff.RetryPolicy {
+	policy := backoff.NewExponentialRetryPolicy(250 * time.Millisecond)
+	policy.SetExpirationInterval(backoff.NoInterval)
+	policy.SetMaximumInterval(30 * time.Second)
```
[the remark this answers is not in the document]
venkat1109: not really a standard. This varies a bit case by case. For example, with calls that are tied to a customer context (most calls today are in history/frontend/matching), the backoff is in a way, derived from the context deadline. Here, we are dealing with a low priority background task - so if things are continuously failing, the choice made here limits calls to 2 reqs/minute

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,137 @@
...
+const scannerTaskListPrefix = "cadence-sys-scanner"
+
+// deleteHandler handles deletions for a given executorTask list
+func (s *Scavenger) deleteHandler(key *taskListKey, state *taskListState) handlerStatus {
+	var err error
+	var nProcessed, nDeleted int
+
+	defer func() { s.deleteHandlerLog(key, state, nProcessed, nDeleted, err) }()
+
+	for nProcessed < maxTasksPerJob {
```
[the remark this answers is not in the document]
venkat1109: good idea, I will add exactly that

service/worker/scanner/tasklist/scavenger.go
venkat1109: I will fix this, but in general, this is something that's going to be pretty much impossible to enforce across the code base - given that we don't have tools today that do this for us - in addition, the rules are unclear. IMO, not worth going after

shreyassrivatsan: Of course. I don't think this is enforceable or even should be, just something I felt while reading the code - that's why marked as a `nit` - your call whether to fix or not..

service/worker/scanner/workflow.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func init() {
+	workflow.RegisterWithOptions(Workflow, workflow.RegisterOptions{Name: scannerWFName})
+	activity.RegisterWithOptions(TaskListScavengerActivity, activity.RegisterOptions{Name: taskListScavengerActivityName})
+}
+
+// Workflow is the workflow that runs the scanner background sub-system
+func Workflow(ctx workflow.Context) error {
+	opts := workflow.ActivityOptions{
```
samarabbas: Do you think having retry logic here makes sense?

service/worker/scanner/executor/executor.go
```diff
@@ -0,0 +1,140 @@
...
+		TaskCount() int64
+	}
+
+	// fixedPoolExecutor is an implemenation of an executor that uses fixes size
+	// goroutine pool. This executor also supports deferred execution of tasks
+	// for fairness
+	fixedPoolExecutor struct {
+		size        int
+		runQ        *runQueue
+		outstanding int64
```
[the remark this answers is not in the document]
venkat1109: It cannot be because of the definition of what outstanding means here - outstanding is the number of tasks that are currently being executed or waiting to be executed in the executor. The queue cannot keep track of tasks which are dequeued and currently being processed

I will add a comment to clarify this

service/worker/scanner/executor/executor.go
```diff
@@ -0,0 +1,140 @@
...
+			close(e.stopC)
+			atomic.StoreInt32(&e.running, 0)
+			e.stopWG.Wait()
+		})
+	}
+}
+
+// Submit is a blocking call that accepts a task for execution
+func (e *fixedPoolExecutor) Submit(task Task) bool {
+	atomic.AddInt64(&e.outstanding, 1)
```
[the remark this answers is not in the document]
venkat1109: check is redundant - but I added it for clarity

service/worker/scanner/executor/executor.go
```diff
@@ -0,0 +1,140 @@
...
+	return e.runQ.add(task)
+}
+
+// TaskCount returns the total of number of tasks currently outstanding
+func (e *fixedPoolExecutor) TaskCount() int64 {
+	return atomic.LoadInt64(&e.outstanding)
```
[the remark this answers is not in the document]
venkat1109: good idea, added the metric

service/worker/scanner/scanner.go
```diff
@@ -0,0 +1,179 @@
...
+			metricsClient: params.MetricsClient,
+			logger:        params.Logger,
+			tallyScope:    params.TallyScope,
+			zapLogger:     zapLogger,
+		},
+	}
+}
+
+// Start starts the scanner
+func (s *Scanner) Start() error {
```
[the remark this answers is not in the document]
venkat1109: its called everytime activity is started

service/worker/scanner/tasklist/handler.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	atomic.AddInt64(&s.stats.tasklist.nDeleted, 1)
+	s.logger.WithFields(bark.Fields{
+		logging.TagDomainID:     key.DomainID,
+		logging.TagTaskListName: key.Name,
+		logging.TagTaskType:     key.TaskType,
+	}).Info("tasklist deleted")
+}
+
+func (s *Scavenger) deleteHandlerLog(key *taskListKey, state *taskListState, nProcessed int, nDeleted int, err error) {
```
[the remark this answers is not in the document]
venkat1109: its being passed to tryDeleteTaskList() - which is called from this method

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -0,0 +1,222 @@
...
+
+type (
+	// Scavenger is the type that holds the state for task list scavenger daemon
+	Scavenger struct {
+		db        p.TaskManager
+		executor  executor.Executor
+		metrics   metrics.Client
+		logger    bark.Logger
+		stats     stats
+		startOnce sync.Once
```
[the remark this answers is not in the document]
venkat1109: good call, I changed the code to use cas for consistency

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+		task struct {
+			nProcessed int64
+			nDeleted   int64
+		}
+	}
+
+	// executorTask is a runnable task that adheres to the executor.Task interface
+	// for the scavenger, each of this task processes a single task list
+	executorTask struct {
```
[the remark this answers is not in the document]
venkat1109: this is a struct definition within the tasklist package and that's the reason its named - executorTask and not taskListExecutorTask - tasklist.taskListExecutorTask is redundant IMO

service/worker/scanner/executor/executor.go
```diff
@@ -0,0 +1,140 @@
...
+			atomic.StoreInt32(&e.running, 0)
+			e.stopWG.Wait()
+		})
+	}
+}
+
+// Submit is a blocking call that accepts a task for execution
+func (e *fixedPoolExecutor) Submit(task Task) bool {
+	atomic.AddInt64(&e.outstanding, 1)
+	return e.runQ.add(task)
```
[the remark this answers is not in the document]
venkat1109: done

service/worker/scanner/executor/runq.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (rq *runQueue) addAndDefer(task Task) {
+	rq.deferredTasks.add(task)
+}
+
+// remove returns the next task from run queue
+// if there are no active tasks in runq, a task
+// from deferredQ is returned
+func (rq *runQueue) remove() (Task, bool) {
```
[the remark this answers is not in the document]
venkat1109: discussed offline - there is no other sane way - i will add comments to clarify the intent here

service/worker/scanner/executor/runq.go
```diff
@@ -0,0 +1,118 @@
...
+	select {
+	case job, ok := <-rq.tasks:
+		if !ok {
```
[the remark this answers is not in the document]
venkat1109: not necessary - only reason task channel can be closed - is when the executor is shutting down

service/worker/scanner/scanner.go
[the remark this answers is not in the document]
venkat1109: done

service/worker/scanner/tasklist/scavenger.go
venkat1109: this code no longer exists

service/worker/scanner/executor/runq.go
```diff
@@ -0,0 +1,118 @@
...
+	select {
+	case rq.tasks <- task:
+		return true
+	case <-rq.stopC:
+		return false
+	}
+}
+
+func (rq *runQueue) addAndDefer(task Task) {
+	rq.deferredTasks.add(task)
```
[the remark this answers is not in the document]
venkat1109: good catch - modified the code to let the caller specify maxDeferredTasks

service/worker/scanner/workflow.go
venkat1109: right now, it doesn't make a lot of sense - since there is only one activity and the workflow retry should suffice - however, when we multiple activities, it does make sense - anyways, i added activity retry as well

service/worker/scanner/workflow.go
```diff
@@ -0,0 +1,95 @@
...
+		ScheduleToStartTimeout: 5 * time.Minute,
+		StartToCloseTimeout:    2 * 24 * time.Hour,
+		HeartbeatTimeout:       5 * time.Minute,
+	}
+	future := workflow.ExecuteActivity(workflow.WithActivityOptions(ctx, opts), taskListScavengerActivityName)
+	future.Get(ctx, nil)
```
mfateev: Why we don't fail workflow on activity error?

service/worker/scanner/executor/executor_test.go
```diff
@@ -0,0 +1,91 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package executor
+
+import (
+	"testing"
```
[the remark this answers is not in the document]
venkat1109: I would like to follow this - but it looks like the IDE auto-formatting somehow messes things up - I never add imports directly - its all automatically added by my IDE

service/worker/scanner/workflow.go
venkat1109: yup - updated the code.

PR #1611, opened 2019-03-27
This PR aims to refactor the existing replication task processing, as…

wxing1292:
… well as improve performance when applying large number of replication tasks concurrently to one workflow.

* Implement sequential task processing logic & UT
* Refactor replication worker task processing logic
* Handle activity / history replication task in sequence & UT
* All replication retryable errors have upper attempt, when exceeding the attempt, replication message will be moved to DLQ


common/collection/initerface.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
samarabbas: fix file name.

common/backoff/retrypolicy.go
```diff
@@ -27,9 +27,10 @@ import (
 )
 
 const (
-	// NoInterval represents Maximim interval
-	NoInterval                      = 0
-	done              time.Duration = -1
+	// NoInterval represents Maximum interval
+	NoInterval = 0
+	// Done indicate that there should be no retry
+	Done              time.Duration = -1
```
venkat1109: why does this need to be public ?

common/collection/interface.go
```diff
@@ -0,0 +1,99 @@
...
+		RemoveIf(key interface{}, fn PredicateFunc) bool
+		// Iter returns an iterator to the map
+		Iter() MapIterator
+		// Size returns the number of items in the map
+		Size() int
+	}
+
+	// MapIterator represents the interface for
+	// map iterators
+	MapIterator interface {
```
venkat1109: I don't see this used anywhere - let's remove it and add it back when we need it

common/collection/priorityQueue.go
```diff
@@ -0,0 +1,96 @@
...
+// THE SOFTWARE.
+
+package collection
+
+import (
+	"container/heap"
+)
+
+type (
+	priorityQueueImpl struct {
```
venkat1109: same comment as above - let's remove this code and add it back when we need it

common/task/interface.go
```diff
@@ -0,0 +1,59 @@
...
+		HandleErr(err error) error
+		// RetryErr check whether to retry after HandleErr(Execute())
+		RetryErr(err error) bool
+		// Ack marks the task as successful completed
+		Ack()
+		// Nack marks the task as unsuccessful completed
+		Nack()
+	}
+
+	// SequentialTask is the interface for tasks which should be executed sequentially
```
venkat1109: i would add some more comments explaining why these tasks are special / called sequential, when to use them

common/task/interface.go
```diff
@@ -0,0 +1,59 @@
...
+// THE SOFTWARE.
+
+package task
+
+import (
+	"github.com/uber/cadence/common"
+)
+
+type (
+	// SequentialTaskProcessor is the generic coroutine pool interface
```
venkat1109: I prefer not using the "coroutine" terminology - you are simply using standard goroutine pools - lets use standard terminology - goroutines, worker, workerPool, poolSize etc

common/task/interface.go
```diff
@@ -0,0 +1,59 @@
...
+	// which process sequential task
+	SequentialTaskProcessor interface {
```
venkat1109: can you add additional comments what "sequential" means, what it does and whey it should be used ?

common/task/interface.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// SequentialTask is the interface for tasks which should be executed sequentially
+	SequentialTask interface {
+		// QueueID return the ID of the queue this task belongs to
+		QueueID() interface{}
+		// TaskID return the ID of the task, this task ID is used for sorting
+		TaskID() int64
+		// HashCode return the hash code of the task
+		HashCode() uint32
```
venkat1109: QueueID() and HashCode() need to satisfy certain properties and are closely related but this definition makes it look like they are simply task properties i.e. all tasks related to same workflow must produce same queueID and same hashCode is not obvious. And when queueIDs are equal, hashCode MUST be equal as well i.e must be Comparable, in java terminology

How about you specifically separating out the partition key part and adding comments about requirements for partition key
```
SequentialTaskPartitionKey interface {
    Key() interface // MUST be go primitive type or struct with primitive types 
    HashCode() uint32
}
```

And then having your SequentialTask interface like 
```
SequentialTask interface {
    TaskID() int64
     PartitionKey() SequentialTaskPartitionKey
}
```

common/task/sequentialTaskProcessor.go
```diff
@@ -0,0 +1,197 @@
...
+		coroutineTaskQueues map[int]chan SequentialTask
+		logger              bark.Logger
+	}
+
+	// SequentialTasks slice of SequentialTask
+	SequentialTasks []SequentialTask
+)
+
+const (
+	coroutineTaskQueueSize           = 100
```
venkat1109: lets define all of these queue / worker pool size in a single place for the replicator

common/task/sequentialTaskProcessor.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+type (
+	sequentialTaskProcessorImpl struct {
+		status       int32
+		shutdownChan chan struct{}
+		waitGroup    sync.WaitGroup
+
+		coroutineSize       int
+		coroutineTaskQueues map[int]chan SequentialTask
```
venkat1109: just use a fixed size slice and array index can be your taskQueueID

common/task/sequentialTaskProcessor.go
```diff
@@ -0,0 +1,197 @@
...
+			err = task.HandleErr(err)
+			if err == nil || !task.RetryErr(err) {
+				break TaskProcessingLoop
+			}
+		}
+	}
+
+	if err != nil {
+		task.Nack()
+	} else {
```
venkat1109: remove else and return - https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

service/worker/replicator/replicationTask.go
```diff
@@ -0,0 +1,291 @@
...
+	if !ok || retryErr.GetRunId() == "" {
+		return err
+	}
+
+	t.metricsClient.IncCounter(metrics.HistoryRereplicationByActivityReplicationScope, metrics.CadenceClientRequests)
+	stopwatch := t.metricsClient.StartTimer(metrics.HistoryRereplicationByActivityReplicationScope, metrics.CadenceClientLatency)
+	defer stopwatch.Stop()
+
+	// this is the retry error
+	beginRunID := retryErr.GetRunId()
```
venkat1109: consider factoring out this common code

common/task/sequentialTaskProcessor.go
```diff
@@ -0,0 +1,197 @@
...
+	SequentialTasks []SequentialTask
+)
+
+const (
+	coroutineTaskQueueSize           = 100
+	coroutineTaskProcessingBatchSize = 100
+)
+
+// NewSequentialTaskProcessor create a new sequential tasks processor
+func NewSequentialTaskProcessor(coroutineSize int, logger bark.Logger) SequentialTaskProcessor {
```
shreyassrivatsan: this seems like a multi-queue version of the scanner executor in #1602.. I have a feeling you can build this on top of that.. That way we do not have two different queuing and processing implementations doing the similar thing..

common/collection/interface.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
shreyassrivatsan: nit: filename `interfaces.go`

common/collection/priorityQueue.go
```diff
@@ -0,0 +1,96 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package collection
```
shreyassrivatsan: is there an implementation we can take, rather than writing this? I see https://godoc.org/github.com/golang-collections/go-datastructures/queue, but not sure how good it is..

common/task/interface.go
```diff
@@ -0,0 +1,59 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package task
+
+import (
+	"github.com/uber/cadence/common"
```
shreyassrivatsan: nit: do not need single import within `( )`

common/task/sequentialTaskProcessor.go
shreyassrivatsan: I have a feeling this should be implemented as a queue of queues. A simple task queue just processing tasks.

And a second wrapper queue that has a map of the single queues - both adhering to the same `Queue` interface. That will separate the two styles of queues from each other - make it easy understand and debug..

service/worker/replicator/replicationTask.go
```diff
@@ -0,0 +1,291 @@
...
+// THE SOFTWARE.
+
+package replicator
+
+import (
+	"context"
+	"time"
+
+	"github.com/dgryski/go-farm"
+
```
shreyassrivatsan: spurious newlines between imports.

service/worker/replicator/replicationTask.go
```diff
@@ -0,0 +1,291 @@
...
+		workflowReplicationTask
+		req *h.ReplicateEventsRequest
+	}
+)
+
+const (
+	replicationTaskRetryDelay = 500 * time.Microsecond
+)
+
+func newActivityReplicationTask(task *replicator.ReplicationTask, msg messaging.Message, logger bark.Logger,
```
shreyassrivatsan: nit: stacking the arguments makes it more readable.. 

```
func newActivityReplicationTask(
    task *replicator.ReplicationTask, 
    msg messaging.Message, 
    logger bark.Logger,
    config *Config, 
    historyClient history.Client, 
    metricsClient metrics.Client,
    historyRereplicator xdc.HistoryRereplicator,
) *activityReplicationTask {
```

Also, ordering of the params is weird. conf, metrics, logger should be next to each other. And conf is normally first or second for constructors.

service/worker/replicator/replicationTask.go
```diff
@@ -0,0 +1,291 @@
...
+			StartedId:         attr.StartedId,
+			StartedTime:       attr.StartedTime,
+			LastHeartbeatTime: attr.LastHeartbeatTime,
+			Details:           attr.Details,
+			Attempt:           attr.Attempt,
+		},
+	}
+}
+
+func newHistoryReplicationTask(task *replicator.ReplicationTask, msg messaging.Message, sourceCluster string, logger bark.Logger,
```
shreyassrivatsan: same as above

service/worker/replicator/replicator.go
```diff
@@ -25,6 +25,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/uber/cadence/common/task"
```
shreyassrivatsan: spurious newlines

common/task/sequentialTaskProcessor.go
shreyassrivatsan: Actually, that may not be possible, but maybe worth you taking a look and see if you see similarities

common/task/interface.go
venkat1109: Alternatively, you can also consider something like this
```
type SequentialTaskPartitionKey definition.WorkflowIdentifier

func (p *SequentialTaskPartitionKey) HashCode() uint32 {...}

SequentialTask interface {
    TaskID() int64
    PartitionKey() SequentialTaskPartitionKey
}
 
```

common/task/sequentialTaskProcessor.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+	}
+
+	select {
+	case <-t.shutdownChan:
+		return
+	case task := <-coroutineTaskQueue:
+		indexTasks(task)
+	BufferLoop:
+		for i := 0; i < coroutineTaskProcessingBatchSize-1; i++ {
```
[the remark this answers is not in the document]
wxing1292: the task batch processing size should be independent of Kafka 
this logic aims to be a generic task processing logic.

service/worker/replicator/processor.go
```diff
@@ -201,157 +178,123 @@ func (p *replicationTaskProcessor) messageProcessLoop(workerWG *sync.WaitGroup,
...
 	sw := p.metricsClient.StartTimer(metrics.DomainReplicationTaskScope, metrics.ReplicatorLatency)
 	defer sw.Stop()
 
 	logger.Debugf("Received domain replication task %v.", task.DomainTaskAttributes)
-	return p.domainReplicator.HandleReceivingTask(task.DomainTaskAttributes)
+	err := p.domainReplicator.HandleReceivingTask(task.DomainTaskAttributes)
+	if err != nil {
+		return err
+	}
+	p.ackMsg(msg, logger)
```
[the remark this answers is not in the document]
wxing1292: i disagree,
decodeMsgAndSubmit aim to pass down the task to for further processing, meaning that decodeMsgAndSubmit can only do nack.

this function (logic) happens to be light weight, so the coroutine will execute the task immediately.
only the logic which can finish the task can do ack

service/worker/replicator/replicationTask.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	historyReplicationTask struct {
+		workflowReplicationTask
+		req *h.ReplicateEventsRequest
+	}
+)
+
+const (
+	replicationTaskRetryDelay = 500 * time.Microsecond
```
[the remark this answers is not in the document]
wxing1292: there is an upper limit of how many times a task can be retried.
the worker should use minimum amount of time applying replication task, reducing the backlog as well as keeping the passive side "up to date" with active side.

if the backoff time is too short, we can come back and tweak it

common/backoff/retrypolicy.go
wxing1292: previously i was trying to make the priority queue do rate limiting / back off and retry.

i can change it back

common/backoff/retry.go
```diff
@@ -48,7 +48,7 @@ func (c *ConcurrentRetrier) Throttle() {
 }
 
 func (c *ConcurrentRetrier) throttleInternal() time.Duration {
-	next := done
+	next := Done
```
venkat1109: can you revert this to private again ?

PR #1614, opened 2019-03-28
Handle concurrent archivals more gracefully

andrewjdawson2016:
- Do check for blob exists before reading blob from persistence, this avoids doing unnecessary persistence reads.
- Add GetTags method blobstore interface
- Add unit tests around concurrent load case
- Remove metrics and logging in case that blob already exists (this is the happy path no metrics or logging is needed).
- Add HistoryBlobReader which knows how to read persistence and get an arbitrary archival history blob.
- Introduce sanity checks to make sure deterministic blob construction is working as expected. This is an important metric to alert on.


common/blobstore/blob/blob.go
```diff
@@ -34,21 +34,46 @@ func NewBlob(body []byte, tags map[string]string) *Blob {
 	}
 }
 
-// DeepCopy returns a deep copy of input blob
-func DeepCopy(blob *Blob) *Blob {
-	if blob == nil {
+// DeepCopy returns a deep copy of blob
+func (b *Blob) DeepCopy() *Blob {
+	if b == nil {
```
samarabbas: Looks like this nil check is not needed anymore.

common/blobstore/metricClient.go
```diff
@@ -68,6 +68,19 @@ func (c *metricClient) Download(ctx context.Context, bucket string, key blob.Key
...
 
+func (c *metricClient) GetTags(ctx context.Context, bucket string, key blob.Key) (map[string]string, error) {
+	c.metricsClient.IncCounter(metrics.BlobstoreClientGetTagsScope, metrics.CadenceClientRequests)
+
+	sw := c.metricsClient.StartTimer(metrics.BlobstoreClientGetTagsScope, metrics.CadenceClientLatency)
+	resp, err := c.client.GetTags(ctx, bucket, key)
+	sw.Stop()
+
+	if err != nil {
+		c.metricsClient.IncCounter(metrics.BlobstoreClientGetTagsScope, metrics.CadenceClientFailures)
```
samarabbas: General pattern is to report errors which count towards availability of that API.  User errors should not be counted as errors like "BlobNotExist".

service/worker/archiver/activities.go
```diff
@@ -98,56 +93,75 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
 			logging.LogFailArchivalUploadAttempt(logger, err, reason, bucket, key.String())
 			return cadence.NewCustomError(errConstructBlob)
 		}
+		if runConstTest {
+			existingBlob, err := downloadBlob(ctx, blobstoreClient, bucket, key)
+			if err != nil {
+				logger.WithError(err).Error("failed to download blob for deterministic construction verification")
+				metricsClient.IncCounter(metrics.ArchiverUploadHistoryActivityScope, metrics.ArchiverCouldNotRunDeterministicConstructionCheckCount)
+			} else if !blob.Equal(existingBlob) {
+				logger.Error("deterministic construction check failed")
```
samarabbas: Can you make sure this logger is tagged correctly with all necessary fields like domain, workflow, run, etc?

service/worker/archiver/activities.go
```diff
@@ -244,11 +258,27 @@ func uploadBlob(ctx context.Context, blobstoreClient blobstore.Client, bucket st
 	return nil
 }
 
+func downloadBlob(ctx context.Context, blobstoreClient blobstore.Client, bucket string, key blob.Key) (*blob.Blob, error) {
+	bCtx, cancel := context.WithTimeout(ctx, blobstoreTimeout)
```
samarabbas: I'm a little worried that we do retries in a tight loop without backoff.  Is there a reason you are not using our regular retry abstraction for retries?

service/worker/archiver/activities.go
```diff
@@ -244,11 +258,27 @@ func uploadBlob(ctx context.Context, blobstoreClient blobstore.Client, bucket st
...
+	blob, err := blobstoreClient.Download(bCtx, bucket, key)
+	cancel()
+	for err != nil {
+		activity.RecordHeartbeat(ctx)
```
samarabbas: Do you think moving heartbeating to separate goroutine makes sense?

common/blobstore/blob/blob.go
andrewjdawson2016: The receiver could still potentially be nil. It is a bug if this method is called anywhere with a nil receiver but it's possible.

common/blobstore/metricClient.go
andrewjdawson2016: I based this off the other metrics clients which follow a pattern like this

```
func (c *metricClient) DescribeDomain(
	ctx context.Context,
	request *shared.DescribeDomainRequest,
	opts ...yarpc.CallOption,
) (*shared.DescribeDomainResponse, error) {

	c.metricsClient.IncCounter(metrics.FrontendClientDescribeDomainScope, metrics.CadenceClientRequests)

	sw := c.metricsClient.StartTimer(metrics.FrontendClientDescribeDomainScope, metrics.CadenceClientLatency)
	resp, err := c.client.DescribeDomain(ctx, request, opts...)
	sw.Stop()

	if err != nil {
		c.metricsClient.IncCounter(metrics.FrontendClientDescribeDomainScope, metrics.CadenceClientFailures)
	}
	return resp, err
}
```

In this case DescribeDomain could return error entity not exists but we still emit a metric based on this error.

I agree with you that these errors should not count against API availability but I think changing this involves changing how we do all our clients?

service/worker/archiver/activities.go
andrewjdawson2016: The blobstore client itself is a retryable blobstore client. So each call to Download will kick off a series of retries which have backoff between them. Is this acceptable or do you want me to use the retryable client inside of an operation passed to Retry function?

service/worker/archiver/activities.go
andrewjdawson2016: My first implementation heartedbeated within a goroutine. I had a conversation with Maxim about doing it this way instead of the goroutine approach. The issue with doing it in a separate goroutine is lets say the activity gets stuck - some RPC call is hanging due to a bug. Since the heartbeat is in a separate go routine we will not get a heartbeat timeout. Instead we will have to wait for the context to timeout.

service/worker/archiver/activities.go
andrewjdawson2016: It is at the top of method we assign logger to result of this method

```
func tagLoggerWithRequest(logger bark.Logger, request ArchiveRequest) bark.Logger {
	return logger.WithFields(bark.Fields{
		logging.TagArchiveRequestDomainID:             request.DomainID,
		logging.TagArchiveRequestWorkflowID:           request.WorkflowID,
		logging.TagArchiveRequestRunID:                request.RunID,
		logging.TagArchiveRequestEventStoreVersion:    request.EventStoreVersion,
		logging.TagArchiveRequestBranchToken:          string(request.BranchToken),
		logging.TagArchiveRequestNextEventID:          request.NextEventID,
		logging.TagArchiveRequestCloseFailoverVersion: request.CloseFailoverVersion,
	})
}
```

PR #1615, opened 2019-03-28
Expose more pending activity information

yycptt:
Currently pendingActivityInfo returned by describeWorkflowExecution doesn't contain information related activity's retry policy. This PR exposes information including attempts, maximumAttempts, expirationTime, etc. 

[Issue 1574](https://github.com/uber/cadence/issues/1574)


service/history/historyEngine.go
```diff
@@ -894,8 +894,18 @@ func (e *historyEngineImpl) DescribeWorkflowExecution(ctx context.Context,
 				return nil, &workflow.InternalServiceError{Message: "Unable to get activity schedule event."}
 			}
 			p.ActivityType = scheduledEvent.ActivityTaskScheduledEventAttributes.ActivityType
-			p.LastStartedTimestamp = common.Int64Ptr(ai.StartedTime.UnixNano())
-			p.Attempt = common.Int32Ptr(ai.Attempt)
+			if state == workflow.PendingActivityStateScheduled {
+				p.ScheduledTimestamp = common.Int64Ptr(ai.ScheduledTime.UnixNano())
+			} else {
```
mfateev: Why is it else? Logically having both does make sense.

service/history/historyEngine.go
```diff
@@ -894,8 +894,18 @@ func (e *historyEngineImpl) DescribeWorkflowExecution(ctx context.Context,
...
+				p.LastStartedTimestamp = common.Int64Ptr(ai.StartedTime.UnixNano())
+			}
+			if ai.HasRetryPolicy {
+				p.Attempt = common.Int32Ptr(ai.Attempt)
+				p.ExpirationTimestamp = common.Int64Ptr(ai.ExpirationTime.UnixNano())
```
andrewjdawson2016: Why do we not do the is not zero check on ExpirationTime like we are doing for MaximumAttempts? The user can give either one or both.

tools/cli/commands.go
```diff
@@ -857,8 +862,15 @@ func convertDescribeWorkflowExecutionResponse(resp *shared.DescribeWorkflowExecu
 			ActivityID:             pa.ActivityID,
 			ActivityType:           pa.ActivityType,
 			State:                  pa.State,
-			HeartbeatDetails:       common.StringPtr(string(pa.HeartbeatDetails)),
-			LastHeartbeatTimestamp: common.StringPtr(convertTime(pa.GetLastHeartbeatTimestamp(), false)),
+			ScheduledTimestamp:     timestampPtrToStringPtr(pa.ScheduledTimestamp, false),
+			LastStartedTimestamp:   timestampPtrToStringPtr(pa.LastStartedTimestamp, false),
+			LastHeartbeatTimestamp: timestampPtrToStringPtr(pa.LastHeartbeatTimestamp, false),
+			Attempt:                pa.Attempt,
+			MaximumAttempts:        pa.MaximumAttempts,
```
andrewjdawson2016: It seems like we want to do some checks here to see if these are zero. If they are zero then just use some string "not set" that way its clear that entering 0 was a not a bug but rather that there just is no termination case based on that condition.

service/history/historyEngine.go
yycptt: I believe when the activity state is scheduled, the started time will be reset to 0. If it has already started, then the scheduled time is not important.

andrewjdawson2016: If you are in PendingActivityStateScheduled will StartTime be set? Will it be set to the last attempt's start time? 

If StartedTime does have a meaningful value when you are in PendingActivityStateScheduled, then I agree we should return it from here always.

service/history/historyEngine.go
yycptt: The expiration time will be set to activity scheduleToCloseTimeout if user doesn't specify or the specified time is shorter than the timeout.

service/history/historyEngine.go
yycptt: In PendingActivityStateScheduled state, the startTime is always 0.

tools/cli/commands.go
yycptt: I agree with you. The problem here is how these information are printed. Currently we are using json.MarshalIndent to do that. To add that check, we basically need to do all the formatting ourselves. That's also why at sever side, I carefully ensure that a nil pointer is returned for these fields instead of a pointer to value 0.

service/history/historyEngine.go
andrewjdawson2016: I see, makes sense.

tools/cli/commands.go
andrewjdawson2016: Sounds reasonable.

service/history/historyEngine.go
andrewjdawson2016: In that case I am of the opinion that your existing implementation is correct.

PR #1618, opened 2019-03-29
Logging refactoring: define interface and migrate existing tags and tag values

longquanzheng:
```
=== RUN   TestDefaultLogger
{"level":"info","msg":"test info","error":"test error","wf-action":"add-activitytask-canceled-event","logging-call-at":"logger_test.go:60"}
 60
--- PASS: TestDefaultLogger (0.00s)
=== RUN   TestThrottleLogger
time="2019-04-11T12:12:02-07:00" level=info msg="Get dynamic config" configDefaultVal=1 configName=frontend.rps configVal=1
{"level":"info","msg":"test info","error":"test error","wf-action":"add-activitytask-canceled-event","logging-call-at":"logger_test.go:94"}
 94
--- PASS: TestThrottleLogger (0.00s)
PASS
```


common/log/interface.go
```diff
@@ -0,0 +1,90 @@
...
+	FatalWithLifecycle(msg string, lifecycle tagValueTypeSysLifecycle, tags ...Fields)
+}
+
+// we intentionally make tagType module-private to enforce using pre-defined tag keys
+type tagType struct {
+	name      string
+	valueType valueTypeEnum
+}
+
+type valueTypeEnum string
```
andrewjdawson2016: Why use a string to represent the enum instead of the more common pattern of using an int?

common/log/tagValues.go
```diff
@@ -0,0 +1,132 @@
...
+	TagValueActionWorkflowCancelFailed            tagValueTypeWorkflowAction = "add-workflow-execution-cancel-failed-event"
+	TagValueActionWorkflowSignalRequested         tagValueTypeWorkflowAction = "add-workflow-execution-signal-requested-event"
+	TagValueActionWorkflowSignalFailed            tagValueTypeWorkflowAction = "add-workflow-execution-signal-failed-event"
+	TagValueActionUnknownEvent                    tagValueTypeWorkflowAction = "add-unknown-event"
+)
+
+// Pre-defined values for TagWorkflowListFilterType
+const (
+	TagValueListWorkflowFilterByID     tagValueTypeWorkflowListFilterType = "WID"
+	TagValueListWorkflowFilterByType   tagValueTypeWorkflowListFilterType = "WType"
```
andrewjdawson2016: The difference in casing looks odd to me it feels like it should be (wtype and status) or (WType and Status). This is just opinion though.

common/log/tags.go
```diff
@@ -0,0 +1,154 @@
...
+	TagArchivalRequestBranchToken          = newTagType("archival-request-branch-token", valueTypeString)
+	TagArchivalRequestNextEventID          = newTagType("archival-request-next-event-id", valueTypeInteger)
+	TagArchivalRequestCloseFailoverVersion = newTagType("archival-request-close-failover-version", valueTypeInteger)
+
+	// archival tags (blob tags)
+	TagArchivalBucket  = newTagType("archival-bucket", valueTypeString)
+	TagArchivalBlobKey = newTagType("archival-blob-key", valueTypeString)
+
+	// archival tags (file blobstore tags)
+	TagArchivalFileBlobstoreBlobPath     = newTagType("archival-file-blobstore-blob-path", valueTypeString)
```
andrewjdawson2016: The tags relating to archival file store can be deleted. They are not used.

common/log/interface.go
longquanzheng: Because that value will be actually passed into ELK and need to be human readable. One bad example is our workflow-event(the name is confusing, we will call it component lifecycle). It use int and turns out super painful for us to search in ELK. See https://github.com/uber/cadence/blob/master/common/logging/events.go#L31
Another good example is Chronos, after Venkat's refactoring, we are using string as value

common/log/tagValues.go
longquanzheng: Good catches. I just copy it from existing code without noticing that. I will change it as you say

common/log/tags.go
longquanzheng: I see, I will follow up with you later.

common/log/throttle.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (tl *throttledLogger) Fatal(msg string, tags ...tag.Tag) {
+	tl.rateLimit(func() {
+		tl.log.Fatal(msg, tags...)
+	})
+}
+
+// Return a logger with the specified key-value pairs set, to be included in a subsequent normal logging call
+func (tl *throttledLogger) WithFields(tags ...tag.Tag) Logger {
```
mfateev: I think it should be renamed to WithTags

common/log/logger.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/uber/cadence/common/log/tag"
+	"go.uber.org/zap"
+)
+
+// Logger is our abstraction for logging
+// Usage examples:
+//  import "github.com/uber/cadence/common/log/tag"
+//  1) logger = logger.WithFields(
+//          tag.TagWorkflowEventID( 123),
```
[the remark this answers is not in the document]
longquanzheng: Good catch, will do it.

common/log/tag/keys.go
```diff
@@ -0,0 +1,507 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
longquanzheng: Sounds good

common/log/throttle.go
longquanzheng: OK

common/log/logger_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	var zapLogger *zap.Logger
+	zapLogger = zap.NewExample()
+
+	dc := dynamicconfig.NewNopClient()
+	cln := dynamicconfig.NewCollection(dc, bark.NewLoggerFromLogrus(logrus.New()))
+	logger := NewThrottledLogger(zapLogger, cln.GetIntProperty(dynamicconfig.FrontendRPS, 1))
+	preCaller := caller(1)
+	logger.WithTags(tag.Error(fmt.Errorf("test error"))).Info("test info", tag.WorkflowAction(tag.ValueActionActivityTaskCanceled))
+
```
mfateev: Can we make values be of Tag type? Then `tag.WorkflowAction(tag.ValueActionActivityTaskCanceled))` becomes  `tag.WorkflowActionActivityTaskCanceled`

longquanzheng: That's really a good idea. Only one concern is I want to keep tag.WorkflowAction in the namespace so that no one mess up the tag key(which will also cause conflict). But another idea is to make it module private, "tag.WorkflowAction" => "tag.workflowAction, and using that to define values.

longquanzheng: Done

common/log/tag/interface.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (t *tagImpl) GetDuration() time.Duration {
+	return t.valueDuration
+}
+
+func (t *tagImpl) GetTime() time.Time {
+	return t.valueTime
+}
+
+func (t *tagImpl) GetObject() interface{} {
```
venkat1109: In go naming convention, this is called GetInterface()

common/log/throttle.go
```diff
@@ -0,0 +1,136 @@
...
+// THE SOFTWARE.
+
+package log
+
+import (
+	"sync/atomic"
+
+	"go.uber.org/zap"
+
+	"github.com/uber/cadence/common/log/tag"
```
venkat1109: imports formatting

common/log/tag/interface.go
```diff
@@ -0,0 +1,192 @@
...
+	ValueTypeTime valueTypeEnum = 7
+	// ValueTypeObject will be converted into string by fmt.Sprintf("%+v")
+	ValueTypeObject valueTypeEnum = 8
+)
+
+// keep this module private
+type valueTypeEnum int
+
+// keep this module private
+type tagImpl struct {
```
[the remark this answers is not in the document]
shreyassrivatsan: +1 to Venkat's suggestion.

common/log/logger_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"strings"
```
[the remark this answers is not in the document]
longquanzheng: sounds good

common/log/tag/interface.go
```diff
@@ -0,0 +1,192 @@
...
+	GetTime() time.Time
+	GetObject() interface{}
+}
+
+var _ Tag = (*tagImpl)(nil)
+
+// the tag value types supported
+const (
+	// ValueTypeString is string value type
+	ValueTypeString valueTypeEnum = 1
```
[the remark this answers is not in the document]
longquanzheng: yes I think iota should work, I changed that from string and forgot about it.

common/log/tag/interface.go
```diff
@@ -0,0 +1,192 @@
...
+
+// keep this module private
+type valueTypeEnum int
+
+// keep this module private
+type tagImpl struct {
+	key       string
+	valueType valueTypeEnum
+
+	valueString   string        //1
```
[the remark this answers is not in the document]
longquanzheng: Sure

common/log/tag/interface.go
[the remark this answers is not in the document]
longquanzheng: That looks like a great idea.
So the reason behind this is we decided to solve the problem of tag conflict which has caused several ELIK ingestion issue in production which made us lost lots of logs. By using this interface, we want to forbidden people from using random string as key, and always generate tag from this logging module.

This idea should also work in the similar with. But only one concern: that it is still possible for people to get around of the restriction:
```
logger.Info("abc", tag.Tag(zap.String("randome name", "some value"))
```
I think it is still OK to make it weaker restriction but gain some perf benefits here. Also, I want to use
```
type Tag *zap.Field
```
With pointer it is even harder to hack into, and more efficient as zap.Field is a struct instead of interface. Without pointer we will end up creating a lot of small struct as well.
Also I will add some comments on this type to let people know that they should NOT do something like
```
field := zap.String("randome name", "some value")
logger.Info("abc", tag.Tag(&field)
```

common/log/tag/interface.go
longquanzheng: OK. btw all of those will become GetField()

common/log/throttle.go
longquanzheng: OK

common/log/tag/interface.go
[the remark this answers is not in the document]
shreyassrivatsan: As Venkat mentioned, if you end up returning a pointer to a struct that will end up on the heap and have to be garbage collected. If you return a struct it all just remains on the stack and doesnt required garbage collection so is a lot more efficient for small structs. That is the reason why zap returns the struct rather than pointer in its corresponding implementation.. https://github.com/uber-go/zap/blob/master/field.go#L37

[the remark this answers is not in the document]
venkat1109: @longquanzheng  - if you are concerned about callers overriding your restriction with tag.Tag(zap.String(...)), I would suggest the following instead of using *zap.Field as a type alias. 

Try
```
type struct Tag {
   field zap.Field
}
```

Because field is private, there is no way for callers outside your package to build Tag without calling one of our methods. 

As for "Without pointer we will end up creating a lot of small struct as well." 

It's ok to create these structs because these are returned by value instead of being returned as reference i.e. they will all be allocated in stack as opposed to heap, which is the optimization  that we are trying to achieve here. And that's also the reason Zap returns zap.Field by value and not by reference - https://github.com/uber-go/zap/blob/master/field.go#L100

Checkout this for more info - https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-escape-analysis.html

common/log/tag/interface.go
venkat1109: yeah, this comment becomes outdated. On the naming side, one more nit. GetFoo() is not standard go idiom. Go idiom is to just have a method called Field() that returns zap.Field

common/log/tag/interface.go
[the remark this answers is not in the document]
longquanzheng: Got it, let's do this way.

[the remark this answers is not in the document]
longquanzheng: It think what you want to say is:
```
type Tag tagImpl

type tagImpl struct{
  field zap.Field
}
```

common/log/logger_test.go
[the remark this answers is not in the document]
longquanzheng: done

PR #1620, opened 2019-03-29
Trying out BuildKite

dmetzgar:
Still having trouble with Cassandra. It installs as root but the integration tests are unable to connect to it.


dmetzgar:
Is anyone able to see the build pipeline? https://buildkite.com/uberopensource/cadence-server

wxing1292:
close in favor of #1643

PR #1625, opened 2019-03-29
Integration test refactor part 2

meiliang86:
- Allow create and start test cluster from config.
- Move direct test cluster calls to go through frontend.
- Allow user to pass in frontend address and skip creating test cluster.


host/onebox.go
```diff
@@ -67,10 +67,12 @@ const maxRpJoinTimeout = 30 * time.Second
 
 var (
 	// EnableEventsV2 indicates whether events v2 is enabled for integration tests
-	EnableEventsV2     = flag.Bool("eventsV2", false, "run integration tests with eventsV2")
-	enableGlobalDomain = flag.Bool("enableGlobalDomain", false, "run integration tests with global domain")
-	enableArchival     = flag.Bool("enableArchival", false, "run integration tests with archival")
-	topicName          = []string{"active", "standby"}
+	EnableEventsV2        = flag.Bool("eventsV2", false, "run integration tests with eventsV2")
+	frontendAddress       = flag.String("frontendAddress", "", "host:port for cadence frontend service")
+	TestClusterConfigFile = flag.String("TestClusterConfigFile", "", "test cluster config file location")
```
venkat1109: can we totally get rid of these flags and expose testClusterConfig from within the test cluster object (for tests that need them today) - if you want to allow an override of testClusterConfigFile through flags, that's cool

meiliang86: Update the PR. Now these flags are now purely for override purpose.

Issue #1317, opened 2018-12-06
Automatically forward calls to active cluster


samarabbas:
Currently cadence fails external API calls like SignalWorkflowExecution/StartWorkflowExecution with DomainNotActiveError, when they are called on a domain which is currently passive on the cluster.  
Instead it will be much more user friendly to forward these calls automatically to Active cluster to make sending events seamless for the user.

PR #1626, opened 2019-03-30
Forward start / signal / signal with start API call to active cluster for whitelisted domains.

wxing1292:
Forward start / signal / signal with start API call to active cluster for whitelisted domains.
This will reduce the amount of client / customer logic handling domain not active error for external event (start / signal).

* UT for domain not active forwarding logic
* Add / update necessary mock and fix typo

solve #1317


common/mocks/FrontendClient.go
```diff
@@ -18,613 +18,827 @@
...
+import mock "github.com/stretchr/testify/mock"
+import shared "github.com/uber/cadence/.gen/go/shared"
+import yarpc "go.uber.org/yarpc"
 
 // FrontendClient is an autogenerated mock type for the Client type
 type FrontendClient struct {
 	mock.Mock
 }
 
-var _ workflowserviceclient.Interface = (*FrontendClient)(nil)
```
longquanzheng: it's better to keep this assertion

common/mocks/PublicClient.go
```diff
@@ -0,0 +1,845 @@
...
+import mock "github.com/stretchr/testify/mock"
+
+import shared "go.uber.org/cadence/.gen/go/shared"
+import yarpc "go.uber.org/yarpc"
+
+// PublicClient is an autogenerated mock type for the Client type
+type PublicClient struct {
+	mock.Mock
+}
+
```
longquanzheng: better to add an assertion to help our IDE as we always do

service/frontend/dcRedirectionHandler.go
```diff
@@ -546,16 +587,37 @@ func (handler *DCRedirectionHandlerImpl) SignalWithStartWorkflowExecution(
...
+	targetDC, err := handler.redirectionPolicy.GetTargetDataCenterByName(request.GetDomain())
 	if err != nil {
 		return nil, err
 	}
 
-	if targetDC == handler.currentClusteName {
-		return handler.frontendHandler.SignalWithStartWorkflowExecution(ctx, request)
+	var ok bool
+	var resp *shared.StartWorkflowExecutionResponse
+	attempt := APIAttemptCount
```
longquanzheng: APIAttemptCount and DomainActivenessAttemptCount looks weird to me. Basically you want to use these two "magic" number to help decide which handler API to use.

My personal suggestion is to remove these two number to make code looks cleaner, and another way to organize your code:

```
var handlers []Handler
if targetDC == handler.currentClusterName {
    handlers :=[]Handler{handler.frontendHandler, 
    handler.service.GetClientBean().GetRemoteFrontendClient(targetDC))
else{
    handlers :=[]Handler{handler.service.GetClientBean().GetRemoteFrontendClient(targetDC),
    handler.frontendHandler)
}

resp, err = handler[0].XXXX(ctx, request)
_, ok = handler.isDomainNotActiveError(err) 
if ok{
    resp, err = handler[1].XXXX(ctx, request)
}   
return resp, err
}

File: service/frontend/dcRedirectionHandler.go [hunk 4]
```diff
@@ -30,23 +35,59 @@ import (
...
+		wfHandler,
+	)
+
+	return dcRedirectionHandler
+}
+
 // NewDCRedirectionHandler creates a thrift handler for the cadence service, frontend and admin
-func NewDCRedirectionHandler(currentClusteName string, redirectionPolicy DCRedirectionPolicy,
-	sevice service.Service, frontendHandler *WorkflowHandler) *DCRedirectionHandlerImpl {
+func NewDCRedirectionHandler(currentClusterName string, domainCache cache.DomainCache, config *Config, redirectionPolicy DCRedirectionPolicy,
```
venkat1109: this method also takes a WorkflowHandler as input i.e. looks like most of the other inputs this constructor takes could also be derived from workflowHandler - only difference between this and other constructor i see is - this method takes a policy directly whereas other takes a redirectPolicyConfig - is this correct ? If yes, can you fix this ?

service/frontend/dcRedirectionHandler.go
```diff
@@ -30,23 +35,59 @@ import (
...
+	// 2. SignalWithStartWorkflowExecution
+	// 3. SignalWorkflowExecution
+	DomainActivenessAttemptCount = 2
+
+	// APIAttemptCount is the attempt for normal API calls
+	APIAttemptCount = 1
+)
+
+// NewDCRedirectionHandlerWithWorkflowHandler create a dc redirection handler with existing frontend handler
+func NewDCRedirectionHandlerWithWorkflowHandler(wfHandler *WorkflowHandler, policy config.DCRedirectionPolicy) *DCRedirectionHandlerImpl {
```
venkat1109: naming nit: both this and other method take workflowHandler as input - I don't see why this is called WithWorkflowHandler ? Consider renaming

service/frontend/dcRedirectionHandler.go
```diff
@@ -564,16 +626,36 @@ func (handler *DCRedirectionHandlerImpl) SignalWorkflowExecution(
...
 	if err != nil {
 		return err
 	}
 
-	if targetDC == handler.currentClusteName {
-		return handler.frontendHandler.SignalWorkflowExecution(ctx, request)
+	var ok bool
+	attempt := APIAttemptCount
+
+	enableDomainNotActiveForwarding, activeCluster, err := handler.enableDomainNotActiveAutoForwarding(request.GetDomain())
```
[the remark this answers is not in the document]
wxing1292: in most case, the domain cache will have the correct active cluster, meaning that we will save an API call round trip in most case.

PR #1627, opened 2019-03-30
Remove dual emit support for metrics

shreyassrivatsan:
Remove the code that dual emits legacy and prometheus style metrics. After this change we only have the new prometheus style metrics.


shreyassrivatsan:
Yeah. Will hold off on this till the release is done and all dashboards/alerts are migrated to the new style metrics..

PR #1629, opened 2019-04-01
Fix: set backoff timer for cron childWorkflow when it's started

yycptt:
https://github.com/uber/cadence/pull/1589 fix the bug when starting a cron workflow, the first decision will be scheduled immediately. The same bug also exists for cron childWorkflow. This PR fixes that.


service/history/historyEngine.go
```diff
@@ -550,7 +550,9 @@ func (e *historyEngineImpl) StartWorkflowExecution(ctx context.Context, startReq
 	timerTasks := []persistence.Task{&persistence.WorkflowTimeoutTask{
 		VisibilityTimestamp: e.shard.GetTimeSource().Now().Add(timeoutDuration),
 	}}
-	if cronBackoffSeconds != 0 {
+
+	// Only schedule the backoff timer task if not child WF and there's first decision task backoff
+	if cronBackoffSeconds != 0 && startRequest.ParentExecutionInfo == nil {
```
samarabbas: Can you also check if we need similar logic on SignalWithStart API?

service/history/historyEngine.go
```diff
@@ -2567,15 +2569,26 @@ func (e *historyEngineImpl) ScheduleDecisionTask(ctx context.Context, scheduleRe
...
 				return nil, ErrWorkflowCompleted
 			}
 
-			// Noop
+			postActions := &updateWorkflowAction{
+				createDecision: true,
+			}
+
+			cronBackoffDuration := msBuilder.GetCronBackoffDuration()
+			if msBuilder.GetNextEventID() == int64(2) && cronBackoffDuration != cron.NoBackoff {
```
samarabbas: Having a check based on event id makes me little worried.  Can you instead add another field on request for AddDecisionTask API to indicate if this is the first decision from caller perspective.  I think it is much safer to let caller decide if this is the first decision and should we backoff and create workflow backoff timer.

service/history/historyEngine.go
yycptt: For now, signalWithStart won't check cronSchedule, so no backoff timer task will be scheduled. This is a bug we need to fix. I have created an issue for it. See https://github.com/uber/cadence/issues/1595

service/history/historyEngine.go
yycptt: Makes sense. Will do.

PR #1634, opened 2019-04-01
 Fix missing cleanup in terminate workflow becuase of size limit

vancexu:
https://github.com/uber/cadence/issues/1619

Tested locally, will add integration tests in another PR, because it will requires a refactor on integration tests.


service/history/workflowExecutionContext.go
```diff
@@ -583,6 +583,23 @@ func (c *workflowExecutionContextImpl) update(transferTasks []persistence.Task,
 					return err
 				}
 				executionInfo.SetLastFirstEventID(firstEvent.GetEventId())
+
+				// add clean up tasks
+				transferTasks = append(transferTasks, &persistence.CloseExecutionTask{})
```
samarabbas: I think we already have a helper method in history engine which generates relevant tasks on workflow completion.  I think we should use the same helper method or create one if it does not exist.  These tasks need to be generated everytime workflow execution completes so we need to make sure this logic remains consistent everywhere.

vancexu: agree, addressed.

PR #1637, opened 2019-04-02
Add blobstore scopes

andrewjdawson2016:
I had the caller side metrics for blobstore but not the receiver side. This add the receiver side metrics.

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -59,6 +59,7 @@ const (
 	History
 	Matching
 	Worker
+	Blobstore
 	NumServices
 )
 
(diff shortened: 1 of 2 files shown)


common/metrics/defs.go
```diff
@@ -842,13 +863,13 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
...
 
-		BlobstoreClientUploadScope:         {operation: "Upload", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientDownloadScope:       {operation: "Download", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientGetTagsScope:        {operation: "GetTags", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientExistsScope:         {operation: "Exists", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientDeleteScope:         {operation: "Delete", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientListByPrefixScope:   {operation: "ListByPrefix", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientBucketMetadataScope: {operation: "BucketMetadata", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
+		BlobstoreClientUploadScope:         {operation: "BlobstoreClientUpload", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
```
shreyassrivatsan: nit: probably do not need `Blobstore` in the operation name as that should show up in the service name I presume?

andrewjdawson2016: This follows the pattern we use for other services for example ```MatchingClientPollForDecisionTaskScope```

PR #1640, opened 2019-04-02
Add Archival Integration Test for ContinueAsNew Workflows

yycptt:
https://github.com/uber/cadence/issues/1555


host/archival_test.go
```diff
@@ -79,26 +87,79 @@ func (s *integrationSuite) TestArchival_NotEnabled() {
...
 
-func (s *integrationSuite) TestArchival_Enabled() {
+func (s *integrationSuite) TestArchival_ContinueAsNew() {
 	s.Equal(cluster.ArchivalEnabled, s.ClusterMetadata.ArchivalConfig().GetArchivalStatus())
 
 	domainID := uuid.New()
-	domain := "archival_domain_enabled"
-	archivalBucket := s.bucketName
+	domain := "archival_domain_continueAsNew"
+	s.createArchivalDomain(domain, domainID, s.bucketName, true)
```
yycptt: I don't think we should create a new domain for each test.

host/archival_test.go
```diff
@@ -23,110 +23,111 @@ package host
...
+	for i := 0; i < retryLimit; i++ {
+		getHistoryResp, err = s.engine.GetWorkflowExecutionHistory(createContext(), &workflow.GetWorkflowExecutionHistoryRequest{
+			Domain: common.StringPtr(s.archivalDomainName),
+			Execution: &workflow.WorkflowExecution{
+				WorkflowId: common.StringPtr(workflowID),
+				RunId:      common.StringPtr(runID),
+			},
+		})
+		if getHistoryResp != nil && getHistoryResp.GetArchived() {
+			break
```
andrewjdawson2016: What do you think about changing this method is ```isHistoryArchived``` and have it return a boolean. That way instead of breaking here you can simply return true. Then in the code that calls this you just do ```s.True(isHistoryArchived(..))``` that way we only need one assertion rather than three and reading a return is more clear than reading a break statement.

host/archival_test.go
```diff
@@ -23,110 +23,111 @@ package host
...
+				PageSize:     1,
+			})
+		} else {
+			var resp *persistence.GetHistoryTreeResponse
+			resp, err = s.testCluster.testBase.HistoryV2Mgr.GetHistoryTree(&persistence.GetHistoryTreeRequest{
+				TreeID: runID,
+			})
+			s.Nil(err)
+			if len(resp.Branches) == 0 && len(resp.ForkingInProgressBranches) == 0 {
+				err = &workflow.EntityNotExistsError{}
```
andrewjdawson2016: This feels odd to be converting the error type. Can we also just have this method return a boolean?

andrewjdawson2016: Once you return a boolean you can also move this code out of the else block.

host/integrationbase.go
```diff
@@ -97,6 +98,16 @@ func (s *IntegrationBase) setupSuite(defaultClusterConfigFile string) {
 		Name:                                   &s.foreignDomainName,
 		WorkflowExecutionRetentionPeriodInDays: &retentionDays,
 	}))
+
+	retentionDays = 0
+	archivalStatus := workflow.ArchivalStatusEnabled
+	s.archivalDomainName = s.randomizeStr("integration-archival-enabled-domain")
+	s.Require().NoError(s.testCluster.GetFrontendClient().RegisterDomain(createContext(), &workflow.RegisterDomainRequest{
+		Name:                                   &s.archivalDomainName,
```
andrewjdawson2016: can we write these in line using the utility functions common.StringPtr() and others like this?

host/testcluster.go
```diff
@@ -106,7 +106,7 @@ func NewCluster(options *TestClusterConfig, logger bark.Logger) (*TestCluster, e
 		EnableWorker:                  options.EnableWorker,
 		EnableEventsV2:                options.EnableEventsV2,
 		EnableVisibilityToKafka:       false,
-		EnableReadHistoryFromArchival: true,
+		EnableReadHistoryFromArchival: options.EnableArchival,
```
andrewjdawson2016: Thanks for catching this.

host/integrationbase.go
yycptt: Oh, I thought we do not have one for archivalStatus. Will update.

host/archival_test.go
yycptt: Makes sense. Will update.

PR #1642, opened 2019-04-03
Archival metric fixes

andrewjdawson2016:


service/worker/archiver/client.go
```diff
@@ -92,7 +93,7 @@ func (c *client) Archive(request *ArchiveRequest) error {
 			logging.TagWorkflowExecutionID: exec.ID,
 			logging.TagWorkflowRunID:       exec.RunID,
 		}).Error("failed to SignalWithStartWorkflow to archival system workflow")
-		c.metricsClient.IncCounter(metrics.ArchiverClientScope, metrics.ArchiverClientSendSignalFailureCount)
```
yycptt: It seems like ArchiverClientSendSignalFailureCount is only used here. If it's not used, shall we delete it from metric/def.go?

andrewjdawson2016: nice catch

common/metrics/defs.go
```diff
@@ -1567,6 +1566,7 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		IndexProcessorCorruptedData:                            {metricName: "index_processor_corrupted_data", oldMetricName: "index-processor.corrupted-data"},
 		ArchiverNonRetryableErrorCount:                         {metricName: "archiver_non_retryable_error", oldMetricName: "archiver.non-retryable-error"},
 		ArchiverSkipUploadCount:                                {metricName: "archiver_skip_upload", oldMetricName: "archiver.skip-upload"},
+		ArchiverRunningDeterministicConstructionCheckCount:     {metricName: "archiver_running_deterministic_construction_check", oldMetricName: "archiver.running-deterministic-construction-check"},
```
shreyassrivatsan: for new metrics you shouldnt need oldmetricname..

PR #1648, opened 2019-04-03
Add domain tag on workflow handler metrics

shreyassrivatsan:
- Move to using the metrics.Scope interface
- Introduce domain tags on more front end metrics


common/metrics/scope.go
```diff
@@ -35,6 +35,11 @@ func newMetricsScope(scope tally.Scope, defs map[int]metricDefinition) Scope {
 	return &metricsScope{scope, defs}
 }
 
+// NoopScope returns a noop scope of metrics
+func NoopScope(serviceIdx ServiceIdx) Scope {
```
venkat1109: thanks for adding this !

PR #1649, opened 2019-04-03
Add Integration test for history size limit

vancexu:
This PR includes an integration test for history size limit case. 
Because we need setup cluster with smaller limit config, this also include a new parameter `HistoryConfig` in Integration Test config. 

(We could do more refactor to move some config nobs to HistoryConfig to make code looks cleaner, if needed)

### DIFF
--- a/host/testcluster.go
+++ b/host/testcluster.go
@@ -54,8 +54,6 @@ type (
 		EnableArchival        bool
 		IsMasterCluster       bool
 		ClusterNo             int
-		NumHistoryShards      int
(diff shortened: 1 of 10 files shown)


host/testcluster.go
```diff
@@ -56,6 +56,7 @@ type (
 		NumHistoryShards      int
```
meiliang86: Can you check if you can move these two into the history config?

vancexu: Ok, updated

PR #1654, opened 2019-04-04
Cancel timer should clear any TimerFired events in buffered events

### LINKED ISSUE #1610: CancelTimer fails if Timer fire gets buffered resulting in non-deterministic error on client
We update the mutableState immediately when a timer fires irrespective of TimerFired event is written to history or is buffered due to outstanding decision.  This results in CancelTimer decision being rejected with CancelTimerFailed event, as during processing of decision it will not be able to find the timer id for the cancelled timer as we already removed it during processing of TimerFired event.  This results in CancelTimerFailed event followed by TimerFired event written to history after we flush the buffer.  Failing of TimerCancel decision is something not expected by clients and it results in non-deterministic error by cadence client.

shreyassrivatsan:
While cancelling timer, if the timer does not exist go over buffered events to check if the timer was already fired and clear it, so that the cancel can succeed.

Fixes #1610

### DIFF
--- a/service/history/historyEngine_test.go
+++ b/service/history/historyEngine_test.go
@@ -4432,7 +4432,7 @@ func (s *engineSuite) TestCancelTimer_RespondDecisionTaskCompleted_TimerFired() 
 	taskToken, _ := json.Marshal(&common.TaskToken{
 		WorkflowID: *we.WorkflowId,
 		RunID:      *we.RunId,
-		ScheduleID: 2,
+		ScheduleID: 6,
 	})
 	identity := "testIdentity"
 	timerID := "t1"
@@ -4442,7 +4442,13 @@ func (s *engineSuite) TestCancelTimer_RespondDecisionTaskCompleted_TimerFired() 
 	// Verify cancel timer with a start event.
 	addWorkflowExecutionStartedEvent(msBuilder, we, "wType", tl, []byte("input"), 100, 100, identity)
 	di := addDecisionTaskScheduledEvent(msBuilder)
-	addDecisionTaskStartedEvent(msBuilder, di.ScheduleID, tl, identity)
+	decisionStartedEvent := addDecisionTaskStartedEvent(msBuilder, di.ScheduleID, tl, identity)
+	decisionCompletedEvent := addDecisionTaskCompletedEvent(msBuilder, di.ScheduleID,
+		*decisionStartedEvent.EventId, nil, identity)
+	addTimerStartedEvent(msBuilder, *decisionCompletedEvent.EventId, timerID, 10)
+	di2 := addDecisionTaskScheduledEvent(msBuilder)
+	addDecisionTaskStartedEvent(msBuilder, di2.ScheduleID, tl, identity)
+	addTimerFiredEvent(msBuilder, di2.ScheduleID, timerID)
 
 	ms := createMutableState(msBuilder)
 	gwmsResponse := &persistence.GetWorkflowExecutionResponse{State: ms}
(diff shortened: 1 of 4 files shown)


service/history/mutableStateBuilder.go
```diff
@@ -2257,18 +2283,31 @@ func (e *mutableStateBuilder) ReplicateTimerFiredEvent(event *workflow.HistoryEv
...
 
-func (e *mutableStateBuilder) AddTimerCanceledEvent(decisionCompletedEventID int64,
-	attributes *workflow.CancelTimerDecisionAttributes, identity string) *workflow.HistoryEvent {
+func (e *mutableStateBuilder) AddTimerCanceledEvent(
+	decisionCompletedEventID int64,
+	attributes *workflow.CancelTimerDecisionAttributes,
+	identity string,
+) *workflow.HistoryEvent {
+	var timerStartedID int64
 	timerID := *attributes.TimerId
```
wxing1292: attributes.GetTimerId() plz

service/history/mutableStateBuilder.go
```diff
@@ -2257,18 +2283,31 @@ func (e *mutableStateBuilder) ReplicateTimerFiredEvent(event *workflow.HistoryEv
...
+		// if timer is not running then check if it has fired in the mutable state.
+		// If so clear the timer from the mutable state. We need to check both the
+		// bufferedEvents and the history builder
+		timerFiredEvent := e.checkAndClearTimerFiredEvent(timerID)
+		if timerFiredEvent == nil {
+			logging.LogInvalidHistoryActionEvent(e.logger, logging.TagValueActionTimerCanceled, e.GetNextEventID(), fmt.Sprintf(
+				"{IsTimerRunning: %v, timerID: %v}", isTimerRunning, timerID))
+			return nil
+		}
+		timerStartedID = *timerFiredEvent.TimerFiredEventAttributes.StartedEventId
```
wxing1292: GetStartedEventId() plz

service/history/mutableStateBuilder.go
```diff
@@ -513,6 +512,33 @@ func (e *mutableStateBuilder) CreateReplicationTask(newRunEventStoreVersion int3
...
+	e.hBuilder.history, timerEvent = checkAndClearTimerFiredEvent(e.hBuilder.history, timerID)
+	return timerEvent
+}
+
+func checkAndClearTimerFiredEvent(events []*workflow.HistoryEvent, timerID string) ([]*workflow.HistoryEvent, *workflow.HistoryEvent) {
+	// go over all history events. if we find a timer fired event for the given
+	// timerID, clear it
+	timerFiredIdx := -1
+	for idx, event := range events {
+		if *event.EventType == workflow.EventTypeTimerFired &&
```
wxing1292: GetEventType() plz

service/history/mutableStateBuilder.go
```diff
@@ -513,6 +512,33 @@ func (e *mutableStateBuilder) CreateReplicationTask(newRunEventStoreVersion int3
 	return t
 }
 
+func (e *mutableStateBuilder) checkAndClearTimerFiredEvent(timerID string) *workflow.HistoryEvent {
+	var timerEvent *workflow.HistoryEvent
+	e.bufferedEvents, timerEvent = checkAndClearTimerFiredEvent(e.bufferedEvents, timerID)
+	if timerEvent != nil {
+		return timerEvent
+	}
```
wxing1292: not sure whether also need to check updateBufferedEvents

service/history/historyEngine_test.go
```diff
@@ -4422,6 +4422,80 @@ func (s *engineSuite) TestCancelTimer_RespondDecisionTaskCompleted_NoStartTimer(
...
+			Identity:         &identity,
+		},
+	})
+	s.Nil(err)
+
+	executionBuilder := s.getBuilder(domainID, we)
+	s.Equal(int64(11), executionBuilder.GetExecutionInfo().NextEventID)
+	s.Equal(int64(7), executionBuilder.GetExecutionInfo().LastProcessedEvent)
+	s.Equal(persistence.WorkflowStateRunning, executionBuilder.GetExecutionInfo().State)
+	s.True(executionBuilder.HasPendingDecisionTask())
```
wxing1292: this probably should not has pending task, since RespondDecisionTaskCompleted is already been called.

wxing1292: one more thing, since the buffered events are potentially touched, could you also verify that buffered events is empty after RespondDecisionTaskCompleted?

shreyassrivatsan: > this probably should not has pending task, since RespondDecisionTaskCompleted is already been called.
this is probably to do with some test setup. let me take a look.. I agree it should not have a pending task.

service/history/mutableStateBuilder.go
shreyassrivatsan: I was not sure about this. I think it doesnt harm to also check updateBufferedEvents here. I can add a check for that too..

service/history/mutableStateBuilder.go
shreyassrivatsan: i'll make these changes.

service/history/mutableStateBuilder.go
shreyassrivatsan: done

PR #1673, opened 2019-04-08
Add archival integration test for multi-blob history workflow

yycptt:
Task: https://github.com/uber/cadence/issues/1652


host/archival_test.go
```diff
@@ -153,7 +174,63 @@ func (s *integrationSuite) isMutableStateDeleted(domainID string, execution *wor
...
+
+	if !s.testClusterConfig.EnableEventsV2 {
+		req := &persistence.GetWorkflowExecutionHistoryRequest{
+			DomainID:     domainID,
+			Execution:    *execution,
+			FirstEventID: common.FirstEventID,
+			NextEventID:  common.EndEventID,
+			PageSize:     pageSize,
+		}
+		resp := &persistence.GetWorkflowExecutionHistoryResponse{
```
andrewjdawson2016: It looks kind of weird to create a fake resp just to make sure you enter the loop the first time. How about instead we move the first GetWorkflowExecutionHistory call outside of the loop. That way the loop condition always operates on real responses?

yycptt: I agree it's kind of weird. But otherwise, there will be 4 lines of duplicated code. How about I also use historySize == 0 as part of the condition to enter the loop the first time?

PR #1681, opened 2019-04-09
sql: support deleting visibility records

### LINKED ISSUE #1666: sql: support deletion of visibility records
Currently, visibility records are never deleted in SQL world. A possible solution here is to delete the visibility record as part of the workflow retention timeout processing and/or archival cleanup processing

venkat1109:
Currently, SQL backed visibility implementation doesn't support deletion of records. This patch does the following:
- Adds a visibility persistence API to delete a record 
- Adds the necessary plumbing via timer_task_processor to call this API

Fixes #1666


common/persistence/sql/storage/mysql/visibility.go
```diff
@@ -106,6 +109,11 @@ func (mdb *DB) ReplaceIntoVisibility(row *sqldb.VisibilityRow) (sql.Result, erro
 	}
 }
 
+// DeleteFromVisibility deletes a row from visibility table if it exist
+func (mdb *DB) DeleteFromVisibility(filter *sqldb.VisibilityFilter) (sql.Result, error) {
+	return mdb.conn.Exec(templateDeleteWorkflowExecution, filter.DomainID, filter.RunID)
```
vancexu: Will this return error if target row is not found?

venkat1109: it won't. Just the RowsAffected() will return 0

PR #1689, opened 2019-04-10
Support shard in event v2 sql store

yux0:
Create a draft PR for supporting shard in event v2 on SQL store


common/persistence/historyV2Store.go
```diff
@@ -79,6 +79,7 @@ func (m *historyV2ManagerImpl) ForkHistoryBranch(request *ForkHistoryBranchReque
 		ForkNodeID:     request.ForkNodeID,
 		NewBranchID:    uuid.New(),
 		Info:           request.Info,
+		ShardID:        request.ShardID,
```
longquanzheng: I am a bit worried about missing some places in application layer. Can you add an assertion here to require shardID? return error if that is empty.

common/persistence/dataInterfaces.go
```diff
@@ -1300,18 +1306,24 @@ type (
 		BranchToken []byte
 		// true means the fork is success, will update the flag, otherwise will delete the new branch
 		Success bool
+		// The shard to update history branch data
+		ShardID int
```
longquanzheng: ShardID should be pointer to make sure it is not a default value. Since zero is also a valid shardID for us.

common/persistence/sql/storage/mysql/eventsV2.go
```diff
@@ -29,24 +29,24 @@ import (
...
 	// below are templates for history_node table
 	addHistoryNodesQry = `INSERT INTO history_node (` +
-		`tree_id, branch_id, node_id, txn_id, data, data_encoding) ` +
-		`VALUES (:tree_id, :branch_id, :node_id, :txn_id, :data, :data_encoding) `
+		`shard_id, tree_id, branch_id, node_id, txn_id, data, data_encoding) ` +
+		`VALUES (:shard_id, :tree_id, :branch_id, :node_id, :txn_id, :data, :data_encoding) `
 
 	getHistoryNodesQry = `SELECT node_id, txn_id, data, data_encoding FROM history_node ` +
-		`WHERE tree_id = ? AND branch_id = ? AND node_id >= ? and node_id < ? ORDER BY tree_id, branch_id, node_id, txn_id LIMIT ? `
+		`WHERE shard_id = ? AND tree_id = ? AND branch_id = ? AND node_id >= ? and node_id < ? ORDER BY tree_id, branch_id, node_id, txn_id LIMIT ? `
```
longquanzheng: ORDER BY should also include shard_id and must be in the same order as primary key, so that we can get the benefit of primary key ordering.

service/frontend/dcRedirectionHandler_test.go
```diff
@@ -89,7 +89,7 @@ func (s *dcRedirectionHandlerSuite) SetupTest() {
 	s.logger = bark.NewLoggerFromLogrus(log2)
 	s.currentClusterName = cluster.TestCurrentClusterName
 	s.alternativeClusterName = cluster.TestAlternativeClusterName
-	s.config = NewConfig(dynamicconfig.NewCollection(dynamicconfig.NewNopClient(), s.logger), false)
+	s.config = NewConfig(dynamicconfig.NewCollection(dynamicconfig.NewNopClient(), s.logger), 0, false)
```
longquanzheng: that is changed here?

common/persistence/historyV2Store.go
yux0: This store is used by Cassandra and Mysql. How about adding the assertion in sqlHistoryV2Manager.go?

common/persistence/dataInterfaces.go
yux0: Will update it

common/persistence/sql/storage/mysql/eventsV2.go
yux0: updated

common/persistence/historyV2Store.go
longquanzheng: I think it should be in the Store, even though Cassandra doesn't need it, it is still required as part of the contract. What we want is to keep our persistence interface as simple as possible, and centralize any logic that can be shared by all different DBs.

yux0: Make sense. Added the validation logic.

common/persistence/dataInterfaces.go
yux0: This change involves many changes in different place. Could you take another look?

common/persistence/historyV2Store.go
```diff
@@ -73,12 +73,19 @@ func (m *historyV2ManagerImpl) ForkHistoryBranch(request *ForkHistoryBranchReque
 	if err != nil {
 		return nil, err
 	}
-
+	shardID, err := getShardID(request.ShardID)
+	if err != nil {
+		m.logger.WithError(err).Error("shardID is not set in fork history branch operation")
```
longquanzheng: We don't need to log this error. The application layer should already have log. We only log something in persistence layer when we need to log some additional information for debugging.

common/persistence/persistence-tests/historyV2PersistenceTest.go
```diff
@@ -711,6 +716,10 @@ func (s *HistoryV2PersistenceSuite) append(branch []byte, events []*workflow.His
 	var resp *p.AppendHistoryNodesResponse
 
 	op := func() error {
+		var shardID *int
+		if s.ShardInfo != nil {
```
longquanzheng: If you have this code, you probably want to have a unit test to assert on the "shardID not set" error. Otherwise I don't think it is useful.

schema/mysql/v57/cadence/schema.sql
```diff
@@ -357,24 +357,26 @@ CREATE TABLE signals_requested_sets (
 
 -- history eventsV2: history_node stores history event data
 CREATE TABLE history_node (
+	shard_id INT NOT NULL,
```
longquanzheng: fix format

service/history/timerQueueActiveProcessor.go
```diff
@@ -747,8 +747,12 @@ Update_History_Loop:
 			return err
 		}
 
+		var shardID *int
+		if t.shard != nil {
+			shardID = common.IntPtr(t.shard.GetShardID())
```
longquanzheng: This check is not correct/needed. If t.shard is null then you should return error. This check here assume that shardID can be null which will get an error from persistence API.

service/history/workflowResetor.go
```diff
@@ -501,13 +511,18 @@ func (w *workflowResetorImpl) replayReceivedSignals(ctx context.Context, receive
 		continueRunID = ""
 
 		var nextPageToken []byte
+		var shardID *int
+		if w.eng != nil && w.eng.shard != nil {
```
longquanzheng: same as above. Our initialization has made sure it won't happen. Otherwise it should just panic without making any process.

schema/mysql/v57/cadence/schema.sql
yux0: updatedd

common/persistence/persistence-tests/historyV2PersistenceTest.go
yux0: removed

service/history/workflowResetor.go
yux0: removed

PR #1692, opened 2019-04-11
Returning empty response for last empty history events read request

yycptt:


common/persistence/cassandra/cassandraHistoryV2Persistence.go
```diff
@@ -193,6 +196,10 @@ func (h *cassandraHistoryV2Persistence) ReadHistoryBranch(request *p.InternalRea
 		}
 	}
 
+	if !found && len(request.NextPageToken) == 0 {
```
longquanzheng: It's better to do this check in HistoryV2Store instead of Cassandra/SQL implementation. 
In some cases fork succeeds but haven't append events yet, we probably don't want to return error here.

yycptt: I agree with you that certralizing logic is better. But in this case, we have two components (Cassandra/SQL) implementing the same interface. I think it's kind of weird if the two implementations have different behaviors and the difference is handled by a higher layer.

yycptt: For the fork case you mentioned, I think the request.NextPageToken will not be empty? So no error will be returned?

longquanzheng: Doing it in HistoryV2Store will let both Cassandra/SQL have the same behavior without writing code in two places. That's why we have this higher layer.
I am not sure why you think difference if we write this code in the higher layer.

The only difference I know is historyV1 have it in lower layer, but essentially we can/should also do it in the higher layer.

For that case, request.NextPageToken is nil, check this line: https://github.com/uber/cadence/blob/master/common/persistence/historyV2Store.go#L372

yycptt: Oh, sorry, I was actually think about V1 where the behavior is different for two implementations. And yes, for V2, right now both SQL and Cassandra will return empty response if there's no row and we can do the check in higher layer to see if an error is needed. I agree and will update.

common/persistence/historyV2Store.go
```diff
@@ -291,6 +291,9 @@ func (m *historyV2ManagerImpl) readHistoryBranch(byBatch bool, request *ReadHist
 	}
 
 	resp, err := m.persistence.ReadHistoryBranch(req)
+	if len(resp.History) == 0 && len(request.NextPageToken) == 0 {
```
andrewjdawson2016: If you get an error will resp be nil? i.e. can you get a nil pointer here?

common/persistence/historyV2Store.go
```diff
@@ -291,6 +291,9 @@ func (m *historyV2ManagerImpl) readHistoryBranch(byBatch bool, request *ReadHist
...
+		err = &workflow.EntityNotExistsError{Message: "Workflow execution history not found."}
```
andrewjdawson2016: I prefer to always return as soon as you can that way the reader knows they are done reading. But its a matter of preference.

common/persistence/sql/sqlHistoryManager.go
```diff
@@ -104,6 +104,9 @@ func (m *sqlHistoryManager) GetWorkflowExecutionHistory(request *p.InternalGetWo
 
 	// TODO: Ensure that no last empty page is requested
 	if err == sql.ErrNoRows || (err == nil && len(rows) == 0) {
+		if len(request.NextPageToken) != 0 {
```
andrewjdawson2016: Are you able to just always return an empty response regardless of NextPageToken, this will get handled correctly at upper layer?

yycptt: I'd like to match the existing behavior in Cassandra V1 here.

common/persistence/historyV2Store.go
yycptt: Updated.

common/persistence/historyV2Store.go
yycptt: Updated.

common/persistence/persistence-tests/historyPersistenceTest.go
```diff
@@ -102,11 +102,27 @@ func (s *HistoryPersistenceSuite) TestGetHistoryEvents() {
...
+	// events should be returned.
+	history, token, err1 = s.GetWorkflowExecutionHistory(domainID, workflowExecution, 1, 4, 1, nil)
+	s.Nil(err1)
+	s.Equal(2, len(history.Events))
+	if len(token) != 0 {
+		history, token, err1 = s.GetWorkflowExecutionHistory(domainID, workflowExecution, 1, 4, 1, token)
+		s.Nil(err1)
+		s.Equal(0, len(token))
+		s.Equal(0, len(history.Events))
+	}
```
longquanzheng: Can you add a test to assert on EntityNotExistsError?

common/persistence/persistence-tests/historyV2PersistenceTest.go
```diff
@@ -260,11 +260,21 @@ func (s *HistoryV2PersistenceSuite) TestReadBranchByPagination() {
 	req.NextPageToken = resp.NextPageToken
```
longquanzheng: same here, we need to add a test to assert on EntityNotExistsError so that we can make sure this behavior won't be broken

common/persistence/sql/sqlHistoryManager.go
longquanzheng: If there is no big change, I think it's better to let both V1/V2 return this error in upper layer instead of lower level. As we always try to keep our lower level code simple so that adding new implementation (like Spanner/DynamoDB/etc) would be much easier.

common/persistence/persistence-tests/historyPersistenceTest.go
yycptt: Updated.

common/persistence/persistence-tests/historyV2PersistenceTest.go
yycptt: Done.

PR #1693, opened 2019-04-11
Remove buffered replication task functionality in favor of resend

wxing1292:


wxing1292:
> Make sure you think through the story of existing buffered tasks after removing the code, if we haven't drained all of them yet.

we have been using re-send for a long time, plus, the existence of re-send will make buffer replication task useless.

PR #1696, opened 2019-04-11
matching: fix bug in ratelimiter, increase visibility

### LINKED ISSUE #1691: Add task list throttling info into DescribeTaskListResponse

venkat1109:
This patch contains the following changes:

Fixes bug in task list rate limiter code path
When history attempts to add a task to matching engine, the sync match code path gets executed. As part of this, the taskListManager attempts to make a reservation in the token bucket rate limiter. If the reservation is too far out in the future, the code decides to forego sync match and instead write the task to persistence. However, the reserved token is not returned to the token bucket here. This patch adds a fix to return this token back.

Increases task list manager visibility
- Added domain scoped metrics for task list manager
- Modified DescribeTaskList API to return the current ratePerSecond for the task list

Fixes #1691


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1251,6 +1251,7 @@ struct TaskListStatus {
   10: optional i64 (js.type = "Long") backlogCountHint
   20: optional i64 (js.type = "Long") readLevel
   30: optional i64 (js.type = "Long") ackLevel
+  35: optional double ratePerSecond
```
mfateev: I would also add it to the PollerInfo. It might help troubleshooting situation when different pollers ask for a different rate.

service/matching/taskListManager.go
```diff
@@ -257,6 +259,7 @@ func newTaskListManagerWithRateLimiter(
 			logging.TagTaskListName: taskList.taskListName,
 		}),
 		metricsClient:       e.metricsClient,
+		scope:               domainTaggedMetricScope(e.domainCache, taskList.domainID, e.metricsClient, metrics.MatchingTaskListMgrScope),
```
shreyassrivatsan: name this `domainScope`?

service/matching/taskListManager.go
```diff
@@ -461,16 +464,21 @@ func (c *taskListManagerImpl) getTask(ctx context.Context) (*getTaskResult, erro
 	case result := <-tasksForPoll:
 		if result.syncMatch {
 			c.metricsClient.IncCounter(scope, metrics.PollSuccessWithSyncCounter)
+			c.scope.IncCounter(metrics.PollSuccessWithSyncCounter)
```
shreyassrivatsan: There is going to be an issue with metrics of this form...

`fetch name:poll_success` will return the total values and the domain specific values too. Its not possible to separate out the aggregate and domain specific series - with the current m3ql language as you cannot say give me series without domain tag. 

I'd suggest removing the legacy aggregate metric and only keep the domain tagged. It would mean changing queries wherever they are used with a `| sum`, but that shouldnt be an issue.

venkat1109: yeah. I realized this after I pushed this for code review. I am planning to use DomainAll() for places where I need both domain specific and aggregate

shreyassrivatsan: Do we really need `domain:all` and dual emission - with the relatively low cardinality we have on domains. 

Counters are easy to aggregate with a sum function.. I'd prefer to leave the use of the all tag just for timers which cannot really be aggregated. Ideally we should use histograms and then we would not need to use the `domain:all` tag anywhere..

service/matching/ratelimiter.go
```diff
@@ -26,6 +26,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"math"
```
shreyassrivatsan: nit: import formatting

service/matching/taskListManager.go
```diff
@@ -725,3 +739,13 @@ func createServiceBusyError(msg string) *s.ServiceBusyError {
 func (c *taskListManagerImpl) isTaskAddedRecently(lastAddTime time.Time) bool {
 	return time.Now().Sub(lastAddTime) <= c.config.MaxTasklistIdleTime()
 }
+
+func domainTaggedMetricScope(cache cache.DomainCache, domainID string, client metrics.Client, scope int) metrics.Scope {
+	entry, err := cache.GetDomainByID(domainID)
+	switch {
```
shreyassrivatsan: interesting use of switch. any reason to do this against a simple `if err != nil`

service/matching/taskListManager.go
venkat1109: Good question - I guess, we can always add the domain:all in future if we ever hit the cardinality limit - then I am as well inclined to omit domain:all for now

service/matching/taskListManager.go
venkat1109: probably not. I was thinking I will have more cases when I initially started writing this, then forgot to cleanup. I will change this to simple err != nil check

idl/github.com/uber/cadence/shared.thrift
venkat1109: done

service/matching/taskListManager.go
venkat1109: done

PR #1698, opened 2019-04-12
Logging refactor #2: rename old logger and Initialize new logger, and apply new logger in frontend

longquanzheng:


common/service/config/log.go
```diff
@@ -57,6 +61,44 @@ func (cfg *Logger) NewBarkLogger() bark.Logger {
 	return bark.NewLoggerFromLogrus(logger)
 }
 
+// NewZapLogger builds and returns a new zap
+// logger for this logging configuration
+func (cfg *Logger) NewZapLogger() *zap.Logger {
+	encodeConfig := zapcore.EncoderConfig{
+		TimeKey:        "ts",
```
wxing1292: plz make this a const

PR #1700, opened 2019-04-12
Enable the functionality to re-apply signals with lower version

wxing1292:
* Allows signal re-application between different run ID
* Modify conflict resolution logic to re-apply signals with lower version

### DIFF
--- a/service/history/historyReplicator.go
+++ b/service/history/historyReplicator.go
@@ -45,6 +45,8 @@ var (
 
 	workflowTerminationReason   = "Terminate Workflow Due To Version Conflict."
 	workflowTerminationIdentity = "worker-service"
+
+	workflowResetReason = "Reset Workflow Due To Events Re-application."
 )
 
 type (
@@ -54,6 +56,7 @@ type (
 
 	historyReplicator struct {
 		shard             ShardContext
+		timeSource        clock.TimeSource
 		historyEngine     *historyEngineImpl
 		historyCache      *historyCache
 		domainCache       cache.DomainCache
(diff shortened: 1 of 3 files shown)


service/history/workflowResetor.go
```diff
@@ -141,19 +157,51 @@ func (w *workflowResetorImpl) ResetWorkflowExecution(ctx context.Context, resetR
 	}
 
 	// before changing mutable state
-	prevRunVersion := currMutableState.GetLastWriteVersion()
+	currPrevRunVersion := currMutableState.GetLastWriteVersion()
 	// terminate the current run if it is running
-	terminateCurr, closeTask, cleanupTask, retError := w.terminateIfCurrIsRunning(currMutableState, request.GetReason(), currExecution)
+	currTerminated, currCloseTask, currCleanupTask, retError := w.terminateIfCurrIsRunning(currMutableState, request.GetReason(), currExecution)
```
longquanzheng: Adding a new flag to reset request(terminateCurrent ) looks more natural to me. As it is also a natural behavior other API( like CLI) may need. Changing the interface is too heavy here.

wxing1292: then should we expose this 'terminate' flag?

wxing1292: one more thing, i cannot directly use the existing exposed API for workflow reset, since cross dc path already has the context & mutable state, and cannot simply just release the lock.

longquanzheng: I see your point. But having two methods with same name in this interface still looks bad to me(it is hard to understand when I comeback later ). Another option is to move the code that locking mutable states into historyEngine and let workflowResetor only does thing about reset. 

The new param in the request probably should named "terminateRunID".

longquanzheng:
A lot of changes are outdated. Can you rebase to the latest master?

service/history/historyReplicator.go
```diff
@@ -103,12 +109,28 @@ var (
 	ErrEmptyHistoryRawEventBatch = &shared.BadRequestError{Message: "encounter empty history batch"}
 	// ErrUnknownEncodingType indicate that the encoding type is unknown
 	ErrUnknownEncodingType = &shared.BadRequestError{Message: "unknown encoding type"}
+	// ErrUnreappliableEvent indicate that the event is not reappliable
+	ErrUnreappliableEvent = &shared.BadRequestError{Message: "event is not "}
```
longquanzheng: Do you miss something here?

PR #1702, opened 2019-04-12
Add public client to bootstrap params

andrewjdawson2016:


common/service/service.go
```diff
@@ -73,6 +72,7 @@ type (
 		DispatcherProvider  client.DispatcherProvider
 		BlobstoreClient     blobstore.Client
 		DCRedirectionPolicy config.DCRedirectionPolicy
+		PublicClient        workflowserviceclient.Interface
```
wxing1292: do we use a different client in the internal repo?

andrewjdawson2016: There will be a follow up diff in the internal repo. This diff needs more context. I can explain in person on monday.

PR #1703, opened 2019-04-12
ES integration test

yycptt:


host/onebox.go
```diff
@@ -143,31 +151,33 @@ type (
...
+		esConfig:                   params.ESConfig,
+		esClient:                   params.ESClient,
+		enableReadVisibilityFromES: params.EnableReadVisibilityFromES,
+		blobstoreClient:            params.Blobstore,
+		historyConfig:              params.HistoryConfig,
+		workerConfig:               params.WorkerConfig,
 	}
 }
 
 func (c *cadenceImpl) enableWorker() bool {
```
vancexu: Is this still needed?

host/onebox.go
```diff
@@ -338,10 +350,23 @@ func (c *cadenceImpl) startFrontend(hosts map[string][]string, startWG *sync.Wai
 	c.frontEndService = service.New(params)
 	c.adminHandler = frontend.NewAdminHandler(
 		c.frontEndService, c.historyConfig.NumHistoryShards, c.metadataMgr, c.historyMgr, c.historyV2Mgr)
-	frontendConfig := frontend.NewConfig(dynamicconfig.NewCollection(params.DynamicConfig, c.logger), false)
+	dc := dynamicconfig.NewCollection(params.DynamicConfig, c.logger)
+	frontendConfig := frontend.NewConfig(dc, c.esConfig.Enable)
+	visibilityMgr := c.visibilityMgr
+	if c.esConfig.Enable && c.enableReadVisibilityFromES {
```
vancexu: I think we should remove `enableReadVisibilityFromES` config from test and only use esEnable as single switch for both write and read with ES. Because we probably always want both rw

docker/buildkite/docker-compose-local.yml
```diff
@@ -48,6 +48,17 @@ services:
       KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
 
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.7.1
```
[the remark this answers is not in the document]
yycptt: updated

docker/buildkite/docker-compose.yml
```diff
@@ -48,6 +48,17 @@ services:
       KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
 
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:6.7.1
```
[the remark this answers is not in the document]
yycptt: updated.

host/onebox.go
yycptt: It's still used several places in the code.

host/onebox.go
yycptt: Removed.

PR #1710, opened 2019-04-16
Handling replication task size limit issue by sending out history rep…

wxing1292:
…lication metadata task, so receiver will fetch history events directly from source cluster, i.e. by passing Kafka.


idl/github.com/uber/cadence/replicator.thrift
```diff
@@ -61,6 +62,15 @@ struct HistoryTaskAttributes {
   120: optional bool resetWorkflow
 }
 
+struct HistoryMetadataTaskAttributes {
+  05: optional list<string> targetClusters
+  10: optional string domainId
+  20: optional string workflowId
+  30: optional string runId
+  40: optional i64 (js.type = "Long") firstEventId
+  50: optional i64 (js.type = "Long") nextEventId
```
samarabbas: Do you think adding version and replicationInfo also makes sense here? It may be useful for things like dedup'ing messages without causing a remote call.

wxing1292: existing history re-replicator, when getting history, will handle continue as new & replication info
version is omitted here since history re-replicator does not require it.

wxing1292: btw, i need to do a double check on the check of kafka error

PR #1711, opened 2019-04-16
Logging: replace with new logger in matching/worker(scanner/indexer/archiver)

longquanzheng:


common/log/tag/tags.go
```diff
@@ -267,11 +267,16 @@ func MetricScope(metricScope int64) Tag {
...
+	return newPredefinedStringTag("store-type", storeType)
+}
+
 // history engine shard
 
 // ShardID returns tag for ShardID
-func ShardID(shardID int64) Tag {
-	return newIntegerTag("shard-id", shardID)
+func ShardID(shardID int) Tag {
+	return newIntegerTag("shard-id", int64(shardID))
```
longquanzheng: I will create new helpers: newInt32 newInt64 to replace newIntegerTag so that we don't have to convert it.

common/log/logger.go
```diff
@@ -45,6 +45,8 @@ import (
 //  Note: msg should be static, it is not recommended to use fmt.Sprintf() for msg.
 //        Anything dynamic should be tagged.
 type Logger interface {
+	internalLogger
```
mfateev: This design choice is weird. Could we have a separate interface that loggerImpl implements?

longquanzheng: Addressed

common/log/replay_logger.go
```diff
@@ -38,11 +37,17 @@ const skipForReplayLogger = skipForDefaultLogger + 1
...
-			zapLogger: logger.toZap(),
+	lg, ok := logger.(*loggerImpl)
+	if ok {
+		logger = &loggerImpl{
+			zapLogger: lg.zapLogger,
 			skip:      skipForReplayLogger,
-		},
+		}
+	} else {
+		logger.Warn("ReplayLogger may not emit callat tag correctly")
```
mfateev: Add explanation why

longquanzheng: OK

PR #1715, opened 2019-04-17
Fix cron schedule and retry policy not replicated to standby bug

yycptt:


yycptt:
Fix https://github.com/uber/cadence/issues/1419

PR #1719, opened 2019-04-18
Verify requestid in StartWorkflowExecutionRequest

lihannan99:


service/history/historyEngine.go
```diff
@@ -3273,6 +3273,9 @@ func validateStartChildExecutionAttributes(parentInfo *persistence.WorkflowExecu
 }
 
 func validateStartWorkflowExecutionRequest(request *workflow.StartWorkflowExecutionRequest, maxIDLengthLimit int) error {
+	if request.RequestId == nil || len(common.StringDefault(request.RequestId)) == 0 {
```
longquanzheng: I think checking len(request.requestID) is sufficient.

lihannan99: updated. Thanks!

lihannan99:
This is fix to https://github.com/uber/cadence/issues/1487

service/history/historyEngine.go
```diff
@@ -3274,6 +3274,9 @@ func validateStartChildExecutionAttributes(parentInfo *persistence.WorkflowExecu
 }
 
 func validateStartWorkflowExecutionRequest(request *workflow.StartWorkflowExecutionRequest, maxIDLengthLimit int) error {
+	if len(common.StringDefault(request.RequestId)) == 0 {
```
yycptt: Can we use len(request.GetRequestId()) == 0 here?

PR #1721, opened 2019-04-19
Logging: replace with new logger in history/common and remove old loggers

longquanzheng:


common/log/loggerimpl/logger.go
```diff
@@ -18,50 +18,48 @@
...
 
-package log
+package loggerimpl
 
 import (
 	"fmt"
-	"path/filepath"
-	"runtime"
-
+	"github.com/stretchr/testify/suite"
```
mfateev: import order

common/persistence/cassandra/cassandraMetadataPersistenceProxy.go
```diff
@@ -112,23 +113,23 @@ func (m *metadataManagerProxy) UpdateDomain(request *p.UpdateDomainRequest) erro
 func (m *metadataManagerProxy) DeleteDomain(request *p.DeleteDomainRequest) error {
 	err := m.metadataMgr.DeleteDomain(request)
 	if err != nil {
-		m.logger.Warnf("Error deleting domain from V1 table: %v", err)
+		m.logger.Warn("Error deleting domain from V1 table", tag.Error(err))
```
mfateev: This is such a common occurrence. May be introduce something like WarnErr(string, error, fields...) to wrap this?

common/persistence/elasticsearch/elasticsearchVisibility.go
```diff
@@ -77,11 +77,11 @@ var (
 )
 
 // NewElasticSearchVisibilityManager create a visibility manager connecting to ElasticSearch
-func NewElasticSearchVisibilityManager(esClient es.Client, index string, config *config.VisibilityConfig, logger bark.Logger) p.VisibilityManager {
+func NewElasticSearchVisibilityManager(esClient es.Client, index string, config *config.VisibilityConfig, logger log.Logger) p.VisibilityManager {
 	return &esVisibilityManager{
 		esClient: esClient,
 		index:    index,
-		logger:   logger.WithField(logging.TagWorkflowComponent, logging.TagValueESVisibilityManager),
+		logger:   logger.WithTags(tag.ComponentESVisibilityManager),
```
mfateev: No tag.Component?

common/service/config/log.go
```diff
@@ -21,8 +21,6 @@
 package config
```
mfateev: May be rename this file?

common/service/dynamicconfig/config.go
```diff
@@ -39,26 +39,23 @@ func NewCollection(client Client, logger bark.Logger) *Collection {
...
-	logger bark.Logger
+	logger log.Logger
 	keys   *sync.Map
 }
 
 func (c *Collection) logNoValue(key Key, err error) {
 	_, loaded := c.keys.LoadOrStore(key, struct{}{})
 	if !loaded {
-		c.logger.Debugf("Failed to fetch key: %s from dynamic config with err: %s", key.String(), err.Error())
+		c.logger.Debug("Failed to fetch key: %s from dynamic config with err: %s", tag.Key(key.String()), tag.Error(err))
```
mfateev: Is key guaranteed to be non nil?

common/persistence/elasticsearch/elasticsearchVisibility.go
longquanzheng: this is a predefined tagValue we made in our first refactoring PR.https://github.com/uber/cadence/commit/e1eae286d4305ec9f1ed70744b8c28267bbacb6b#diff-53c1076ea850f00548dda5f08f711613R93

common/log/loggerimpl/logger.go
longquanzheng: Yes, I will clean up all the import order before checking in

common/persistence/cassandra/cassandraMetadataPersistenceProxy.go
longquanzheng: Better not. Firstly I don't think this is very common. A lot of them also have other tags.
Secondly I will also need to fix the callat tag if introducing such helper. We already had similar discussion about more helpers, like LifeCycle helpers and the answer is no.

common/service/dynamicconfig/config.go
longquanzheng: Yes I believe so, from the calling path (otherwise we already have panics)

common/service/config/log.go
longquanzheng: Probably next PR? let's narrow down the scope because I already have too many conflicts with everyone.

longquanzheng: Actually renaming this file should have conflict with others. What is a better name for you? For me log.go looks fine because it's not a pkg name. Usually file name just for reference to help understand what's in the file.

common/cache/domainCache_test.go
```diff
@@ -21,28 +21,28 @@
 package cache
 
 import (
-	"os"
+	"github.com/uber/cadence/common/log/loggerimpl"
```
meiliang86: Import order

common/cluster/metadata.go
```diff
@@ -22,10 +22,9 @@ package cluster
 
 import (
 	"fmt"
-	"github.com/uber-common/bark"
-	"github.com/uber/cadence/common/logging"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/log/tag"
```
meiliang86: Fix groups

common/log/logger_bench_test.go
```diff
@@ -20,6 +20,7 @@
 package log
 
 import (
+	"github.com/uber/cadence/common/log/loggerimpl"
```
meiliang86: fix import

common/log/logger_test.go
```diff
@@ -23,15 +23,16 @@ package log
 import (
 	"bytes"
 	"fmt"
+	"github.com/uber/cadence/common/log/loggerimpl"
```
meiliang86: fix import

meiliang86:
`func newErrorTag(key string, value error) Tag {
`
This should not take key as it is ignored.

meiliang86:
Please fix imports before checking in.

longquanzheng:
> `func newErrorTag(key string, value error) Tag { `
> This should not take key as it is ignored.

We want to pass in "error" key so that the key is explicitly reserved

PR #1722, opened 2019-04-19
sql: replace wide rows with blobs

venkat1109:
Today, some of the mysql tables related to mutable state have 30+ columns and growing. Using very wide rows result in the following disadvantages:

- Adding support for a new type of sql engine involves tedious work. And so is making changes to existing SQL queries that span several lines
- New features / improvements will involve adding new columns i.e. schema change. This can potentially affect availability for some sql based storage engines. In addition, its an additional step during deployment / rollout
- Performance Impact: A test performed on executions table show that write throughput can be improved by 2x and read throughput can be improved by about 50% by using blobs instead of wide rows. This test was performed on MySQL 5.7 (InnoDB) with multiple threads doing random reads / writes to the executions table.

Given the above, this patch changes the MySQL tables to use thriftrw blobs instead of wide rows. The change greatly simplified the sql queries and also some tables which used a combination of blobs + columns. Even with this change, there are a few exceptions where we still need columns in addition to blobs. These are:
 - Cases where we need conditional updates. Example - range_id, last_event_id
 - Cases where one or two columns are updated frequently. Example - lastHeartbeatTime
 - Tables which are deprecated (like eventsv1) are left untouched

Note on versioning: This patch does not add a "version" field to the thriftrw blob types. This is something that can be added when we need it


shreyassrivatsan:
Do we need some form of backward compatibility here? Are people using SQL outside? If so how will they migrate to this version only dealing with blobs.

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -183,102 +184,92 @@ func (m *sqlExecutionManager) GetWorkflowExecution(request *p.GetWorkflowExecuti
...
-		ClientLibraryVersion:         execution.ClientLibraryVersion,
-		ClientFeatureVersion:         execution.ClientFeatureVersion,
-		ClientImpl:                   execution.ClientImpl,
-		SignalCount:                  int32(execution.SignalCount),
-		HistorySize:                  execution.HistorySize,
-		CronSchedule:                 execution.CronSchedule,
+		TaskList:                     info.GetTaskList(),
+		WorkflowTypeName:             info.GetWorkflowTypeName(),
+		WorkflowTimeout:              info.GetWorkflowTimeoutSeconds(),
+		DecisionTimeoutValue:         info.GetDecisionTaskTimeoutMinutes(),
```
mfateev: I"m confused. Why is it in minutes?

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -183,102 +184,92 @@ func (m *sqlExecutionManager) GetWorkflowExecution(request *p.GetWorkflowExecuti
...
+		StartTimestamp:               time.Unix(0, info.GetStartTimeNanos()),
+		LastUpdatedTimestamp:         time.Unix(0, info.GetLastUpdatedTimeNanos()),
+		CreateRequestID:              info.GetCreateRequestID(),
+		DecisionVersion:              info.GetDecisionVersion(),
+		DecisionScheduleID:           info.GetDecisionScheduleID(),
+		DecisionStartedID:            info.GetDecisionStartedID(),
+		DecisionRequestID:            info.GetDecisionRequestID(),
+		DecisionTimeout:              info.GetDecisionTimeout(),
+		DecisionAttempt:              info.GetDecisionAttempt(),
+		DecisionTimestamp:            info.GetDecisionTimestampNanos(),
```
mfateev: Why StartTimestamp is converted using time.Unix and this one is not?

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1395,90 +1398,82 @@ func createExecutionFromRequest(
...
+		LastEventTaskID:              &request.LastEventTaskID,
+		LastProcessedEvent:           &request.LastProcessedEvent,
+		StartTimeNanos:               common.Int64Ptr(nowTimestamp.UnixNano()),
+		LastUpdatedTimeNanos:         common.Int64Ptr(nowTimestamp.UnixNano()),
+		CreateRequestID:              &request.RequestID,
+		DecisionVersion:              &request.DecisionVersion,
+		DecisionScheduleID:           &request.DecisionScheduleID,
+		DecisionStartedID:            &request.DecisionStartedID,
+		DecisionTimeout:              &request.DecisionStartToCloseTimeout,
+		DecisionAttempt:              common.Int64Ptr(0),
```
mfateev: emptyStr and no zeroPtr?

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1644,10 +1641,16 @@ func createTransferTasks(tx sqldb.Tx, transferTasks []p.Task, shardID int, domai
 			//d.logger.Fatal("Unknown Transfer Task.")
 		}
 
-		transferTasksRows[i].TaskID = task.GetTaskID()
-		transferTasksRows[i].TaskType = task.GetType()
-		transferTasksRows[i].Version = task.GetVersion()
-		transferTasksRows[i].VisibilityTimestamp = task.GetVisibilityTimestamp()
+		info.TaskType = common.Int16Ptr(int16(task.GetType()))
+		info.Version = common.Int64Ptr(task.GetVersion())
+		info.VisibilityTimestampNanos = common.Int64Ptr(task.GetVisibilityTimestamp().UnixNano())
```
mfateev: May be helper function for these?

venkat1109:
> Do we need some form of backward compatibility here? Are people using SQL outside? If so how will they migrate to this version only dealing with blobs.

SQL support is officially in alpha. We have a few customers playing with it today but they are aware things can change. There will be an announcement in the public slack room after this is landed.
And yes, it will be very hard to do changes like this once we are out of alpha - one of the main reasons I front loaded this PR

common/persistence/sql/sqlTaskManager.go
```diff
@@ -72,15 +73,23 @@ func (m *sqlTaskManager) LeaseTaskList(request *persistence.LeaseTaskListRequest
 		TaskType: common.Int64Ptr(int64(request.TaskType))})
 	if err != nil {
 		if err == sql.ErrNoRows {
+			tlInfo := &sqlblobs.TaskListInfo{
+				AckLevel:         &ackLevel,
+				Kind:             common.Int16Ptr(int16(request.TaskListKind)),
+				ExpiryTimeNanos:  common.Int64Ptr(0),
+				LastUpdatedNanos: common.Int64Ptr(time.Now().UnixNano()),
```
mfateev: Could we have something like timeNowPtr()?

common/persistence/sql/sqlExecutionManager.go
venkat1109: Good question. I don't know why - I must have read some code and thought DecisionTimeoutValue is in minutes. I hate when struct members represent numbers and have no units suffix - I will double confirm if DecisionTimeoutValue is actually seconds and if so, change the sqlblob also to be in seconds

common/persistence/sql/sqlExecutionManager.go
venkat1109: Because the existing persistence interface (which is the input / output to/from the persistence APIs)  passes a time.Time for StartTimestamp and uint64 for DecisionTimestamp

common/persistence/sql/sqlExecutionManager.go
venkat1109: will do

common/persistence/sql/sqlTaskManager.go
venkat1109: done

common/persistence/persistenceInterface.go
```diff
@@ -512,7 +513,8 @@ func NewDataBlob(data []byte, encodingType common.EncodingType) *DataBlob {
 		return nil
 	}
 	if encodingType != "thriftrw" && data[0] == 'Y' {
-		panic(fmt.Sprintf("Invlid incoding: \"%v\"", encodingType))
+		debug.PrintStack()
```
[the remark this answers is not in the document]
venkat1109: removed

PR #1724, opened 2019-04-22
Auto reset of open workflow by binary checksum

longquanzheng:


common/persistence/cassandra/cassandraMetadataPersistence.go
```diff
@@ -97,6 +100,7 @@ type (
 	cassandraMetadataPersistence struct {
 		cassandraStore
 		currentClusterName string
+		thriftEncoder      codec.BinaryEncoder
```
samarabbas: Usually we take care of any serialization/deserialization at a layer above and keep the storage layer completely agnostic of encoding/decoding.  Either use strongly typed structures all the way to db or pass in EncodedData.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1591,14 +1628,23 @@ func batchQueryHelper(batch *gocql.Batch, stmt string, useCondition bool, condit
...
-	replicationState *p.ReplicationState, cqlNowTimestamp int64, useCondition bool, condition int64) {
+	replicationState *p.ReplicationState, cqlNowTimestamp int64, useCondition bool, condition int64) error {
 	if executionInfo.ParentDomainID == "" {
 		executionInfo.ParentDomainID = emptyDomainID
 	}
 	if executionInfo.ParentRunID == "" {
 		executionInfo.ParentRunID = emptyRunID
 	}
 
+	var resetPointsBlob []byte
```
samarabbas: repeated logic in multiple places.  Please refactor.

service/history/mutableStateBuilder.go
```diff
@@ -1611,17 +1612,80 @@ func (e *mutableStateBuilder) CreateTransientDecisionEvents(di *decisionInfo, id
...
+	_, ok := resetPoints[binChecksum]
+	if ok {
+		return
+	}
+
+	if len(resetPoints) == maxResetPoints {
+		// if exceeding the max limit, do rotation by taking the oldest one out
+		var minP *workflow.ResetPointInfo
+		minK := ""
+		for k, p := range resetPoints {
```
samarabbas: Can we just have an ordered data structure?  we can also keep a map for lookup but just having lookup makes this logic unnecessarily complex.

service/history/transferQueueActiveProcessor.go
```diff
@@ -331,6 +331,12 @@ func (t *transferQueueActiveProcessorImpl) processDecisionTask(task *persistence
 		return nil
 	}
 
+	// compare binaryChecksum and auto-reset workflow if it has decision completed by bad binary(defined by domain userResetBinaries)
+	err = t.historyService.resetor.AutoResetWorkflow(context, msBuilder)
```
samarabbas: Let's do reset as a separate transfer task.

service/history/workflowResetor.go
```diff
@@ -881,3 +887,141 @@ func (w *workflowResetorImpl) replicateResetEvent(baseMutableState mutableState,
 	newMsBuilder.UpdateReplicationStateLastEventID(clusterMetadata.GetCurrentClusterName(), lastEvent.GetVersion(), lastEvent.GetEventId())
 	return
 }
+
+func (w *workflowResetorImpl) AutoResetWorkflow(currWorkflowContext workflowExecutionContext, currMutableState mutableState) (retError error) {
```
samarabbas: Is this logic only called through transfer task?  Can we separate out decision to reset from actual reset into two separate piece of logic.  I expect decision to reset to be really fast which does not need any database reads.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -857,6 +869,17 @@ struct DomainConfiguration {
   40: optional i32 archivalRetentionPeriodInDays
   50: optional ArchivalStatus archivalStatus
   60: optional string archivalBucketOwner
+  70: optional ResetBinaries userResetBinaries
+}
+
+struct ResetBinaries{
```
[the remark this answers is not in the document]
longquanzheng: Sure

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -857,6 +869,17 @@ struct DomainConfiguration {
...
+  10: optional map<string, ResetBinaryInfo> infos
+}
+
+struct ResetBinaryInfo{
```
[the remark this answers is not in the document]
longquanzheng: Sure

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -857,6 +869,17 @@ struct DomainConfiguration {
   40: optional i32 archivalRetentionPeriodInDays
   50: optional ArchivalStatus archivalStatus
   60: optional string archivalBucketOwner
+  70: optional ResetBinaries userResetBinaries
```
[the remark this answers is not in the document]
longquanzheng: Sure

common/persistence/cassandra/cassandraMetadataPersistence.go
longquanzheng: I wanted to do that at the beginning but found that SQL is taking thrift struct and serialize/deserialize inside for the whole workflowExecutionInfo blob. https://github.com/uber/cadence/blob/a19d2cb577aac678663c95482c49b35c739f302c/idl/github.com/uber/cadence/sqlblobs.thrift#L67

But you are also right, maybe we should keep consistent that these small blob should be serialize/deserialize in upper layer. For SQL, I should pass in a blob.  We can talk about that offline.

service/history/mutableStateBuilder.go
longquanzheng: Sure

service/history/transferQueueActiveProcessor.go
longquanzheng: Sure

service/history/workflowResetor.go
longquanzheng: Reset need to read database (current/base mutableState + base history) and have to be expensive. Either reading them in transferTask or inside this reset logic.
The only way to save perf is to reduce the chance of reset  failures because of:
1) base deleted
2) base not resettable
3) current not resettable
4) domainConfig change

PR #1725, opened 2019-04-22
Domain handler logic change

wxing1292:
* Add creation / update of local domain back
* Partially enable domain V1 -> V2 migration
* When creating local domain, will create record in domain V2 table.
* Domain update will use domain table with larger version, i.e. after using CLI migrating domain record from V1 to V2, update API will only update record in domain V2 table.
* Move domain CRUD test from using integration test to unit test for speed.


dmetzgar:
Do you want to add a CLI parameter to set the global domain flag?

wxing1292:
> Do you want to add a CLI parameter to set the global domain flag?

please see #1740 which add the CLI support

service/frontend/domainHandler.go
```diff
@@ -77,33 +76,40 @@ func (d *domainHandlerImpl) registerDomain(ctx context.Context,
...
-	// TODO remove the IsGlobalDomainEnabled check once cross DC is public
-	if clusterMetadata.IsGlobalDomainEnabled() && !clusterMetadata.IsMasterCluster() {
-		return errNotMasterCluster
+	if registerRequest.GetName() == "" {
+		return errDomainNotSet
 	}
-	if !clusterMetadata.IsGlobalDomainEnabled() {
+
+	if !d.clusterMetadata.IsGlobalDomainEnabled() {
+		registerRequest.IsGlobalDomain = common.BoolPtr(false)
```
longquanzheng: I would return an error if a customer want a global domain which is not supported, rather than silently change it to local domain.

longquanzheng:
BTW please try to make the title fit the limit if possible.

PR #1727, opened 2019-04-23
CLI: Support Memo feature

vancexu:
To start workflow with memo, use `-memo` flag (or `-memo_file`) with json input, eg:
``` ./cadence --do samples-domain workflow run --tl helloWorldGroup --wt main.Workflow --et 60 --dt 10 -i '"vancexu"' -memo '{"key1" : 1234321, "key2": "a string is very long", "map": {"mKey": 12343, "asd": "asdf"}} "vancexu" 123' -memo_key 'k1 k2 k3'```

To view result, use `-print_memo` (or `-pme`), eg:
Screen Shot 2019-04-23 at 3 57 23 PM

### DIFF
--- a/tools/cli/commands.go
+++ b/tools/cli/commands.go
@@ -65,6 +65,15 @@ const (
 
 	workflowStatusNotSet = -1
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
+
+	memoKeyForCLI = "CLI"
+)
+
+type jsonType int
+
+const (
+	jsonTypeInput jsonType = iota
+	jsonTypeMemo
 )
 
 // SetFactory is used to set the ClientFactory global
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tools/cli/commands.go [hunk 1]
```diff
@@ -500,6 +510,11 @@ func startWorkflowHelper(c *cli.Context, shouldPrintProgress bool) {
 	if c.IsSet(FlagCronSchedule) {
 		startRequest.CronSchedule = common.StringPtr(c.String(FlagCronSchedule))
 	}
+	memoStr := processJSONMemo(c)
+	if len(memoStr) != 0 {
+		fields := map[string][]byte{memoKeyForCLI: []byte(processJSONMemo(c))}
```
mfateev: I think the key should be passed as a parameter.

vancexu: I think this is simplest for user to use. 
An alternative is: provide one optional "memo-key" in start (but now user have to put same key when list)

PR #1732, opened 2019-04-24
CLI support list workflow output in json

vancexu:
https://github.com/uber/cadence/issues/1728
Add support to show list result in json, result is parsable json list, for machine processing.
Screen Shot 2019-04-26 at 5 43 13 PM

default json is not decoded not human readable, to read as human use `-pf` similar to show history events, which will try decode bytes to string. NOTE this -pf is not parsable json. 
Screen Shot 2019-04-26 at 5 44 52 PM


vancexu:
Unit tests will pass when https://github.com/uber/cadence/pull/1727 is merged

tools/cli/commands.go
```diff
@@ -1041,6 +1090,54 @@ func listClosedWorkflow(client client.Client, pageSize int, earliestTime, latest
...
+
+	return result, nextPageToken
+}
+
+// default will print decoded raw
+func printListResults(executions []*s.WorkflowExecutionInfo, inJson bool) {
+	for _, execution := range executions {
+		if inJson {
+			j, _ := json.Marshal(execution)
+			fmt.Println(string(j))
```
mfateev: Will it produce the parsable JSON list of execution structures?

vancexu: I see your point. It currently produces json lines,  
updated to produce a JSON list (also update top comments with latest result)

PR #1735, opened 2019-04-24
Add CLI to migration domain from V1 table to V2 table

wxing1292:


common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -270,6 +272,7 @@ func (m *CassandraMetadataPersistenceV2) UpdateDomain(request *p.UpdateDomainReq
 	return nil
 }
 
+// GetDomain gert a domain
```
dmetzgar: gert

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -92,14 +92,24 @@ const (
 )
 
 type (
-	cassandraMetadataPersistenceV2 struct {
+	// CassandraMetadataPersistenceV2 domain metadata v2 impl
+	// TODO make this struct private, once domain migration admin command is finished
```
vancexu: Another way to consider is to keep this private, but create a public function which takes interface MetadataManager, and cast to cassandraMetadataPersistenceV2 impl inside, then use this function in cli to do migrate

common/persistence/cassandra/cassandraMetadataPersistence.go
```diff
@@ -100,8 +100,16 @@ type (
...
+// NewMetadataPersistenceWithSession is used to create an instance of HistoryManager implementation
+func NewMetadataPersistenceWithSession(session *gocql.Session, clusterName string, logger log.Logger) p.MetadataStore {
+	return &cassandraMetadataPersistence{
+		cassandraStore:     cassandraStore{session: session, logger: logger},
+		currentClusterName: clusterName,
+	}
+}
+
+// NewMetadataPersistence is used to create an instance of HistoryManager implementation
+func NewMetadataPersistence(cfg config.Cassandra, clusterName string, logger log.Logger) (p.MetadataStore,
```
longquanzheng: nit: I don't think we need to expose this as public method. I would keep it private until we need it.

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
longquanzheng: +1 like we discussed offline

tools/cli/admin.go
```diff
@@ -245,6 +245,40 @@ func newAdminDomainCommands() []cli.Command {
 				AdminGetDomainIDOrName(c)
 			},
 		},
+		{
+			Name:  "migrate",
```
longquanzheng: Actually we don't have to merge this to master. This command is only for temporary and we won't use it anymore after migration. Adding a new command here will just confuse our customer. I would suggest making a private branch. And you don't have to wait for release and then adhoc deployment script to publish to every adhoc host.

wxing1292:
close in favor of #1924

PR #1737, opened 2019-04-25
Bug fix: schedule to start timer not fire after reset sticky tasklist

yux0:
reset sticky tasklist before sticky schedule to start time will cause workflow stuck

### DIFF
--- a/host/integration_test.go
+++ b/host/integration_test.go
@@ -2978,7 +2978,7 @@ func (s *integrationSuite) TestStickyTasklistResetThenTimeout() {
 	we, err0 := s.engine.StartWorkflowExecution(createContext(), request)
 	s.Nil(err0)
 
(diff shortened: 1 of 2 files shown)


host/integration_test.go
```diff
@@ -2945,6 +2945,175 @@ WaitForStickyTimeoutLoop:
...
+				s.Equal(workflow.TimeoutTypeScheduleToStart, event.DecisionTaskTimedOutEventAttributes.GetTimeoutType())
+				stickyTimeout = true
+				break WaitForStickyTimeoutLoop
+			}
+		}
+		time.Sleep(time.Second)
+	}
+	s.True(stickyTimeout, "Decision not timed out.")
+
+	for i := 0; i < 3; i++ {
```
wxing1292: what does this block tests?

PR #1739, opened 2019-04-25
sql: add cadence-sql-tool and refactor cadence-cassandra-tool

venkat1109:
This patch adds a tool similar to cadence-cassandra-tool for SQL based persistence. The tool reuses code as much as possible from the cadence-cassandra-tool. To enable this, a good amount of refactoring was needed to separate out the common pieces.


schema/mysql/v57/cadence/database.sql
```diff
@@ -1 +1 @@
-CREATE DATABASE cadence;
\ No newline at end of file
+CREATE DATABASE cadence character set utf8;
```
wxing1292: utf8mb4?

schema/mysql/v57/visibility/database.sql
```diff
@@ -1 +1 @@
-CREATE DATABASE cadence_visibility;
\ No newline at end of file
+CREATE DATABASE cadence_visibility character set utf8;
```
wxing1292: utf8mb4?

schema/mysql/v57/cadence/database.sql
venkat1109: utf8mb4 will limit the workflow_id length to 191 bytes instead of 255 bytes. On the other hand, the benefit you get with mb4 is support for additional unicode characters.. and it helps avoid some SQL injection attacks. We discussed this a bit last week and decided to go with utf8 because:
 - We don't need the support for additional unicode characters - we will update the documentation to make this clear
 - Cadence is not a customer facing service and probably will never be. So, the SQL injection is less of a concern, although its still legit

PR #1740, opened 2019-04-25
More domain changes related to re-enable local domain

wxing1292:
* Add CLI support for domain creation, with local / global domain
* Add protection in domain replication against domain name / UUID collision


tools/cli/domain.go
```diff
@@ -93,6 +93,10 @@ func newDomainCommands() []cli.Command {
 					Name:  FlagClustersWithAlias,
 					Usage: "Clusters",
 				},
+				cli.StringFlag{
```
vancexu: Better to use boolFlag

wxing1292: sure

denyska:
Please update the documentation to reflect that global_domain is required.













$ docker run --network=host --rm ubercadence/cli:master --domain samples-domain domain register 


Error: Option global_domain must be provided.

PR #1742, opened 2019-04-25
 Fix delete history when database timeout

longquanzheng:
TODO: also need to deal with MySQL


wxing1292:
we should also take the SQL timeout into consideration and combine those 2 (if SQL side actually provide specific timeout error)

form my understanding, SQL side throws context.DeadlineExceeded?

wxing1292:
the call chain look like this:
1. https://github.com/uber/cadence/blob/6ad9097d83109a4f03c9b979f7c5f20573114b9a/common/persistence/sql/common.go#L76
2. https://github.com/uber/cadence/blob/78d84ea199b0bba2828df3461e4e7b9a40edf5fd/common/persistence/sql/storage/mysql/db.go#L62
3. https://github.com/golang/go/blob/master/src/database/sql/sql.go#L2027

wxing1292:
according to https://play.golang.org/p/exW3pkfzs1m

context.DeadlineExceeded is probably correct for SQL

wxing1292:
actually, I think we should abandon the deletion, completely, then have a background job do the cleanup.

basically if the creation logic see timeout error, we cannot determine whether the creation of mutable state is a success or not, even calling get mutable state afterwards cannot guarantee anything, due to race condition (i.e. really slow network)

PR #1745, opened 2019-04-26
sql: allow configuration of connection pool settings

venkat1109:
This patch exposes sql connection pool settings to config yaml. Some refactoring of the persistence factory was needed to achieve this. One behavior difference is introduced (for SQL only) as part of this diff. Previously, we use a separate connection pool (sqlx.DB object) for each type of persistence manager (taskMgr, executionMgr etc). This makes connection pool settings like maxConns difficult unless we also expose separate settings for each one of the managers. On the other hand, for MySQL, there is no clear advantage to maintaining separate connection pools today. There could be some benefits wrt priority (i.e. execution manager gets more maxConns than taskMgr), but this is best enforced at higher layers (quotas & ratelimiting). So, this patch puts all managers under one connection pool.

### DIFF
--- a/common/persistence/persistence-factory/factory.go
+++ b/common/persistence/persistence-factory/factory.go
@@ -137,17 +137,8 @@ func New(
 		metricsClient: metricsClient,
 		logger:        logger,
 	}
-	defaultCfg := cfg.DataStores[cfg.DefaultStore]
-	visibilityCfg := cfg.DataStores[cfg.VisibilityStore]
 	limiters := buildRatelimiters(cfg)
-	factory.datastores = map[storeType]Datastore{
-		storeTypeTask:       newStore(defaultCfg, limiters[cfg.DefaultStore], clusterName, 0, logger),
-		storeTypeShard:      newStore(defaultCfg, limiters[cfg.DefaultStore], clusterName, 0, logger),
-		storeTypeMetadata:   newStore(defaultCfg, limiters[cfg.DefaultStore], clusterName, 0, logger),
-		storeTypeExecution:  newStore(defaultCfg, limiters[cfg.DefaultStore], clusterName, 0, logger),
-		storeTypeHistory:    newStore(defaultCfg, limiters[cfg.DefaultStore], clusterName, cfg.HistoryMaxConns, logger),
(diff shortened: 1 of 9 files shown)


common/persistence/persistence-factory/factory.go
```diff
@@ -305,29 +296,37 @@ func (f *factoryImpl) getCassandraConfig() *config.Cassandra {
...
-		return ds
-	}
-	ds.factory = newCassandraStore(*cfg.Cassandra, clusterName, maxConnsOverride, logger)
-	return ds
-}
+func (f *factoryImpl) init(clusterName string, limiters map[string]tokenbucket.TokenBucket) {
+	f.datastores = make(map[storeType]Datastore, len(storeTypes))
+	defaultCfg := f.config.DataStores[f.config.DefaultStore]
+	defaultDataStore := Datastore{ratelimit: limiters[f.config.DefaultStore]}
+	switch {
```
wxing1292: can we add a check if both defaultCfg.Cassandra and defaultCfg.SQL are not nil?

common/persistence/persistence-factory/factory.go
```diff
@@ -305,29 +296,37 @@ func (f *factoryImpl) getCassandraConfig() *config.Cassandra {
...
 	}
-	return sql.NewFactory(cfg, clusterName, logger)
-}
 
-func newCassandraStore(cfg config.Cassandra, clusterName string, maxConnsOverride int, logger log.Logger) DataStoreFactory {
-	if maxConnsOverride > 0 {
-		cfg.MaxConns = maxConnsOverride
+	visibilityCfg := f.config.DataStores[f.config.VisibilityStore]
+	visibilityDataStore := Datastore{ratelimit: limiters[f.config.VisibilityStore]}
+	switch {
```
wxing1292: same here

common/persistence/persistence-factory/factory.go
venkat1109: that's done at config.Validate() at initialization time. Over here, its a fatal condition and the default case fatals

common/persistence/persistence-factory/factory.go
venkat1109: same response as above

PR #1750, opened 2019-04-29
Auto reset persistence - metadataManager/Store

longquanzheng:
As part 2 of https://github.com/uber/cadence/pull/1724

### DIFF
--- a/common/persistence/cassandra/cassandraMetadataPersistenceV2.go
+++ b/common/persistence/cassandra/cassandraMetadataPersistenceV2.go
@@ -23,6 +23,8 @@ package cassandra
 import (
 	"fmt"
 
+	"github.com/uber/cadence/common"
+
 	"github.com/gocql/gocql"
 	workflow "github.com/uber/cadence/.gen/go/shared"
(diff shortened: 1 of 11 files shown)


common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -328,11 +339,12 @@ func (m *cassandraMetadataPersistenceV2) GetDomain(request *p.GetDomainRequest)
 	if info.Data == nil {
 		info.Data = map[string]string{}
 	}
+	config.BadBinaries = p.NewDataBlob(badBinariesData, common.EncodingType(badBinariesDataEncoding))
```
venkat1109: do you need  "!= nil" check or can the DataBlob can take nil ?

longquanzheng: NewDataBlob inside already handled that : https://github.com/uber/cadence/blob/master/common/persistence/persistenceInterface.go#L589

PR #1753, opened 2019-04-30
Revert #1611 for replication task solution which has better performance

wxing1292:
Compared to #1611, this PR shows ~60% reduction in replication task processing mean latency.

Revert "This PR aims to refactor the existing replication task processing, as… (#1611)"
* This reverts commit 9878a8f0797fe44f5442aa57fcbf0acaf99cb492.

This PR aims to refactor the existing replication task processing, as well as improve performance when applying large number of replication tasks concurrently to one workflow.

* Implement generic priority queue & UT
* Implement generic concurrent priority queue
* Move collection interface into one file
* Implement generic multi-sequential-task-queue processing logic & UT
* Refactor replication worker task processing logic
* Handle activity / history replication task using squential task queue & UT
* All replication retryable errors have upper attempt & duration, when exceeding the attempt | duration, replication message will be moved to DLQ


common/collection/initerface.go
```diff
@@ -0,0 +1,99 @@
...
+
+package collection
+
+type (
+	// Queue is the interface for queue
+	Queue interface {
+		// PeekItem returns the first item of the queue
+		Peek() interface{}
+		// Offer push an item to the queue
+		Offer(item interface{})
```
wxing1292: could you point me to the standard naming convention?

common/collection/initerface.go
```diff
@@ -0,0 +1,99 @@
...
+		// Poll pop an item from the queue
+		Poll() interface{}
+		// IsEmpty indicate if the queue is empty
+		IsEmpty() bool
+		// Size size of the queue
+		Size() int
```
wxing1292: i know, just one thing, if we are going to have a base interface for all collection we have, meaning including our concurrent hashmap, then does len make sense for map?

common/collection/initerface.go
```diff
@@ -0,0 +1,99 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package collection
+
+type (
+	// Queue is the interface for queue
+	Queue interface {
+		// PeekItem returns the first item of the queue
```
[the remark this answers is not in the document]
wxing1292: ok

common/task/interface.go
```diff
@@ -47,19 +46,27 @@ type (
...
-		TaskID() int64
+	}
+
+	// SequentialTaskQueue is the generic task queue interface which group
+	// sequential tasks to be executed one by one
+	SequentialTaskQueue interface {
+		// QueueID return the ID of the queue, as well as the tasks inside (same)
+		QueueID() interface{}
+		// Offer push an task to the task set
+		Offer(task SequentialTask)
```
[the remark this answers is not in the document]
wxing1292: sorry which collection.Queue interface?

common/task/sequentialTaskProcessor.go
```diff
@@ -90,105 +81,85 @@ func (t *sequentialTaskProcessorImpl) Stop() {
 }
 
 func (t *sequentialTaskProcessorImpl) Submit(task SequentialTask) error {
-	hashCode := int(task.HashCode()) % t.coroutineSize
-	taskQueue := t.coroutineTaskQueues[hashCode]
+
+	taskqueue := task.NewTaskQueue()
+	taskqueue.Offer(task)
+
+	_, fnEvaluated, err := t.taskqueues.PutOrDo(
```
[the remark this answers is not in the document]
wxing1292: sure, will add time / backlog related metrics for task queue processing

common/collection/initerface.go
```diff
@@ -0,0 +1,99 @@
...
+	HashFunc func(interface{}) uint32
+
+	// ActionFunc take a key and value, do calculation and return err
+	ActionFunc func(key interface{}, value interface{}) error
+	// PredicateFunc take a key and value, do calculation and return boolean
+	PredicateFunc func(key interface{}, value interface{}) bool
+
+	// ConcurrentTxMap is a generic interface for any
+	// implementation of a dictionary or a key value
+	// lookup table that is thread safe, and
```
wxing1292: done

common/task/interface.go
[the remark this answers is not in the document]
venkat1109: your common/collection.Queue interface

wxing1292:
> Left some comments related to metrics & naming. It's great that this approach is has better latency characteristics but it also bugs me a bit that you don't have a complete understanding of why the previous approach that relied on batch processing fares poorly (i.e. what's special about the workload characteristics that make the batch processing not work)
> 
> ShipIt after fixing the comments !

My understanding, the existing implementation does not have any flexibility dealing with kafka out of order. (Yep, i am saying that message is heavily out of order, due to kafka).

There are only few parts which participate in the "message order" related code pats:
1. replication queue in history service, this implementation is really simple, i.e. just read a replication task, get the corresponding history, and send it out. I do not expect this part to heavily reorder the message to be sent out since our testing setup's replication message size is more or less of the same size.
2. kafka client -> kafka server (send path)
3. kafka server -> kafka client (recv path)
4. this task processing logic

common/collection/initerface.go
wxing1292: sure, will change the function name to add / remove / peek / len

common/task/sequentialTaskProcessor.go
```diff
@@ -90,105 +81,85 @@ func (t *sequentialTaskProcessorImpl) Stop() {
 }
 
 func (t *sequentialTaskProcessorImpl) Submit(task SequentialTask) error {
-	hashCode := int(task.HashCode()) % t.coroutineSize
-	taskQueue := t.coroutineTaskQueues[hashCode]
+
+	taskqueue := task.NewTaskQueue()
```
wxing1292: changing the NewTaskQueue interface method to input factory function which can generate corresponding task queue for each type of task

common/collection/initerface.go
venkat1109: Are there good reasons to name these Offer() and Poll() as opposed to Add() and Remove() ? If not, can you stick with the standard Add()/Remove() naming convention ?

common/collection/initerface.go
venkat1109: Go standard naming convention for this is Len()

common/collection/initerface.go
venkat1109: comment truncated

common/task/sequentialTaskProcessor.go
venkat1109: it's a bit weird to ask a task to return a taskQ on which it should be enqueued for processing by a workerPool owned by the calling thread. Can you instead consider having your tasks implement one method - PartitionKey() Key  or GroupingKey() Key - where you can alias Key to WorkflowIdentifier for now - that way, you can create the queue here instead of creating the queue from within the task

PR #1761, opened 2019-04-30
worker: auto-register system domain

venkat1109:
This patch adds code to automatically register cadence-system domain if it doesn't exist at worker bootstrap time. This is specifically needed for docker backed by mysql.


service/worker/service.go
```diff
@@ -252,12 +259,55 @@ func (s *Service) startArchiver(base service.Service, pFactory persistencefactor
...
+
+func (s *Service) registerSystemDomain(pFactory persistencefactory.Factory, clusterName string) {
+	metadataV2, err := pFactory.NewMetadataManager(persistencefactory.MetadataV2)
+	if err != nil {
+		s.logger.Fatal("error creating metadataV2Mgr", tag.Error(err))
+	}
+	defer metadataV2.Close()
+	_, err = metadataV2.CreateDomain(&persistence.CreateDomainRequest{
+		Info: &persistence.DomainInfo{
+			ID:          uuid.New(),
```
wxing1292: maybe we should hard code the uuid?

venkat1109: Went back & forth on this one - hardcoded it in the end

PR #1775, opened 2019-05-02
File based dynamic config client

yycptt:
This implements https://github.com/uber/cadence/issues/1587

File based dynamic config client will read config from a file and store it in memory. Periodically, it will check the config file to see if it has been updated. If so, read and parse the file again.

This PR only contains the client implementation and unit tests. Right now, the file based client is not used by the cadence service. I can do that in a separate PR if needed.


common/service/dynamicconfig/fileBasedClient.go
```diff
@@ -0,0 +1,280 @@
...
+	client := &fileBasedClient{
+		config: config,
+		doneCh: doneCh,
+		logger: logger,
+	}
+	if err := client.update(); err != nil {
+		return nil, err
+	}
+	go func() {
+		ticker := time.NewTicker(client.config.PollInterval)
```
vancexu: PollInterval may need a default 5s to be defensive.

common/service/dynamicconfig/fileBasedClient.go
```diff
@@ -0,0 +1,280 @@
...
+		for {
+			select {
+			case <-ticker.C:
+				err := client.update()
+				if err != nil {
+					client.logger.Error("Failed to update dynamic config", tag.Timestamp(time.Now()))
+				}
+				client.logger.Info("Updated dynamic config", tag.Timestamp(time.Now()))
```
vancexu: would be better to put in else

yycptt: Updated.

common/service/dynamicconfig/fileBasedClient.go
yycptt: Added a validate function on start to check both poll interval and whether the config file exists.

PR #1778, opened 2019-05-02
Check bad binary in respondDecisionTask and pollForDecisionTask

longquanzheng:
In our design meeting we decided to check in PollForDecisionTask.
But as this requires customer to upgrade cadence-client, we should also check in decisionTaskComplete.


service/frontend/workflowHandler.go
```diff
@@ -491,6 +495,19 @@ func (wh *WorkflowHandler) PollForDecisionTask(
 	return resp, nil
 }
 
+func (wh *WorkflowHandler) checkBadBinary(domainEntry *cache.DomainCacheEntry, binaryChecksum string) error {
+	if domainEntry.GetConfig().BadBinaries.Binaries != nil {
+		badBinaries := domainEntry.GetConfig().BadBinaries.Binaries
+		_, ok := badBinaries[binaryChecksum]
+		if ok {
+			return &gen.BadRequestError{
```
samarabbas: I think we should fail the decision and also return error back to user for logging.

longquanzheng: Done

service/history/historyEngine.go
```diff
@@ -1289,6 +1289,13 @@ Update_History_Loop:
 		sizeChecker.completedID = completedID
 		sizeChecker.msBuilder = msBuilder
 
+		binChecksum := request.GetBinaryChecksum()
+		if _, ok := domainEntry.GetConfig().BadBinaries.Binaries[binChecksum]; ok {
+			failDecision = true
+			failCause = workflow.DecisionTaskFailedCauseBadBinary
+			failMessage = fmt.Sprintf("binary %v is already marked as bad deployment", binChecksum)
+			goto End_Of_Process_Decision_Loop
```
vancexu: I think it would be better to avoid using goto, otherwise it will become the first goto in cadence..

longquanzheng: OK

service/history/historyEngine.go
```diff
@@ -1289,479 +1289,486 @@ Update_History_Loop:
...
-				targetDomainID := domainID
-				attributes := d.ScheduleActivityTaskDecisionAttributes
-				// First check if we need to use a different target domain to schedule activity
-				if attributes.Domain != nil {
-					// TODO: Error handling for ActivitySchedule failed when domain lookup fails
-					domainEntry, err := e.shard.GetDomainCache().GetDomain(*attributes.Domain)
-					if err != nil {
-						return nil, &workflow.InternalServiceError{Message: "Unable to schedule activity across domain."}
+		binChecksum := request.GetBinaryChecksum()
+		if _, ok := domainEntry.GetConfig().BadBinaries.Binaries[binChecksum]; !ok {
```
shreyassrivatsan: Is there a way to structure this code without having to nest the whole decision loop? This makes the code in this function even more difficult to understand.

longquanzheng: Another option is to use goto. But what we actually need is to refactor this big for-loop.

longquanzheng: Make sure you selected "Hide whitespace changes" in the diff setting

longquanzheng: issue created: https://github.com/uber/cadence/issues/1783

service/history/historyEngine.go
```diff
@@ -1289,478 +1289,485 @@ Update_History_Loop:
...
-					metrics.DecisionTypeScheduleActivityCounter)
-				targetDomainID := domainID
-				attributes := d.ScheduleActivityTaskDecisionAttributes
-				// First check if we need to use a different target domain to schedule activity
-				if attributes.Domain != nil {
-					// TODO: Error handling for ActivitySchedule failed when domain lookup fails
-					domainEntry, err := e.shard.GetDomainCache().GetDomain(*attributes.Domain)
-					if err != nil {
-						return nil, &workflow.InternalServiceError{Message: "Unable to schedule activity across domain."}
+		binChecksum := request.GetBinaryChecksum()
```
samarabbas: what happens if no checksum is included in the request?

longquanzheng: GetBinaryChecksum will return "" and then won't match any key in the map.
I will add a unit test

service/history/historyEngine.go
shreyassrivatsan: maybe switch the if and else.. in the current code it is very difficult to figure out what's the effect of bad binaries existing as its buried after the full process_decision_loop..

longquanzheng: yes, agreed. Bowei and I had the same discussion. Already updated with that.

PR #1780, opened 2019-05-03
Refactor workflow creation logic for code reusability & maintainability 

wxing1292:
* Refactor active workflow creation logic
* Refactor passive workflow creation logic


service/history/historyEngine.go
```diff
@@ -21,7 +21,7 @@
 package history
 
 import (
-	"context"
+	ctx "context"
```
vancexu: this shortcut will be confusing since local var ctx is always used in code.

service/history/workflowExecutionContext.go
```diff
@@ -46,15 +46,20 @@ const (
 
 type (
 	workflowExecutionContext interface {
-		appendHistoryEvents(history []*workflow.HistoryEvent, transactionID int64, doLastEventValidation bool) (int, error)
+		appendFirstBatchHistoryForContinueAsNew(newStateBuilder mutableState, transactionID int64) error
+		appendFirstBatchEventsForActive(msBuilder mutableState) (int, error)
+		appendFirstBatchEventsForStandby(msBuilder mutableState, history []*workflow.HistoryEvent) (int, error)
 		clear()
 		continueAsNewWorkflowExecution(context []byte, newStateBuilder mutableState, transferTasks []persistence.Task, timerTasks []persistence.Task, transactionID int64) error
+		createWorkflowExecution(
```
vancexu: why `now` is also a param?

service/history/workflowExecutionContext.go
```diff
@@ -46,15 +46,20 @@ const (
 
 type (
 	workflowExecutionContext interface {
-		appendHistoryEvents(history []*workflow.HistoryEvent, transactionID int64, doLastEventValidation bool) (int, error)
+		appendFirstBatchHistoryForContinueAsNew(newStateBuilder mutableState, transactionID int64) error
+		appendFirstBatchEventsForActive(msBuilder mutableState) (int, error)
+		appendFirstBatchEventsForStandby(msBuilder mutableState, history []*workflow.HistoryEvent) (int, error)
```
vancexu: do we really need to expose 2 appendFirstBatchEvents 1 for active 1 for standby? looks like standby is only used in active

service/history/workflowExecutionContext.go
wxing1292: active side and passive side has different "now", active use the real clock, passive use the event time.

service/history/workflowExecutionContext.go
wxing1292: passive actually do use it, in here: https://github.com/uber/cadence/pull/1780/files#diff-3f1276f2a53b19aa2890fb159f550a2dR687

service/history/historyEngine_test.go
```diff
@@ -5187,6 +5236,9 @@ func copyActivityInfo(sourceInfo *persistence.ActivityInfo) *persistence.Activit
 		MaximumInterval:          sourceInfo.MaximumInterval,
 		ExpirationTime:           sourceInfo.ExpirationTime,
 		MaximumAttempts:          sourceInfo.MaximumAttempts,
+		NonRetriableErrors:       sourceInfo.NonRetriableErrors,
+		//// Not written to database - This is used only for deduping heartbeat timer creation
+		// LastHeartbeatTimeoutVisibility: sourceInfo.LastHeartbeatTimeoutVisibility,
```
yux0: remove?

wxing1292: nop, 
1. this is test code
2. i want to put this here for reference, since previous code does not copy / paste everything

yux0: remove?

vancexu: why //// not //?

service/history/workflowExecutionContext.go
```diff
@@ -717,6 +802,50 @@ func (c *workflowExecutionContextImpl) update(transferTasks []persistence.Task,
...
+	var err error
+	if msBuilder.GetEventStoreVersion() == persistence.EventStoreVersionV2 {
+		historySize, err = c.shard.AppendHistoryV2Events(&persistence.AppendHistoryNodesRequest{
+			IsNewBranch: true,
+			Info:        historyGarbageCleanupInfo(c.domainID, c.workflowExecution.GetWorkflowId(), c.workflowExecution.GetRunId()),
+			BranchToken: msBuilder.GetCurrentBranch(),
+			Events:      history,
+			// It is ok to use 0 for TransactionID because RunID is unique so there are
+			// no potential duplicates to override.
+			TransactionID: 0,
```
yux0: I am not sure. But it might need to also pass the shard id to support mysql.

PR #1781, opened 2019-05-03
Bugfix: workflow reset return wrong run ID

wxing1292:


Gopkg.lock
```diff
@@ -989,6 +989,7 @@
     "github.com/hashicorp/go-version.git",
     "github.com/iancoleman/strcase",
     "github.com/jmoiron/sqlx",
+    "github.com/m3db/prometheus_client_golang/prometheus",
```
longquanzheng: remove this line

wxing1292: why? dep seems always wang to add this

PR #1782, opened 2019-05-03
config: remove statsd validation

jozuenoon:
Since we are trying now to setup cadence with Prometheus metrics (recently added) and drop statsd entirely it should be possible to omit statsd configuration altogether. However due to config validation checks cadence will fail on startup.

### DIFF
--- a/common/service/config/config.go
+++ b/common/service/config/config.go
@@ -263,9 +263,9 @@ type (
 	// Statsd contains the config items for statsd metrics reporter
 	Statsd struct {
 		// The host and port of the statsd server
-		HostPort string `yaml:"hostPort" validate:"nonzero"`
+		HostPort string `yaml:"hostPort"`
 		// The prefix to use in reporting to statsd
-		Prefix string `yaml:"prefix" validate:"nonzero"`
+		Prefix string `yaml:"prefix"`
 		// FlushInterval is the maximum interval for sending packets.
 		// If it is not specified, it defaults to 1 second.
 		FlushInterval time.Duration `yaml:"flushInterval"`


venkat1109:
@jozuenoon  - can you provide an example of config that fails validation ? Even though statsd requires some members to be set, statsd, by itself is optional:

```
// Metrics contains the config items for metrics subsystem
	Metrics struct {
		// M3 is the configuration for m3 metrics reporter
		M3 *m3.Configuration `yaml:"m3"`
		// Statsd is the configuration for statsd reporter
		Statsd *Statsd `yaml:"statsd"`
		// Prometheus is the configuration for prometheus reporter
		Prometheus *prometheus.Configuration `yaml:"prometheus"`
		// Tags is the set of key-value pairs to be reported
		// as part of every metric
		Tags map[string]string `yaml:"tags"`
	}
```

jozuenoon:
I believe this is relevant part of it at least:
```
services:
    frontend:
        rpc:
            port: 7933
            bindOnIP: 0.0.0.0
        metrics:
            prometheus:
                timerType: histogram
                listenAddress: 0.0.0.0:8000

    matching:
        rpc:
            port: 7935
            bindOnIP: 0.0.0.0
        metrics:
            prometheus:
                timerType: histogram
                listenAddress: 0.0.0.0:8001

    history:
        rpc:
            port: 7934
            bindOnIP: 0.0.0.0
        metrics:
            prometheus:
                timerType: histogram
                listenAddress: 0.0.0.0:8002

    worker:
        rpc:
            port: 7939
            bindOnIP: 0.0.0.0
        metrics:
            prometheus:
                timerType: histogram
                listenAddress: 0.0.0.0:8003
```

End the error I get:
```
"2019/05/03 14:44:39 Config file corrupted.Services[worker](value).Metrics.Statsd.HostPort: zero value"  
```

jozuenoon:
Related helm chart update if anybody up for testing also: https://github.com/banzaicloud/banzai-charts/pull/788

shreyassrivatsan:
@jozuenoon can you post the full config you are using. I am able to start cadence server all the roles with the prometheus config without any statsd or m3 config..

jozuenoon:
Thanks! It seems like it was config file naming mismatch. Finally figured out that I need add `cassandra` to config file template name, so it's now ingested with `start.sh` script.

PR #1784, opened 2019-05-03
Remove unused visibility code

wxing1292:


service/history/transferQueueActiveProcessor.go
```diff
@@ -21,7 +21,7 @@
 package history
 
 import (
-	"context"
+	ctx "context"
```
vancexu: I think it has little benefits to shortcut context. 
People get used to use `context` package already and when people read ctx.Func they may have question mark..

wxing1292: the thing is, most context used by history service is workflow execution context...
we can determine whether make that ctx and go's context a context

vancexu: my preference is using context for standard go package, and name workflow context explicit wfCtx, so that whoever read code will have no suprise

PR #1786, opened 2019-05-03
Print auto-reset points in describeWorkflow CLI

longquanzheng:
tested locally:
```
longer@~/gocode/src/github.com/uber/cadence:(descWF)$ ./cadence --do samples-domain   wf desc -w cron_3f994e77-804d-4553-b2b0-900813ae54e3
{
  "ExecutionConfiguration": {
    "taskList": {
      "name": "cronGroup"
    },
    "executionStartToCloseTimeoutSeconds": 1200,
    "taskStartToCloseTimeoutSeconds": 60,
    "childPolicy": "TERMINATE"
  },
  "WorkflowExecutionInfo": {
    "Execution": {
      "workflowId": "cron_3f994e77-804d-4553-b2b0-900813ae54e3",
      "runId": "f6345419-f79b-4ff8-b867-f1b18b131e82"
    },
    "Type": {
      "name": "main.SampleCronWorkflow"
    },
    "StartTime": "2019-05-02T13:42:06-07:00",
    "CloseTime": "2019-05-02T16:21:31-07:00",
    "CloseStatus": "TIMED_OUT",
    "HistoryLength": 2130,
    "ParentDomainID": "10000000-0000-f000-f000-000000000000",
    "ParentExecution": {
      "workflowId": "",
      "runId": "30000000-0000-f000-f000-000000000000"
    },
    "AutoResetPoints": {
      "points": [
        {
          "binaryChecksum": "641a58009fe186a0d8e58f36ce37cd1e",
          "runId": "f6345419-f79b-4ff8-b867-f1b18b131e82",
          "firstDecisionCompletedId": 421,
          "createdTimeNano": 1556829820725600000,
          "resettable": true
        }
      ]
    }
  },
  "PendingActivities": null
}
Auto Reset Points:
+----------------------------------+--------------------------------+------------+
|         BINARY CHECKSUM          |          CREATE TIME           | RESETTABLE |
+----------------------------------+--------------------------------+------------+
| 641a58009fe186a0d8e58f36ce37cd1e | 2019-05-02 13:43:40.7256 -0700 | true       |
|                                  | PDT                            |            |
+----------------------------------+--------------------------------+------------+
```


wxing1292:
in terms of code, LGTM

in terms of style, @vancexu

longquanzheng:
In terms of printing style, it's similar to https://github.com/uber/cadence/pull/1751 
It's according to what we decided in the design doc.

But feel free to point out if I am wrong.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -309,6 +309,7 @@ struct WorkflowExecutionInfo {
   80: optional WorkflowExecution parentExecution
   90: optional i64 (js.type = "Long") executionTime
   100: optional Memo memo
+  110: optional ResetPoints autoResetPoints
```
vancexu: Do you think it is possible to move this out to DescribeResponse? Reason is WorkflowExecutionInfo is used multiple places in visibility, unless you want such things in list result, I think it would be better to move out.

tools/cli/commands.go
```diff
@@ -967,6 +967,24 @@ func describeWorkflowHelper(c *cli.Context, wid, rid string) {
 		o = convertDescribeWorkflowExecutionResponse(resp)
 	}
 	prettyPrintJSONObject(o)
+	if resp.WorkflowExecutionInfo.AutoResetPoints != nil && len(resp.WorkflowExecutionInfo.AutoResetPoints.Points) > 0 {
```
vancexu: Can we add a flag, reason is https://github.com/uber/cadence/issues/1728 we may want most command be able to print parseable json. 
This table looks an add-on, I would not want to see it in `wf desc`, but would like to only see this when `wf desc -preset`

idl/github.com/uber/cadence/shared.thrift
longquanzheng: Addressed offline.

tools/cli/commands.go
longquanzheng: Will do.

PR #1787, opened 2019-05-03
Handle execution time query

yycptt:
Since the execution time is set to 0 for workflows that are not cron or currently during retry, we need to  add some special handling to the query user provided to the new List, Scan and Count visibility API. 

The basic idea here is that, whenever user query contains a range query on execution time, add another condition that the execution time must greater than 0.


yycptt:
https://github.com/uber/cadence/issues/1754

PR #1788, opened 2019-05-03
Fix yarpc registration data race

wxing1292:


wxing1292:
close in favor of #1828

PR #1790, opened 2019-05-06
Parse time in visibility query string

yycptt:
Right now in the query string of the new visibility API, user can only use UnixNano time when doing filtering (on startTime, closeTime or executionTime). 

This PR adds additional time parsing functionality to the API, so user can use RFC3339 time format when writing queries.


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -793,3 +811,46 @@ func checkPageSize(request *p.ListWorkflowExecutionsRequestV2) {
...
+	}, func(obj *fastjson.Object, key string, v *fastjson.Value) {
+		timeStr := string(v.GetStringBytes())
+
+		// try to parse time
+		parsedTime, err := time.Parse(defaultDateTimeFormat, timeStr)
+		if err == nil {
+			obj.Set(key, fastjson.MustParse(fmt.Sprintf(`"%v"`, parsedTime.UnixNano())))
+		}
+
+		// do nothing if failed to parse
```
vancexu: We probably shouldn't eat this error; caller should be able to return BadRequest when timeStr not parseable

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -793,3 +811,46 @@ func checkPageSize(request *p.ListWorkflowExecutionsRequestV2) {
 		request.PageSize = 1000
 	}
 }
+
+func processAllValuesForKey(dsl *fastjson.Value, keyFilter func(k string) bool,
```
vancexu: Would be cool to have unit tests for all these awesome fined grained functions. 
The goal is to have as much coverage as possible, also easier for debugging when test fails.

PR #1794, opened 2019-05-06
Auto-reset: do not trigger/process reset task when reset point has e…

longquanzheng:
…xpired

### DIFF
--- a/service/history/workflowResetor.go
+++ b/service/history/workflowResetor.go
@@ -864,11 +864,9 @@ func FindAutoResetPoint(badBinaries *workflow.BadBinaries, autoResetPoints *work
 	for _, p := range autoResetPoints.Points {
 		bin, ok := badBinaries.Binaries[p.GetBinaryChecksum()]
 		if ok && p.GetResettable() {
-			if p.GetExpiringTimeNano() > 0 {
-				if nowNano > p.GetExpiringTimeNano() {
-					// reset point has expired and we may already deleted the history
-					continue
-				}
+			if p.GetExpiringTimeNano() > 0 && nowNano > p.GetExpiringTimeNano() {
+				// reset point has expired and we may already deleted the history
+				continue
 			}
 			return bin.GetReason(), p
 		}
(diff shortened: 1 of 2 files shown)


service/history/workflowResetor_test.go
```diff
@@ -4204,4 +4219,31 @@ func TestFindAutoResetPoint(t *testing.T) {
...
+		Binaries: map[string]*workflow.BadBinaryInfo{
+			"none":    {},
+			"expired": {},
+		},
+	}, &workflow.ResetPoints{
+		Points: []*workflow.ResetPointInfo{
+			pt0, pt1, pt3, pt4, pt5,
+		},
+	})
+	fmt.Println(pt.GetBinaryChecksum())
```
shreyassrivatsan: spurious print?

service/history/workflowResetor.go
```diff
@@ -860,9 +860,16 @@ func FindAutoResetPoint(badBinaries *workflow.BadBinaries, autoResetPoints *work
 	if badBinaries == nil || badBinaries.Binaries == nil || autoResetPoints == nil || autoResetPoints.Points == nil {
 		return
 	}
+	nowNano := time.Now().UnixNano()
 	for _, p := range autoResetPoints.Points {
 		bin, ok := badBinaries.Binaries[p.GetBinaryChecksum()]
 		if ok && p.GetResettable() {
+			if p.GetExpiringTimeNano() > 0 {
+				if nowNano > p.GetExpiringTimeNano() {
```
shreyassrivatsan: nit: combine these conditions into a single if statement

shreyassrivatsan:
Approved with a couple of comments

service/history/workflowResetor.go
longquanzheng: Sure

service/history/workflowResetor_test.go
longquanzheng: Oh thanks for pointing it out.

PR #1795, opened 2019-05-06
Switch to File Based Dynamic Config Client

yycptt:


common/service/config/config.go
```diff
@@ -57,6 +57,9 @@ type (
 		ElasticSearch elasticsearch.Config `yaml:"elasticsearch"`
 		// PublicClient is config for connecting to cadence frontend
 		PublicClient PublicClient `yaml:"publicClient"`
+		// DCClientConfig is the config for setting up the file based dynamic config client
+		// Filepath should be relative to the root directory
+		DCClientConfig dynamicconfig.FileBasedClientConfig `yaml:"dcClientConfig"`
```
vancexu: Let's rename to "DynamicConfigClient", 
dc may not sounds familiar to customers, and team may confuse it with data center.

config/base.yaml
```diff
@@ -1,3 +1,7 @@
 log:
   stdout: true
   level: info
+
+dcClientConfig:
```
vancexu: Let's move it to `development.yaml`. 
Since this is good to have but still optional

cmd/server/server.go
```diff
@@ -111,7 +112,11 @@ func (s *server) startService() common.Daemon {
 		log.Fatalf("error creating ringpop factory: %v", err)
 	}
 
-	params.DynamicConfig = dynamicconfig.NewNopClient()
+	params.DynamicConfig, err = dynamicconfig.NewFileBasedClient(&s.cfg.DCClientConfig, params.Logger.WithTags(tag.Service(params.Name)), s.doneC)
+	if err != nil {
+		log.Printf("error creating file based dynamic config client: %v", err)
```
vancexu: I would make it:
"cannot create filed based dynamic config client, so will use no-op config client instead. error: %v"

config/dynamicconfig.yaml
```diff
@@ -0,0 +1,37 @@
+# Use this file to override the default dynamic config value (they are specified
```
vancexu: Let's move sample config to path `config/dynamic_config/dev.yaml`

common/service/config/config.go
yycptt: Update.

config/base.yaml
yycptt: Done

cmd/server/server.go
yycptt: Done.

config/dynamicconfig.yaml
yycptt: Done

PR #1801, opened 2019-05-07
metrics/prometheus: fix invalid metrics names

jozuenoon:
Some extra metrics names which were missed at https://github.com/uber/cadence/pull/1770.

Errors I got look like following:
```
descriptor Desc{fqName: "memory.heapidle", help: "memory.heapidle gauge", constLabels: {}, variableLabels: []} is invalid: "memory.heapidle" is not a valid metric name
```


shreyassrivatsan:
This should be superseded by #1814

jozuenoon:
Thanks! Closing then.

PR #1802, opened 2019-05-07
Initial skeleton implementation for 3+DC history replication task processing logic

wxing1292:


service/history/stateBuilder.go
```diff
@@ -102,23 +104,26 @@ func (b *stateBuilderImpl) getNewRunTimerTasks() []persistence.Task {
 func (b *stateBuilderImpl) applyEvents(domainID, requestID string, execution shared.WorkflowExecution,
 	history []*shared.HistoryEvent, newRunHistory []*shared.HistoryEvent, eventStoreVersion, newRunEventStoreVersion int32) (*shared.HistoryEvent,
 	*decisionInfo, mutableState, error) {
-	var lastEvent *shared.HistoryEvent
+
+	if len(history) == 0 {
+		return nil, nil, nil, errors.NewInternalFailureError(ErrMessageHistorySizeZero)
```
yux0: Is this fixing an existing bug or 3dc related? Looks like the behavior changed

service/history/stateBuilder.go
```diff
@@ -102,23 +104,26 @@ func (b *stateBuilderImpl) getNewRunTimerTasks() []persistence.Task {
...
 	for _, event := range history {
-		lastEvent = event
 		// NOTE: stateBuilder is also being used in the active side
 		if b.msBuilder.GetReplicationState() != nil {
+			// this function must be called within the for loop, in case
+			// history event version changed during for loop
 			b.msBuilder.UpdateReplicationStateVersion(event.GetVersion(), true)
 		}
-		b.msBuilder.GetExecutionInfo().LastEventTaskID = event.GetTaskId()
+		b.msBuilder.GetExecutionInfo().LastEventTaskID = lastEvent.GetTaskId()
```
yux0: event.GetTaskId?

service/history/historyReplicatorV2.go
```diff
@@ -0,0 +1,488 @@
...
+
+	// current workflow is completed
+	if currentState == persistence.WorkflowStateCompleted {
+		// allow the application of workflow creation if currentLastWriteVersion > incomingVersion
+		// because this can be caused by missing replication events
+		// proceed to create workflow
+		createMode = persistence.CreateWorkflowModeWorkflowIDReuse
+		prevRunID = currentRunID
+		prevLastWriteVersion = currentLastWriteVersion
+		return context.createWorkflowExecution(
```
yux0: If the workflow is already started why create workflow execution again?

service/history/stateBuilder.go
wxing1292: good catch

service/history/stateBuilder.go
wxing1292: there is no reason to use the applyEvents function with 0 history

service/history/historyReplicatorV2.go
wxing1292: basically, the replication path is trying to be optimistic.

each time, when creating a workflow, we need to know either there is no current record of that workflow, or that workflow is closed, plus its run ID and last write version.

the overall logic is
1. try to create the workflow optimistically, i.e. assuming that there is no existing workflow in DB
2. either 1 goes through, or see a WorkflowExecutionAlreadyStartedError
3. if see WorkflowExecutionAlreadyStartedError, then use the info in the error to continue create the workflow

PR #1809, opened 2019-05-08
Deprecate clustersInfo configuration in favor of new clusterMetadata…

wxing1292:
… configuration, which provides better flexibility

* Add conversion function which turn clustersInfo into clusterMetadata
* State builder in history service will always set the replication state whenever applying an event.
* Fix & update workflow execution reset test code.

New config change allows a cluster to be disabled, which can be useful for cluster migration using cross DC

Before:
```
clustersInfo:
  enableGlobalDomain: false
  failoverVersionIncrement: 10
  masterClusterName: "active"
  currentClusterName: "active"
  clusterInitialFailoverVersion:
    active: 0
  clusterAddress:
    active:
      rpcName: "cadence-frontend"
      rpcAddress: "127.0.0.1:7933"
```

After:
```
clusterMetadata:
  enableGlobalDomain: false
  failoverVersionIncrement: 10
  masterClusterName: "active"
  currentClusterName: "active"
  clusterInformation:
    active:
      enabled: true
      initialFailoverVersion: 0
      rpcName: "cadence-frontend"
      rpcAddress: "127.0.0.1:7933"

```


common/service/config/config.go
```diff
@@ -237,13 +242,39 @@ type (
...
+	// Deprecated: please use ClusterInformation instead
 	Address struct {
 		// RPCName indicate the remote service name
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service IP address
 		RPCAddress string `yaml:"rpcAddress"`
 	}
 
+	// ClustersInformation contains the all cluster which participated in cross DC
+	ClustersInformation struct {
```
longquanzheng: Can we use any other name other than ClustersInfomation. The reason is we have ClusterInfomation. For example: GlobalReplicationConfig

common/service/config/config.go
```diff
@@ -219,6 +222,8 @@ type (
 	}
 
 	// ClustersInfo contains the all cluster names and active cluster
+	//
+	// Deprecated: please use ClusterInformation instead
 	ClustersInfo struct {
```
longquanzheng: I would recommend to remove all configs in ClustersInfo(make it an empty struct) and keep our code simple, and panic if ClustersInfo is specified.

common/service/config/config.go
```diff
@@ -237,13 +242,39 @@ type (
...
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service IP address
 		RPCAddress string `yaml:"rpcAddress"`
 	}
 
+	// ClustersInformation contains the all cluster which participated in cross DC
+	ClustersInformation struct {
+		EnableGlobalDomain bool `yaml:"enableGlobalDomain"`
+		// VersionIncrement is the increment of each cluster version when failover happens
+		VersionIncrement int64 `yaml:"versionIncrement"`
```
longquanzheng: Let's go back to FailoverVersionIncrement as we said

cmd/server/server.go
```diff
@@ -123,22 +124,30 @@ func (s *server) startService() common.Daemon {
...
 		params.Logger,
 		params.MetricsClient,
-		enableGlobalDomain,
-		s.cfg.ClustersInfo.FailoverVersionIncrement,
-		s.cfg.ClustersInfo.MasterClusterName,
-		s.cfg.ClustersInfo.CurrentClusterName,
-		s.cfg.ClustersInfo.ClusterInitialFailoverVersions,
-		s.cfg.ClustersInfo.ClusterAddress,
+		dc.GetBoolProperty(dynamicconfig.EnableGlobalDomain, clustersInformation.EnableGlobalDomain),
+		clustersInformation.VersionIncrement,
```
longquanzheng: Please rename back for all: version->failoverVersion to be consistent

PR #1822, opened 2019-05-09
Adding conflict resolver version 2 for 3dc branch conflict resolving

yux0:
This change is related to 3 DC. In the new conflict resolver, it gets the result using PaginateHistory to get a batch result of BatchHistory. On each history batch, it calculates the first and next event id. 

#1455

No unit test added at this moment as the code requires updates when versionHistories introduce into the mutable state


service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,191 @@
...
+		if batchSize > eventsToApply {
+			history = history[0:eventsToApply]
+		}
+		eventsToApply -= int64(len(history))
+
+		if len(history) == 0 {
+			break
+		}
+
+		firstEvent := history[0]
```
wxing1292: plz put the application of history event logic into a dedicated function so the for loop is simple

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,191 @@
...
+			sBuilder = newStateBuilder(r.shard, resetMutableStateBuilder, r.logger)
+		}
+
+		// NOTE: passing 0 as newRunEventStoreVersion is safe here, since we don't need the newMutableState of the new run
+		_, _, _, err = sBuilder.applyEvents(domainID, requestID, execution, history, nil, resetMutableStateBuilder.GetEventStoreVersion(), 0)
+		if err != nil {
+			r.logger.Error("conflict resolution err applying events.", tag.Error(err))
+			return nil, err
+		}
+		resetMutableStateBuilder.executionInfo.SetLastFirstEventID(lastFirstEventID)
```
wxing1292: as long as state builder is applying history event by batch, the SetLastFirstEventID, SetNextEventID can be omitted (done by state builder)

wxing1292: same applies to UpdateReplicationStateLastEventID

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,191 @@
...
+
+	for iter.HasNext() {
+		item, err := iter.Next()
+		if err != nil {
+			r.logger.Error("Conflict resolution err getting history.", tag.Error(err))
+			return nil, err
+		}
+
+		history := item.(*shared.History).Events
+		batchSize := int64(len(history))
```
wxing1292: the verification logic here is not correct, should assert on last event on the last batch has the event ID == replayEventID

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,191 @@
...
+
+	var resetMutableStateBuilder *mutableStateBuilder
+	var sBuilder stateBuilder
+	var lastEvent *shared.HistoryEvent
+	var lastFirstEventID int64
+
+	eventsToApply := replayNextEventID - common.FirstEventID
+
+	iter := collection.NewPagingIterator(func(paginationToken []byte) ([]interface{}, []byte, error) {
+		var paginateItems []interface{}
```
wxing1292: plz merge this function with getHistoryBatch

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+func newConflictResolverV2(shard ShardContext, context workflowExecutionContext, historyMgr persistence.HistoryManager, historyV2Mgr persistence.HistoryV2Manager,
+	logger log.Logger) *conflictResolverImpl {
+
+	return &conflictResolverImpl{
+		shard:           shard,
+		clusterMetadata: shard.GetService().GetClusterMetadata(),
+		context:         context,
+		historyMgr:      historyMgr,
```
wxing1292: we only plan to support history events v2, this should not be necessary

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/uber/cadence/common/cluster"
+	"github.com/uber/cadence/common/collection"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/persistence"
+)
+
+const (
+	defaultPageSize   = 100
+	eventStoreVersion = persistence.EventStoreVersionV2
```
wxing1292: plz rename to the following, or otherwise the names are confusing since these variables are shared within the package 
conflictResolverDefaultPageSize
conflictResolverEventStoreVersion

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		_, _, _, err = sBuilder.applyEvents(domainID, requestID, execution, history, nil, eventStoreVersion, eventStoreVersion)
+		if err != nil {
+			r.logger.Error("conflict resolution err applying events.", tag.Error(err))
+			return nil, err
+		}
+		resetMutableStateBuilder.IncrementHistorySize(len(history))
+	}
+	//Sanity check on last event id of the last history batch
+	if *lastEvent.EventId != replayEventID {
```
wxing1292: GetEventId()

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+		historyMgr:      historyMgr,
+		historyV2Mgr:    historyV2Mgr,
+		logger:          logger,
+	}
+}
+
+func (r *conflictResolverV2Impl) reset(prevRunID string, requestID string, replayEventID int64, info *persistence.WorkflowExecutionInfo, localVersionHistories persistence.VersionHistories) (mutableState, error) {
+	domainID := r.context.getDomainID()
+	execution := *r.context.getExecution()
+	branchToken := info.GetCurrentBranch()
```
wxing1292: this should come from LAC's corresponding branch, for now make it a empty []byte and add a TODO

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+	msBuilder, err := r.resetMutableState(resetMutableStateBuilder, info, lastEvent, replayEventID, prevRunID)
+	if err != nil {
+		r.logger.Error("conflict resolution err reset workflow.", tag.Error(err))
+	}
+	//TODO: use LCA on the msBuilder.getVersionHistories() vs localVersionHistories
+	//r.resolveVersionHistoryConflict(localVersionHistories, msBuilder.GetVersionHistories())
+	return msBuilder, err
+}
+
+func (r *conflictResolverV2Impl) resolveVersionHistoryConflict(local persistence.VersionHistories, remote persistence.VersionHistories) error {
```
wxing1292: what does this function do?

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+	return resetMutableStateBuilder, sBuilder
+}
+
+func (r *conflictResolverV2Impl) resetMutableState(msBuilder *mutableStateBuilder,
+	execution *persistence.WorkflowExecutionInfo,
+	lastEvent *shared.HistoryEvent,
+	replayEventID int64,
+	prevRunID string) (mutableState, error) {
+	// reset branchToken to the original one(it has been set to a wrong branchToken in applyEvents for startEvent)
+	msBuilder.executionInfo.BranchToken = msBuilder.GetCurrentBranch()
```
wxing1292: the branch should change, for now use a empty []byte and add a TODO

service/history/conflictResolverV2.go
```diff
@@ -0,0 +1,186 @@
...
+	// similarly, in case of resetWF, the runID in startEvent is incorrect
+	msBuilder.executionInfo.RunID = execution.RunID
+	msBuilder.executionInfo.StartTimestamp = execution.StartTimestamp
+	// the last updated time is not important here, since this should be updated with event time afterwards
+	msBuilder.executionInfo.LastUpdatedTimestamp = execution.StartTimestamp
+	sourceCluster := r.clusterMetadata.ClusterNameForFailoverVersion(lastEvent.GetVersion())
+	msBuilder.UpdateReplicationStateLastEventID(sourceCluster, lastEvent.GetVersion(), replayEventID)
```
wxing1292: the following 2 lines are handled by state builder
```
sourceCluster := r.clusterMetadata.ClusterNameForFailoverVersion(lastEvent.GetVersion())
msBuilder.UpdateReplicationStateLastEventID(sourceCluster, lastEvent.GetVersion(), replayEventID)
```

service/history/conflictResolverV2.go
yux0: removed

service/history/conflictResolverV2.go
yux0: remvoed

service/history/conflictResolverV2.go
yux0: updated

service/history/conflictResolverV2.go
yux0: updated

service/history/conflictResolverV2.go
yux0: removed

service/history/conflictResolverV2.go
yux0: removed

Issue #1709, opened 2019-04-16
Visibility: Support customized searchable fields


vancexu:
Cadence need to provide customized info that are searchable:
1. user may provide such info when they start workflow
2. user may update such info when workflow is running 
3. user use list API to search all workflows with certain info

Currently Visibility contains only following information that user can search with ListWorkflowExecutions API, 
```
	DomainID     
	WorkflowID    
	RunID       
	WorkflowType  
	StartTime    
	ExecutionTime 
	CloseTime    
	CloseStatus  
	HistoryLength 
```
Currently, Visibility also allow user can provide non-searchable info in visibility through Memo feature.

vancexu:
In addition, we need to support sorting on customized fields

PR #1825, opened 2019-05-09
Support custom search attributes on start workflow

vancexu:
This PR contains:
1. A new API `GetSearchAttributes`, which will be used by customers to learn what keys (and value type) they can use. Implemented through dynamic config.
2. Ability to support search attributes on start workflow. This is majorly about persist info in mutable state; pipe to ES; various validation on frontend


dmetzgar:
#1709

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -254,6 +254,15 @@ enum ArchivalStatus {
   ENABLED,
 }
 
+enum IndexedValueType {
+  STRING,
+  KEYWORD,
```
samarabbas: what is the difference between STRING and KEYWORD?

idl/github.com/uber/cadence/sqlblobs.thrift
```diff
@@ -120,6 +120,7 @@ struct WorkflowExecutionInfo {
   114: optional string clientImpl
   115: optional binary autoResetPoints
   116: optional string autoResetPointsEncoding
+  118: optional map<string, binary> searchAttributes
```
samarabbas: Maybe we can keep searchAttributes as a binary blob?

schema/cassandra/cadence/schema.cql
```diff
@@ -73,6 +73,7 @@ CREATE TYPE workflow_execution (
   last_event_task_id               bigint,
   auto_reset_points                blob, -- the resetting points for auto-reset feature
   auto_reset_points_encoding       text, -- encoding for auto_reset_points_data
+  search_attributes                map<text, blob>
```
samarabbas: No reason to keep this as a map.

service/frontend/workflowHandler.go
```diff
@@ -3225,3 +3267,201 @@ func (wh *WorkflowHandler) getArchivedHistory(
...
+		case gen.IndexedValueTypeInt:
+			converted[k] = gen.IndexedValueTypeInt
+		case gen.IndexedValueTypeDouble:
+			converted[k] = gen.IndexedValueTypeDouble
+		case gen.IndexedValueTypeBool:
+			converted[k] = gen.IndexedValueTypeBool
+		case gen.IndexedValueTypeDatetime:
+			converted[k] = gen.IndexedValueTypeDatetime
+		default:
+			wh.GetLogger().Fatal("unknown index value type", tag.Value(v))
```
samarabbas: Can we log an error instead of Fatal here?  We want to fail the call with InternalServiceError rather than crashing the host.

service/frontend/workflowHandler.go
```diff
@@ -3225,3 +3267,201 @@ func (wh *WorkflowHandler) getArchivedHistory(
 		Archived:      common.BoolPtr(true),
 	}, nil
 }
+
+func (wh *WorkflowHandler) convertIndexedKeys(keys map[string]interface{}) map[string]gen.IndexedValueType {
```
samarabbas: I'm not sure what this is converting? It seems to return the same key/value pair back.

service/frontend/workflowHandler.go
```diff
@@ -3225,3 +3267,201 @@ func (wh *WorkflowHandler) getArchivedHistory(
...
+		case gen.IndexedValueTypeDatetime:
+			converted[k] = gen.IndexedValueTypeDatetime
+		default:
+			wh.GetLogger().Fatal("unknown index value type", tag.Value(v))
+		}
+	}
+	return converted
+}
+
+func (wh *WorkflowHandler) validateSearchAttributes(input *gen.SearchAttributes, domain string) error {
```
samarabbas: I would put this in common helpers as later other services like history also needs access to this logic.

service/frontend/workflowHandler.go
```diff
@@ -3225,3 +3267,201 @@ func (wh *WorkflowHandler) getArchivedHistory(
...
+			Error("total size of search attributes exceed limit")
+		return fmt.Errorf("total size %d exceed limit", totalSize)
+	}
+
+	return nil
+}
+
+// validateListRequestForQuery validate that search attributes in listRequest query is legal,
+// and add prefix for custom keys
+func (wh *WorkflowHandler) validateListRequestForQuery(listRequest *gen.ListWorkflowExecutionsRequest) error {
```
samarabbas: Can you have a separate QueryValidator component instead of having separate functions for individual validation defined as functions.

service/worker/indexer/processor.go
```diff
@@ -234,10 +236,21 @@ func (p *indexProcessor) generateESDoc(msg *indexer.Message, keyToKafkaMsg strin
 	return doc
 }
 
+func (p *indexProcessor) decodeSearchAttrBinary(bytes []byte, key string) interface{} {
```
samarabbas: Can we rely on Thrift encode/decode rather than decoding payload ourself.

common/definition/indexedKeys.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	Memo = "Memo"
+)
+
+// Attr is prefix of custom search attributes
+const Attr = "Attr"
+
+// defaultIndexedKeys defines all searchable keys
+var defaultIndexedKeys = map[string]interface{}{
+	DomainID:      shared.IndexedValueTypeKeyword,
```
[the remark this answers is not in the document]
vancexu: make sense. updated

common/persistence/persistenceInterface.go
```diff
@@ -252,6 +253,7 @@ type (
 		BranchToken       []byte
 		CronSchedule      string
 		ExpirationSeconds int32
+		SearchAttributes  map[string][]byte
```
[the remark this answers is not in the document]
vancexu: keep as discussed

idl/github.com/uber/cadence/shared.thrift
vancexu: Keyword is can only be used as exact match when query, 
for example UUID "1234-567-XXX-XXX" will be matched when query "567" when it saved as STRING Type, but not be matched when saved as KEYWORD.

service/frontend/workflowHandler.go
vancexu: Updated

service/frontend/workflowHandler.go
vancexu: Go requires such type cast from interface{} to real type. I renamed it `convertIndexedKeysToThrift`

service/frontend/workflowHandler.go
vancexu: Updated

service/worker/indexer/processor.go
vancexu: keep as discussed

service/frontend/workflowHandler.go
vancexu: updated

vancexu:
I will add xdc integration test in separate PR

common/util.go
```diff
@@ -437,3 +441,151 @@ func GetSizeOfMapStringToByteArray(input map[string][]byte) int {
 	}
 	return res + golandMapReserverNumberOfBytes
 }
+
+// VisibilityQueryValidator for sql query validation
+type VisibilityQueryValidator struct {
```
samarabbas: can you move this to a different file?

PR #1826, opened 2019-05-10
Remove unused file and docker release instruction(moved to internal)

longquanzheng:
I moved the release instruction to our internal wiki


CONTRIBUTING.md
```diff
@@ -96,3 +96,25 @@ go test -v <path> -run <TestSuite> -testify.m <TestSpercificTaskName>
...
+```yaml
+  cadence:
+    image: ubercadence/server:THE.LATEST.VERSION 
+    ports
+```
+6. Create the tar.gz file and upload to release page
+```bash
+cd github.com/uber/cadence/docker
+tar -cvf docker.tar.gz *
+```
```
dmetzgar: This whole section should be in the internal docs. People contributing to Cadence server won't have permission to push to our Docker or add a release.

longquanzheng: Yes, that's true. It's also nice to have the release process in a single place rather than split into external and internal. I will move it.

PR #1828, opened 2019-05-10
Fix yarpc registration data race

wxing1292:


service/frontend/dcRedirectionHandler.go
```diff
@@ -23,6 +23,12 @@ package frontend
 import (
 	"context"
 
+	"github.com/uber/cadence/.gen/go/health/metaserver"
```
longquanzheng: fix import.

PR #1830, opened 2019-05-10
Not retry on reset bad request

longquanzheng:
If for any reason the reset task failed and we know it is not retryable we should skip the task to unblock transfer Queue.

That corruption usually means critical bug in our system. We should set up alert on the metric

### DIFF
--- a/service/history/transferQueueActiveProcessor.go
+++ b/service/history/transferQueueActiveProcessor.go
@@ -998,8 +998,8 @@ func (t *transferQueueActiveProcessorImpl) processResetWorkflow(task *persistenc
 			logger.Error("Auto-Reset workflow failed and not retryable. The reset point is corrupted.", tag.Error(err))
 			return nil
 		}
-		logger.Error("Auto-Reset workflow failed", tag.Error(err))
 		// log this error and retry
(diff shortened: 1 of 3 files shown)


service/history/transferQueueActiveProcessor.go
```diff
@@ -991,7 +991,15 @@ func (t *transferQueueActiveProcessorImpl) processResetWorkflow(task *persistenc
...
 	if err != nil {
+		if _, ok := err.(*workflow.BadRequestError); ok {
+			// This means the reset point is corrupted and not retry able.
+			// There must be a bug in our system that we must fix.(for example, history is not the same in active/passive)
+			t.metricsClient.IncCounter(metrics.TransferQueueProcessorScope, metrics.AutoResetPointCorruptionCounter)
+			logger.Error("Auto-Reset workflow failed and not retryable. The reset point is corrupted.", tag.Error(err))
+			return nil
+		}
 		logger.Error("Auto-Reset workflow failed", tag.Error(err))
+		// log this error and retry
```
shreyassrivatsan: nit: comment should be on line above

longquanzheng: Sure

PR #1831, opened 2019-05-11
Break passive transient decision into 2 batch of events for workflow execution reset

wxing1292:
Existing cross DC history replication logic & passive side events storage logic treat transient decision (1st batch decision task scheduled + started, 2nd batch decision task completed + other events) as one single batch of events to be replicated to passive side, this behavior is correct since passive side need to guarantee the transactionality. 

Above behavior however break the new workflow execution reset logic, since this functionality assumes that decision task completed event will always be the first event in an event batch.

This PR breaks the replicated single batch into 2 batches, same as the active side, for workflow execution reset.

Correctness is guaranteed by existing transient decision cross DC integration test.


wxing1292:
close in favor of #1839

PR #1832, opened 2019-05-13
Add instance ID to runtime metrics

shreyassrivatsan:
- Support for adding instance tag to runtime metrics.
- Currently not used in default cadence server

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -21,7 +21,8 @@
 package metrics
 
 const (
-	domain         = "domain"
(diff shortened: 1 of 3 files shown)


common/metrics/tags.go
```diff
@@ -31,38 +32,38 @@ type Tag interface {
 	Value() string
 }
 
-type domainTag struct {
+type _domainTag struct {
```
longquanzheng: We don't use underscore at all in our code base. I don't see this is a good start here to introduce new code style.

PR #1838, opened 2019-05-13
Fix bug: signal workflow will trigger cron workflow execution

yux0:
the signal workflow execution should not trigger cron workflow execution

### DIFF
--- a/host/signalworkflow_test.go
+++ b/host/signalworkflow_test.go
@@ -586,6 +586,66 @@ CheckHistoryLoopForSignalSent:
 	s.Equal("history-service", *signalEvent.WorkflowExecutionSignaledEventAttributes.Identity)
 }
 
+func (s *integrationSuite) TestSignalWorkflow_Cron_NoDecisionTaskCreated() {
+	id := "integration-signal-workflow-test-cron"
+	wt := "integration-signal-workflow-test-cron-type"
+	tl := "integration-signal-workflow-test-cron-tasklist"
+	identity := "worker1"
+	cronSpec := "@yearly"
+
+	workflowType := &workflow.WorkflowType{}
+	workflowType.Name = common.StringPtr(wt)
+
+	taskList := &workflow.TaskList{}
+	taskList.Name = common.StringPtr(tl)
+
+	// Start workflow execution
+	request := &workflow.StartWorkflowExecutionRequest{
+		RequestId:                           common.StringPtr(uuid.New()),
+		Domain:                              common.StringPtr(s.domainName),
+		WorkflowId:                          common.StringPtr(id),
+		WorkflowType:                        workflowType,
+		TaskList:                            taskList,
+		Input:                               nil,
+		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(100),
+		TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(1),
+		Identity:                            common.StringPtr(identity),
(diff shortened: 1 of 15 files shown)


host/signalworkflow_test.go
```diff
@@ -586,6 +586,66 @@ CheckHistoryLoopForSignalSent:
...
+		WorkflowExecution: &workflow.WorkflowExecution{
+			WorkflowId: common.StringPtr(id),
+			RunId:      common.StringPtr(*we.RunId),
+		},
+		SignalName: common.StringPtr(signalName),
+		Input:      signalInput,
+		Identity:   common.StringPtr(identity),
+	})
+	s.Nil(err)
+	resp, err := s.engine.GetWorkflowExecutionHistory(createContext(), &workflow.GetWorkflowExecutionHistoryRequest{
```
yycptt: I am not sure but will this create a race condition? If you get the history before the decision task for the signal is added to history, then the test can still pass.

yux0: Updated the integration test to check the signal event created but no decision task creation

host/signalworkflow_test.go
```diff
@@ -586,6 +586,71 @@ CheckHistoryLoopForSignalSent:
...
+		Domain: common.StringPtr(s.domainName),
+		Execution: &workflow.WorkflowExecution{
+			WorkflowId: common.StringPtr(id),
+			RunId:      common.StringPtr(*we.RunId),
+		},
+		MaximumPageSize: common.Int32Ptr(1000),
+		WaitForNewEvent: common.BoolPtr(false),
+	})
+	isSignalEventZCreated := false
```
yycptt: isSignalEventCreated

yux0: Deleted after the new test case

host/signalworkflow_test.go
yux0: Updated with the new test approach

host/signalworkflow_test.go
```diff
@@ -586,6 +586,121 @@ CheckHistoryLoopForSignalSent:
...
+	s.Nil(err)
+
+	// decider logic
+	activityScheduled := false
+	activityData := int32(1)
+	var decisionTaskDelay time.Duration
+	dtHandler := func(execution *workflow.WorkflowExecution, wt *workflow.WorkflowType,
+		previousStartedEventID, startedEventID int64, history *workflow.History) ([]byte, []*workflow.Decision, error) {
+		decisionTaskDelay = time.Now().Sub(now)
+		if !activityScheduled {
```
yycptt: I don't think you need anything related to activity. Just complete workflow execution directly in the dtHandler should be enough.

host/signalworkflow_test.go
```diff
@@ -586,6 +586,92 @@ CheckHistoryLoopForSignalSent:
...
+		return nil, []*workflow.Decision{{
+			DecisionType: common.DecisionTypePtr(workflow.DecisionTypeCompleteWorkflowExecution),
+			CompleteWorkflowExecutionDecisionAttributes: &workflow.CompleteWorkflowExecutionDecisionAttributes{
+				Result: []byte("Done."),
+			},
+		}}, nil
+	}
+
+	// activity handler
+	atHandler := func(execution *workflow.WorkflowExecution, activityType *workflow.ActivityType,
```
yycptt: This can also be deleted.

PR #1839, opened 2019-05-13
Handle transient decision completion with 2 history event batches.

wxing1292:
Existing cross DC history replication logic & passive side events storage logic treat transient decision completion (1st batch decision task scheduled + started, 2nd batch decision task completed + other events) as one single batch of events to be replicated to passive side, this behavior is correct since passive side need to guarantee the transactionality.

Above behavior however break the new workflow execution reset logic, since this functionality assumes that decision task completed event will always be the first event in an event batch.

This PR breaks the single replication message which contains 2 event batches to 2 replication message which each contains 1 event batch. 
Additional logic is added to ReplicateDecisionTaskStartedEvent to handle case if 1st batch is replicated but 2nd batch is not, basically set the decision attempt to 0 so passive side, when turning into active can properly timeout / fail existing decision.

Correctness is guaranteed by existing transient decision cross DC integration test.


samarabbas:
I'm still not certain why we need separate handling for transient decision.  If the decision fails why can't the state machine on passive also go into transient decision than it expects the next transaction to have multiple batches, similar to state machine on active side.

wxing1292:
> I'm still not certain why we need separate handling for transient decision. If the decision fails why can't the state machine on passive also go into transient decision than it expects the next transaction to have multiple batches, similar to state machine on active side.

transient decision has 2 parts:
1. active side in transient decision, this is correctly handled by passive side
2. active side try to end the transient decision, with decision task scheduled + started as first batch,  decision task completed + decisions + others in the second batch.

the 2 above breaks the existing model that one batch of events is one transaction.
if say first batch is replicated to passive while second batch is not, and passive turns into active, history will be corrupted.

PR #1849, opened 2019-05-14
Start emitting build info metrics

### LINKED ISSUE #1840: No build version info is exposed via /metrics for prometheus
/metrics should be exposing build version information.

shreyassrivatsan:
Pass in build info/version information through LDFLAGS

Fixes #1840


common/metrics/runtime.go
```diff
@@ -50,6 +86,20 @@ func NewRuntimeMetricsReporter(scope tally.Scope, reportInterval time.Duration,
 		lastNumGC:      memstats.NumGC,
 		quit:           make(chan struct{}),
 	}
+	rReporter.buildInfoScope = scope.Tagged(
+		map[string]string{
+			"revision":      Revision,
```
samarabbas: How about we define these tags under other tag definitions under defs.go

shreyassrivatsan: Done. Defined them in tags.go along with the other tags.

PR #1856, opened 2019-05-15
Create workflow execution request should specify workflow state & clo…

wxing1292:
…se status

This is for supporting zombie workflow creation as well as creating workflow with cron schedule.
Cron workflow, when create, should be "created" state, until first decision task is scheduled.


common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -700,7 +720,8 @@ func (s *ExecutionManagerSuite) TestUpdateWorkflow() {
 	s.Equal(int32(20), info1.WorkflowTimeout)
 	s.Equal(int32(13), info1.DecisionTimeoutValue)
 	s.Equal([]byte(nil), info1.ExecutionContext)
-	s.Equal(p.WorkflowStateCreated, info1.State)
+	s.Equal(p.WorkflowStateRunning, info1.State)
+	s.Equal(p.WorkflowCloseStatusNone, info0.CloseStatus)
```
vancexu: should be info1. similar for the following copy 

(please leave other error you found in this unchanged, I have a pending PR address those)

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -90,6 +90,8 @@ func (s *ExecutionManagerSuite) TestCreateWorkflowExecutionBrandNew() {
 		WorkflowTypeName:     workflowType,
 		WorkflowTimeout:      workflowTimeout,
 		DecisionTimeoutValue: decisionTimeout,
+		State:                p.WorkflowStateRunning,
```
venkat1109: can you make sure we have a test that passes all combinations of {state,closeStatus} to each of {createWorkflow, updateWorkflow} APIs and makes sure GetWorkflowExecution() returns the expected value ? In the past, we have had atleast two bugs due to the lack of such a test

wxing1292: sure

PR #1860, opened 2019-05-15
Remove all defers in archival workflow code

andrewjdawson2016:


service/worker/archiver/archiver.go
```diff
@@ -101,8 +101,6 @@ func (a *archiver) Finished() []uint64 {
 
 func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metrics.Client, request ArchiveRequest) {
 	sw := metricsClient.StartTimer(metrics.ArchiverScope, metrics.ArchiverHandleRequestLatency)
-	defer sw.Stop()
```
shreyassrivatsan: What's the rationale of removing this defer?  The effective code is the same here.. If someone adds a new path they will very likely miss this..

PR #1861, opened 2019-05-16
Rate limit archive request 

yycptt:
https://github.com/uber/cadence/issues/1854


service/worker/archiver/client.go
```diff
@@ -66,18 +70,28 @@ func NewClient(
...
+		rateLimiter:   tokenbucket.New(requestRPS, clock.NewRealTimeSource()),
 	}
 }
 
 // Archive starts an archival task
 func (c *client) Archive(request *ArchiveRequest) error {
 	c.metricsClient.IncCounter(metrics.ArchiverClientScope, metrics.CadenceRequests)
+
+	if ok, _ := c.rateLimiter.TryConsume(1); !ok {
+		errMsg := "Too many archive requests to archival workflow"
```
venkat1109: declare the error message once at the top of the file and reuse it across calls

service/worker/archiver/client.go
```diff
@@ -57,6 +60,7 @@ type (
 		logger        log.Logger
 		cadenceClient cclient.Client
 		numWorkflows  dynamicconfig.IntPropertyFn
+		rateLimiter   tokenbucket.TokenBucket
```
venkat1109: You want this limit to be a dynamic config func() rather than a static integer. That way, you can pick a change to the ratelimit without having to restart the process. Take a look at this - which does something similar:
https://github.com/uber/cadence/blob/master/service/matching/ratelimiter.go

service/worker/archiver/client.go
yycptt: Done.

service/worker/archiver/client.go
yycptt: Will do this in another PR.

PR #1863, opened 2019-05-16
Initial work on s3 blobstore

jontro:
I've created an initial implementation of the blobstore interface for s3.

Some notes 
- Credentials are assumed to be configured according to the section in "Configuring Credentials" https://docs.aws.amazon.com/sdk-for-go/api/
- All buckets have to be in the same region, it would maybe make sense to be able to configure this per bucket

Is there an easy way to test archival of a workflow in a development environment?


common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	"github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common/backoff"
+	"github.com/uber/cadence/common/blobstore"
+	"github.com/uber/cadence/common/blobstore/blob"
+	"go.uber.org/multierr"
+	"io/ioutil"
+	"math"
+	"net/url"
```
andrewjdawson2016: Lets order imports such that all go lib packages occur first then an enter then all other packages.

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+			}
+
+		}
+		return err
+	}
+	return nil
+}
+
+func s3download(ctx context.Context, s3api s3iface.S3API, bucket string, key string) ([]byte, error) {
+
```
andrewjdawson2016: remove extra enter

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+				return blobstore.ErrBucketNotExists
+			}
+
+		}
+		return err
+	}
+	return nil
+}
+
+func s3download(ctx context.Context, s3api s3iface.S3API, bucket string, key string) ([]byte, error) {
```
andrewjdawson2016: For all of these methods which take a context can we include a call to something to ensure there is a context timeout something like 

```
func (t *terrablobBlobstore) ensureContextTimeout(ctx context.Context) (context.Context, context.CancelFunc) {
	if _, ok := ctx.Deadline(); ok {
		return ctx, func() {}
	}
	return context.WithTimeout(ctx, defaultBlobstoreTimeout)
}
````

File: common/blobstore/s3store/client.go [hunk 4]
```diff
@@ -0,0 +1,311 @@
...
+	}()
+
+	body, err := ioutil.ReadAll(result.Body)
+
+	return body, err
+}
+
+func s3gettags(ctx context.Context, s3api s3iface.S3API, bucket string, key string) (map[string]string, error) {
+
+	result, err := s3api.GetObjectTaggingWithContext(ctx, &s3.GetObjectTaggingInput{
```
andrewjdawson2016: Do you need to close anything on this result resource like you do for download?

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+
+	tags, err := s3gettags(ctx, c.s3cli, bucket, key.String())
+	if err != nil {
+		return nil, err
+	}
+	return tags, nil
+}
+
+func (c *client) Exists(ctx context.Context, bucket string, key blob.Key) (bool, error) {
+
```
andrewjdawson2016: Remove extra enter

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+		Bucket: aws.String(bucket),
+		Prefix: aws.String(prefix),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchBucket {
+			return nil, blobstore.ErrBucketNotExists
+		}
+		return nil, ErrListFiles
+	}
+	var keys = make([]blob.Key, len(results.Contents))
```
andrewjdawson2016: Since you actually know the number of items that will be in this slice you can declare it with both length and capacity. This is more efficient (e.g. make([]int, 5, 5)

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+	lifecycleResults, err := c.s3cli.GetBucketLifecycleWithContext(ctx, &s3.GetBucketLifecycleInput{
+		Bucket: aws.String(bucket),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchBucket {
+			return nil, blobstore.ErrBucketNotExists
+		}
+		return nil, err
+	}
+	var retentionDays = math.MaxInt64
```
andrewjdawson2016: We use the value of 0 to indicate unbounded retention

```
	if resp.Configuration.ArchivalRetentionPeriodInDays != nil {
		formatStr = formatStr + "ArchivalRetentionInDays: %v\n"
		archivalRetentionDays := resp.Configuration.GetArchivalRetentionPeriodInDays()
		if archivalRetentionDays == 0 {
			descValues = append(descValues, "unlimited")
		} else {
			descValues = append(descValues, archivalRetentionDays)
		}
	}
```

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchBucket {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (c *client) IsRetryableError(err error) bool {
```
andrewjdawson2016: In the case of filestore I return false for all error types because it just for testing locally. But when running in prod we will want S3 to have a correct policy around retryable vs non-retryable errors.

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+	return false
+}
+
+func (c *client) GetRetryPolicy() backoff.RetryPolicy {
+	policy := backoff.NewExponentialRetryPolicy(0)
```
andrewjdawson2016: Same comment with this retry policy. We want a policy that enables for retrying.

common/blobstore/s3store/config.go
```diff
@@ -0,0 +1,63 @@
...
+// THE SOFTWARE.
+
+package s3store
+
+import (
+	"errors"
+)
+
+type (
+	// Config describes the configuration needed to construct a blobstore client backed by file system
```
andrewjdawson2016: "backed by S3"

common/blobstore/s3store/config.go
```diff
@@ -0,0 +1,63 @@
...
+	Config struct {
+		Region        string         `yaml:"region"`
+		DefaultBucket BucketConfig   `yaml:"defaultBucket"`
```
andrewjdawson2016: The reason filestore has DefaultBucket and CustomBucket is on start up it actually creates these buckets. But in the case of S3 blobstore we won't be doing that. There should not be any DefaultBucket or CustomBuckets as part of this config. S3 blobstore should just know how to take a bucket name in its public APIs and attempt to operate on that bucket.

common/blobstore/s3store/config.go
```diff
@@ -0,0 +1,63 @@
...
+	BucketConfig struct {
+		Name string `yaml:"name"`
+	}
+)
+
+// Validate validates config
+func (c *Config) Validate() error {
+	validateBucketConfig := func(b BucketConfig) error {
+		// s3 buckets must be between 3 and 63 charachters long
+		if len(b.Name) < 3 {
```
andrewjdawson2016: Can you also check that the name is not longer than 63?

common/service/config/config.go
```diff
@@ -22,6 +22,7 @@ package config
 
 import (
 	"encoding/json"
+	"github.com/uber/cadence/common/blobstore/s3store"
```
andrewjdawson2016: Same comment as above regarding import order

common/blobstore/s3store/config.go
jontro: What do you think about the use case of having different buckets in different regions? Or is one sufficient? Doing this lookup on demand is quite expensive but could of course be cached if so.

common/blobstore/s3store/client.go
jontro: The aws sdk has built-in retry logic, should we disable that one and move it out to this? https://github.com/aws/aws-sdk-go/blob/master/aws/client/default_retryer.go

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+			if aerr.Code() == s3.ErrCodeNoSuchKey {
+				return nil, ErrReadFile
+			}
+
+		}
+		return nil, err
+
+	}
+
+	defer func() {
```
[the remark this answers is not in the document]
jontro: Should we put it before the error handling? The examples I've loked at puts it after (https://docs.aws.amazon.com/sdk-for-go/api/ search for defer)

common/blobstore/s3store/client.go
jontro: AFAIK there is no need to do this for GetObjectTagging as it does not involve any streams

common/blobstore/s3store/client.go
jontro: Fixed with the caveat that the bucket can have multiple retention rules. I also checked that the rules are enabled. It'll return the largest expiration it can find now

common/blobstore/s3store/client.go
jontro: See comment above

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+func (c *client) Download(ctx context.Context, bucket string, key blob.Key) (*blob.Blob, error) {
+
+	b, err := s3download(ctx, c.s3cli, bucket, key.String())
+	if err != nil {
+		if err == ErrReadFile {
+			return nil, blobstore.ErrBlobNotExists
+		}
+		return nil, err
+	}
+	tags, err := s3gettags(ctx, c.s3cli, bucket, key.String())
```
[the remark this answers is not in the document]
jontro: Unfortunately I have not found a way to do this. We get the tag count in a field and can do a an optimization if the tag count is 0

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,311 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return blob.NewBlob(b, tags), nil
+}
+
+func (c *client) GetTags(ctx context.Context, bucket string, key blob.Key) (map[string]string, error) {
+
+	tags, err := s3gettags(ctx, c.s3cli, bucket, key.String())
```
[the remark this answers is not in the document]
jontro: See comment above

common/blobstore/s3store/client.go
jontro: Should I add this to all implemented methods in the beginning?

common/blobstore/s3store/client.go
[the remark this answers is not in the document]
shreyassrivatsan: The defer should only run if the result exists, i.e `err == nil`.. So this seems to be the right place for it.. 

It should be when the object the defer depends on becomes available.. In this case right after error handling if any..

common/blobstore/s3store/config_test.go
```diff
@@ -0,0 +1,126 @@
...
+						Name: "test-custom-bucket-name",
+					},
+				},
+			},
+			isValid: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		if tc.isValid {
```
shreyassrivatsan: nit: nice to use subtests.. so 
```
t.Run(testname, func (t *testing.T) {
    if tc.IsValue ...
})
```

common/blobstore/s3store/config.go
jontro: Removed it for now

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil, err
+	}
+
+	s3cli := s3.New(sess)
+	return &client{
+		s3cli: s3cli,
+	}, nil
+}
+
+func (c *client) Upload(ctx context.Context, bucket string, key blob.Key, blob *blob.Blob) error {
```
andrewjdawson2016: Can we ensure these contexts have a timeout?

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,478 @@
...
+	client := s.constructClient()
+	b := blob.NewBlob([]byte("body version 1"), map[string]string{})
+	key, err := blob.NewKeyFromString("blob.blob")
+	s.NoError(err)
+	s.NoError(client.Upload(context.Background(), defaultBucketName, key, b))
+	downloadBlob, err := client.Download(context.Background(), defaultBucketName, key)
+	s.NoError(err)
+	s.NotNil(downloadBlob)
+	s.assertBlobEquals(map[string]string{}, "body version 1", downloadBlob)
+	_, err = client.Delete(context.Background(), defaultBucketName, key)
```
andrewjdawson2016: If you are going to call Delete in this test then lets check the blob does not exist after calling delete. Another test that is needed is to Upload/Download/Upload with same key and ensure existing blob gets replaced.

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,478 @@
...
+
+	b := blob.NewBlob([]byte("blob body"), map[string]string{"tagKey": "tagValue"})
+	key, err := blob.NewKeyFromString("blob.blob")
+	s.NoError(err)
+	s.Equal(blobstore.ErrBucketNotExists, client.Upload(context.Background(), "bucket-not-exists", key, b))
+}
+
+/*
+Not sure how an error would fail
+func (s *ClientSuite) TestUpload_Fail_ErrorOnWrite() {
```
andrewjdawson2016: Some of these tests only make sense in the context of filestore. No need to leave these commented out tests in this file.

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,478 @@
...
+	key, err := blob.NewKeyFromString("blobname.ext")
+	s.NoError(err)
+	b, err := client.Download(context.Background(), defaultBucketName, key)
+	s.Equal(blobstore.ErrBlobNotExists, err)
+	s.Nil(b)
+}
+
+/*
+There is no invalid blob format
+func (s *ClientSuite) TestDownload_Fail_BlobFormatInvalid() {
```
andrewjdawson2016: Same comment here about commented out tests

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,478 @@
...
+	//	s.NoError(writeFile(filepath.Join(dir, defaultBucketName, key.String()), []byte("invalid")))
+
+	b, err := client.Download(context.Background(), defaultBucketName, key)
+	s.Equal(blobstore.ErrBlobDeserialization, err)
+	s.Nil(b)
+}
+*/
+
+func (s *ClientSuite) TestUploadDownload_Success_CustomBucket() {
+
```
andrewjdawson2016: Remove extra enter

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,478 @@
...
+	client := s.constructClient()
+	b := blob.NewBlob([]byte("blob body"), map[string]string{})
+	key, err := blob.NewKeyFromString("blob.blob")
+	s.NoError(err)
+	customBucketName := fmt.Sprintf("%v-%v", customBucketNamePrefix, 3)
+	s.NoError(client.Upload(context.Background(), customBucketName, key, b))
+	downloadBlob, err := client.Download(context.Background(), customBucketName, key)
+	s.NoError(err)
+	s.NotNil(downloadBlob)
+	s.assertBlobEquals(map[string]string{}, "blob body", downloadBlob)
```
andrewjdawson2016: Can we include some tests with non-empty tags?

common/blobstore/s3store/config_test.go
andrewjdawson2016: I am okay with the existing pattern. This is how we write other config tests throughout code base.

common/blobstore/s3store/README.md
```diff
@@ -0,0 +1,22 @@
+# Amazon S3 blobstore
+
+## Using localstack for local development
+1. Install awscli from [here][https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-install.html]
+2. Install localstack from [here][https://github.com/localstack/localstack#installing]
+3. Launch localstack with `SERVICES=s3 localstack start`
+4. Create a bucket using `aws --endpoint-url=http://localhost:4572 s3 s3://cadence-development` 
+5. Configure archival with the following configuration
+```
+archival:
```
andrewjdawson2016: Can we also include this config in development.yaml commented out?

cmd/server/server.go
```diff
@@ -184,7 +185,11 @@ func (s *server) startService() common.Daemon {
 	params.PublicClient = workflowserviceclient.New(dispatcher.ClientConfig(common.FrontendServiceName))
 
 	if params.ClusterMetadata.ArchivalConfig().ConfiguredForArchival() {
-		params.BlobstoreClient, err = filestore.NewClient(&s.cfg.Archival.Filestore)
+		if s.cfg.Archival.Filestore != nil {
```
andrewjdawson2016: Can we return an error if both configs are specified.

common/blobstore/s3store/config_test.go
shreyassrivatsan: Sure, but we should move to subtests as they are more grokkable on failures - subtests are newer than when the cadence code base started so were not used initially is my guess.. Also, they allow you to parallelize tests to make them faster if we want to make them faster, rather than running them serial..

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (c *client) Exists(ctx context.Context, bucket string, key blob.Key) (bool, error) {
+	_, err := c.s3cli.HeadObjectWithContext(ctx, &s3.HeadObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key.String()),
+	})
+	if err != nil {
+		aerr, ok := err.(awserr.Error)
+		if ok && (aerr.Code() == s3.ErrCodeNoSuchKey || aerr.Code() == "NotFound") {
```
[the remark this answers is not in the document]
jontro: The s3 api currently returns "NotFound" which have no string literal in the api. The s3.ErrCodeNoSuchKey can be removed or factored into a generic helper method. See https://github.com/aws/aws-sdk-go/issues/1208

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,275 @@
...
+	// ErrConstructKey could not construct key
+	ErrConstructKey = &shared.BadRequestError{Message: "could not construct key"}
+)
+
+type client struct {
+	s3cli s3iface.S3API
+}
+
+// NewClient returns a new Client backed by S3
+func NewClient(cfg *Config) (blobstore.Client, error) {
```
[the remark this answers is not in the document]
jontro: What do you think about adding a function like NewS3cliFromConfig that lives in client.go to keep the specifics out from server.go?

common/blobstore/s3store/client.go
jontro: Fixed

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,275 @@
...
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
+
+	s3Config := &aws.Config{
+		Endpoint:         cfg.Endpoint,
+		Region:           aws.String(cfg.Region),
+		S3ForcePathStyle: aws.Bool(cfg.S3ForcePathStyle),
+	}
+	sess, err := session.NewSession(s3Config)
```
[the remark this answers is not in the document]
jontro: The session should not be doing anything that needs to be closed. Everything is started and cleaned up after each method call afaik

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (c *client) GetRetryPolicy() backoff.RetryPolicy {
+	policy := backoff.NewExponentialRetryPolicy(0)
+	policy.SetMaximumAttempts(1)
+	return policy
+}
+
+func s3gettags(ctx context.Context, s3api s3iface.S3API, bucket string, key string) (map[string]string, error) {
+	result, err := s3api.GetObjectTaggingWithContext(ctx, &s3.GetObjectTaggingInput{
```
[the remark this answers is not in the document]
jontro: It's just a standard call, it will not open any streams

common/blobstore/s3store/client_test.go
jontro: Added this, though maybe this whole test should be removed as a custom bucket name isn't anything special with the s3store any more

jontro:
Hopefully I have fixed all feedback and left comments otherwise. One thing, I can't see why the tests are failing, locally the unit tests I've touched passes.

Gopkg.toml
```diff
@@ -150,3 +150,8 @@ ignored = ["github.com/uber/cadence/.gen"]
 [[constraint]]
   name = "github.com/valyala/fastjson"
   version = "1.4.1"
+
```
andrewjdawson2016: Extra enter.

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,321 @@
...
+	ctx, cancel := c.ensureContextTimeout(ctx)
+	defer cancel()
+	tags, err := getTags(ctx, c.s3cli, bucket, key.String())
+	if err != nil {
+		return nil, err
+	}
+	return tags, nil
+}
+
+// Exists returns false when the bucket does not exist or the bucket and key does not exist
```
andrewjdawson2016: Filestore currently returns error if bucket does not exist (e.g. bucket not existing is an error but key not existing is just false, nil) I think both APIs are reasonable but I want them to be the same. Lets change this to return an error if bucket does not exist.

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,321 @@
...
+func (c *client) Exists(ctx context.Context, bucket string, key blob.Key) (bool, error) {
+	ctx, cancel := c.ensureContextTimeout(ctx)
+	defer cancel()
+	_, err := c.s3cli.HeadObjectWithContext(ctx, &s3.HeadObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key.String()),
+	})
+	if err != nil {
+		aerr, ok := err.(awserr.Error)
+		if ok && (aerr.Code() == "NotFound") {
```
andrewjdawson2016: Why do we need this hardcoded string? Is there no error code or constant in s3 lib we can check against?

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,321 @@
...
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchBucket {
+			return nil, blobstore.ErrBucketNotExists
+		}
+		return nil, err
+	}
+	retentionDays := 0
+	for _, v := range lifecycleResults.Rules {
+		if *v.Status == "Enabled" && retentionDays < int(*v.Expiration.Days) {
```
andrewjdawson2016: Any S3 library constant we can use instead of this hardcoded string?

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,321 @@
...
+		Bucket: aws.String(bucket),
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == s3.ErrCodeNoSuchBucket {
+			return nil, blobstore.ErrBucketNotExists
+		}
+		return nil, err
+	}
+	retentionDays := 0
+	for _, v := range lifecycleResults.Rules {
```
andrewjdawson2016: Are we iterating through these rules in chronological order? If so will retention days always be correct at the end of this loop? The way I am reading this code is we select the highest Expiration.Days that has ever been seen even if that is higher than the current Expiration.Days? Maybe I am just misunderstanding these APIs?

common/blobstore/s3store/client.go
```diff
@@ -0,0 +1,321 @@
...
+		if aerr, ok := err.(awserr.Error); ok && (aerr.Code() == s3.ErrCodeNoSuchBucket || aerr.Code() == "NotFound") {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+func (c *client) IsRetryableError(err error) bool {
+	return c.isStatusCodeRetryable(err) || request.IsErrorRetryable(err) || request.IsErrorThrottle(err)
```
andrewjdawson2016: Nice :)

common/blobstore/s3store/client.go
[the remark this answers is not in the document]
andrewjdawson2016: Okay :)

common/blobstore/s3store/client_test.go
```diff
@@ -0,0 +1,355 @@
...
+	downloadBlob, err := client.Download(context.Background(), defaultBucketName, key)
+	s.NoError(err)
+	s.NotNil(downloadBlob)
+	s.assertBlobEquals(map[string]string{}, "body version 1", downloadBlob)
+	b = blob.NewBlob([]byte("body version 2"), map[string]string{"key": "value"})
+	s.NoError(client.Upload(context.Background(), defaultBucketName, key, b))
+	downloadBlob, err = client.Download(context.Background(), defaultBucketName, key)
+	s.NoError(err)
+	s.NotNil(downloadBlob)
+	s.assertBlobEquals(map[string]string{"key": "value"}, "body version 2", downloadBlob)
```
andrewjdawson2016: Great!

common/blobstore/s3store/client_test.go
andrewjdawson2016: Yes agreed. Let's remove.

common/blobstore/s3store/client.go
jontro: Problem is it would be more inefficient. I.e. I would need to call a different api method. The same error is returned from the api whether or not the bucket exists. What do you think?

common/blobstore/s3store/client.go
jontro: Fixed!

Gopkg.toml
jontro: Fixed, somehow merging with master this automatically. Good find!

Issue #1800, opened 2019-05-07
3+DC - Ability To Create / Update Workflow With Zombie Status 


wxing1292:
The ability to create workflow with zombie status allows replication stack to backfill finished workflow execution sent by remote, while not affecting local running workflow

PR #1866, opened 2019-05-16
Adding support of zombie workflow in persistence layer

wxing1292:
* Enable creation of zombie workflow
* Enable update non current workflow turning it into zombie state

solve #1800

### DIFF
--- a/common/persistence/cassandra/cassandraPersistence.go
+++ b/common/persistence/cassandra/cassandraPersistence.go
@@ -1350,7 +1350,7 @@ func (d *cassandraPersistence) CreateWorkflowExecutionWithinBatch(request *p.Int
 		// 2. workflow close status being WorkflowCloseStatusNone
 		if request.State != p.WorkflowStateZombie || request.CloseStatus != p.WorkflowCloseStatusNone {
 			return &workflow.InternalServiceError{Message: fmt.Sprintf(
-				"Cannot create zombie workflow with state: %v, close sataus: %v",
+				"Cannot create zombie workflow with state: %v, close status: %v",
 				request.State,
 				request.CloseStatus,
 			)}
--- a/common/persistence/cassandra/cassandraPersistence.go
+++ b/common/persistence/cassandra/cassandraPersistence.go
@@ -1106,6 +1106,14 @@ func (d *cassandraPersistence) CreateWorkflowExecution(request *p.InternalCreate
 	cqlNowTimestamp := p.UnixNanoToDBTimestamp(time.Now().UnixNano())
 	batch := d.session.NewBatch(gocql.LoggedBatch)
 
+	if request.CreateWorkflowMode == p.CreateWorkflowModeZombie {
+		if err := d.assertNotCurrentExecution(
+			request.DomainID,
+			request.Execution.GetWorkflowId(),
+			request.Execution.GetRunId()); err != nil {
+			return nil, err
+		}
+	}
 	if err := d.CreateWorkflowExecutionWithinBatch(request, batch, cqlNowTimestamp); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


common/persistence/sql/sqlExecutionManager.go
```diff
@@ -609,26 +623,37 @@ func (m *sqlExecutionManager) updateWorkflowExecutionTx(tx sqldb.Tx, request *p.
...
-			runID,
-			runID,
-			executionInfo.CreateRequestID,
-			executionInfo.State,
-			executionInfo.CloseStatus,
-			startVersion,
-			lastWriteVersion); err != nil {
+		switch executionInfo.State {
+		case p.WorkflowStateZombie:
+			if err := assertNotCurrentExecution(tx, shardID, domainID, workflowID, runID); err != nil {
```
yux0: Can this NotCurrentExecution check move to executionStore so that we don't need to have them in both cass and sql

wxing1292: sql and cassandra side use different ways to check whether the workflow to be updated is the current workflow

sql actually have the current record, in persistenc logic (while doing create / update), while cassandra does not.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1815,31 +1834,45 @@ func (d *cassandraPersistence) UpdateWorkflowExecution(request *p.InternalUpdate
 		d.createTimerTasks(batch, startReq.TimerTasks, startReq.DomainID, startReq.Execution.GetWorkflowId(),
 			startReq.Execution.GetRunId(), cqlNowTimestamp)
 	} else {
-		startVersion := common.EmptyVersion
-		lastWriteVersion := common.EmptyVersion
-		if request.ReplicationState != nil {
-			startVersion = request.ReplicationState.StartVersion
-			lastWriteVersion = request.ReplicationState.LastWriteVersion
+		switch executionInfo.State {
+		case p.WorkflowStateZombie:
```
yux0: Why this only do a sanity check?

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -95,7 +96,14 @@ func (m *sqlExecutionManager) createWorkflowExecutionTx(tx sqldb.Tx, request *p.
 
 	lastWriteVersion := common.EmptyVersion
 
-	if row != nil {
+	// current workflow record check
+	if row == nil {
```
yux0: Should this proceed if the row is nil and the create workflow mode is not zombie

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1576,6 +1601,14 @@ func createOrUpdateCurrentExecution(
 				Message: fmt.Sprintf("CreateWorkflowExecution operation failed. Failed to insert into current_executions table. Error: %v", err),
 			}
 		}
+	case p.CreateWorkflowModeZombie:
+		if request.State != p.WorkflowStateZombie || request.CloseStatus != p.WorkflowCloseStatusNone {
+			return &workflow.InternalServiceError{Message: fmt.Sprintf(
+				"Cannot create zombie workflow with state: %v, close sataus: %v",
```
yux0: *status

common/persistence/cassandra/cassandraPersistence.go
wxing1292: i do not think there is a conditional get operation in cassandra
we need to have this check here, at least for the initial testing for 3+DC
for prod run, we can decide later

common/persistence/sql/sqlExecutionManager.go
wxing1292: good catch

PR #1870, opened 2019-05-17
Refactor DC redirection handler to be more generic

wxing1292:
* Deprecate DCRedirectionPolicyForwarding
* Add new DCRedirectionPolicySelectedAPIsForwarding, which will forward the following APIs to active cluster is domain is whitelisted:
	1. StartWorkflowExecution
	2. SignalWithStartWorkflowExecution
	3. SignalWorkflowExecution
	4. RequestCancelWorkflowExecution
	5. TerminateWorkflowExecution
* Add corredponding UT


service/frontend/dcRedirectionPolicy.go
```diff
@@ -33,49 +34,58 @@ const (
...
+	SelectedAPIsForwardingRedirectionPolicy struct {
+		currentClusterName string
+		config             *Config
+		domainCache        cache.DomainCache
 	}
 )
 
+// selectedAPIsForwardingRedirectionPolicyWhitelistedAPIs contains a list of APIs which can be redirected
+var selectedAPIsForwardingRedirectionPolicyWhitelistedAPIs = map[string]struct{}{
+	"StartWorkflowExecution":           {},
```
shreyassrivatsan: Using API names like this seems a little error prone.

service/frontend/dcRedirectionHandler.go
```diff
@@ -142,16 +152,22 @@ func (handler *DCRedirectionHandlerImpl) DescribeTaskList(
 	request *shared.DescribeTaskListRequest,
 ) (*shared.DescribeTaskListResponse, error) {
 
-	targetDC, err := handler.redirectionPolicy.GetTargetDataCenterByName(request.GetDomain())
-	if err != nil {
-		return nil, err
-	}
+	var apiName = "DescribeTaskList"
```
shreyassrivatsan: Is there a way to get the name of function we are in, so that we dont have to spell this out?

wxing1292:
> Looks good.. This is much neater now..
> 
> I am not a big fan of so much duplicative code.. Is there a way to implement this using yarpc middleware? https://godoc.org/go.uber.org/yarpc/api/middleware#ApplyOnewayInbound It maybe a larger refactor but something for the future..

i am going to land this PR now since customer is asking for it

Issue #1846, opened 2019-05-13
3+DC - Persistence of VersionHistory & BranchToken


wxing1292:

yux0:
#1872 merged

PR #1872, opened 2019-05-17
add version histories object in database

yux0:
This is 3DC related work to persist version histories data in the database. Future work to wire it up in mutable state will be in a separate PR.


common/persistence/dataInterfaces.go
```diff
@@ -811,6 +831,7 @@ type (
 		CurrReplicationTasks []Task
 		CurrTransferTasks    []Task
 		CurrTimerTasks       []Task
+		VersionHistories     *workflow.VersionHistories
```
wxing1292: there should be 2 version histories
one for current workflow, one for new workflow after reset

common/persistence/persistenceInterface.go
```diff
@@ -268,6 +270,7 @@ type (
 		ReplicationState         *ReplicationState
 		BufferedEvents           []*DataBlob
 		BufferedReplicationTasks map[int64]*InternalBufferedReplicationTask
+		//VersionHistories         *DataBlob
```
wxing1292: uncomment this one?

service/history/mutableStateBuilder.go
```diff
@@ -80,6 +80,7 @@ type (
 		deleteBufferedReplicationEvent *int64
 
 		executionInfo    *persistence.WorkflowExecutionInfo // Workflow mutable state info.
+		versionHistories *workflow.VersionHistories
```
wxing1292: plz add a new function which initialize ms builder with version history

common/persistence/persistenceInterface.go
yux0: removed

service/history/mutableStateBuilder.go
yux0: Done

common/persistence/dataInterfaces.go
yux0: This one is redundant. The VersionHistories current locates in the execution info which this request already has current one and an insert one.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -447,7 +447,10 @@ workflow_state = ? ` +
 	templateUpdateWorkflowExecutionConditionSuffix = ` IF next_event_id = ?`
 
 	templateUpdateWorkflowExecutionQuery = `UPDATE executions ` +
```
wxing1292: let us create a dedicated query just for version history.
for forward / backward compatibility, create an issue.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -457,7 +460,11 @@ workflow_state = ? ` +
 		`and task_id = ? `
 
 	templateUpdateWorkflowExecutionWithReplicationQuery = `UPDATE executions ` +
```
wxing1292: plz see above

dmetzgar:
#1846

common/persistence/dataInterfaces.go
```diff
@@ -549,6 +566,7 @@ type (
 		ReplicationState         *ReplicationState
 		BufferedEvents           []*workflow.HistoryEvent
 		BufferedReplicationTasks map[int64]*BufferedReplicationTask
+		VersionHistories         VersionHistories
```
wxing1292: make this a pointer, so we can use a null value for possible backward / forward compatibility issue

common/persistence/cassandra/cassandraPersistence.go
wxing1292: before actual update, we should do an assertion that at most one (replication state / version history) is set (for global domain workflow)

for local domain workflow, let us set the next event ID to empty event ID, current branch token to be empty.

common/persistence/dataInterfaces.go
```diff
@@ -537,6 +537,23 @@ type (
...
+	}
+
+	// VersionHistory provides operations on version history
+	VersionHistory struct {
+		BranchToken []byte
+		History     []VersionHistoryItem
+	}
+
+	// VersionHistories contains a set of VersionHistory
+	VersionHistories struct {
```
wxing1292: we also need to store the index of current branch / version history

common/persistence/versionHistory.go
```diff
@@ -23,83 +23,70 @@ package persistence
 import (
 	"fmt"
 	"github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common"
 	"reflect"
 )
 
-type (
```
wxing1292: i think it is ok to leave these definitions here
data interface file is too large

common/persistence/dataInterfaces.go
yux0: Done

common/persistence/dataInterfaces.go
yux0: Done

common/persistence/versionHistory.go
yux0: I will leave the structs here as it will be easier for refactor.

common/persistence/cassandra/cassandraPersistence.go
yux0: Done

common/persistence/cassandra/cassandraPersistence.go
yux0: Done

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1598,6 +1673,12 @@ func (d *cassandraPersistence) GetWorkflowExecution(request *p.GetWorkflowExecut
 	}
 	state.BufferedReplicationTasks = bufferedReplicationTasks
 	state.VersionHistories = p.NewDataBlob(result["version_histories"].([]byte), common.EncodingType(result["version_histories_encoding"].(string)))
+	if state.VersionHistories != nil && state.ReplicationState != nil {
```
wxing1292: i think we should do this check on create and update

common/persistence/executionStore.go
```diff
@@ -658,14 +702,49 @@ func (m *executionManagerImpl) ResetWorkflowExecution(request *ResetWorkflowExec
...
+				"current replication state and current version histories cannot both be set."),
+		}
+	}
+	if request.InsertVersionHistories != nil && request.InsertReplicationState != nil {
+		return &workflow.InternalServiceError{
+			Message: fmt.Sprintf("Reset workflow execution with invalid request:" +
+				"insert replication state and insert version histories cannot both be set."),
+		}
+	}
+	return nil
```
wxing1292: missing the branch token & next event ID check

PR #1875, opened 2019-05-17
Returning scheduledTime/startedTime for Polling Decision tasks

longquanzheng:


longquanzheng:
> it is unfortunate that cassandra use decision timestamp for started time
> is it possible for mysql to use decision schedule timestamp / decision start timestamp?

@venkat1109  What is the status of MySQL (especially for outside)? It is OK to change a existing column name which is not backward compatible?

PR #1876, opened 2019-05-17
persistence: add documentation

venkat1109:
This patch adds basic documentation for persistence. This includes how to setup config for different persistence implementations, how to add support for new database etc.

### DIFF
--- a/docs/persistence.md
+++ b/docs/persistence.md
@@ -1,0 +1,128 @@
+# Overview
+Cadence has a well defined API interface at the persistence layer. Any database that supports multi-row transactions on
+a single shard or partition can be used as persistence for cadence. This includes cassandra, dynamoDB, auroraDB, MySQL,
+Postgres and may others. There are currently two supported database implementations at the persistence layer - 
+cassandra and MySQL. This doc shows how to run cadence with cassandra and mysql. It also describes the steps involved
+in adding support for a new database at the persistence layer.
+ 
+# Getting started on mac
+## Cassandra
+### Start cassandra
+```
+brew install cassandra
+brew services start cassandra
+```
+### Install cadence schema
+```
+cd $GOPATH/github.com/uber/cadence
+make install-schema
+```
+
+### Start cadence server
+```
+cd $GOPATH/github.com/uber/cadence
+./cadence-server start --services=frontend,matching,history,worker
+```  
+ 
+## MySQL
+### Start MySQL server
+```
+brew install mysql
+brew services start mysql
+```
+### Install cadence schema
+```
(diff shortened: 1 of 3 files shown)


docs/persistence.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+# Configuration
+## Common to all persistence implementations
+There are two major sub-subsystems within cadence that need persistence - cadence-core and visibility. cadence-core is
+the workflow engine that uses persistence to store state tied to domains, workflows, workflow histories, task lists 
+etc. cadence-core powers almost all of the cadence APIs. cadence-core could be further broken down into multiple 
+subs-systems that have slightly different persistence workload characteristics. But for the purpose of simplicity, we 
+don't expose these sub-systems today but this may change in future. Visibility is the sub-system that powers workflow 
+search. This includes APIs such as ListOpenWorkflows, ListClosedWorkflows or ListWorkflowsWithTag etc. Today, it is
```
vancexu: Probably remove ListWorkflowsWithTag because we don't have it on Cassandra or MySQL.

docs/persistence.md
```diff
@@ -0,0 +1,128 @@
...
+# Configuration
+## Common to all persistence implementations
+There are two major sub-subsystems within cadence that need persistence - cadence-core and visibility. cadence-core is
+the workflow engine that uses persistence to store state tied to domains, workflows, workflow histories, task lists 
+etc. cadence-core powers almost all of the cadence APIs. cadence-core could be further broken down into multiple 
+subs-systems that have slightly different persistence workload characteristics. But for the purpose of simplicity, we 
+don't expose these sub-systems today but this may change in future. Visibility is the sub-system that powers workflow 
+search. This includes APIs such as ListOpenWorkflows, ListClosedWorkflows or ListWorkflowsWithTag etc. Today, it is 
+possible to run a cadence server with cadence-core backed by one database and cadence-visibility backed by another kind 
+of database. The top level persistence configuration looks like the following:
```
vancexu: Might worth to mention cadence support visibility on ElasticSearch which support search attributes and large scale (10M+ workflows)

vancexu: And I will add a doc for elasticsearch setup

venkat1109: I updated the docs to indicate that the full feature set of visibility is supported only in elastic search and that, a limited feature set is supported in cassandra /mysql. Would be awesome if you add the setup instructions / additional details for ES later.

docs/persistence.md
venkat1109: 👍

dmetzgar:
Fixes #1841

PR #1879, opened 2019-05-18
Add metrics for archival request counts

andrewjdawson2016:
These metrics enable us to produce availability graphs for archival.

### DIFF
--- a/service/worker/archiver/archiver.go
+++ b/service/worker/archiver/archiver.go
@@ -100,6 +100,7 @@ func (a *archiver) Finished() []uint64 {
 }
 
 func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metrics.Client, request ArchiveRequest) {
+	metricsClient.IncCounter(metrics.ArchiverScope, metrics.CadenceRequests)
 	sw := metricsClient.StartTimer(metrics.ArchiverScope, metrics.ArchiverHandleRequestLatency)
 	logger = tagLoggerWithRequest(logger, request)
 	ao := workflow.ActivityOptions{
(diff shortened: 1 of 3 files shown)


service/worker/archiver/archiver.go
```diff
@@ -131,6 +133,7 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
 			NonRetriableErrorReasons: deleteHistoryActivityNonRetryableErrors,
 		},
 	}
+	metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverDeleteWithRetriesCount)
```
shreyassrivatsan: so effectively `ArchiverDeleteWithRetriesCount` will be equal to `ArchiverUploadWithRetriesCount` and be same as cadence requests..  I am ok with multiple metrics, but why do we need them?

service/worker/archiver/archiver.go
```diff
@@ -113,6 +114,7 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
 		},
 	}
 	actCtx := workflow.WithActivityOptions(ctx, ao)
+	metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverUploadWithRetriesCount)
```
shreyassrivatsan: technically you can use the `timertype:count` series for this and do not need a new separate emit here.. So then you will have the counts line up with the timer metrics exactly..

service/worker/archiver/archiver_test.go
```diff
@@ -72,6 +75,7 @@ func (s *archiverSuite) TearDownTest() {
 }
 
 func (s *archiverSuite) TestHandleRequest_UploadFails_NonRetryableError() {
+
```
yycptt: Empty line.

service/worker/archiver/archiver.go
shreyassrivatsan: something like this - fetch name:activity_info_count service:cadence-history deployment:prod-phx2  operation:sessionstats stats_type:count  timertype:count | transformNull 0 | aliasByTags name domain

andrewjdawson2016: Oh interesting.

andrewjdawson2016:
Using existing latency metrics instead of these counter metrics.

PR #1884, opened 2019-05-20
Separate workflow resetor from history replicator

wxing1292:


service/history/historyReplicator.go
```diff
@@ -1102,3 +1094,12 @@ func newRetryTaskErrorWithHint(msg string, domainID string, workflowID string, r
 		NextEventId: common.Int64Ptr(nextEventID),
 	}
 }
+
+func notify(shard ShardContext, historyEngine *historyEngineImpl,
```
longquanzheng: Should we move this to shardContext?

Issue #1757, opened 2019-04-30
3+DC - History Cleanup With Multiple Branches


wxing1292:

yux0:
#1885 merged

PR #1885, opened 2019-05-21
3DC multi branches cleanup

yux0:
In 3+DC, the mutable state may have multi-branches info. When a workflow is cleanup, all branches need to be deleted. Currently, the logic is to remove all non current branches (not max version), this is due to archival only archive current branch. In the future, when archival supports archive multiple branches. The cleanup can delete all branches.


dmetzgar:
Fixes #1757

service/history/timerQueueProcessorBase.go
```diff
@@ -704,15 +705,20 @@ func (t *timerQueueProcessorBase) deleteWorkflowHistory(task *persistence.TimerT
 				tag.TaskType(task.GetTaskType()))
 			versionHistories := msBuilder.GetAllBranches()
 
-			for _, versionHistory := range versionHistories.Histories {
+		CleanupLoop:
+			for idx, versionHistory := range versionHistories.Histories {
+				if reflect.DeepEqual(versionHistory.BranchToken, msBuilder.GetCurrentBranch()) {
```
wxing1292: bytes.Equal

service/history/timerQueueProcessorBase.go
```diff
@@ -704,15 +705,20 @@ func (t *timerQueueProcessorBase) deleteWorkflowHistory(task *persistence.TimerT
 				tag.TaskType(task.GetTaskType()))
 			versionHistories := msBuilder.GetAllBranches()
 
-			for _, versionHistory := range versionHistories.Histories {
+		CleanupLoop:
+			for idx, versionHistory := range versionHistories.Histories {
```
wxing1292: probably also need to do a null check here

service/history/timerQueueProcessorBase.go
yux0: Done

service/history/timerQueueProcessorBase.go
yux0: Done

PR #1887, opened 2019-05-21
Add version histories info into GetMutableSate response

yux0:
Add version histories info into GetMutableSate response


common/persistence/versionHistory.go
```diff
@@ -165,3 +167,17 @@ func (h *VersionHistories) AddHistory(item VersionHistoryItem, local VersionHist
 func (h *VersionHistories) GetHistories() []VersionHistory {
 	return h.versionHistories
 }
+
+// ToThrift return thrift format of version histories
+func (h *VersionHistories) ToThrift() *shared.VersionHistories {
+	tHistories := &shared.VersionHistories{}
+	for _, history := range h.versionHistories {
+		tHistory := &shared.VersionHistory{BranchToken: history.branchToken}
```
wxing1292: i think we should make a copy here

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1413,3 +1413,20 @@ struct HistoryBranch{
   20: optional string branchID
   30: optional list<HistoryBranchRange>  ancestors
 }
+
+// VersionHistoryItem contains signal eventID and the corresponding version
+struct VersionHistoryItem{
+  10: optional i64 eventID
```
wxing1292: `i64 (js.type = "Long") `

service/history/mutableState.go
```diff
@@ -115,6 +115,7 @@ type (
 		GetStartEvent() (*workflow.HistoryEvent, bool)
 		GetContinueAsNew() *persistence.CreateWorkflowExecutionRequest
 		GetCurrentBranch() []byte
+		GetAllBranches() persistence.VersionHistories
```
wxing1292: GetAllVersionHistory?

service/history/mutableStateBuilder.go
```diff
@@ -81,6 +81,7 @@ type (
 
 		executionInfo    *persistence.WorkflowExecutionInfo // Workflow mutable state info.
 		replicationState *persistence.ReplicationState
+		versionHistories persistence.VersionHistories
```
wxing1292: plz change this to a pointer so we can use `null`

common/persistence/versionHistory.go
```diff
@@ -165,3 +167,17 @@ func (h *VersionHistories) AddHistory(item VersionHistoryItem, local VersionHist
 func (h *VersionHistories) GetHistories() []VersionHistory {
 	return h.versionHistories
 }
+
+// ToThrift return thrift format of version histories
+func (h *VersionHistories) ToThrift() *shared.VersionHistories {
```
wxing1292: suggestion, also add a function `ToThrift ` on VersionHistory

service/history/historyEngine.go
```diff
@@ -581,6 +581,7 @@ func (e *historyEngineImpl) getMutableState(ctx ctx.Context,
 
 	executionInfo := msBuilder.GetExecutionInfo()
 	execution.RunId = context.getExecution().RunId
+	versionHistories := msBuilder.GetAllBranches()
```
wxing1292: could you also add a TODO in GetMutableState function (the one above), specifying that we need to take branching into consideration?

common/persistence/versionHistory.go
yux0: Done

idl/github.com/uber/cadence/shared.thrift
yux0: Done

service/history/mutableStateBuilder.go
yux0: Done

common/persistence/versionHistory.go
yux0: Good point. done

PR #1895, opened 2019-05-23
Add check for total size on respond decision task

andrewjdawson2016:
Questions:
1. Do we really care about each task's individual size or do we only care about the sum? If we only care about the sum size this logic can be simplified.
2. If we do care about the total size and individual sizes separately then what default config values do we want to give to total size? Currently I am using the same defaults for both individual size and total size.

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -983,20 +983,25 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(ctx ctx.Context,
 }
 
 type decisionBlobSizeChecker struct {
-	sizeLimitWarn  int
-	sizeLimitError int
-	domainID       string
-	workflowID     string
-	runID          string
-	metricsClient  metrics.Client
-	logger         log.Logger
-	msBuilder      mutableState
-	completedID    int64
+	sizeLimitWarn       int
+	sizeLimitError      int
+	totalSizeLimitWarn  int
+	totalSizeLimitError int
+	totalSize           int
+	domainID            string
+	workflowID          string
+	runID               string
+	metricsClient       metrics.Client
+	logger              log.Logger
+	msBuilder           mutableState
+	completedID         int64
 }
 
 func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(blob []byte, message string) (bool, error) {
+	blobSize := len(blob)
+	c.totalSize = c.totalSize + blobSize
 	err := common.CheckEventBlobSizeLimit(
-		len(blob),
+		blobSize,
 		c.sizeLimitWarn,
 		c.sizeLimitError,
 		c.domainID,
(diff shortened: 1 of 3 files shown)


service/history/historyEngine.go
```diff
@@ -1628,6 +1659,9 @@ Update_History_Loop:
 					return nil, &workflow.BadRequestError{Message: fmt.Sprintf("Unknown decision type: %v", *d.DecisionType)}
 				}
 			}
+			if err := sizeChecker.failWorkflowIfTotalSizeExceedsLimit(); err != nil {
```
samarabbas: Looks like the existing logic already has some bugs.  We are not setting the isComplete flag which will result us not creating the appropriate timer and transfer tasks for completed workflows.

service/history/service.go
```diff
@@ -226,12 +228,14 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int, enableVisibilit
...
 
-		BlobSizeLimitError:     dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 2*1024*1024),
-		BlobSizeLimitWarn:      dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 256*1024),
-		HistorySizeLimitError:  dc.GetIntPropertyFilteredByDomain(dynamicconfig.HistorySizeLimitError, 200*1024*1024),
-		HistorySizeLimitWarn:   dc.GetIntPropertyFilteredByDomain(dynamicconfig.HistorySizeLimitWarn, 50*1024*1024),
-		HistoryCountLimitError: dc.GetIntPropertyFilteredByDomain(dynamicconfig.HistoryCountLimitError, 200*1024),
-		HistoryCountLimitWarn:  dc.GetIntPropertyFilteredByDomain(dynamicconfig.HistoryCountLimitWarn, 50*1024),
+		BlobSizeLimitError:      dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 2*1024*1024),
+		BlobSizeLimitWarn:       dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 256*1024),
+		TotalBlobSizeLimitError: dc.GetIntPropertyFilteredByDomain(dynamicconfig.BlobSizeLimitError, 2*1024*1024),
```
samarabbas: This is more like transaction size limit rather than blob size.  Let's rename this appropriately.
Currently the underlying DB limit is 15MB, so we should probably use 14MB as error threshold and 10MB as warning threshold.

service/history/historyEngine.go
```diff
@@ -1021,6 +1026,32 @@ func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(blob []byte
 	return true, nil
 }
 
+func (c *decisionBlobSizeChecker) failWorkflowIfTotalSizeExceedsLimit() error {
+	err := common.CheckEventBlobSizeLimit(
```
samarabbas: I'm not convinced enforcing this limit by accumulating the user blob sizes is the right approach.  This is more like system protection and needs to be enforced as the total size of transaction when we are appending history events to events table.

PR #1897, opened 2019-05-23
Enforce transaction size limit

andrewjdawson2016:


wxing1292:
TransactionSizeError should be translated to BadRequest in history handler?

common/persistence/cassandra/cassandraHistoryV2Persistence.go
```diff
@@ -107,6 +113,11 @@ func convertCommonErrors(operation string, err error) error {
 // AppendHistoryNodes upsert a batch of events as a single node to a history branch
 // Note that it's not allowed to append above the branch's ancestors' nodes, which means nodeID >= ForkNodeID
 func (h *cassandraHistoryV2Persistence) AppendHistoryNodes(request *p.InternalAppendHistoryNodesRequest) error {
+	if len(request.Events.Data) > h.transactionSizeLimit {
+		return &p.TransactionSizeError{
+			Msg: "AppendHistoryNodes operation failed, transaction size was too large",
```
venkat1109: include requested_tx_size: len(request.Events.Data) in the error string. Same comment in other places where the size limit is checked

service/history/historyEngine.go
```diff
@@ -1604,6 +1610,7 @@ Update_History_Loop:
 						return nil, err
 					}
 					if failWorkflow {
+						isComplete = true
```
venkat1109: was this a bug that existed before ? I don't follow how the rest of the changes are related to this

andrewjdawson2016: This is a bug that existed. It should have been a separate PR.

config/development.yaml
```diff
@@ -8,11 +8,13 @@ persistence:
         hosts: "127.0.0.1"
         keyspace: "cadence"
         consistency: "One"
+        maxTransactionSize: 14000000
```
wxing1292: 14 MB?
i think we should target < 1MB for testing and < 1.8 MB for prod

common/metrics/defs.go
```diff
@@ -1143,6 +1143,7 @@ const (
 	PersistenceErrCurrentWorkflowConditionFailedCounter
 	PersistenceErrTimeoutCounter
 	PersistenceErrBusyCounter
+	PersistenceErrTransactionSizeCounter
```
samarabbas: I think you can get rid of this counter as we already emit the transaction size from application layer.

common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -93,6 +97,11 @@ func (h *cassandraHistoryPersistence) Close() {
 }
 
 func (h *cassandraHistoryPersistence) AppendHistoryEvents(request *p.InternalAppendHistoryEventsRequest) error {
+	if len(request.Events.Data) > h.transactionSizeLimit {
```
samarabbas: I think we can enforce this at the common persistence layer, so we don't have to do it for each implementation of the store.

common/service/config/config.go
```diff
@@ -178,6 +178,8 @@ type (
 		MaxQPS int `yaml:"maxQPS"`
 		// MaxConns is the max number of connections to this datastore for a single keyspace
 		MaxConns int `yaml:"maxConns"`
+		// TransactionSizeLimit is the max data size of a single transaction to this datastore
+		TransactionSizeLimit int `yaml:"maxTransactionSize"`
```
samarabbas: Let's move this as a dynamic config knob.

andrewjdawson2016:
This diff does the following:
1. Enforces transaction size limit at persistence layer
2. Fixes existing bug in RespondDecisionTaskCompleted. Workflows which were being failed by us due to a size check validation failure were not being cleaned up.
3. Makes two small refactors in RespondDecisionTaskCompleted.
4. Fails workflow if transaction size error occurs.

andrewjdawson2016:
Screen Shot 2019-05-26 at 2 19 24 PM

Here is an example of a workflow history which results from triggering a transaction size beyond limit.

dmetzgar:
Fixes #1894

common/constants.go
```diff
@@ -107,3 +107,8 @@ const (
 	// below which a warning will be logged
 	CriticalLongPollTimeout = time.Second * 20
 )
+
+const (
+	// DefaultTransactionSizeLimit is the largest allowed transaction size to persistence
+	DefaultTransactionSizeLimit = 14 * 1024 * 1024
```
wxing1292: consider 5 MB?

andrewjdawson2016: @samarabbas What do you think about this? My understanding was the limit on cassandra and mysql was more than 14MB. But looking at this https://docs.datastax.com/en/cql/3.3/cql/cql_reference/refLimits.html it looks like the limit might be lower?

samarabbas: Currently the size of each transaction is bounded by cassandra config value of 'max_mutation_size_in_kb'.  The default value is 16MB for this config knob.  I think we should allow transactions size to go as much as the allowed limit by persistence choice.  I think 14MB makes sense as a default.

service/history/historyEngine.go
```diff
@@ -1728,11 +1696,48 @@ Update_History_Loop:
...
+				if err != nil {
+					return nil, err
+				}
+				transferTasks = []persistence.Task{tranT}
+				timerTasks = []persistence.Task{timerT}
+				transactionID, err3 = e.shard.GetNextTransferTaskID()
+				if err3 != nil {
+					return nil, err3
+				}
+				if continueAsNewBuilder != nil {
```
samarabbas: This cannot have continueAsNewBuilder.  Let's remove this branch.

service/history/historyEngine.go
```diff
@@ -1728,11 +1696,48 @@ Update_History_Loop:
...
+				// clears mutable state if error is returned
+				msBuilder, err = context.loadWorkflowExecution()
+				if err != nil {
+					return nil, err
+				}
+				attributes := &workflow.FailWorkflowExecutionDecisionAttributes{
+					Reason:  common.StringPtr(common.FailureReasonTransactionSizeExceedsLimit),
+					Details: []byte(updateErr.Error()),
+				}
+				if evt := msBuilder.AddFailWorkflowEvent(completedID, attributes); evt == nil {
```
samarabbas: Probably we should use "Force Terminate" pattern and complete the workflow with Terminate event.  We do the same thing when workflow breaches overall history size and event limit.

PR #1899, opened 2019-05-23
Fix logging tags mismatch for component and domainName 

longquanzheng:


service/history/historyEngine.go
```diff
@@ -1110,7 +1110,7 @@ Update_History_Loop:
 			e.throttledLogger.Warn("the number of auto-reset points is exceeding the limit, will do rotating.",
 				tag.WorkflowDomainName(domainEntry.GetInfo().Name),
 				tag.WorkflowID(workflowExecution.GetWorkflowId()),
-				tag.WorkflowDomainName(workflowExecution.GetRunId()),
+				tag.WorkflowRunID(workflowExecution.GetRunId()),
```
samarabbas: Can we tag domainID here too.

longquanzheng: Sure

PR #1904, opened 2019-05-24
Task latency metrics for unsync match

longquanzheng:
For https://github.com/uber/cadence/issues/396


service/matching/matchingEngine.go
```diff
@@ -255,6 +257,7 @@ func (e *matchingEngineImpl) AddActivityTask(addRequest *m.AddActivityTaskReques
 		WorkflowID:             addRequest.Execution.GetWorkflowId(),
 		ScheduleID:             addRequest.GetScheduleId(),
 		ScheduleToStartTimeout: addRequest.GetScheduleToStartTimeoutSeconds(),
+		CreatedTime:            time.Now(),
```
wxing1292: in the future, try use the time source

PR #1905, opened 2019-05-24
Support ElasticSearch in CrossDC integration test

vancexu:
This PR also include XDC integration test of SearchAttributes


Makefile
```diff
@@ -27,7 +27,7 @@ BUILD := ./build
 TOOLS_CMD_ROOT=./cmd/tools
 INTEG_TEST_ROOT=./host
 INTEG_TEST_DIR=host
-INTEG_TEST_XDC_ROOT=./hostxdc
+INTEG_TEST_XDC_ROOT=./host/hostxdc
```
yycptt: I think /host/xdc looks better than /host/hostxdc. What do you think?

host/util.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"github.com/olivere/elastic"
+	"github.com/stretchr/testify/suite"
+	"io/ioutil"
+	"time"
+)
+
+// CreateESClient create ElasticSearch client for test
+func CreateESClient(s suite.Suite, url string) *elastic.Client {
```
yycptt: Naming of this file is a little bit confusing to me. How about moving these methods to elasticsearch_test.go? or change the file name to something like es_util.go?

Makefile
vancexu: updated

host/util.go
vancexu: I would keep it. 
The purpose is to put shared code for host and hostxdc in one place. I expect more shared code (not only related to es) in future, which should live together.

PR #1906, opened 2019-05-24
Improve perf for emitting domainTag metrics in matching

longquanzheng:


service/matching/matchingEngine.go
```diff
@@ -622,8 +622,8 @@ func (e *matchingEngineImpl) createPollForDecisionTaskResponse(context *taskCont
 		}
 		token, _ = e.tokenSerializer.Serialize(taskoken)
 		if context.syncResponseCh == nil {
-			scope := domainTaggedMetricScope(e.domainCache, task.DomainID, e.metricsClient, metrics.MatchingPollForDecisionTaskScope)
-			scope.RecordTimer(metrics.AsyncMatchLatency, time.Since(task.CreatedTime))
+			scope := e.metricsClient.Scope(metrics.MatchingPollForDecisionTaskScope)
+			scope.Tagged(metrics.DomainTag(context.domainName)).RecordTimer(metrics.AsyncMatchLatency, time.Since(task.CreatedTime))
```
venkat1109: does scope.Tagged() deal with empty string as parameter correctly ?

service/matching/taskListManager.go
```diff
@@ -372,13 +376,19 @@ func (c *taskListManagerImpl) GetTaskContext(
 		WorkflowId: common.StringPtr(task.WorkflowID),
 		RunId:      common.StringPtr(task.RunID),
 	}
+
+	// reload from domainCache in case it got empty result during construction
+	if len(c.domainName) == 0 {
+		c.domainName, c.domainScope = domainNameAndMetricScope(c.domainCache, c.taskListID.domainID, c.metricsClient, metrics.MatchingTaskListMgrScope)
```
venkat1109: i don't think is thread safe.  Even if it is, it will be a benign race, which will still trigger race detector errors. You will either need a lock or use an atomic type

service/matching/matchingEngine.go
longquanzheng: I don't know but it's always better to have a checker. Will add it here.

service/matching/taskListManager.go
longquanzheng: Sure

service/matching/taskListManager.go
```diff
@@ -372,13 +378,24 @@ func (c *taskListManagerImpl) GetTaskContext(
 		WorkflowId: common.StringPtr(task.WorkflowID),
 		RunId:      common.StringPtr(task.RunID),
 	}
+
+	// reload from domainCache in case it got empty result during construction
+	if len(c.domainName) == 0 {
```
venkat1109: can you create a helper for this whole block of code which checks if domainName is empty and if yes, populates it in a thread safe manner ?

venkat1109: c.domainName != "" is the go standard way

longquanzheng: I don't like arguing about this. But still wondering where it comes from as standard. 
https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go

And we do have both in our codebase.
So I will change it anyway if you think it is needed.

longquanzheng: I don't get this. Probably need to talk offline.

venkat1109: this isn't a blocker and yes, we have both in our code base today, which is unfortunate

longquanzheng: Updated

PR #1908, opened 2019-05-28
matching: metrics related fixes

venkat1109:
Follow patch to #1906. The domainScope was assigned within a lock however it was accessed throughout the code without acquiring any locks. This patch fixes the bug and also gets rid of the locks, in favor of atomics.

There was also another existing bug - the metricsClient was never initialized in the constructor. Also, fixes this problem


service/matching/taskListManager.go
```diff
@@ -741,7 +727,37 @@ func (c *taskListManagerImpl) isTaskAddedRecently(lastAddTime time.Time) bool {
...
 }
 
-// if domainCache return error, it will return "" as domainNamne and a scope without domainName tagged
+func (c *taskListManagerImpl) domainScope() metrics.Scope {
+	scope := c.domainScopeValue.Load().(metrics.Scope)
+	if scope != nil {
+		return scope
+	}
+	c.tryInitDomainNameAndScope()
+	return c.domainScopeValue.Load().(metrics.Scope)
```
shreyassrivatsan: what does this return if the init domain scope does not happen?

venkat1109: constructor always initializes to a valid value. If domain lookup fails, the atomic is initialized with ordinary scope (without domain tag). tryInit() will always transition from one valid state to another.. so, to answer your question, Load() will never return nil

venkat1109:
> So it doesn't improve much perf compared to directly using domainCache
This patch caches the domain metric scope once it's created not the domain metadata. That's about it. DomainCache caches the domain metadata. And I tend to disagree that it's complicated.

longquanzheng:
> > So it doesn't improve much perf compared to directly using domainCache
> > This patch caches the domain metric scope once it's created not the domain metadata. That's about it. DomainCache caches the domain metadata. And I tend to disagree that it's complicated.

I am fine to forget about complexity as it's hard to argue. Adding another atomicValue is unnecessary. How about just use domainCache whenever you need to translate from domainID to domainName?

venkat1109:
> > > So it doesn't improve much perf compared to directly using domainCache
> > > This patch caches the domain metric scope once it's created not the domain metadata. That's about it. DomainCache caches the domain metadata. And I tend to disagree that it's complicated.
> 
> I am fine to forget about complexity as it's hard to argue. Adding another atomicValue is unnecessary. How about just use domainCache whenever you need to translate from domainID to domainName?

Discussed offline.

PR #1916, opened 2019-05-28
Add firstRunID and correct originalRunID

longquanzheng:


service/history/mutableStateBuilder.go
```diff
@@ -1306,7 +1306,7 @@ func (e *mutableStateBuilder) DeleteSignalRequested(requestID string) {
 
 func (e *mutableStateBuilder) addWorkflowExecutionStartedEventForContinueAsNew(domainID string,
```
shreyassrivatsan: The caller of this method will have to pass in `firstRunID` instead of `originalRunID`..

shreyassrivatsan: or maybe both?

longquanzheng: originalRunID is in execution.GetRunID. I don't think we should pass duplicated info here. We previously called firstRunID as originalRunID but later Max decided to change it to firstRunID and reserve originalRunID as the runID when WF start event is written.

shreyassrivatsan: Ok, makes sense.. We still need to change the caller to pass in the `firstRunID` right?

longquanzheng: I See. Thanks for pointing it out

meiliang86: Can you add comments to these two fields explaining the meaning? This can be easily forget.

longquanzheng: Sure, added to AddWorkflowExecutionStartedEvent

PR #1921, opened 2019-05-29
Change how domain scopes are created for front end handler metrics

shreyassrivatsan:
- create startRequestProfileWithDomain method and use in places where domain name is passed in
- create domainGetter interface so we can retrieve domain from request body
- update timer logic to dual emit to all internally if domain tag is used
- create an internal stopwatch with same methods as tally.Stopwatch to assist with the above.

### DIFF
--- a/common/metrics/scope.go
+++ b/common/metrics/scope.go
@@ -27,17 +27,18 @@ import (
 )
 
 type metricsScope struct {
-	scope tally.Scope
-	defs  map[int]metricDefinition
+	scope    tally.Scope
+	defs     map[int]metricDefinition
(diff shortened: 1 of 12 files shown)


common/metrics/scope.go
```diff
@@ -27,17 +27,18 @@ import (
 )
 
 type metricsScope struct {
-	scope tally.Scope
-	defs  map[int]metricDefinition
+	scope    tally.Scope
+	defs     map[int]metricDefinition
+	isDomain bool
```
venkat1109: consider naming this hasDomainTag or isDomainTagged

PR #1924, opened 2019-05-30
Add automated logic to migration domain from V1 table to V2 table

wxing1292:


common/persistence/cassandra/cassandraMetadataPersistenceProxy.go
```diff
@@ -74,6 +76,27 @@ func (m *metadataManagerProxy) GetDomain(request *p.GetDomainRequest) (*p.Intern
 	resp, err = m.metadataMgr.GetDomain(request)
 	if err == nil {
 		resp.TableVersion = p.DomainTableVersionV1
+		if !resp.IsGlobalDomain {
```
longquanzheng: Add a TODO comment to  delete / revert the PR after next next release

wxing1292: there is already a TODO at the very top
basically, the proxy file should be deleted once the migration is done

PR #1931, opened 2019-05-30
Archival: Delete blob activity

yycptt:


service/worker/archiver/activities.go
```diff
@@ -117,18 +120,25 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
 		if err != nil && err != blobstore.ErrBlobNotExists {
-			logger.Error(uploadErrorMsg, tag.UploadFailReason("could not get blob tags"), tag.ArchivalBucket(bucket), tag.ArchivalBlobKey(key.String()))
+			logger.Error(uploadErrorMsg, tag.UploadFailReason("could not get blob tags"), tag.ArchivalBucket(request.BucketName), tag.ArchivalBlobKey(key.String()))
 			return err
 		}
+
+		// if we can get tag, it means this blob already exists
+		// Based on if it's the last blob, we need to break the loop or continue
 		runConstTest := false
 		blobAlreadyExists := err == nil
```
andrewjdawson2016: Here is an example of what I mean by making code clear enough it does not need comments. I could have written this code as `if err == nil`. But that would be confusing. So instead I created a named variable to describe what I mean when `err == nil` and my if uses that variable.

service/worker/archiver/activities.go
```diff
@@ -226,6 +243,70 @@ func deleteHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 	return nil
 }
 
+// deleteBlobActivity deletes uploaded history blobs from blob store.
+// method will retry all retryable operations until context expires.
+// method will always return either: nil, contextTimeoutErr or an error from deleteBlobActivityNonRetryableErrors.
+func deleteBlobActivity(ctx context.Context, request ArchiveRequest) (err error) {
```
andrewjdawson2016: Can we add a TODO: as part of method comment that says this method should take a list of blobs uploaded once correct heartbeating is supported?

service/worker/archiver/archiver.go
```diff
@@ -114,14 +114,39 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
+	err := workflow.ExecuteActivity(actCtx, uploadHistoryActivityFnName, request).Get(actCtx, nil)
+	if err != nil {
+		logger.Error("failed to upload history, will delete all uploaded blobs and moving on to deleting history without archiving", tag.Error(err))
 		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverUploadFailedAllRetriesCount)
 	} else {
 		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverUploadSuccessCount)
 	}
 	uploadSW.Stop()
 
+	// Uncomment this part if we need to delete uploaded blobs
```
andrewjdawson2016: Let's make this code live now. I do not see any risk.

service/worker/archiver/activities.go
```diff
@@ -301,6 +382,28 @@ func downloadBlob(ctx context.Context, blobstoreClient blobstore.Client, bucket
...
+func deleteBlob(ctx context.Context, blobstoreClient blobstore.Client, bucket string, key blob.Key) (bool, error) {
+	dCtx, cancel := context.WithTimeout(ctx, blobstoreTimeout)
+	deleted, err := blobstoreClient.Delete(dCtx, bucket, key)
+	cancel()
+	for err != nil {
+		if err == blobstore.ErrBlobNotExists {
+			return false, nil
+		}
+		if !blobstoreClient.IsRetryableError(err) {
+			return deleted, cadence.NewCustomError(errDeleteBlob, err.Error())
```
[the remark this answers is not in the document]
yycptt: I am pretty sure only error.Reason will be used to determine if retry is needed. Check this line, https://github.com/uber/cadence/blob/master/service/history/historyEngine.go#L1896.

Sure I can err.Error() in all other places.

service/worker/archiver/activities.go
```diff
@@ -117,18 +120,25 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		historyBlobReader = NewHistoryBlobReader(NewHistoryBlobIterator(request, container, domainName, clusterName))
 	}
 	blobstoreClient := container.Blobstore
+
+	// start uploading blobs
```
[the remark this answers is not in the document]
yycptt: Thanks for the explanation. I will keep in mind. 

Actually, I just want to add some empty lines between these line to make it a little bit easier to read.

service/worker/archiver/activities.go
yycptt: Got it!

service/worker/archiver/activities.go
```diff
@@ -41,23 +41,27 @@ import (
...
 	errConstructBlob = "failed to construct blob"
 	errDownloadBlob  = "could not download existing blob"
+	errDeleteBlob    = "could not delete existing blob"
 
 	errDeleteHistoryV1 = "failed to delete history from events_v1"
 	errDeleteHistoryV2 = "failed to delete history from events_v2"
 )
 
 var (
-	uploadHistoryActivityNonRetryableErrors = []string{errGetDomainByID, errGetTags, errUploadBlob, errReadBlob, errEmptyBucket, errConstructBlob, errDownloadBlob}
```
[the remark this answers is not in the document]
yycptt: My fault. It's introduced during conflict resolution when merging master.

service/worker/archiver/activities.go
yycptt: Done.

service/worker/archiver/archiver.go
yycptt: Done.

service/worker/archiver/activities.go
```diff
@@ -104,8 +108,7 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		scope.IncCounter(metrics.ArchiverSkipUploadCount)
 		return nil
 	}
-	bucket := domainCacheEntry.GetConfig().ArchivalBucket
-	if len(bucket) == 0 {
+	if len(request.BucketName) == 0 {
```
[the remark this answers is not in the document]
yycptt: Done.

service/worker/archiver/activities.go
```diff
@@ -226,6 +243,70 @@ func deleteHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+	for {
+		key, err := NewHistoryBlobKey(request.DomainID, request.WorkflowID, request.RunID, request.CloseFailoverVersion, pageToken)
+		if err != nil {
+			logger.Error("could not construct blob key", tag.Error(err))
+			return cadence.NewCustomError(errConstructKey, err.Error())
+		}
+
+		deleted, err := deleteBlob(ctx, blobstoreClient, request.BucketName, key)
+		if err != nil {
+			logger.Error("failed to delete blob", tag.ArchivalBucket(request.BucketName), tag.ArchivalBlobKey(key.String()), tag.Error(err))
```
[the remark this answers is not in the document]
yycptt: Done.

service/worker/archiver/activities.go
```diff
@@ -226,6 +243,70 @@ func deleteHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+		pageToken++
+	}
+
+	// delete index blob
+	indexBlobKey, err := NewHistoryIndexBlobKey(request.DomainID, request.WorkflowID, request.RunID)
+	if err != nil {
+		logger.Error("could not construct index blob key", tag.Error(err))
+		return cadence.NewCustomError(errConstructKey, err.Error())
+	}
+	if _, err := deleteBlob(ctx, blobstoreClient, request.BucketName, indexBlobKey); err != nil {
```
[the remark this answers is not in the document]
yycptt: Makes sense. Now we have the invariant that if a version exists in the index blob, history for that version is complete.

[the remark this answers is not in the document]
yycptt: Also, when deleting versions. There are three cases: 
1. There are multiple versions, one of them is deleted. We need to upload a new blob.
2. The version we need to delete is the only version left. We need to delete the index blob.
3. The version we need to delete does not exist in the tag. Do nothing.

service/worker/archiver/activities.go
```diff
@@ -226,6 +235,96 @@ func deleteHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+		logger.Error("could not construct index blob key", tag.Error(err))
+		return cadence.NewCustomError(errConstructKey, err.Error())
+	}
+	existingVersions, err := getTags(ctx, blobstoreClient, request.BucketName, indexBlobKey)
+	if err != nil && err != blobstore.ErrBlobNotExists {
+		logger.Error("could not get index blob tags", tag.ArchivalBlobKey(indexBlobKey.String()), tag.Error(err))
+		return err
+	}
+	if err != blobstore.ErrBlobNotExists {
+		if indexBlobWithVersion := deleteVersion(request.CloseFailoverVersion, existingVersions); indexBlobWithVersion != nil {
```
andrewjdawson2016: Can we call this indexBlobWithoutVersion?

service/worker/archiver/archiver.go
```diff
@@ -114,14 +114,38 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
+	if err != nil {
+		logger.Error("failed to upload history, will delete all uploaded blobs and moving on to deleting history without archiving", tag.Error(err))
 		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverUploadFailedAllRetriesCount)
 	} else {
 		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverUploadSuccessCount)
 	}
 	uploadSW.Stop()
 
+	if err != nil {
+		// If uploading or version check fails, delete all uploaded blobs
```
andrewjdawson2016: This comment is a little misleading. Delete blobs should not care if its deleting blobs because upload failed or version check failed. If version check failed uploadHistory will return an error. So we just blindly delete uploaded blobs on any error from upload - which might be a version check error.

yycptt: Deleted. You are right it's misleading. I did this because when I wrote this code, I thought the version check may be done outside the upload activity.

PR #1939, opened 2019-06-01
Add retention period check

yycptt:
https://github.com/uber/cadence/issues/1926


host/xdc/integration_domain_failover_test.go
```diff
@@ -132,10 +132,11 @@ func (s *integrationClustersTestSuite) TestDomainFailover() {
 	domainName := "test-domain-for-fail-over-" + common.GenerateRandomString(5)
```
wxing1292: you can delete this file.

PR #1941, opened 2019-06-03
Add DNS ringpop discovery provider

dlh:
Add a new discovery provider to simplify deployments for DNS-based service
discovery platforms, such as HashiCorp Consul and Kubernetes (headless
services).


common/service/config/ringpop.go
```diff
@@ -202,6 +209,42 @@ func (factory *RingpopFactory) getChannel(dispatcher *yarpc.Dispatcher) (*tcg.Ch
 	return ch, nil
 }
 
+type minimalResolver interface {
+	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+}
+
+type dnsProvider struct {
+	Resolver        minimalResolver
+	UnresolvedHosts []string
```
venkat1109: can you make this unresolved a set ? that way, we can also be sure that there are no duplicates in the input list and so, no duplicates on the output list of resolved addrs. The population of set can be done in a constructor of form newDNSProvider(hosts []string, resolver dnsHostResolver) *dnsProvider {}

common/service/config/ringpop.go
```diff
@@ -202,6 +209,42 @@ func (factory *RingpopFactory) getChannel(dispatcher *yarpc.Dispatcher) (*tcg.Ch
...
+		if !exists {
+			resolved, err = provider.Resolver.LookupHost(context.Background(), host)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		for _, r := range resolved {
+			if !exists {
+				resolvedHosts[host] = append(resolvedHosts[host], r)
```
venkat1109: can you move this code inside of your !exists block before this loop ?

dlh:
@venkat1109 I pushed an update to address your review

common/service/config/ringpop.go
```diff
@@ -202,7 +210,65 @@ func (factory *RingpopFactory) getChannel(dispatcher *yarpc.Dispatcher) (*tcg.Ch
...
+		}
+
+		resolved, exists := resolvedHosts[host]
+		if !exists {
+			resolved, err = provider.Resolver.LookupHost(context.Background(), host)
+			if err != nil {
+				provider.Logger.Warn("Could not resolve host", tag.Address(host), tag.Error(err))
+				continue
+			}
+			for _, r := range resolved {
```
venkat1109: are there reasons to do a slice copy here instead of simply doing  ```resolvedHosts[host] = resolved``` ?

dlh: 🤦‍♂ Nope no reason. I updated it to be direct assignment.

Issue #1901, opened 2019-05-23
Add version check when archiving history


yycptt:
With the new archiving state, mutable state will not be deleted during archiving, it's possible that the history will be overwritten by some higher version event. In that case, we need to stop archiving. 

Basic idea is that if version of the history events read is larger than the version specified in the archival request, then we need to stop.

yycptt:
After discussion, we don't need to perform version check while reading and uploading history. Instead, after uploading has finished, do a version check to ensure "current branch" hasn't changed. If the check fails, delete all uploaded history blobs. Otherwise, delete mutable state and history.

yycptt:
Update: we will still perform the version check during uploading. The main purpose is visibility (emit metrics in such case) and performance (stop early)

PR #1947, opened 2019-06-04
Archival: Add failover version check when uploading history

yycptt:
#1901


service/worker/archiver/util.go
```diff
@@ -96,6 +97,17 @@ func shouldRun(probability float64) bool {
 	return rand.Intn(int(1.0/probability)) == 0
 }
 
+func isHistoryMutated(historyBlob *HistoryBlob, request *ArchiveRequest) bool {
```
andrewjdawson2016: calling this historyMutated is more standard go: https://golang.org/doc/effective_go.html#Getters

service/worker/archiver/util.go
```diff
@@ -96,6 +97,17 @@ func shouldRun(probability float64) bool {
...
+	lastFailoverVersion := common.Int64Default(historyBlob.Header.LastFailoverVersion)
+	if lastFailoverVersion > request.CloseFailoverVersion {
+		return true
+	}
+
+	isLast := common.BoolDefault(historyBlob.Header.IsLast)
```
andrewjdawson2016: The return statement is a little long to reason about. Can we short circut here and say 
```
if !isLast {
    return false
}
```

PR #1951, opened 2019-06-04
Refactor decision logic removing for loop

wxing1292:
* Remove workflowStartEventProvider in favor of mutableState GetStartEvent function
* Refactor handleDecisionTaskStarted to use history engine 
* Remove getWorkflowStartedEvent function

### DIFF
--- a/service/history/decisionHandler.go
+++ b/service/history/decisionHandler.go
@@ -25,11 +25,10 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/uber/cadence/common/cache"
-
 	h "github.com/uber/cadence/.gen/go/history"
 	workflow "github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/metrics"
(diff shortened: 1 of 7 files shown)


service/history/decisionHandler.go
```diff
@@ -110,7 +109,10 @@ func (handler *decisionHandlerImpl) handleDecisionTaskScheduled(
 				transferTasks:  []persistence.Task{&persistence.RecordWorkflowStartedTask{}},
 			}
 
-			startEvent, _ := msBuilder.GetStartEvent()
+			startEvent, found := msBuilder.GetStartEvent()
+			if !found {
```
yux0: Will this check break cron workflow?

service/history/decisionHandler.go
```diff
@@ -295,17 +285,17 @@ func (handler *decisionHandlerImpl) handleDecisionTaskCompleted(
...
-	context, release, err0 := handler.historyCache.getOrCreateWorkflowExecutionWithTimeout(ctx, domainID, workflowExecution)
-	if err0 != nil {
-		return nil, err0
+	context, release, err := handler.historyCache.getOrCreateWorkflowExecutionWithTimeout(ctx, domainID, workflowExecution)
+	if err != nil {
+		return nil, err
 	}
 	defer func() { release(retError) }()
 
 Update_History_Loop:
```
yux0: Could this also be replaced with updateWorkflowExecutionWithAction?

service/history/decisionHandler.go
wxing1292: these are missing checks. without this, code below will encounter nil pointer exception

service/history/decisionHandler.go
wxing1292: handling of decision task completed is really complex.
for now, i prefer stop here

Issue #1456, opened 2019-02-12
3+DC - GetMutableState Modification With Branch Changes


wxing1292:

yux0:
Assign this task to me with an open PR #1953

wxing1292:
#2391

PR #1953, opened 2019-06-05
Handle current branch changes in Getmutablestate

yux0:
Handle current branch changes in Getmutablestate. When the caller is using Getmutablestate with long polling, the API should return immediately when the branch token changed.


dmetzgar:
Fixes #1456

yux0:
Completed in #2396

PR #1957, opened 2019-06-05
Refactor decision task handling for retry / cron continue as new logic

wxing1292:
* Merge retry / cron continue as new logic into one function
* Move cron package into backoff package for unified 'backoff.NoBackoff', instead of 'cron.NoBackoff' and 'common.NoRetryBackoff'


service/history/stateBuilder.go
```diff
@@ -558,8 +558,8 @@ func (b *stateBuilderImpl) scheduleWorkflowTimerTask(event *shared.HistoryEvent,
 	timeout := now.Add(time.Duration(msBuilder.GetExecutionInfo().WorkflowTimeout) * time.Second)
 
 	cronSchedule := b.msBuilder.GetExecutionInfo().CronSchedule
-	cronBackoffDuration := cron.GetBackoffForNextSchedule(cronSchedule, now)
-	if cronBackoffDuration != cron.NoBackoff {
+	cronBackoffDuration := retry.GetBackoffForNextSchedule(cronSchedule, now)
+	if cronBackoffDuration != retry.NoBackoff {
```
yycptt: Package name looks a little bit weird.

common/retry/retrypolicy.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package backoff
+package retry
```
yycptt: If the package name is "retry", may be consider rename RetryPolicy -> Policy.

wxing1292: according to the agreement, rename to backoff

PR #1961, opened 2019-06-06
Add archival size metrics

andrewjdawson2016:
This will enable us to create a histogram of blob sizes and total history upload sizes.

### DIFF
--- a/service/worker/archiver/activities.go
+++ b/service/worker/archiver/activities.go
@@ -124,6 +124,7 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 	blobstoreClient := container.Blobstore
 
 	handledLastBlob := false
+	var totalUploadSize int64
 	for pageToken := common.FirstBlobPageToken; !handledLastBlob; pageToken++ {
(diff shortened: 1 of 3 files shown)


service/worker/archiver/activities.go
```diff
@@ -173,6 +174,9 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 			logger.Error(uploadErrorMsg, tag.ArchivalUploadFailReason(reason), tag.ArchivalBlobKey(key.String()))
 			return cadence.NewCustomError(errConstructBlob, err.Error())
 		}
+		currBlobSize := int64(len(blob.Body))
+		scope.RecordTimer(metrics.ArchiverBlobSize, time.Duration(currBlobSize))
```
samarabbas: Probably it will be useful to dual emit these as domain specific metric.

shreyassrivatsan: I would go further and say you should start emitting all archival metrics as domain tagged if its easy to do that..

andrewjdawson2016: I am already emitting all metrics as domain specific metrics.

PR #1968, opened 2019-06-07
Archival blob integrity check

andrewjdawson2016:


service/worker/archiver/activities.go
```diff
@@ -225,6 +232,53 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		logger.Error(uploadErrorMsg, tag.ArchivalUploadFailReason(errorDetails(err)), tag.ArchivalBlobKey(indexBlobKey.String()), tag.Error(err))
 		return err
 	}
+	if runBlobIntegrityCheck {
```
yycptt: Can we move this to a separate function? The code of upload activity is getting longer and longer.

service/worker/archiver/activities.go
```diff
@@ -225,6 +232,53 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+			}
+			fetchedHistory.Events = append(fetchedHistory.Events, resp.History.Events...)
+			req.NextPageToken = resp.NextPageToken
+		}
+		equal, reason := historiesEqual(fetchedHistory, uploadedHistory)
+		if err != nil {
+			scope.IncCounter(metrics.ArchiverCouldNotRunBlobIntegrityCheckCount)
+			logger.Error("failed to check if histories are equal", tag.Error(err))
+		}
+		if !equal {
```
yycptt: Shall we return an error in this case?

service/worker/archiver/util.go
```diff
@@ -127,3 +130,33 @@ func errorDetails(err error) string {
 	err.(*cadence.CustomError).Details(&details)
 	return details
 }
+
+func historiesEqual(fetchedHistory *clientShared.History, uploadedHistory *shared.History) (bool, string) {
```
yycptt: Return error instead of string?

service/worker/archiver/activities.go
andrewjdawson2016: If we return an error that delete blobs will clean up these blobs. That will make it so we cannot debug why this call failed.

service/worker/archiver/util.go
andrewjdawson2016: same comment as above.

service/worker/archiver/activities.go
andrewjdawson2016: Once we heartbeat progress we can return error as part of the struct and set blobsToDelete to empty.

service/worker/archiver/activities.go
```diff
@@ -225,6 +232,53 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+			if !resp.GetArchived() {
+				scope.IncCounter(metrics.ArchiverBlobIntegrityCheckFailedCount)
+				logger.Error("blob integrity check failed history was not archived")
+				return nil
+			}
+			fetchedHistory.Events = append(fetchedHistory.Events, resp.History.Events...)
+			req.NextPageToken = resp.NextPageToken
+		}
+		equal, reason := historiesEqual(fetchedHistory, uploadedHistory)
+		if err != nil {
```
venkat1109: this can never happen, you already return from the method on all error paths above. This code can be removed

service/worker/archiver/activities.go
```diff
@@ -225,6 +232,53 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+			scope.IncCounter(metrics.ArchiverCouldNotRunBlobIntegrityCheckCount)
+			logger.Error("failed to check if histories are equal", tag.Error(err))
+		}
+		if !equal {
+			scope.IncCounter(metrics.ArchiverBlobIntegrityCheckFailedCount)
+			logger.Error("uploaded history does not match fetched history", tag.ArchivalBlobIntegrityCheckFailReason(reason))
```
venkat1109: also log - (domainID, workflowID, runID, version) for debugging

service/worker/archiver/activities.go
venkat1109: +1 - consider factoring out into its own function

service/worker/archiver/activities.go
```diff
@@ -225,6 +232,53 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+			}
+			if !resp.GetArchived() {
+				scope.IncCounter(metrics.ArchiverBlobIntegrityCheckFailedCount)
+				logger.Error("blob integrity check failed history was not archived")
+				return nil
+			}
+			fetchedHistory.Events = append(fetchedHistory.Events, resp.History.Events...)
+			req.NextPageToken = resp.NextPageToken
+		}
+		equal, reason := historiesEqual(fetchedHistory, uploadedHistory)
```
venkat1109: talked offline, comparing hashes would be a full integrity check  + avoids having to cache the whole history in memory

service/worker/archiver/activities.go
yycptt: I think the logger already included these information.

service/worker/archiver/activities.go
andrewjdawson2016: I want to leave it as part of the same method for now. By making it its own function you have to pass two many parameters and it doesn't buy you much because the whole logic is already inside an if condition, so the reader can easily mentally skip over the section already.

service/worker/archiver/util.go
andrewjdawson2016: Thank you :) I ended up deleting the function all together.

service/worker/archiver/activities.go
andrewjdawson2016: The logger I use everywhere in this activity is tagged with everything useful already.

service/worker/archiver/activities.go
andrewjdawson2016: Thanks for taking through this. I this new approach better.

PR #1973, opened 2019-06-07
Fix list api pagination with custom order

vancexu:
This PR will make sure list API works correctly when using order by.


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -570,12 +571,87 @@ func addMustQuery(dsl *fastjson.Value, queryString string) {
...
+	isSorted := dsl.Exists(dslFieldSort)
+	var sortField string
+
+	if !isSorted { // set default sorting by StartTime desc
+		dsl.Set(dslFieldSort, fastjson.MustParse(jsonSortForOpen))
+		sortField = definition.StartTime
+	} else { // user provide sorting using order by
+		// sort validation on length
+		if len(dsl.GetArray(dslFieldSort)) > 1 {
+			return "", errors.New("only one field can be used to sort")
```
yycptt: Return bad request in this case?

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -570,12 +571,87 @@ func addMustQuery(dsl *fastjson.Value, queryString string) {
...
+		}
+		// sort validation to exclude IndexedValueTypeString
+		obj, _ := dsl.GetArray(dslFieldSort)[0].Object()
+		obj.Visit(func(k []byte, v *fastjson.Value) {
```
yycptt: How about adding a comment here which says fastjson doesn't support get keys of an object? People may be confused why there's only one key in this object and Visit() is used.

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -570,12 +571,87 @@ func addMustQuery(dsl *fastjson.Value, queryString string) {
...
+			if err != nil {
+				return "", err
+			}
+		case string: // field not present, ES will return "-Infinity" or "Infinity"
+			sortVal = fmt.Sprintf(`"%s"`, token.SortValue.(string))
+		}
+	case workflow.IndexedValueTypeKeyword:
+		if token.SortValue != nil {
+			sortVal = fmt.Sprintf(`"%s"`, token.SortValue.(string))
+		} else { // // field not present, ES will return null (so token.SortValue is nil)
```
yycptt: nit: //

common/persistence/elasticsearch/esVisibilityStore.go
vancexu: ListWorkflowExecutions

common/persistence/elasticsearch/esVisibilityStore.go
vancexu: added

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -514,10 +515,10 @@ func getSQLFromCountRequest(request *p.CountWorkflowExecutionsRequest) string {
...
 }
 
-func getCustomizedDSLFromSQL(sql string, domainID string) (*fastjson.Value, bool, error) {
+func getCustomizedDSLFromSQL(sql string, domainID string) (*fastjson.Value, error) {
 	dslStr, _, err := elasticsql.Convert(sql)
 	if err != nil {
-		return nil, false, err
+		return nil, err
 	}
 	dsl := fastjson.MustParse(dslStr) // dsl.String() will be a compact json without spaces
```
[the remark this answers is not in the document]
vancexu: changed

PR #1975, opened 2019-06-07
Decision task attribute validation should result in decision failure, not error

wxing1292:


service/history/decisionTaskHandler.go
```diff
@@ -307,10 +328,26 @@ func (handler *decisionTaskHandlerImpl) handleDecisionCompleteWorkflow(
...
 
 	if handler.hasUnhandledEventsBeforeDecisions {
-		handler.failDecision = true
-		handler.failDecisionCause = workflow.DecisionTaskFailedCauseUnhandledDecision.Ptr()
+		return handler.handlerFailDecision(workflow.DecisionTaskFailedCauseUnhandledDecision, "")
+	}
+
+	err := handler.attrValidator.validateCompleteWorkflowExecutionAttributes(attr)
+	if err != nil {
+		if _, ok := err.(*workflow.BadRequestError); ok {
```
shreyassrivatsan: can we make this a method, `processBadRequestError` or something, that either modifies the error if its bad request or returns it as is?

shreyassrivatsan: I guess it has to take in the fail decision reason..

service/history/historyEngine.go
```diff
@@ -864,7 +864,10 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(ctx ctx.Context,
 				return nil, &h.EventAlreadyStartedError{Message: "Activity task already started."}
 			}
 
-			msBuilder.AddActivityTaskStartedEvent(ai, scheduleID, requestID, request.PollRequest.GetIdentity())
+			_, err := msBuilder.AddActivityTaskStartedEvent(ai, scheduleID, requestID, request.PollRequest.GetIdentity())
+			if err != nil {
```
shreyassrivatsan: nit: `if err := ...`

service/history/historyEngine.go
```diff
@@ -1937,7 +1941,10 @@ func (e *historyEngineImpl) failDecision(context workflowExecutionContext, sched
 		return nil, err
 	}
 
-	msBuilder.AddDecisionTaskFailedEvent(scheduleID, startedID, cause, nil, request.GetIdentity(), "", "", "", 0)
+	_, err = msBuilder.AddDecisionTaskFailedEvent(scheduleID, startedID, cause, nil, request.GetIdentity(), "", "", "", 0)
+	if err != nil {
```
shreyassrivatsan: nit: `if err := ...`

service/history/mutableStateBuilder.go
```diff
@@ -3372,7 +3584,11 @@ func (e *mutableStateBuilder) ReplicateChildWorkflowExecutionTimedOutEvent(event
 	return nil
 }
 
-func (e *mutableStateBuilder) CreateActivityRetryTimer(ai *persistence.ActivityInfo, failureReason string) persistence.Task {
+func (e *mutableStateBuilder) CreateActivityRetryTimer(
+	ai *persistence.ActivityInfo,
+	failureReason string,
+) persistence.Task {
+
```
shreyassrivatsan: thanks for doing all these!

service/history/decisionTaskHandler.go
wxing1292: already done

PR #1979, opened 2019-06-08
Archival: Fix non-deterministic history read

yycptt:
https://github.com/uber/cadence/issues/1980


yycptt:
I think currentPageToken and nextPageToken should be deleted from blob header.

service/worker/archiver/history_blob_iterator.go
```diff
@@ -104,14 +113,15 @@ func (i *historyBlobIterator) Next() (*HistoryBlob, error) {
 	if !i.HasNext() {
 		return nil, errIteratorDepleted
 	}
-	events, nextPersistencePageToken, historyEndReached, err := i.readBlobEvents(i.persistencePageToken)
+	events, nextPersistencePageToken, historyEndReached, numEventsToSkip, err := i.readBlobEvents(i.persistencePageToken, i.numEventsToSkip)
```
andrewjdawson2016: This method is returning a lot of things now, what do you think about making a struct for the result?

service/worker/archiver/history_blob_iterator_test.go
```diff
@@ -500,6 +500,89 @@ func (s *HistoryBlobIteratorSuite) TestNext_Success_TenCallsToNext() {
 	s.False(itr.HasNext())
 }
 
+func (s *HistoryBlobIteratorSuite) TestReadBlobEvents_Success_SameHistoryDifferentPage() {
```
andrewjdawson2016: Awesome.

service/worker/archiver/history_blob_iterator.go
yycptt: In my draft pr for recording upload progess using heartbeating, the state of the iterator will be a struct and I can make this function return that struct as part of its result. So how about I making the change in that diff?

andrewjdawson2016: Sounds good

andrewjdawson2016:
You have an interesting thought about deleting currPageToken and nextPageToken on the blob header. Why do you want to do this? My original reason for including these on the blob header was I thought it was a nice property that the keys for current blob and next blob could be constructed given only the blob itself. If we remove these fields then given a blob somewhere in code you cannot reconstruct the key.

yycptt:
Sorry. Ignore what I said. I thought they are persistence page token.

PR #1981, opened 2019-06-08
Archival blob integrity check

andrewjdawson2016:


service/worker/archiver/history_blob_downloader.go
```diff
@@ -0,0 +1,150 @@
...
+	b, err := d.blobstoreClient.Download(ctx, request.ArchivalBucket, key)
+	if err != nil {
+		return nil, err
+	}
+	unwrappedBlob, wrappingLayers, err := blob.Unwrap(b)
+	if err != nil {
+		return nil, err
+	}
+	historyBlob := &HistoryBlob{}
+	switch *wrappingLayers.EncodingFormat {
```
yycptt: How about adding a default case and returns an error? I understand that right now there's only one encoding, but if we add a new one and forget to make a change here, an error should be returned.

service/worker/archiver/history_blob_downloader.go
```diff
@@ -0,0 +1,150 @@
...
+	case blob.JSONEncoding:
+		if err := json.Unmarshal(unwrappedBlob.Body, historyBlob); err != nil {
+			return nil, err
+		}
+	}
+	if *historyBlob.Header.IsLast {
```
yycptt: nil check in case blob is corrupted? or use common.BoolDefault()?

service/worker/archiver/history_blob_downloader.go
```diff
@@ -0,0 +1,150 @@
...
+		token = nil
+	} else {
+		token.BlobstorePageToken = *historyBlob.Header.NextPageToken
```
yycptt: same here

service/worker/archiver/history_blob_downloader.go
```diff
@@ -0,0 +1,150 @@
...
+
+type (
+	// DownloadBlobRequest is request to DownloadBlob
+	DownloadBlobRequest struct {
+		NextPageToken        []byte
+		ArchivalBucket       string
+		DomainID             string
+		WorkflowID           string
+		RunID                string
+		CloseFailoverVersion *int64
```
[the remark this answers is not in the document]
yycptt: Oh, I see, since it's optional.

service/worker/archiver/activities.go
```diff
@@ -225,6 +232,39 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
+				return nil
+			}
+			for _, e := range resp.HistoryBlob.Body.Events {
+				fetchedHistoryEventHashes = append(fetchedHistoryEventHashes, hash(e.String()))
+			}
+			req.NextPageToken = resp.NextPageToken
+		}
+		if !hashesEqual(fetchedHistoryEventHashes, uploadedHistoryEventHashes) {
+			scope.IncCounter(metrics.ArchiverBlobIntegrityCheckFailedCount)
+			logger.Error("uploaded history does not match fetched history")
```
venkat1109: print additional context for for debugging - (wfID, runID, failoverVersion)

service/worker/archiver/client_worker.go
```diff
@@ -53,17 +53,20 @@ type (
...
+		Logger           log.Logger
+		ClusterMetadata  cluster.Metadata
+		HistoryManager   persistence.HistoryManager
+		HistoryV2Manager persistence.HistoryV2Manager
+		Blobstore        blobstore.Client
+		DomainCache      cache.DomainCache
+		Config           *Config
+
+		// the following are only set in testing code
+		HistoryBlobReader     HistoryBlobReader
```
venkat1109: if this is only used in testing, can you make these things private - assuming the tests that are using it are within the same package ?

service/worker/archiver/activities.go
andrewjdawson2016: Already included on the logger. We tag logger with all interesting fields at the start of the activity.

service/worker/archiver/client_worker.go
andrewjdawson2016: We cannot because they will then not get propagated as part of the context.

PR #2012, opened 2019-06-12
Refactor reset mutable state API

wxing1292:
* Supporting insertion of transfer / timer / replication task.
* Add more check on last write version and workflow state

### DIFF
--- a/common/persistence/persistenceInterface.go
+++ b/common/persistence/persistenceInterface.go
@@ -368,19 +368,28 @@ type (
 
 	// InternalResetMutableStateRequest is used to reset workflow execution state  for Persistence Interface
 	InternalResetMutableStateRequest struct {
-		PrevRunID        string
+		// previous workflow information
+		PrevRunID            string
+		PrevLastWriteVersion int64
+		PrevState            int
+
 		ExecutionInfo    *InternalWorkflowExecutionInfo
 		ReplicationState *ReplicationState
 		Condition        int64
 		RangeID          int64
 
-		// Mutable state
+		// mutable state pending info
 		InsertActivityInfos       []*InternalActivityInfo
 		InsertTimerInfos          []*TimerInfo
 		InsertChildExecutionInfos []*InternalChildExecutionInfo
 		InsertRequestCancelInfos  []*RequestCancelInfo
 		InsertSignalInfos         []*SignalInfo
 		InsertSignalRequestedIDs  []string
+
+		// replication/ transfer / timer task
+		InsertReplicationTasks []Task
+		InsertTransferTasks    []Task
+		InsertTimerTasks       []Task
 	}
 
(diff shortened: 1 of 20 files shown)


common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1880,23 +1885,92 @@ func createTimerTasks(
...
 }
 
-func assertAndUpdateCurrentExecution(tx sqldb.Tx, shardID int, domainID sqldb.UUID, workflowID string, newRunID sqldb.UUID, previousRunID sqldb.UUID,
+func assertNotCurrentExecution(tx sqldb.Tx, shardID int, domainID sqldb.UUID, workflowID string, runID sqldb.UUID) error {
+
+	assertFn := func(currentRow *sqldb.CurrentExecutionsRow) error {
+		if bytes.Equal(currentRow.RunID, runID) {
+			return &p.ConditionFailedError{Msg: fmt.Sprintf(
+				"Assertion on current record failed failed. Current run ID was %v, expected %v",
+				currentRow.RunID,
```
venkat1109: this log is misleading - expected and got will always be same - I think you want to return error saying that the records are expected to be different but they are same

common/persistence/persistenceInterface.go
```diff
@@ -368,19 +368,28 @@ type (
 
 	// InternalResetMutableStateRequest is used to reset workflow execution state  for Persistence Interface
 	InternalResetMutableStateRequest struct {
-		PrevRunID        string
+		// previous workflow information
+		PrevRunID            string
```
venkat1109: Does it make sense to create a separate persistence type for a "WorkflowRun" - that contains (runID, lastWriteVersion, state) ? Looks like these params are needed / passed everywhere for assertion.

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -1880,23 +1885,92 @@ func createTimerTasks(
 	return nil
 }
 
-func assertAndUpdateCurrentExecution(tx sqldb.Tx, shardID int, domainID sqldb.UUID, workflowID string, newRunID sqldb.UUID, previousRunID sqldb.UUID,
+func assertNotCurrentExecution(tx sqldb.Tx, shardID int, domainID sqldb.UUID, workflowID string, runID sqldb.UUID) error {
```
venkat1109: I don't see this being used anywhere, do you need this ?

wxing1292: this is from the 3+DC branch, probably leaked by rebase
i will remove this

common/persistence/persistenceInterface.go
wxing1292: created an issue for this: #2013

PR #2019, opened 2019-06-13
Fix int overflow in list api time filter

vancexu:
Currently when user use MaxInt64 in StartTimeFilter of ListOpenWorkflowExecutions ... APIs against ES, it will return no results because int overflow and ES not able to handle correctly process MaxInt64.

This PR fixed this issue.

### DIFF
--- a/common/persistence/elasticsearch/esVisibilityStore.go
+++ b/common/persistence/elasticsearch/esVisibilityStore.go
@@ -686,9 +686,16 @@ func (v *esVisibilityStore) getSearchResult(request *p.ListWorkflowExecutionsReq
 		rangeQuery = elastic.NewRangeQuery(es.CloseTime)
 	}
 	// ElasticSearch v6 is unable to precisely compare time, have to manually add resolution 1ms to time range.
+	// Also has to use string instead of int64 to avoid data conversion issue,
+	// 9223372036854775807 to 9223372036854776000 (long overflow)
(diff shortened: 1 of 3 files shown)


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -686,9 +686,16 @@ func (v *esVisibilityStore) getSearchResult(request *p.ListWorkflowExecutionsReq
 		rangeQuery = elastic.NewRangeQuery(es.CloseTime)
 	}
 	// ElasticSearch v6 is unable to precisely compare time, have to manually add resolution 1ms to time range.
+	// Also has to use string instead of int64 to avoid data conversion issue,
+	// 9223372036854775807 to 9223372036854776000 (long overflow)
+	if request.LatestStartTime > math.MaxInt64-oneMilliSecondInNano { // prevent latestTime overflow
+		request.LatestStartTime = math.MaxInt64 - oneMilliSecondInNano
+	}
```
wxing1292: additional check on the following?
1. EarliestStartTime <=LatestStartTime
2. 
```
if request. EarliestStartTime < math.MinInt64+oneMilliSecondInNano { // prevent earliestTime overflow
	request.EarliestStartTime = math.MinInt64+oneMilliSecondInNano
}
```

vancexu: updated

PR #2026, opened 2019-06-14
Add latency metrics for each visibility message to ES 

vancexu:
I also contains a small refactor of archival tests


common/metrics/fake.go
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: could you change this file name to noop.go?

service/worker/archiver/replay_metrics_client.go
```diff
@@ -67,7 +67,7 @@ func (r *replayMetricsClient) AddCounter(scope int, counter int, delta int64) {
 // StartTimer starts a timer for the given metric name. Time will be recorded when stopwatch is stopped.
 func (r *replayMetricsClient) StartTimer(scope int, timer int) tally.Stopwatch {
 	if workflow.IsReplaying(r.ctx) {
-		return nopStopwatch()
+		return metrics.NopStopwatch()
```
wxing1292: noop?

service/worker/indexer/esProcessor.go
```diff
@@ -64,6 +65,11 @@ type (
 		logger        log.Logger
 		metricsClient metrics.Client
 	}
+
+	kafkaMessageWithMetrics struct { // value of esProcessorImpl.mapToKafkaMsg
```
wxing1292: you can also make this struct implement messaging.Message and make the implemented ack / nack call stopwatch.stop

common/metrics/fake.go
vancexu: renamed

service/worker/archiver/replay_metrics_client.go
vancexu: FYI 

> In computer science, a NOP, no-op, or NOOP (pronounced "no op"; short for no operation) is an assembly language instruction, programming language statement, or computer protocol command that does nothing.

service/worker/indexer/esProcessor.go
vancexu: updated, I think your suggestion is cleaner

PR #2035, opened 2019-06-14
Add explanation about auto-reset

longquanzheng:


dmetzgar:
I think the reason this continues to build is because you create a branch in this repo. I filter with !gh-pages, but it doesn't have a way to filter by what the target repo is of a PR. So we'll just have to ignore build failures for gh-pages.

PR #2045, opened 2019-06-17
Fix null-check when creating s3 archival store

RikardLegge:
Found this bug whilst experimenting with cadence so no related issue exists.

With the initial implementation of the s3 support for archival, a few logical errors related to the existance of a `BlobstoreClient` where added causing nil de-references later.

What this PR changes is:
- BlobStore must exist if archival is enabled.
- If s3store can not be constructed, the error is passed out to the parent scope causing the error branch to run.
- Flip the error check when constructing the s3store to create a blob store if a s3 client could be constructed from the config, not when an error occurres.


cmd/server/server.go
```diff
@@ -191,14 +192,18 @@ func (s *server) startService() common.Daemon {
...
+			s3cli, err = s3store.ClientFromConfig(s.cfg.Archival.S3store)
+			if err == nil {
 				params.BlobstoreClient = s3store.NewClient(s3cli)
 			}
 		}
 		if err != nil {
 			log.Fatalf("error creating blobstore: %v", err)
 		}
+		if params.BlobstoreClient == nil {
+			log.Fatalf("no blobstore created, please add a filrestore or s3store configuration to archival")
```
shreyassrivatsan: nit, typo: `filestore`

cmd/server/server.go
```diff
@@ -191,14 +192,18 @@ func (s *server) startService() common.Daemon {
 			params.BlobstoreClient, err = filestore.NewClient(s.cfg.Archival.Filestore)
 		}
 		if s.cfg.Archival.S3store != nil {
-			s3cli, err := s3store.ClientFromConfig(s.cfg.Archival.S3store)
-			if err != nil {
+			var s3cli s3iface.S3API
+			s3cli, err = s3store.ClientFromConfig(s.cfg.Archival.S3store)
```
shreyassrivatsan: nit: `s3cli, err :=`, do need the var

cmd/server/server.go
```diff
@@ -191,14 +192,18 @@ func (s *server) startService() common.Daemon {
...
+			if err == nil {
 				params.BlobstoreClient = s3store.NewClient(s3cli)
 			}
 		}
 		if err != nil {
 			log.Fatalf("error creating blobstore: %v", err)
 		}
+		if params.BlobstoreClient == nil {
```
yycptt: I think we should perform this check (if neither fliestore nor s3 store is specified) on L190.

shreyassrivatsan: +1 to that.

cmd/server/server.go
RikardLegge: Fixed in 12c430e0172fd8446861626da6aec7f32b693f17

cmd/server/server.go
RikardLegge: Changde in 12c430e0172fd8446861626da6aec7f32b693f17

cmd/server/server.go
RikardLegge: Would that not shadow the err variable, making the later error check never run?
Might have misunderstood you comment, but this is what I'm basing my change on:
https://play.golang.org/p/QvMipO3_ov7

andrewjdawson2016: How about instead of worrying about this shadowing simply log.Fatal whenever you get an error, it also lets you have better error messages.

```
params.BlobstoreClient, err := filestore.NewClient(s.cfg.Archival.Filestore)
if err != nil {
    log.Fatal("failed to create file based blobstore" ...)
}
...
```

PR #2046, opened 2019-06-17
Update register domain command to include new required flag

longquanzheng:


longquanzheng:
> Thanks for doing this.
> 
> I ran into this issue myself, didn't know the docs where in `gh-pages`!

:p  And it has a nice website here: https://cadenceworkflow.io/docs/08_cli

PR #2047, opened 2019-06-17
3+DC version history / persistence layer refactor

wxing1292:
* Refactor version history adding more methods
* Add check when changing workflow state / close status


common/persistence/dataInterfaces.go
```diff
@@ -2310,6 +2317,52 @@ func (e *WorkflowExecutionInfo) GetCurrentBranch() []byte {
...
+		}
+		if state == WorkflowStateCompleted && closeStatus == WorkflowCloseStatusNone {
+			// TODO may want to further assert that close status being
+			//  WorkflowCloseStatusTerminated or
+			//  WorkflowCloseStatusTimedOut
+			return invalidErr
+		}
+
+	case WorkflowStateRunning:
+		if state == WorkflowStateCompleted && closeStatus == WorkflowCloseStatusNone {
```
venkat1109: looks like this state machine check is not exhaustive. Either make it exhaustive or another option is not to do the validation in persistence at all - since it's really the upper layer which understands these states. We just happen to define them here for convenience.

common/persistence/versionHistory.go
```diff
@@ -21,174 +21,513 @@
...
-	if len(items) == 0 {
-		panic("version history items cannot be empty")
+// NewVersionHistoryItem create a new version history item
+func NewVersionHistoryItem(
+	inputEventID int64,
+	inputVersion int64,
+) *VersionHistoryItem {
+
+	if inputEventID < 0 || inputVersion < 0 {
+		panic("invalid version history item event ID / version")
```
venkat1109: can you also emit the values of inputEventID and inputVersion for debugging ?

common/persistence/dataInterfaces.go
```diff
@@ -551,20 +558,20 @@ type (
...
+		eventID int64
+		version int64
 	}
 
 	// VersionHistory provides operations on version history
 	VersionHistory struct {
-		BranchToken []byte
-		History     []VersionHistoryItem
+		branchToken []byte
+		items       []*VersionHistoryItem
```
venkat1109: nit: I like the renaming but are there reasons to change this from slice of items to slice of pointer to items ? Given that each historyItem is two int64s, it is very cache friendly to have these as slice of items instead of pointers.

common/persistence/dataInterfaces.go
```diff
@@ -551,20 +558,20 @@ type (
...
 	}
 
 	// VersionHistories contains a set of VersionHistory
 	VersionHistories struct {
-		CurrentBranch int32
-		Histories     []VersionHistory
+		currentBranchIndex int
+		histories          []*VersionHistory
```
venkat1109: same question as above.

common/persistence/dataInterfaces.go
wxing1292: was think about the same reason when using slice of structs, i.e. cache friendly
the reason to change to pointer is, currently correctness is more important than performance, and the focus is to make the logic works end to end.

using a slice of struct maybe good for perf, but now (during the development), perf is not a major concern and i prefer correctness (less bug).

after the initial dev, i may end up changing this back

common/persistence/versionHistory.go
wxing1292: sure

common/persistence/dataInterfaces.go
wxing1292: good catch

common/persistence/dataInterfaces.go
```diff
@@ -143,6 +147,9 @@ const (
 	// TransferTaskTransferTargetRunID is the the dummy run ID for transfer tasks of types
 	// that do not have a target workflow
 	TransferTaskTransferTargetRunID = "30000000-0000-f000-f000-000000000002"
+
+	// indicate invalid workflow state transition
+	invalidStateTransitionMsg = "unable to change workflow state from %v to %v, close status %v"
```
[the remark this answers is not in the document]
wxing1292: sure

common/persistence/dataInterfaces.go
wxing1292: #2053

common/persistence/versionHistory.go
```diff
@@ -21,174 +21,517 @@
...
-	if len(items) == 0 {
-		panic("version history items cannot be empty")
+// NewVersionHistoryItem create a new version history item
+func NewVersionHistoryItem(
+	inputEventID int64,
+	inputVersion int64,
+) *VersionHistoryItem {
+
+	if inputEventID < 0 || inputVersion < 0 {
+		panic(fmt.Sprintf(
```
meiliang86: You have plans to change this and other panics to return error in future, right?

common/persistence/versionHistory.go
```diff
@@ -21,174 +21,517 @@
...
+	return nil, notFoundErr
+}
+
+// SetBranchToken the set the branch token
+func (v *VersionHistory) SetBranchToken(
+	inputToken []byte,
+) error {
+
+	if len(v.branchToken) != 0 {
+		return &shared.BadRequestError{Message: "cannot branch token is already set"}
```
meiliang86: message does not read...

common/persistence/versionHistory.go
```diff
@@ -21,174 +21,517 @@
...
+	token := make([]byte, len(inputToken))
+	copy(token, inputToken)
+	v.branchToken = token
+	return nil
+}
+
+// GetBranchToken return the branch token
+func (v *VersionHistory) GetBranchToken() []byte {
+	token := make([]byte, len(v.branchToken))
+	copy(token, v.branchToken)
```
meiliang86: Why copying this everywhere?

common/persistence/versionHistory.go
```diff
@@ -21,174 +21,517 @@
...
-func (v *VersionHistory) IsAppendable(item VersionHistoryItem) bool {
-	if len(v.History) == 0 {
-		panic("version history cannot be empty")
+// IsLCAAppendable checks if a LCA version history item is appendable
+func (v *VersionHistory) IsLCAAppendable(
+	item *VersionHistoryItem,
+) bool {
+
+	if len(v.items) == 0 {
+		panic("version history not initialized")
```
meiliang86: Way too many panics. Maybe you should start checking errors now...

common/persistence/versionHistory.go
wxing1292: i need to think about this.

based on my current assumption
event ID < 0 is bug, should fail fast
version == common.EmptyVersion ( < 0 ) can happen for local workflow, however, this is not supported yet

common/persistence/versionHistory.go
wxing1292: good catch

common/persistence/versionHistory.go
wxing1292: i want to make this immutable, to achieve less bug, perf is not a concern, for now

common/persistence/versionHistory.go
wxing1292: errors should be something that can happen,
panics is for definite code bug,

this is my current assumption

wxing1292: but good point, too many panics

common/persistence/dataInterfaces.go
```diff
@@ -2310,6 +2317,146 @@ func (e *WorkflowExecutionInfo) GetCurrentBranch() []byte {
 	return e.BranchToken
 }
 
+// UpdateWorkflowStateCloseStatus update the workflow state
+func (e *WorkflowExecutionInfo) UpdateWorkflowStateCloseStatus(
```
venkat1109: consider moving this to a separate go file. datainterfaces.go was originally intended to just have the interfaces defined. This is more than a simple get/set method for this file.

wxing1292: sure, will do in the next PR

Issue #2042, opened 2019-06-14
Cross DC - Add addition check about cross domain calls


wxing1292:
We should currently block the followings:
1. global domain -> local domain
2. local domain -> global domain
3. global domain -> global domain if domains are not the same

PR #2048, opened 2019-06-17
Disable not-well-defined decision cross domain call

wxing1292:
* Local domain to local domain cross domain call is still valid
* Local domain to global domain cross domain call is disabled
* Global domain to itself cross domain call is still valid
* Global domain to local domain cross domain call is disabled
* Global domain to other global domain cross domain call is disabled
* Affected decisions are
  * ScheduleActivity
  * CancelExternalWorkflowExecution
  * SignalExternalWorkflowExecution
  * StartChildExecution

solve #2042

### DIFF
--- a/service/history/decisionChecker.go
+++ b/service/history/decisionChecker.go
@@ -110,9 +110,18 @@ func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(
 }
 
 func (v *decisionAttrValidator) validateActivityScheduleAttributes(
+	domainID string,
+	targetDomainID string,
 	attributes *workflow.ScheduleActivityTaskDecisionAttributes,
 	wfTimeout int32,
 ) error {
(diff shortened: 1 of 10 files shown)


service/history/decisionChecker.go
```diff
@@ -452,3 +462,53 @@ func (v *decisionAttrValidator) validateStartChildExecutionAttributes(
 
 	return nil
 }
+
+func (v *decisionAttrValidator) validateCrossDomainCall(
```
venkat1109: Are there reasons to expose new method ? Can you not automatically embed this validation within the existing validateFooAttributes() methods ? I like the fact that there is one validate method for a given attribute type. That way, caller code doesn't have to change when we add new types of validations for the same set of attributes.

wxing1292: move this check to within corresponding decision check

PR #2049, opened 2019-06-17
Add docker compose for Cadence on ElasticSearch

vancexu:
Create docker `ubercadence/server-es:master` for people to use elasticsearch features like search attributes and List/Scan/CountWorkflow.

Usage:
```
cd docker
docker-compose -f docker-compose-es.yml up
```


docker/Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN go get -u golang.org/x/lint/golint
 
 # get and compile cadence-server
 RUN git clone https://github.com/uber/cadence.git /go/src/github.com/uber/cadence
-RUN cd /go/src/github.com/uber/cadence && git checkout $git_branch && make bins_nothrift
+RUN cd /go/src/github.com/uber/cadence && git fetch && git checkout $git_branch && make bins_nothrift
```
wxing1292: `RUN git clone https://github.com/uber/cadence.git /go/src/github.com/uber/cadence`
should already pull the newest change?

docker/Dockerfile
```diff
@@ -57,6 +58,8 @@ EXPOSE 7933 7934 7935 7939
 COPY ./start.sh $CADENCE_HOME/start.sh
 COPY ./config_template.yaml $CADENCE_HOME/config/docker_template_cassandra.yaml
 COPY ./config_template_mysql.yaml $CADENCE_HOME/config/docker_template_mysql.yaml
+RUN mkdir $CADENCE_HOME/config/dynamicconfig/
```
wxing1292: `mkdir -p`

docker/Dockerfile
vancexu: need fetch before checkout to branch other than master.

PR #2050, opened 2019-06-17
Add metrics to DC redirection handler

wxing1292:
solve #2041

### DIFF
--- a/service/frontend/dcRedirectionHandler.go
+++ b/service/frontend/dcRedirectionHandler.go
@@ -22,6 +22,7 @@ package frontend
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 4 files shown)


service/frontend/dcRedirectionHandler.go
```diff
@@ -885,3 +1038,17 @@ func (handler *DCRedirectionHandlerImpl) TerminateWorkflowExecution(
 
 	return err
 }
+
+func (handler *DCRedirectionHandlerImpl) beforeCall() time.Time {
+	return handler.timeSource.Now()
```
vancexu: Can timeSource be nil?

wxing1292: this is initialized in the new function

PR #2054, opened 2019-06-18
Acquire domain ID & workflow ID lock before creating a new workflow

wxing1292:
* Add getOrCreatCurrentWorkflowExecution for current workflow execution lock
* Start workflow execution / signal with start workflow execution, before creating the workflow, will lock on current workflow execution, i.e. run ID being ""
* Rename getOrCreateWorkflowExecutionWithTimeout -> getOrCreateWorkflowExecution
* Rename getOrCreateWorkflowExecution -> getOrCreateWorkflowExecutionForBackground
* Reformat workflow execution context

### DIFF
--- a/service/history/historyCache.go
+++ b/service/history/historyCache.go
@@ -78,6 +78,11 @@ func (c *historyCache) getOrCreateCurrentWorkflowExecution(
 	workflowID string,
 ) (workflowExecutionContext, releaseWorkflowExecutionFunc, error) {
 
+	scope := metrics.HistoryCacheGetOrCreateCurrentScope
+	c.metricsClient.IncCounter(scope, metrics.CacheRequests)
+	sw := c.metricsClient.StartTimer(scope, metrics.CacheLatency)
+	defer sw.Stop()
+
 	// using empty run ID as current workflow run ID
 	runID := ""
 	execution := workflow.WorkflowExecution{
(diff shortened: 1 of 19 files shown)


service/history/historyCache.go
```diff
@@ -72,38 +72,65 @@ func newHistoryCache(shard ShardContext) *historyCache {
...
+			c.shard,
+			c.executionManager,
+			c.logger,
+		), func(error) {}, nil
+	}
 
+	key := definition.NewWorkflowIdentifier(domainID, workflowID, runID)
+	workflowCtx, cacheHit := c.Get(key).(workflowExecutionContext)
+	if !cacheHit {
+		c.metricsClient.IncCounter(metrics.HistoryCacheGetOrCreateScope, metrics.CacheMissCounter)
```
venkat1109: scope name should be different given that this is a different method

service/history/historyCache.go
```diff
@@ -72,38 +72,65 @@ func newHistoryCache(shard ShardContext) *historyCache {
 	}
 }
 
-func (c *historyCache) getOrCreateWorkflowExecution(domainID string,
-	execution workflow.WorkflowExecution) (workflowExecutionContext, releaseWorkflowExecutionFunc, error) {
-	return c.getOrCreateWorkflowExecutionWithTimeout(context.Background(), domainID, execution)
-}
+func (c *historyCache) getOrCreatCurrentWorkflowExecution(
```
venkat1109: this func is mostly the same as getOrCreateWorkflowExecution() except for a flag and another validation call. Consider factoring out the common pieces into a helper method that both of these  can call.

service/history/historyCache.go
wxing1292: good catch

service/history/historyCache.go
wxing1292: sure

PR #2069, opened 2019-06-19
Limit archival retry timeout & Improve blob downloader error msg

yycptt:
https://github.com/uber/cadence/issues/2040
https://github.com/uber/cadence/issues/2037

### DIFF
--- a/service/worker/archiver/archiver.go
+++ b/service/worker/archiver/archiver.go
@@ -103,12 +103,12 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
 	sw := metricsClient.StartTimer(metrics.ArchiverScope, metrics.ArchiverHandleRequestLatency)
 	logger = tagLoggerWithRequest(logger, request)
 	ao := workflow.ActivityOptions{
-		ScheduleToStartTimeout: 10 * time.Minute,
-		StartToCloseTimeout:    5 * time.Minute,
+		ScheduleToStartTimeout: 2 * time.Minute,
+		StartToCloseTimeout:    2 * time.Minute,
 		RetryPolicy: &cadence.RetryPolicy{
 			InitialInterval:          time.Second,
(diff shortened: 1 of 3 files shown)


service/worker/archiver/archiver.go
```diff
@@ -103,12 +103,12 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
 	ao := workflow.ActivityOptions{
-		ScheduleToStartTimeout: 10 * time.Minute,
-		StartToCloseTimeout:    5 * time.Minute,
+		ScheduleToStartTimeout: 2 * time.Minute,
+		StartToCloseTimeout:    2 * time.Minute,
 		RetryPolicy: &cadence.RetryPolicy{
 			InitialInterval:          time.Second,
 			BackoffCoefficient:       2.0,
-			ExpirationInterval:       10 * time.Minute,
+			ExpirationInterval:       4 * time.Minute,
```
andrewjdawson2016: Seems fine, we can tune this if we find a need. But I think this is plenty long.

service/worker/archiver/history_blob_downloader.go
```diff
@@ -62,6 +63,12 @@ type (
 	}
 )
 
+var (
+	errArchivalTokenCorrupted      = &gen.BadRequestError{Message: "Next page token is corrupted."}
+	errRequestedHistoryNotArchived = &gen.BadRequestError{Message: "Requested workflow history has not been archived yet."}
```
andrewjdawson2016: I know these are two different errors and we understand how they are different but I suspect the difference is not meaningful to most people reading the errors. How about we combine these errors into one error?

service/worker/archiver/archiver.go
yycptt: If we want to use a lower value, we also need to decrease schedule to start timeout.

PR #2077, opened 2019-06-20
Archival: refactor upload activity

yycptt:
https://github.com/uber/cadence/issues/2030


service/worker/archiver/activities.go
```diff
@@ -97,229 +96,7 @@ const (
...
-			}
-			req.NextPageToken = resp.NextPageToken
-		}
-		if !hashesEqual(fetchedHistoryEventHashes, uploadedHistoryEventHashes) {
-			scope.IncCounter(metrics.ArchiverBlobIntegrityCheckFailedCount)
-			logger.Error("uploaded history does not match fetched history")
-		}
-	}
-	return result, nil
+	return NewHistoryBlobUploader().UploadHistory(ctx, &request)
```
andrewjdawson2016: beautiful - I like this approach so much more than before.

PR #2081, opened 2019-06-21
Disable workflow execution first batch deletion

wxing1292:
* Do not delete the first event batch if persistence of mutable state fails
* Record context deadline exceeds / cancelled error as context timeout metrics


wxing1292:
> Will this change cause event leak?

yes, the thing is, we cannot clearly tell which events can be safely deleted.

PR #2083, opened 2019-06-21
Header propagation for cron jobs and retries

shreyassrivatsan:


shreyassrivatsan:
> Should we add a unit test for this?

Yeah. Added one for the cron code path.. Wasn't there before.

PR #2086, opened 2019-06-22
Refactor domain processing logic, with more check on replication config

wxing1292:
solves #2084


service/frontend/domainAttrValidator.go
```diff
@@ -0,0 +1,153 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package frontend
+
+import (
+	"fmt"
+
+	"github.com/uber/cadence/common/cluster"
+
```
yycptt: nit: can we delete these empty lines in this and in other files.

PR #2096, opened 2019-06-24
schema-tool: do not error when schema is already up-to-date

venkat1109:
Fixes https://github.com/uber/cadence/issues/2087

### DIFF
--- a/tools/common/schema/updatetask_test.go
+++ b/tools/common/schema/updatetask_test.go
@@ -44,6 +44,10 @@ func (s *UpdateTaskTestSuite) SetupSuite() {
 
 func (s *UpdateTaskTestSuite) TestReadSchemaDir() {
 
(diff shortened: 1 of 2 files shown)


tools/common/schema/updatetask_test.go
```diff
@@ -57,6 +61,16 @@ func (s *UpdateTaskTestSuite) TestReadSchemaDir() {
 	s.NotNil(err)
 	_, err = readSchemaDir(tmpDir, "0.5", "10.3")
 	s.NotNil(err)
+	_, err = readSchemaDir(tmpDir, "1.5", "1.5")
+	s.NotNil(err)
```
shreyassrivatsan: nit: use `Error` and `NoError`

sagikazarmark:
Is there a chance to release this as a hotfix?

PR #2101, opened 2019-06-25
Guarantee that timer fire after schedule time + timeout time

wxing1292:


service/history/timerBuilder.go
```diff
@@ -222,10 +222,26 @@ func (tb *timerBuilder) GetUserTimer(timerID string) (bool, *persistence.TimerIn
...
+	expireTimeNanoSeconds := expireTime % 1e9
+
+	referenceTime := reference.UnixNano()
+	referenceTimeSeconds := referenceTime / 1e9
+	referenceTimeNanoSeconds := referenceTime % 1e9
+
+	// do a tiny sleep here to guarantee timer fire after schedule time + timeout time
+	if expireTimeSeconds <= referenceTimeSeconds {
+		if expireTimeNanoSeconds > referenceTimeNanoSeconds {
+			time.Sleep(time.Nanosecond * time.Duration(expireTimeNanoSeconds-referenceTimeNanoSeconds))
```
shreyassrivatsan: Can you check what the recommended minimum for `time.Sleep` is? Super short sleeps are not always handled well..

shreyassrivatsan: Somewhat related, but take a look at this - https://github.com/golang/go/issues/17696.. Windows minimum timer duration is 1ms or 15.6ms.. I am not sure what Linux exposes to users.. I remember reading for super short nanosecond sleeps its better to do (cannot find reference now).. 

```
while time.Now() < expTime {
    yield()
}
```

wxing1292: did 2 tests, one on mac, one on linux (4.something)
time interval on sleep 1 nano second produces double digits micro seconds interval.

just keep this PR open, in case java side needs is.

Issue #2095, opened 2019-06-24
Add documentation for the go-client session framework


yycptt:
Add the documentation for session to cadenceworkflow.io

PR #2105, opened 2019-06-25
Add Session documentation

yycptt:
#2095

### DIFF
--- a/docs/03_goclient/15_sessions.md
+++ b/docs/03_goclient/15_sessions.md
@@ -1,0 +1,134 @@
+# Sessions
+
+The session framework provides a straight-forward interface for scheduling multiple activities on a single worker without having users manually specify the task list name. It also includes features like **concurrent session limitation** and **worker failure detection**.
+
+## Use Cases
+
+- **File Processing**: User may want to implement a workflow that can download a file, process it and then upload the modified version. If these three steps are implemented as three different activities, all of them should be executed by the same worker. This is a perfect use case of the session framework.
+
+- **Machine Learning Model Training**: Similar to file processing, training a machine learning model also involves three stages: download the data set, optimize the model and upload the trained parameter. However, as each model may consume a large amount of GPU memory, the number of concurrent models being trained should be limited. If the training of each model is implemented as a session, this requirement can easily be achieved by using the concurrent session limitation feature.
+
+## Basic Usage
+
+Before using the session framework to write your workflow code, your worker needs to be configured to process sessions. To do that, set the `EnableSessionWorker` field of `worker.Options` to `true` when starting your worker.
+
+The most important APIs provided by the session framework are `workflow.CreateSession()` and `workflow.CompleteSession()`. The basic idea is that all the activities executed within a session will be processed by the same worker and these two APIs allow users to create new sessions and close them after all activities finish execution.
+
+Heres a more detailed description of these two APIs:
+```go
+type SessionOptions struct {
+  // ExecutionTimeout: required, no default
+  //     Specifies the maximum amount of time the session can run
+  ExecutionTimeout time.Duration
+
+  // CreationTimeout: required, no default
+  //     Specifies how long session creation can take before returning an error
+  CreationTimeout  time.Duration
+}
+
+func CreateSession(ctx Context, sessionOptions *SessionOptions) (Context, error)
+```
+
+`CreateSession()` takes in workflow.Context, sessionOptions and returns a new session context. When it's called, it will check the task list name specified in the ActivityOptions (or the one specified in StartWorkflowOptions) and create the session on one of the workers which is polling that task list. `CreateSession()` will return an error if the context passed in already contains an open session or all the workers are currently busy and unable to handle new sessions (check the **Concurrent Session Limitation** section for more details).
+
+The returned session context should be used to execute all activities belonging to the session. The context will be cancelled if the worker executing this session dies or `CompleteSession()` is called. When using the returned session context to execute activities, a `workflow.ErrSessionFailed` error may be returned if the session framework detects that the worker executing this session has died. The failure of user activities won't affect the state of the session, so user still needs to handler the errors returned from their activites and call `CompleteSession()` if necessary.
+
+```go
+func CompleteSession(ctx Context)
+```
+
+`CompleteSession()` takes in a session context and closes it. When it's called, the session context will be canceled (means all user activities using that session context will also be canceled) and the resources reserved at the worker will be released, so make sure `CompleteSession()` is called when you no longer need the session. As it's safe to call `CompleteSession()` on a failed session, typical usage is running it as a `defer` function after a session is successfully created.
+
+### Sample Code
+
+```go
+func FileProcessingWorkflow(ctx workflow.Context, fileID string) (err error) {
+  ao := workflow.ActivityOptions{
+    ScheduleToStartTimeout: time.Second * 5,
+    StartToCloseTimeout:    time.Minute,
+  }
+  ctx = workflow.WithActivityOptions(ctx, ao)
+
+  so := &workflow.SessionOptions{
+    CreationTimeout:  time.Minute,
+    ExecutionTimeout: time.Minute,
+  }
+  sessionCtx, err := workflow.CreateSession(ctx, so)
+  if err != nil {
+    return err
+  }
+  defer workflow.CompleteSession(sessionCtx)
+
+  var fInfo *fileInfo
+  err = workflow.ExecuteActivity(sessionCtx, downloadFileActivityName, fileID).Get(sessionCtx, &fInfo)
+  if err != nil {
+    return err
+  }
+
+  var fInfoProcessed *fileInfo
+  err = workflow.ExecuteActivity(sessionCtx, processFileActivityName, *fInfo).Get(ctx, &fInfoProcessed)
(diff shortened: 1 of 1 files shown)


docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+  // The hostname of the worker that is executing the session
+  HostName          string
+
+  // ... other unexported fields
+}
+
+func GetSessionInfo(ctx Context) *SessionInfo
+```
+
+The session context also stores some metadata of the session, which can be retrieved by the `GetSessionInfo()` API. For now, there are only two exported fields in the returned `workflow.SessionInfo`. We may consider exposing more information in the future.
```
dmetzgar: I would remove the "for now... in the future" part as we'll forget that this exists in the documentation esp if there's no issue tracking it. Just talk about the fields that are available.

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Recreate Session
+
+For long-running sessions, user may want to use the ContinueAsNew feature to split the workflow into multiple runs, but still needs all the activities to be executed by the same worker. The `RecreateSession()`  API is designed for such use case.
+
+```go
+func RecreateSession(ctx Context, recreateToken []byte, sessionOptions *SessionOptions) (Context, error)
+```
+
+Its usage is the same as `CreateSession()` except that it also takes in a `recrateToken`, which is needed to create a new session on the same worker as the previous one. User can get the token by calling the `GetRecreateToken()` method of the `SessionInfo` object and pass it to the next run.
```
dmetzgar: Fix spelling of "`recrateToken`"

dmetzgar: "User can" -> "You can"

dmetzgar: "and pass it to the next run" seems like it can be removed.

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+```go
+token := workflow.GetSessionInfo(sessionCtx).GetRecrateToken()
```
dmetzgar: Fix spelling of `GetRecrateToken`

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+```
+
+## Q&A
+
+### Is there a complete sample code?
+Yes, the fileprocessing example in the cadence-sample repo has been updated to use the session framework.
```
dmetzgar: Add a link to the sample repo. Even better, link directly to the file processing example.

dmetzgar: "fileprocessing" -> "file processing"

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+### What happens to my activity if worker dies?
```
dmetzgar: "if **the** worker dies"

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+If your activity has already been scheduled, it will be cancelled. If not, you will get an `workflow.ErrSessionFailed` error when you call `workflow.ExecuteActivity()`.
+
+### Is the concurrent session limitation per process or per host?
+It's per worker process, so make sure there's only one worker process running on the host if you plan to use that feature.
+
+
+## Future Work
+
+* **Support automatic session re-establishing**:
```
dmetzgar: Add links to issues for these bullet points

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+  Right now a session is considered failed if the worker process dies. However, for some use cases, user only cares if the worker host is alive or not. For these uses cases, we should automatically re-establish the session if the worker process is restarted.
```
dmetzgar: If you want this to show up on the next line, try adding three spaces at the end of the previous line

dmetzgar: "However, for some use case, **you may** only **care**"

dmetzgar: "we should" -> "you should"

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+* **Support fine-grained concurrent session limitation**:
+  The current implementation assumes that all sessions are consuming the same type of resource and there's only one global limitation. In the future, we may allow users to specify what type of resource their session will consume and enforce different limitations on different type of resources.
```
dmetzgar: "In the future, we may allow users" -> "Our plan is to allow you"

"their session" -> "your session"

"different type" -> "different types"

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,134 @@
...
+
+  // CreationTimeout: required, no default
+  //     Specifies how long session creation can take before returning an error
+  CreationTimeout  time.Duration
+}
+
+func CreateSession(ctx Context, sessionOptions *SessionOptions) (Context, error)
+```
+
+`CreateSession()` takes in workflow.Context, sessionOptions and returns a new session context. When it's called, it will check the task list name specified in the ActivityOptions (or the one specified in StartWorkflowOptions) and create the session on one of the workers which is polling that task list. `CreateSession()` will return an error if the context passed in already contains an open session or all the workers are currently busy and unable to handle new sessions (check the **Concurrent Session Limitation** section for more details).
```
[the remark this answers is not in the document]
yycptt: It will keep retrying until the user specified creation timeout.

mfateev: I find this confusing. It says that it takes workflow.Context and returns session context. I assumed that session context is a different type. I think we nee to rewrite this sentence.

docs/03_goclient/15_sessions.md
mfateev: They weren't: https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/fileprocessing/workflow.go

yycptt: I think they will by the time I land the documentation. I just don't have the permission to land the change in the samples repo.

docs/03_goclient/15_sessions.md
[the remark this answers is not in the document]
dmetzgar: You should say this in the documentation.

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,136 @@
...
+
+Its usage is the same as `CreateSession()` except that it also takes in a `recreateToken`, which is needed to create a new session on the same worker as the previous one. You can get the token by calling the `GetRecreateToken()` method of the `SessionInfo` object.
+
+```go
+token := workflow.GetSessionInfo(sessionCtx).GetRecreateToken()
+```
+
+## Q&A
+
+### Is there a complete sample code?
```
purplesusan: remove "code" (Is there a complete example?)

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,136 @@
...
+Yes, the [file processing example](https://github.com/samarabbas/cadence-samples/blob/master/cmd/samples/fileprocessing/workflow.go) in the cadence-sample repo has been updated to use the session framework.
+
+### What happens to my activity if the worker dies?
+If your activity has already been scheduled, it will be cancelled. If not, you will get an `workflow.ErrSessionFailed` error when you call `workflow.ExecuteActivity()`.
```
purplesusan: get an -> get a

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Is the concurrent session limitation per process or per host?
+It's per worker process, so make sure there's only one worker process running on the host if you plan to use that feature.
+
+
+## Future Work
+
+* **[Support automatic session re-establishing](https://github.com/uber-go/cadence-client/issues/775)**   
+Right now a session is considered failed if the worker process dies. However, for some use cases, you may only care if the worker host is alive or not. For these uses cases, session should be automatically re-established if the worker process is restarted.
```
purplesusan: care if the -> care whether
cases, session -> cases, the session

docs/03_goclient/15_sessions.md
```diff
@@ -0,0 +1,136 @@
...
+  defer workflow.CompleteSession(sessionCtx)
+
+  var fInfo *fileInfo
+  err = workflow.ExecuteActivity(sessionCtx, downloadFileActivityName, fileID).Get(sessionCtx, &fInfo)
+  if err != nil {
+    return err
+  }
+
+  var fInfoProcessed *fileInfo
+  err = workflow.ExecuteActivity(sessionCtx, processFileActivityName, *fInfo).Get(ctx, &fInfoProcessed)
```
dmetzgar: For these three activities, you use `sessionCtx` for the Get method on the first one, but `ctx` for the other two. Is this correct?

yycptt: Thanks for the catch. It's my fault. I don't think it will cause any problem as I have tested the code locally. But it's always good to keep it consistent and avoid any confusion. I will update soon. Thanks.

Issue #2088, opened 2019-06-24
Archival: Add Archiver Interface


yycptt:
Task 1 of https://github.com/uber/cadence/issues/2082

Create Archiver Interface which looks like the following:
```
HistoryArchiver interface {
	Archive(context, URI, Request) error
	Get(context, URI, Request) (Response, error)
	Describe(context, Request) (Response, error)
}
VisibilityArchiver interface {
	Archive(context, URI, Request) error
	Query(context, URI, Request) (Response, error)
	Describe(context, Request) (Response, error)
}
```
 
Also plumb the interface through the code base (bootstrap param, frontend, timer queue processor, archiver worker)

PR #2106, opened 2019-06-25
Add Archiver interface

yycptt:
#2088

### DIFF
--- a/common/archiver/options.go
+++ b/common/archiver/options.go
@@ -1,0 +1,78 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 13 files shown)


service/frontend/service.go
```diff
@@ -200,7 +200,7 @@ func (s *Service) Start() {
 	}
 
 	metricsBlobstore := blobstore.NewMetricClient(params.BlobstoreClient, base.GetMetricsClient())
-	wfHandler := NewWorkflowHandler(base, s.config, metadata, history, historyV2, visibility, kafkaProducer, metricsBlobstore)
+	wfHandler := NewWorkflowHandler(base, s.config, metadata, history, historyV2, visibility, kafkaProducer, metricsBlobstore, params.HistoryArchivers, params.VisibilityArchivers)
```
yycptt: Another idea is adding these two variables to serviceImpl and add new methods to the service interface to return them.

common/archiver/options.go
```diff
@@ -0,0 +1,78 @@
...
+	catalog := &ArchiveFeatureCatalog{}
+	for _, opt := range opts {
+		opt(catalog)
+	}
+	return catalog
+}
+
+// GetHeartbeatArchiveOption returns an ArchiveOption for enabling heartbeating.
+// It should be used when the Archive method is invoked inside an activity.
+func GetHeartbeatArchiveOption() ArchiveOption {
```
andrewjdawson2016: Naming is a bit odd here in that the option is called HeartbeatArchive but but the thing you add to the catalog is called ProgressManager. I think these names should be consistent with each other.

common/archiver/options.go
```diff
@@ -0,0 +1,78 @@
...
+	return func(catalog *ArchiveFeatureCatalog) {
+		catalog.ProgressManager = &heartbeatProgessManager{}
+	}
+}
+
+type heartbeatProgessManager struct{}
+
+func (h *heartbeatProgessManager) RecordProgress(ctx context.Context, progress interface{}) error {
```
andrewjdawson2016: Why does this method return an error if it always returns nil?

service/frontend/service.go
andrewjdawson2016: I like the current approach. It seems like they should follow the same approach as existing items like kafkaProducer.

common/archiver/options.go
yycptt: I think other implementations of the Progress Manager interface may return an error.

common/archiver/options.go
yycptt: I think heartbeating is just one way of managing progress. Different implementations can have different implementation of progress manager, so I think it will be better to show the fact that this option is using heartbeating in the function name. 

I agree that the name should indicate the thing added to the catalog is ProgressManager, but I can't come up with a good name. Maybe "GetHeartbeatProgressManagingArchiveOption()"? Too long?

PR #2116, opened 2019-06-26
Refactor Cassandra / MySQL persistence layer

wxing1292:
* Add dedicated handling for WorkflowMutation and WorkflowSnapshot
* Move Cassandra util code into dedicated file


common/persistence/cassandra/cassandraPersistenceUtil.go
```diff
@@ -0,0 +1,1709 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
venkat1109: fix copyright - should be 2019 for this new file

common/persistence/cassandra/cassandraPersistenceUtil.go
```diff
@@ -0,0 +1,1709 @@
...
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
```
venkat1109: can you just name this file util.go ? cassandra/cassandraPersistenceUtil.go is redundant

wxing1292: this is only for execution table, i.e. cassandraPersistence's helper function, not visibility

wxing1292: btw, after this, i plan to also refactor the remaining ones, i.e. create workflow / continue as new and reset workflow execution, before 3DC reset mutable state change

PR #2119, opened 2019-06-27
Support DNS mode for public client

longquanzheng:


client/clientBean.go
```diff
@@ -195,3 +219,143 @@ func (p *ipDispatcherProvider) Get(name string, address string) (*yarpc.Dispatch
...
+	dispatcher := yarpc.NewDispatcher(yarpc.Config{
+		Name: crossDCCaller,
+		Outbounds: yarpc.Outbounds{
+			serviceName: transport.Outbounds{
+				Unary:       outbound,
+				ServiceName: serviceName,
+			},
+		},
+	})
+	err = dispatcher.Start()
```
shreyassrivatsan: nit: 

`if err = dispa... ; err != nil {`

client/clientBean.go
```diff
@@ -195,3 +219,143 @@ func (p *ipDispatcherProvider) Get(name string, address string) (*yarpc.Dispatch
...
+		adr := fmt.Sprintf("%v:%v", ip, d.port)
+		newPeers[adr] = struct{}{}
+	}
+
+	updates := &peer.ListUpdates{
+		Additions: make([]peer.Identifier, 0),
+		Removals:  make([]peer.Identifier, 0),
+	}
+	changed := false
+	//remove if it doesn't exist anymore
```
shreyassrivatsan: nit: space on comments after `//`

cmd/server/server.go
```diff
@@ -152,7 +151,26 @@ func (s *server) startService() common.Daemon {
 		s.cfg.Archival.DefaultBucket,
 		enableReadFromArchival(),
 	)
-	params.DispatcherProvider = client.NewIPYarpcDispatcherProvider()
+
+	// TODO this DispatcherProvider will also be used for clientBean to create dispatchers for remoteCluster in CrossDC re-replication.
```
shreyassrivatsan: When adding `TODO` either point to an issue or point to `@<githubid>` otherwise these are sure to be orphaned.

client/clientBean.go
```diff
@@ -195,3 +219,143 @@ func (p *ipDispatcherProvider) Get(name string, address string) (*yarpc.Dispatch
...
+			} else {
+				d.logger.Info("No change in DNS lookup", tag.Address(d.dnsAddress))
+			}
+			sleepDu := now.Add(d.interval).Sub(now)
+			time.Sleep(sleepDu)
+		}
+	}()
+}
+
+func (d *dnsUpdater) refresh() (map[string]struct{}, *peer.ListUpdates, bool, error) {
```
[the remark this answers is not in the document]
longquanzheng: I see. I think I can break that refresh() into two smaller functions to limit to 2 returned values

common/service/config/config.go
```diff
@@ -329,9 +329,14 @@ type (
 	}
 
 	// PublicClient is config for connecting to cadence frontend
+	// Currently supporting single IP address(HostPort) and DNS and Port. Only one of them is required.
 	PublicClient struct {
 		// HostPort is the host port to connect on
 		HostPort string `yaml:"hostPort" validate:"nonzero"`
```
sagikazarmark: What happens if I don't configure hostPort? Will the validation fail?

common/service/config/config.go
```diff
@@ -329,9 +329,14 @@ type (
...
+		// DNS is the DNS name to connect on
+		DNSPort string `yaml:"DNSPort" validate:"nonzero"`
```
sagikazarmark: Note that current the name of the config in the file is `DNSPort`.

common/service/config/config.go
```diff
@@ -329,9 +329,14 @@ type (
 	}
 
 	// PublicClient is config for connecting to cadence frontend
+	// Currently supporting single IP address(HostPort) and DNS and Port. Only one of them is required.
 	PublicClient struct {
 		// HostPort is the host port to connect on
 		HostPort string `yaml:"hostPort" validate:"nonzero"`
+		// DNS is the DNS name to connect on
```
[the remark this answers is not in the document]
longquanzheng: Actually, putting IP address in the dns host name also works. I think we can remove the old config to make it simpler. And we don't have that TODO issue anymore.

client/clientBean.go
longquanzheng: done all comments

cmd/server/server.go
longquanzheng: not applicable anymore

common/service/config/config.go
longquanzheng: The new code also works with IP address, so I would keep using the same config of hostport and let it non-zero

common/service/config/config.go
longquanzheng: I just update the code, No DNSPort anymore.

client/clientBean.go
```diff
@@ -44,7 +42,8 @@ import (
 )
 
 const (
-	ipPortRegex = `\b(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9]):[1-9]\d*\b`
+	ipPortRegex            = `\b(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9]):[1-9]\d*\b`
```
shreyassrivatsan: use something standard like this for IP validation https://golang.org/pkg/net/#ParseIP .. 

I am not sure if there ip:port validation, but the string can be split and worked with..

client/clientBean.go
```diff
@@ -184,44 +185,11 @@ func (h *clientBeanImpl) GetRemoteFrontendClient(cluster string) frontend.Client
...
-	err = dispatcher.Start()
-	if err != nil {
-		return nil, err
-	}
-	return dispatcher, nil
-}
-
 // NewDNSYarpcDispatcherProvider create a dispatcher provider which handles with IP address
 func NewDNSYarpcDispatcherProvider(logger log.Logger, interval time.Duration) DispatcherProvider {
+	if interval <= 0 {
```
shreyassrivatsan: should min be 0 or something 1 second ?

client/clientBean.go
```diff
@@ -306,24 +274,24 @@ func (d *dnsUpdater) Start() {
 	}()
 }
 
-func (d *dnsUpdater) refresh() (map[string]struct{}, *peer.ListUpdates, bool, error) {
+func (d *dnsUpdater) refresh() (*dnsRefreshResult, error) {
```
shreyassrivatsan: neat signature now!

client/clientBean.go
```diff
@@ -280,23 +248,23 @@ func (d *dnsUpdater) Start() {
...
-			newPeers, updates, changed, err := d.refresh()
+			res, err := d.refresh()
 			if err != nil {
 				d.logger.Error("Failed to update DNS", tag.Error(err), tag.Address(d.dnsAddress))
 			}
-			if changed {
-				if len(updates.Additions) > 0 {
-					d.logger.Info("Add new peers by DNS lookup", tag.Address(d.dnsAddress), tag.Addresses(identifiersToStringList(updates.Additions)))
+			if res.changed {
+				if len(res.updates.Additions) > 0 {
```
shreyassrivatsan: Seems like you could just log this in the refresh function and do not need to return it..

client/clientBean.go
longquanzheng: actually I should remove this line. ipPortRegex is not being used anymore after I remove the old code.

client/clientBean.go
longquanzheng: addressed offline

Issue #2098, opened 2019-06-24
matching: high throughput task list 


venkat1109:
Cadence tasks lists are light weight task queues that are used to dispatch activity and decision task to cadence workers. Task lists are specific to a cadence domain but can be shared by multiple workflows within a domain. The throughput of a task list is directly proportional to the number of events generated per sec by the workflows that use that task list. Internally, task lists are owned and managed by cadence-matching service. Within matching service, task list ownership is determined by consistent hashing and the host responsible for a task list acquires a lease on the task list using CAS primitives at the persistence layer.  Given this architecture, it’s clear that a single task list throughput cannot scale beyond a point.

While this limitation can be documented and also easily worked around by using multiple task lists within a domain, there is no automatic way to prevent customers from accidentally relying on this limitation. The goal here is to provide a single task list offering for cadence customers that can provide high throughput / low latency without needing any code changes from customers.

venkat1109:
Proposal
--

**High Level Overview**
On a high level, each task list will have N partitions behind the scenes where N is configurable. When history needs to add a task to a task list, a partition is picked by task_id % N at the routing / hashing layer. When workers poll for a task list, frontend will randomly pick a partition and forward the poll request to that partition. Rate limits at task list level will be split equally among all the partitions. Sync match will continue to work as is on each of the individual task list partitions. Except for the hashing / routing layer used to talk to matching service, the rest of the system will work as is unaware of the partitioning mechanism. 

**Optimization for Low Throughput Scenarios**
One problem with the approach described above is that it adds additional latency for task dispatch when throughput is low and there are multiple partitions. For example, let’s say there are two partitions on a task list; partition-0 has a backlog of 5 tasks and partition-1 has none. Now, when a poller arrives at frontend, it picks a random partition, say partition-1 and forwards the poller to the host owning that task list. Since partition-1 has no tasks, we can either block until poll timeout or immediately return and have frontend try another partition. Neither of these approaches are ideal. To solve this problem, we organize the task lists into a k-ary tree where k is configurable. Basic idea is that when addTask or pollTask arrives at a non-root partition and  the partition is idle, then forward the request to the parent with some rate limiting. Specifically, the approach requires  the following rules:

*AddTask (partition int, task Task)*
- Try local sync match first
[...]

PR #2125, opened 2019-06-27
matching: refactor taskListManager to separate out individual components

venkat1109:
First step towards #2098. This patch separates out individual components within taskListManager into their own class / files. Most important parts are:

- the piece of code that matches a task producer with a task consumer is separated out into a new class called TaskMatcher
- taskReader used to acquire a ratelimit token, read a task from db and then block infinitely until a poller arrives. This meant ratelimit not being honored in certain cases. This behavior is now fixed. This is the only behavior change in this PR.  

Other things:
- taskListManager.AddTask() now takes a context param (previously never honored the context)
- taskContext is replaced with internalTask. The RecordXXXStarted() methods are removed from the taskContext since they don't really belong there
- taskReader is moved into its own class - but it still takes taskListManager as a param because of the tight coupling. I tried to break this but there are optimizations that necessitate access to internal state of taskListManager. So this part of refactoring is purely for readability - from a encapsulation perspective, it doesn't buy anything
- All configs are moved to one file

### DIFF
--- a/service/matching/matcher.go
+++ b/service/matching/matcher.go
@@ -25,7 +25,6 @@ import (
 	"errors"
 	"time"
 
-	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/metrics"
 	"golang.org/x/time/rate"
 )
@@ -50,9 +49,6 @@ type TaskMatcher struct {
 const (
 	_defaultTaskDispatchRPS    = 100000.0
 	_defaultTaskDispatchRPSTTL = 60 * time.Second
-	// maxRateLimitWaitTime is the max amount of time that we are willing to wait for a
-	// ratelimit token to be available in future
-	maxRateLimitWaitTime = int64(200 * time.Millisecond)
 )
 
 var errTasklistThrottled = errors.New("cannot add to tasklist, limit exceeded")
@@ -107,7 +103,7 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
 			err = <-task.syncResponseCh
 			return true, err
 		}
-		return false, err
+		return false, nil
 	default: // no poller waiting for tasks
 		if rsv != nil {
 			// there was a ratelimit token we consumed
@@ -193,12 +189,9 @@ func (tm *TaskMatcher) ratelimit(ctx context.Context) (*rate.Reservation, error)
 		return nil, nil
 	}
 
-	// how long can we wait for a ratelimit token to be available in future
-	timeout := time.Duration(common.MinInt64(int64(deadline.Sub(time.Now())), maxRateLimitWaitTime))
-
 	rsv := tm.limiter.Reserve()
 	// If we have to wait too long for reservation, give up and return
-	if !rsv.OK() || rsv.Delay() > timeout {
+	if !rsv.OK() || rsv.Delay() > deadline.Sub(time.Now()) {
 		if rsv.OK() { // if we were indeed given a reservation, return it before we bail out
 			rsv.Cancel()
 		}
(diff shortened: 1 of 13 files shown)


service/matching/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+			// there was a ratelimit token we consumed
+			// return it since we did not really do any work
+			rsv.Cancel()
+		}
+		return false, nil
+	}
+}
+
+// MustOffer blocks until a consumer is found to handle this task
+// Returns error only when context is canceled or the ratelimit is set to zero (allow nothing)
```
longquanzheng: Should it be rate exceeds threshold instead of ratelimit being zero?

service/matching/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		if err := tm.limiter.Wait(ctx); err != nil {
+			return nil, err
+		}
+		return nil, nil
+	}
+
+	// how long can we wait for a ratelimit token to be available in future
+	timeout := time.Duration(common.MinInt64(int64(deadline.Sub(time.Now())), maxRateLimitWaitTime))
```
longquanzheng: Can you explain more about the idea?
Looks like if no deadline, which means deadline can be forever, it will wait forever (limiter.Wait(ctx) )
But if deadline is set, it won't only wait for up to maxRateLimitWaitTime. 

Why not being consistent that always honor the deadline, or both using upper bound maxRateLimitWaitTime?

venkat1109: I don't totally like this part of the code either - its needed to get desired existing production behavior. Basic idea is this 

- when context deadline is set, the call comes from history - which cannot block forever - we try to block sometime to sync match - but after that, we simply write the task to db
- when context deadline is not set, the call comes from taskReader (reads backlog from db and tries to dispatch to poller) - in this case, enforcing a shorter timeout doesn't really make sense. The caller wants to block infinitely until a token can be obtained.   

How about this - I will move the part that enforces maxRateLimitWaitTime to the caller instead of being done here ? That way, the code here becomes super simple - it simply uses the context as it is provided.

service/matching/matcher.go
venkat1109: see comment below. this blocks until ratelimit token becomes available.

service/matching/taskReader.go
```diff
@@ -22,114 +22,156 @@ package matching
...
+	}
+)
+
+func newTaskReader(tlMgr *taskListManagerImpl) *taskReader {
+	ctx, cancel := context.WithCancel(context.Background())
+	return &taskReader{
+		tlMgr:               tlMgr,
+		cancelCtx:           ctx,
+		cancelFunc:          cancel,
+		taskBuffer:          make(chan *persistence.TaskInfo, tlMgr.config.GetTasksBatchSize()-1),
```
longquanzheng: I know this is from existing code. But do you know why it needs to be -1?

service/matching/matcher.go
longquanzheng: I see. I misread the "context MUST NOT have a deadline associated"
In addition of this comment, maybe also add a check in the method in case someone misuse this method in some other place

service/matching/matcher.go
longquanzheng: yes that make sense and the code will be much easier to understand and debug

service/matching/taskReader.go
venkat1109: my best guess: the deliverBufferedTasks() goroutine always dequeues the head of the buffer and blocks trying to dispatch that task to a poller - so, when there are no pollers, the amount of tasks buffered will be len(taskBuffer) + 1 ; the (-1) was an attempt to offset that additional item - I will add a comment about this

PR #2140, opened 2019-06-28
POC of randomized event generator for history events

yux0:
POC of randomized event generator for history events. This generator is to generate history events to test 3+DC logic.


host/xdc/event_generator.go
```diff
@@ -0,0 +1,374 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
wxing1292: i would put this in common/xdc

wxing1292: and copyright 2019

host/xdc/event_generator.go
```diff
@@ -0,0 +1,374 @@
...
+type (
+	// Model is a group of edges
+	Model interface {
+		AddEdge(...Edge)
+		ListEdges() []Edge
+	}
+
+	// Generator generates random events based on the defined models
+	Generator interface {
+		AddInitialEntryVertex(...Vertex)
```
wxing1292: we might need more description about the terminology
let us sync on Monday

host/xdc/history_event_test.go
```diff
@@ -0,0 +1,247 @@
...
+
+	ewfModel := NewHistoryEventModel()
+	ewfSignal := NewHistoryEvent(shared.EventTypeSignalExternalWorkflowExecutionInitiated.String())
+	ewfSignalFailed := NewHistoryEvent(shared.EventTypeSignalExternalWorkflowExecutionFailed.String())
+	ewfSignaled := NewHistoryEvent(shared.EventTypeExternalWorkflowExecutionSignaled.String())
+	ewfCancel := NewHistoryEvent(shared.EventTypeRequestCancelExternalWorkflowExecutionInitiated.String())
+	ewfCancelFail := NewHistoryEvent(shared.EventTypeRequestCancelExternalWorkflowExecutionFailed.String())
+	ewfCanceled := NewHistoryEvent(shared.EventTypeExternalWorkflowExecutionCancelRequested.String())
+
+	dtCompleteToEWFSignal := NewConnection(dtComplete, ewfSignal)
```
wxing1292: try use variable names that is more clear, like decision timeout? i guess for this case

host/xdc/event_generator.go
yux0: Updated the location and the copyright header

host/xdc/event_generator.go
yux0: Added more detail about the interfaces.

host/xdc/history_event_test.go
yux0: Replace all the short names with readable names

PR #2142, opened 2019-06-29
Refactor workflow reset path

wxing1292:
* Refactor workflow reset path in Cassandra / MySQL persistence layer
* Remove redudent check on Cassandra: update /delete activity / timer / child workflow / cancel workflow / signal workflow / buffered events
* Keeping condition check on update execution, thus batch operation will still check update condition
* Refactor shard context to use allocateTransferIDsLocked and allocateTimerIDsLocked helper functions


service/history/shardContext.go
```diff
@@ -383,35 +387,17 @@ func (s *shardContextImpl) CreateWorkflowExecution(request *persistence.CreateWo
...
-		}
-		s.logger.Debug(fmt.Sprintf("Assigning replication task ID: %v", id))
-		task.SetTaskID(id)
-		transferMaxReadLevel = id
-	}
-
-	defer s.updateMaxReadLevelLocked(transferMaxReadLevel)
-
-	if err = s.allocateTimerIDsLocked(
+	if err := s.allocateTaskIDsLocked(
```
meiliang86: Return transferMaxReadLevel instead?

PR #2144, opened 2019-07-01
Only delete first event batch if the error is generated by Cadence pe…

wxing1292:
…rsistence layer


service/history/historyEngine.go
```diff
@@ -383,16 +383,16 @@ func (e *historyEngineImpl) StartWorkflowExecution(
 	startRequest *h.StartWorkflowExecutionRequest,
 ) (resp *workflow.StartWorkflowExecutionResponse, retError error) {
```
shreyassrivatsan: can get rid of `resp` and `retError`

wxing1292: it is used on line 408

PR #2145, opened 2019-07-01
Batch terminating workflow

longquanzheng:
Tested locally:
start wf and child wfs:
```
longer@~/gocode/src/github.com/uber/cadence:(batch-term)$ cadence --do samples-domain wf list --query "WorkflowType='main.SampleParentWorkflow' and CloseTime=missing order by CloseTime ASC" -ps 100
        WORKFLOW TYPE       |                     WORKFLOW ID                     |                RUN ID                | START TIME | EXECUTION TIME | END TIME
  main.SampleParentWorkflow | parentworkflow_952970e3-f223-48a8-8e2d-5db9d5871663 | ffd9bb6d-140a-45da-be12-52b4e31e873c | 17:26:51   | 17:26:51       | 16:00:00
  main.SampleParentWorkflow | parentworkflow_efe85202-f1a7-4c17-9a2f-dcb7e35c3b2c | f5e41f08-bd4b-425f-a9d8-6c1347824a6d | 17:27:02   | 17:27:02       | 16:00:00
  main.SampleParentWorkflow | parentworkflow_4dc12065-cc8d-4e42-8da7-18a3aeee187d | ede21ac0-8465-480d-8c25-50da1dbe2dbe | 17:26:54   | 17:26:54       | 16:00:00
  main.SampleParentWorkflow | parentworkflow_e25501b5-c859-462e-a517-eb050dc22b26 | e94eda6e-e1bf-416b-8779-2bf4ed48d876 | 17:27:05   | 17:27:05       | 16:00:00
  main.SampleParentWorkflow | parentworkflow_68bb29ea-eb36-4b4d-985e-18d3275f290b | e4c70df5-3d3a-4879-9fec-cde5656e2d9d | 17:27:02   | 17:27:02       | 16:00:00
  main.SampleParentWorkflow | parentworkflow_73d2b47d-a47c-45e0-816d-110ed03a5ac0 | e235d9dd-ee76-4712-93df-d843dc7821dd | 17:27:02   | 17:27:02       | 16:00:00
...
...
...
...
```

Start batch terminate wf
```
longer@~/gocode/src/github.com/uber/cadence:(batch-term)$ cadence --do cadence-system-global wf start -w "test-batch-14" -wt cadence-sys-batch-workflow -et 100000 -dt 100 -i "{\"DomainName\":\"samples-domain\",\"Query\":\"WorkflowType='main.SampleParentWorkflow' and CloseTime=missing\",\"Reason\":\"test-reason\",\"BatchType\":\"terminate\",\"TerminateParams\":{\"TerminateChildren\":true}}" --tl cadence-sys-batcher-tasklist
Started Workflow Id: test-batch-14, run Id: c56ceeeb-655d-46f6-9d00-e693f4acf638
```

target wf terminated:
```
longer@~/gocode/src/github.com/uber/cadence:(batch-term)$ cadence --do samples-domain wf list --query "WorkflowType='main.SampleParentWorkflow' and CloseTime=missing order by CloseTime ASC" -ps 100
  WORKFLOW TYPE | WORKFLOW ID | RUN ID | START TIME | EXECUTION TIME | END TIME
longer@~/gocode/src/github.com/uber/cadence:(batch-term)$ cadence --do samples-domain wf list --open
  WORKFLOW TYPE | WORKFLOW ID | RUN ID | START TIME | EXECUTION TIME
```

Also tested heartbeat:
```
  "PendingActivities": [
    {
      "ActivityID": "0",
      "ActivityType": {
        "name": "cadence-sys-batch-activity"
      },
      "State": "STARTED",
      "LastStartedTimestamp": "2019-07-02T17:27:16-07:00",
      "HeartbeatDetails": "{\"PageToken\":\"eyJGcm9tIjowLCJTb3J0VmFsdWUiOm51bGwsIlRpZUJyZWFrZXIiOiIiLCJTY3JvbGxJRCI6IkRuRjFaWEo1VkdobGJrWmxkR05vQlFBQUFBQUFBQUhtRmpOU1pFNXRXREJvVkhsNVZuSjVTME5HUm0xQ1prRUFBQUFBQUFBQjV4WXpVbVJPYlZnd2FGUjVlVlp5ZVV0RFJrWnRRbVpCQUFBQUFBQUFBZWdXTTFKa1RtMVlNR2hVZVhsV2NubExRMFpHYlVKbVFRQUFBQUFBQUFIcEZqTlNaRTV0V0RCb1ZIbDVWbko1UzBOR1JtMUNaa0VBQUFBQUFBQUI2aFl6VW1ST2JWZ3dhRlI1ZVZaeWVVdERSa1p0UW1aQiJ9\",\"CurrentPage\":18,\"TotalEstimate\":39,\"SuccessCount\":36,\"ErrorCount\":0}\n",
      "LastHeartbeatTimestamp": "2019-07-02T17:27:30-07:00",
      "Attempt": 0,
      "ExpirationTimestamp": "2039-06-27T16:27:16-08:00"
    }
  ]
```

### DIFF
--- a/service/worker/batcher/batcher.go
+++ b/service/worker/batcher/batcher.go
@@ -108,7 +108,7 @@ func (s *Batcher) Start() error {
 		MetricsScope:              s.tallyScope,
 		BackgroundActivityContext: context.WithValue(context.Background(), batcherContextKey, s),
 	}
-	worker := worker.New(s.svcClient, common.SystemLocalDomainName, batcherTaskListName, workerOpts)
+	worker := worker.New(s.svcClient, common.SystemGlobalDomainName, batcherTaskListName, workerOpts)
 	return worker.Start()
 }
 
@@ -135,8 +135,8 @@ func (s *Batcher) createGlobalSystemDomainIfNotExists() error {
 		return nil
 	}
 
-	if s.cfg.ClusterMetadata.IsMasterCluster() && !s.cfg.ClusterMetadata.IsGlobalDomainEnabled() {
-		return fmt.Errorf("not master cluster, retry on describe domain only")
+	if s.cfg.ClusterMetadata.IsGlobalDomainEnabled() && !s.cfg.ClusterMetadata.IsMasterCluster() {
+		return fmt.Errorf("not master of the global cluster, retry on describe domain only")
 	}
 
 	ctx, cancel = context.WithTimeout(context.Background(), rpcTimeout)
--- a/service/worker/batcher/batcher.go
+++ b/service/worker/batcher/batcher.go
@@ -71,19 +71,14 @@ type (
 		TallyScope tally.Scope
 	}
 
-	// batcherContext is the context object that get's
-	// passed around within the scanner workflows / activities
-	batcherContext struct {
+	// Batcher is the background sub-system that execute workflow for batch operations
+	// It is also the context object that get's passed around within the scanner workflows / activities
+	Batcher struct {
 		cfg           Config
 		svcClient     workflowserviceclient.Interface
 		metricsClient metrics.Client
 		tallyScope    tally.Scope
 		logger        log.Logger
-	}
-
-	// Batcher is the background sub-system that execute workflow for batch operations
-	Batcher struct {
-		context batcherContext
 	}
 )
 
@@ -91,13 +86,11 @@ type (
 func New(params *BootstrapParams) *Batcher {
 	cfg := params.Config
 	return &Batcher{
-		context: batcherContext{
-			cfg:           cfg,
-			svcClient:     params.ServiceClient,
-			metricsClient: params.MetricsClient,
-			tallyScope:    params.TallyScope,
-			logger:        params.Logger.WithTags(tag.ComponentBatcher),
-		},
+		cfg:           cfg,
+		svcClient:     params.ServiceClient,
+		metricsClient: params.MetricsClient,
+		tallyScope:    params.TallyScope,
+		logger:        params.Logger.WithTags(tag.ComponentBatcher),
 	}
 }
 
@@ -112,10 +105,10 @@ func (s *Batcher) Start() error {
 
 	// start worker for batch operation workflows
 	workerOpts := worker.Options{
-		MetricsScope:              s.context.tallyScope,
-		BackgroundActivityContext: context.WithValue(context.Background(), batcherContextKey, s.context),
+		MetricsScope:              s.tallyScope,
+		BackgroundActivityContext: context.WithValue(context.Background(), batcherContextKey, s),
 	}
-	worker := worker.New(s.context.svcClient, common.SystemLocalDomainName, batcherTaskListName, workerOpts)
+	worker := worker.New(s.svcClient, common.SystemLocalDomainName, batcherTaskListName, workerOpts)
 	return worker.Start()
 }
 
@@ -132,28 +125,28 @@ func (s *Batcher) createGlobalSystemDomainIfNotExistsWithRetry() error {
 
 func (s *Batcher) createGlobalSystemDomainIfNotExists() error {
 	ctx, cancel := context.WithTimeout(context.Background(), rpcTimeout)
-	_, err := s.context.svcClient.DescribeDomain(ctx, &shared.DescribeDomainRequest{
+	_, err := s.svcClient.DescribeDomain(ctx, &shared.DescribeDomainRequest{
 		Name: common.StringPtr(common.SystemGlobalDomainName),
 	})
 	cancel()
 
 	if err == nil {
-		s.context.logger.Info("Global system domain already exists", tag.WorkflowDomainName(common.SystemGlobalDomainName))
+		s.logger.Info("Global system domain already exists", tag.WorkflowDomainName(common.SystemGlobalDomainName))
 		return nil
 	}
 
-	if s.context.cfg.ClusterMetadata.IsMasterCluster() && !s.context.cfg.ClusterMetadata.IsGlobalDomainEnabled() {
+	if s.cfg.ClusterMetadata.IsMasterCluster() && !s.cfg.ClusterMetadata.IsGlobalDomainEnabled() {
 		return fmt.Errorf("not master cluster, retry on describe domain only")
 	}
 
 	ctx, cancel = context.WithTimeout(context.Background(), rpcTimeout)
-	err = s.context.svcClient.RegisterDomain(ctx, s.getDomainCreationRequest())
+	err = s.svcClient.RegisterDomain(ctx, s.getDomainCreationRequest())
 	cancel()
 	if err != nil {
-		s.context.logger.Error("Error creating global system domain", tag.Error(err))
+		s.logger.Error("Error creating global system domain", tag.Error(err))
 		return err
 	}
-	s.context.logger.Info("Domain created successfully")
+	s.logger.Info("Domain created successfully")
 	return nil
 }
 
@@ -162,15 +155,15 @@ func (s *Batcher) getDomainCreationRequest() *shared.RegisterDomainRequest {
 		Name:                                   common.StringPtr(common.SystemGlobalDomainName),
 		WorkflowExecutionRetentionPeriodInDays: common.Int32Ptr(common.SystemDomainRetentionDays),
 		EmitMetric:                             common.BoolPtr(true),
-		SecurityToken:                          common.StringPtr(s.context.cfg.AdminOperationToken()),
+		SecurityToken:                          common.StringPtr(s.cfg.AdminOperationToken()),
 		IsGlobalDomain:                         common.BoolPtr(false),
 	}
 
-	if s.context.cfg.ClusterMetadata.IsGlobalDomainEnabled() {
+	if s.cfg.ClusterMetadata.IsGlobalDomainEnabled() {
 		req.IsGlobalDomain = common.BoolPtr(true)
-		req.ActiveClusterName = common.StringPtr(s.context.cfg.ClusterMetadata.GetMasterClusterName())
+		req.ActiveClusterName = common.StringPtr(s.cfg.ClusterMetadata.GetMasterClusterName())
 		var clusters []*shared.ClusterReplicationConfiguration
-		for name, c := range s.context.cfg.ClusterMetadata.GetAllClusterInfo() {
+		for name, c := range s.cfg.ClusterMetadata.GetAllClusterInfo() {
 			if !c.Enabled {
 				continue
 			}
--- a/service/worker/batcher/batcher.go
+++ b/service/worker/batcher/batcher.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/opentracing/opentracing-go"
 	"github.com/uber-go/tally"
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/backoff"
@@ -107,6 +108,7 @@ func (s *Batcher) Start() error {
 	workerOpts := worker.Options{
 		MetricsScope:              s.tallyScope,
 		BackgroundActivityContext: context.WithValue(context.Background(), batcherContextKey, s),
+		Tracer:                    opentracing.GlobalTracer(),
 	}
 	worker := worker.New(s.svcClient, common.SystemGlobalDomainName, batcherTaskListName, workerOpts)
 	return worker.Start()
--- a/service/worker/batcher/workflow.go
+++ b/service/worker/batcher/workflow.go
@@ -26,9 +26,11 @@ import (
 	"time"
 
 	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/backoff"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"go.uber.org/cadence"
+	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
 	"go.uber.org/cadence/.gen/go/shared"
 	"go.uber.org/cadence/activity"
 	"go.uber.org/cadence/workflow"
@@ -209,12 +211,7 @@ func BatchActivity(ctx context.Context, batchParams BatchParams) error {
 		// TODO https://github.com/uber/cadence/issues/2154
 		//  Need to improve scan concurrency because it will hold an ES resource until the workflow finishes.
 		//  And we can't use list API because terminate / reset will mutate the result.
-		resp, err := batcher.svcClient.ScanWorkflowExecutions(ctx, &shared.ListWorkflowExecutionsRequest{
-			PageSize:      common.Int32Ptr(int32(pageSize)),
-			Domain:        common.StringPtr(batchParams.DomainName),
-			NextPageToken: hbd.PageToken,
-			Query:         common.StringPtr(batchParams.Query),
-		})
+		resp, err := scanWorkflowsWithRetry(ctx, batcher.svcClient, hbd.PageToken, batchParams)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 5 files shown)


service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+				attempts:  0,
+				hbd:       hbd,
+			}
+		}
+		batchCount := len(resp.Executions)
+
+		// wait for counters indicate this batch is done
+	Loop:
+		for {
+			select {
```
shreyassrivatsan: why counters? Why not just loop here waiting for responses with error or not.. That seems more useful - in case you want to return back errors/persist them somewhere.. 

I dont think this helps with performance also as the different go routines will have to contend on the atomic increment.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+
+			switch batchParams.BatchType {
+			case BatchTypeTerminate:
+				err = processOneTerminateTask(ctx, limiter, task, batchParams)
+			}
+			if err != nil {
+				getActivityLogger(ctx).Error("Failed to process task", tag.Error(err))
+				// put back to the channel if less than attemptsOnError
+				task.attempts++
+				if task.attempts >= batchParams.AttemptsOnRetryableError {
```
shreyassrivatsan: So expectation is that `err` if returned has to be retryable? That seems prone to errors.. The error handling here should deal with different types of errors - we shouldn't expect the specific methods to handle the retryable logic.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+		err = batcher.svcClient.TerminateWorkflowExecution(newCtx, &shared.TerminateWorkflowExecutionRequest{
+			Domain:            common.StringPtr(param.DomainName),
+			WorkflowExecution: &wf,
+			Reason:            common.StringPtr(param.Reason),
+			Identity:          common.StringPtr(batchWFTypeName),
+		})
+		cancel()
+		if err != nil {
+			_, ok := err.(*shared.EntityNotExistsError)
+			if !ok {
```
shreyassrivatsan: Shouldn't we return the error in either case? If parent terminate fails even for a retryable error we shouldnt try to terminate children.. We should try to terminate parent first..

Gopkg.toml
```diff
@@ -32,7 +32,7 @@ ignored = ["github.com/uber/cadence/.gen"]
 
 [[constraint]]
   name = "github.com/apache/thrift"
-  version = "0.9.3"
+  version = "0.10.0"
```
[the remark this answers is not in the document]
longquanzheng: Because client has updated to that version: https://github.com/uber-go/cadence-client/blob/master/Gopkg.toml#L62
dep ensure -update will not work without resolving this conflict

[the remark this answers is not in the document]
longquanzheng: Also need to update client because it needs latest client to be able to get children info from describeWF API

service/worker/batcher/workflow.go
longquanzheng: That's because we need to continue on terminating all childWFs in this case. Do you have any idea to implement if we returning early here?

service/worker/batcher/workflow.go
longquanzheng: The most cost is I/O to cadence server and ratelimiter and most of them times they are waiting for them. Because we have to use multiple go routines, waiting for return will not work. Also I wan to return counters to do heartbeat for visibility of the progress. Without these atomic counters, we can use channel to return counters. Or do you have any preference?

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+	defaultActivityHeartBeatTimeout = time.Minute
 )
+
+const (
+	BatchTypeTerminate = "terminate"
+	BatchTypeReset     = "reset"
+)
+
+type (
+	BatchParams struct {
```
vancexu: Please check and fix lint warning before merge

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+		params.Concurrency = defaultConcurrency
+	}
+	if params.AttemptsOnRetryableError <= 0 {
+		params.AttemptsOnRetryableError = defaultAttemptsOnRetryableError
+	}
+	if params.ActivityHeartBeatTimeout <= 0 {
+		params.ActivityHeartBeatTimeout = defaultActivityHeartBeatTimeout
+	}
+	switch params.BatchType {
+	case BatchTypeReset:
```
vancexu: I feel this switch could be simpler, could we just:
```
case Terminate:
  // do something or nothing
default:
  // not support
```
and later for each support type we could add a case.

(The current usage of `fallthrough` makes it a little bit harder to read IMHO)

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+type (
+	BatchParams struct {
+		// Target domain to execute batch operation
+		DomainName string
+		// To query the target workflows for processing
+		QueryCondition string
+		// Reaason for the operation
+		Reason string
+		// Supporting: reset,terminate
+		BatchType string
```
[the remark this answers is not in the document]
longquanzheng: This will be passing from JSON so it will be hardcoded string anyway.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+		// Number of attempts for each workflow to process in case of retryable error before giving up
+		AttemptsOnRetryableError int
+		ActivityHeartBeatTimeout time.Duration
+	}
+
+	HeartBeatDetails struct {
+		PageToken   []byte
+		CurrentPage int
+		// This is just an estimation for visibility
+		TotalEstimate int64
```
vancexu: Just curious what is the purpose of this TotalEstimate?

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+		// TODO https://github.com/uber/cadence/issues/2154
+		//  Need to improve scan concurrency because it will hold a ES resource for a long time.
+		//  we can't use list API because terminate / reset will mutate the result.
+		resp, err := batcher.svcClient.ScanWorkflowExecutions(ctx, &shared.ListWorkflowExecutionsRequest{
+			PageSize:      common.Int32Ptr(int32(pageSize)),
+			Domain:        common.StringPtr(batchParams.DomainName),
+			NextPageToken: hbd.PageToken,
+			Query:         common.StringPtr(batchParams.QueryCondition),
+		})
+		if err != nil {
```
vancexu: Since it is using service client, I think we need retry to handle transient internal server error (such as service busy error) here instead of activity retry.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+
+	return nil
+}
+
+func startTaskProcessor(
+	ctx context.Context, batchParams BatchParams, taskCh chan taskDetail,
+	limiter *rate.Limiter, doneCount, errCount *int32) {
+
+	for {
+		select {
```
vancexu: Need to add done ctx or channel handing, otherwise this startTaskProcessor goroutine will leak. 
One leak case: goroutines start but no tasks produced in taskCh.

service/worker/batcher/workflow.go
```diff
@@ -132,8 +133,11 @@ func BatchWorkflow(ctx workflow.Context, batchParams BatchParams) error {
 }
 
 func setDefaultParams(params BatchParams) (BatchParams, error) {
-	if params.BatchType == "" || params.Reason == "" || params.DomainName == "" || params.QueryCondition == "" {
-		return BatchParams{}, fmt.Errorf("must provide required parameters: BatchType/Reason/DomainName/QueryCondition")
+	if params.BatchType == "" ||
```
shreyassrivatsan: It will be nice to split setDefault and validation.. Call `setDefault` first.

And then call validate which returns an error.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,332 @@
...
+	BatchTypeTerminate = "terminate"
+	BatchTypeReset     = "reset"
+)
+
+type (
+	// BatchParams is the parameters for batch operation workflow
+	BatchParams struct {
+		// Target domain to execute batch operation
+		DomainName string
+		// To get the target workflows for processing
```
shreyassrivatsan: This comment will fail lint.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,328 @@
...
+		cancel()
+		if err != nil {
+			_, ok := err.(*shared.EntityNotExistsError)
+			if !ok {
+				return err
+			}
+		}
+		wfs = wfs[1:]
+		newCtx, cancel = context.WithDeadline(ctx, time.Now().Add(rpcTimeout))
+		resp, err := batcher.svcClient.DescribeWorkflowExecution(newCtx, &shared.DescribeWorkflowExecutionRequest{
```
vancexu: The processing of children surprise me: we allow user to query but force to terminate child also without option.
I think we should use child policy in parent here to decide whether to do same to children or not.

service/worker/batcher/workflow.go
shreyassrivatsan: As discussed, let's wait on a response channel for tasks to complete and accumulate error/success counts in this method.

service/worker/batcher/workflow.go
longquanzheng: When describeWF, you can see the HeartbeatDetails. With this customers can have visibility of the progress.

service/worker/batcher/workflow.go
longquanzheng: Addressed offline: adding parameter to allow not terminate children,and add todo for future after we fix the ChildPolicy issue in cadence server

service/worker/batcher/workflow.go
longquanzheng: Done

service/worker/batcher/workflow.go
longquanzheng: OK

service/worker/batcher/workflow.go
longquanzheng: OK

service/worker/batcher/workflow.go
longquanzheng: done

service/worker/batcher/batcher.go
```diff
@@ -112,10 +105,10 @@ func (s *Batcher) Start() error {
 
 	// start worker for batch operation workflows
 	workerOpts := worker.Options{
-		MetricsScope:              s.context.tallyScope,
-		BackgroundActivityContext: context.WithValue(context.Background(), batcherContextKey, s.context),
+		MetricsScope:              s.tallyScope,
```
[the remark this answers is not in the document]
longquanzheng: done

service/worker/batcher/workflow.go
longquanzheng:
```
longer@~/gocode/src/github.com/uber/cadence:(batch-term)$ make lint
./install-dep.sh
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/cask).
No changes to formulae.

Warning: dep 0.5.4 is already installed and up-to-date
To reinstall 0.5.4, run `brew reinstall dep`
dep ensure
Running linter
```
This is OK because the upper level(BatchParams) already satisfy Linter's requirement

service/worker/batcher/workflow.go
longquanzheng: done

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+
+	return resp, err
+}
+
+func startTaskProcessor(
+	ctx context.Context,
+	batchParams BatchParams,
+	taskCh chan taskDetail,
+	respCh chan error,
+	limiter *rate.Limiter) {
```
shreyassrivatsan: when stacking this line should end as
```
	respCh chan error,
	limiter *rate.Limiter,
) {
```

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case task := <-taskCh:
+			if isDone(ctx) {
```
shreyassrivatsan: I dont think we need this check because of the upper case.. If the context gets cancelled while processing the task the RPC calls etc, will return an error and this method will return..

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+	wfs := []shared.WorkflowExecution{task.execution}
+	for len(wfs) > 0 {
+		wf := wfs[0]
+
+		err := limiter.Wait(ctx)
+		if err != nil {
+			getActivityLogger(ctx).Error("Failed to wait for rateLimiter", tag.Error(err))
+		}
+
+		newCtx, cancel := context.WithDeadline(ctx, time.Now().Add(rpcTimeout))
```
shreyassrivatsan: use `context.WithTimeout`

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+		})
+		cancel()
+		if err != nil {
+			_, ok := err.(*shared.EntityNotExistsError)
+			if !ok {
+				return err
+			}
+		}
+		wfs = wfs[1:]
+		newCtx, cancel = context.WithDeadline(ctx, time.Now().Add(rpcTimeout))
```
shreyassrivatsan: use `context.WithTimeout()`

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+					taskCh <- task
+				}
+			} else {
+				respCh <- nil
+			}
+		}
+	}
+}
+
+func processOneTerminateTask(ctx context.Context, limiter *rate.Limiter, task taskDetail, batchParams BatchParams) error {
```
shreyassrivatsan: nit: name `processTerminateTask`

service/worker/batcher/workflow.go
```diff
@@ -198,10 +198,7 @@ func BatchActivity(ctx context.Context, batchParams BatchParams) error {
 	}
 
 	if startOver {
-		resp, err := batcher.svcClient.CountWorkflowExecutions(ctx, &shared.CountWorkflowExecutionsRequest{
-			Domain: common.StringPtr(batchParams.DomainName),
-			Query:  common.StringPtr(batchParams.Query),
-		})
+		resp, err := countWorkflowsWithRetry(ctx, batcher.svcClient, batchParams)
```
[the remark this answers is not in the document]
longquanzheng: That's probably not true. I thought about the idea of skipping this error here. But the fact is that we also call ListWorkflows before terminating everything, and they are both calling thru ES -- if count fails, most likely List will not succeed. 
Count API is same important as List API, we should provide same availability. 
So to provide a consistent user experience(always seeing visibility/progress), I decided to also return error here.

service/worker/batcher/workflow.go
```diff
@@ -20,7 +20,408 @@
...
+)
+
+const (
+	// BatchTypeTerminate is batch type for terminating workflows
+	BatchTypeTerminate = "terminate"
+)
+
+type (
+	// TerminateParams is the parameters for terminating workflow
+	TerminateParams struct {
```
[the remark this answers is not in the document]
longquanzheng: Agree, it should default to true.

service/worker/batcher/workflow.go
longquanzheng: Actually it's better to check it. If both ctx is canceled and a task comes, but task got processed first, then it will try to consume ratelimiter with the canceled and then return errors. 
See https://godoc.org/golang.org/x/time/rate#Limiter.WaitN

Since go select doesn't guarantee ordering, this could run into a loop until select eventually go into ctx.Done(). 
And there is no harm to check it here.

PR #2152, opened 2019-07-01
Move frontend token bucket under quota policy

shreyassrivatsan:


common/quotas/interfaces.go
```diff
@@ -0,0 +1,36 @@
...
+// and more complex rate limiting functionality.
+type Policy interface {
+	// Allow attempts to allow a request to go through. The method returns
+	// immediately with a true or false indicating if the request can make
+	// progress
+	Allow() bool
+
+	// Wait will wait upto duration d for the request to go through. The method
+	// returns a true or false indicating if the request can make progress.
+	Wait(d time.Duration) bool
```
longquanzheng: Looks like this method is not being used yet in this PR. I don't feel like this is a good signature that you will use in the future. Venkat already have something similar here: https://github.com/uber/cadence/blob/master/service/matching/ratelimiter.go#L92
Maybe you can do the same, or remove it until when you are going to use it.

shreyassrivatsan: I'll remove it for now and bring it back if necessary.. For the API quota case I do not think that API is necessary.

PR #2157, opened 2019-07-02
Expose last failure reason & worker identity in DescribeWorkflowExecution

yycptt:
https://github.com/uber/cadence/issues/698


idl/github.com/uber/cadence/sqlblobs.thrift
```diff
@@ -153,6 +153,8 @@ struct ActivityInfo {
   60: optional i64 (js.type = "Long") retryExpirationTimeNanos
   62: optional double retryBackoffCoefficient
   64: optional list<string> retryNonRetryableErrors
+  66: optional string retryLastFailureReason
```
vancexu: Why it is naming differently in sql?

PR #2162, opened 2019-07-03
Tracing and context propagation documentation

shreyassrivatsan:


docs/03_goclient/16_tracing.md
```diff
@@ -0,0 +1,64 @@
+# Tracing and Context Propagation
+
+## Tracing
+
+The Go client provides distributed tracing support through [opentracing](https://opentracing.io/). Tracing can be
+configured by providing a `opentracing.Tracer` implementation in `Client` and `Worker` instantiation. Tracing allows
```
dmetzgar: For a lot of these types, I'd prefer to link to their documentation on godoc. For instance, opentracing.Tracer could link to https://godoc.org/github.com/opentracing/opentracing-go#Tracer

docs/03_goclient/16_tracing.md
```diff
@@ -0,0 +1,64 @@
...
+On the server side Cadence provides a mechanism to propagate what it calls headers across different workflow
+transitions.
+
+```go
+struct Header {
+  10: optional map<string, binary> fields
+}
+```
+
+The client leverages this to pass around selected context information. `HeaderReader` and `HeaderWriter` are interfaces
```
dmetzgar: HeaderWriter: https://godoc.org/go.uber.org/cadence/internal#HeaderWriter   
HeaderReader: https://godoc.org/go.uber.org/cadence/internal#HeaderReader

docs/03_goclient/16_tracing.md
```diff
@@ -0,0 +1,64 @@
...
+type ContextPropagator interface {
+  Inject(context.Context, HeaderWriter) error
+
+  Extract(context.Context, HeaderReader) (context.Context, error)
+
+  InjectFromWorkflow(Context, HeaderWriter) error
+
+  ExtractToWorkflow(Context, HeaderReader) (Context, error)
+}
+```
```
dmetzgar: Can we add a section that shows how to retrieve headers in a workflow and an activity? Also, how do I make sure they're propagated when I call StartWorkflow/ExecuteWorkflow?

docs/03_goclient/16_tracing.md
```diff
@@ -0,0 +1,64 @@
...
+struct Header {
+  10: optional map<string, binary> fields
+}
+```
+
+The client leverages this to pass around selected context information. `HeaderReader` and `HeaderWriter` are interfaces
+that allow reading and writing to the cadence server headers. The client already provides [implementations](https://github.com/uber-go/cadence-client/blob/master/internal/headers.go)
+for these. `HeaderWriter` sets a field in the header. Headers is a map, so setting a value for the the same key
+multiple times will overwrite the previous values. `HeaderReader` iterates through the headers map and runs the
+provided handler function on each key value pair, allowing the user to deal with the fields one is interested in.
```
purplesusan: allowing the user -> allowing you
one is -> you are

docs/03_goclient/16_tracing.md
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type HeaderReader interface {
+  ForEachKey(handler func(string, []byte) error) error
+}
+```
+
+### Context Propagators
+
+Context propagators require implementing the following 4 methods to propagate selected context across a workflow.
```
purplesusan: This would work better as a bulleted list with each of the methods a list item, following a colon after "workflow". For example:

Context propagators require implementing the following 4 methods to propagate selected context across a workflow:

- `Inject`...
- `InjectFromWorkflow` ...
- `Extract` and `ExtractWorkflow` ...

The [tracing context propagator]...

docs/03_goclient/16_tracing.md
shreyassrivatsan: The user does not really have to anything once he writes a context propagator. Clarified that in the document.

PR #2167, opened 2019-07-03
Add search attribute to DescribeWorkflow

vancexu:
Add search attributes to DescribeWorkflow API, 
Also change CLI to reflect this. 

image

image


service/history/historyEngine.go
```diff
@@ -756,10 +756,11 @@ func (e *historyEngineImpl) DescribeWorkflowExecution(
...
 			},
-			Type:            &workflow.WorkflowType{Name: common.StringPtr(executionInfo.WorkflowTypeName)},
-			StartTime:       common.Int64Ptr(executionInfo.StartTimestamp.UnixNano()),
-			HistoryLength:   common.Int64Ptr(msBuilder.GetNextEventID() - common.FirstEventID),
-			AutoResetPoints: executionInfo.AutoResetPoints,
+			Type:             &workflow.WorkflowType{Name: common.StringPtr(executionInfo.WorkflowTypeName)},
+			StartTime:        common.Int64Ptr(executionInfo.StartTimestamp.UnixNano()),
+			HistoryLength:    common.Int64Ptr(msBuilder.GetNextEventID() - common.FirstEventID),
+			AutoResetPoints:  executionInfo.AutoResetPoints,
+			SearchAttributes: &workflow.SearchAttributes{IndexedFields: executionInfo.SearchAttributes},
```
longquanzheng: Could *executionInfo.SearchAttributes* be null?

tools/cli/flags.go
```diff
@@ -441,8 +443,8 @@ func getFlagsForDescribe() []cli.Flag {
 func getFlagsForDescribeID() []cli.Flag {
 	return []cli.Flag{
 		cli.BoolFlag{
-			Name:  FlagPrintRawTimeWithAlias,
-			Usage: "Print raw time stamp",
+			Name:  FlagPrintRawWithAlias,
+			Usage: "Print raw time stamp and properties as they are",
```
longquanzheng: and **other** properties

Or maybe just say "print raw properties as they are stored".

tools/cli/flags.go
```diff
@@ -101,6 +101,8 @@ const (
 	FlagPrintFullyDetailWithAlias   = FlagPrintFullyDetail + ", pf"
 	FlagPrintRawTime                = "print_raw_time"
```
longquanzheng: Is it still being used by other command? If not we may delete it.

service/history/historyEngine.go
vancexu: Yes it could, but that's fine.

tools/cli/flags.go
vancexu: used by others

tools/cli/flags.go
vancexu: changed

service/history/historyEngine.go
longquanzheng: Then desc result will be weird: SearchAttributes will be not nil but it's content is nil. In this case usually we would just make SearchAttributes is nil, so that when reading this field, only need to check if searchAttr is nil or not

PR #2169, opened 2019-07-05
Update CLI documentation for local dev quickstart

khaale:
https://github.com/uber/cadence/blob/master/docker/README.md contains instruction for CLI usage
```
For example to register new domain 'test-domain' with 1 retention day
`docker run --network=host --rm ubercadence/cli:master --do test-domain domain register -rd 1`
```
which fails with the following message
```
Error: Option global_domain must be provided.
```

I suppose that it worth adding `--global_domain false` to documentation.


vancexu:
Thank you @khaale for the contribution! 

This is a breaking change on CLI, which we shouldn't add. 
We are fixing it to make the CLI command in original instruction works.

Thanks again 
Bowei

dmetzgar:
Actually, I meant to close this. #2172 is checked in.

PR #2171, opened 2019-07-08
Refactor workflow history size / count related code logic

wxing1292:
* Move history size / count limit check into decition handler
* Change exceeding history size / count limit behavior from force termination to fail workflow
* Move workflow stats metrics related code to dedicated file

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -77,7 +77,7 @@ const (
 	FailureReasonHeartbeatExceedsLimit = "HEARTBEAT_EXCEEDS_LIMIT"
 	// FailureReasonDecisionBlobSizeExceedsLimit is the failureReason for decision blob exceeds size limit
 	FailureReasonDecisionBlobSizeExceedsLimit = "DECISION_BLOB_SIZE_EXCEEDS_LIMIT"
-	// FailureReasonSizeExceedsLimit is reason to terminate workflow when history size or count exceed limit
+	// FailureReasonSizeExceedsLimit is reason to fail workflow when history size or count exceed limit
 	FailureReasonSizeExceedsLimit = "HISTORY_EXCEEDS_LIMIT"
 	// FailureReasonTransactionSizeExceedsLimit is the failureReason for when transaction cannot be committed because it exceeds size limit
 	FailureReasonTransactionSizeExceedsLimit = "TRANSACTION_SIZE_EXCEEDS_LIMIT"
--- a/common/util.go
+++ b/common/util.go
@@ -73,12 +73,12 @@ const (
 	FailureReasonFailureDetailsExceedsLimit = "FAILURE_DETAILS_EXCEEDS_LIMIT"
 	// FailureReasonCancelDetailsExceedsLimit is failureReason for cancel details exceeds limit
 	FailureReasonCancelDetailsExceedsLimit = "CANCEL_DETAILS_EXCEEDS_LIMIT"
-	//FailureReasonHeartbeatExceedsLimit is failureReason for heartbeat exceeds limit
+	// FailureReasonHeartbeatExceedsLimit is failureReason for heartbeat exceeds limit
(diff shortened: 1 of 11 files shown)


common/util.go
```diff
@@ -73,12 +73,12 @@ const (
...
 	// FailureReasonCancelDetailsExceedsLimit is failureReason for cancel details exceeds limit
 	FailureReasonCancelDetailsExceedsLimit = "CANCEL_DETAILS_EXCEEDS_LIMIT"
-	//FailureReasonHeartbeatExceedsLimit is failureReason for heartbeat exceeds limit
+	// FailureReasonHeartbeatExceedsLimit is failureReason for heartbeat exceeds limit
 	FailureReasonHeartbeatExceedsLimit = "HEARTBEAT_EXCEEDS_LIMIT"
 	// FailureReasonDecisionBlobSizeExceedsLimit is the failureReason for decision blob exceeds size limit
 	FailureReasonDecisionBlobSizeExceedsLimit = "DECISION_BLOB_SIZE_EXCEEDS_LIMIT"
-	// TerminateReasonSizeExceedsLimit is reason to terminate workflow when history size or count exceed limit
-	TerminateReasonSizeExceedsLimit = "HISTORY_EXCEEDS_LIMIT"
+	// FailureReasonSizeExceedsLimit is reason to terminate workflow when history size or count exceed limit
```
vancexu: Why change to fail instead of terminate? I thought terminate is cleaner and more proper in this case, since customer may not expect or handle workflows terminate state.

vancexu: Why change to fail instead of terminate? I thought terminate is cleaner and more proper in this case, since customer may not expect or handle workflows terminate state.

vancexu: BTW comment not changed.

service/history/workflowExecutionContext.go
```diff
@@ -944,7 +860,7 @@ func (c *workflowExecutionContextImpl) update(
 	c.msBuilder.GetExecutionInfo().LastUpdatedTimestamp = c.timeSource.Now()
 
 	// for any change in the workflow, send a event
-	c.shard.NotifyNewHistoryEvent(newHistoryEventNotification(
+	_ = c.shard.NotifyNewHistoryEvent(newHistoryEventNotification(
```
vancexu: will this pass lint check?

service/history/decisionChecker.go
```diff
@@ -117,6 +135,46 @@ func (c *decisionBlobSizeChecker) failWorkflowIfBlobSizeExceedsLimit(
...
+			tag.WorkflowEventCount(historyCount))
+
+		attributes := &workflow.FailWorkflowExecutionDecisionAttributes{
+			Reason:  common.StringPtr(common.FailureReasonSizeExceedsLimit),
+			Details: []byte("Workflow history size / count exceeds limit."),
+		}
+
+		if _, err := c.mutableState.AddFailWorkflowEvent(c.completedID, attributes); err != nil {
+			return false, &workflow.InternalServiceError{
+				Message: "Unable to add fail workflow event.",
```
shreyassrivatsan: Should also propagate `err` here

service/history/workflowExecutionContext.go
shreyassrivatsan: nit: do not need `_ =`

common/util.go
wxing1292: because existing logic fails a workflow in decision handling:
https://github.com/uber/cadence/blob/master/service/history/decisionChecker.go#L113

the reason to move the size / count check here is
1. decision is inevitable
2. workflow execution context has too much "business logic"

service/history/workflowExecutionContext.go
wxing1292: NotifyNewHistoryEvent function can return error, however in this case, the error should not be returned since workflow change notification is not really important.

BTW `_ = funcation call` is to make the linter happy

wxing1292: this is to make linter happy

PR #2172, opened 2019-07-08
Make global domain check optional when register domain in CLI

vancexu:
Currently it is too strict to force user add cluster info for local domain.  For example, https://github.com/uber/cadence/pull/2169 user should only need
 `cadence --do samples-domain2 d re` instead of  
`cadence --do samples-domain2 d re --global_domain false --active_cluster active --clusters active`

This PR makes global_domain flag optional. It use server side check so that it make sure:
if server enable cross dc then global_domain flag is must provided.

### DIFF
--- a/service/frontend/domainHandler.go
+++ b/service/frontend/domainHandler.go
@@ -116,7 +116,7 @@ func (d *domainHandlerImpl) registerDomain(
 
 	activeClusterName := d.clusterMetadata.GetCurrentClusterName()
 	// input validation on cluster names
-	if registerRequest.ActiveClusterName != nil {
+	if len(registerRequest.GetActiveClusterName()) != 0 {
 		activeClusterName = registerRequest.GetActiveClusterName()
 	}
 	clusters := []*persistence.ClusterReplicationConfig{}
--- a/service/frontend/domainHandler.go
+++ b/service/frontend/domainHandler.go
@@ -116,7 +116,7 @@ func (d *domainHandlerImpl) registerDomain(
 
 	activeClusterName := d.clusterMetadata.GetCurrentClusterName()
 	// input validation on cluster names
-	if len(registerRequest.GetActiveClusterName()) != 0 {
+	if registerRequest.ActiveClusterName != nil {
 		activeClusterName = registerRequest.GetActiveClusterName()
 	}
 	clusters := []*persistence.ClusterReplicationConfig{}
--- a/service/frontend/domainHandler_GlobalDomainDisabled_test.go
+++ b/service/frontend/domainHandler_GlobalDomainDisabled_test.go
@@ -165,6 +165,7 @@ func (s *domainHandlerGlobalDomainDisabledSuite) TestRegisterGetDomain_AllDefaul
 	err := s.handler.registerDomain(context.Background(), &shared.RegisterDomainRequest{
 		Name:                                   common.StringPtr(domainName),
 		WorkflowExecutionRetentionPeriodInDays: common.Int32Ptr(retention),
+		ActiveClusterName:                      common.StringPtr(""),
 	})
 	s.Nil(err)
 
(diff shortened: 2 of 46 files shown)


tools/cli/commands.go
```diff
@@ -208,10 +208,16 @@ func RegisterDomain(c *cli.Context) {
 	}
 
 	var activeClusterName string
+	var clusters []*shared.ClusterReplicationConfiguration
+	if !isGlobalDomain { // set default for local domain
+		activeClusterName = "active"
+		clusters = append(clusters, &shared.ClusterReplicationConfiguration{
```
shreyassrivatsan: nit: when stacking make this the below for clarity.

```
		clusters = append(
                      clusters, 
                      &shared.ClusterReplicationConfiguration{
			    ClusterName: common.StringPtr("active"),
		      },
                  )

```

wxing1292:
the cluster info is not required:
https://github.com/uber/cadence/blob/master/service/frontend/domainHandler.go#L117

the one thing that should be done is to drop the requirement of --global_domain flag, so if CLI not seeing that, set "IsGlobalDomain" in the register request to be null and let server do the check

vancexu:
> the cluster info is not required:
> https://github.com/uber/cadence/blob/master/service/frontend/domainHandler.go#L117
> 
> the one thing that should be done is to drop the requirement of --global_domain flag, so if CLI not seeing that, set "IsGlobalDomain" in the register request to be null and let server do the check

Agree. 
There is a small missing check in frontend leads to default not set correctly. 
After this PR in local domain register will be simple as previous `cadence --do samples-domain2 d re`

service/frontend/domainHandler.go
```diff
@@ -116,7 +116,7 @@ func (d *domainHandlerImpl) registerDomain(
 
 	activeClusterName := d.clusterMetadata.GetCurrentClusterName()
 	// input validation on cluster names
-	if registerRequest.ActiveClusterName != nil {
+	if len(registerRequest.GetActiveClusterName()) != 0 {
```
wxing1292: there is additional cluster name check below, plz change this back

service/frontend/domainHandler_GlobalDomainDisabled_test.go
```diff
@@ -165,6 +165,7 @@ func (s *domainHandlerGlobalDomainDisabledSuite) TestRegisterGetDomain_AllDefaul
 	err := s.handler.registerDomain(context.Background(), &shared.RegisterDomainRequest{
 		Name:                                   common.StringPtr(domainName),
 		WorkflowExecutionRetentionPeriodInDays: common.Int32Ptr(retention),
+		ActiveClusterName:                      common.StringPtr(""),
```
wxing1292: same as above

tools/cli/commands.go
```diff
@@ -183,9 +183,7 @@ func RegisterDomain(c *cli.Context) {
 		}
 	}
 	isGlobalDomain := false
```
wxing1292: this should be a pointer, and set to nil

tools/cli/commands.go
```diff
@@ -183,9 +183,7 @@ func RegisterDomain(c *cli.Context) {
...
-	if !c.IsSet(FlagIsGlobalDomain) {
-		ErrorAndExit(fmt.Sprintf("Option %s must be provided.", FlagIsGlobalDomain), err)
-	} else {
+	if c.IsSet(FlagIsGlobalDomain) {
 		isGlobalDomain, err = strconv.ParseBool(c.String(FlagIsGlobalDomain))
```
wxing1292: when the FlagIsGlobalDomain is set, then use this value to set the `isGlobalDomain` pointer

tools/cli/domain.go
```diff
@@ -95,7 +95,7 @@ func newDomainCommands() []cli.Command {
 				},
 				cli.StringFlag{
 					Name:  FlagIsGlobalDomainWithAlias,
-					Usage: "Flag to indicate whether domain is a global domain",
+					Usage: "Flag to indicate whether domain is a global domain, default to false if not provided",
```
wxing1292: there should be no default value

PR #2174, opened 2019-07-08
Replace archiver with provider 

yycptt:
This diff only replace history/visibility archiver in bootstrap params, frontend, history, worker with archiver provider.

Currently, it's not used anywhere. The switch from blobstore interface to archiver interface will be done later, after static config and domain record are updated.


host/onebox.go
```diff
@@ -424,7 +424,7 @@ func (c *cadenceImpl) startFrontend(hosts map[string][]string, startWG *sync.Wai
 	frontendConfig := frontend.NewConfig(dc, c.historyConfig.NumHistoryShards, c.workerConfig.EnableIndexer)
 	c.frontendHandler = frontend.NewWorkflowHandler(
 		c.frontEndService, frontendConfig, c.metadataMgr, c.historyMgr, c.historyV2Mgr,
-		c.visibilityMgr, kafkaProducer, params.BlobstoreClient, nil, nil)
+		c.visibilityMgr, kafkaProducer, params.BlobstoreClient, nil)
```
andrewjdawson2016: Can you file a task to write integration using archiver?

yycptt: Task Created. #2187

PR #2176, opened 2019-07-08
matching: make taskListID and taskListName partition aware

venkat1109:
This patch does 3 things:
 - makes taskListID type partition aware
-  introduces new type "qualifiedTaskListName" to represent internal task list names with metadata
- adds validation to prohibit tasklist names with reserved prefix

Related to #2098

### DIFF
--- a/service/history/decisionChecker.go
+++ b/service/history/decisionChecker.go
@@ -195,7 +195,7 @@ func (v *decisionAttrValidator) validateActivityScheduleAttributes(
 		return &workflow.BadRequestError{Message: "ScheduleActivityTaskDecisionAttributes is not set on decision."}
 	}
 
-	if err := v.validateTaskListName(attributes.TaskList, ""); err != nil {
+	if _, err := v.validatedTaskList(attributes.TaskList, ""); err != nil {
 		return err
 	}
 
@@ -496,9 +496,11 @@ func (v *decisionAttrValidator) validateContinueAsNewWorkflowExecutionAttributes
 	}
 
 	// Inherit Tasklist from previous execution if not provided on decision
-	if err := v.validateTaskListName(attributes.TaskList, executionInfo.TaskList); err != nil {
+	tl, err := v.validatedTaskList(attributes.TaskList, executionInfo.TaskList)
+	if err != nil {
 		return err
 	}
+	attributes.TaskList = tl
 
 	// Inherit workflow timeout from previous execution if not provided on decision
 	if attributes.GetExecutionStartToCloseTimeoutSeconds() <= 0 {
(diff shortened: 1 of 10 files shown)


service/history/decisionChecker.go
```diff
@@ -23,6 +23,8 @@ package history
 import (
 	"fmt"
```
shreyassrivatsan: remove newline.. run `make fmt`

service/history/decisionChecker.go
```diff
@@ -191,8 +195,8 @@ func (v *decisionAttrValidator) validateActivityScheduleAttributes(
 		return &workflow.BadRequestError{Message: "ScheduleActivityTaskDecisionAttributes is not set on decision."}
 	}
 
-	if attributes.TaskList == nil || attributes.TaskList.GetName() == "" {
-		return &workflow.BadRequestError{Message: "TaskList is not set on decision."}
+	if err := v.validateTaskListName(attributes.TaskList, ""); err != nil {
```
shreyassrivatsan: can you set this to be a const `noTasklist` or something?

service/history/decisionChecker.go
```diff
@@ -581,6 +581,31 @@ func (v *decisionAttrValidator) validateStartChildExecutionAttributes(
 	return nil
 }
 
+func (v *decisionAttrValidator) validateTaskListName(tl *workflow.TaskList, defaultVal string) error {
```
shreyassrivatsan: This method does more than just `validate` - it changes the passed in task list with the `defaultVal`.. Perhaps `validateAndSetTaskListName` ?

service/matching/tasklist.go
```diff
@@ -0,0 +1,150 @@
...
+// IsRoot returns true if this task list is a root partition
+func (tn *qualifiedTaskListName) IsRoot() bool {
+	return tn.partition == 0
+}
+
+// Parent returns the name of the parent task list
+// input:
+//   degree: Number of children at each level of the tree
+// Returns empty string if this task list is the root
+func (tn *qualifiedTaskListName) Parent(degree int) string {
```
shreyassrivatsan: `degree` must be > 1

shreyassrivatsan: Also, shouldn't degree be configured on the task list.. Why is this an input to the method?

service/history/decisionChecker.go
venkat1109: this code changed a bit, this comment is no longer relevant

service/matching/tasklist.go
venkat1109: if its not an input, then this type needs to take in the dynamic config func as input in the constructor. the latter option results in passing dynamic config everywhere in the unit test. But the actual Parent() method is only going to be called in one or two places. So, I  decided to pass this a param to this function.

service/history/decisionChecker.go
venkat1109: weird, I did run make fmt

venkat1109: fixed

service/history/decisionChecker.go
venkat1109: done

PR #2186, opened 2019-07-09
Refactor workflow events serialization path

wxing1292:
This is part of the on going refactor for workflow execution context / mutable state

* Introduce WorkflowEvents as serialized workflow events container
* Introduce ToSerializedEvents function to history builder
* Move serialization of workflow history events from history store / history store v2 to history builder
* WorkflowEvents, along with WorkflowMutation / WorkflowSnapshot will be generated as a result of transaction done on mutable state


common/persistence/dataInterfaces.go
```diff
@@ -760,6 +760,15 @@ type (
 		Encoding common.EncodingType // optional binary encoding type
 	}
 
+	// WorkflowEvents is used as generic workflow history events transaction container
+	WorkflowEvents struct {
```
venkat1109: How about:

- Naming this WorkflowEventBatch
- Then creating another type called WorkflowEvents { []WorkflowEventBatch} 

Each of the two types can have a Size() method which returns the size corresponding to the type. That way, you can avoid returning 3 parameters from methods like historyBuilder.ToSerializedEvents()

service/history/historyBuilder.go
```diff
@@ -1066,3 +1078,124 @@ func (b *historyBuilder) GetHistory() *workflow.History {
...
+	eventsSize := 0
+	workflowEventsSeq := []*persistence.WorkflowEvents{}
+
+	if len(b.transientHistory) != 0 {
+		serializedEvents, err := b.serializeEvents(branchToken, b.transientHistory, encoding)
+		if err != nil {
+			return nil, 0, err
+		}
+		if serializedEvents != nil {
+			eventsSize += len(serializedEvents.EventsBlob.Data)
```
venkat1109: see comment above about adding a Size() method

common/persistence/dataInterfaces.go
```diff
@@ -1209,12 +1217,12 @@ type (
 		Info string
 		// The branch to be appended
 		BranchToken []byte
-		// The batch of events to be appended. The first eventID will become the nodeID of this batch
-		Events []*workflow.HistoryEvent
+		// Node ID is the first event ID of the events blob
+		NodeID int64
```
yux0: Is this NodeID the same as the node id in events v2?

service/history/historyBuilder.go
```diff
@@ -1066,3 +1078,121 @@ func (b *historyBuilder) GetHistory() *workflow.History {
...
+	firstEventID := firstEvent.GetEventId()
+	lastEventID := lastEvent.GetEventId()
+	version := firstEvent.GetVersion()
+
+	blob, err := b.payloadSerializer.SerializeBatchEvents(events, encoding)
+	if err != nil {
+		return nil, err
+	}
+
+	// sanity check
```
yux0: can this move outside of serializeEvents?

wxing1292:
close in favor of #2196

PR #2188, opened 2019-07-09
Add README for Implementing Archiver Interface

yycptt:
#2181


common/archiver/README.md
```diff
@@ -0,0 +1,69 @@
...
+  // Record current progress
+  if featureCatalog.ProgressManager != nil {
+    if err := featureCatalog.ProgressManager.RecordProgress(ctx, progress); err != nil {
+      // log some error message and return error if needed. 
+    }
+  }
+}
+```
+
+4. The `archiver` package provides a utility class called `HistoryIterator` which is a wrapper of `HistoryReader`. Its usage is simpler than the `HistoryReader` given in the `BootstrapContainer`, so archiver implementations can choose to use it when reading workflow histories. Check the `historyIterator.go` file for more details. Sample usage can be found in the filestore historyArchiver implementation.
```
andrewjdawson2016: My understanding was that HistoryReader would go away and we only needed HistoryIterator now that we can reconstruct its state?

common/archiver/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+1. Create a new directory in the `archiver` folder. The structure should looks like the following:
+
+```
+./common/archiver
+  - filestore/                  -- Filestore implementation 
+  - provider/
+      - provider.go             -- Provider of archiver instances
+  - yourImplementation/
+      - historyArchiver.go      -- HistoryArchiver implementation
```
andrewjdawson2016: Add files for historyArchiver_test.go and visibilityArchiver_test.go

common/archiver/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Each archiver implementation should define a format of URI, which is used to describe the type of implementation and the target location of `Archive()` or `Get()` operation. For example, the URI for filestore uses the format: `file:///path/to/dir`. The `ValidateURI()` method is responsible for checking if a given string is a valid URI.
+
+The `Archive()` method should archive workflow histories or visibility records described by the archive request to the location specified by the URI, while the `Get()` method is responsible for retrieving those archived data.
+
+Since the `Archive()` method will be called from different places (for example, timerQueueProcessor and worker), it may want to do different things. Therefore, a list of `ArchiveOption` will be passed to this method. These options will be applied to an `ArchiveFeatureCatalog`, and by checking the fields of that catalog, the `Archive()` method can figure out what to do. Right now, the only feature in the catalog is `ProgressManager` which can be used to record and load archive progress. More features can be added if needed. The correct way to use it is shown in the code sample below.
```
andrewjdawson2016: I think its more instructive to say the "Archive method may be invoked differently by different callers. For examples some callers may automatically retry, while others only try once. Therefore, a list of ArchiveOptions..."

common/archiver/README.md
yycptt: Do you mean we should provide HistoryIterator instead of HistoryReader in the BootstrapParams? And then export the reset() method of HistoryIterator so that it can be called inside the Archive() method? 

But currently implementation of HistoryIterator is linked to a particular archive request. If that problem is solved, an other problem we need to address is that: the config for the historyIterator is stored in the provider, so the caller of the GetHistoryArchiver() should give the provider a historyReader and historyIterator will be created inside the provider. Or we store the config to history/worker service instead of provider. Both method seems weird to me.

common/archiver/README.md
yycptt: Done.

common/archiver/README.md
yycptt: Updated.

common/archiver/README.md
yycptt: My bad. It should be HistoryManager, not HistoryReader

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
+## What
+
+This README explains how to create new implementations for the History and Visibility Archiver Interface.
+
+## Steps
+
+1. Create a new directory in the `archiver` folder. The structure should looks like the following:
```
purplesusan: looks -> look

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+  - filestore/                      -- Filestore implementation 
+  - provider/
+      - provider.go                 -- Provider of archiver instances
+  - yourImplementation/
+      - historyArchiver.go          -- HistoryArchiver implementation
+      - historyArchiver_test.go     -- Unit tests for HistoryArchiver
+      - visibilityArchiver.go       -- VisibilityArchiver implementations
+      - visibilityArchiver_test.go  -- Unit tests for VisibilityArchiver
+```
+2. Each Archiver will be given a `BootstrapContainer` when it's created and the archiver implementation can (will) use them to read history, emit logs/metrics etc. Check `interface.go` for more details.
```
purplesusan: Check -> See

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+3. Implement the Archiver interface. Both Archiver interface contains three methods:
```
purplesusan: interface contains -> interfaces contain

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+  
+```go
+func Archive(ctx context.Context, URI string, request *ArchiveRequest, opts ...ArchiveOption) error
+
+func Get(ctx context.Context, URI string, request *GetRequest) (*GetResponse, error)
+
+func ValidateURI(URI string) error
+```
+
+Each archiver implementation should define a format of URI, which is used to describe the type of implementation and the target location of `Archive()` or `Get()` operation. For example, the URI for filestore uses the format: `file:///path/to/dir`. The `ValidateURI()` method is responsible for checking if a given string is a valid URI.
```
purplesusan: location of -> location of the

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+The `Archive()` method should archive workflow histories or visibility records described by the archive request to the location specified by the URI, while the `Get()` method is responsible for retrieving those archived data.
+
+`Archive()` method may be invoked differently by different callers. For examples some callers may automatically retry, while others only try once. Therefore, a list of `ArchiveOption` will be passed to this method. These options will be applied to an `ArchiveFeatureCatalog`, and by checking the fields of that catalog, the `Archive()` method can figure out what to do. Right now, the only feature in the catalog is `ProgressManager` which can be used to record and load archive progress. More features can be added if needed. The correct way to use it is shown in the code sample below.
```
purplesusan: `Archive()` method -> The `Archive()` method
For examples -> For example

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+  // Record current progress
+  if featureCatalog.ProgressManager != nil {
+    if err := featureCatalog.ProgressManager.RecordProgress(ctx, progress); err != nil {
+      // log some error message and return error if needed. 
+    }
+  }
+}
+```
+
+4. The `archiver` package provides a utility class called `HistoryIterator` which is a wrapper of `HistoryManager`. Its usage is simpler than the `HistoryManager` given in the `BootstrapContainer`, so archiver implementations can choose to use it when reading workflow histories. Check the `historyIterator.go` file for more details. Sample usage can be found in the filestore historyArchiver implementation.
```
purplesusan: Check -> See

common/archiver/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+5. The `archiver` package also provides some common error types and messages in `constants.go`.
+
+6. After you have finished the implementation, modify `./provider/provider.go` file so that the `ArchiverProvider` knows how to create an instance of your archiver. Also, add configs for you archiver to static yaml config files and modify the `HistoryArchival` and `VisibilityArchival` struct in the `../common/service/config.go` accordingly.
```
purplesusan: modify -> modify the

PR #2190, opened 2019-07-09
Add batch signal/cancel support

longquanzheng:


service/worker/batcher/workflow.go
```diff
@@ -422,10 +441,11 @@ func processSignalTask(ctx context.Context, limiter *rate.Limiter, task taskDeta
 			return err
 		}
 	}
+	activity.RecordHeartbeat(ctx, task.hbd)
 	return nil
 }
 
-func processTerminateTask(ctx context.Context, limiter *rate.Limiter, task taskDetail, batchParams BatchParams) error {
+func processTerminateOrCancelTask(ctx context.Context, limiter *rate.Limiter, task taskDetail, batchParams BatchParams) error {
```
shreyassrivatsan: Can we change this to something like the below.. So that we separate out the child logic from the actual task which is a just a simple RPC call.. 

```
func processChildRecursiveTask(ctx context.Context,... , terminateChild bool, task func() error)
```

and call this with 
```
processChildRecursiveTask(ctx, ..., terminateChildren, func() error {
return batcher.svcClient.TerminateWorkflowExecution(newCtx, &shared.TerminateWorkflowExecutionRequest{
			WorkflowExecution: &wf,					Domain:            common.StringPtr(batchParams.DomainName),
			Reason:            common.StringPtr(batchParams.Reason),					WorkflowExecution: &wf,
			Identity:          common.StringPtr(batchWFTypeName),					Reason:            common.StringPtr(batchParams.Reason),
		})					Identity:          common.StringPtr(batchWFTypeName),
			})

})
```

longquanzheng: That will work for the serice call but not the other part -- How about the TeriminateParams and CancelParams? Maybe combine one? Right now they only have one : terminateChildren and cancelChildren. We can combine them into one but could be a little confusing as the WF parameter. If we called it childPolicy(true/false/nil), it is incorrect because it doen't apply to reset/signal.

shreyassrivatsan: You can just pass in a boolean to the `processChildRecursiveTask` - `applyOnChildren`.. That isnt something that depends on whether it is cancel or terminate.

longquanzheng: sounds good

service/worker/batcher/workflow.go
```diff
@@ -304,7 +333,19 @@ func startTaskProcessor(
 
 			switch batchParams.BatchType {
 			case BatchTypeTerminate:
-				err = processTerminateTask(ctx, limiter, task, batchParams, client)
+				err = processChildRecursiveTask(ctx, limiter, task, batchParams, client,
```
shreyassrivatsan: So looking at this further.. Everything can go through the same method. When `applyOnChild` is null, it will only act on current, else will recursively look at children. So even signal can go through same path with `applyOnChild = null`

service/worker/batcher/workflow.go
```diff
@@ -326,8 +367,28 @@ func startTaskProcessor(
...
+		_, ok := err.(*shared.EntityNotExistsError)
+		if !ok {
+			return err
+		}
+	}
+	activity.RecordHeartbeat(ctx, task.hbd)
+	return nil
+}
+
+func processChildRecursiveTask(
```
shreyassrivatsan: stack params here. one param per line if it cannot fit in a single line.

service/worker/batcher/workflow.go
```diff
@@ -304,7 +333,23 @@ func startTaskProcessor(
 
 			switch batchParams.BatchType {
 			case BatchTypeTerminate:
-				err = processTerminateTask(ctx, limiter, task, batchParams, client)
+				err = processChildRecursiveTask(ctx, limiter, task, batchParams, client,
```
shreyassrivatsan: nit: can just be named `processTask` now..

longquanzheng: Makes sense. Thank you so much for those useful feedbacks to improve the code quality. Learned a lot!

Issue #2092, opened 2019-06-24
Archival: Update Domain Record


yycptt:
Task 5 of https://github.com/uber/cadence/issues/2082
 
1. Update the schema of domain record:
    Add the following fields:
        Visibility Status
        Visibility URI
        History Status
        History URI
    Delete existing archival related fields
 
2. Update register, update, describe domain (including the cli part)
3. Again at the time temporary store bucket name in history_uri field

Issue #2091, opened 2019-06-24
Archival: Update Static Cluster Level Config


yycptt:
Task 4 of https://github.com/uber/cadence/issues/2082
 
1. Update cluster level config to the following:
```
archival:
	visibility:
    	    write_status: {enabled, paused, disabled} // has dynamic config overwrite
    	    read_enabled: {true, false} // has dynamic config overwrite
    	    s3:
                // config needed to connect to S3 (timeouts, creds etc…)
    	       // others
	history:
    	    write_status: {enabled, paused, disabled} // has dynamic config overwrite
    	    read_enabled: {true, false} // has dynamic config overwrite
    	    s3:
        	// config needed to connect to S3 (timeouts, creds etc…)
    	        // others
domain:
	archival:
    	    visibility:
        	default_uri: // some default URI
        	default_status: {true, false}
    	    history:
        	default_uri: // some default URI
        	default_status: {true, false}
```
 

2. Also create dynamic config for the fields that should support overwrite
3. Initialize archiver on start up in /cmd/server.go
4. Update register, update domain
5. At this time temporary store bucket name in default_uri field

PR #2191, opened 2019-07-09
Archival: Update Cluster Static Config

yycptt:
#2091 

This diff does not update domain config. Logic in register, update domain related to archival has been deleted. Related tests are also deleted. These things will be updated as part of #2092.


common/cluster/archival_status.go
```diff
@@ -87,6 +82,16 @@ func (a *ArchivalConfig) EnableReadFromArchival() bool {
 	return a.enableReadFromArchival
 }
 
+func (a *ArchivalConfig) isDomainDefaultValid() bool {
+	if a.domainDefaultStatus == ArchivalPaused {
+		return false
+	}
+
+	URISet := len(a.domainDefaultURI) != 0
```
yycptt: I think we should also check if the URI is valid here. But before I do that, I think we should not make ValidateURI a method of an archiver instance. Instead if should be an exported function of the archiver package. Otherwise, whenever we want to validate a URI, we need to create an archiver instance. We can also consider add another exported method to the provider package to validate any given URI, regardless of their scheme.

common/service/config/config.go
```diff
@@ -283,16 +282,26 @@ type (
...
-		// DefaultBucket is the default bucket used for archival in case domain does not specify override
-		DefaultBucket string `yaml:"defaultBucket"`
-		// Filestore the configuration for file based blobstore
-		Filestore *filestore.Config `yaml:"filestore"`
-		// S3store the configuration for amazon s3 based blobstore
-		S3store *s3store.Config `yaml:"s3store"`
+	}
+
+	// VisibilityArchival contains the config for visibility archival
+	VisibilityArchival struct {
```
yycptt: Although this struct currently contains the same fields as the HistoryArichival struct above, they will be different as there will be different implementation for history and visibility archival and their configs (will be new field in these structs) are different.

common/archiver/provider/provider.go
```diff
@@ -25,20 +25,21 @@ import (
...
 	"github.com/uber/cadence/common/archiver"
 	"github.com/uber/cadence/common/archiver/filestore"
-	"github.com/uber/cadence/common/log/tag"
 )
 
 var (
 	// ErrUnknownScheme is the error for unknown archiver scheme
 	ErrUnknownScheme = errors.New("unknown archiver scheme")
+	// ErrEmptyBootStrapContainer is the error for empty bootstrap container
+	ErrEmptyBootStrapContainer = errors.New("empty bootstrap container")
```
andrewjdawson2016: "empty bootstrap container can only be provided if archiver has already been created"

common/archiver/provider/provider.go
```diff
@@ -64,51 +63,43 @@ type (
...
+	}
 
 	switch scheme {
 	case filestore.URIScheme:
-		container := *p.historyContainer
-		container.Logger = container.Logger.WithTags(tag.Service(serviceName)).WithTags(tags...)
-		historyArchiver := filestore.NewHistoryArchiver(container, p.historyArchiverConfigs.FileStore)
-		p.historyArchivers[key] = historyArchiver
-		return historyArchiver, nil
+		p.historyArchivers[serviceName] = filestore.NewHistoryArchiver(*container, p.historyArchiverConfigs.FileStore)
```
andrewjdawson2016: Why is this serviceName and not key?

common/archiver/provider/provider.go
```diff
@@ -64,51 +63,43 @@ type (
...
+		return nil, ErrEmptyBootStrapContainer
+	}
 
 	switch scheme {
 	case filestore.URIScheme:
-		visibilityArchiver := filestore.NewVisibilityArchiver(*p.visibilityContainer, p.visibilityArchiverConfigs.FileStore)
-		p.visibilityArchivers[key] = visibilityArchiver
-		return visibilityArchiver, nil
+		p.visibilityArchivers[serviceName] = filestore.NewVisibilityArchiver(*container, p.visibilityArchiverConfigs.FileStore)
+		return p.visibilityArchivers[serviceName], nil
```
andrewjdawson2016: Key?

common/archiver/provider/provider.go
```diff
@@ -64,51 +63,43 @@ type (
...
-		visibilityContainer:       visibilityContainer,
 		historyArchiverConfigs:    historyArchiverConfigs,
 		visibilityArchiverConfigs: visibilityArchiverConfigs,
 	}
 }
 
-func (p *archiverProvider) GetHistoryArchiver(scheme, serviceName string, tags ...tag.Tag) (archiver.HistoryArchiver, error) {
-	key := p.getArchiverKey(scheme, serviceName)
-	if historyArchiver, ok := p.historyArchivers[key]; ok {
+func (p *archiverProvider) GetHistoryArchiver(scheme, serviceName string, container *archiver.HistoryBootstrapContainer) (archiver.HistoryArchiver, error) {
```
andrewjdawson2016: I kind of think it makes sense now to have CreateHistoryArchiver and CreateVisibilityArchiver methods. The reason is its kind of weird to understand when you can or cannot provide BootstrapContainer. For example right now in if archiver already exists and bootstrapContainer is provided the client will get back an archiver but it will not be based on the bootstrap container they provided. Thoughts?

common/cluster/archival_status.go
```diff
@@ -47,38 +48,32 @@ const (
...
+		domainDefaultURI:       domainDefaultURI,
 	}
-	if !ac.isValid() {
-		return &ArchivalConfig{
-			status:                 ArchivalDisabled,
-			defaultBucket:          "",
-			enableReadFromArchival: false,
-		}
+	if !ac.isDomainDefaultValid() {
+		ac.domainDefaultStatus = ArchivalDisabled
```
andrewjdawson2016: If the domain is not valid lets also set the configs for cluster: status and enableReadFromArchival to their disabled states. To make this more clear lets include a private helper method `disabledConfig()`

common/cluster/archival_status.go
```diff
@@ -47,38 +48,32 @@ const (
...
 	return a.status
 }
 
 // ConfiguredForArchival returns true if cluster is configured to handle archival, false otherwise.
-// If cluster is configured for archival then defaultBucket will be set.
 func (a *ArchivalConfig) ConfiguredForArchival() bool {
-	if !a.isValid() {
-		return false
-	}
 	return a.status != ArchivalDisabled
```
andrewjdawson2016: Is is a mistake on my part, but lets just treat ArchivalPaused as not configured either. Currently I want paused == disabled throughout the system.

common/cluster/archival_status.go
```diff
@@ -87,6 +82,16 @@ func (a *ArchivalConfig) EnableReadFromArchival() bool {
 	return a.enableReadFromArchival
 }
 
+func (a *ArchivalConfig) isDomainDefaultValid() bool {
+	if a.domainDefaultStatus == ArchivalPaused {
```
andrewjdawson2016: Lets create separate types for clusterArchivalStatus and domainArchivalStatus. I think it will improve clarity as we later plumb this config throughout the code.

common/cluster/archival_status.go
andrewjdawson2016: It's a good thought but may not be easy. Consider the case of S3 - to check if a bucket is valid you actually need to make a call to blobstore to see if it exists. In the case of filestore we can do validation just based on a regex of the URI but that is not true of all archiver implementations. I think its reasonable that validateURI is part of archiver methods and just rely on calling that in frontend whenever a domain is actually enabled for archival.

common/cluster/archival_status.go
```diff
@@ -32,8 +32,9 @@ type (
 	// ArchivalConfig is an immutable representation of the current archival configuration of the cluster
 	ArchivalConfig struct {
 		status                 ArchivalStatus
-		defaultBucket          string
 		enableReadFromArchival bool
+		domainDefaultStatus    ArchivalStatus
```
andrewjdawson2016: Looks like you do not have getters for these methods?

andrewjdawson2016: I think now that we have two types of entities in this archivalConfig: domainDefaults and clusterDefaults the getters for this file might look a little weird. Lets revisit what these getters should look like when we make domain CRUD changes to workflowHandler.go

common/cluster/metadata.go
```diff
@@ -127,19 +131,41 @@ func NewMetadata(
 		panic("Cluster info initial versions have duplicates")
 	}
 
-	status, err := getArchivalStatus(archivalStatus)
+	historyArchivalStatus := dc.GetStringProperty(dynamicconfig.HistoryArchivalStatus, archivalClusterConfig.History.Status)()
```
andrewjdawson2016: Yes :) I prefer having this here instead of server.go

common/cluster/archival_status.go
andrewjdawson2016: One option is to introduce two new structs clusterConfig and domainDefaultConfig. Then just have getters to these two methods and have these structs themselves have the needed methods on them.

common/service/config/config.go
```diff
@@ -61,6 +58,8 @@ type (
 		// DynamicConfigClient is the config for setting up the file based dynamic config client
 		// Filepath should be relative to the root directory
 		DynamicConfigClient dynamicconfig.FileBasedClientConfig `yaml:"dynamicConfigClient"`
+		// Domain is the default config for every domain
+		Domain Domain `yaml:"domain"`
```
andrewjdawson2016: Let's get another opinion about this name. Can you ask Samar his opinion.

common/service/config/config.go
```diff
@@ -283,16 +282,26 @@ type (
 
 	// Archival contains the config for archival
 	Archival struct {
-		// Status is the status of archival either: enabled, disabled, or paused
+		// History is the config for the history archival
+		History HistoryArchival `yaml:"historyArchival"`
```
andrewjdawson2016: Where does one give archiver connection info?

common/cluster/archival_status.go
yycptt: Will improve this part when modifying frontend domain handler as it's not clear now how these fields will be used.

common/archiver/provider/provider.go
yycptt: Since now we have the create method,  I will keep the current error message as it should always be not nil.

config/development.yaml
```diff
@@ -82,19 +82,27 @@ dcRedirectionPolicy:
...
-    defaultBucket: "cadence-development"
-    customBuckets:
-      - "custom-bucket-1"
-      - "custom-bucket-2"
-#  s3store:
-#    region: "us-east-1"
-#    endpoint: "http://127.0.0.1:4572"
-#    s3ForcePathStyle: true
+  history:
+    status: "disabled"
```
yycptt: I need to disable it for now since the reset part of the system is still using blobstore, which will be nil.

common/cluster/archival_status.go
```diff
@@ -29,11 +29,15 @@ type (
 	// ArchivalStatus represents the archival status of the cluster
 	ArchivalStatus int
```
yycptt: I keep the naming of the cluster level config as otherwise it will be referred as cluster.ClusterArchivalStatus outside the package.

common/cluster/archival_status.go
```diff
@@ -44,50 +48,53 @@ const (
...
+		return newDisabledArchivalConfig()
 	}
 	return ac
 }
 
-// GetDefaultBucket returns the default bucket for ArchivalConfig
-func (a *ArchivalConfig) GetDefaultBucket() string {
-	return a.defaultBucket
+// ClusterConfiguredForArchival returns true if cluster is configured to handle archival, false otherwise.
+func (a *ArchivalConfig) ClusterConfiguredForArchival() bool {
```
yycptt: Keep only this method for now. More methods will be added if necessary when I update the frontend domain handler

common/service/config/config.go
samarabbas: Looks like I'm missing some context here.  Anything which is overridden by domain usually goes to DynamicConfig rather than static yaml based config.  Our DynamicConfig already has all the infrastructure needed for providing defaults are overriding those defaults per domain.  Let's have a quick discussion before adding this to static config.

samarabbas: Use case here is more about various defaults used for domains.  I would rather rename this to DomainDefaults.

common/archiver/provider/provider.go
```diff
@@ -25,20 +25,27 @@ import (
...
 type (
 	// ArchiverProvider returns history or visibility archiver based on the scheme and serviceName.
-	// The archiver for each combination of scheme and serviceName will be created only once and cached.
+	// Create() method will create a new archiver (overwritting the existing one) and cache it.
+	// Get() method will find the existing archiver for a given scheme and serviceName pair and return it.
+	// If that archiver does not exist, an ErrArchiverNotExist error will be returned.
 	ArchiverProvider interface {
-		GetHistoryArchiver(scheme string, serviceName string, tags ...tag.Tag) (archiver.HistoryArchiver, error)
-		GetVisibilityArchiver(scheme string, serviceName string, tags ...tag.Tag) (archiver.VisibilityArchiver, error)
+		CreateHistoryArchiver(scheme string, serviceName string, container *archiver.HistoryBootstrapContainer) (archiver.HistoryArchiver, error)
```
yycptt: I will change the provider interface as we discussed, in the domain config diff. No need to review this part now.

common/archiver/historyIterator.go
```diff
@@ -114,7 +116,10 @@ func NewHistoryIterator(
 		shardID:           request.ShardID,
 		eventStoreVersion: request.EventStoreVersion,
 		branchToken:       request.BranchToken,
-		config:            config,
+		config: &historyIteratorConfig{
+			historyPageSize:       historyPageSize,
+			targetHistoryBlobSize: targetHistoryBlobSize,
```
yycptt: I'd like to flatten these two fields to the root level of historyIterator.

common/archiver/provider/provider.go
```diff
@@ -25,86 +25,109 @@ import (
...
 	switch scheme {
 	case filestore.URIScheme:
-		container := *p.historyContainer
-		container.Logger = container.Logger.WithTags(tag.Service(serviceName)).WithTags(tags...)
-		historyArchiver := filestore.NewHistoryArchiver(container, p.historyArchiverConfigs.FileStore)
-		p.historyArchivers[key] = historyArchiver
+		if p.historyArchiverConfigs.Filestore == nil {
+			return nil, ErrArchiverConfigNotFound
+		}
+		historyArchiver, err := filestore.NewHistoryArchiver(*container, p.historyArchiverConfigs.Filestore)
```
andrewjdawson2016: :)

common/cluster/archival_status.go
andrewjdawson2016: makes sense

common/cluster/archival_status.go
```diff
@@ -44,50 +48,53 @@ const (
...
+		DomainDefaultStatus:    domainDefaultStatus,
+		DomainDefaultURI:       domainDefaultURI,
 	}
-	if !ac.isValid() {
-		return &ArchivalConfig{
-			status:                 ArchivalDisabled,
-			defaultBucket:          "",
-			enableReadFromArchival: false,
-		}
+	if !ac.isDomainDefaultValid() {
```
andrewjdawson2016: Can we still call this method isValid? We are checking that the whole ArchivalConfig is valid its just that the only condition that we need to actually check relates to domain fields.

common/cluster/archival_status.go
andrewjdawson2016: Makes sense.

config/development.yaml
```diff
@@ -82,19 +82,27 @@ dcRedirectionPolicy:
...
+    enableReadFromArchival: false
+    archiverProvider:
```
andrewjdawson2016: This feels to me like it should be invalid config. If cluster is disabled then providing any archiverProviders should panic on startup.

config/development.yaml
```diff
@@ -82,19 +82,27 @@ dcRedirectionPolicy:
...
+      filestore:
+        fileMode: "0600"
+        dirMode: "0700"
+      # s3store:
```
andrewjdawson2016: Lets not include this S3 for the time being. I suspect its going to look different than this now that we are using archiver not blobstore. I don't want this to be misleading.

config/development.yaml
```diff
@@ -82,19 +82,27 @@ dcRedirectionPolicy:
...
+      #   endpoint: "http://127.0.0.1:4572"
+      #   s3ForcePathStyle: true
+  visibility:
+    status: "disabled"
+    enableReadFromArchival: false
+
+domainDefaults:
+  archival:
+    history:
+      defaultStatus: "disabled"
```
andrewjdawson2016: In a similar way to avoid it should be invalid to provide a URI if status is disabled. Cluster should panic on startup if this config is given.

yycptt: Will be done in the domain config change diff.

common/cluster/archival_status.go
yycptt: I think the method name show the fact that we are only check domain defaults and any config for the cluster level config is valid.

common/archiver/historyIterator.go
```diff
@@ -95,7 +93,7 @@ func NewHistoryIterator(
 	request *ArchiveHistoryRequest,
 	historyManager persistence.HistoryManager,
 	historyV2Manager persistence.HistoryV2Manager,
-	config *HistoryIteratorConfig,
+	targetHistoryBlobSize int,
 	initialState []byte,
 	sizeEstimator SizeEstimator,
```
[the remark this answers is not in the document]
yycptt: Will create a separate diff for this issue.

PR #2195, opened 2019-07-10
Add filtering in list results

vancexu:
ES has a known issue that range query on nanoseconds doesn't work as expected, so we previously add 1 ms resolution to old APIs like ListOpen/CloseWorkflow.

But this may surprise users such as 
image
first record has closeTime 156277796496704**7**000 but query request has smaller upper bound 156277796496704**6**000

This change ensure user see correct records, so old APIs behavior is consistence in both ES and Cassandra
image
 
NOTE:
This will cause page not always full (will be a breaking change if user somehow has such assumption)


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -217,7 +227,12 @@ func (v *esVisibilityStore) ListOpenWorkflowExecutionsByType(
 		}
 	}
 
-	return v.getListWorkflowExecutionsResponse(searchResult.Hits, token, definition.StartTime, request.PageSize)
+	isRecordValid := func(rec *p.VisibilityWorkflowExecutionInfo) bool {
```
yycptt: Looks like it's the same function as L181. Declare it as a normal function instead of a lambda function so it can be reused?

yycptt: or declare it in getListWorkflowExecutionsResponse() and pass isOpen to that function?

PR #2200, opened 2019-07-11
Refactor workflow execution context & mutable state

wxing1292:
* Introduce WorkflowEvents, which contains events to be persisted
* Cleanup workflow execution continue as new path, now all create workflow persistence request is generated using mutable state
* Remove appendFirstBatchHistoryForContinueAsNew, handle the persistence of continue as new history events in update method
* Move mutable state helper functions to dedicated file
* Introduce CloseTransactionAsSnapshot as close function for workflow creation
* Introduce CloseTransactionAsMutation as close function for workflow update


wxing1292:
Note: the mutable state / workflow execution context re-write is still ongoing, meaning that there are things half done

wxing1292:
@yycptt plz verify the continue as new path (especially the cron related logic)

service/history/mutableStateBuilder.go
```diff
@@ -3629,6 +3490,272 @@ func (e *mutableStateBuilder) GetContinueAsNew() *persistence.WorkflowSnapshot {
...
+	e.deleteSignalRequestedID = ""
+	e.continueAsNew = nil
+	e.continueAsNewWorkflowEvents = nil
+	e.clearBufferedEvents = false
+	if e.updateBufferedEvents != nil {
+		e.bufferedEvents = append(e.bufferedEvents, e.updateBufferedEvents...)
+		e.updateBufferedEvents = nil
+	}
+	if len(e.bufferedEvents) > e.config.MaximumBufferedEventsBatch() {
+		return ErrBufferedEventsLimitExceeded
```
venkat1109: what does it mean when you return error at this stage ? What is the application supposed to do ? Can we return this error much earlier before we even flush events etc i.e. check if len(buffered) + len(updateBuffered) > limit ?

service/history/mutableStateUtil.go
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
venkat1109: s/2017/2019

service/history/mutableStateUtil.go
```diff
@@ -0,0 +1,185 @@
...
+
+import (
+	"github.com/uber/cadence/common/persistence"
+)
+
+func convertPendingActivityInfos(
+	inputs map[int64]*persistence.ActivityInfo,
+) []*persistence.ActivityInfo {
+
+	outputs := []*persistence.ActivityInfo{}
```
venkat1109: you know the size beforehand. Allocate exactly len(inputs) capacity slice. Same comment for rest of the methods

service/history/stateBuilder.go
```diff
@@ -129,6 +129,10 @@ func (b *stateBuilderImpl) applyEvents(domainID, requestID string, execution sha
 		switch event.GetEventType() {
 		case shared.EventTypeWorkflowExecutionStarted:
 			attributes := event.WorkflowExecutionStartedEventAttributes
+			domainEntry, err := b.domainCache.GetDomainByID(domainID)
+			if err != nil {
+				return nil, nil, nil, err
```
venkat1109: this method signature definitely needs some revisiting. It takes about 6 params and returns 4 results

service/history/mutableStateBuilder.go
```diff
@@ -3629,6 +3490,272 @@ func (e *mutableStateBuilder) GetContinueAsNew() *persistence.WorkflowSnapshot {
...
+	}
+
+	e.hasBufferedEventsInPersistence = len(e.bufferedEvents) > 0
+	e.condition = e.GetNextEventID()
+	return nil
```
yycptt: reset pendingTimertasks etc.?

service/history/mutableStateBuilder.go
```diff
@@ -83,14 +83,21 @@ type (
...
+
+		// in memory only attributes
+		// indicates whether there are buffered events in persistence
 		hasBufferedEventsInPersistence bool
+		// indicates the previous next event ID
+		condition int64
+
+		pendingTransferTasks    []persistence.Task
+		pendingReplicationTasks []persistence.Task
+		pendingTimerTasks       []persistence.Task
```
yycptt: Can we rename these three fields to something like transactionTransferTasks?

service/history/mutableStateBuilder.go
```diff
@@ -3629,6 +3490,272 @@ func (e *mutableStateBuilder) GetContinueAsNew() *persistence.WorkflowSnapshot {
...
+func (e *mutableStateBuilder) AddTransferTasks(transferTasks ...persistence.Task) {
+	e.pendingTransferTasks = append(e.pendingTransferTasks, transferTasks...)
+}
+
+// TODO convert AddTransferTasks to prepareTimerTasks
+func (e *mutableStateBuilder) AddTimerTasks(timerTasks ...persistence.Task) {
+	e.pendingTimerTasks = append(e.pendingTimerTasks, timerTasks...)
+}
+
+func (e *mutableStateBuilder) ToWorkflowMutation(now time.Time) (*persistence.WorkflowMutation, []*persistence.WorkflowEvents, error) {
```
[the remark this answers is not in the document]
wxing1292: sure, but this function is not commit, not yet. this function aims to generate objects to be persisted.

one more thing, this function is generating a diff, there is another function which generates a snapshot.

service/history/mutableStateBuilder.go
```diff
@@ -3629,6 +3490,272 @@ func (e *mutableStateBuilder) GetContinueAsNew() *persistence.WorkflowSnapshot {
...
+		Condition: e.condition,
+	}
+
+	if err := e.closeSession(); err != nil {
+		return nil, nil, err
+	}
+	return workflowMutation, workflowEventsSeq, nil
+}
+
+func (e *mutableStateBuilder) ToWorkflowSnapshot(now time.Time) (*persistence.WorkflowSnapshot, []*persistence.WorkflowEvents, error) {
```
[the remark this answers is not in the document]
wxing1292: if commit is the key word (meaning generating a snapshot or diff but does not persist it to DB), what about CommitMutation and CommitSnapshot?

service/history/mutableStateUtil.go
wxing1292: sure

service/history/mutableStateUtil.go
wxing1292: sure

service/history/stateBuilder.go
wxing1292: agreed, let us do it one thing at a time :)

service/history/mutableStateBuilder.go
wxing1292: right now this is handled by failing the decision
```
c.failInflightDecision()
```
ref: https://github.com/uber/cadence/blob/master/service/history/workflowExecutionContext.go#L750

The refactor need to incrementally change related logic

service/history/mutableStateBuilder.go
```diff
@@ -3629,6 +3490,272 @@ func (e *mutableStateBuilder) GetContinueAsNew() *persistence.WorkflowSnapshot {
...
+	e.pendingReplicationTasks = append(e.pendingReplicationTasks,
+		convertSyncActivityInfos(
+			e.pendingActivityInfoIDs,
+			e.syncActivityTasks,
+		)...)
+
+	return workflowEventsSeq, nil
+}
+
+func (e *mutableStateBuilder) eventsToReplicationTask(events []*workflow.HistoryEvent) []persistence.Task {
```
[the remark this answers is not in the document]
wxing1292: if changing the signature, then this function can return nil and caller need to do nil check.

service/history/mutableStateBuilder.go
wxing1292: good catch

service/history/mutableStateBuilder.go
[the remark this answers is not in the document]
yycptt: Maybe something like CreateWorkflowMutation, GenerateWorkflowMutation, CreateXXXDiff, etc.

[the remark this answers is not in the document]
wxing1292: change to 
CloseTransactionForWorkflowSnapshot
and 
CloseTransactionForWorkflowMutation

PR #2206, opened 2019-07-12
Instructions for docker-compose in production

longquanzheng:


docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+   ```
+  b. Go to the domains table directory and clear all data files(**NOT** folders)
+  ```
+  rm data/cadence/domains-aec8fcd0a4ee11e98c689b89c16b65a4/*.*
+  ``` 
+  c. Copy the snapshots to the table directory:
+  ```
+  cp data/cadence/domains-aec8fcd0a4ee11e98c689b89c16b65a4/snapshots/my-test-backup/* data/cadence/domains-aec8fcd0a4ee11e98c689b89c16b65a4/
+  ```
+  d. Finally restart the Cassandra container and Cadence service
```
purplesusan: add a comma after Finally
add a period at the end

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+  d. Finally restart the Cassandra container and Cadence service   
+
+Typical production setup
+=========================
+In a typical production setting, dependencies (Cassandra / MySQL and Statsd / Prometheus server) are
+managed / started independently of the cadence-server. 
+The recommended way is to use [Cadence Helm Chart](https://hub.helm.sh/charts/banzaicloud-stable/cadence), if you have Kubernetes.
+
+ 
+For hooking up with other deployment platform, you can use the following command:
```
purplesusan: For hooking up with -> To hook into 
platform -> platforms

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
 
 Please note that SQL support is still in active developement and it is not production ready yet.
 
-Quickstart for production
+Onebox production setup with docker-compose
```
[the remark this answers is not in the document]
longquanzheng: I see. Let's use One Host then. I used OneBox because this is what we call it internally in AWS.

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
-a production setting, use the following command:
+For a simple production setup, start with our **docker-compose-production-example.yml** for a single host Cadence cluster as the follow.
 
+#### 1. Modify the database mounting path in the example.yaml
+You may want to set the Cassandra data location because:
+ 1) You need a disk partition that has enough space
+ 2) It is better to have a separated storage hardware for database for performance, backup plan and availability(with RAID)
+ 3) Setting a mounting path can avoid losing data when you run **docker-compose down**
+
+Save it into docker-compose.yaml.
```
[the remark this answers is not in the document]
longquanzheng: How about:
Save the edited example YAML file as **docker-compose.yaml**.  ?

docker/README.md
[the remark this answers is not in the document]
purplesusan: cool, thanks.

docker/README.md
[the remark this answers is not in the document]
purplesusan: That works. Thanks!

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+Onebox production setup with docker-compose
 =========================
-In a typical production setting, dependencies (cassandra / statsd server) are
-managed / started independently of the cadence-server. To use the container in
-a production setting, use the following command:
+For a simple production setup, start with our **docker-compose-production-example.yml** for a single host Cadence cluster as the follow.
 
+#### 1. Modify the database mounting path in the example.yaml
+You may want to set the Cassandra data location because:
+ 1) You need a disk partition that has enough space
```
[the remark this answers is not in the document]
longquanzheng: 1. is already taken by "#### 1. Modify the database mou..."
Can I use * a. ?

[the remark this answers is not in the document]
purplesusan: Yes, a. would be a good alternative.

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
 =========================
-In a typical production setting, dependencies (cassandra / statsd server) are
-managed / started independently of the cadence-server. To use the container in
-a production setting, use the following command:
+For a simple production setup, start with our **docker-compose-production-example.yml** for a single host Cadence cluster as described in the following steps.
 
+#### 1. Modify the database mounting path in the example YAML file
+You may want to set the Cassandra data location because:
+* a. You need a disk partition that has enough space
+* b. You prefer to have separated storage hardware for have a separated storage hardware for database for performance, backup plan, and availability (with RAID)
```
dmetzgar: Rewrite to "b. Separate storage hardware for your database gives you better control over performance, availability, and backup strategy"

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+* c. Setting a mounting path can avoid losing data when you run **docker-compose down**
+
+Save the edited example YAML file as docker-compose.yaml.
+
+See the [Docker Documentation](https://docs.docker.com/compose/production/) for more information about using docker-compose in production.
```
dmetzgar: Seems this section belongs in the section below.

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
 
-Quickstart for production
+One host production setup with docker-compose
 =========================
-In a typical production setting, dependencies (cassandra / statsd server) are
-managed / started independently of the cadence-server. To use the container in
-a production setting, use the following command:
+For a simple production setup, start with our **docker-compose-production-example.yml** for a single host Cadence cluster as described in the following steps.
 
+#### 1. Modify the database mounting path in the example YAML file
```
dmetzgar: This heading style doesn't match the rest of the file. Can you change the other headings to use the `# Heading` style?

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+* c. Modify the version tag in Cadence server to the new release in docker-compose.yaml. Then run it again:
+```
+docker-compose up -d
+```
+
+#### 4. Backup & Restore
+* **Backup**: Use nodetool [snapshot](http://cassandra.apache.org/doc/latest/tools/nodetool/snapshot.html) command to create hot snapshots.
+ To access nodetool, you can install nodetool, or log on to the Cassandra container. Get your Cassandra docker container ID by:
+ ```
+ docker ps
```
dmetzgar: Put a `$` in front of the command so that it's easier to distinguish the command from its output.

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+* b. Shut down docker-compose by using the following command:
+```
+docker-compose stop
+```
+* c. Modify the version tag in Cadence server to the new release in docker-compose.yaml. Then run it again:
+```
+docker-compose up -d
+```
+
+#### 4. Backup & Restore
```
dmetzgar: Backup and restore should be their own headings instead of bullet points under this heading.

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
 
 Please note that SQL support is still in active developement and it is not production ready yet.
 
-Quickstart for production
+One host production setup with docker-compose
 =========================
-In a typical production setting, dependencies (cassandra / statsd server) are
-managed / started independently of the cadence-server. To use the container in
-a production setting, use the following command:
+For a simple production setup, start with our **docker-compose-production-example.yml** for a single host Cadence cluster as described in the following steps.
```
dmetzgar: There are a couple things missing here. The first is that there is no explanation of why the yaml file is set to 0.5.9 version of Cadence. If that's done on purpose to have them walk through the upgrade scenario, then call that out.

The other thing missing is some indication of when the schema upgrade is needed. Is there anything they should look for in the release notes? Should they consider a change in minor version to indicate a schema change?

docker/README.md
```diff
@@ -62,12 +62,90 @@ docker-compose -f docker-compose-mysql.yml down
...
+
+#### 2. Run docker-compose
+```
+docker-compose up -d
+```
+Try out our [samples](https://github.com/uber-common/cadence-samples) to make sure the server is running well.
+
+#### 3. Upgrade Cadence server
+* a. First, you need to upgrade the database schema. You can find instructions for using the cadence-cassandra-tool to upgrade schema [here](https://github.com/uber/cadence/tree/master/tools/cassandra#update-schema-as-part-of-a-release).
+* b. Shut down docker-compose by using the following command:
```
dmetzgar: We can't do in-place upgrades?

docker/README.md
dmetzgar: Also, we should have a MySQL version of the docker-compose. Most people will be using MySQL.

longquanzheng: I would update that 0.5.9 to 0.x.y-auto-setup and let them choose the latest one to start with.
 
Currently there is no place to know whether or not a schema changes is needed. I will include that into our release process. Let us make sure very release telling it. If they run with old schema, the server won't start successfully. 

Agreed that we should have a MySQL version. Issue created https://github.com/uber/cadence/issues/2406 We need to practice how to use the commands to do backup like Cassandra

docker/README.md
longquanzheng: Done

docker/README.md
longquanzheng: done

docker/README.md
longquanzheng: Yes, you are right.

docker/README.md
longquanzheng: Ideally yes, we should be able to do that. But with docker-compse, you have to stop and then start so that it will pick up the new docker image.
I have tried "docker-compose restart", it simply restart the container without reading the changes from compose file.

docker/README.md
longquanzheng: Done

docker/README.md
longquanzheng: Done

PR #2208, opened 2019-07-15
Domain specific rate limiter

shreyassrivatsan:
Just the basic implementation with benchmarks. Will be used in subsequent diffs.


common/quotas/policy.go
```diff
@@ -43,3 +49,57 @@ func (s *simpleRateLimitPolicy) Allow() bool {
...
+
+func (d *DomainRateLimitPolicy) Allow(domain string) bool {
+	// check if we have a per-domain limiter - if not create a default one for
+	// the domain.
+	d.RLock()
+	limiter, ok := d.domainLimiters[domain]
+	d.RUnlock()
+	if !ok {
+		d.Lock()
+		limiter = rate.NewLimiter(rate.Limit(defaultRps), defaultRps)
```
wxing1292: 1. this line can be out of the lock
2. before putting the limiter into the map, do a double check?

PR #2213, opened 2019-07-16
Use dynamic rate limiter for workflow handler

shreyassrivatsan:


common/quotas/ratelimiter.go
```diff
@@ -123,3 +126,21 @@ func (rl *RateLimiter) shouldUpdate(maxDispatchPerSecond *float64) bool {
...
 	}
 }
+
+type dynamicRateLimiter struct {
+	rps dynamicconfig.IntPropertyFn
+	rl  *RateLimiter
+}
+
+// NewDynamicRateLimiter returns a rate limiter which handles dynamic config
+func NewDynamicRateLimiter(rps dynamicconfig.IntPropertyFn) Policy {
```
venkat1109: Given that this class is in common/quotas, how about taking a func() that returns float64 instead of int ? there are cases, for instance in matching service, where the ratelimit is specified by user and could be fractional

venkat1109: Also, its also preferable to refactor NewRateLimiter() above to take a func() instead of taking a pointer

shreyassrivatsan: I did 1. Yes, ideally `NewRateLimiter` takes a `func()` rather than a pointer, but without the task list actually keeping track of the current RPS its going to be difficult to refactor that.. I'll think a bit more if there is a way I can change that - otherwise I am going to leave this as is for a later refactor.

shreyassrivatsan: The way the RPS is determined in the matching scenario, I do not see how we can get away from using a pointer in `NewRateLimiter`

PR #2216, opened 2019-07-16
Add synchronous request reply design doc

andrewjdawson2016:


docs/design/2215-synchronous-request-reply.md
```diff
@@ -0,0 +1,102 @@
+# Proposal: Synchronous Request Reply
+
+Authors: Maxim Fateev (mfateev), Andrew Dawson (andrewjdawson2016)
```
dmetzgar: Put the @ symbol in front of the aliases so GitHub links them

docs/design/2215-synchronous-request-reply.md
```diff
@@ -0,0 +1,102 @@
...
+2. Long poll on query result changes
+
+3. Update operation that should complete within a single decision
+
+All three together cover all the listed use cases.
+
+### Guarantee read after write consistency between signal and query
+
+Currently there are situations when a query emitted after a signal sees a stale workflow state without the signal update yet delivered. This forces complications to a workflow logic as signal then query for updated state interaction pattern is not consistent. After this race condition is fixed signal then query can be used for request-reply like scenarios. The drawback of this approach is that it is not most optimal from the performance point of view as it requires at least two decisions and multiple database updates. But together with Long poll on query result changes feature it is a reasonable way to implement update and then wait for a long running update operations.
+
```
dmetzgar: Fix a few typos in the paragraph above to make it easier to read:

> Currently, there is a race condition when a query emitted after a signal sees a stale workflow state without the update from the signal. Accounting for this inconsistency complicates the workflow logic. Fixing this race condition allows the signal-then-query pattern to be used for request-reply scenarios. The drawback of this approach is suboptimal performance since it requires at least two decisions and multiple database updates. But, together with the long polling on query result changes feature, it is a reasonable way to implement update-and-wait for long-running update operations.

PR #2220, opened 2019-07-16
Implement Frontend Service for Batch Job(Pending on meeting for API discussion)

longquanzheng:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1484,7 +1484,7 @@ struct BatchCancelRequest{
 
 struct StartBatchJobRequest{
     10: optional string domain
-    20: optional string jobID
+    20: optional SearchAttributes searchAttributes
```
shreyassrivatsan: why do you need `SearchAttributes` while starting batch job? Also its odd to have it as 2nd parameter. If we are anyway breaking contract move it down..

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1484,7 +1484,7 @@ struct BatchCancelRequest{
...
     30: optional BatchOperationType operationType
     40: optional string sourceQuery
     50: optional string reason
```
shreyassrivatsan: Overall this input request struct looks odd.. 

what is header, requestID, identity and startToCloseTimeoutSeconds used for? 

I think things like sourceQuery, reason, rps should be clubbed into a single struct that applies to batch in general for clarity.. searchAttributes also falls into this bucket..

idl/github.com/uber/cadence/cadence.thrift
```diff
@@ -577,4 +577,61 @@ service WorkflowService {
...
     )
 
+  /**
+  * StartBatchJob starts a new batch job
+  **/
+  shared.StartBatchJobResponse StartBatchJob(1: shared.StartBatchJobRequest startRequest)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.InternalServiceError internalServiceError,
+      3: shared.WorkflowExecutionAlreadyStartedError sessionAlreadyExistError,
```
mfateev: What workflow execution has to do with batch?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+    Signal
+}
+
+struct BatchSignalRequest{
+    10: optional string signalName
+    20: optional binary signalData
+}
+
+struct BatchTerminateRequest{
+    10: optional bool applyOnChildren
```
mfateev: Let's don't put it to not deprecate as soon as parent policy is implemented.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+}
+
+struct BatchCancelRequest{
+    10: optional bool applyOnChildren
```
mfateev: ditto

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+
+struct DescribeBatchJobRequest{
+    10: optional string domain,
+    20: optional string jobID
+}
+
+struct DescribeBatchJobResponse{
+    10: optional i64 totalEstimate
+    20: optional i64 successCount
+    30: optional i64 failureCount
```
mfateev: Why do we even have user facing failures?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+    40: optional WorkflowExecutionCloseStatus status
```
mfateev: What is this? Why batch request has workflowexecutionclosestatus?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+    50: optional i32 currentPage
```
mfateev: This doesn't make sense for the user unless it is possible to kill batch job and restart from that page one day later.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1462,3 +1462,91 @@ struct HistoryBranch{
...
+    //    70: optional i64 rps
+}
+
+struct ListBatchJobsRequest{
+    10: optional string domain
+}
+
+struct BatchJobInfo{
+    10: optional string jobID
+    20: optional WorkflowExecutionCloseStatus status
```
mfateev: ditto

idl/github.com/uber/cadence/cadence.thrift
longquanzheng: Removed.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: After the meeting I think it a little more that we still need it. 
ParentClosePolicy won't help because a lot of exsiting domains will be using abandon as default or setting abandon and change mind later because. 
We thought replacement is after we support RootWorkflowID as search attributes, then customer can use batch with querying RootWorkflowID. But it is not correct:  a customer can have multiple trees and the RootWorkflowIDs are from another query. In that case this is the only way a customer can kill the trees because our ES query doesn't support subQuery(or JOIN). 
And and it is always easier for customers to understand with this parameter rather than using RootWorkflowID.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: With this service, customers can start many batch jobs and the only way to manage(find) them is to put some tags. 
@mfateev  I can remove it if you both think it is unnecessary for now. We can definitely do it later when needed.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: BatchJob will be a first class member as Activity/Timer/ChildWorkflow/etc in our clientSDK. 
Similar to start a workflow:
Header is for jeager propagation. 
RequestedID is for dedup. 
identity is also used for our client SDK.
startToCloseTimeoutSeconds is timeout for the batchjob. 

How do you call the struct? I don't see this is a convention in our IDL yet. Btw there will be another layer above in our client SDK.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: There can be some unretryable failures anyway. For example, when resetting to a bad binary, the badbinary can be expired due to the previous run was deleted(without achival). Or it max out the retry limit.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: I reused this enum. For a better name we can create the same one like BatchJobCloseStatus. But essentially they are the same because batch job is essentially a workflow, and we don't need to map from BatchJobCloseStatus to WorkflowExecutionCloseStatus if we reusing it.

idl/github.com/uber/cadence/shared.thrift
longquanzheng: That's true. I was thinking to provide visibility of progress but looks like we already have it with SuccessCount/FailureCount/TotalEstimate

idl/github.com/uber/cadence/shared.thrift
shreyassrivatsan: I think a separate enum is preferable for clarity.

idl/github.com/uber/cadence/shared.thrift
shreyassrivatsan: I prefer to put all the general BatchOperation options into a separate struct or at least improve the ordering here. The general options and then the batch specific options..

PR #2221, opened 2019-07-17
Refactor workflow execution context & mutable state

wxing1292:
* Remove redundant append first event batch APIs
* Remove mutableStateSessionUpdates in favor of WorkflowSnapshot & WorkflowMutation
* Move force fail decision for failover / buffered events size limit into CloseTransactionAsMutation & CloseTransactionAsSnapshot
* Mutable state will now keep track of all transfer / timer tasks by AddTransferTasks & AddTimerTasks
* Remove the requirement for transactionID from shard when create / update workflow
* Remove special handling of LimitExceededError since buffered events size limit is handled before close transaction


service/history/mutableStateBuilder.go
```diff
@@ -3684,31 +3671,39 @@ func (e *mutableStateBuilder) prepareEventsAndReplicationTasks() ([]*persistence
...
-		convertSyncActivityInfos(
-			e.pendingActivityInfoIDs,
-			e.syncActivityTasks,
-		)...)
+		e.syncActivityToReplicationTask(closeAsActive)...,
+	)
+
+	if !closeAsActive && len(e.insertReplicationTasks) > 0 {
+		return nil, &workflow.InternalServiceError{
+			Message: "should no generate replication task when close transaction as passive",
```
venkat1109: should not*

service/history/mutableStateBuilder.go
```diff
@@ -3777,6 +3792,122 @@ func (e *mutableStateBuilder) validateNoEventsAfterWorkflowFinish(
 	}
 }
 
+func (e *mutableStateBuilder) closeTransactionCheckActiveness(
```
venkat1109: consider renaming to closeTransactionIfActive()

service/history/workflowExecutionContext.go
```diff
@@ -151,11 +134,10 @@ type (
 		metricsClient     metrics.Client
 		timeSource        clock.TimeSource
 
-		locker                locks.Mutex
-		msBuilder             mutableState
-		stats                 *persistence.ExecutionStats
-		updateCondition       int64
-		createReplicationTask bool
+		locker          locks.Mutex
```
venkat1109: nit: mutex is a well understood name for this

service/history/conflictResolver.go
```diff
@@ -119,6 +119,9 @@ func (r *conflictResolverImpl) reset(
 				r.shard.GetEventsCache(),
 				r.logger,
 				firstEvent.GetVersion(),
+				// if can see replication task, meaning that domain is
+				// global domain with > 1 target clusters
+				true,
```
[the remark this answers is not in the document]
wxing1292: this is on the replication receiver side, meaning that if logic is being executed, it is guaranteed that the mutable state to be generated can do replication

Issue #2187, opened 2019-07-09
Archival: Create Integration test for Archival Interface


yycptt:
As we move to the new Archiver interface, we also need a set of integration tests for that interface.

yycptt:
This should be done as part of #2093

Issue #2094, opened 2019-06-24
Archival: Archival Small History without Sending Signal to Archival Workflow


yycptt:
Task 7 of https://github.com/uber/cadence/issues/2082
 
Create dynamic config for the threshold of history size below which history should be archived by the timer queue processor without sending signal to archival workflow
 
Update timer queue processor to use archiver

yycptt:
Should be done as part of https://github.com/uber/cadence/issues/2093

Issue #2093, opened 2019-06-24
Archival: Delete Blobstore Interface


yycptt:
Task 6 of https://github.com/uber/cadence/issues/2082
 
Delete blobstore interface and all its implementations
Switch to new Archiver interface, includes frontend, archiver worker.

PR #2222, opened 2019-07-17
Archival: Switch to Archiver Interface

yycptt:
#2093 #2094 #2187


common/metrics/defs.go
```diff
@@ -1692,68 +1628,55 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
...
+		ESProcessorRequests:                      {metricName: "es_processor_requests"},
+		ESProcessorRetries:                       {metricName: "es_processor_retries"},
+		ESProcessorFailures:                      {metricName: "es_processor_errors"},
+		ESProcessorCorruptedData:                 {metricName: "es_processor_corrupted_data"},
+		ESProcessorProcessMsgLatency:             {metricName: "es_processor_process_msg_latency", metricType: Timer},
+		IndexProcessorCorruptedData:              {metricName: "index_processor_corrupted_data"},
+		IndexProcessorProcessMsgLatency:          {metricName: "index_processor_process_msg_latency", metricType: Timer},
+		ArchiverNonRetryableErrorCount:           {metricName: "archiver_non_retryable_error"},
+		ArchiverHistoryMutatedCount:              {metricName: "archiver_history_mutated"},
+		ArchiverTotalUploadSize:                  {metricName: "archiver_total_upload_size", metricType: Timer},
```
yycptt: I keep the metrics for mutated history and total upload size which I think should be useful for other implementations.

common/metrics/defs.go
```diff
@@ -1412,6 +1357,8 @@ const (
 	WorkflowCronBackoffTimerCount
 	WorkflowCleanupDeleteCount
 	WorkflowCleanupArchiveCount
+	WorkflowCleanupSendArchiveSignalFailedCount
```
yycptt: There's an old metrics called ArchiverClientSendSignalFailureCount which is not used anywhere. I moved it here.

common/metrics/defs.go
```diff
@@ -1412,6 +1357,8 @@ const (
...
+	WorkflowCleanupArchiveFailedCount
```
yycptt: This counts the failed archive operation in timer queue processor.

service/history/historyEngine.go
```diff
@@ -178,9 +179,18 @@ func NewEngineWithShardContext(
...
 	}
+	historyArchiverBootstrapContainer := &archiver.HistoryBootstrapContainer{
+		HistoryManager:   historyManager,
+		HistoryV2Manager: historyV2Manager,
+		Logger:           shard.GetLogger(),
+		MetricsClient:    shard.GetMetricsClient(),
+		ClusterMetadata:  shard.GetClusterMetadata(),
+		DomainCache:      shard.GetDomainCache(),
+	}
+	historyEngImpl.archiverProvider.RegisterBootstrapContainer(common.HistoryServiceName, historyArchiverBootstrapContainer, &archiver.VisibilityBootstrapContainer{})
```
yycptt: I am not 100% sure whether the registration should live here or when a new history handler is created.

service/frontend/workflowHandler.go
```diff
@@ -3216,23 +3209,39 @@ func (wh *WorkflowHandler) getArchivedHistory(
...
 	if err != nil {
 		return nil, wh.error(err, scope)
 	}
+
+	resp, err := historyArchiver.Get(ctx, archivalURI, &archiver.GetHistoryRequest{
+		DomainID:      domainID,
+		WorkflowID:    request.GetExecution().GetWorkflowId(),
+		RunID:         request.GetExecution().GetRunId(),
+		NextPageToken: request.GetNextPageToken(),
+		PageSize:      int(request.GetMaximumPageSize()),
```
andrewjdawson2016: We are treating pageSize as number of events is this how persistence treats it? I thought in persistence it represents number of batches?

service/history/historyEngine.go
andrewjdawson2016: It should live at the very top level - service/history/service.go::Start()

service/history/service.go
```diff
@@ -196,6 +198,8 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int, enableVisibilit
 		TimerProcessorMaxPollInterval:                         dc.GetDurationProperty(dynamicconfig.TimerProcessorMaxPollInterval, 5*time.Minute),
 		TimerProcessorMaxPollIntervalJitterCoefficient:        dc.GetFloat64Property(dynamicconfig.TimerProcessorMaxPollIntervalJitterCoefficient, 0.15),
 		TimerProcessorMaxTimeShift:                            dc.GetDurationProperty(dynamicconfig.TimerProcessorMaxTimeShift, 1*time.Second),
+		TimerProcessorHistoryArchivalSizeLimit:                dc.GetIntProperty(dynamicconfig.TimerProcessorHistoryArchivalSizeLimit, 2*1024*1024),
+		TimerProcessorHistoryArchivalTimeLimit:                dc.GetDurationProperty(dynamicconfig.TimerProcessorHistoryArchivalTimeLimit, 60*time.Second),
```
andrewjdawson2016: Are there any timelimits on how long a timer task can run for - if so this number needs to be smaller than that bound.

service/history/timerQueueProcessorBase.go
```diff
@@ -632,50 +634,94 @@ func (t *timerQueueProcessorBase) deleteWorkflow(task *persistence.TimerTaskInfo
...
-	if err := t.historyService.archivalClient.Archive(req); err != nil {
-		t.logger.Error("failed to initiate archival", tag.Error(err),
-			tag.WorkflowID(task.WorkflowID),
-			tag.WorkflowRunID(task.RunID),
-			tag.WorkflowDomainID(task.DomainID),
-			tag.ShardID(t.shard.GetShardID()),
-			tag.TaskID(task.GetTaskID()),
-			tag.FailoverVersion(task.GetVersion()),
-			tag.TaskType(task.GetTaskType()))
+	executionStats, err := workflowContext.loadExecutionStats()
```
andrewjdawson2016: Can you check everything in this file with Wenquan?

service/history/timerQueueProcessorBase.go
```diff
@@ -632,50 +634,94 @@ func (t *timerQueueProcessorBase) deleteWorkflow(task *persistence.TimerTaskInfo
...
+	if err != nil {
 		return err
 	}
+	archiveInline := executionStats.HistorySize < int64(t.config.TimerProcessorHistoryArchivalSizeLimit())
+	taggedLogger := t.logger.WithTags(
+		tag.Error(err),
```
andrewjdawson2016: error is nil here.

service/history/timerQueueProcessorBase.go
```diff
@@ -632,50 +634,94 @@ func (t *timerQueueProcessorBase) deleteWorkflow(task *persistence.TimerTaskInfo
...
+		tag.WorkflowID(task.WorkflowID),
+		tag.WorkflowRunID(task.RunID),
+		tag.WorkflowDomainID(task.DomainID),
+		tag.ShardID(t.shard.GetShardID()),
+		tag.TaskID(task.GetTaskID()),
+		tag.FailoverVersion(task.GetVersion()),
+		tag.TaskType(task.GetTaskType()),
+	)
+	if !archiveInline {
```
andrewjdawson2016: Can we change this to be `if archiveInline` I find it easier to read positive statements than negative statements.

service/history/timerQueueProcessorBase.go
```diff
@@ -632,50 +634,94 @@ func (t *timerQueueProcessorBase) deleteWorkflow(task *persistence.TimerTaskInfo
...
+			RunID:                task.RunID,
+			EventStoreVersion:    msBuilder.GetEventStoreVersion(),
+			BranchToken:          msBuilder.GetCurrentBranch(),
+			NextEventID:          msBuilder.GetNextEventID(),
+			CloseFailoverVersion: msBuilder.GetLastWriteVersion(),
+			URI:                  domainCacheEntry.GetConfig().HistoryArchivalURI,
+		}
+
+		// send signal before deleting mutable state to make sure archival is idempotent
+		if err := t.historyService.archivalClient.Archive(req); err != nil {
```
andrewjdawson2016: What do you think about moving both cases into the archivalClient? That way when you implement DB scanner you do not have to duplicate this code?

service/history/timerQueueProcessorBase.go
```diff
@@ -632,50 +634,94 @@ func (t *timerQueueProcessorBase) deleteWorkflow(task *persistence.TimerTaskInfo
...
+		}
+	}
+
 	if err := t.deleteCurrentWorkflowExecution(task); err != nil {
 		return err
 	}
 	if err := t.deleteWorkflowExecution(task); err != nil {
 		return err
 	}
+	if archiveInline {
```
andrewjdawson2016: In order to handle this in the model where client does both cases just return a result from client which is a struct including a field describing how archival was done.

service/worker/archiver/activities.go
```diff
@@ -23,80 +23,74 @@ package archiver
...
-	errDeleteBlob     = "could not delete existing blob"
 
 	errDeleteHistoryV1 = "failed to delete history from events_v1"
 	errDeleteHistoryV2 = "failed to delete history from events_v2"
 
-	errHistoryMutated = "history was mutated during uploading"
-
-	errActivityPanic = "cadenceInternal:Panic"
+	errActivityPanic       = "cadenceInternal:Panic"
+	errTimeoutStartToClose = "cadenceInternal:Timeout START_TO_CLOSE"
```
andrewjdawson2016: I don't think you need to include timeout errors as non-retryable errors - those are seen as the workflow getting an error rather than the activity returning an error.

service/worker/archiver/activities.go
```diff
@@ -23,80 +23,74 @@ package archiver
...
-// method will retry all retryable operations until context expires.
-// archival will be skipped and no error will be returned if cluster or domain is not figured for archival.
-// an error will be returned if context timeout, request is invalid, or failed to get domain cache entry.
-// all other errors (and the error details) will be included in the uploadResult.
-// the result will also contain a list of blob keys to delete when there's an error.
-func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (result *uploadResult, err error) {
-	return NewHistoryBlobUploader().UploadHistory(ctx, &request)
+// uploadHistoryActivity is used to archive a workflow execution history.
+// method will retry all errors except timeout errors and archiver.ErrArchiveNonRetriable.
+// it's the history archiver's responsibility to retry individual operations inside the Archive() method.
```
andrewjdawson2016: Lets leave line 58 out of the comment. Its totally up to the implementor if they want to retry or not - its not their responsibility to retry.

service/worker/archiver/activities.go
```diff
@@ -23,80 +23,74 @@ package archiver
...
+	if err != nil {
+		logger.Error(carchiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason("failed to extract archival scheme"), tag.ArchivalURI(request.URI))
+		return carchiver.ErrArchiveNonRetriable
+	}
+	historyArchiver, err := container.ArchiverProvider.GetHistoryArchiver(scheme, common.WorkerServiceName)
+	if err != nil {
+		logger.Error(carchiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason("failed to get history archiver"), tag.Error(err))
+		return carchiver.ErrArchiveNonRetriable
+	}
+	return historyArchiver.Archive(ctx, request.URI, &carchiver.ArchiveHistoryRequest{
```
andrewjdawson2016: This looks great.

service/worker/archiver/archiver.go
```diff
@@ -115,57 +114,15 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
 	uploadSW := metricsClient.StartTimer(metrics.ArchiverScope, metrics.ArchiverUploadWithRetriesLatency)
-	var result *uploadResult
-	err := workflow.ExecuteActivity(actCtx, uploadHistoryActivityFnName, request).Get(actCtx, &result)
-	if err == nil && result != nil {
-		err = errors.New(result.ErrorWithDetails)
-	}
+	err := workflow.ExecuteActivity(actCtx, uploadHistoryActivityFnName, request).Get(actCtx, nil)
 	if err != nil {
-		logger.Error("failed to upload history, will delete all uploaded blobs and moving on to deleting history without archiving", tag.Error(err))
+		logger.Error("failed to upload history, will move on to deleting history without archiving", tag.Error(err))
```
andrewjdawson2016: "failed to archive history" its possible archiving history does not mean uploading.

service/worker/archiver/client_worker.go
```diff
@@ -120,6 +108,16 @@ func NewClientWorker(container *BootstrapContainer) ClientWorker {
...
 	}
+	historyArchiverBootstrapContainer := &carchiver.HistoryBootstrapContainer{
+		HistoryManager:   container.HistoryManager,
+		HistoryV2Manager: container.HistoryV2Manager,
+		Logger:           container.Logger,
+		MetricsClient:    container.MetricsClient,
+		ClusterMetadata:  container.ClusterMetadata,
+		DomainCache:      container.DomainCache,
+	}
+	container.ArchiverProvider.RegisterBootstrapContainer(common.WorkerServiceName, historyArchiverBootstrapContainer, &carchiver.VisibilityBootstrapContainer{})
```
andrewjdawson2016: Registration should be done at top level e.g. service.go

service/frontend/workflowHandler.go
yycptt: I think it's also number of events in persistence.

service/history/service.go
yycptt: I don't see any timelimits, but a timer task should be completed very fast. I will change this to 3 seconds.

service/history/timerQueueProcessorBase.go
yycptt: I think pass `archiveInline` to the client.Archive() method makes more sense. Basically, you give the archive client a request and how tell it how archive should be done. Making the client figure out how archive should be done on its own seems the wrong level of abstraction to me. Also `workflowContext` and the dynamicconfig need to be passed to the `Archive()` method to support which I think is weird.

service/history/timerQueueProcessorBase.go
yycptt: Good point! Updated.

service/history/timerQueueProcessorBase.go
yycptt: Done in archival client.

service/frontend/service.go
```diff
@@ -199,8 +200,19 @@ func (s *Service) Start() {
 		kafkaProducer = &mocks.KafkaProducer{}
 	}
 
-	metricsBlobstore := blobstore.NewMetricClient(params.BlobstoreClient, base.GetMetricsClient())
-	wfHandler := NewWorkflowHandler(base, s.config, metadata, history, historyV2, visibility, kafkaProducer, metricsBlobstore, params.ArchiverProvider)
+	domainCache := cache.NewDomainCache(metadata, base.GetClusterMetadata(), base.GetMetricsClient(), base.GetLogger())
```
andrewjdawson2016: I think you want to start the domainCache here. Can you check with Samar on this.

service/worker/archiver/client.go
```diff
@@ -103,11 +162,12 @@ func (c *client) Archive(request *ArchiveRequest) error {
 		DecisionTaskStartToCloseTimeout: workflowTaskStartToCloseTimeout,
 		WorkflowIDReusePolicy:           cclient.WorkflowIDReusePolicyAllowDuplicate,
 	}
-	_, err := c.cadenceClient.SignalWithStartWorkflow(context.Background(), workflowID, signalName, *request, workflowOptions, archivalWorkflowFnName, nil)
+	_, err := c.cadenceClient.SignalWithStartWorkflow(ctx, workflowID, signalName, *request, workflowOptions, archivalWorkflowFnName, nil)
```
andrewjdawson2016: I think passing ctx here will not work because if inline failed ctx will be timed out at this point

service/worker/archiver/client.go
```diff
@@ -75,20 +87,67 @@ func NewClient(
...
+
+	return c.sendArchiveSignal(ctx, request.ArchiveRequest, taggedLogger)
+}
+
+func (c *client) archiveInline(ctx context.Context, request *ClientRequest, taggedLogger log.Logger) (err error) {
+	defer func() {
+		if err != nil {
+			c.metricsClient.IncCounter(metrics.ArchiverClientScope, metrics.ArchiverClientInlineArchiveFailureCount)
+			taggedLogger.Error("failed to perform workflow history archival inline", tag.Error(err))
+			err = c.sendArchiveSignal(ctx, request.ArchiveRequest, taggedLogger)
```
andrewjdawson2016: Can we call this in Archive method. Calling signal in the inline method feels like the wrong level of abstraction to me. These two methods should simply know how to do their respective types of archival - the flow control belongs one level higher.

service/worker/archiver/client.go
yycptt: Changed to context.Background() for now. Question is after we have a reliable estimate of time needed by typical inline archival, do we want to change back to `ctx`. Basically if inline already takes too much time and blocking other tasks, do we want to spend more time to send the signal or just let it picked up by DB scanner.

service/worker/archiver/client.go
yycptt: Makes sense. Updated.

PR #2223, opened 2019-07-17
Ignore decision schedule to start timer in passive processing

wxing1292:
This is because schedule to start timer for decision is only used by sticky decision, and this sticky decision is only used locally, when failover happen, remote active cluster's corresponding decision is non sticky


service/history/timerQueueStandbyProcessor.go
```diff
@@ -330,6 +330,13 @@ func (t *timerQueueStandbyProcessorImpl) processActivityTimeout(timerTask *persi
 
 func (t *timerQueueStandbyProcessorImpl) processDecisionTimeout(timerTask *persistence.TimerTaskInfo, lastAttempt bool) error {
 
+	// decision schedule to start timer task is a special snowflake.
+	// the schedule to start timer is for sticky decision, which is
+	// not applicable for cross DC
```
venkat1109: not applicable on the passive cluster

PR #2224, opened 2019-07-17
Add info struct to domain policy

shreyassrivatsan:


common/quotas/policy.go
```diff
@@ -83,7 +80,8 @@ func (d *DomainRateLimitPolicy) Allow(domain string) bool {
 
 	if !ok {
 		// create a new limiter
-		domainLimiter := rate.NewLimiter(rate.Limit(domainRps), domainRps)
+		initialRps := d.domainRPS()
+		domainLimiter := NewRateLimiter(&initialRps, _defaultRPSTTL, int(d.domainRPS()))
```
wxing1292: should we consider allowing temporary burst?

shreyassrivatsan: Current config allows the burst to take up the full limit, but we do not actually expose a get N tokens API, so the burst really isnt useful at this time. Its more like a placeholder.. But with more complicated policy it will come into play..

common/quotas/ratelimiter.go
```diff
@@ -126,19 +126,28 @@ func (rl *RateLimiter) shouldUpdate(maxDispatchPerSecond *float64) bool {
...
 // NewDynamicRateLimiter returns a rate limiter which handles dynamic config
-func NewDynamicRateLimiter(rps RPSFunc) Policy {
+func NewDynamicRateLimiter(rps RPSFunc) *DynamicRateLimiter {
 	initialRps := rps()
 	rl := NewRateLimiter(&initialRps, _defaultRPSTTL, int(rps()))
-	return &dynamicRateLimiter{rps, rl}
+	return &DynamicRateLimiter{rps, rl}
 }
 
-func (d *dynamicRateLimiter) Allow() bool {
```
wxing1292: `dynamicRateLimiter` is different than `DomainRateLimitPolicy`
`dynamicRateLimiter` should not really require any input information other than perhaps how many token to be consumed.

one more thing, `DomainRateLimitPolicy` is not a policy (which should only contains configurations)
perhaps the `DomainRateLimitPolicy` could be better named something like multi-stage-rate-limiter

common/quotas/policy.go
wxing1292: it does not has to be related to `consume N token` for some heavily lifting APIs.
my point is allow customer to temporary use more tokens, that is what token bucket for.

common/quotas/ratelimiter.go
shreyassrivatsan: I like that.. that's a better name..

common/quotas/policy.go
shreyassrivatsan: Ah yes, that was an oversight. Updated..

PR #2226, opened 2019-07-18
Use golang rate limiter everywhere

shreyassrivatsan:


common/quotas/interfaces.go
```diff
@@ -28,6 +30,18 @@ type Info struct {
 	Domain string
 }
 
+// Limiter corresponds to basic rate limiting functionality.
+type Limiter interface {
+	// Allow attempts to allow a request to go through. The method returns
+	// immediately with a true or false indicating if the request can make
+	// progress
+	Allow() bool
```
wxing1292: maybe it is just me: 
Allow -> Request()
&
Wait -> RequestWithContext() ?

shreyassrivatsan: Yeah, its not the best name, but its consistent with the golang rate limiter, so prefer not to deviate.

PR #2228, opened 2019-07-18
Make search attributes works with reset in xdc

vancexu:


vancexu:
Abort this in favor of Refactor reset mutable state path #2230, changes include there

PR #2229, opened 2019-07-18
matching: support for task list partitions

### LINKED ISSUE #2098: matching: high throughput task list

venkat1109:
Fixes #2098


service/matching/matcher.go
```diff
@@ -162,19 +274,79 @@ func (tm *TaskMatcher) PollForQuery(ctx context.Context) (*internalTask, error)
...
 func (tm *TaskMatcher) UpdateRatelimit(rps *float64) {
-	tm.limiter.UpdateMaxDispatch(rps)
+	if rps == nil {
+		return
+	}
+	rate := *rps
+	nPartitions := tm.numPartitions()
+	if rate > float64(nPartitions) {
+		// divide the rate equally across all partitions
+		rate = rate / float64(tm.numPartitions())
```
shreyassrivatsan: probably `numpartitions-1` if it is greater than 1.. as the parent partition should not do any work in a fully loaded system when the limit will actually apply..

service/matching/matcher.go
```diff
@@ -105,7 +103,20 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
 			return true, err
 		}
 		return false, nil
-	default: // no poller waiting for tasks
+	default:
+		// no poller waiting for tasks, try forwarding this task to the
+		// root partition if available
+		select {
+		case token := <-tm.fwdrAddReqTokenC():
```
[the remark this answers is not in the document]
venkat1109: first version of my code had the maxOutstanding logic within the forwarder. But it doesn't work because the channel needs to be exposed so that the caller can block on the channel along with other things (i.e. wait for a local poller or a forwarding token to be available simultaneously in a select{} block). The job of matcher is to still match producer with consumer. It now also enforces maxOutstanding / ratelimits

service/matching/matcher.go
venkat1109: > as the parent partition should not do any work in a fully loaded system when the limit will actually apply
This isn't true unless I am misunderstanding your statement. Parent partition is just another partition and when the system is fully loaded, it gets roughly 1/N of the traffic

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (lb *defaultLoadBalancer) pickPartition(
+	domainID string,
+	taskList shared.TaskList,
+	taskListType int,
+	forwardedFrom string,
+	nPartitions dynamicconfig.IntPropertyFnWithTaskListInfoFilters,
+) string {
+
+	if forwardedFrom != "" || taskList.GetKind() == shared.TaskListKindSticky {
```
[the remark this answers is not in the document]
venkat1109: Nope, multiple levels of forwarding is ok. May be a comment might help here.  When forwardedFrom is set, LB doesn't pick a partition because partition is already picked by the caller

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+
+// NewLoadBalancer returns an instance of matching load balancer that
+// can help distribute api calls across task list partitions
+func NewLoadBalancer(
+	domainIDToName func(string) (string, error),
+	dc *dynamicconfig.Collection,
+) LoadBalancer {
+	return &defaultLoadBalancer{
+		domainIDToName:   domainIDToName,
+		nReadPartitions:  dc.GetIntPropertyFilteredByTaskListInfo(dynamicconfig.MatchingNumTasklistReadPartitions, 1),
```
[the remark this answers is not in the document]
venkat1109: let's talk offline. I believe the code has all the logic to deal with changing number of partitions. It's possible I missed something.

service/matching/matcher.go
```diff
@@ -115,25 +126,111 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
...
+			token.release()
+			if err != nil {
+				// forwarder returns error only when the call is rate limited. To
+				// avoid a busy loop on such rate limiting events, we only attempt to make
+				// the next forwarded call after this childCtx expires. Till then, we block
+				// hoping for a local poller match
+				select {
+				case tm.taskC <- task:
+					return nil
+				case <-childCtx.Done():
```
shreyassrivatsan: If I am not mistaken the cancel you call above will cause the child context to be done here, so this wont block for local poller, it will exit the select immediately.

service/matching/matcher.go
```diff
@@ -115,25 +126,111 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
...
+		}
+		tm.scope().IncCounter(metrics.PollSuccessCounter)
+		return task, nil
+	case task := <-tm.queryTaskC:
+		tm.scope().IncCounter(metrics.PollSuccessWithSyncCounter)
+		tm.scope().IncCounter(metrics.PollSuccessCounter)
+		return task, nil
+	default:
+	}
+
```
shreyassrivatsan: would be good to have some comments here.

client/clientfactory.go
```diff
@@ -141,7 +155,8 @@ func (cf *rpcClientFactory) NewMatchingClientWithTimeout(
 		return matchingserviceclient.New(dispatcher.ClientConfig(common.MatchingServiceName)), nil
 	}
 
-	client := matching.NewClient(timeout, longPollTimeout, common.NewClientCache(keyResolver, clientProvider))
+	lb := matching.NewLoadBalancer(domainIDToName, cf.dynConfig)
```
wxing1292: lb -> loadBalancer

service/matching/matcher.go
```diff
@@ -115,25 +126,111 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
...
+			return ctx.Err()
+		}
 	}
 }
 
 // Poll blocks until a task is found or context deadline is exceeded
 // On success, the returned task could be a query task or a regular task
 // Returns ErrNoTasks when context deadline is exceeded
 func (tm *TaskMatcher) Poll(ctx context.Context) (*internalTask, error) {
+	// try local match first
```
shreyassrivatsan: can we share code between Poll and PollForQuery

service/matching/matcher.go
shreyassrivatsan: Because we try local sync match first on a fully loaded system I thought tasks would rarely make it to the parent. but i see the parent is treated as an equivalent partition in the `pick` logic (i guess parent could starve otherwise), so yes `1/N` should be fine

client/matching/client.go
```diff
@@ -44,86 +45,124 @@ type clientImpl struct {
 	timeout         time.Duration
 	longPollTimeout time.Duration
 	clients         common.ClientCache
+	lb              LoadBalancer
```
wxing1292: lb -> loadBalancer

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
wxing1292: loadBalancer.go plz

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+			taskList shared.TaskList,
+			taskListType int,
+			forwardedFrom string,
+		) string
+	}
+
+	defaultLoadBalancer struct {
+		nReadPartitions  dynamicconfig.IntPropertyFnWithTaskListInfoFilters
+		nWritePartitions dynamicconfig.IntPropertyFnWithTaskListInfoFilters
+		domainIDToName   func(string) (string, error)
```
wxing1292: domainIDToName DomainIDToNameFunc

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+const (
+	taskListPartitionPrefix = "/__cadence_sys/"
+)
+
+// NewLoadBalancer returns an instance of matching load balancer that
+// can help distribute api calls across task list partitions
+func NewLoadBalancer(
+	domainIDToName func(string) (string, error),
```
wxing1292: domainIDToName DomainIDToNameFunc

service/matching/matcher.go
```diff
@@ -115,25 +126,112 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
...
+	}
+
+	fwdrTokenC := tm.fwdrAddReqTokenC()
+
+	for {
+		select {
+		case tm.queryTaskC <- task:
+			<-task.responseC
+			return nil, nil
+		case token := <-fwdrTokenC:
```
wxing1292: should this be in the default?

service/matching/matcher.go
```diff
@@ -115,25 +126,112 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
...
 		return ctx.Err()
+	default:
+	}
+
+forLoop:
+	for {
+		select {
+		case tm.taskC <- task:
+			return nil
+		case token := <-tm.fwdrAddReqTokenC():
```
wxing1292: should this be in the default?

service/matching/matcher.go
```diff
@@ -148,12 +246,27 @@ func (tm *TaskMatcher) Poll(ctx context.Context) (*internalTask, error) {
 	case <-ctx.Done():
 		tm.scope().IncCounter(metrics.PollTimeoutCounter)
 		return nil, ErrNoTasks
+	case token := <-tm.fwdrPollReqTokenC():
+		if task, err := tm.fwdr.ForwardPoll(ctx); err == nil {
```
wxing1292: before actually polling from root / remote, maybe another attempt of polling from `tm.taskC` and `tm.queryTaskC`?

service/matching/matcher.go
```diff
@@ -162,19 +275,79 @@ func (tm *TaskMatcher) PollForQuery(ctx context.Context) (*internalTask, error)
...
```
wxing1292: same above

service/matching/matcher.go
wxing1292: basically, i think polling from local should have higher priority, event if task manager has token to poll from remote

client/matching/lb.go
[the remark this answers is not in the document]
venkat1109: yup. that's the idea.

service/matching/matcher.go
venkat1109: good match. Missed this as I moved around code.

service/matching/matcher.go
venkat1109: I did consider this and then decided to drop this idea for the first round and here is the reasoning:

Before entering the select{} block which has both local task channels and forwarderTokenC, we always first exclusively attempt to match only with local task channels i.e. we always prioritize local pollers in the first select{} block before entering the second. In a busy task list, the first select{} should almost always match against a local poller. On the other hand, when its not matched, the task list probably isn't all that busy and its ok to forward. Having said that, my plan is to observe the forwardedFrom / matching metrics from a real stress test and use that to decide if adding this additional nested select{} is worth it. Till then, I prefer to start simple.

service/matching/matcher.go
venkat1109: yup. done.

service/matching/forwarder.go
```diff
@@ -0,0 +1,305 @@
...
+) *Forwarder {
+	rps := float64(cfg.ForwarderMaxRatePerSecond())
+	fwdr := &Forwarder{
+		cfg:                   cfg,
+		client:                client,
+		taskListID:            taskListID,
+		taskListKind:          kind,
+		outstandingTasksLimit: int32(cfg.ForwarderMaxOutstandingTasks()),
+		outstandingPollsLimit: int32(cfg.ForwarderMaxOutstandingPolls()),
+		limiter:               quotas.NewRateLimiter(&rps, _defaultTaskDispatchRPSTTL, 1),
```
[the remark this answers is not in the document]
venkat1109: done

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+	// LoadBalancer is the interface for implementers of
+	// component that distributes add/poll api calls across
+	// available task list partitions when possible
+	LoadBalancer interface {
+		// PickWritePartition returns the task list partition for adding
+		// an activity or decision task. The input is the name of the
+		// original task list (with no partition info). When forwardedFrom
+		// is non-empty, this call is forwardedFrom from a child partition
+		// to a parent partition in which case, no load balancing should be
+		// performed
```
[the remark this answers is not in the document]
venkat1109: this should be a contract of the interface itself

client/matching/lb.go
venkat1109: that would introduce a circular dependency. On the other hand, I can move the type definition of DomainIDToNameFunc() within this package. But that would be weird for the caller- who interacts only with client package

client/matching/lb.go
venkat1109: see comment above

client/matching/lb.go
```diff
@@ -0,0 +1,136 @@
...
+	forwardedFrom string,
+	nPartitions dynamicconfig.IntPropertyFnWithTaskListInfoFilters,
+) string {
+
+	if forwardedFrom != "" || taskList.GetKind() == shared.TaskListKindSticky {
+		return taskList.GetName()
+	}
+
+	if strings.HasPrefix(taskList.GetName(), taskListPartitionPrefix) {
+		// this should never happen when forwardedFrom is false
```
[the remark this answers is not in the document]
venkat1109: done

client/clientfactory.go
venkat1109: done

PR #2233, opened 2019-07-19
Introduce pub-sub on mutable state

andrewjdawson2016:
- Introduces new component of workflowExecutionContext which is used to watch for mutableState changes.
- Follow up diff will remove historyEventNotifier in favor of this new pub-sub.

### DIFF
--- a/service/history/mutableStatePubSub.go
+++ b/service/history/mutableStatePubSub.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package history
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/pborman/uuid"
+	"github.com/uber/cadence/common/persistence"
+)
+
+type (
+	// MutableStatePubSub is used to get updates on mutable state changes.
+	MutableStatePubSub interface {
+		Publish(*persistence.WorkflowMutableState)
+		Subscribe() (string, <-chan struct{})
+		Unsubscribe(string)
+		LatestMutableState() *persistence.WorkflowMutableState
+	}
+
+	mutableStatePubSub struct {
+		// An atomic is used to keep track of mutable state rather than pushing updates on a channel.
+		// If a channel (buffered or unbuffered) is used, then at some point either blocking will occur or a mutable state update will be dropped.
+		// Since clients only care about getting the latest mutable state, simplify notifying clients
+		// of updates and providing the ability to get the latest satisfies all current use cases.
+		latestMS atomic.Value
+		lockableSubscribers
+	}
+
+	lockableSubscribers struct {
+		sync.RWMutex
+		subscribers map[string]chan struct{}
+	}
+)
+
+func (ls *lockableSubscribers) add() (string, chan struct{}) {
+	ls.Lock()
+	defer ls.Unlock()
+	id := uuid.New()
+	ch := make(chan struct{}, 1)
+	ls.subscribers[id] = ch
+	return id, ch
+}
+
+func (ls *lockableSubscribers) remove(id string) {
+	ls.Lock()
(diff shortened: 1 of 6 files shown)


service/history/mutableStatePubSub.go
```diff
@@ -0,0 +1,118 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/pborman/uuid"
+	"github.com/uber/cadence/common/persistence"
+)
+
+type (
+	// MutableStatePubSub is used to get updates on mutable state changes.
+	MutableStatePubSub interface {
```
wxing1292: suggestion: workflowWatcher?

service/history/workflowExecutionContext.go
```diff
@@ -118,6 +117,7 @@ type (
 			baseRunID string,
 			baseRunNextEventID int64,
 		) (retError error)
+		mutableStatePubSub() MutableStatePubSub
```
wxing1292: Get*

service/history/workflowExecutionContext.go
```diff
@@ -489,7 +491,9 @@ func (c *workflowExecutionContextImpl) updateWorkflowExecutionWithNew(
 	// TODO remove updateCondition in favor of condition in mutable state
 	c.updateCondition = currentWorkflow.ExecutionInfo.NextEventID
 
+	c.pubSub.Publish(c.msBuilder.CopyToPersistence())
```
wxing1292: plz make sure this is a deep copy

wxing1292:
one more thing, you may also consider adding the notification functionality to workflow reset / conflict resolution.

service/history/mutableStatePubSub.go
shreyassrivatsan: +1 workflowWatcher seems a better name.

service/history/mutableStatePubSub.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, ch := range ls.subscribers {
+		select {
+		case ch <- struct{}{}:
+		default:
+		}
+	}
+}
+
+// NewMutableStatePubSub returns new MutableStatePubSub.
+// This should only be called from workflowExecutionContext.go.
```
shreyassrivatsan: Why is this the case? Can you say why?

service/history/mutableStatePubSub.go
```diff
@@ -0,0 +1,118 @@
...
+		sync.RWMutex
+		subscribers map[string]chan struct{}
+	}
+)
+
+func (ls *lockableSubscribers) add() (string, chan struct{}) {
+	ls.Lock()
+	defer ls.Unlock()
+	id := uuid.New()
+	ch := make(chan struct{}, 1)
```
venkat1109: nit: move the allocations outside of the lock

service/history/mutableStatePubSub.go
```diff
@@ -0,0 +1,118 @@
...
+	ls.subscribers[id] = ch
```
venkat1109: does this id need to be unique across workflowExecutionContexts for the same workflow ? If not, consider using a map[int]chan struct{} - it likely will be faster and result in less allocations. The integer keys can be generated using a simple atomic 64-bit increment

service/history/mutableStatePubSub.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/pborman/uuid"
+	"github.com/uber/cadence/common/persistence"
+)
+
+type (
+	// MutableStatePubSub is used to get updates on mutable state changes.
+	MutableStatePubSub interface {
+		Publish(*persistence.WorkflowMutableState)
+		Subscribe() (string, <-chan struct{})
```
venkat1109: [not a blocker] consider returning a struct subscriber { id, ch }..  just makes the code read better on the caller. your unsubscribe can also take a subscriber as input

andrewjdawson2016: Doing this is actually a little tricky because I want to return a read only channel to the caller. But if I return a subscriber struct then the channel has to be read&write so internally I can send on it.

service/history/mutableStatePubSub.go
andrewjdawson2016: I will use int keys instead of uuid. But for now I will reuse the same Mutex primitive used for subscribers rather than introducing a new atomic concurrency primitive for this.

service/history/workflowWatcher.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: 2019

service/history/workflowWatcher.go
```diff
@@ -0,0 +1,122 @@
...
+package history
+
+import (
+	"sync"
+	"sync/atomic"
+)
+
+type (
+	// WatcherUpdate contains the union of all workflow state subscribers care about.
+	WatcherUpdate struct {
```
wxing1292: WatcherSnapshot?
Watchee?

service/history/workflowWatcher.go
```diff
@@ -0,0 +1,122 @@
...
+		sync.Mutex
+		nextSubscriberID int64
+		subscribers      map[int64]chan struct{}
+	}
+)
+
+func (ls *lockableSubscribers) add() (int64, chan struct{}) {
+	ls.Lock()
+	defer ls.Unlock()
+	id := ls.nextSubscriberID
```
wxing1292: you can also just use a uuid

service/history/workflowWatcher.go
```diff
@@ -0,0 +1,122 @@
...
+	return &workflowWatcher{
+		lockableSubscribers: lockableSubscribers{
+			subscribers: make(map[int64]chan struct{}),
+		},
+	}
+}
+
+// Publish is used to indicate an update has been successfully persisted and subscribers should be notified.
+func (w *workflowWatcher) Publish(update *WatcherUpdate) {
+	if update == nil {
```
wxing1292: will this ever be null?

service/history/workflowWatcher.go
andrewjdawson2016: This is generated code so I think the script should be updated rather than the generated code.

service/history/workflowWatcher.go
andrewjdawson2016: Okay I will call it WatcherSnapshot.

service/history/workflowWatcher.go
andrewjdawson2016: Yes the first version used uuid, but Venkat suggested using int to save on allocations. I do not think either is much better than the other.

PR #2240, opened 2019-07-19
Introduce esql, potentially used in report API and Listworkflow API

jysui123:
- esql library for sql translation is added inside /common/elasticsearch/esql
- potentially useful in Visibility Listworkflow API and report API.


Gopkg.lock
```diff
@@ -653,6 +653,104 @@
   revision = "f266f90e9c4d5894364039a324a05d061f2f34e2"
   version = "v3.3.11"
 
+[[projects]]
```
vancexu: please remove this, it shouldn't be there

common/elasticsearch/esql/.gitignore
```diff
@@ -0,0 +1,27 @@
+# Binaries for programs and plugins
```
vancexu: please remove file

common/elasticsearch/esql/.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: go
```
vancexu: please remove this file

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,135 @@
...
+## ES V2.x vs ES V6.5
+|Item|ES V2.x|ES v6.5|
+|:-:|:-:|:-:|
+|missing check|{"missing": {"field": "xxx"}}|{"must_not": {"exist": {"field": "xxx"}}}|
+|group by multiple columns|nested "aggs" field|"composite" flattened grouping|
+
+
+## Attentions
+- We assume all the text are store in type `keyword`, i.e., ES does not break the text into separated words. And we use tag `term` to match the whole text rather than a word in the text.
+- `must_not` in ES does not share the same logic as `NOT` in sql
```
vancexu: Please remove because this detail mislead people to overthink.

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,135 @@
...
+- If you want to apply aggregation on some fields, they should be in type `keyword` in ES (set type of a field by put mapping to your table)
```
vancexu: why only `keyword`? Can it be other basic types like long, boolean?

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,135 @@
...
+- `COUNT(colName)` will include documents w/ null values in that column in ES SQL API, while in esql we exclude null valued documents
+- ES SQL API and esql do not support `SELECT DISTINCT`, but we can achieve the same result by `SELECT * FROM table GROUP BY colName`
```
vancexu: I think instead of "but we can...", it would be better to say "a work around is to use", otherwise it makes me think you added dedup logic in group by.

common/elasticsearch/esql/esql.go
```diff
@@ -0,0 +1,164 @@
...
+	process       Replace // if selected by filterProcess, change the column value
+	cadence       bool
+	pageSize      int
+	bucketNumber  int
+}
+
+// SetDefault ...
+// all members goes to default
+// should not be called if there is potential race condition
+func (e *ESql) SetDefault() {
```
vancexu: Why do we need this? If not necessary, just remove

common/elasticsearch/esql/esql.go
```diff
@@ -0,0 +1,164 @@
...
+		process:       nil,
+		replace:       nil,
+		filterReplace: nil,
+		filterProcess: nil,
+	}
+}
+
+// SetReplace ... set up user specified column name replacement policy
+// should not be called if there is potential race condition
+func (e *ESql) SetReplace(filterArg Filter, replaceArg Replace) {
```
vancexu: Can we add support for multiple filters? Also let's discuss we may need rename to ProcessColumn and ProcessValue

common/elasticsearch/esql/esql.go
```diff
@@ -0,0 +1,164 @@
...
+// usage:
+//  - dsl, sortField, err := e.ConvertPretty(sql, pageParam1, pageParam2, ...)
+//
+// arguments:
+//  - sql: the sql query needs conversion in string format
+//  - pagination: variadic arguments that indicates es search_after for pagination
+//
+// return values:
+//  - dsl: the elasticsearch dsl json style string
+//  - sortField: string array that contains all column names used for sorting. useful for pagination.
```
vancexu: sortFields

common/elasticsearch/esql/esql.go
```diff
@@ -0,0 +1,164 @@
...
+
+// ConvertPretty ...
+// Transform sql to elasticsearch dsl, and prettify the output json
+//
+// usage:
+//  - dsl, sortField, err := e.ConvertPretty(sql, pageParam1, pageParam2, ...)
+//
+// arguments:
+//  - sql: the sql query needs conversion in string format
+//  - pagination: variadic arguments that indicates es search_after for pagination
```
vancexu: Need more description and example, otherwise users will have no idea how to use pagination. 
Also does it mean we only supporting search_after?

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,135 @@
...
+func myProcess(timeStr string) (string, error) {
+    // convert formatted time string to unix nano integer
+    parsedTime, _ := time.Parse(defaultDateTimeFormat, timeStr)
+    return fmt.Sprintf("%v", parsedTime.UnixNano()), nil
+}
+// with the 2 policies , converted dsl is equivalent to
+// "SELECT myColA FROM myTable WHERE myColB < 10 AND dateTime = '1561678568048000000'
+// in which the time is in unix nano format
+e := NewESql()
+e.SetReplace(myFilter1, myReplace)     // set up filtering policy
```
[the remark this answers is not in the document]
jysui123: That's right. Replace is to replace colname and process is to convert the query target value.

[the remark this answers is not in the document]
jysui123: Not now. If user want to do it, multiple policies must be wrapped by a single function by user.

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+## ES V2.x vs ES V6.5
+|Item|ES V2.x|ES v6.5|
+|:-:|:-:|:-:|
+|missing check|{"missing": {"field": "xxx"}}|{"must_not": {"exist": {"field": "xxx"}}}|
+|group by multiple columns|nested "aggs" field|"composite" flattened grouping|
+
+
+## Attentions
+- We assume all the text are store in type `keyword`, i.e., ES does not break the text into separated words. And we use tag `term` to match the whole text rather than a word in the text.
```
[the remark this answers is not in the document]
jysui123: Probably we need discuss on this. Example here:
document: {"field": "abc  bcd"}
query1: "SELECT * FROM dummy WHERE  field = 'abc'"
query2: "SELECT * FROM dummy WHERE  field = 'abc  bcd'"

if stored in keyword type: query1 doesn't match, query2 matches
if stored in text type: query1 matches, query2 doesn't match

if user want to match part of the string, LIKE or REGEXP is better choice.

common/elasticsearch/esql/README.md
jysui123: Yeah, you're right.

common/elasticsearch/esql/esql.go
jysui123: Sure.

common/elasticsearch/esql/esql.go
jysui123: If you mean scroll then yes, now only search after is supported. Let's discuss it later.

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,162 @@
+# ESQL: Translate SQL to Elasticsearch DSL
+Use SQL to query Elasticsearch. ES V6 compatible.
+
+## Supported features
+- [x] =, !=, <, >, <=, >=, <>, ()
+- [x] comparison with arithmetics (e.g. colA + 1 < colB * 2)
+- [x] arithmetic operators: +, -, *, /, %, >>, <<, ()
+- [x] AND, OR, NOT
```
vancexu: nit, the format is italic, good to make them consistent

common/elasticsearch/esql/README.md
```diff
@@ -0,0 +1,162 @@
+# ESQL: Translate SQL to Elasticsearch DSL
+Use SQL to query Elasticsearch. ES V6 compatible.
+
+## Supported features
+- [x] =, !=, <, >, <=, >=, <>, ()
+- [x] comparison with arithmetics (e.g. colA + 1 < colB * 2)
```
vancexu: add note only for number type, and think about supporting date type.

common/elasticsearch/esql/select.go
```diff
@@ -0,0 +1,523 @@
...
+		dslQuerySlice = append(dslQuerySlice, fmt.Sprintf(`"%v": %v`, tag, content))
+	}
+	dsl = "{" + strings.Join(dslQuerySlice, ",") + "}"
+	return dsl, sortField, nil
+}
+
+func (e *ESql) convertWhereExpr(expr sqlparser.Expr, parent sqlparser.Expr) (string, error) {
+	var err error
+	if expr == nil {
+		err = fmt.Errorf("esql: invalid where expression, where expression should not be nil")
```
vancexu: miss return

common/elasticsearch/esql/esql.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Convert ...
+// Transform sql to elasticsearch dsl string
+//
+// usage:
+//  - dsl, sortField, err := e.Convert(sql, pageParam1, pageParam2, ...)
+//
+// arguments:
+//  - sql: the sql query needs conversion in string format
+//  - pagination: variadic arguments that indicates es search_after for
```
vancexu: remove for at the end

common/elasticsearch/esql/README.md
jysui123: It's not italic on my side...

PR #2243, opened 2019-07-21
Use domain rate limiter in workflow handler

shreyassrivatsan:


service/frontend/service.go
```diff
@@ -98,7 +98,7 @@ func NewConfig(dc *dynamicconfig.Collection, numHistoryShards int, enableVisibil
 		ESIndexMaxResultWindow:              dc.GetIntProperty(dynamicconfig.FrontendESIndexMaxResultWindow, 10000),
 		HistoryMaxPageSize:                  dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendHistoryMaxPageSize, common.GetHistoryMaxPageSize),
 		RPS:                                 dc.GetIntProperty(dynamicconfig.FrontendRPS, 1200),
-		DomainRPS:                           dc.GetIntProperty(dynamicconfig.FrontendDomainRPS, 400),
+		DomainRPS:                           dc.GetIntProperty(dynamicconfig.FrontendDomainRPS, 1200),
```
wxing1292: is 1.2K little too high?

shreyassrivatsan: Yes, it is high and will not do anything as global = domain specific limits. Plan is to use dynamic config to bring this down to 400 and ensure it doesnt mess things up. Later we can change the default accordingly.

PR #2245, opened 2019-07-22
Fix type conversion for indexed value type

vancexu:


common/util.go
```diff
@@ -464,3 +464,20 @@ func GetArchivalScheme(URI string) (string, error) {
 	}
 	return URI[:sepIdx], nil
 }
+
+// ConvertIndexedValueTypeToThriftType takes fieldType as interface{} and convert to IndexedValueType.
+// Because different implementation of dynamic config client may lead to different types
+func ConvertIndexedValueTypeToThriftType(fieldType interface{}, logger log.Logger) workflow.IndexedValueType {
+	switch fieldType.(type) {
```
wxing1292: switch type := fieldType.(type) ?

wxing1292: link: https://tour.golang.org/methods/16

Issue #2141, opened 2019-06-28
Invalid Cron delay for @every


mfateev:
I'd expect that it starts every minute just like with */1 schedule. Now it is workflow execution time plus 1 minute. Workflow execution time is less than a minute each time

I think that delay should not be 60 seconds each time but should vary depending on how long previous execution took
```
#schedule: "@every 1m"                           WorkflowExecutionStarted .   WorkflowExecutionContinuedAsNew    Delay
2019-06-27T15:42:13.464-0700    INFO   blah    2019-06-27T15:41:13-07:00    2019-06-27T15:42:37-07:00          60
2019-06-27T15:43:37.634-0700    INFO  blah     2019-06-27T15:42:37-07:00    2019-06-27T15:44:01-07:00          60
2019-06-27T15:45:01.318-0700    INFO  blah     2019-06-27T15:44:01-07:00    2019-06-27T15:45:25-07:00          60
2019-06-27T15:46:25.314-0700    INFO   blah    2019-06-27T15:45:25-07:00    2019-06-27T15:46:49-07:00          60
```

spastukhov:
It looks like GetBackoffForNextSchedule function in https://github.com/uber/cadence/blob/master/common/backoff/cron.go should use previous workflow instance start time as an argument to next rather than current time.

yux0:
With the new fix, one assumption is the workflow start time always correct, which is not true. The start time could be vary based on the last execution close time

PR #2253, opened 2019-07-23
Fix cron calculation bug with @every expression

yux0:
Fix cron calculation bug with @every expression.
#2141

### DIFF
--- a/common/backoff/cron.go
+++ b/common/backoff/cron.go
@@ -56,7 +56,7 @@ func GetBackoffForNextSchedule(cronSchedule string, startTime time.Time, closeTi
 	startUTCTime := startTime.In(time.UTC)
 	closeUTCTime := closeTime.In(time.UTC)
 	nextScheduleTime := schedule.Next(startUTCTime)
-	//Calculate the next schedule start time which is nearest to the close time
+	// Calculate the next schedule start time which is nearest to the close time
 	for nextScheduleTime.Before(closeUTCTime) {
 		nextScheduleTime = schedule.Next(nextScheduleTime)
 	}
--- a/common/backoff/cron.go
+++ b/common/backoff/cron.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
@@ -43,8 +43,8 @@ func ValidateSchedule(cronSchedule string) error {
 }
 
 // GetBackoffForNextSchedule calculates the backoff time for the next run given
-// a cronSchedule and current time
-func GetBackoffForNextSchedule(cronSchedule string, nowTime time.Time) time.Duration {
+// a cronSchedule, workflow start time and workflow close time
+func GetBackoffForNextSchedule(cronSchedule string, startTime time.Time, closeTime time.Time) time.Duration {
 	if len(cronSchedule) == 0 {
 		return NoBackoff
 	}
@@ -53,17 +53,22 @@ func GetBackoffForNextSchedule(cronSchedule string, nowTime time.Time) time.Dura
 	if err != nil {
 		return NoBackoff
 	}
-
-	nowTime = nowTime.In(time.UTC)
-	backoffInterval := schedule.Next(nowTime).Sub(nowTime)
+	startUTCTime := startTime.In(time.UTC)
+	closeUTCTime := closeTime.In(time.UTC)
+	nextScheduleTime := schedule.Next(startUTCTime)
+	//Calculate the next schedule start time which is nearest to the close time
+	for nextScheduleTime.Before(closeUTCTime) {
+		nextScheduleTime = schedule.Next(nextScheduleTime)
+	}
+	backoffInterval := nextScheduleTime.Sub(closeUTCTime)
 	roundedInterval := time.Second * time.Duration(math.Ceil(backoffInterval.Seconds()))
 	return roundedInterval
 }
 
 // GetBackoffForNextScheduleInSeconds calculates the backoff time in seconds for the
 // next run given a cronSchedule and current time
-func GetBackoffForNextScheduleInSeconds(cronSchedule string, nowTime time.Time) int32 {
-	backoffDuration := GetBackoffForNextSchedule(cronSchedule, nowTime)
+func GetBackoffForNextScheduleInSeconds(cronSchedule string, startTime time.Time, closeTime time.Time) int32 {
+	backoffDuration := GetBackoffForNextSchedule(cronSchedule, startTime, closeTime)
 	if backoffDuration == NoBackoff {
 		return 0
 	}
--- a/common/backoff/cron_test.go
+++ b/common/backoff/cron_test.go
@@ -57,7 +57,7 @@ func TestCron(t *testing.T) {
 				end, _ = time.Parse(time.RFC3339, tt.endTime)
 			}
 			backoff := GetBackoffForNextSchedule(tt.cron, start, end)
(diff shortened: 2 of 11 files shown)


host/integration_test.go
```diff
@@ -613,7 +613,7 @@ func (s *integrationSuite) TestWorkflowRetry() {
 		RetryPolicy: &workflow.RetryPolicy{
 			InitialIntervalInSeconds:    common.Int32Ptr(int32(initialIntervalInSeconds)),
 			MaximumAttempts:             common.Int32Ptr(int32(maximumAttempts)),
-			MaximumIntervalInSeconds:    common.Int32Ptr(1),
+			MaximumIntervalInSeconds:    common.Int32Ptr(10),
```
yycptt: Let's change this back to 1 to make sure the maximum interval is enforced.

service/history/historyEngine.go
```diff
@@ -748,10 +746,10 @@ func (e *historyEngineImpl) DescribeWorkflowExecution(
 	// For now execution time will be calculated based on start time and cron schedule/retry policy
 	// each time DescribeWorkflowExecution is called.
 	backoffDuration := time.Duration(0)
-	if executionInfo.HasRetryPolicy && (executionInfo.Attempt > 0) {
-		backoffDuration = time.Duration(float64(executionInfo.InitialInterval)*math.Pow(executionInfo.BackoffCoefficient, float64(executionInfo.Attempt-1))) * time.Second
-	} else if len(executionInfo.CronSchedule) != 0 {
-		backoffDuration = backoff.GetBackoffForNextSchedule(executionInfo.CronSchedule, executionInfo.StartTimestamp)
+	startEvent, ok := msBuilder.GetStartEvent()
+	if ok &&
+		startEvent.GetWorkflowExecutionStartedEventAttributes().GetFirstDecisionTaskBackoffSeconds() > 0 {
```
yycptt: Seems like this check is not needed?

common/backoff/cron.go
```diff
@@ -53,17 +53,22 @@ func GetBackoffForNextSchedule(cronSchedule string, nowTime time.Time) time.Dura
 	if err != nil {
 		return NoBackoff
 	}
-
-	nowTime = nowTime.In(time.UTC)
-	backoffInterval := schedule.Next(nowTime).Sub(nowTime)
+	startUTCTime := startTime.In(time.UTC)
+	closeUTCTime := closeTime.In(time.UTC)
+	nextScheduleTime := schedule.Next(startUTCTime)
+	//Calculate the next schedule start time which is nearest to the close time
```
shreyassrivatsan: nit space after //

common/backoff/cron_test.go
```diff
@@ -33,29 +33,75 @@ func Test_NextCronSchedule(t *testing.T) {
 	// every day cron
 	now, _ := time.Parse(time.RFC3339, "2018-12-17T08:00:00-08:00") // UTC: 2018-12-17 16:00:00 +0000 UTC
```
shreyassrivatsan: let's make this a table driven test.. something like -

```
	testCases := []struct {
	}{
}

	for _, tc := range testCases {
		t.Run(name, func(t *testing.T) {...}
	}

```

common/backoff/cron_test.go
```diff
@@ -33,29 +33,75 @@ func Test_NextCronSchedule(t *testing.T) {
...
+
 	// invalid cron spec
 	cronSpec = "invalid-cron-spec"
-	backoff = GetBackoffForNextSchedule(cronSpec, now)
+	backoff = GetBackoffForNextSchedule(cronSpec, now, now)
 	a.Equal(NoBackoff, backoff)
 }
+
+func Test_Every(t *testing.T) {
+	cronSpec1 := "*/1 * * * *"
```
shreyassrivatsan: same as above, table driven tests..

service/history/stateBuilder.go
```diff
@@ -656,14 +656,21 @@ func (b *stateBuilderImpl) scheduleWorkflowTimerTask(event *shared.HistoryEvent,
 	timerTasks := []persistence.Task{}
 	now := time.Unix(0, event.GetTimestamp())
 	timeout := now.Add(time.Duration(msBuilder.GetExecutionInfo().WorkflowTimeout) * time.Second)
+	backoffDuration := backoff.NoBackoff
+	startWorkflowAttribute := event.GetWorkflowExecutionStartedEventAttributes()
+	if startWorkflowAttribute.GetFirstDecisionTaskBackoffSeconds() > 0 {
+		backoffDuration = time.Duration(event.GetWorkflowExecutionStartedEventAttributes().GetFirstDecisionTaskBackoffSeconds()) * time.Second
```
shreyassrivatsan: nit: can make this a small helper method used by this and history.

service/history/timerQueueProcessor2_test.go
```diff
@@ -361,6 +368,7 @@ func (s *timerQueueProcessor2Suite) TestWorkflowTimeout_Cron() {
 	schedule := "@every 30s"
 
 	builder := newMutableStateBuilderWithEventV2(s.mockShard, s.mockEventsCache, s.logger, we.GetRunId())
+	s.mockShardManager.On("UpdateShard", mock.Anything).Return(nil).Maybe()
```
shreyassrivatsan: why is this `Maybe()`. call should either be made or not, otherwise the test isnt really deterministic.

yux0: This test is not deterministic. I tried to use Once(). But it will fail when running the test as the suite and succeed when running as a single test

shreyassrivatsan: i think we should figure out why its not deterministic and fix that. otherwise this will come to bite us later..

host/integration_test.go
yux0: Make sense. Updated

service/history/historyEngine.go
yux0: removed

common/backoff/cron.go
yux0: updated

service/history/stateBuilder.go
yux0: Removed the history engine logic. So I will keep this as it is

common/backoff/cron_test.go
yux0: changed to table driven test. One minor thing is that it is not straightforward to know which test data fail. Is there any good practice to know which test data fail?

shreyassrivatsan: give each test a name and pass that in as first parameter of `t.Run` then the test failures will show up with test name..

common/backoff/cron_test.go
```diff
@@ -27,35 +27,36 @@ import (
...
+	{"@every 5h", "2018-12-17T08:00:00+00:00", "2018-12-18T00:00:00+00:00", time.Hour * 4},
+}
 
-	// invalid cron spec
-	cronSpec = "invalid-cron-spec"
-	backoff = GetBackoffForNextSchedule(cronSpec, now)
-	a.Equal(NoBackoff, backoff)
+func TestCron(t *testing.T) {
+	for idx, tt := range crontests {
+		t.Run(string(idx), func(t *testing.T) {
```
shreyassrivatsan: pass in a test name here instead of idx for better clarity of what is failing

yux0: With name, it is hard to maintain. I added a message to indicate what is the input and what should be the output

service/history/timerQueueProcessor2_test.go
yux0: This test is testing async method. So the calls might vary.

common/backoff/cron_test.go
shreyassrivatsan: Sure that's also good.. idx makes it clear..

PR #2264, opened 2019-07-23
Implement Sticky TTL

longquanzheng:
For https://github.com/uber/cadence/issues/2261

### DIFF
--- a/service/history/mutableStateBuilder.go
+++ b/service/history/mutableStateBuilder.go
@@ -107,6 +107,7 @@ type (
 		config          *Config
 		timeSource      clock.TimeSource
 		logger          log.Logger
+		domainName      string
 	}
 )
 
@@ -116,6 +117,7 @@ func newMutableStateBuilder(
 	shard ShardContext,
 	eventsCache eventsCache,
 	logger log.Logger,
+	domainName string,
 ) *mutableStateBuilder {
 	s := &mutableStateBuilder{
 		updateActivityInfos:             make(map[*persistence.ActivityInfo]struct{}),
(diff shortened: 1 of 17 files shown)


service/history/service.go
```diff
@@ -252,6 +255,7 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int, enableVisibilit
 		SearchAttributesNumberOfKeysLimit: dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesNumberOfKeysLimit, 100),
 		SearchAttributesSizeOfValueLimit:  dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesSizeOfValueLimit, 2*1024),
 		SearchAttributesTotalSizeLimit:    dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesTotalSizeLimit, 40*1024),
+		StickyTTL:                         dc.GetDurationPropertyFilteredByDomain(dynamicconfig.StickyTTL, time.Hour*3),
```
meiliang86: Use max int64 to make the behavior unchanged if it's not specified? At least make it a bigger default value.

longquanzheng: OK, let me make it one year

service/history/mutableStateBuilder.go
```diff
@@ -586,7 +590,14 @@ func (e *mutableStateBuilder) assignTaskIDToEvents() error {
 }
 
 func (e *mutableStateBuilder) IsStickyTaskListEnabled() bool {
-	return len(e.executionInfo.StickyTaskList) > 0
+	if e.executionInfo.StickyTaskList == "" {
+		return false
+	}
+	maxDu := e.config.StickyTTL(e.domainName)
```
wxing1292: what is MaxDu?

PR #2268, opened 2019-07-24
Improve archival related code

yycptt:


common/archiver/interface.go
```diff
@@ -81,29 +73,18 @@ type (
 
 	// HistoryArchiver is used to archive history and read archived history
 	HistoryArchiver interface {
-		Archive(ctx context.Context, URI string, request *ArchiveHistoryRequest, opts ...ArchiveOption) error
-		Get(ctx context.Context, URI string, request *GetHistoryRequest) (*GetHistoryResponse, error)
-		ValidateURI(URI string) error
+		Archive(ctx context.Context, URI URI, request *ArchiveHistoryRequest, opts ...ArchiveOption) error
```
andrewjdawson2016: Lets make these all unnamed variables, naming is useful when its not clear what the interface does without names but here everything is clear without names.

common/archiver/interface.go
```diff
@@ -81,29 +73,18 @@ type (
...
-
 	// VisibilityBootstrapContainer contains components needed by all visibility Archiver implementations
 	VisibilityBootstrapContainer struct{}
 
 	// VisibilityArchiver is used to archive visibility and read archived visibility
 	VisibilityArchiver interface {
-		Archive(ctx context.Context, URI string, request *ArchiveVisibilityRequest, opts ...ArchiveOption) error
-		Get(ctx context.Context, URI string, request *GetVisibilityRequest) (*GetVisibilityResponse, error)
-		ValidateURI(URI string) error
+		ValidateURI(URI URI) error
```
andrewjdawson2016: Same here.

common/archiver/options.go
```diff
@@ -28,20 +28,25 @@ import (
...
 	}
 
+	// NonRetriableError returns an error indicating archiver has encountered an non-retriable error
+	NonRetriableError func() error
+
 	// ProgressManager is used to record and load archive progress
 	ProgressManager interface {
 		RecordProgress(ctx context.Context, progress interface{}) error
 		LoadProgress(ctx context.Context, valuePtr interface{}) error
+		HasProgress(ctx context.Context) bool
```
andrewjdawson2016: Nice :)

common/archiver/provider/provider.go
```diff
@@ -45,9 +47,9 @@ type (
 			serviceName string,
 			historyContainer *archiver.HistoryBootstrapContainer,
 			visibilityContainter *archiver.VisibilityBootstrapContainer,
-		)
-		GetHistoryArchiver(scheme string, serviceName string) (archiver.HistoryArchiver, error)
-		GetVisibilityArchiver(scheme string, serviceName string) (archiver.VisibilityArchiver, error)
+		) error
+		GetHistoryArchiver(URI archiver.URI, serviceName string) (archiver.HistoryArchiver, error)
```
andrewjdawson2016: I think you should just take `scheme string` here. My earlier suggestion was that you could take `uri string` but that was before realized that URI had to live in archiver package. Now that URI belongs to archiver rather than provider I think provider should just take scheme.

common/archiver/provider/provider.go
```diff
@@ -82,18 +84,31 @@ func NewArchiverProvider(
...
-	p.visibilityContainers[serviceName] = visibilityContainter
+) error {
+	if _, ok := p.historyContainers[serviceName]; ok && historyContainer != nil {
+		return ErrBootstrapContainerAlreadyRegistered
+	}
+	if _, ok := p.visibilityContainers[serviceName]; ok && visibilityContainter != nil {
+		return ErrBootstrapContainerAlreadyRegistered
+	}
+
+	if historyContainer != nil {
```
andrewjdawson2016: I don't think you need these nil checks

common/cluster/archival_status.go
```diff
@@ -31,19 +31,21 @@ type (
 	// ArchivalStatus represents the archival status of the cluster
 	ArchivalStatus int
 
-	// ArchivalConfig is an immutable representation of the current archival configuration of the cluster
+	// ArchivalConfig is an immutable representation of the archival configuration of the cluster
+	// This config is determined at cluster startup time
```
andrewjdawson2016: Can we include a comment in server.go: 

"Dynamic archival config is accessed once on cluster startup than never accessed again. This is done so as to keep archival status and and the initialization of archiver.Provider in sync."

andrewjdawson2016: This comment should be added in addition to the comment you added here. I suggest adding this comment because it is pretty confusing why this archival config requires a restart.

andrewjdawson2016: Can we also add a TODO in server go that says "TODO: Once archival pause is implemented archival config can be made truly dynamic."

config/development.yaml
```diff
@@ -83,18 +83,23 @@ dcRedirectionPolicy:
 
 archival:
   history:
-    status: "disabled"
-    enableReadFromArchival: false
+    status: "enabled"
+    enableReadFromArchival: true
```
andrewjdawson2016: enableRead

config/development.yaml
```diff
@@ -83,18 +83,23 @@ dcRedirectionPolicy:
...
+    provider:
+      filestore:
+        fileMode: "0700"
+        dirMode: "0600"
   visibility:
     status: "disabled"
     enableReadFromArchival: false
```
andrewjdawson2016: Same

service/worker/archiver/client.go
```diff
@@ -128,12 +128,12 @@ func (c *client) archiveInline(ctx context.Context, request *ClientRequest, tagg
 		}
 	}()
 	c.metricsClient.IncCounter(metrics.ArchiverClientScope, metrics.ArchiverClientInlineArchiveAttemptCount)
-	scheme, err := common.GetArchivalScheme(request.ArchiveRequest.URI)
+	parsedURI, err := carchiver.NewURI(request.ArchiveRequest.URI)
```
andrewjdawson2016: Can we call it URI also in the activites.go can we just call it URI

common/archiver/URI.go
```diff
@@ -0,0 +1,93 @@
...
+		String() string
+	}
+
+	uri struct {
+		url *url.URL
+	}
+)
+
+// NewURI constructs a new archiver URI from string.
+func NewURI(s string) (URI, error) {
```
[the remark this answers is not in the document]
yycptt: Added tests. I returned RawPath in this case, as this URI will used by different implementations, I don't think it should do the escape and return the modified path. If an implementation need an escaped path, it can write its own logic to do so.

common/archiver/filestore/historyArchiver.go
```diff
@@ -115,44 +113,45 @@ func newHistoryArchiver(
 
 func (h *historyArchiver) Archive(
 	ctx context.Context,
-	URI string,
+	URI archiver.URI,
 	request *archiver.ArchiveHistoryRequest,
```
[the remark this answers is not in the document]
yycptt: I would consider where and what as two different things. 

Also if by merging them we can end up with the same data structure as the archival signal, I think that will be great. But now we need to add more fields to the signal (TaskID, and maybe more later), I don't think merging URI into request is necessary.

PR #2271, opened 2019-07-24
Enforce decision task start to close timeout upper limit

wxing1292:
* Force override decisoin task start to close timeout to 240s if input is larger
* Increase passive task processing overdue threshold


meiliang86:
Can you file a task to add validation in go client? Java client already has this validation.

wxing1292:
> Can you file a task to add validation in go client? Java client already has this validation.

https://github.com/uber-go/cadence-client/issues/801

PR #2275, opened 2019-07-25
Fix cron integration test 

yux0:
Fix cron integration test


host/integration_test.go
```diff
@@ -1007,6 +997,13 @@ func (s *integrationSuite) TestCronWorkflow() {
 		time.Sleep(200 * time.Millisecond)
 	}
 	s.NotNil(closedExecutions)
+	firstExecutionTime := closedExecutions[0].GetExecutionTime()
+	for i := 1; i != 4; i++ {
+		executionInfo := closedExecutions[i]
+		executionTime := executionInfo.GetExecutionTime()
+		s.Equal(int(0), int(executionTime/1000000000 - firstExecutionTime/1000000000 ) % 3)
```
yycptt: Can we add a comment here explaining why %3 and /1000000000?

host/integration_test.go
```diff
@@ -1973,13 +1970,16 @@ func (s *integrationSuite) TestCronChildWorkflowExecution() {
...
 	}
 	s.NotNil(closedExecutions)
-	for i := 0; i != 4; i++ {
+	firstExecutionTime := closedExecutions[0].GetExecutionTime()
+	for i := 1; i != 4; i++ {
 		executionInfo := closedExecutions[i]
 		if executionInfo.GetExecution().GetWorkflowId() == childID {
-			s.Equal(targetBackoffDuration.Nanoseconds(), executionInfo.GetExecutionTime()-executionInfo.GetStartTime())
+			executionTime := executionInfo.GetExecutionTime()
+			s.Equal(int(0), int(executionTime/1000000000 - firstExecutionTime/1000000000 ) % 3)
```
yycptt: ditto

host/integration_test.go
```diff
@@ -1007,6 +997,13 @@ func (s *integrationSuite) TestCronWorkflow() {
...
+	firstExecutionTime := closedExecutions[0].GetExecutionTime()
+	for i := 1; i != 4; i++ {
+		executionInfo := closedExecutions[i]
+		executionTime := executionInfo.GetExecutionTime()
+		s.Equal(int(0), int(executionTime/1000000000 - firstExecutionTime/1000000000 ) % 3)
+
+	}
 	for i := 0; i != 4; i++ {
 		executionInfo := closedExecutions[i]
 		s.Equal(targetBackoffDuration.Nanoseconds(), executionInfo.GetExecutionTime()-executionInfo.GetStartTime())
```
yycptt: Do we still need this check?

yycptt: and the check on L1021?

host/integration_test.go
yycptt: Also, I think we still need to check the execution time for the first execution, which should be 3s after the start time.

yux0: updated

host/integration_test.go
yux0: Removed them

host/integration_test.go
yux0: For the first execution which is the terminated one, the execution time equals to the start time

host/integration_test.go
yycptt: Sorry, I think I was wrong here. The check on L1021 is testing whether the execution time in the result of DescribeWorkflow is correct or not. Since it's describing the first run, the difference between start and execution time is always 3s.

host/integration_test.go
yycptt: I think by first execution, I mean the first run, which should be continued as new. 

For the last run, which is terminated, can you explain why the execution time equals to the start time? Am I missing something here?

yycptt: When listing closed workflow, the returned execution ordered by increasing close time or decreasing close time? If the first execution is terminated, seems like it's the latter.

host/integration_test.go
yux0: Correct. L1010is duplicated. We still need L1021.

host/integration_test.go
yux0: The latter. The first one is the termination and it will verify below

PR #2279, opened 2019-07-25
Update thrift IDL to support consistent query after signal pattern

andrewjdawson2016:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1333,6 +1340,12 @@ struct RespondQueryTaskCompletedRequest {
   40: optional string errorMessage
 }
 
+struct WorkflowQueryResult {
+  10: optional QueryResultType resultType
+  20: optional binary queryAnswer
```
shreyassrivatsan: `queryResult` to be consistent with the struct name?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1333,6 +1340,12 @@ struct RespondQueryTaskCompletedRequest {
...
+  30: optional string errorMessage
```
mfateev: Should't it be consistent with activity failure which contains both error type and binary details?

idl/github.com/uber/cadence/shared.thrift
andrewjdawson2016: I thought about this but the result is really either an error or an answer. It is still a result if an error is returned from the query handler.

idl/github.com/uber/cadence/shared.thrift
andrewjdawson2016: I think the reason reason ActivityFailure has both is because ActivityReturns errors that look like this 

```
// NewCustomError create new instance of *CustomError with reason and optional details.
// Use CustomError for any use case specific errors that cross activity and child workflow boundaries.
func NewCustomError(reason string, details ...interface{}) *CustomError {
	return internal.NewCustomError(reason, details...)
}
```

Query does not return such an error. If you look at the RespondQueryTaskCompletedRequest it looks like 

```
struct RespondQueryTaskCompletedRequest {
  10: optional binary taskToken
  20: optional QueryTaskCompletedType completedType
  30: optional binary queryResult
  40: optional string errorMessage
}
```

idl/github.com/uber/cadence/shared.thrift
shreyassrivatsan: I think its ok to say `result` even if its an error.. 

Another option is `Response`.. `WorkflowQueryResponse` and `queryResponse`.. I have not really seen answer being used in APIs in general.

idl/github.com/uber/cadence/shared.thrift
mfateev: I know. I think currently we have to reuse queryResult to pass error details which is not very clean design.

idl/github.com/uber/cadence/shared.thrift
andrewjdawson2016: If its called result then there will be a type WorkflowQueryResult.result where result can be nil.

Also if we call it WorkflowQueryResponse this is weird because we suffix things with Response if they are responses to API calls. 

I think answer is the right term, because a query returning an error is also a successful completion of the query from the cadence point of view - so error can be seen as a result just as much as result can be.

shreyassrivatsan: Ok, yeah. that's odd.. Makes sense then..

PR #2280, opened 2019-07-25
Add component for generate event attributes and event batching

yux0:
Add component for generate event attributes and event batching

### DIFF
--- a/common/xdc/event_generator.go
+++ b/common/xdc/event_generator.go
@@ -27,6 +27,11 @@ import (
 	"time"
 )
 
+var (
+	defaultBatchFunc = func(batch []Vertex) bool {
+		return len(batch) == 0
+	}
+)
 type (
 	// Model represents a state transition graph that contains all the relationships of Vertex
 	Model interface {
@@ -55,6 +60,8 @@ type (
 		ListGeneratedVertex() []Vertex
 		// Reset resets the generator to initial state
 		Reset()
+		// SetCanDoBatchOnNextVertex sets a function that used in GetNextVertex to return batch result
+		SetCanDoBatchOnNextVertex(func([]Vertex) bool)
 	}
 
 	// Vertex represents a state in the model. A state represents a type of an Cadence event
@@ -110,6 +117,7 @@ type (
 		exitVertexes        map[Vertex]bool
 		randomEntryVertexes []Vertex
 		dice                *rand.Rand
+		canDoBatch          func([]Vertex) bool
 	}
 
 	// Connection is the edge
@@ -137,6 +145,7 @@ func NewEventGenerator() Generator {
 		exitVertexes:        make(map[Vertex]bool),
 		randomEntryVertexes: make([]Vertex, 0),
 		dice:                rand.New(rand.NewSource(time.Now().Unix())),
+		canDoBatch:          defaultBatchFunc,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


common/xdc/event_generator.go
```diff
@@ -35,40 +41,61 @@ type (
...
 		AddExitVertex(...Vertex)
+		// RandomEntryVertex is a random entry point which can be access at any state of the generator
 		AddRandomEntryVertex(...Vertex)
+		// AddModel loads model into the generator
+		// AddModel can load multiple models and models will be joint if there is common vertexes
 		AddModel(Model)
+		// HasNextVertex determines if there is more vertex to generate
 		HasNextVertex() bool
+		// GetNextVertex generates next vertex batch
 		GetNextVertex() []Vertex
```
wxing1292: GetNextVertices

common/xdc/event_generator.go
```diff
@@ -35,40 +41,61 @@ type (
...
+		// ListGeneratedVertex lists the pasted generated vertexes
 		ListGeneratedVertex() []Vertex
```
wxing1292: ListGeneratedVertices

common/xdc/event_generator.go
```diff
@@ -35,40 +41,61 @@ type (
...
+		// The name of the vertex. Usually, this will be the Cadence event type
 		SetName(string)
 		GetName() string
 		Equals(Vertex) bool
+		// IsStrictOnNextVertex means if the vertex must be followed by its children
+		// When IsStrictOnNextVertex set to true, it means this event can only follow by its neighbors
 		SetIsStrictOnNextVertex(bool)
 		IsStrictOnNextVertex() bool
+		// MaxNextVertex means the max neighbors can branch out from this vertex
+		// MaxNextVertex means the max neighbors can branch out from this vertex
```
wxing1292: copy past err

common/xdc/event_generator.go
```diff
@@ -239,7 +304,7 @@ func (g *EventGenerator) getNextPossibleVertex() Vertex {
 		}
 		neighbors := g.connections[leaf]
 		for _, nextV := range neighbors {
-			if nextV.GetCondition() == nil || (nextV.GetCondition() != nil && nextV.GetCondition()(g.previousVertexes)) {
+			if nextV.GetCondition() == nil || (nextV.GetCondition() != nil && nextV.GetCondition()()) {
```
wxing1292: this is equivalent to `nextV.GetCondition() == nil ||  nextV.GetCondition()()`

wxing1292: on line 312, it it guaranteed to be `isAccessible == false` ?

common/xdc/event_generator.go
```diff
@@ -239,7 +304,7 @@ func (g *EventGenerator) getNextPossibleVertex() Vertex {
...
 				isAccessible = true
 				break
```
wxing1292: once `isAccessible == true` this function can basically return?

common/xdc/event_generator.go
```diff
@@ -26,6 +26,12 @@ import (
...
 
+var (
+	defaultBatchFunc = func(batch []Vertex) bool {
+		return len(batch) == 0
+	}
+)
+
 type (
 	// Model represents a state transition graph that contains all the relationships of Vertex
 	Model interface {
```
wxing1292: it will make the code cleaner if different interfaces / implementations are put in different files

yux0: Moved to a different file

common/xdc/event_generator.go
yux0: Yes. This function is to check if there is a path to move forward. I updated the break to a reurn

common/xdc/event_generator.go
yux0: On 312, yes it is guaranteed or it will break the loop

PR #2282, opened 2019-07-25
Delete archival config from cluster metadata

yycptt:


common/archiver/archivalMetadata.go
```diff
@@ -103,3 +156,21 @@ func getDomainArchivalStatus(str string) (shared.ArchivalStatus, error) {
 	}
 	return shared.ArchivalStatusDisabled, fmt.Errorf("invalid archival status of %v for domain, valid status are: {\"\", \"disabled\", \"enabled\"}", str)
 }
+
+// GetTestArchivalMetadata return an archival metadata instance for test use
```
andrewjdawson2016: This should exist in testing code file?

common/archiver/archivalMetadata.go
```diff
@@ -51,13 +63,54 @@ const (
...
+		visibilityConfig: NewArchivalConfig(
+			visibilityStatus,
+			visibilityReadEnabled,
+			domainDefaults.Visibility.Status,
+			domainDefaults.Visibility.URI,
+		),
+	}
+}
+
+func (metadata *archivalMetadata) GetHistoryConfig() *ArchivalConfig {
```
andrewjdawson2016: Hummm.... its not actually true that ArchivalConfig is immutable the way we are returning it and exporting fields. What was the reason you wanted to remove the getters and return a reference to struct directly?

common/service/service.go
```diff
@@ -109,6 +111,8 @@ type (
 		messagingClient        messaging.Client
 		dynamicCollection      *dynamicconfig.Collection
 		dispatcherProvider     client.DispatcherProvider
+		archivalMetadata       archiver.ArchivalMetadata
```
andrewjdawson2016: This looks nice - good idea.

common/archiver/archivalMetadata.go
yycptt: Removed.

common/archiver/archivalMetadata.go
yycptt: As discussed, created an interface and provided getter methods.

PR #2283, opened 2019-07-25
Add create es index for local docker test

vancexu:


service/frontend/adminHandler.go
```diff
@@ -155,6 +156,13 @@ func (adh *AdminHandler) AddSearchAttribute(ctx context.Context, request *admin.
 			return &gen.BadRequestError{Message: fmt.Sprintf("Unknown value type, %v", v)}
 		}
 		err := adh.params.ESClient.PutMapping(ctx, index, definition.Attr, k, valueType)
+		if err != nil && strings.HasPrefix(err.Error(), "elastic: Error 404") {
```
yycptt: I think the elastic package provides a function to check if the error is 404.

PR #2287, opened 2019-07-26
Simplify archival activity retries 

andrewjdawson2016:


service/worker/archiver/activities.go
```diff
@@ -86,79 +79,49 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		NextEventID:          request.NextEventID,
 		CloseFailoverVersion: request.CloseFailoverVersion,
 	}, carchiver.GetHeartbeatArchiveOption(), carchiver.GetNonRetriableErrorOption(errUploadNonRetriable))
+	if err != nil && err.Error() == errUploadNonRetriable.Error() {
+		return cadence.NewCustomError(errUploadNonRetriable.Error())
```
yycptt: Add logging if the call fails? (Sorry, this is my fault, I forget to do the logging.)
Convert to Cadence custom error in defer()?
Why only non-retriable error is converted to custom error?

service/worker/archiver/activities.go
```diff
@@ -86,79 +79,49 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
...
 	}
-	err := container.HistoryManager.DeleteWorkflowExecutionHistory(deleteHistoryReq)
+	err = container.HistoryManager.DeleteWorkflowExecutionHistory(deleteHistoryReq)
 	if err == nil {
 		return nil
 	}
-	op := func() error {
-		return container.HistoryManager.DeleteWorkflowExecutionHistory(deleteHistoryReq)
-	}
-	for err != nil {
```
yycptt: Neither of the two retry levels are needed?

PR #2296, opened 2019-07-27
Some code / log cleanup

wxing1292:
* Reduce unnecessary logs
* Remove unnecessary processor start delay
* Remove unnecessary errors
* Format code

### DIFF
--- a/service/history/queueProcessor.go
+++ b/service/history/queueProcessor.go
@@ -43,7 +43,6 @@ import (
 type (
 	// QueueProcessorOptions is options passed to queue processor implementation
 	QueueProcessorOptions struct {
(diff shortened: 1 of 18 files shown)


service/history/queueProcessor.go
```diff
@@ -155,8 +162,6 @@ func (p *queueProcessorBase) notifyNewTask() {
 }
 
 func (p *queueProcessorBase) processorPump() {
-	<-time.NewTimer(backoff.NewJitter().JitDuration(p.options.StartDelay(), 0.99)).C
```
yux0: Why we don't need this jitter

wxing1292: this randomized delay was added as part of 2DC, from what i have seen in the past year, this does not really do anything

PR #2298, opened 2019-07-28
Refactor business & persistence layer for nDC

wxing1292:
* Introduce ConflictResolveWorkflowMode indicating whether to update current record
* Add UT for persistence layer conflict resolve workflow execution
* Wire nDC transaction manager to persistence layer
* Add mocks for nDCBranchMgr, nDCStateRebuilder, nDCTransactionMgr and nDCWorkflow
* Add UT for nDC transaction manager


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1600,49 +1602,123 @@ func (d *cassandraPersistence) ConflictResolveWorkflowExecution(request *p.Inter
...
+				shardID,
+				rowTypeExecution,
+				domainID,
+				workflowID,
+				permanentRunID,
+				defaultVisibilityTimestamp,
+				rowTypeExecutionTaskID,
+				prevRunID,
+			)
+		} else {
```
yux0: can this merge of the else if above and just update the prevRunID?

wxing1292: yes, it can, however, it will create a if else within a if else, not sure which solution is better...

PR #2308, opened 2019-07-29
Add docker CLI instructions for host address

longquanzheng:
We had that before but somehow lost it when migrating to a different branch: https://github.com/uber/cadence/pull/1869/files

### DIFF
--- a/docs/08_cli.md
+++ b/docs/08_cli.md
@@ -14,7 +14,12 @@ Example of using the docker image to describe a domain:
 docker run --rm ubercadence/cli:master --domain samples-domain domain describe
 ```
 
-You could use docker image `ubercadence/cli`, by replacing all the following `./cadence ...` with `docker run --rm ubercadence/cli:master --address <HOST_DNS>:7933 ...` . For old versions of docker, HOST_DNS is just 127.0.0.1; for 18.03 onwards, you may have to use host.docker.internal. For more info check https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds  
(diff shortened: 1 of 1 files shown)


docs/08_cli.md
```diff
@@ -14,6 +14,14 @@ Example of using the docker image to describe a domain:
 docker run --rm ubercadence/cli:master --domain samples-domain domain describe
 ```
 
+You could use docker image `ubercadence/cli`, by replacing all the following `./cadence ...` with `docker run --rm ubercadence/cli:master --address <HOST_DNS>:7933 ...` .
```
dmetzgar: I think you're not reading this section. Check out original line 24.

dmetzgar: This line is unnecessary and the <HOST_DNS> part is confusing.

docs/08_cli.md
```diff
@@ -14,6 +14,14 @@ Example of using the docker image to describe a domain:
 docker run --rm ubercadence/cli:master --domain samples-domain domain describe
 ```
 
+You could use docker image `ubercadence/cli`, by replacing all the following `./cadence ...` with `docker run --rm ubercadence/cli:master --address <HOST_DNS>:7933 ...` . 
+
+You may get "connection refused" error for 18.03 onwards versions of docker. You have to use host.docker.internal. For more info check https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds
```
dmetzgar: This part makes sense but I would reword it:
> On Docker versions 18.03 and later, you may get a "connection refused" error. You can work around this by setting the host to "host.docker.internal" (see [here](https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds) for more info).

longquanzheng: done

PR #2310, opened 2019-07-29
Use local system domain for batch operations

longquanzheng:
Tested locally


service/frontend/dcRedirectionPolicy_test.go
```diff
@@ -330,40 +329,6 @@ func (s *selectedAPIsForwardingRedirectionPolicySuite) TestGetTargetDataCenter_G
 	s.Equal(2*len(selectedAPIsForwardingRedirectionPolicyWhitelistedAPIs), alternativeClustercallCount)
 }
 
-func (s *selectedAPIsForwardingRedirectionPolicySuite) TestGetTargetDataCenter_GlobalDomain_Enforce_Forwarding_AlternativeClusterToCurrentCluster() {
```
shreyassrivatsan: why is test deleted?

longquanzheng: Because YARPC doesn't provide a way to test -- no way to put this CallOptions into context ourselves. It's done internally.

PR #2314, opened 2019-07-30
Use go mod

wxing1292:
* Modify Makefile accordingly

After this PR is landed, Cadence repo can live at anywhere.
For development, please move the Cadence repo out of GOPATH


shreyassrivatsan:
let's land this after you are back..

wxing1292:
> let's land this after you are back..

sure

Makefile
```diff
@@ -81,13 +81,13 @@ PKG_TEST_DIRS := $(filter-out $(INTEG_TEST_ROOT)%,$(TEST_DIRS))
 #   Packages are specified as import paths.
 GOCOVERPKG_ARG := -coverpkg="$(PROJECT_ROOT)/common/...,$(PROJECT_ROOT)/service/...,$(PROJECT_ROOT)/client/...,$(PROJECT_ROOT)/tools/..."
 
-dep-ensured:
-	./install-dep.sh
-	dep ensure
+go-mod:
+	go mod tidy
+	go mod vendor
```
sagikazarmark: This is only necessary if you want to vendor dependencies and commit the vendor directory. Since you are not doing it with dep, I'm guessing you don't really need it.

Makefile
```diff
@@ -81,13 +81,13 @@ PKG_TEST_DIRS := $(filter-out $(INTEG_TEST_ROOT)%,$(TEST_DIRS))
 #   Packages are specified as import paths.
 GOCOVERPKG_ARG := -coverpkg="$(PROJECT_ROOT)/common/...,$(PROJECT_ROOT)/service/...,$(PROJECT_ROOT)/client/...,$(PROJECT_ROOT)/tools/..."
 
-dep-ensured:
-	./install-dep.sh
-	dep ensure
+go-mod:
+	go mod tidy
```
sagikazarmark: Go modules are installed automatically when you run `go build` or `go test`. There is no need to run tidy every time.

Makefile
```diff
@@ -81,13 +81,13 @@ PKG_TEST_DIRS := $(filter-out $(INTEG_TEST_ROOT)%,$(TEST_DIRS))
...
+	go mod vendor
 
 yarpc-install:
-	go get './vendor/go.uber.org/thriftrw'
-	go get './vendor/go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc'
+	go get 'go.uber.org/thriftrw'
```
sagikazarmark: FYI This will actually update the dependencies to their latest version every time you run this.

Makefile
```diff
@@ -97,38 +97,38 @@ thriftc: yarpc-install $(THRIFTRW_GEN_SRC)
 copyright: cmd/tools/copyright/licensegen.go
 	GOOS= GOARCH= go run ./cmd/tools/copyright/licensegen.go --verifyOnly
 
-cadence-cassandra-tool: dep-ensured $(TOOLS_SRC)
+cadence-cassandra-tool: go-mod $(TOOLS_SRC)
```
sagikazarmark: As said above, go-mod is not necessary here.

Makefile
wxing1292: ah, that is true, previously the yarpc was using something from vendor dir
```
yarpc-install:
	go get './vendor/go.uber.org/thriftrw'
	go get './vendor/go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc'
```
basically the orioginal code was trying to use the same version of thriftrw and thriftrw-plugin-yarpc for code generation.

the thing is, `go get './vendor/go.uber.org/thriftrw'` can work
however `go get './vendor/go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc'` cannot, due to missing dir (probably due to something similar to https://github.com/golang/go/issues/32504).

any suggestion?

sagikazarmark: If you want to make sure that the same version is used for both libraries, you should probably use the [replace directive](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive)

wxing1292: that is not what I meant, not version control.

the thriftrw and thriftrw-plugin-yarpc are both tool dependency, and previously `go.uber.org/yarpc/encoding/thrift/thriftrw-plugin-yarpc` cannot be successfully imported due to it being a `main package`

I think the correct solution is `tools dependency`

sagikazarmark: Ah, okay, got it.

I would consider using a tool like [gobin](https://github.com/myitcv/gobin) then. (I'm personally not a huge fan of the tool dependency pattern)

Here is an example: https://github.com/sagikazarmark/modern-go-application/blob/master/Makefile#L186-L202

Dockerfile
```diff
@@ -5,14 +5,9 @@ FROM golang:1.12.7-alpine AS builder
 
 RUN apk add --update --no-cache ca-certificates make git curl mercurial bzr
 
-WORKDIR /go/src/github.com/uber/cadence
-
-RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | INSTALL_DIRECTORY=/usr/local/bin sh
-COPY Gopkg.* ./
```
sagikazarmark: This should be replaced with go.* copied into the image and `go mod download` ran in order to build an image cache.

Here is a snippet from our [Dockerfile](https://github.com/banzaicloud/pipeline/blob/master/Dockerfile):

```
ENV GOFLAGS="-mod=readonly"
ARG GOPROXY

COPY go.* .
RUN go mod download
```

The readonly flag makes sure that no new dependencies are installed, thus making sure that there are no surprises around dependencies.

The GOPROXY arg allows using a go modules proxy. It's optional, but allows a developers to build an image with a proxy.

Dockerfile
```diff
@@ -79,7 +74,7 @@ CMD /start.sh
 # Cadence CLI
 FROM alpine AS cadence-cli
 
-COPY --from=builder /go/src/github.com/uber/cadence/cadence /usr/local/bin
+COPY --from=builder /cadence /usr/local/bin
```
sagikazarmark:
```suggestion
COPY --from=builder /cadence/cadence /usr/local/bin
```

Makefile
```diff
@@ -81,11 +81,8 @@ PKG_TEST_DIRS := $(filter-out $(INTEG_TEST_ROOT)%,$(TEST_DIRS))
 #   Packages are specified as import paths.
 GOCOVERPKG_ARG := -coverpkg="$(PROJECT_ROOT)/common/...,$(PROJECT_ROOT)/service/...,$(PROJECT_ROOT)/client/...,$(PROJECT_ROOT)/tools/..."
 
-dep-ensured:
-	./install-dep.sh
-	dep ensure
-
 yarpc-install:
+	go mod vendor
```
sagikazarmark: I still think using [gobin](https://github.com/myitcv/gobin) would be better here. In the same issue that you linked eariler (golang/go#32504) it seems to me that some tooling will be the final solution eventually, not `tools.go`.

docker/buildkite/Dockerfile
```diff
@@ -30,5 +30,5 @@ RUN go get golang.org/x/tools/cmd/cover
 
 # The golang Docker sets the $GOPATH to be /go
```
sagikazarmark: I'm guessing this comment doesn't really make sense anymore.

Dockerfile
wxing1292: thanks for pointing out, the `ENV GOFLAGS="-mod=readonly"` and `ARG GOPROXY` are added.

just 2 questions
1. on line 10 there is a `COPY . .`, which should copy the  `go.mod` and `go.sum`?
2. why `go mod download`? I tested building the docker image locally, and it seems working.

Makefile
wxing1292: for this issue, I need some opinion from my colleagues @shreyassrivatsan @samarabbas @dmetzgar, any idea?

FYI, this PR will not be landed until I get back from, roughly 1 week later.
In the meantime, I will try to do some investigation

sagikazarmark: By using gobin, those two tools won't be added to `go.mod` which means less dependencies. Also, you don't need `go mod vendor`.

Dockerfile
sagikazarmark: In order to utilize the docker build cache, it's a best practice to place things at the beginning of the build that has the least chance to change.

In this case, by downloading dependencies separately first you can spare some time with the build when only the code changes, but not the list of dependencies (which is usually the case).

Otherwise every time something changes, you would have to download the dependencies (which takes time) even if they haven't actually changed.

Although this does not add much in a build environment (if you can't utilize the build cache), it's really helpful when building locally. It doesn't take more time to build the image this way, so this kind of optimization usually make sense.

Makefile
wxing1292: > By using gobin, those two tools won't be added to `go.mod` which means less dependencies. Also, you don't need `go mod vendor`.

for our use case, we prefer to use the binary from dependency `go.uber.org/thriftrw` and `go.uber.org/yarpc` (we are already depending on these 2 lib).
assuming that my understanding of gobin is correct ([ref](https://github.com/myitcv/gobin#examples-using--m)), tool dependency still needs to be declared.

so (still assuming that my understanding is correct) the difference is basically whether it is a good idea to include another dependency.

Dockerfile
wxing1292: i see

Makefile
sagikazarmark: Ah, okay. If they are library dependencies as well, I guess this solution is ok.

PR #2321, opened 2019-07-30
Add Workflow Close Status to Query Results

yycptt:
https://github.com/uber/cadence/issues/2080

CLI change is not included in this diff.

I need to sync idl to the client side before I can update the cli to display close status.


service/history/historyEventNotifier.go
```diff
@@ -60,8 +60,15 @@ type (
...
 
-func newHistoryEventNotification(domainID string, workflowExecution *gen.WorkflowExecution,
-	lastFirstEventID int64, nextEventID int64, previousStartedEventID int64, isWorkflowRunning bool) *historyEventNotification {
+func newHistoryEventNotification(
+	domainID string,
+	workflowExecution *gen.WorkflowExecution,
+	lastFirstEventID int64,
+	nextEventID int64,
+	previousStartedEventID int64,
+	isWorkflowRunning bool,
```
wxing1292: since you are returning closeStatus as additional attr, could you also change isWorkflowRunning to workflow state, i.e. created / running / completed

yycptt: Sure. and the getMutableState history API will also return workflow state instead of isWorkflowRunning?

yycptt:
We need some discussion on the user experience. 

For example: 
    Adding new client side API for returning the close status when querying workflow, or just change the existing API?
    Is there any other information we want to add besides close status?

service/history/historyEventNotifier.go
wxing1292: depends, if you want to make the change, then it requires 2 release
1. return both isWorkflowRunning and workflowState
2. remove isWorkflowRunning

Gopkg.toml
```diff
@@ -140,7 +140,7 @@ ignored = ["github.com/uber/cadence/.gen"]
   version = "1.1.0"
 
 [[constraint]]
-  name = "github.com/hashicorp/go-version"
+  name = "github.com/hashicorp/go-version.git"
```
dmetzgar: Was this intentional? I think Wenquan just removed this.

yycptt: Will remove this after I merge master branch into my branch. My base commit doesn't have this fix https://github.com/uber/cadence/pull/2319.

PR #2323, opened 2019-07-30
Improve Archival logic in TimerQueueProcessor

yycptt:
Also only execute the deleteHistory activity as local activity.


service/worker/archiver/activities.go
```diff
@@ -71,7 +71,7 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		logger.Error(carchiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason("failed to get history archiver"), tag.Error(err))
 		return errUploadNonRetriable
 	}
-	err = historyArchiver.Archive(ctx, URI, &carchiver.ArchiveHistoryRequest{
+	return historyArchiver.Archive(ctx, URI, &carchiver.ArchiveHistoryRequest{
```
yycptt: @andrewjdawson2016 Sorry I made you add the log message. I thought about it, it should be the archiver's responsibility to do the logging.

service/worker/archiver/activities.go
```diff
@@ -82,15 +82,6 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 		NextEventID:          request.NextEventID,
 		CloseFailoverVersion: request.CloseFailoverVersion,
 	}, carchiver.GetHeartbeatArchiveOption(), carchiver.GetNonRetriableErrorOption(errUploadNonRetriable))
-	if err == nil {
```
andrewjdawson2016: I think we want this code?

service/worker/archiver/client.go
```diff
@@ -46,6 +47,11 @@ type (
 		ArchiveInline  bool
 	}
 
+	// ClientResponse is the archive response returned from the archiver client
+	ClientResponse struct {
+		ArchivedInline bool
```
andrewjdawson2016: On the ClientRequest ArchiveInline should be called AttemptArchiveInline

service/worker/archiver/handler.go
```diff
@@ -135,31 +135,12 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
-			NonRetriableErrorReasons: deleteHistoryActivityNonRetryableErrors,
-		},
-	}
-	actCtx = workflow.WithActivityOptions(ctx, ao)
-	if err := workflow.ExecuteActivity(actCtx, deleteHistoryActivityFnName, request).Get(actCtx, nil); err != nil {
-		logger.Error("failed to delete history, this means zombie histories are left", tag.Error(err))
-		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverDeleteFailedAllRetriesCount)
+	if err != nil {
+		logger.Error("deleting history though local activity failed, this means zombie histories are left", tag.Error(err))
+		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverDeleteLocalFailedAllRetriesCount)
```
andrewjdawson2016: Can we change the metric name to just delete now that we are only trying to delete local?

service/worker/archiver/handler.go
```diff
@@ -135,31 +135,12 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
...
-			ExpirationInterval:       5 * time.Minute,
-			NonRetriableErrorReasons: deleteHistoryActivityNonRetryableErrors,
-		},
-	}
-	actCtx = workflow.WithActivityOptions(ctx, ao)
-	if err := workflow.ExecuteActivity(actCtx, deleteHistoryActivityFnName, request).Get(actCtx, nil); err != nil {
-		logger.Error("failed to delete history, this means zombie histories are left", tag.Error(err))
-		metricsClient.IncCounter(metrics.ArchiverScope, metrics.ArchiverDeleteFailedAllRetriesCount)
+	if err != nil {
+		logger.Error("deleting history though local activity failed, this means zombie histories are left", tag.Error(err))
```
andrewjdawson2016: Can we remove the word local from the log. It suggests that we might try to delete some other way.

service/worker/archiver/activities.go
yycptt: Added them back. Idea is it's ok to log in both the activity and in archiver.

PR #2324, opened 2019-07-31
Implement decision Heartbeat timeout 

longquanzheng:
Test will be implemented later before landing.
For https://github.com/uber/cadence/issues/2305


service/history/decisionHandler.go
```diff
@@ -465,36 +466,56 @@ Update_History_Loop:
...
 		var newDecisionTaskScheduledID int64
+		var timeoutDecision bool
 		if createNewDecisionTask {
-			di, err := msBuilder.AddDecisionTaskScheduledEvent()
+			var newDecision *decisionInfo
+			var err error
+			if request.GetForceCreateNewDecisionTask() {
+				// heartbeating decision, timeout the heartbeat based on heartbeat timeout
+				timeout := handler.config.DecisionHeartbeatTimeout(domainEntry.GetInfo().Name)
+				if time.Now().After(time.Unix(0, currentDecision.OriginalScheduledTimestamp).Add(timeout)) {
```
meiliang86: And if there is no new decision

service/history/stateBuilder.go
```diff
@@ -158,7 +158,8 @@ func (b *stateBuilderImpl) applyEvents(domainID, requestID string, execution sha
 		case shared.EventTypeDecisionTaskScheduled:
 			attributes := event.DecisionTaskScheduledEventAttributes
 			di, err := b.msBuilder.ReplicateDecisionTaskScheduledEvent(event.GetVersion(), event.GetEventId(),
-				attributes.TaskList.GetName(), attributes.GetStartToCloseTimeoutSeconds(), attributes.GetAttempt(), event.GetTimestamp())
+				attributes.TaskList.GetName(), attributes.GetStartToCloseTimeoutSeconds(), attributes.GetAttempt(),
+				event.GetTimestamp(), event.GetTimestamp())
```
wxing1292: add a comment here saying the original start time is not used in passive processing logic, this is only a place holder?

service/history/mutableStateBuilder.go
```diff
@@ -1344,8 +1346,8 @@ func (e *mutableStateBuilder) ReplicateWorkflowExecutionStartedEvent(
 	return nil
 }
 
-func (e *mutableStateBuilder) AddDecisionTaskScheduledEvent() (*decisionInfo, error) {
-
+// originalScheduledTimestamp is to record the first scheduled decision during decision heartbeat.
+func (e *mutableStateBuilder) AddDecisionTaskScheduledEventAsHeartbeat(originalScheduledTimestamp int64) (*decisionInfo, error) {
```
wxing1292:
```
AddDecisionTaskScheduledEventAsHeartbeat(
    originalScheduledTimestamp int64,
) (*decisionInfo, error)
```

service/history/mutableStateBuilder_test.go
```diff
@@ -472,7 +472,7 @@ func (s *mutableStateSuite) prepareTransientDecisionCompletionFirstBatchReplicat
 		decisionScheduleEvent.DecisionTaskScheduledEventAttributes.TaskList.GetName(),
 		decisionScheduleEvent.DecisionTaskScheduledEventAttributes.GetStartToCloseTimeoutSeconds(),
 		decisionScheduleEvent.DecisionTaskScheduledEventAttributes.GetAttempt(),
-		0,
+		0, 0,
```
wxing1292: can you put it in the next line?

service/history/mutableStateBuilder_test.go
```diff
@@ -522,7 +522,7 @@ func (s *mutableStateSuite) prepareTransientDecisionCompletionFirstBatchReplicat
 		newDecisionScheduleEvent.DecisionTaskScheduledEventAttributes.TaskList.GetName(),
 		newDecisionScheduleEvent.DecisionTaskScheduledEventAttributes.GetStartToCloseTimeoutSeconds(),
 		newDecisionScheduleEvent.DecisionTaskScheduledEventAttributes.GetAttempt(),
-		0,
+		0, 0,
```
wxing1292: same

service/history/service.go
```diff
@@ -155,6 +155,8 @@ type Config struct {
 
 	// StickyTTL is to expire a sticky tasklist if no update more than this duration
 	StickyTTL dynamicconfig.DurationPropertyFnWithDomainFilter
```
wxing1292: can you put these 2 lines at line 55
moreover 
StickyTTL -> MaxStickyDecisionTTL?
DecisionHeartbeatTimeout -> MaxDecisionHeartbeatTimeout?

we need to keep the code clean, i.e. making variable names readable, not `el`, not `le`, not `maxDu`

longquanzheng: What do you mean MaxDecisionTimeout? Is there a minimumTimeout? And where is el/le/maxDu here?

service/history/mutableStateBuilder_test.go
longquanzheng: OK sure

service/history/mutableStateBuilder_test.go
longquanzheng: ok

service/history/mutableStateBuilder.go
longquanzheng: Why? does it look very long here?

service/history/stateBuilder.go
longquanzheng: Sure, I will add a comment

service/history/service.go
longquanzheng: Also why it has to be at line 55?

service/history/mutableStateBuilder.go
wxing1292: 1. we had similar conversation during the standup
2. yes, it is too long
3. i have already converted most of functions in mutable state builder

service/history/service.go
wxing1292: 1. when naming a variable, we need to make the variable name self-explain.
if you call config.DecisionHeartbeatTimeout() then are you overriding the decision timeout? even if customer provide such timeout? and do not say right now that is what we allow, there is no guarantee that we will allow customized decision heartbeat timeout.

2. about `el`, `le`, `maxDu`, I was referring to variable names you come up with: 
https://github.com/uber/cadence/blame/4f7cb5dc21319c934952717d12c1337aed49d093/common/persistence/historyV2Store.go#L322
https://github.com/uber/cadence/blob/5dcef3f1f2df37fff3f873120329b6ceca94ca87/service/history/mutableStateBuilder.go#L596

3.  StickyTTL is not specific enough

4. if you do not want to move these 2 dynamic config to line 55, can you move `MaxDecisionStartToCloseSeconds ` to here, so all decision related configs are in one location.

longquanzheng: 1. So DecisionHeartBeatTimeout may not be the best name, but MaxDecisionHeartbeatTimtout is even worse. What is "Max" here mean? Do we have median/minimum? And I don't see we have a convention to use Max with Timeout. Timeout is just the Max value implicitly.

And for your confusion, it is not overriding decision timeout, because it is "DecisionHeartbeatTimeout" NOT "DecisionTimeout". DecisionHeartbeat and Decision are different things. 

It may not be the best name, but we got the name from standup and I am not the one who created it.

2. el/le can be improved in later PR if you have concern. And maxDu is already fixed in the PR. 

3. I don't see StickyDecisionTTL is better than StickyTTL. Because Sticky is always for decision. Do we have StickyActivityTTL? 

4. Yes, Grouping similar settings makes sense to me.

longquanzheng: Another reason I don't like StickyDecisionTTL is it seems like the TTL is for decision, but it's actually for the tasklist. So it may be better to say StickyTasklistTTL. But I think they are not too much different.

wxing1292: if a name is not self-explaining, then we should improve it, regardless of who came up with the name.

StickyTTL is not clear enough, if the config is attached to sticky decision & sticky decision's tasklist, then we should name it StickyDecisionTasklistTTL. `Sticky` itself is not clear.

Heartbeat timeout in the existing context only apply to one entity, i.e. activity heartbeat timeout. the new `decision heartbeat timeout` that is being introduced here is little bit different, i.e. this timeout is applied across decisions (with different decision (schedule) IDs).
that is why this `decision heartbeat timeout` thing is confusing. 

I suggest that the name of this config should be reconsidered.

longquanzheng: I can't see a better name right now better decision heartbeat timeout. 
StickyTTL in my opinion is fine, StickyDecisionTasklistTTL is an overkill. It is not always true that the longer name the better. It's super odd to have a name like "StickyDecisionTasklist". I would rather take StickyTasklistTTL as it is what we are talking everyday.

longquanzheng: So I will merge it if you can't give a better name than decision heartbeat timeout. But don't worry, we can still change it since it is not customer facing.

longquanzheng: And if you look at our code base, we have a lot of names as StickyXXX:
StickyTimeout
StickyContext
StickyAttributes
...

If following your logic, all of them will have to be:
StickyDecisionTasklistTimeout
StickyDecisionTasklistContext
StickyDecisionTasklistAttributes

There is always some context to understand the code, there nothing ambiguity in this case. And there is already lots of comments there. We don't want to use code as code comments IMO

wxing1292: for local variable, the context is clear, Sticky* means what is means.

for service level variable, we should do better to make less confusion, especially this project is open sourced.

FYI, i would give `MaxDecisionContinuationTimeout` a try, `Max` is there since this is probably a server side protection, and we probably will introduce a `DecisionContinuationTimeout` for customer to specify

longquanzheng: I think we can leave this "Max" term later when it is really there. Otherwise it cause more confusion. Like I said, what does Max mean if there is no comparison? Because at least in short term, there's no plan to have this for customer to specify. DecisionContinuation sounds odd to me as well. You are inventing new word to describe this behavior. basically we have RespondDecisionTaskComplete with ForCreateNewDecision = true. You are trying to make a new word "DecisionContinuation" to replace what they made "DecisionHeartbeat". In my opinion, they are both odd, and Heartbeat is slightly better than "Continuation". Because decision is alway continue......

longquanzheng: BTW I am fine to rename that StickyTTL, and I still think StickTasklistTTL is much better than StickyDecisionTasklistTTL. I open an issue to do this. Because this is already in staging/production. Simply chaning it here will break our customer because this PR is for different purpose. https://github.com/uber/cadence/issues/2332

wxing1292: `Because decision is alway continue` -> actually, not, for normal case, there is a new decision each time.

for handling local activity, each time, we are still creating a new decision. for server protection, you are adding a timeout for these continued decision.

the continuation here clearly demonstrates that the timeout is applied to a set of continued decisions.

longquanzheng: What you described is not true. Please take a look at the code again. Decision can continue even without ForceCreateNewDecision = true:
```
createNewDecisionTask := !isComplete && (hasUnhandledEvents ||
			request.GetForceCreateNewDecisionTask() || activityNotStartedCancelled)
```
What I am protecting here is simply ONLY for GetForceCreateNewDecisionTask like I said.

wxing1292: > BTW I am fine to rename that StickyTTL, and I still think StickTasklistTTL is much better than StickyDecisionTasklistTTL. I open an issue to do this. Because this is already in staging/production. Simply chaning it here will break our customer.
> #2332

if StickTasklistTTL applies to both activity and decision, then i am all for it

longquanzheng: Yes, it also applies to Local activities and Query. And the original reason to have it is to fix the Query timeout for our customer if you still remember that.

wxing1292: > What you described is not true. Please take a look at the code again. Decision can continue even without ForceCreateNewDecision = true:
> 
> ```
> createNewDecisionTask := !isComplete && (hasUnhandledEvents ||
> 			request.GetForceCreateNewDecisionTask() || activityNotStartedCancelled)
> ```
> 
> What I am protecting here is simply ONLY for GetForceCreateNewDecisionTask like I said.

then it is still across multiple decisions: `createNewDecisionTask`

wxing1292: seems that we are not really convincing each other, let us create an issue about this.

longquanzheng: Makes sense. Let's leave this for now. It's never late to do refactoring and rename: https://github.com/uber/cadence/issues/2334

PR #2325, opened 2019-07-31
New Dockerfiles

sagikazarmark:
This PR introduces new Dockerfiles, ones which work better in production environments
(the currently included one is more suited for local environments where people quickly spin up Cadence to try it)
and follow Docker best pratices.

I created everything next to the current setup, (the old Dockerfiles are still available in the repo).

## Notable changes

- Dockerfiles moved to the repository root
- Code is copied from the filesystem instead of being cloned from a git repository
- All images are rewritten in a way that they can better utilise build cache
- All images use Alpine instead of Debian
- CGO is disabled (it does not seem to cause any trouble)
- The server images is split into two images:
    - The *normal* server image does not contain anything else, but Cadence.
    - The *allinone* server image contains most of the tools the current image includes (migration tools, etc)
- All images are backward compatible with earlier images (in case of the server this is only true for the *allinone* image)
- Binaries are in `/usr/local/bin`
- Schema and configs are in `/etc/cadence`

## Motivation

I'm working with Cadence for almost a year now, but running it in production using the official Docker images wasn't always easy.
Recently, as I worked on the Helm chart, I found a number of problems/things to improve. Originally, I wanted to create custom images,
but I decided to submit them for your consideration as well.

## Building the images

The server and the cli are probably straightforward:

```bash
docker build -t ubercadence/server:local .
docker build -t ubercadence/cli:local -f Dockerfile.cli .
```

The allinone image is a bit tricky:

```bash
docker build -t ubercadence/server:local-allinone -f Dockerfile.allinone --build-arg CADENCE_IMAGE=ubercadence/server:local .
```

Basically, it uses the "production" image as a base and adds the remaining scripts and dependencies to the final image.


.dockerignore
```diff
@@ -0,0 +1 @@
+vendor/
```
longquanzheng: Can we get rid of this file?

sagikazarmark: Why? This ensures that dep will always download the required dependencies according to the `Gopkg.lock` file in the repo. It also makes the build context smaller.

Dockerfile
```diff
@@ -0,0 +1,61 @@
+# Build cadence binaries
```
longquanzheng: Do we have to put the Dockerfiles in root directory? It would be nice if we could keep it inside docker folder.

sagikazarmark: It's kind of a standard for Dockerfiles to be in the repository root. Also, I wanted to keep as much of the existing setup as possible, so that any kind of automated builds you might have didn't break.

It's not a requirement, but as I said, it's the standard place.

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+FROM golang:1.12.7-alpine AS builder
+
+RUN apk add --update --no-cache ca-certificates make git curl mercurial bzr
+
+# Install dep
+ENV DEP_VERSION=0.5.4
+RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | INSTALL_DIRECTORY=/usr/local/bin DEP_RELEASE_TAG=v${DEP_VERSION} sh
```
longquanzheng: we always use the latest dep: https://github.com/uber/cadence/blob/master/install-dep.sh

sagikazarmark: That's generally not a good practice if you want to ensure repeatable builds.

Dockerfile.cli
```diff
@@ -0,0 +1,33 @@
+# Build cadence binaries
+FROM golang:1.12.7-alpine AS builder
+
+RUN apk add --update --no-cache ca-certificates make git curl mercurial bzr
+
+# Install dep
+ENV DEP_VERSION=0.5.4
```
longquanzheng: same here.

docker/config/dynamicconfig/development_es.yaml
```diff
@@ -0,0 +1,29 @@
+frontend.enableClientVersionCheck:
```
longquanzheng: can we copy it from the /config/dynamicconfig? So that we don't have to maintain two copies. Because this file is changing fast.

sagikazarmark: Sure

Dockerfile
longquanzheng: I see, let's follow the standard then.

Dockerfile
longquanzheng: That's fine. I think what we decided is always using latest one, if it doesn't build, we fix it right away.

Dockerfile
longquanzheng: Is it also standard to have Dockerfile.allinone and Dockerfile.CLI in the root? Can we only have one? Otherwise the root looks messy.

.dockerignore
longquanzheng: Can you explain a little bit why do we need this? 
Seems like the way it works changes. It used to clone the repo and build from there. But here  you are building from current repo, right? But I still don't understand what this is for. Because dep always download from Gopkg.lock. "makes the build contextw smaller" do you mean the it relates to the image?

Dockerfile.allinone
```diff
@@ -0,0 +1,10 @@
+ARG CADENCE_IMAGE=ubercadence/server:latest
```
longquanzheng: Is it also standard to have Dockerfile.allinone and Dockerfile.CLI in the root? Can we only have one? Otherwise the root looks messy.

sagikazarmark: Multi stage Dockerfiles can be built until a specific target, so we can merge the three into one, but it would make the build process slightly more complicated. I don't think there is any rule for multiple dockerfiles, but I don't find it messy at all.

That being said, the production and the allinone images could also be merged and configured with build arguments (`--build-arg ALLINONE=true`). Would it be okay to have one for the cli and one for the two server images?

longquanzheng: I think using that argument would be a better idea. CLI doesn't have to use different dockerfile

docker/docker_start.sh
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash -x
```
longquanzheng: There seem to be a lot of duplicated code. What is the main difference start.sh and docker_start.sh?

sagikazarmark: `docker_start.sh` is the old starter script with some additional environment setup. Binaries are looked up from a different place.

`start.sh` is the "new" code. The previous file can be deleted, it has only been preserved so that the current Dockerfile does not break in any way.

longquanzheng: OK. I think you can remove the old script as well as dockerfile. We don't need to keep both

.dockerignore
sagikazarmark: Yeah, one of the biggest changes here is instead of cloning the repo, docker uses the working copy to build the image. This is aligned with the current best practices of docker image building.

Files listed in docker ignore won't be copied into the image by `ADD` and `COPY` commands. Ignored files are also not passed as build context, which is the list of files and their content. When building an image, the commands in the dockerfile use the so called build cache. When a command's result will be the same as another command from a previous build (eg. copy file A into the image. if fileA has the same content as in a previous build, docker will use that layer from the already built image, instead of copying the file again). This is an oversimplified explanation of what build context is and how it relates to build cache. To put it simple: it makes the image build slightly faster, but does not have an effect on the image size.

The reason why vendor is ignored is to make sure as much building happens within the docker context as possible. Basically, everything that's not source code copying, should happen within docker.

Another reason is the aforementioned build cache. This way if `Gopkg.lock` didn't change since the last build, docker will simply use an earlier layer from the cache, gaining some speed and storage, but again, it does not affect the final image size.

The thing is: it's hard to explain why these practices make sense and why they make things easier without an in-depth knowledge of Docker, so I would suggest reading the best practices in the Docker documentation, because I'm pretty sure it explains these things much-much better, than I do.

Dockerfile.allinone
sagikazarmark: > CLI doesn't have to use different dockerfile

What do you mean? I proposed two separate solutions, the build arg one would still have a separate dockerfile for the cli. Alternatively, we can get rid of the cli dockerfile altogether, but that would require some documentation changes (prepend cadence commands with `cadence`)

docker/docker_start.sh
sagikazarmark: ok

sagikazarmark:
Okay, I think I found a way to merge all the three dockerfiles, but it comes at a price:

```
docker build -t ubercadence/server:local .
docker build -t ubercadence/cli:local --build-arg TARGET=cli .
docker build -t ubercadence/server:local-allinone --build-arg TARGET=allinone .
```

The price is that regardless which image you want to build, you have to build all three of them (except if you use BuildKit, which skips unused stages).

Does this work for you?

Dockerfile.allinone
sagikazarmark: I think this one is resolved by the single dockerfile

longquanzheng:
> Okay, I think I found a way to merge all the three dockerfiles, but it comes at a price:
> 
> ```
> docker build -t ubercadence/server:local .
> docker build -t ubercadence/cli:local --build-arg TARGET=cli .
> docker build -t ubercadence/server:local-allinone --build-arg TARGET=allinone .
> ```
> 
> The price is that regardless which image you want to build, you have to build all three of them (except if you use BuildKit, which skips unused stages).
> 
> Does this work for you?

Thanks for the update. I think it's fine.

longquanzheng:
@samarabbas  do you also want to take a look?

.dockerignore
longquanzheng: Got it. That makes a lot of sense to me. Thanks!

h7kanna:
There is a pending request to introduce go mod https://github.com/uber/cadence/pull/2314. May be Dockerfile should include go mod instead of dep.

sagikazarmark:
@h7kanna that's really a minor change to the Dockerfile. But I'd be happy to help out with go modules as well.

sagikazarmark:
@longquanzheng do you think you can release a patch version using this Dockerfile? I would use the changes in the helm chart.

longquanzheng:
> @longquanzheng do you think you can release a patch version using this Dockerfile? I would use the changes in the helm chart.

We started to follow the semantic versioning recently, so only do patch for bug fixes. But we usually do release every 2 weeks so it soon be there soon.
There is a release 0.7.0 ongoing this week. So next week this change will be released in 0.8.0

sagikazarmark:
Thanks! Any chance we can backport this PR to the planned 0.7 release?

dmetzgar:
@sagikazarmark @longquanzheng is there any change needed to our [buildkite docker files](https://github.com/uber/cadence/tree/master/docker/buildkite)?

longquanzheng:
> Thanks! Any chance we can backport this PR to the planned 0.7 release?

Let me check with my teammate. It should be OK.

longquanzheng:
> @sagikazarmark @longquanzheng is there any change needed to our [buildkite docker files](https://github.com/uber/cadence/tree/master/docker/buildkite)?

I am doing it in this PR: https://github.com/uber/cadence/pull/2331
@dmetzgar

sevein:
The `TARGET` trick is amazing, thanks @sagikazarmark!

PR #2335, opened 2019-08-01
Update archival relate metrics and logger tags

yycptt:


common/metrics/defs.go
```diff
@@ -1265,21 +1265,23 @@ const (
...
+	HistoryArchiverArchiveTransientErrorCount
 	HistoryArchiverHistoryMutatedCount
-	HistoryArchiverTotalUploadSize
+	HistoryArchiverUploadedHistorySize
+	HistoryArchiverHistorySize
 
 	// The following metrics are only used by internal archiver implemention.
 	// TODO: move them to internal repo once cadence plugin model is in place.
 	HistoryArchiverBlobExistsCount
+	HistoryArchiverUploadedBlobSize
```
andrewjdawson2016: rename: HistoryArchiverBlobSize - it might be the size of a blob which we did not upload

common/metrics/defs.go
```diff
@@ -1265,21 +1265,23 @@ const (
 	SequentialTaskQueueProcessingLatency
 	SequentialTaskTaskProcessingLatency
 
+	HistoryArchiverArchiveNonRetryableErrorCount
+	HistoryArchiverArchiveTransientErrorCount
 	HistoryArchiverHistoryMutatedCount
-	HistoryArchiverTotalUploadSize
+	HistoryArchiverUploadedHistorySize
```
andrewjdawson2016: rename: HistoryArchiverUploadedSize - like you pointed out this metric does not indicate history size it just indicates load on archiver, so no need to include history in metric name

service/worker/archiver/handler.go
```diff
@@ -104,7 +104,7 @@ func handleRequest(ctx workflow.Context, logger log.Logger, metricsClient metric
 	logger = tagLoggerWithRequest(logger, request)
 	ao := workflow.ActivityOptions{
 		ScheduleToStartTimeout: 1 * time.Minute,
-		StartToCloseTimeout:    2 * time.Minute,
+		StartToCloseTimeout:    1 * time.Minute,
```
andrewjdawson2016: Can we also update ExpirationInterval on delete to be 5 minutes. Now that we deleted the real delete activity I think local should have more retry time.

PR #2340, opened 2019-08-02
Archival Bug Fix

yycptt:
1. More validation on archival config
2. Also update the fileMode and dirMode used for local filestore.
3. Return detailed error instead of "Invalid URI" in DomainHandler.


common/archiver/archivalMetadata.go
```diff
@@ -73,25 +74,70 @@ const (
...
 	domainDefaults *config.ArchivalDomainDefaults,
 ) ArchivalMetadata {
+	historyConfig := NewArchivalConfig(
+		historyStatus,
+		historyReadEnabled,
+		domainDefaults.History.Status,
+		domainDefaults.History.URI,
+	)
+	if historyConfig.ClusterConfiguredForArchival() {
+		// Only check dynamic config when archival is enabled in static config.
```
andrewjdawson2016: Can you add to this comment a little. Basically summarize our one way config conversation.

common/archiver/archivalMetadata.go
```diff
@@ -73,25 +74,70 @@ const (
...
+		// Dynamic archival config is accessed once on cluster startup than never accessed again.
+		// This is done so as to keep archival status and and the initialization of archiver.Provider in sync.
+		// TODO: Once archival pause is implemented archival config can be made truly dynamic.
+		dynamicHistoryStatusString := dc.GetStringProperty(dynamicconfig.HistoryArchivalStatus, historyStatus)()
+		dynamicHistoryStatus, err := getClusterArchivalStatus(dynamicHistoryStatusString)
+		if err != nil {
+			panic(err)
```
andrewjdawson2016: I am a little scared about panic based on a dynamic config value. It seems fine to panic based on static config because there is nothing you can do in that case. But here if someone changes dynamic config to invalid value then restarts the cluster all the nodes will start panic on startup and we will have an outage. Can we instead just default to disabled if we cannot read archival config. 

If someone meant to enable but actually entered "emabled" (or some other mispelling) its reasonable that archival will stay disabled.

andrewjdawson2016: Sorry I should have thought about that earlier.

common/archiver/archivalMetadata.go
```diff
@@ -154,7 +210,7 @@ func (a *archivalConfig) GetDomainDefaultURI() string {
 
 func (a *archivalConfig) isValid() bool {
 	URISet := len(a.domainDefaultURI) != 0
-	return (URISet && a.ClusterConfiguredForArchival()) || (!URISet && !a.ClusterConfiguredForArchival())
+	return (URISet && a.ClusterConfiguredForArchival()) || (!a.enableRead && a.domainDefaultStatus == shared.ArchivalStatusDisabled && !URISet && !a.ClusterConfiguredForArchival())
```
andrewjdawson2016: Can you factor this out into three lines

validEnabled := ...
validDisabled := ...
return validEnabled || validDisabled

config/development.yaml
```diff
@@ -87,8 +87,8 @@ archival:
     enableRead: true
     provider:
       filestore:
-        fileMode: "0700"
-        dirMode: "0600"
+        fileMode: "0766"
```
andrewjdawson2016: 0666?

No one should need to execute right?

common/archiver/archivalMetadata.go
```diff
@@ -73,25 +74,70 @@ const (
...
+		visibilityStatus,
+		visibilityReadEnabled,
+		domainDefaults.Visibility.Status,
+		domainDefaults.Visibility.URI,
+	)
+	if visibilityConfig.ClusterConfiguredForArchival() {
+		dynamicVisibilityStatusString := dc.GetStringProperty(dynamicconfig.VisibilityArchivalStatus, visibilityStatus)()
+		dynamicVisibilityStatus, err := getClusterArchivalStatus(dynamicVisibilityStatusString)
+		if err != nil {
+			panic(err)
```
andrewjdawson2016: Same comment here on the panic

common/archiver/archivalMetadata.go
```diff
@@ -73,25 +74,70 @@ const (
...
-			domainDefaults.History.URI,
-		),
-		visibilityConfig: NewArchivalConfig(
-			visibilityStatus,
-			visibilityReadEnabled,
-			domainDefaults.Visibility.Status,
-			domainDefaults.Visibility.URI,
-		),
+		historyConfig:    historyConfig,
+		visibilityConfig: visibilityConfig,
```
andrewjdawson2016: What do you think about just setting visibilityConfig: NewDisabledArchivalConfg()

The advantage of this I see is it becomes really obvious that visibility archival cannot be enabled

PR #2351, opened 2019-08-06
use workflow execution time to calculate cron backoff

yux0:
use workflow execution time to calculate cron backoff.
Before, it use the workflow start time, which could be vary based on the workflow close time.
The execution time is the workflow actual execution time.

### DIFF
--- a/service/history/mutableStateBuilder.go
+++ b/service/history/mutableStateBuilder.go
@@ -772,22 +772,24 @@ func (e *mutableStateBuilder) GetRetryBackoffDuration(errReason string) time.Dur
 	return getBackoffInterval(info.Attempt, info.MaximumAttempts, info.InitialInterval, info.MaximumInterval, info.BackoffCoefficient, e.timeSource.Now(), info.ExpirationTime, errReason, info.NonRetriableErrors)
 }
 
-func (e *mutableStateBuilder) GetCronBackoffDuration() time.Duration {
+func (e *mutableStateBuilder) GetCronBackoffDuration() (time.Duration, error) {
 	info := e.executionInfo
 	if len(info.CronSchedule) == 0 {
-		return backoff.NoBackoff
+		return backoff.NoBackoff, nil
 	}
-
 	// TODO: decide if we can add execution time in execution info.
 	executionTime := e.executionInfo.StartTimestamp
 	// This only call when doing ContinueAsNew. At this point, the workflow should have a start event
 	workflowStartEvent, found := e.GetStartEvent()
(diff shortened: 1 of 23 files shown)


service/history/mutableStateBuilder.go
```diff
@@ -777,7 +777,14 @@ func (e *mutableStateBuilder) GetCronBackoffDuration() time.Duration {
 	if len(info.CronSchedule) == 0 {
 		return backoff.NoBackoff
 	}
-	return backoff.GetBackoffForNextSchedule(info.CronSchedule, e.executionInfo.StartTimestamp, e.timeSource.Now())
+
+	workflowStartEvent, found := e.GetStartEvent()
+	executionTime := e.executionInfo.StartTimestamp
+	if found {
```
wxing1292: if this found / not found applies to both initial run of cron workflow & following run of cron workflow, plz include a comment, because normally `e.GetStartEvent()` should be able to load the start event

service/history/mutableStateBuilder.go
```diff
@@ -777,7 +777,14 @@ func (e *mutableStateBuilder) GetCronBackoffDuration() time.Duration {
...
+		firstDecisionTaskBackoffInSeconds := time.Duration(workflowStartEvent.GetWorkflowExecutionStartedEventAttributes().GetFirstDecisionTaskBackoffSeconds()) * time.Second
```
wxing1292: plz indent this line

service/history/mutableStateBuilder.go
yux0: No. Only for continueasnew cron workflow. Not the inital run. So I can just ignore it? When continueasnew, there should be the start event.

service/history/mutableStateBuilder.go
yux0: Good catch

service/history/mutableStateBuilder.go
yux0: I updated the comment but keep this found. Just to be safe.

wxing1292: no, not exactly

if `e.GetStartEvent()` is for following runs, i.e. calling the existing run to calculate the next run's property, then it must be `found == true`.

so i guess you need to modify this method signature to include a `error`

wxing1292: the basic idea is that we should not ignore potential error case

yux0: Yes. I updated the method to return an error.

PR #2364, opened 2019-08-09
Update README.md

meiliang86:


service/worker/README.md
```diff
@@ -40,6 +40,15 @@ bin/kafka-topics.sh --create --zookeeper localhost:2181 --replication-factor 1 -
...
 ```
 
+5. Create global domains
+```
+cadence --do sample domain register --gd true --ac active --cl active standby
+```
+
+6. Failover between zones:
+```
+cadence --do sample domain update --ac standby
```
longquanzheng: This command only failovers to standby, can you add another one for active? Or maybe use <standby/active>

PR #2382, opened 2019-08-13
Add domain tags to Decision/Activity timeout metrics

longquanzheng:


service/history/timerQueueActiveProcessor.go
```diff
@@ -430,10 +430,17 @@ ExpireActivityTimers:
 			}
 		}
 
+		domainEntry, err := t.shard.GetDomainCache().GetDomainByID(msBuilder.GetExecutionInfo().DomainID)
+		if err != nil {
+			return &workflow.InternalServiceError{Message: "Unable to get domain from cache by domainID."}
```
yux0: nit: first letter lower case for the error message.

service/history/timerQueueActiveProcessor.go
```diff
@@ -520,17 +533,26 @@ func (t *timerQueueActiveProcessorImpl) processDecisionTimeout(
 		return nil
 	}
 
+	domainEntry, err := t.shard.GetDomainCache().GetDomainByID(msBuilder.GetExecutionInfo().DomainID)
+	if err != nil {
+		return &workflow.InternalServiceError{Message: "Unable to get domain from cache by domainID."}
```
yux0: same

service/history/timerQueueActiveProcessor.go
longquanzheng: OK

service/history/timerQueueActiveProcessor.go
```diff
@@ -430,10 +430,17 @@ ExpireActivityTimers:
...
+		domainEntry, err := t.shard.GetDomainCache().GetDomainByID(msBuilder.GetExecutionInfo().DomainID)
+		if err != nil {
+			return &workflow.InternalServiceError{Message: "unable to get domain from cache by domainID."}
+		}
+
 		switch timeoutType {
 		case workflow.TimeoutTypeScheduleToClose:
 			{
-				t.metricsClient.IncCounter(metrics.TimerActiveTaskActivityTimeoutScope, metrics.ScheduleToCloseTimeoutCounter)
+				t.metricsClient.Scope(metrics.TimerActiveTaskActivityTimeoutScope).
```
vancexu: Should we tag scope out so that scope and `.Scope` and `.Tag` will be reused?

longquanzheng: It can't be reused if you look into the code. The scopes are different for each type.

vancexu: Probably I didn't explain clear. I saw several duplicates
```
t.metricsClient.Scope(metrics.TimerActiveTaskActivityTimeoutScope).
					Tagged(metrics.DomainTag(domainEntry.GetInfo().Name)).
```
and I think we could reduce that. https://github.com/uber/cadence/pull/2386

PR #2386, opened 2019-08-14
Refactor metrics with domain tag

vancexu:
Small efforts to reduce some duplicates


service/history/timerQueueActiveProcessor.go
```diff
@@ -842,3 +834,12 @@ func (t *timerQueueActiveProcessorImpl) updateWorkflowExecution(
 
 	return nil
 }
+
+func (t *timerQueueActiveProcessorImpl) getMetricScopeWithDomainTag(scope int, domainID string) (metrics.Scope, error) {
+	domainEntry, err := t.shard.GetDomainCache().GetDomainByID(domainID)
+	if err != nil {
+		return nil, &workflow.InternalServiceError{Message: "unable to get domain from cache by domainID."}
```
longquanzheng: Actually we should return this error instead of making a new one.

vancexu: make sense. updated

PR #2404, opened 2019-08-16
IDL change to ignore queries on workflows in closed or terminated state

andrewjdawson2016:


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1316,10 +1321,12 @@ struct QueryWorkflowRequest {
   10: optional string domain
   20: optional WorkflowExecution execution
   30: optional WorkflowQuery query
+  40: optional QueryIgnoreState ignoreState
```
mfateev: I propose calling it QueryRejectIf with : NotOpen, Failed values

PR #2413, opened 2019-08-19
Reject query based on workflow state

andrewjdawson2016:


service/history/historyEngine.go
```diff
@@ -637,6 +637,11 @@ func (e *historyEngineImpl) getMutableState(
 		BranchToken:            msBuilder.GetCurrentBranch(),
 	}
 
+	if !msBuilder.IsWorkflowExecutionRunning() {
```
wxing1292: 3+DC will have additional zombie state, perhaps, you should check whether the closeStatus is not 'none'

PR #2417, opened 2019-08-20
Use one set of worker queues for timer processing

shreyassrivatsan:
Pull out worker queue logic - initialize and use in timer queue implementations.


service/history/timerQueueTaskProcessor.go
```diff
@@ -0,0 +1,322 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: plz also port the existing tests to `timerQueueTaskProcessor_test.go`

service/history/timerQueueProcessorBase_test.go
```diff
@@ -66,7 +66,8 @@ type (
 		notificationChan chan struct{}
```
wxing1292: FYI timerQueueProcessor_test.go and timerQueueProcessor2_test.go needs rewrite (to become a unit test, not integration test which includes db)

service/history/timerQueueTaskProcessor.go
shreyassrivatsan: done

service/history/timerQueueProcessorBase_test.go
shreyassrivatsan: yes, I saw that.. will change as part of future diffs..

PR #2419, opened 2019-08-20
New mutable state task refresher for mutable state task rebuild

wxing1292:
* New mutable state task refresher which can re-generate tasks according to the current state of mutable state
* Add batch event ID to request cancel info & signal info for mutable state task refresher
* Format state builder
* Modifying mutableStateTaskGenerator accordingly
* Rename GetInFlightDecisionTask -> GetInFlightDecision
* Rename GetPendingDecision -> GetDecisionInfo
* Rename GetAllRequestCancels -> GetPendingRequestCancelInfos
* Rename GetAllSignalsToSend -> GetPendingSignalInfos
* Rename HasPendingDecisionTask -> HasPendingDecision
* Rename HasProcessedOrPendingDecisionTask -> HasProcessedOrPendingDecision


service/history/mutableState.go
```diff
@@ -120,37 +120,38 @@ type (
 		GetChildExecutionInfo(int64) (*persistence.ChildExecutionInfo, bool)
 		GetChildExecutionInitiatedEvent(int64) (*workflow.HistoryEvent, bool)
 		GetCompletionEvent() (*workflow.HistoryEvent, bool)
+		GetDecisionInfo(int64) (*decisionInfo, bool)
```
longquanzheng: The other renaming looks good except for this one. GetDecisionInfo(initiatedID) sounds like it is able to get any decision but it is actually getting current decision. I would prefer GetPendingDecisionInfo or GetCurrentDecisionInfo

longquanzheng: There is already a GetPendingDecision(). What this does is Get and check if the decisionID is the same.

idl/github.com/uber/cadence/sqlblobs.thrift
```diff
@@ -181,6 +181,7 @@ struct ChildExecutionInfo {
 
 struct SignalInfo {
   10: optional i64 (js.type = "Long") version
+  11: optional i64 (js.type = "Long") initiatedEventBatchID
```
longquanzheng: Thanks for adding these

service/history/mutableState.go
```diff
@@ -120,37 +120,38 @@ type (
...
 		GetLastFirstEventID() int64
 		GetLastWriteVersion() int64
 		GetNextEventID() int64
 		GetPreviousStartedEventID() int64
-		GetPendingDecision(int64) (*decisionInfo, bool)
 		GetPendingActivityInfos() map[int64]*persistence.ActivityInfo
 		GetPendingTimerInfos() map[string]*persistence.TimerInfo
 		GetPendingChildExecutionInfos() map[int64]*persistence.ChildExecutionInfo
+		GetPendingRequestCancelInfos() map[int64]*persistence.RequestCancelInfo
+		GetPendingSignalInfos() map[int64]*persistence.SignalInfo
```
longquanzheng: Can we call it GetPendingSignalExternalInfos()? GetPendingSignalInfos sounds like getting received signals.

service/history/mutableStateBuilder.go
```diff
@@ -4258,7 +4282,7 @@ func (e *mutableStateBuilder) closeTransactionHandleDecisionFailover(
 	}
 
 	// Handling mutable state turn from standby to active, while having a decision on the fly
-	di, ok := e.GetInFlightDecisionTask()
+	di, ok := e.GetInFlightDecision()
```
longquanzheng: nit: di=> decisionInfo

service/history/mutableStateTaskGenerator.go
```diff
@@ -112,14 +117,11 @@ func newMutableStateTaskGenerator(
...
+	nowTimestamp int64,
 	event *shared.HistoryEvent,
 ) error {
 
-	// TODO when refactoring workflow reset functionality,
-	//  the logic below should be changed slightly, i.e. 'now' should not be derived from
-	//  start event, but the reset time
-
-	now := time.Unix(0, event.GetTimestamp())
+	now := time.Unix(0, nowTimestamp)
```
longquanzheng: I remember we want to use timeSource for testiability, is it possible to do it here?

service/history/mutableStateTaskRefresher.go
```diff
@@ -0,0 +1,394 @@
...
+) error {
+
+	if !r.mutableState.HasPendingDecision() {
+		// no decision task at all
+		return nil
+	}
+
+	decision, ok := r.mutableState.GetPendingDecision()
+	if !ok {
+		return &shared.InternalServiceError{Message: "impossible case: cannot get pending decision"}
```
longquanzheng: it's better to remove "impossible case" because it doesn't give any valuable information. As an error, then it returns here, it is already possible.

service/history/mutableStateTaskRefresher.go
```diff
@@ -0,0 +1,394 @@
...
+			scheduleEvent,
+		); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *mutableStateTaskRefresherImpl) refreshTasksForCancelExternalWorkflow(
```
longquanzheng: nit: use RequestCancel insteand of Cancel to be consistent.

service/history/mutableState.go
wxing1292: we actually do `GetActivityInfo` which returns activity info and a bool, just following the same pattern here

GetPendingDecisionInfo() basically just get the pending decision, like GetPendingActivityInfos

service/history/mutableState.go
wxing1292: sure

service/history/mutableStateBuilder.go

service/history/mutableStateTaskRefresher.go
wxing1292: what if we actually have a bug? handling impossible case (i.e. return error rather than silently ignore the error) helps debugging, and in the future, refactoring.

service/history/mutableStateBuilder.go
wxing1292: changing di -> decision or otherwise will collide with the decisionInfo type

service/history/mutableStateTaskRefresher.go
wxing1292: sure

wxing1292:
for other var renaming, e.g. ai -> acrtivityInfo, si -> signalInfo, rci -> requestCancelInfo will do another PR

service/history/mutableStateTaskRefresher.go
longquanzheng: I understand what you are trying to give. This error is more critical than other errors like ConditionFailed/Timeout/ShardLost etc. When this error happens, it means we could have a bug. 

Imagine you run a command line and it returns an error telling you this is impossible... Whether it is impossible or not is too subjective to tell at that point. What you need is provide the fact what is causing error.

So I would suggest to reword it to "cannot get pending decision, it could be a bug". It will be more objective.

wxing1292: done

PR #2421, opened 2019-08-21
Admin CLI add delete visibility from ES

vancexu:
image
image


tools/cli/admin.go
```diff
@@ -520,6 +524,42 @@ func newAdminElasticSearchCommands() []cli.Command {
...
+					Name:  FlagMuttleyDestinationWithAlias,
+					Usage: "Optional muttely destination to ElasticSearch cluster",
+				},
+				cli.StringFlag{
+					Name:  FlagIndex,
+					Usage: "ElasticSearch target index",
+				},
+				cli.StringFlag{
+					Name:  FlagInputFileWithAlias,
+					Usage: "Input file of indexer.Message in json format, separated by newline",
```
longquanzheng: Please explain more detail about the file format

tools/cli/adminElasticSearchCommands.go
```diff
@@ -202,6 +205,59 @@ func AdminIndex(c *cli.Context) {
...
+	ratelimiter := tokenbucket.New(rps, clock.NewRealTimeSource())
+
+	file, err := os.Open(inputFileName)
+	if err != nil {
+		ErrorAndExit("Cannot open input file", nil)
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
+
+	scanner.Scan() // skip first line
```
longquanzheng: what is the first line and why it has to skip?

tools/cli/admin.go
vancexu: addressed

tools/cli/adminElasticSearchCommands.go
vancexu: added in helper message. First line is list result table header, see top comments for use example

PR #2429, opened 2019-08-22
Make archival config more dynamic

yycptt:
https://github.com/uber/cadence/issues/2397


common/archiver/archivalMetadata.go
```diff
@@ -148,34 +123,53 @@ func NewArchivalConfig(
 	}
 
 	return &archivalConfig{
-		clusterStatus:       clusterStatus,
-		enableRead:          enableRead,
-		domainDefaultStatus: domainDefaultStatus,
-		domainDefaultURI:    domainDefaultURI,
+		staticClusterStatus:  staticClusterStatus,
+		dynamicClusterStatus: dc.GetStringProperty(dynamicconfig.HistoryArchivalStatus, staticClusterStatusStr),
```
andrewjdawson2016: NewArchivalConfig seems like it is meant to work for both visibility and history but as its written this won't work for both?

common/archiver/archivalMetadata.go
```diff
@@ -148,34 +123,53 @@ func NewArchivalConfig(
...
+	// even dynamic config says archival is enabled, we should ignore that.
+	// Only when archival is enabled in static config, should we check if there's any difference between static config and dynamic config.
+	if a.staticClusterStatus != ArchivalEnabled {
+		return a.staticClusterStatus
+	}
+
+	dynamicStatusStr := a.dynamicClusterStatus()
+	dynamicStatus, err := getClusterArchivalStatus(dynamicStatusStr)
+	if err != nil {
+		dynamicStatus = a.staticClusterStatus
```
andrewjdawson2016: I think that if we ever fail to read dynamic config we should always default to disabled state even if archival config is enabled. Thoughts?

common/archiver/archivalMetadata.go
yycptt: Makes sense. I've updated the function to take dynamic function as parameter.

common/archiver/archivalMetadata.go
yycptt: I can argue both ways and I don't have a strong opinion on this. so changed to disabled when dynamic config is invalid.

PR #2432, opened 2019-08-22
CLI: add scan command

vancexu:


tools/cli/workflow.go
```diff
@@ -148,6 +148,16 @@ func newWorkflowCommands() []cli.Command {
 				ListAllWorkflow(c)
 			},
 		},
+		{
+			Name:    "scan",
+			Aliases: []string{"sc", "scanall"},
+			Usage: "scan workflow executions (only work with ElasticSearch). " +
```
longquanzheng: nit: only works if Cadence server enabled visibility with ElasticSearch

tools/cli/flags.go
```diff
@@ -406,6 +406,43 @@ func getFlagsForListAll() []cli.Flag {
 	}
 }
 
+func getFlagsForScan() []cli.Flag {
+	return []cli.Flag{
+		cli.IntFlag{
+			Name:  FlagPageSizeWithAlias,
```
longquanzheng: you can set the default value here so that the --help will show the default value

tools/cli/flags.go
```diff
@@ -406,6 +406,43 @@ func getFlagsForListAll() []cli.Flag {
...
+			Usage: "For tune page size",
```
longquanzheng: maybe "page size of result"

tools/cli/app.go
```diff
@@ -27,7 +27,7 @@ import (
 const (
 	// Version is the controlled version string. It should be updated every time
 	// before we release a new version.
-	Version = "0.8.1"
+	Version = "0.8.2"
```
yycptt: I think we have 0.8.3.

vancexu: Only for CLI. Compatible with server 0.8.*

Issue #2393, opened 2019-08-15
Repliator queue for metadata


meiliang86:
This can be used for domain replication and other stuffs like config later.

PR #2440, opened 2019-08-23
Cadence queue for domain replication

meiliang86:


common/persistence/cassandra/cassandraQueue.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/uber/cadence/common/backoff"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/common/service/config"
+)
+
+const (
+	templateEnqueueMessageQuery   = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
+	templateGetLastMessageIDQuery = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
+	templateGetMessagesQuery      = `SELECT message_id, message_payload ` +
```
longquanzheng: I think we need to add "Limit ?" otherwise it will always return all records.

common/persistence/dataInterfaces.go
```diff
@@ -22,6 +22,7 @@ package persistence
 
 import (
 	"fmt"
+	"github.com/uber/cadence/.gen/go/replicator"
```
longquanzheng: nit: fix the format before landing

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -0,0 +1,200 @@
...
+	return &cassandraQueue{
+		cassandraStore: cassandraStore{session: session, logger: logger},
+		logger:         logger,
+		queueType:      queueType,
+		retryPolicy:    retryPolicy,
+	}, nil
+}
+
+func (q *cassandraQueue) EnqueueMessage(messagePayload []byte) error {
+	err := backoff.Retry(
```
longquanzheng: I don't think we need backoff retry policy here. Persistence API are expected to be very low latency, so that it's much easier for us to troubleshoot with Cassandra. the retry should be in application layer. So the number requests sent to Cassandra should be constant and deterministic.

common/persistence/cassandra/cassandraQueue.go
meiliang86: fixed.

common/persistence/dataInterfaces.go
meiliang86: done

common/persistence/sql/sqlQueue.go
```diff
@@ -0,0 +1,113 @@
...
+		queueType:   queueType,
+		logger:      logger,
+		retryPolicy: retryPolicy,
+	}, nil
+
+}
+
+func (q *sqlQueue) EnqueueMessage(messagePayload []byte) error {
+	err := backoff.Retry(func() error {
+		return q.tryEnqueue(messagePayload)
```
longquanzheng: I think best way to do enque in MySQL will be different. We probably should use write lock to avoid retrying

common/queue/interface.go
```diff
@@ -0,0 +1,29 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package queue
+
+type (
+	// Queue is a general queue interface
+	Queue interface {
+		Publish(msgs interface{}) error
```
wxing1292: only publish? for general purpose queue?

wxing1292: also this interface (queue should be within collection package)

service/worker/replicator/domainReplicationMessageProcessor.go
```diff
@@ -0,0 +1,153 @@
...
+
+// TODO: need to make sure only one worker is processing per source DC
+// TODO: store checkpoints in DB
+func (p *domainReplicationMessageProcessor) processorLoop() {
+	timer := time.NewTimer(getWaitDuration())
+
+	for {
+		select {
+		case <-timer.C:
+			ctx, cancel := context.WithTimeout(context.Background(), fetchTaskRequestTimeout)
```
wxing1292: can you put these logic in a function?

common/persistence/persistence-factory/factory.go
```diff
@@ -79,6 +81,7 @@ type (
 		NewExecutionStore(shardID int) (p.ExecutionStore, error)
 		// NewVisibilityStore returns a new visibility store
 		NewVisibilityStore() (p.VisibilityStore, error)
+		NewQueue(queueType int) (p.Queue, error)
```
wxing1292: NewDomainReplicationQueue?

common/persistence/persistenceInterface.go
```diff
@@ -142,6 +142,18 @@ type (
 		CountWorkflowExecutions(request *CountWorkflowExecutionsRequest) (*CountWorkflowExecutionsResponse, error)
 	}
 
+	// Queue is a store to enqueue and get messages
+	Queue interface {
+		EnqueueMessage(messagePayload []byte) error
+		GetMessages(lastMessageID int, maxCount int) ([]*QueueMessage, error)
```
wxing1292: EnqueueMessage & DequeueMessages?

common/queue/interface.go
meiliang86: Seems like this is not used. I'm removing it.

common/persistence/persistence-factory/factory.go
meiliang86: This is a general queue which takes a queue type parameter.

common/persistence/persistenceInterface.go
meiliang86: done

service/worker/replicator/domainReplicationMessageProcessor.go
meiliang86: done

common/persistence/cassandra/cassandraQueue.go
meiliang86: ID conflict error is implementation detail that should be hide inside queue implementation. Caller should not try to parse and understand this error.

common/persistence/sql/sqlQueue.go
meiliang86: done

common/persistence/cassandra/cassandraQueue.go
wxing1292: I agree with @longquanzheng persistence layer should only handle converting to / from DB
no business logic, or retry logic should be put here

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -0,0 +1,190 @@
...
+
+	return &cassandraQueue{
+		cassandraStore: cassandraStore{session: session, logger: logger},
+		logger:         logger,
+		queueType:      queueType,
+		retryPolicy:    retryPolicy,
+	}, nil
+}
+
+func (q *cassandraQueue) EnqueueMessage(messagePayload []byte) error {
```
wxing1292: could you plz format all func like this:
```go
EnqueueMessage(
    messagePayload []byte,
) error
```

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (q *cassandraQueue) getNextMessageID() (int, error) {
+	query := q.session.Query(templateGetLastMessageIDQuery, q.queueType)
+
+	result := make(map[string]interface{})
+	err := query.MapScan(result)
+	if err != nil {
+		if err == gocql.ErrNotFound {
+			return 0, nil
```
wxing1292: i guess this should return entity not exists

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -0,0 +1,190 @@
...
+	if iter == nil {
+		return nil, &workflow.InternalServiceError{
+			Message: "DequeueMessages operation failed. Not able to create query iterator.",
+		}
+	}
+
+	var result []*persistence.QueueMessage
+	message := make(map[string]interface{})
+	for iter.MapScan(message) {
+		payload := message["message_payload"].([]byte)
```
wxing1292: can you create a dedicated function which does the conversion?

common/persistence/sql/sqlQueue.go
```diff
@@ -0,0 +1,86 @@
...
+		lastMessageID, err := tx.GetLastEnqueuedMessageIDForUpdate(q.queueType)
+		if err != nil {
+			if err == sql.ErrNoRows {
+				lastMessageID = -1
+			} else {
+				return fmt.Errorf("failed to get last enqueued message id: %v", err)
+			}
+		}
+
+		_, err = tx.InsertIntoQueue(&sqldb.QueueRow{QueueType: q.queueType, MessageID: lastMessageID + 1, MessagePayload: messagePayload})
```
wxing1292: can we have dedicated function which does these conversion?

service/frontend/workflowHandler.go
```diff
@@ -3466,3 +3479,44 @@ type domainWrapper struct {
...
+
+	if request == nil {
+		return nil, wh.error(errRequestNotSet, scope)
+	}
+
+	if wh.domainReplicationQueue == nil {
+		return nil, wh.error(errors.New("domain replication queue not enabled for cluster"), scope)
+	}
+
+	lastMessageID := -1
```
wxing1292: could you plz make this a const?

meiliang86:
#2393

PR #2442, opened 2019-08-23
Add warn log when deleting duplicate history when WorkflowExecutionAl…

longquanzheng:
…readyStartedError


service/history/historyEngine.go
```diff
@@ -459,6 +459,14 @@ func (e *historyEngineImpl) StartWorkflowExecution(
 	if err != nil {
 		if t, ok := err.(*persistence.WorkflowExecutionAlreadyStartedError); ok {
 			if t.StartRequestID == *request.RequestId {
+				// This is the only code path that deleting history could cause history corruption(missing first batch).
+				// We will use this warn log to verify this issue: https://github.com/uber/cadence/issues/2441
+				e.logger.Warn("encounter WorkflowExecutionAlreadyStartedError, deleting duplicated history",
+					tag.WorkflowDomainName(domainEntry.GetInfo().Name),
```
wxing1292: plz add this log in deleteEvents function, as a defer
plz also log the branch token & tree ID

PR #2476, opened 2019-08-29
Admin tools for deleting task and restart shard

lihannan99:


tools/cli/adminCommands.go
```diff
@@ -362,6 +362,74 @@ func AdminGetShardID(c *cli.Context) {
...
 
+// AdminRemoveTask describes history host
+func AdminRemoveTask(c *cli.Context) {
+	adminClient := cFactory.ServerAdminClient(c)
+
+	sid := c.Int(FlagShardID)
+	taskID := c.Int64(FlagRemoveTaskID)
+	typeID := c.Int(FlagRemoveTypeID)
+	fmt.Println("adminRemoveTask", sid, taskID, typeID)
+	if !c.IsSet(FlagShardID) {
```
vancexu: please use `getRequiredOption` to avoid such check

tools/cli/adminCommands.go
```diff
@@ -362,6 +362,74 @@ func AdminGetShardID(c *cli.Context) {
...
+		ErrorAndExit("type_id id is needed", nil)
+		return
+	}
+
+	ctx, cancel := newContext(c)
+	defer cancel()
+
+	req := &shared.RemoveTaskRequest{}
+
+	if c.IsSet(FlagShardID) {
```
vancexu: with required flag, you also don't need if check when create request below

tools/cli/adminCommands.go
```diff
@@ -362,6 +362,74 @@ func AdminGetShardID(c *cli.Context) {
...
+	}
+}
+
+// AdminShardManagement describes history host
+func AdminShardManagement(c *cli.Context) {
+	adminClient := cFactory.ServerAdminClient(c)
+
+	sid := c.Int(FlagShardID)
+
+	if !c.IsSet(FlagShardID) {
```
vancexu: ditto

PR #2482, opened 2019-08-29
Query API

andrewjdawson2016:
This adds a query API to history engine and plumbs it through frontend. This API will not be used anywhere yet because no client supports consistent query yet. After this diff lands there will be four more diffs before the server side for consistent query is done:
1. Handle RespondDecisionTaskCompleted and RespondDecisionTask failed to update query state machine.
2. Dispatch queries to matching on startDecisionTask.
3. Create a decision task to dispatch query if there is not already one outstanding.
4. Integration tests around consistent query.


client/history/client.go
```diff
@@ -645,6 +645,30 @@ func (c *clientImpl) SyncActivity(
 	return err
 }
 
+func (c *clientImpl) QueryWorkflow(
+	ctx context.Context,
+	request *h.QueryWorkflowRequest,
+	opts ...yarpc.CallOption) (*h.QueryWorkflowResponse, error) {
```
shreyassrivatsan: nit:
```
opts ...yarpc.CallOption,
) (*h.QueryWorkflowResponse, error) {
```

idl/github.com/uber/cadence/history.thrift
```diff
@@ -705,4 +715,18 @@ service HistoryService {
...
+  * QueryWorkflow returns query result for a specified workflow execution
+  **/
+  QueryWorkflowResponse QueryWorkflow(1: QueryWorkflowRequest queryRequest)
+	throws (
+	  1: shared.BadRequestError badRequestError,
+	  2: shared.InternalServiceError internalServiceError,
+	  3: shared.EntityNotExistsError entityNotExistError,
+	  4: shared.QueryFailedError queryFailedError,
+	  5: shared.LimitExceededError limitExceededError,
+      6: shared.ServiceBusyError serviceBusyError,
```
shreyassrivatsan: nit: spacing seems messed up here.

service/frontend/workflowHandler.go
```diff
@@ -2758,11 +2753,40 @@ func (wh *WorkflowHandler) QueryWorkflow(
 		response.GetClientFeatureVersion(),
 		response.GetClientImpl(),
 	)
-
 	queryRequest.Execution.RunId = response.Execution.RunId
-	if len(response.StickyTaskList.GetName()) != 0 && clientFeature.SupportStickyQuery() {
-		matchingRequest.TaskList = response.StickyTaskList
-		stickyDecisionTimeout := response.GetStickyTaskListScheduleToStartTimeout()
+	if clientFeature.SupportConsistentQuery() {
```
shreyassrivatsan: so currently this path is not exercised right?

service/history/handler.go
```diff
@@ -993,6 +993,42 @@ func (h *Handler) ResetWorkflowExecution(ctx context.Context,
...
+	if domainID == "" {
+		return nil, h.error(errDomainNotSet, scope, domainID, "")
+	}
+
+	if ok := h.rateLimiter.Allow(); !ok {
+		return nil, h.error(errHistoryHostThrottle, scope, domainID, "")
+	}
+
+	workflowID := request.GetExecution().GetWorkflowId()
+	engine, err1 := h.controller.GetEngine(workflowID)
```
shreyassrivatsan: these can just be `err` do not need `err1` and `err2`

service/history/historyEngine.go
```diff
@@ -578,6 +580,48 @@ func (e *historyEngineImpl) GetMutableState(
 	return response, nil
 }
 
+func (e *historyEngineImpl) QueryWorkflow(
+	ctx ctx.Context,
+	request *h.QueryWorkflowRequest,
+) (*h.QueryWorkflowResponse, error) {
+	context, release, retError := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetExecution())
```
shreyassrivatsan: can just use err here. do not really need `retError` and `err`, as you are returning `retError` on failure.

service/frontend/workflowHandler.go
andrewjdawson2016: Correct

service/history/handler.go
andrewjdawson2016: Yes I agree but we use this pattern throughout all other methods in this file.

shreyassrivatsan: If something is an unnecessary pattern we should change it - so over time the codebase gets cleaned up. I normally like changing the file I am working on - so at least future uses will not use the pattern.. 

In general using 2 error variables can get misleading, unless necessary - especially if they are not scoped like the below. You have 2 valid error variable and people may miss what to return and work with.. 

```
if customErr := op1; customErr != nil {
```

PR #2485, opened 2019-08-29
update cron integration test

yux0:
This is a fix for cron integration test. As we are not using one time source for workflow timestamps. The start time and close time can be different in seconds. This change is to ignore the diff between start time and close time in a cron workflow.


host/integration_test.go
```diff
@@ -1018,8 +1018,10 @@ func (s *integrationSuite) TestCronWorkflow() {
 		executionInfo := closedExecutions[i]
 		// Roundup to compare on the precision of seconds
 		expectedBackoff := executionInfo.GetExecutionTime()/1000000000 - lastExecution.GetExecutionTime()/1000000000
+		// We are using different time source and the close time and start time can be different
+		executionTimeDiff := executionInfo.GetStartTime()/1000000000 - lastExecution.GetCloseTime()/1000000000
```
wxing1292: plz 
1. add a comment here explaining how the execution time is calaulated
2. cut an issue to remove this patch once unified time source is implemented

yux0: updated

PR #2486, opened 2019-08-29
format tests

yux0:
Move integration test framework to common and do code formatting.


common/event_generator.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
```
wxing1292: plz move to common/test dir

PR #2491, opened 2019-08-30
Integrate Visibility Archival into the system

yycptt:
Task 3 of https://github.com/uber/cadence/issues/2452

- Archive visibility record when workflow closes
- New Frontend API for querying archived records
- Integration test for visibility archival

Advanced query API for archived visibility records will be added later.


idl/github.com/uber/cadence/shared.thrift
```diff
@@ -911,6 +911,7 @@ struct History {
 
 struct WorkflowExecutionFilter {
   10: optional string workflowId
+  20: optional string runId
```
andrewjdawson2016: @vancexu is it correct to add this to this struct?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1312,6 +1313,21 @@ struct ListWorkflowExecutionsResponse {
   20: optional binary nextPageToken
 }
 
+struct ListArchivedWorkflowExecutionsRequest {
+  10: optional string domain
+  20: optional i32 maximumPageSize
+  30: optional binary nextPageToken
+  40: optional StartTimeFilter CloseTimeFilter
```
andrewjdawson2016: naming?

service/frontend/workflowHandler.go
```diff
@@ -2317,6 +2317,118 @@ func (wh *WorkflowHandler) ListOpenWorkflowExecutions(
...
+	if !wh.GetArchivalMetadata().GetVisibilityConfig().ClusterConfiguredForArchival() {
+		return nil, wh.error(&gen.BadRequestError{Message: "Cluster is not configured for visibility archival"}, scope)
+	}
+
+	entry, err := wh.domainCache.GetDomain(listRequest.GetDomain())
+	if err != nil {
+		return nil, wh.error(err, scope)
+	}
+
+	if entry.GetConfig().VisibilityArchivalStatus != shared.ArchivalStatusEnabled {
```
andrewjdawson2016: Check on read as well?

service/frontend/workflowHandler.go
```diff
@@ -2317,6 +2317,118 @@ func (wh *WorkflowHandler) ListOpenWorkflowExecutions(
 	return resp, nil
 }
 
+// ListArchivedWorkflowExecutions - retrieves archived info for closed workflow executions in a domain
+func (wh *WorkflowHandler) ListArchivedWorkflowExecutions(
+	ctx context.Context,
+	listRequest *gen.ListArchivedWorkflowExecutionsRequest,
```
andrewjdawson2016: We talked about this supporting a simplified SQL syntax?

service/history/transferQueueProcessorBase.go
```diff
@@ -300,7 +302,39 @@ func (t *transferQueueProcessorBase) recordWorkflowClosed(
...
+		if err != nil {
+			return err
+		}
+		visibilityArchiver, err := t.shard.GetService().GetArchiverProvider().GetVisibilityArchiver(archiveURI.Scheme(), common.HistoryServiceName)
+		if err != nil {
+			return err
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), t.shard.GetConfig().TransferProcessorVisibilityArchivalTimeLimit())
+		defer cancel()
+		return visibilityArchiver.Archive(ctx, archiveURI, &archiver.ArchiveVisibilityRequest{
```
andrewjdawson2016: Emit metric on failure

andrewjdawson2016: We want to be able to compute a success graph like we do for histories. Make sure we have all the metrics we need to do this.

idl/github.com/uber/cadence/shared.thrift
yycptt: I have talked to @vancexu about this before. He's ok with this change.

idl/github.com/uber/cadence/shared.thrift
yycptt: The actual filtering is done on closeTime. But the struct type has a bad naming. I can follow the existing patten (name the variable as `StartTimeFilter`, but internally filtering on close time), but it seems kind of weird to me. Or adding a new struct for closeTime?

service/frontend/workflowHandler.go
yycptt: I know. I will do that in another pr later as it will contain logic for parsing, translating, validating the SQL query (since it applies to all visibility archivers). I think for the first version of visibility archival it's ok not having SQL syntax. 

Or I can add a new API (or new field to the request struct) for the SQL syntax and change the visibility archiver interface correspondingly so that each archiver can enforce different rules on the SQL syntax. Thoughts?

idl/github.com/uber/cadence/shared.thrift
vancexu: I think it is ok to have RunID as optional filter.

idl/github.com/uber/cadence/shared.thrift
vancexu: For clearness, how about creating TimeFilter (instead of using previous wrong StartTimeFilter) since this is new API doesn't have compatibility issue)
Also property name should be lower case `closeTimeFilter`

service/frontend/workflowHandler.go
```diff
@@ -2317,6 +2317,118 @@ func (wh *WorkflowHandler) ListOpenWorkflowExecutions(
...
+
+	if err := wh.versionChecker.checkClientVersion(ctx); err != nil {
+		return nil, wh.error(err, scope)
+	}
+
+	if listRequest == nil {
+		return nil, wh.error(errRequestNotSet, scope)
+	}
+
+	if ok := wh.allow(listRequest); !ok {
```
vancexu: what is this `allow`? can we rename to be more specific?

service/frontend/workflowHandler.go
```diff
@@ -2317,6 +2317,118 @@ func (wh *WorkflowHandler) ListOpenWorkflowExecutions(
...
+
+	if listRequest.CloseTimeFilter.GetEarliestTime() > listRequest.CloseTimeFilter.GetLatestTime() {
+		return nil, wh.error(&gen.BadRequestError{Message: "EarliestTime in CloseTimeFilter should not be larger than LatestTime"}, scope)
+	}
+
+	if listRequest.GetMaximumPageSize() <= 0 {
+		listRequest.MaximumPageSize = common.Int32Ptr(int32(wh.config.VisibilityMaxPageSize(listRequest.GetDomain())))
+	}
+
+	if wh.isListRequestPageSizeTooLarge(listRequest.GetMaximumPageSize(), listRequest.GetDomain()) {
```
vancexu: Why archival vis need such check? probably just remove.

service/frontend/workflowHandler.go
yycptt: Makes sense. Updated.

service/frontend/workflowHandler.go
yycptt: Updated frontend and visibility archiver interface to accept sql string. Created a query parser interface for filestore. Will implement this interface later.

service/frontend/workflowHandler.go
yycptt: Removed.

service/frontend/workflowHandler.go
yycptt: This is for rate limiting and this function is used throughout the entire file. I can create another pr to address this issue.

service/history/transferQueueProcessorBase.go
yycptt: Emitted metrics when visibility archive is attempted or failed. For latency metrics, I think they should be emitted by the archiver itself.

PR #2521, opened 2019-09-06
Wire NDC implementation end to end

wxing1292:
* Refactor replication queue, add handling of NDC replication task
* Fix issue that branch token should be set before task refreshment during mutable state rebuild
* Fix dead lock issue when performing NDC workflow reset
* Reorder task ID generation in ConflictResolveWorkflowExecution, ResetWorkflowExecution, making sure current workflow tasks are handled before new workflow tasks
* Add NDC handling for continue as new workflow in state builder, history engine, workflow resetter
* Fix cassandra schema


service/history/shardContext.go
```diff
@@ -698,16 +698,6 @@ func (s *shardContextImpl) ConflictResolveWorkflowExecution(request *persistence
 	// assign IDs for the transfer/replication tasks
```
yux0: Also move the comments?

service/history/historyEngine.go
```diff
@@ -345,38 +345,51 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
...
+) (mutableState, error) {
 
 	domainName := domainEntry.GetInfo().Name
-	if e.config.EnableEventsV2(domainName) && e.config.EnableNDC(domainName) {
-		return newMutableStateBuilderWithVersionHistories(
+	enableEventsV2 := e.config.EnableEventsV2(domainName)
+	enableNDC := e.config.EnableNDC(domainName)
+
+	var newMutableState mutableState
+	if enableEventsV2 && enableNDC {
```
yux0: So if event v2 is disable, it won't create mutable state with version histories?

service/history/replicatorQueueProcessor.go
```diff
@@ -639,3 +548,323 @@ func (p *replicatorQueueProcessorImpl) getTasks(readLevel int64) (*replicator.Re
...
+}
+
+func (p *replicatorQueueProcessorImpl) getEventsBlob(
+	branchToken []byte,
+	firstEventID int64,
+	nextEventID int64,
+) (*shared.DataBlob, error) {
+
+	var eventBatchBlobs []*persistence.DataBlob
+	var pageToken []byte
```
yux0: no need to declare the pageToken here

service/history/historyEngine.go
wxing1292: yes, how does it work without events v2?

PR #2539, opened 2019-09-13
Add compatibility support for MySQL "tx_isolation" in docker config

barakalon:
This replaces this PR: https://github.com/uber/cadence/pull/2493

Here’s some context: https://forum.pydio.com/t/solved-mysql8-error-when-running-pydio-cells-docker-failed-to-initialize-storage-failed-to-open-migrations-creating-dex-migration-table-error-1193-unknown-system-variable-tx-isolation/2129

> the problem are are changes made in MySQL 5.7 and 8 - ‘tx_isolation’ has been depreciated in 5.7.20 and an alias has been added called ’transaction_isolation. In MySQL 8 tx_isolation has been removed entirely leaving the former alias.
>
> The means that cells will not run with MySQL 8 currently. Code changes are needed to allow both, pre 5.7.20 (tx_isolation) and 5.7.20 and later (transaction_isolation). It should however not be a big task I guess, as this variable is rarely used. An option would be by to use the “set transaction” statement instead of setting the isolation level as connection parameter. “set transaction” is supported in all versions and will not cause compatibility issues - it will, however, add a slight overhead due to the additional statement needed.

This change isn't ideal, but it quickly enables folks (like myself) who are trying to deploy Cadence via Docker to use an existing Aurora MySQL instance, which still uses the `tx_isolation` parameter.


barakalon:
I'm unable to view the buildkite failures.
Can anyone provide more information on what went wrong?

longquanzheng:
> I'm unable to view the buildkite failures.
> Can anyone provide more information on what went wrong?

Sure, I am looking at it

PR #2545, opened 2019-09-16
Factor out all methods on decision state machine

andrewjdawson2016:


service/history/mutableStateBuilder.go
```diff
@@ -170,12 +170,8 @@ func newMutableStateBuilder(
...
 	}
 	s.hBuilder = newHistoryBuilder(s, logger)
-
-	s.taskGenerator = newMutableStateTaskGenerator(
-		shard.GetDomainCache(),
-		s.logger,
-		s,
-	)
+	s.taskGenerator = newMutableStateTaskGenerator(shard.GetDomainCache(), s.logger, s)
+	s.dtm = newMutableStateDecisionTaskManager(s)
```
wxing1292: plz change `dtm` to something meaningful & clear

service/history/mutableStateBuilder.go
```diff
@@ -1110,15 +1106,8 @@ func (e *mutableStateBuilder) getDecisionInfo() *decisionInfo {
...
-func (e *mutableStateBuilder) GetDecisionInfo(
-	scheduleEventID int64,
-) (*decisionInfo, bool) {
-
-	decision := e.getDecisionInfo()
-	if scheduleEventID == decision.ScheduleID {
-		return decision, true
-	}
-	return nil, false
+func (e *mutableStateBuilder) GetDecisionInfo(scheduleEventID int64) (*decisionInfo, bool) {
```
wxing1292: format like this plz, this will make the input vars more clear
```
GetDecisionInfo(
	scheduleEventID int64,
) (*decisionInfo, bool)
```

service/history/mutableStateBuilder.go
```diff
@@ -1973,120 +1619,23 @@ func (e *mutableStateBuilder) AddDecisionTaskCompletedEvent(
...
-	return nil
+func (e *mutableStateBuilder) ReplicateDecisionTaskTimedOutEvent(timeoutType workflow.TimeoutType) error {
+	return e.dtm.ReplicateDecisionTaskTimedOutEvent(timeoutType)
 }
 
-func (e *mutableStateBuilder) AddDecisionTaskScheduleToStartTimeoutEvent(
-	scheduleEventID int64,
-) (*workflow.HistoryEvent, error) {
-
-	opTag := tag.WorkflowActionDecisionTaskTimedOut
```
wxing1292: i believe the following checks should be kept in mutable state
```go
opTag := tag.WorkflowActionDecisionTaskTimedOut
if err := e.checkMutability(opTag); err != nil {
	return nil, err
}
```

wxing1292: same applies to all places

service/history/mutableStateDecisionTaskManager.go
```diff
@@ -0,0 +1,726 @@
...
+	workflow "github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/errors"
+	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/persistence"
+)
+
+type (
+	mutableStateDecisionTaskManager interface {
+		ReplicateDecisionTaskScheduledEvent(
```
wxing1292: could arrange the methods like
```
AddDecisionTaskScheduledEvent
ReplicateDecisionTaskScheduledEvent

AddDecisionTaskStartedEvent
ReplicateDecisionTaskStartedEvent
...
```

service/history/mutableStateDecisionTaskManager_mock.go
```diff
@@ -0,0 +1,402 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: 2019

service/history/mutableStateDecisionTaskManager.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+)
+
+func newMutableStateDecisionTaskManager(msb *mutableStateBuilder) mutableStateDecisionTaskManager {
+	return &mutableStateDecisionTaskManagerImpl{
+		msb: msb,
+	}
+}
+
+func (m *mutableStateDecisionTaskManagerImpl) ReplicateDecisionTaskScheduledEvent(
```
wxing1292: could you plz arrange the functions in this way?

1. decision schedule functions, i.e. add then replicate
2. decision start functions
3. decision completed functions
4. decision timeout / fail, etc

service/history/mutableStateDecisionTaskManager.go
```diff
@@ -0,0 +1,726 @@
...
+	"math"
+	"time"
+
+	workflow "github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/errors"
+	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/persistence"
+)
+
```
wxing1292: plz also add a line here
```
//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination mutableStateDecisionTaskManager_mock.go
```

wxing1292:
double check, there is not logic change, just moving of existing logic?

service/history/mutableStateDecisionTaskManager.go
wxing1292: and one more thing, `mutableStateDecisionTaskManager` is a long name, what about `decisionMgr` ? or `decisionStateMachine`

service/history/mutableStateDecisionTaskManager.go
andrewjdawson2016: There is not a clean mapping between the two for example ```AddDecisionTaskScheduledEventAsHeartbeat``` has not equal replicate method. So I think this grouping will look confusing.

service/history/mutableStateBuilder.go
andrewjdawson2016: I like the property that all the logic of knowing how to update a decision task in accordance with the current state of the workflow lives in one file. Why do you prefer split this logic into two places?

service/history/mutableStateDecisionTaskManager.go
andrewjdawson2016: I reordered in this way but I kept the replicate methods separate.

service/history/mutableStateBuilder.go
wxing1292: the mutability check should exist in mutable state, which in general controls all logic
decision state machine should be the logic maintaining the decision part, nothing else

service/history/mutableStateDecisionTaskManager.go
wxing1292: AddDecisionTaskScheduledEventAsHeartbeat has no counter part, that is true,
but for readability, we should group related logics in on section.

andrewjdawson2016: The logics are grouped in related ways. We have all the Replicate methods together. We have all the Add*Events together etc...

service/history/mutableStateBuilder.go
andrewjdawson2016: I would agree if the file I was adding was a pure state machine. But as we spoke about offline on Thursday it is not this. Really all this file is a grouping of related methods from mutableStateBuilder. 

If we only want the logic of decision in this new file then we also need to remove things like: checking buffered events and writing to hBuilder. 

In short I think decisionInfo updates are too tightly duplicated with other parts of mutableStateBuilder to pull it out cleanly. So instead I think just all the methods related to decisions should be pulled out.

service/history/mutableStateDecisionTaskManager.go
andrewjdawson2016: Resolved offline

Issue #2420, opened 2019-08-20
Allowing reset workflows without decision started event


longquanzheng:
In case of workflow timeout, there is no decision started event, current reset doesn't allow to reset

PR #2560, opened 2019-09-17
Implement of reapply logic 

yux0:


service/history/nDCEventsReapplication.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/cache"
+	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/cluster"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/metrics"
+)
+
+type (
+	nDCEventsReapplication struct {
```
wxing1292: plz make this a interface
suggestion: call the interface nDCEventsReapplier
suggestion: call the implementation nDCEventsReapplierImpl

wxing1292: also add the mock by this
```go
//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination nDCEventsReapplier_mock.go
```

service/history/nDCEventsReapplication.go
```diff
@@ -0,0 +1,262 @@
...
+		resetor:         resetor,
+	}
+}
+func (r *nDCEventsReapplication) reapplyEvents(
+	ctx ctx.Context,
+	domainID string,
+	workflowID string,
+	historyEvents []*shared.HistoryEvent,
+) (retError error) {
+	reapplyEvents := []*shared.HistoryEvent{}
```
wxing1292: plz add a TODO here, stating that there should be a reapply policy in the future

service/history/nDCEventsReapplication.go
```diff
@@ -0,0 +1,262 @@
...
+		switch event.GetEventType() {
+		case shared.EventTypeWorkflowExecutionSignaled:
+			reapplyEvents = append(reapplyEvents, event)
+		}
+	}
+
+	if len(reapplyEvents) == 0 {
+		return nil
+	}
+	context, msBuilder, releaseFunc, err := r.getCurrentWorkflowMutableState(ctx, domainID, workflowID)
```
wxing1292: I would suggest for here, only operate on the mutable state passed in (as a parameter), let outside handle the context.updateWorkflow

service/history/nDCEventsReapplication.go
```diff
@@ -0,0 +1,262 @@
...
+			return ErrUnreappliableEvent
+		}
+	}
+
+	r.logger.Info(fmt.Sprintf("reapplying %v signals", numSignals))
+	now := r.timeSource.Now()
+	return context.updateWorkflowExecutionAsActive(now)
+}
+
+func (r *nDCEventsReapplication) reapplyEventsToCurrentClosedWorkflow(
```
wxing1292: Please leave out the reapply to closed workflow for now, we either need to rewrite the existing code, supporting #2420 or introduce this back when reapply policy is introduced

service/history/nDCEventsReapplication.go
```diff
@@ -0,0 +1,262 @@
...
+	msBuilder, err := context.loadWorkflowExecution()
+	if err != nil {
+		// no matter what error happen, we need to retry
+		release(err)
+		return nil, nil, nil, err
+	}
+	return context, msBuilder, release, nil
+}
+
+func (r *nDCEventsReapplication) prepareWorkflowMutation(
```
wxing1292: when we do reapply, we should treat the current cluster as active (close transaction function in mutable state will do domain activeness check). so this logic is not needed

this function was used in 2DC since the overall reapply logic is different

service/history/nDCEventsReapplication.go
yux0: Renamed this file

service/history/nDCEventsReapplication.go
yux0: done

service/history/nDCEventsReapplication.go
yux0: done

service/history/nDCEventsReapplication.go
yux0: ignore the closed workflow for now. Added a todo here

service/history/nDCEventsReapplication.go
yux0: removed this method

PR #2562, opened 2019-09-18
Add protection on SQL injection for visibility

vancexu:
Currently SQL injection will panic instead of throwing error.

This PR fixes these.

### DIFF
--- a/common/elasticsearch/validator/queryValidator.go
+++ b/common/elasticsearch/validator/queryValidator.go
@@ -22,6 +22,7 @@ package validator
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	workflow "github.com/uber/cadence/.gen/go/shared"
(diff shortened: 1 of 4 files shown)


common/elasticsearch/validator/queryValidator.go
```diff
@@ -77,17 +78,20 @@ func (qv *VisibilityQueryValidator) validateListOrCountRequestForQuery(whereClau
 		whereClause := strings.TrimSpace(whereClause)
 		// #nosec
 		if common.IsJustOrderByClause(whereClause) { // just order by
-			placeholderQuery = "SELECT * FROM dummy " + whereClause
+			placeholderQuery = fmt.Sprintf("SELECT * FROM dummy %s", whereClause)
```
yycptt: What's the difference compare to the old one?

vancexu: not much difference but might be more proper to some security tool.

PR #2570, opened 2019-09-19
Add buffered events version change protection

wxing1292:
* When mutable state is initialized, a copy of domain cache entry is passed in, instead of domain name for ease of use
* Update some tests, use mocked domain cache entry, instead of mock metadata manager
* Add additional return parameter: transaction policy for nDCWorkflow suppressWorkflowBy API
* Allow current workflow to be set as zombie during conflicting resolution
* Fix issue that NDC branch manager should flush buffer as active
* When mutable state is initialized, buffered events version checking will be triggered, handing 4 cases of version changes, decision will be failed and persisted to DB before passing mutable state to caller.
   * active -> passive => fail decision & flush buffer using last write version
   * active -> active => fail decision & flush buffer using last write version
   * passive -> active => fail decision using current version, no buffered events
   * passive -> passive => no buffered events, since always passive, nothing to be done

### DIFF
--- a/common/persistence/operationModeValidator.go
+++ b/common/persistence/operationModeValidator.go
@@ -135,6 +135,7 @@ func ValidateConflictResolveWorkflowModeState(
 	switch mode {
 	case ConflictResolveWorkflowModeUpdateCurrent:
 		// it is ok that currentWorkflowMutation is null, for 2 DC
+		// Note: current workflow mutation can be in zombie state, for the update
 		if resetWorkflowState == WorkflowStateZombie ||
 			(newWorkflowState != nil && *newWorkflowState == WorkflowStateZombie) {
 			return &workflow.InternalServiceError{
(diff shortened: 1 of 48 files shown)


common/persistence/operationModeValidator.go
```diff
@@ -135,9 +135,9 @@ func ValidateConflictResolveWorkflowModeState(
 	switch mode {
 	case ConflictResolveWorkflowModeUpdateCurrent:
 		// it is ok that currentWorkflowMutation is null, for 2 DC
+		// Note: current workflow mutation can be in zombie state, for the update
 		if resetWorkflowState == WorkflowStateZombie ||
-			(newWorkflowState != nil && *newWorkflowState == WorkflowStateZombie) ||
-			(currentWorkflowState != nil && *currentWorkflowState == WorkflowStateZombie) {
```
yux0: Why current workflow will not be zombie?

wxing1292: this validation is checking whether the status to be written down to DB make sense.

current workflow, if present, can be terminated or zombiefied so target workflow and / or new workflow can be persisted.
since the policy is `ConflictResolveWorkflowModeUpdateCurrent` so the target and / or new workflow must be non zombie

wxing1292: the removal of check `(currentWorkflowState != nil && *currentWorkflowState == WorkflowStateZombie)` basically current workflow ending up being a zombie workflow. 
this should be allowed since current workflow may not be generated by this cluster

PR #2571, opened 2019-09-19
Fix archiver provider race condition

yycptt:


common/archiver/provider/provider.go
```diff
@@ -53,6 +54,8 @@ type (
 	}
 
 	archiverProvider struct {
+		m *sync.RWMutex
```
andrewjdawson2016: You can just make archiverProvider sync.RWMutex then you can do p.RLock and p.RUnlock

yycptt: Updated.

PR #2582, opened 2019-09-22
Handle Unhandled Errors

carlcortright:
Looking through the codebase there were several errors that were either unhandled when they should be handled or handled without empty assignment. This PR logs them in most cases and returns them in others.


cmd/server/cadence.go
```diff
@@ -21,6 +21,7 @@
 package main
 
 import (
+	"github.com/uber/cadence/common/log/tag"
```
longquanzheng: can you fix the import order?

cmd/server/cadence.go
```diff
@@ -36,7 +37,9 @@ var validServices = []string{historyService, matchingService, frontendService, w
 // main entry point for the cadence server
 func main() {
 	app := buildCLI()
-	app.Run(os.Args)
+	if err := app.Run(os.Args); err != nil {
+		log.Fatal("failed to run entry point for cadence server", tag.Error(err))
```
longquanzheng: This won't work. Our tagging only works with the special logger. I think you can just :
```
log.Fatal("failed to run entry point for cadence server, err:", err)
```

common/persistence/persistence-tests/metadataPersistenceV2Test.go
```diff
@@ -22,6 +22,7 @@ package persistencetests
 
 import (
 	"fmt"
+	"github.com/uber/cadence/common/log/tag"
```
longquanzheng: import order

common/persistence/sql/common.go
```diff
@@ -58,7 +62,9 @@ func (m *sqlStore) txExecute(operation string, f func(tx sqldb.Tx) error) error
 	}
 	err = f(tx)
 	if err != nil {
-		tx.Rollback()
+		if err := tx.Rollback(); err != nil {
+			err = errors.Wrap(err, "failed to rollback transaction")
```
longquanzheng: why do we need to wrap it?

service/history/workflowResetor.go
```diff
@@ -776,11 +778,19 @@ func (w *workflowResetorImpl) ApplyResetEvent(
 		return retError
 	}
 	defer func() {
-		w.eng.historyV2Mgr.CompleteForkBranch(&persistence.CompleteForkBranchRequest{
+		err := w.eng.historyV2Mgr.CompleteForkBranch(&persistence.CompleteForkBranchRequest{
```
longquanzheng: we actually don't want to return error even completeFork failed. There will be a background jog takes care of it. For this error let's simply log it.

service/matching/forwarder.go
```diff
@@ -226,8 +226,7 @@ func (fwdr *Forwarder) ForwardPoll(ctx context.Context) (*internalTask, error) {
 			ForwardedFrom: &fwdr.taskListID.name,
 		})
 		if err != nil {
-			fwdr.handleErr(err)
-			return nil, err
+			return nil, fwdr.handleErr(err)
```
longquanzheng: This is not equivalent. The handleErr did the transfomation of error on purpose, I don't think we should return errForwarderSlowDown to API.  I think we want to ignore the error returned from handlerErr.

service/worker/archiver/util.go
```diff
@@ -98,6 +99,11 @@ func errorDetails(err error) string {
 	if _, ok := err.(*cadence.CustomError); !ok {
 		return details
 	}
-	err.(*cadence.CustomError).Details(&details)
+	// If no details can be extracted the Details function will return with an err no data message that we want to
```
longquanzheng: @yycptt  @andrewjdawson2016  can anyone of you take a look at this?

service/worker/indexer/esProcessor.go
```diff
@@ -297,14 +301,14 @@ func newKafkaMessageWithMetrics(kafkaMsg messaging.Message, stopwatch *tally.Sto
...
-	km.message.Ack()
+	_ = km.message.Ack()
 	if km.swFromAddToAck != nil {
 		km.swFromAddToAck.Stop()
 	}
 }
 
 func (km *kafkaMessageWithMetrics) Nack() {
-	km.message.Nack()
+	_ = km.message.Nack()
```
longquanzheng: why some of the Nack/Act errors are logged but some not?

service/worker/archiver/util.go
yycptt: I think the idea there is that, if the error has some details and the details can be parsed in to string, then return the details as string, otherwise just return an empty string.
How about changing the function to the following:
```
func errorDetails(err error) string {
    var details string
    customErr, ok := err.(*cadence.CustomError)
    if !ok {
        return ""
    }
    if err := customErr.Details(&details); err != nil {
        return ""
    }
    return details
}
```

tools/common/schema/test/updatetest.go
```diff
@@ -63,32 +63,40 @@ func (tb *UpdateSchemaTestBase) SetupSuiteBase(db DB) {
 
 // TearDownSuiteBase tears down the test suite
 func (tb *UpdateSchemaTestBase) TearDownSuiteBase() {
-	tb.db.DropDatabase(tb.DBName)
+	// Test suite utility - ignore errors
```
longquanzheng: looks like the tests are failing:
```

FAIL: TestUpdateSchemaTestSuite (13.86s) | 7s
-- | --
  | require.go:765:
  | Error Trace:	updatetest.go:83
  | updateTask_test.go:63
  | Error:      	Expected nil, but got: gocql.errorFrame{frameHeader:gocql.frameHeader{version:0x84, flags:0x0, stream:1920, op:0x0, length:145, customPayload:map[string][]uint8(nil), warnings:[]string(nil)}, code:8704, message:"Cannot drop user type update_test_3381071236719446981.cluster_replication_config as it is still used by user type domain_replication_config"}
  | Test:       	TestUpdateSchemaTestSuite
  | --- FAIL: TestUpdateSchemaTestSuite/TestDryrun (13.57s)


```

common/persistence/sql/common.go
carlcortright: We don't just updated it to just swap the error.

service/worker/indexer/esProcessor.go
carlcortright: I was logging opportunistically. If there was easy access to a logger I'd log it, otherwise just explicitly ignored.

carlcortright:
@longquanzheng @samarabbas apologies for the delay on this one. Just iterated on your comments. Let me know if there's anything else we should update or if it's good to merge!

cmd/tools/cli/main.go
```diff
@@ -21,14 +21,16 @@
 package main
 
 import (
-	"os"
-
 	"github.com/uber/cadence/tools/cli"
+	"log"
+	"os"
```
yux0: Import ordering

yux0: "log"
"os"

"github.com/uber/cadence/tools/cli"

longquanzheng:
> @longquanzheng @samarabbas apologies for the delay on this one. Just iterated on your comments. Let me know if there's anything else we should update or if it's good to merge!

can you run "make fmt" and "make lint" to make sure it passes all the format requirements?

wxing1292:
Currently the build is blocked by missing 
```
# github.com/uber/cadence/service/history
service/history/timerBuilder.go:241:4: undefined: log
service/history/workflowResetor.go:26:2: imported and not used: "github.com/pkg/errors"
```

I will try to create a PR tomorrow to `carlcortright:cc/handle-errors`

wxing1292:
Or maybe this is easier, could you apply this patch?
[fix-build.txt](https://github.com/uber/cadence/files/3715547/fix-build.txt)

PR #2597, opened 2019-09-25
Introduce in memory decision task state machine

andrewjdawson2016:
This diff does the following

- Introduces methods on mutable state to handle in memory decision tasks
- Consume these in memory decision tasks for query
- Update query state machine and query registry to not include expired state
- Fix the two places I missed including the mutability check from an earlier diff


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/dgryski/go-farm v0.0.0-20190423205320-6a90982ecee2
 	github.com/eapache/go-resiliency v1.2.0 // indirect
 	github.com/emirpasic/gods v0.0.0-20190624094223-e689965507ab
+	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a // indirect
```
wxing1292: My guess, if you use go 1.13 this line will not be included

service/history/historyEngine.go
```diff
@@ -585,44 +585,98 @@ func (e *historyEngineImpl) GetMutableState(
...
+	}
+	ttl := e.shard.GetConfig().LongPollExpirationInterval(domainCache.GetInfo().Name)
+	timer := time.NewTimer(ttl)
+	defer timer.Stop()
+
 	context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetExecution())
 	if err != nil {
 		return nil, err
 	}
-	ms, err := context.loadWorkflowExecution()
```
wxing1292: plz add a `defer func() { release(retError) }()` here
I know there is a release fn call after error checking and `msBuilder.GetQueryRegistry()`
the defer is more of a best practice

service/history/historyEngine.go
```diff
@@ -585,44 +585,98 @@ func (e *historyEngineImpl) GetMutableState(
...
-		return nil, err
+
+	// after query is finished removed query from query registry and remove any potentially created in memory decision task
+	defer func() {
+		queryRegistry.removeQuery(queryID)
+		context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetExecution())
+		if err != nil {
+			retResp = nil
+			retErr = err
+		}
```
wxing1292: same here

service/history/historyEngine.go
```diff
@@ -585,44 +585,98 @@ func (e *historyEngineImpl) GetMutableState(
...
+		release(nil)
+	}()
+
+	// ensure decision task exists to dispatch query on
+retryLoop:
+	for i := 0; i < conditionalRetryCount; i++ {
+		context, release, err = e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetExecution())
+		if err != nil {
+			return nil, err
+		}
```
wxing1292: same here

service/history/mutableState.go
```diff
@@ -217,5 +217,12 @@ type (
 
 		CloseTransactionAsMutation(now time.Time, transactionPolicy transactionPolicy) (*persistence.WorkflowMutation, []*persistence.WorkflowEvents, error)
 		CloseTransactionAsSnapshot(now time.Time, transactionPolicy transactionPolicy) (*persistence.WorkflowSnapshot, []*persistence.WorkflowEvents, error)
+
+		AddInMemoryDecisionTaskScheduled(time.Duration) error
```
wxing1292: could you put these APIs at very top?

service/history/mutableStateBuilder.go
```diff
@@ -1505,12 +1505,20 @@ func (e *mutableStateBuilder) ReplicateWorkflowExecutionStartedEvent(
...
+	if err := e.checkMutability(opTag); err != nil {
+		return err
+	}
 	return e.decisionTaskManager.AddFirstDecisionTaskScheduled(startEvent)
 }
 
 func (e *mutableStateBuilder) AddDecisionTaskScheduledEvent(
 	bypassTaskGeneration bool,
 ) (*decisionInfo, error) {
+	opTag := tag.WorkflowActionDecisionTaskScheduled
```
wxing1292: nice

service/history/mutableStateBuilder.go
```diff
@@ -3569,6 +3577,30 @@ func (e *mutableStateBuilder) CloseTransactionAsSnapshot(
 	return workflowSnapshot, workflowEventsSeq, nil
 }
 
+func (e *mutableStateBuilder) AddInMemoryDecisionTaskScheduled(ttl time.Duration) error {
```
wxing1292: could you plz put all business APIs before close transaction function?

service/history/mutableStateDecisionTaskManager_mock.go
```diff
@@ -1,6 +1,6 @@
 // The MIT License (MIT)
 //
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
wxing1292: 2019

service/history/historyEngine.go
```diff
@@ -585,44 +585,98 @@ func (e *historyEngineImpl) GetMutableState(
 func (e *historyEngineImpl) QueryWorkflow(
 	ctx ctx.Context,
 	request *h.QueryWorkflowRequest,
-) (*h.QueryWorkflowResponse, error) {
+) (retResp *h.QueryWorkflowResponse, retErr error) {
+	domainCache, err := e.shard.GetDomainCache().GetDomainByID(request.GetDomainUUID())
```
wxing1292: i would suggest 
1. moving this logic into a dedicated file.
2. breaking the 3 logic (loading mutable state and do something) into 3 functions
3. it is ok to have a for loop for retry, but plz move the logic within the for loop into a function

go.mod
andrewjdawson2016: thanks :)

service/history/historyEngine.go
andrewjdawson2016: Oh thank you - I did it this way because I did not realize it was idempotent.

PR #2598, opened 2019-09-25
Improve MySQL db query index usage

longquanzheng:
I found that there is a way to get rid of the filtering "txn_id < ?" which can cause perf for MySQL query.


wxing1292:
Assuming that there is a million stale records and page size is 100.
the pagination token:
```go
	filter := &sqldb.HistoryNodeFilter{
		TreeID:    sqldb.MustParseUUID(request.TreeID),
		BranchID:  sqldb.MustParseUUID(request.BranchID),
		MinNodeID: &minNodeID,
		MaxNodeID: &maxNodeID,
		PageSize:  &request.PageSize,
		ShardID:   request.ShardID,
	}
```
will not change at all, meaning that the query events v2 API will end in an endless loop.

We can talk tomorrow

wxing1292:
Talked to @longquanzheng, the min node id is actually changed by 1, so filter will be different

PR #2621, opened 2019-09-30
Fix bootstarp errors

vancexu:
https://github.com/uber/cadence/issues/2617

### DIFF
--- a/service/worker/service.go
+++ b/service/worker/service.go
@@ -23,6 +23,7 @@ package worker
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -344,6 +345,7 @@ func (s *Service) ensureSystemDomainExists(pFactory persistencefactory.Factory, 
 
 func (s *Service) ensureDomainAvailable() {
 	client := cclient.NewClient(s.params.PublicClient, common.SystemLocalDomainName, &cclient.Options{})
+	// Use TerminateWorkflow to check whether domain is refreshed in cache or not
 	err := client.TerminateWorkflow(context.Background(), "wid-not-exist", "", "test reason", nil)
 	retryCount := 0
 	for err != nil && retryCount <= 10 {
@@ -360,7 +362,7 @@ func (s *Service) ensureDomainAvailable() {
 }
 
 func isErrSystemDomainNotExist(err *cshared.EntityNotExistsError) bool {
-	return err.Message == fmt.Sprintf("Domain: %s", common.SystemLocalDomainName)
+	return strings.Contains(err.Message, common.SystemLocalDomainName)
 }
 
 func (s *Service) registerSystemDomain(pFactory persistencefactory.Factory, clusterName string) {
--- a/service/worker/service.go
+++ b/service/worker/service.go
@@ -21,6 +21,8 @@
 package worker
 
 import (
+	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


service/worker/service.go
```diff
@@ -327,17 +331,38 @@ func (s *Service) ensureSystemDomainExists(pFactory persistencefactory.Factory,
...
-	case *shared.EntityNotExistsError:
-		s.logger.Info("cadence-system domain does not exist, attempting to register domain")
-		s.registerSystemDomain(pFactory, clusterName)
-	default:
+	if err == nil {
+		s.ensureDomainAvailable()
+	} else {
+		if _, ok := err.(*shared.EntityNotExistsError); ok {
+			s.logger.Info("cadence-system domain does not exist, attempting to register domain")
+			s.registerSystemDomain(pFactory, clusterName)
```
yycptt: Ensure domain available after registration?

service/worker/service.go
```diff
@@ -327,17 +331,38 @@ func (s *Service) ensureSystemDomainExists(pFactory persistencefactory.Factory,
...
 		s.logger.Fatal("failed to verify if cadence system domain exists", tag.Error(err))
 	}
 }
 
+func (s *Service) ensureDomainAvailable() {
+	client := cclient.NewClient(s.params.PublicClient, common.SystemLocalDomainName, &cclient.Options{})
+	err := client.TerminateWorkflow(context.Background(), "wid-not-exist", "", "test reason", nil)
+	retryCount := 0
+	for err != nil && retryCount <= 10 {
+		nonExistErr, ok := err.(*cshared.EntityNotExistsError)
```
yycptt: What about other error types like service busy? Do we need to keep retrying in that case?

service/worker/service.go
```diff
@@ -327,17 +331,38 @@ func (s *Service) ensureSystemDomainExists(pFactory persistencefactory.Factory,
...
+			s.logger.Info("cadence-system domain does not exist, attempting to register domain")
+			s.registerSystemDomain(pFactory, clusterName)
+		}
 		s.logger.Fatal("failed to verify if cadence system domain exists", tag.Error(err))
 	}
 }
 
+func (s *Service) ensureDomainAvailable() {
+	client := cclient.NewClient(s.params.PublicClient, common.SystemLocalDomainName, &cclient.Options{})
+	err := client.TerminateWorkflow(context.Background(), "wid-not-exist", "", "test reason", nil)
```
yycptt: Consider adding a comment explaining why `TerminateWorkflow` is used here?

service/worker/service.go
```diff
@@ -327,17 +331,38 @@ func (s *Service) ensureSystemDomainExists(pFactory persistencefactory.Factory,
...
+			err = client.TerminateWorkflow(context.Background(), "wid-not-exist", "", "test reason", nil)
+			retryCount++
+		} else {
+			break
+		}
+	}
+}
+
+func isErrSystemDomainNotExist(err *cshared.EntityNotExistsError) bool {
+	return err.Message == fmt.Sprintf("Domain: %s", common.SystemLocalDomainName)
```
yycptt: Equality check seems a bit fragile. Is it possible to just check part of the message like `SystemLocalDomainName` or `Domain`?

service/worker/service.go
vancexu: logically we should but registerSystemDomain already wait for cache refresh, so avoid additional call here.

service/worker/service.go
vancexu: added

service/worker/service.go
vancexu: We should only wait on this DomainNotExist error. Retry on other error make little sense.

service/worker/service.go
vancexu: Addressed.

PR #2634, opened 2019-10-04
Remove Deprecated code of  Events V1

longquanzheng:
One year ago, I introduced History Events V2 with more functionality(forking branches) and performance(removing LWT). However we had to keep V1 running for backward compatibility. After one year, we have found that there is no more workflows running with V1. It's time to remove the V1 code and the logic of backward compatibility. 

TODO:
1. Rename V2 interface to get rid of "V2" since there is no V1 anymore.
2. Make SetHistoryTree transparent in mutableStateBuilder to make it more easy to maintain.


wxing1292:
@yycptt if possible, could you use the paging iterator?
[link](https://github.com/uber/cadence/blob/master/common/collection/pagingIterator.go#L38)

yycptt:
@wxing1292 I think they are a little bit different. The history iterator used by archiver needs the ability to get current state of the iterator and create a new iterator based on some previous states. I don't think the paging iterator supports that. The functionality is needed because archiver workflow needs to periodically record its current progress and resume from last saved state if worker restarts. 

There are some other issues like keeping current iterator state when error occurs, performing lookahead for nextPageToken etc. If you want, we can talk about it in more detail offline.

PR #2636, opened 2019-10-04
Emit tasklist tag for workflow completion event

meiliang86:


service/history/workflowExecutionContext.go
```diff
@@ -699,7 +699,8 @@ func (c *workflowExecutionContextImpl) updateWorkflowExecutionWithNew(
 	// emit workflow completion stats if any
 	if currentWorkflow.ExecutionInfo.State == persistence.WorkflowStateCompleted {
 		if event, err := c.msBuilder.GetCompletionEvent(); err == nil {
-			emitWorkflowCompletionStats(c.metricsClient, domainName, event)
+			taskList := currentWorkflow.ExecutionInfo.TaskList
```
wxing1292: if this is for decision, plz also take sticky tasklist into consideration

meiliang86: This is for customers to filter. sticky tasklist name does not make sense to them.

PR #2638, opened 2019-10-04
upgrade version which server can support

andrewjdawson2016:


meiliang86:
Make sure you release this docker image before the client.

PR #2642, opened 2019-10-07
Wire Kafka with NDC

wxing1292:


service/worker/replicator/replicationTask.go
```diff
@@ -225,30 +301,44 @@ func (t *activityReplicationTask) HandleErr(err error) error {
 		return err
 	}
 
-	retryErr, ok := t.convertRetryTaskError(err)
-	if !ok || retryErr.GetRunId() == "" {
+	retryV1Err, okV1 := t.convertRetryTaskError(err)
+	_, okV2 := t.convertRetryTaskV2Error(err)
```
vancexu: shouldn't this be called  only when okv1 is false? In that case you don't need to put okv1 and okv2 together

service/worker/replicator/replicationTask.go
```diff
@@ -225,30 +301,44 @@ func (t *activityReplicationTask) HandleErr(err error) error {
...
-	endEventID := t.taskID + 1 // the next event ID should be at activity schedule ID + 1
-	resendErr := t.historyRereplicator.SendMultiWorkflowHistory(
-		t.queueID.DomainID, t.queueID.WorkflowID,
-		beginRunID, beginEventID, endRunID, endEventID,
-	)
-
-	if resendErr != nil {
-		t.logger.Error("error resend history", tag.Error(resendErr))
-		// should return the replication error, not the resending error
+	if !okV1 && !okV2 {
```
vancexu: This block looks unreachable, consider what you have in line 306

service/worker/replicator/replicationTask.go
wxing1292: forgot to remove 
```go
if !okV1 || retryV1Err.GetRunId() == "" {
    return err
}
```

PR #2648, opened 2019-10-09
Domain replication followup changes

meiliang86:
- Check host identity before processing domain replication tasks
- Add metrics for domain replication queue ack level

### DIFF
--- a/common/persistence/domainReplicationQueue.go
+++ b/common/persistence/domainReplicationQueue.go
@@ -148,7 +148,7 @@ func (q *domainReplicationQueueImpl) purgeAckedMessages() error {
 	}
 
 	q.metricsClient.
-		Scope(metrics.FrontendDomainReplicationQueueScope, metrics.SourceClusterTag(q.clusterName)).
+		Scope(metrics.FrontendDomainReplicationQueueScope).
 		UpdateGauge(metrics.DomainReplicationTaskAckLevel, float64(minAckLevel))
 
 	q.ackLevelUpdated = false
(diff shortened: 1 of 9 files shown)


common/persistence/domainReplicationQueue.go
```diff
@@ -51,8 +59,10 @@ func NewDomainReplicationQueue(queue Queue, logger log.Logger) DomainReplication
 
 type (
 	domainReplicationQueueImpl struct {
-		logger              log.Logger
 		queue               Queue
+		clusterName         string
```
wxing1292: current cluster name?

common/persistence/domainReplicationQueue.go
```diff
@@ -137,6 +147,10 @@ func (q *domainReplicationQueueImpl) purgeAckedMessages() error {
 		return fmt.Errorf("failed to purge messages: %v", err)
 	}
 
+	q.metricsClient.
+		Scope(metrics.FrontendDomainReplicationQueueScope, metrics.SourceClusterTag(q.clusterName)).
```
wxing1292: i believe the source cluster tag is not necessary

common/persistence/domainReplicationQueue.go
meiliang86: It's called clusterName everywhere else so I'm going to keep it as is.

common/persistence/domainReplicationQueue.go
meiliang86: You are right. I'll remove.

meiliang86:
#2393

PR #2655, opened 2019-10-10
Deprecate historyV1 in archiver history iterator test

yycptt:


yycptt:
Will merge after https://github.com/uber/cadence/pull/2634 is merged.

PR #2669, opened 2019-10-15
Adds prometheus config to config_template.yml

kewu21:
Doesn't have `timerType` specified because it will cause error when
reporting histogram metrics

### DIFF
--- a/docker/config_template.yaml
+++ b/docker/config_template.yaml
@@ -89,6 +89,7 @@ services:
         {{- else if .Env.PROMETHEUS_ENDPOINT }}
         metrics:
             prometheus:
+                timerType: {{ default .Env.PROMETHEUS_TIMER_TYPE "histogram" }}
                 listenAddress: {{ .Env.PROMETHEUS_ENDPOINT }}
         {{- end }}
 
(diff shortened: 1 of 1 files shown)


docker/config_template.yaml
```diff
@@ -86,6 +86,10 @@ services:
             statsd:
                 hostPort: {{ .Env.STATSD_ENDPOINT }}
                 prefix: "cadence-frontend"
+        {{- else if .Env.PROMETHEUS_ENDPOINT }}
+        metrics:
+            prometheus:
```
wxing1292: prefer to add the timerType here, how about this
```yaml
timerType: {{ default .Env.PROMETHEUS_TIMER_TYPE "histogram" }}
```
then if "histogram" does not work for your setup, env PROMETHEUS_TIMERTYPE can be overrided to "summary"
ref: https://github.com/uber-go/tally/blob/1018d35f8be346a864b3ce25f2693c80da7c3fd6/prometheus/reporter.go#L254

kewu21: @wxing1292 good point, will do.

kewu21:
@wxing1292 the comment is addressed. PTAL, thanks!

PR #2672, opened 2019-10-15
Abort domain cache refreshment retry during shutdown

yux0:


wxing1292:
@yux0 which part is broken?

common/cache/domainCache.go
```diff
@@ -378,17 +378,18 @@ func (c *domainCache) GetDomainName(
...
 	defer timer.Stop()
 
 	for {
 		select {
 		case <-c.shutdownChan:
 			return
 		case <-timer.C:
-			for err := c.refreshDomains(); err != nil; err = c.refreshDomains() {
+			timer.Reset(DomainCacheRefreshInterval)
+			if err := c.refreshDomains(); err != nil {
```
wxing1292: how about adding a check within the for loop
```go
select {
case <-c.shutdownChan:
    return
default:
    c.logger.Error("Error refreshing domain cache", tag.Error(err))
    time.Sleep(DomainCacheRefreshFailureRetryInterval)
}

PR #2675, opened 2019-10-16
Delete CompleteFork API from History Persistence

longquanzheng:
We no longer need it because fork will never do clean up when second phase commit fails. We use history scanner to clean up in the background instead.

This would reduce a lot of complexity and improve performance.

TODO:
we could also combine ForkBranch into AppendNodes API to furthure optimize perf. But it may lose some simplicity


service/worker/archiver/activities.go
```diff
@@ -23,12 +23,12 @@ package archiver
 import (
 	"context"
 	"errors"
+	"github.com/uber/cadence/common/persistence"
```
yycptt: import order.

schema/mysql/v57/cadence/schema.sql
```diff
@@ -226,7 +226,7 @@ CREATE TABLE history_tree (
   tree_id        BINARY(16) NOT NULL,
   branch_id      BINARY(16) NOT NULL,
   --
-  in_progress    BOOLEAN NOT NULL,
+  in_progress    BOOLEAN NOT NULL, -- Deprecated
```
wxing1292: i guess this can be deleted

schema/cassandra/cadence/schema.cql
```diff
@@ -350,7 +350,7 @@ CREATE TABLE history_tree (
   tree_id           uuid,
   branch_id         uuid,
   ancestors         list<frozen<branch_range>>,
-  in_progress       boolean, -- For fork operation to prevent race condition to leak event data when forking branches
+  in_progress       boolean, -- Deprecated: For fork operation to prevent race condition to leak event data when forking branches
```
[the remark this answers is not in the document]
wxing1292: could you plz also delete the corresponding `in_progress` in the schema update files?

schema/mysql/v57/cadence/schema.sql
wxing1292: could you plz also delete the corresponding `in_progress` in the schema update files?

service/worker/archiver/activities.go
longquanzheng: Good catch. Will fix it.

schema/cassandra/cadence/schema.cql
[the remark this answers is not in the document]
longquanzheng: Sure. That makes sense.

schema/mysql/v57/cadence/schema.sql
sevein: Out of curiosity, wouldn't be safer to delete the column from a new migration?

longquanzheng: @sevein Are you saying that, if someone use old commit of code but install the schema from this new commit, then it will break. I agree but this usually very rare because people should use the schema files from the same commit I think. But this PR does break some implicit best practice that schema migration changes should be incremental and immutable...

I think we are just do special case here. Because adding one new schema version just for deleting this deprecated column doesn't buy much, only giving us one more version which is not ideal. 

Thanks for your thoughts anyway.

sevein: Thank you, understood!

PR #2678, opened 2019-10-17
Consistent query support

andrewjdawson2016:
This diff provides server side support for consistent query. Consistent query means that queries which are issued after events will be based on a workflow state with those events applied. 

The default for query is to run in eventual consistency mode. The API level change for consistent query is to introduce another consistency mode called STRONG.


idl/github.com/uber/cadence/history.thrift
```diff
@@ -335,12 +336,11 @@ struct SyncActivityRequest {
 
 struct QueryWorkflowRequest {
   10: optional string domainUUID
-  20: optional shared.WorkflowExecution execution
-  30: optional shared.WorkflowQuery query
```
wxing1292: is this backwards compatible?

wxing1292: or this is not released yet?

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1093,7 +1100,7 @@ struct PollForDecisionTaskResponse {
   90: optional TaskList WorkflowExecutionTaskList
   100:  optional i64 (js.type = "Long") scheduledTimestamp
   110:  optional i64 (js.type = "Long") startedTimestamp
-  120:  optional list<WorkflowQuery> queries
+  120:  optional map<string, WorkflowQuery> queries
```
wxing1292: same here

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1110,7 +1117,7 @@ struct RespondDecisionTaskCompletedRequest {
   60: optional bool returnNewDecisionTask
   70: optional bool forceCreateNewDecisionTask
   80: optional string binaryChecksum
-  90: optional list<WorkflowQueryResult> queryResults
+  90: optional map<string, WorkflowQueryResult> queryResults
```
wxing1292: same here

service/history/decisionHandler.go
```diff
@@ -456,11 +456,29 @@ Update_History_Loop:
...
+		} else {
+			for _, bid := range bufferedQueries {
+				if _, ok := queryResults[bid]; !ok {
+					hasUnhandledQueries = true
+					break
+				}
+			}
+		}
+
+		createNewDecisionTask := msBuilder.IsWorkflowExecutionRunning() && (hasUnhandledEvents || request.GetForceCreateNewDecisionTask() || activityNotStartedCancelled || hasUnhandledQueries)
```
wxing1292: this maybe a long term thing, but for respond decision task completed, plz keep in mind that the code is still hard to read, and we need to further refactor this.

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
...
 	request *h.QueryWorkflowRequest,
 ) (retResp *h.QueryWorkflowResponse, retErr error) {
-	domainCache, err := e.shard.GetDomainCache().GetDomainByID(request.GetDomainUUID())
+
+	context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetRequest().GetExecution())
 	if err != nil {
 		return nil, err
 	}
-	context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetExecution())
+	defer release(retErr)
```
wxing1292: there is a difference between
`defer release(retErr)` vs `defer func() { release(retErr) } ()`
the first one basically get the value of retErr right away, while the second one will capture the retErr when defer function is invoked

here we should use the second one

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
...
+	ms, err := context.loadWorkflowExecution()
```
wxing1292: plz do not use `ms` try `mutableState` for readibility

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
 	ctx ctx.Context,
 	request *h.QueryWorkflowRequest,
 ) (retResp *h.QueryWorkflowResponse, retErr error) {
-	domainCache, err := e.shard.GetDomainCache().GetDomainByID(request.GetDomainUUID())
+
+	context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetRequest().GetExecution())
```
wxing1292: plz put this function logic into a dedicated file, history engine is already too large

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
...
+					},
+				},
+			}, nil
+		}
+	}
+
+	execInfo := ms.GetExecutionInfo()
+	clientFeature := client.NewFeatureImpl(execInfo.ClientLibraryVersion, execInfo.ClientFeatureVersion, execInfo.ClientImpl)
+
+	de, err := e.shard.GetDomainCache().GetDomainByID(request.GetDomainUUID())
```
wxing1292: what is `de` ?

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
...
-		}
-		msBuilder, err := context.loadWorkflowExecution()
-		if err != nil {
-			release(err)
-			retResp = nil
-			retErr = err
-			return
-		}
-		msBuilder.DeleteInMemoryDecisionTask()
+	scope := e.metricsClient.Scope(metrics.HistoryQueryWorkflowScope)
```
wxing1292: metricsScope ?

service/history/historyEngine.go
```diff
@@ -690,108 +698,173 @@ func (e *historyEngineImpl) QueryWorkflow(
...
+		sw := scope.StartTimer(metrics.DirectQueryDispatchLatency)
+		defer sw.Stop()
+		mresp, err := e.queryDirectlyThroughMatching(ctx, msResp, clientFeature, request.GetDomainUUID(), req)
+		return mresp, err
 	}
 
-	// at this point query has been buffered and there is a pending decision task to dispatch the query on
+	// If we get here it means query could not be dispatched through matching directly, so it must be buffered to be dispatched on the next decision task.
+	// There is a hard guarantee in the system that at this point a new decision task will get generated.
+	qr := ms.GetQueryRegistry()
```
wxing1292: `qr` -> `queryRegistry` ?

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (q *queryImpl) getQueryTermCh() <-chan struct{} {
+	q.RLock()
+	defer q.RUnlock()
+
+	return q.termCh
+}
+
+func (q *queryImpl) completeQuery(queryResult *shared.WorkflowQueryResult) error {
```
wxing1292: this format plz
```go
completeQuery(
    queryResult *shared.WorkflowQueryResult,
) error
```

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	q.state = queryStateCompleted
+	q.queryResult = queryResult
+	close(q.termCh)
+	return nil
+}
+
+func validateQueryResult(queryResult *shared.WorkflowQueryResult) error {
+	if queryResult == nil {
+		return errors.New("query result is nil")
```
wxing1292: internal server error?

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	validAnswered := queryResult.GetResultType().Equals(shared.QueryResultTypeAnswered) && queryResult.Answer != nil && queryResult.ErrorDetails == nil && queryResult.ErrorReason == nil
+	validFailed := queryResult.GetResultType().Equals(shared.QueryResultTypeFailed) && queryResult.Answer == nil && queryResult.ErrorDetails != nil && queryResult.ErrorReason != nil
```
wxing1292: indentation?

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return q.termCh
+}
+
+func (q *queryImpl) completeQuery(queryResult *shared.WorkflowQueryResult) error {
+	q.Lock()
+	defer q.Unlock()
+
+	if q.state == queryStateCompleted {
+		return errors.New("query already completed")
```
wxing1292: maybe it is a good idea to define all these err at the very top

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+		queryResult: q.queryResult,
+		state:       q.state,
+	}
+}
+
+func (q *queryImpl) getQueryTermCh() <-chan struct{} {
+	q.RLock()
+	defer q.RUnlock()
+
+	return q.termCh
```
wxing1292: this is a getter, unless the termCh can be changed concurrently, lock is unnecessary

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+
+	queryImpl struct {
+		sync.RWMutex
+
+		id          string
+		queryInput  *shared.WorkflowQuery
+		queryResult *shared.WorkflowQueryResult
+		termCh      chan struct{}
+		state       queryState
+	}
```
wxing1292:
```go
	queryImpl struct {
		id          	string
		termCh	chan struct{}
		
		sync.RWMutex
		queryInput	*shared.WorkflowQuery
		queryResult	*shared.WorkflowQueryResult
		state       		queryState
	}
```

wxing1292: generally, things are immutable should be put at the very top
things are guarded by lock should be put under the lock

service/history/query.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (q *queryImpl) getQuerySnapshot() *querySnapshot {
+	q.RLock()
+	defer q.RUnlock()
+
+	return &querySnapshot{
+		id:          q.id,
+		queryInput:  q.queryInput,
+		queryResult: q.queryResult,
```
wxing1292: this is returning a pointer out, i.e. caller can manipulate the query input / query result while NOT holding a lock
if the intention of this API is truly creating a snapshot, then a deep copy should be made

idl/github.com/uber/cadence/history.thrift
andrewjdawson2016: Not released yet, so this is fine.

idl/github.com/uber/cadence/shared.thrift
andrewjdawson2016: same

idl/github.com/uber/cadence/shared.thrift
andrewjdawson2016: same

service/history/historyEngine.go
andrewjdawson2016: Thanks for catching this

service/history/historyEngine.go
andrewjdawson2016: ok

service/history/historyEngine.go
andrewjdawson2016: As we spoke about offline this should be done as part of a larger refactor where all the logics in historyEngine are split into different files.

service/history/historyEngine.go
andrewjdawson2016: domain entry

service/history/historyEngine.go
andrewjdawson2016: In other places here we we do the following 

```e.metricsClient.IncCounter(metrics.HistoryRespondActivityTaskFailedScope, metrics.StaleMutableStateCounter)```

Instead of redeclaring the scope every time a metric is emitted I think a better pattern is to create a scope object and use that to emit metrics.

service/history/historyEngine.go
andrewjdawson2016: ok

service/history/query.go
andrewjdawson2016: ok

service/history/query.go
andrewjdawson2016: It does not really matter much because this metric is never returned to the user, it will only be emitted to log errors. But logically it is internal service error so I changed it to such.

service/history/query.go
andrewjdawson2016: done

service/history/query.go
andrewjdawson2016: Just the way it is showing in github. But I made the line shorter maybe that will display better.

service/history/query.go
andrewjdawson2016: sounds good, made the change.

service/history/query.go
andrewjdawson2016: This API cannot afford the overhead of making a truly deep copy because query inputs/results can be big. Here I return a shallow copy for some protection but its not perfect. Since queryRegistry and query are just an implementation detail of mutable state I think its okay to not have perfect encapsulation here.

wxing1292: an suggestion is to make this function take in an action function, which will do the logic for the caller.

wxing1292: snapshot should not return pointer, which will let the caller has the ability to change the internal state.

andrewjdawson2016: Spoke offline we renamed this to getInternalState.

PR #2680, opened 2019-10-17
Rewrite passive task processing life cycle

wxing1292:
* Add dedicated dynamic config for passive task processing: verification retry duration; fetch history from remote duration. 
* Group passive task processing post processing actions into one dedicated file.
* Add new type taskInfo grouping task attempt & start processing time for metrics / logging.


service/history/taskProcessor.go
```diff
@@ -45,6 +44,14 @@ type (
 	taskInfo struct {
 		processor taskExecutor
 		task      queueTaskInfo
+
+		attempt   int
+		startTime time.Time
+		logger    log.Logger // TODO logger initialization logic
```
yycptt: What's this TODO? I think we have a `initializeLoggerForTask` function, right?

PR #2699, opened 2019-10-22
History DLQ implementation

meiliang86:


common/metrics/defs.go
```diff
@@ -949,10 +953,12 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		PersistenceDeleteTaskScope:                               {operation: "PersistenceDelete"},
 		PersistenceGetCurrentExecutionScope:                      {operation: "GetCurrentExecution"},
 		PersistenceGetTransferTasksScope:                         {operation: "GetTransferTasks"},
-		PersistenceGetReplicationTasksScope:                      {operation: "GetReplicationTasks"},
 		PersistenceCompleteTransferTaskScope:                     {operation: "CompleteTransferTask"},
 		PersistenceRangeCompleteTransferTaskScope:                {operation: "RangeCompleteTransferTask"},
+		PersistenceGetReplicationTasksScope:                      {operation: "GetReplicationTasks"},
 		PersistenceCompleteReplicationTaskScope:                  {operation: "CompleteReplicationTask"},
+		PersistencePutReplicationTaskToDLQScope:                  {operation: "PersistencePutReplicationTaskToDLQ"},
+		PersistenceGetReplicationTasksFromDLQScope:               {operation: "PersistenceGetReplicationTasksFromDLQ"},
```
wxing1292: PersistenceReplicationTaskToDLQScope
PersistenceReplicationTaskFromDLQScope
?
seems more symmetrical?

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -63,6 +63,9 @@ const (
 	rowTypeReplicationDomainID   = "10000000-5000-f000-f000-000000000000"
 	rowTypeReplicationWorkflowID = "20000000-5000-f000-f000-000000000000"
 	rowTypeReplicationRunID      = "30000000-5000-f000-f000-000000000000"
+	// Row Constants for Replication Task DLQ Row. Source cluster name will be used as WorkflowID.
+	rowTypeDLQDomainID = "10000000-6000-f000-f000-000000000000"
+	rowTypeDLQRunID    = "30000000-6000-f000-f000-000000000000"
```
wxing1292:
```go
rowTypeDLQDomainID = "10000000-6000-f000-f000-000000000000"
rowTypeDLQWorkflowID = "10000000-6000-f000-f000-000000000000"
rowTypeDLQRunID    = "30000000-6000-f000-f000-000000000000"
```
and put the source cluster name in the task or blob

common/persistence/dataInterfaces.go
```diff
@@ -2422,3 +2439,22 @@ func SplitHistoryGarbageCleanupInfo(info string) (domainID, workflowID, runID st
 	workflowID = info[len(domainID)+1 : workflowEnd]
 	return
 }
+
+// NewGetReplicationTasksFromDLQRequest creates a new GetReplicationTasksFromDLQRequest
+func NewGetReplicationTasksFromDLQRequest(
+	sourceClusterName string,
```
wxing1292: why do we need a source cluster?

service/history/replicationTaskProcessor.go
```diff
@@ -278,6 +364,32 @@ func isTransientRetryableError(err error) bool {
...
+	select {
+	case <-p.done:
+		p.logger.Info("ReplicationTaskProcessor shutting down.")
+		return false
+	default:
+		return true
+	}
+}
+
+func toPersistenceReplicationInfo(info map[string]*shared.ReplicationInfo) map[string]*persistence.ReplicationInfo {
```
wxing1292: should the kafka deprecation support 2DC case?

wxing1292: also format plz

service/history/replicationTaskProcessor.go
```diff
@@ -269,6 +270,91 @@ func (p *ReplicationTaskProcessor) processTaskOnce(replicationTask *r.Replicatio
 	return err
 }
 
+func (p *ReplicationTaskProcessor) putReplicationTaskToDLQ(replicationTask *r.ReplicationTask) error {
+	request, err := p.generateDLQRequest(replicationTask)
+	if err != nil {
+		p.logger.Error("Failed to generate DLQ replication task.", tag.Error(err))
+		// We cannot deserialize the task. Dropping it.
+		return nil
```
wxing1292: why dropping a task?

service/history/replicationTaskProcessor.go
```diff
@@ -202,40 +205,38 @@ Loop:
...
 }
 
 func (p *ReplicationTaskProcessor) processTaskOnce(replicationTask *r.ReplicationTask) error {
 	var err error
 	var scope int
 	switch replicationTask.GetTaskType() {
 	case r.ReplicationTaskTypeDomain:
-		scope = metrics.DomainReplicationTaskScope
-		err = p.handleDomainReplicationTask(replicationTask)
+		// Domain replication task should be handled in worker (domainReplicationMessageProcessor)
```
wxing1292: if this case is not expected, either panic or return an error

service/history/replicationTaskProcessor.go
```diff
@@ -153,6 +151,7 @@ Loop:
 		// for each iteration, do close check first
 		select {
 		case <-p.done:
+			p.logger.Info("ReplicationTaskProcessor shutting down.")
```
wxing1292: can you separate the logic into dedicated functions?

service/history/replicationTaskProcessor.go
```diff
@@ -202,40 +205,38 @@ Loop:
...
 			return
 		}
 	}
 }
 
-func (p *ReplicationTaskProcessor) processTask(replicationTask *r.ReplicationTask) {
-	var err error
+func (p *ReplicationTaskProcessor) processTask(replicationTask *r.ReplicationTask) error {
+	err := backoff.Retry(func() error {
+		return p.processTaskOnce(replicationTask)
```
wxing1292: i would suggest copy pasting logic from worker (or move those logic to common package and reference it here)
ref: https://github.com/uber/cadence/blob/master/service/worker/replicator/replicationTask.go

service/history/replicationTaskFetcher.go
```diff
@@ -192,8 +191,8 @@ Loop:
 				// We don't receive tasks from previous fetch so processors are all sleeping.
 				f.logger.Debug("Skip fetching as no processor is asking for tasks.")
 				timer.Reset(backoff.JitDuration(
-					time.Duration(f.config.AggregationIntervalSecs)*time.Second,
```
wxing1292: can we move logic within the for loop into smaller functions?

service/history/handler.go
```diff
@@ -1575,6 +1572,25 @@ func (h *Handler) ReapplyEvents(
 	return nil
 }
 
+// GetReplicationTasksFromDLQ is only used for integration test verification purpose
+func (h *Handler) GetReplicationTasksFromDLQ(sourceClusterName string) ([]*persistence.ReplicationTaskInfo, error) {
```
wxing1292:
```go
func (h *Handler) GetReplicationTasksFromDLQ(
    sourceClusterName string,
) ([]*persistence.ReplicationTaskInfo, error) {
```

and why specifying sourceClusterName?

service/history/handler.go
```diff
@@ -1575,6 +1572,25 @@ func (h *Handler) ReapplyEvents(
...
+	var tasksFromDLQ []*persistence.ReplicationTaskInfo
+	for i := 0; i < h.config.NumberOfShards; i++ {
+		shardItem := h.controller.getHistoryShardItem(i)
+		if shardItem != nil {
```
wxing1292: if shard item is not present in this host, then just giveup?
i know this API is for integration test, but why creating an API just for test?

common/metrics/defs.go
meiliang86: I think the current names are more meaningful.

common/persistence/cassandra/cassandraPersistence.go
meiliang86: Discussed offline. We will have source cluster in the DLQ because source task ID can have collision. On the other side cluster name string is error prone, so we will use initial failover version.

common/persistence/dataInterfaces.go
meiliang86: DLQ has a source cluster index in it. So get operations will be per source cluster.

service/history/replicationTaskProcessor.go
meiliang86: It should support 2DC case. Anything here you think that does not work for 2DC?

service/history/replicationTaskProcessor.go
meiliang86: DLQ has the same schema as replication task. So if we cannot deserialize the task then there is not much we can do. We can potentially create a new blob column to store the payload. Not sure if it's worth it.

service/history/replicationTaskProcessor.go
meiliang86: Done

service/history/handler.go
meiliang86: Removed this method.

service/history/replicationTaskFetcher.go
meiliang86: done :)

service/history/replicationTaskProcessor.go
meiliang86: done

service/history/replicationTaskProcessor.go
meiliang86: I'll remove the code in worker after we deprecate kafka.

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -907,37 +908,49 @@ func (m *sqlExecutionManager) GetReplicationTasks(
...
+			MinTaskID: readLevel,
+			MaxTaskID: maxReadLevelInclusive,
+			PageSize:  request.BatchSize,
+		})
+
+	if err != nil && err != sql.ErrNoRows {
+		return nil, &workflow.InternalServiceError{
+			Message: fmt.Sprintf("GetReplicationTasks operation failed. Select failed: %v", err),
 		}
 	}
```
wxing1292: a switch case may make this looks cleaner

common/persistence/sql/sqlExecutionManager.go
```diff
@@ -984,17 +997,47 @@ func (m *sqlExecutionManager) CompleteReplicationTask(
...
+	if err != nil {
+		if err != sql.ErrNoRows {
+			return nil, &workflow.InternalServiceError{
+				Message: fmt.Sprintf("GetReplicationTasks operation failed. Select failed: %v", err),
+			}
+		}
+	}
+	if len(rows) == 0 {
+		return &p.GetReplicationTasksResponse{}, nil
+	}
```
wxing1292: i would rather make this a switch case, like below, and let populateGetReplicationTasksResponse deal with empty rows
```go
switch err {
case nil:
    return m.populateGetReplicationTasksResponse(rows, request.MaxReadLevel)
case sql.ErrNoRows:
    return &p.GetReplicationTasksResponse{}, nil
default:
    return err
}
```

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -205,14 +205,29 @@ type (
...
 	// can be used to filter results through a WHERE clause
 	ReplicationTasksFilter struct {
 		ShardID   int
-		TaskID    *int64
-		MinTaskID *int64
-		MaxTaskID *int64
-		PageSize  *int
+		MinTaskID int64
+		MaxTaskID int64
+		PageSize  int
```
wxing1292: any reason to changing from pointer?
seems sql side convention is using pointer

service/history/replicationTaskFetcher.go
```diff
@@ -33,7 +33,7 @@ import (
 	"github.com/uber/cadence/common/cluster"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
-	"github.com/uber/cadence/common/service/config"
+	c "github.com/uber/cadence/common/service/config"
```
wxing1292: can we find a better name? `c` is not good enough

common/persistence/sql/sqlExecutionManager.go
meiliang86: done

common/persistence/sql/sqlExecutionManager.go
meiliang86: done

common/persistence/sql/storage/sqldb/interfaces.go
meiliang86: We don't need them to be pointers at all. We used to use pointer and then dereference it in SQL layer.

PR #2703, opened 2019-10-23
WIP: Generic task ack manager implementation

yycptt:
Tests will be added later, when the entire task processing logic refactor is done.

### DIFF
--- a/common/task/ackMgr.go
+++ b/common/task/ackMgr.go
@@ -49,10 +49,9 @@ type (
 	}
 
 	ackMgrImpl struct {
-		ackMgrBase    AckMgrBase
-		logger        log.Logger
-		metricsClient metrics.Client
-		metricsScope  metrics.Scope
+		ackMgrBase   AckMgrBase
+		logger       log.Logger
+		metricsScope metrics.Scope
 
 		sync.RWMutex
 		iterator         collection.Iterator
@@ -67,7 +66,6 @@ type (
 func NewAckMgr(
 	ackMgrBase AckMgrBase,
 	logger log.Logger,
-	metricsClient metrics.Client,
 	metricsScope metrics.Scope,
 	ackLevel SequenceID,
 	readLevel SequenceID,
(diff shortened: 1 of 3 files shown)


common/task/ackMgr.go
```diff
@@ -0,0 +1,176 @@
...
+
+type (
+	// AckMgr is the interface for reading/acknowledging Tasks
+	AckMgr interface {
+		ReadTasks() ([]Info, bool, error)
+		CompleteTask(SequenceID)
+		AckLevel() SequenceID
+		ReadLevel() SequenceID
+		UpdateAckLevel() SequenceID
+		PendingTasks() int64
```
wxing1292: plz keep the managing pending task functionality in ack manager
the io functionality should be implemented by each replication / transfet / timer component

common/task/ackMgr.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	// AckMgrBase is the interface for the functionality that AckMgr relies on
+	AckMgrBase interface {
+		PaginationFn(SequenceID, SequenceID) collection.PaginationFn
+		UpdateMaxReadLevel() SequenceID
+	}
```
wxing1292: this interface may not be necessary, since each replication / transfer / timer component (event loop) will do the IO loading task then adding new task to ack manager

common/task/task.go
```diff
@@ -21,18 +21,32 @@
...
+	}
+
+	// SequenceIDs is a list of task SequenceID
+	SequenceIDs []SequenceID
+
+	// SequenceID determines the total order of tasks
+	SequenceID struct {
+		VisibilityTimestamp time.Time
+		TaskID              int64
 	}
```
wxing1292: this is too specific, can we make a task key interface? basically ack manager does not need to understand this key interface implementation details, but have the ability to sort it for acking

yycptt: How about reusing the `task.Info` type as the key? The reason is that when implementing multi-cursor, each ackMgr needs additional information (beyond taskID and timestamp) about a task so that it knows how to split the outstanding tasks. Reusing also avoid unnecessarily type conversions.

PR #2719, opened 2019-10-26
Set default event store version to v2 in apply events

yux0:
Set default event store version to v2 in apply events. This change is needed before service upgrade to 0.10


wxing1292:
could you make change in history replicator, apply events API?
that API is the only entrance

PR #2721, opened 2019-10-26
Modify state builder to use standardized task generation logic

wxing1292:
* Rewrite state builder
* Rewrite UT

### DIFF
--- a/service/history/nDCHistoryReplicator.go
+++ b/service/history/nDCHistoryReplicator.go
@@ -46,7 +46,7 @@ var (
 
 type (
 	stateBuilderProvider func(
-		msBuilder mutableState,
+		mutableState mutableState,
 		logger log.Logger) stateBuilder
 
 	mutableStateProvider func(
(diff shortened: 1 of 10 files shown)


service/history/stateBuilder.go
```diff
@@ -78,528 +72,565 @@ var _ stateBuilder = (*stateBuilderImpl)(nil)
...
+				parentDomainID,
+				execution,
+				requestID,
+				event,
+			); err != nil {
+				return nil, err
+			}
+
+			if err := taskGenerator.generateRecordWorkflowStartedTasks(
+				b.unixNanoToTime(event.GetTimestamp()),
```
yycptt: nit: create a variable for the converted event timestamp?

service/history/stateBuilder.go
```diff
@@ -26,17 +26,19 @@ import (
 	"time"
 
 	"github.com/pborman/uuid"
+
```
yycptt: nit: remove empty line

service/history/nDCHistoryReplicator.go
```diff
@@ -149,10 +149,18 @@ func newNDCHistoryReplicator(
 			return newNDCWorkflowResetter(shard, transactionMgr, domainID, workflowID, baseRunID, newContext, newRunID, logger)
 		},
 		newStateBuilder: func(
-			msBuilder mutableState,
+			state mutableState,
```
yycptt: mutableState?

wxing1292: cannot name it mutable state since above mutable state is already used

service/history/stateBuilder.go
wxing1292: this is due to the new goimports sorting

PR #2732, opened 2019-10-28
Consistent query do not generate decision task for unhandled queries

andrewjdawson2016:
Additional refactoring (in the future) needed: #2740


service/history/decisionHandler.go
```diff
@@ -576,6 +548,25 @@ Update_History_Loop:
 			}
 
 			return nil, updateErr
+		} else {
```
meiliang86: You don't need else here?

service/history/decisionHandler.go
```diff
@@ -576,6 +548,25 @@ Update_History_Loop:
...
+			// at this point the update is successful, so answer all the queries which we have answers for
```
meiliang86: Also move the section to a separate method. This method is already way too long.

service/history/decisionHandler.go
```diff
@@ -576,6 +548,25 @@ Update_History_Loop:
...
+			qr := msBuilder.GetQueryRegistry()
+			for id, result := range req.GetCompleteRequest().GetQueryResults() {
+				if err := qr.completeQuery(id, result); err != nil {
```
meiliang86: Do you need log here?

service/history/decisionHandler.go
```diff
@@ -576,6 +548,25 @@ Update_History_Loop:
...
+					handler.metricsClient.IncCounter(metrics.HistoryRespondDecisionTaskCompletedScope, metrics.CompleteQueryFailedCount)
+				}
+			}
+
+			// if no decision task was created then it means no buffered events came in during this decision task's handling
+			// this means all buffered queries can be dispatched directly through matching at this point
+			if !createNewDecisionTask {
+				buffered := qr.getBufferedSnapshot()
+				for _, id := range buffered {
+					if err := qr.unblockQuery(id); err != nil {
```
meiliang86: Also log here?

service/history/historyEngine.go
```diff
@@ -173,6 +173,8 @@ var (
 	ErrEventsAterWorkflowFinish = &workflow.InternalServiceError{Message: "error validating last event being workflow finish event"}
 	// ErrQueryTimeout is the error indicating query timed out before being answered
 	ErrQueryTimeout = errors.New("query timed out")
+	// ErrQueryEnteredInvalidState is error indicating query entered invalid state
+	ErrQueryEnteredInvalidState = &workflow.InternalServiceError{Message: "query entered invalid state, this should be impossible"}
```
meiliang86: I think the convention is to panic in these impossible cases.

service/history/historyEngine.go
```diff
@@ -824,16 +826,31 @@ func (e *historyEngineImpl) QueryWorkflow(
...
-		switch result.GetResultType() {
-		case workflow.QueryResultTypeAnswered:
-			return &h.QueryWorkflowResponse{
-				Response: &workflow.QueryWorkflowResponse{
-					QueryResult: result.GetAnswer(),
-				},
-			}, nil
-		case workflow.QueryResultTypeFailed:
-			return nil, &workflow.QueryFailedError{Message: result.GetErrorMessage()}
+		switch state.state {
```
meiliang86: Move these into separate method.

service/history/query.go
```diff
@@ -113,6 +115,22 @@ func (q *queryImpl) completeQuery(
 	return nil
 }
 
+func (q *queryImpl) unblockQuery() error {
+	q.Lock()
+	defer q.Unlock()
+
+	if q.terminalState() {
+		return errAlreadyInTerminalState
```
meiliang86: Should you just simply return nil here? Otherwise it does not seem there is a way out? Meaning that queryRegistryImpl.unblockQuery will not remove it from the map

service/history/decisionHandler.go
andrewjdawson2016: https://github.com/uber/cadence/issues/2740

I do not want to refactor this method piece by piece. I think we need to take a full pass over this method so we can refactor in a consistent manner. For now I prefer to leave it as one chunk of code.

andrewjdawson2016: I agree this method needs to be refactored but I think doing it piece wise and then refactoring it will be harder.

service/history/historyEngine.go
andrewjdawson2016: I personally prefer panic but it seems like in other places we return this error. For example:

```
	switch wfIDReusePolicy {
	case workflow.WorkflowIdReusePolicyAllowDuplicateFailedOnly:
		if _, ok := FailedWorkflowCloseState[prevCloseState]; !ok {
			msg := "Workflow execution already finished successfully. WorkflowId: %v, RunId: %v. Workflow ID reuse policy: allow duplicate workflow ID if last run failed."
			return getWorkflowAlreadyStartedError(msg, prevStartRequestID, execution.GetWorkflowId(), prevRunID)
		}
	case workflow.WorkflowIdReusePolicyAllowDuplicate:
		// as long as workflow not running, so this case has no check
	case workflow.WorkflowIdReusePolicyRejectDuplicate:
		msg := "Workflow execution already finished. WorkflowId: %v, RunId: %v. Workflow ID reuse policy: reject duplicate workflow ID."
		return getWorkflowAlreadyStartedError(msg, prevStartRequestID, execution.GetWorkflowId(), prevRunID)
	default:
		return &workflow.InternalServiceError{Message: "Failed to process start workflow reuse policy."}
	}
```

There are other places in code where we panic in impossible cases. Do you have a strong preference about which to use here?

service/history/historyEngine.go
andrewjdawson2016: Here I also prefer to keep it as one chunk of code. The reason is all other methods in historyEngine do this. I think it's confusing to see all the methods look one way and then see this method spread into several functions.

Wenquan and I talked about a solution which I like: https://github.com/uber/cadence/issues/2748

Basically most of the methods in historyEngine should be their own file like what Wenquan did with decisionHandler. Then these files can be nicely factored.

service/history/query.go
andrewjdawson2016: I think I prefer to return this error here because if you return nil it makes the logic in queryRegistry weird 

```
	if err := q.completeQuery(queryResult); err != nil {
		return err
	}
	delete(r.buffered, id)
	r.completed[id] = q
	return nil
```

If you return nil here then after ```complete``` is called "successfully" then you have to run checks for what state you are in. 

Also query does not does have to move to a terminal state to be removed from the map. The only way things get removed from queryRegistry is by having ```delete``` called. So the pattern for using query registry should be 

```
qr := getRegistry
queryID := qr.Buffer()
defer qr.Remove(queryID) // this removes from all maps regardless of the progress query has made
```

andrewjdawson2016: Since ```delete``` force removes a query from map regardless of state I think the right pattern for coding state transitions is to return an error whenever the state transition did not occur.

service/history/decisionHandler.go
andrewjdawson2016: added

service/history/decisionHandler.go
andrewjdawson2016: added

service/history/historyEngine.go
wxing1292: the convention is to return internal server err

service/history/queryRegistry.go
```diff
@@ -40,11 +40,15 @@ type (
 		hasCompleted() bool
 		getCompletedSnapshot() []string
 
+		hasUnblocked() bool
+		getUnblockedSnapshot() []string
```
wxing1292: what is a snapshot?

wxing1292: i would prefer to get a rough idea of what the api do from reading the api.

wxing1292: possible api names?
```go
		hasBufferedQuery() bool
		getBufferedIDs() []string
		hasCompletedQuery() bool
		getCompletedIDs() []string

		hasUnblockedQuery() bool
		getUnblockedIDs() []string
```

service/history/queryRegistry.go
```diff
@@ -91,6 +97,20 @@ func (r *queryRegistryImpl) getCompletedSnapshot() []string {
...
+	defer r.RUnlock()
+
+	return len(r.unblocked) > 0
+}
+
+func (r *queryRegistryImpl) getUnblockedSnapshot() []string {
+	r.RLock()
+	defer r.RUnlock()
+
+	return getIDs(r.unblocked)
```
wxing1292: make the `getIDs` function a method of  `* queryRegistryImpl`

PR #2746, opened 2019-10-30
Fix workflow execution state

wxing1292:


wxing1292:
close due to mis-understanding

wxing1292:
@yux0 actually your are right

PR #2747, opened 2019-10-31
Support 2DC/NDC compatibility

yux0:
Support 2DC/NDC compatibility


service/history/historyReplicator.go
```diff
@@ -135,8 +135,8 @@ func newHistoryReplicator(
 				shard,
 				logger,
 				msBuilder,
-				func(mutableState mutableState) mutableStateTaskGenerator {
-					return newMutableStateTaskGenerator(shard.GetDomainCache(), logger, mutableState)
+				func(msBuilder mutableState) mutableStateTaskGenerator {
+					return newMutableStateTaskGenerator(shard.GetDomainCache(), logger, msBuilder)
```
wxing1292: revert this change?

service/history/historyReplicator.go
```diff
@@ -947,14 +947,17 @@ func (r *historyReplicator) terminateWorkflow(
 					domainID,
 					workflowID,
 					runID,
-					msBuilder.GetNextEventID(),
+					mutableState.GetNextEventID(),
 				)
 			}
 
 			// setting the current version to be the last write version
-			msBuilder.UpdateReplicationStateVersion(currentLastWriteVersion, true)
```
wxing1292: all `UpdateReplicationStateVersion` in this file may need to be changed to `UpdateCurrentVersion`

service/history/historyReplicator.go
yux0: reverted

service/history/historyReplicator.go
yux0: replaced

PR #2749, opened 2019-10-31
Fix workflow completed race for signal/cancel/terminate

vancexu:
https://github.com/uber/cadence/issues/2717


service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
 	createDecision bool
 }
 
-func (e *historyEngineImpl) updateWorkflowExecutionWithAction(
+type workflowLockedContext struct {
+	context workflowExecutionContext
+	release releaseWorkflowExecutionFunc
+	err     error
+}
+
```
wxing1292: can we put this is a dedicated file? preferably workflowExecutionUtil

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+type updateWorkflowActionFunc func(builder mutableState) (*updateWorkflowAction, error)
```
wxing1292:
```go
type updateWorkflowActionFunc func(mutableState mutableState) (*updateWorkflowAction, error)
```

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+	if execution.GetRunId() != "" {
+		workflowContext := e.getWorkflowLockedContext(ctx, domainID, execution)
+		return e.updateWorkflowExecutionWithContextAndAction(workflowContext, action)
+	}
+
+	// for execution without runID,
+	// reload current run in case continueAsNew caused unexpected workflow already completed error
+	var release releaseWorkflowExecutionFunc
+	for attempt := 0; attempt < retryCountForContextReload; attempt++ {
+		context, release, err0 := e.historyCache.getOrCreateWorkflowExecution(ctx, domainID, execution)
```
wxing1292: err0?

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+		if err0 != nil {
+			return err0
+		}
+		msBuilder, err1 := context.loadWorkflowExecution()
```
wxing1292: let's try to deprecate the name msBuilder

wxing1292: why err1?

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+		if err1 != nil {
+			release(err1)
+			return err1
+		}
+
+		if msBuilder.IsWorkflowExecutionRunning() {
+			workflowContext := &workflowLockedContext{
+				context, release, err0,
+			}
```
wxing1292: can we create a new function for this struct?
initialization of a struct like this is really confusing

service/history/historyEngine.go
wxing1292: missing a 
```go
defer func() { release(retError) } ()
```
?

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+			return e.updateWorkflowExecutionWithContextAndAction(workflowContext, action)
+		}
+
+		// workflow is completed because of continue as new, fetch latest run
```
wxing1292: how do you know the workflow is completed because of continue as new without checking the close status?
the current workflow can be terminated, completed, failed, continue-as-newed, timeouted

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+		resp, err2 := e.historyCache.getCurrentExecutionWithRetry(&persistence.GetCurrentExecutionRequest{
```
wxing1292: err2?

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
+		}
+		if context.getExecution().GetRunId() == resp.RunID {
+			release(ErrWorkflowCompleted)
+			return ErrWorkflowCompleted
+		}
+		execution.RunId = common.StringPtr(resp.RunID)
+	}
+	if release != nil {
+		release(ErrMaxAttemptsExceeded)
+	}
```
wxing1292: plz remove this in favor of the above defer

service/history/historyEngine.go
```diff
@@ -2303,13 +2377,36 @@ Update_History_Loop:
 
 		err = context.updateWorkflowExecutionAsActive(e.shard.GetTimeSource().Now())
 		if err == ErrConflict {
```
wxing1292: this err `ErrConflict` is deprecated, please remove all for loop retry for this err

service/history/historyEngine.go
```diff
@@ -59,6 +59,7 @@ import (
 
 const (
 	conditionalRetryCount                     = 5
+	retryCountForContextReload                = 50
```
wxing1292: 5 is enough

service/history/historyEngine.go
```diff
@@ -1608,7 +1609,7 @@ func (e *historyEngineImpl) RequestCancelWorkflowExecution(
 		RunId:      request.WorkflowExecution.RunId,
 	}
 
-	return e.updateWorkflowExecution(ctx, domainID, execution, true,
+	return e.updateWorkflowExecutionWithContextReload(ctx, domainID, execution, true,
```
wxing1292: can we unify updateWorkflowExecutionWithContextReload and updateWorkflowExecutionWithActionAndContextReload?

service/history/historyEngine.go
```diff
@@ -2254,20 +2255,93 @@ type updateWorkflowAction struct {
...
 	domainID string,
 	execution workflow.WorkflowExecution,
-	action func(builder mutableState) (*updateWorkflowAction, error),
+	action updateWorkflowActionFunc,
 ) (retError error) {
 
+	// for execution with runID, just update with action
+	if execution.GetRunId() != "" {
+		workflowContext := e.getWorkflowLockedContext(ctx, domainID, execution)
+		return e.updateWorkflowExecutionWithContextAndAction(workflowContext, action)
```
wxing1292: missing a defer here?

service/history/historyEngine.go
vancexu: ok

service/history/historyEngine.go
vancexu: updated

service/history/historyEngine.go
vancexu: ok

service/history/historyEngine.go
vancexu: yes. 
I actually do had that, but in next commit which includes change APIs getOrCreateWorkflowExecution etc to return `*workflowLockedContext`

service/history/historyEngine.go
vancexu: not really. context lock is released when perform action inside `updateWorkflowExecutionWithContextAndAction`

service/history/historyEngine.go
vancexu: nop, `release` is manually done because of inside for loop

service/history/historyEngine.go
vancexu: ok, updated.

service/history/historyEngine.go
vancexu: Updated, Please double check if I understand you correctly.

service/history/historyEngine.go
vancexu: I can just remove `updateWorkflowExecutionWithContextReload`, but it would cause caller to call something like 
```
updateWorkflowExecutionWithActionAndContextReload(ctx, domainID, execution,
		getUpdateWorkflowActionFunc(createDecisionTask, action))
```

Since we already use things like `updateWorkflowExecution` and `updateWorkflowExecutionWithAction`, I think it's better to have this retry context version `updateWorkflowExecutionWithContextReload` and `updateWorkflowExecutionWithActionAndContextReload`

service/history/historyEngine.go
vancexu: ok

service/history/historyEngine.go
vancexu: I tried to change
```
		err = context.updateWorkflowExecutionAsActive(e.shard.GetTimeSource().Now())
		if err == ErrConflict {
			continue UpdateHistoryLoop
		}
		return err
```
 to 
```
		return context.updateWorkflowExecutionAsActive(e.shard.GetTimeSource().Now())
```
but encountered several unit test failed. how about we change it separately?

service/history/historyEngine.go
wxing1292: we already have too many wrapped version of update workflow execution, can we simplify / unify that?

service/history/historyEngine.go
wxing1292: for this `ErrConflict` sure.

service/history/workflowExecutionUtil.go
```diff
@@ -139,3 +139,21 @@ func terminateWorkflow(
 	)
 	return err
 }
+
+type workflowLockedContext struct {
+	context workflowExecutionContext
+	release releaseWorkflowExecutionFunc
+	err     error
+}
```
wxing1292: please put all struct definition at the very top

service/history/historyEngine.go
```diff
@@ -2254,20 +2254,84 @@ type updateWorkflowAction struct {
 	createDecision bool
 }
 
-func (e *historyEngineImpl) updateWorkflowExecutionWithAction(
+type updateWorkflowActionFunc func(mutableState mutableState) (*updateWorkflowAction, error)
+
+func (e *historyEngineImpl) updateWorkflowExecutionWithActionAndContextReload(
```
wxing1292: this function is confusing, in one path, it is a pass through; another path, release function is called to handled error cases.
the general idea of the release function is: that function should always be called in a defer function whenever seeing it in the first place, to guard against potential bug / panic

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,24 @@ import (
 	"github.com/uber/cadence/common"
 )
 
+type workflowLockedContext struct {
+	context workflowExecutionContext
+	release releaseWorkflowExecutionFunc
+	err     error
+}
```
wxing1292: if the intention is to combine common used structures related to workflow mutation, then plz also add mutable state within this struct, like: https://github.com/uber/cadence/blob/master/service/history/nDCWorkflow.go#L48

so overall probably this struct should looks like this
```go
type workflowLockedContext struct {
	ctx ctx.Context
	context workflowExecutionContext
	mutableState mutableState
	releaseFn releaseWorkflowExecutionFunc	
}
```

wxing1292: and please do not include the error within this struct

service/history/historyEngine.go
```diff
@@ -2335,16 +2411,24 @@ func (e *historyEngineImpl) updateWorkflowExecution(
...
+	return func(builder mutableState) (*updateWorkflowAction, error) {
+		err := action(builder)
+		if err != nil {
+			return nil, err
+		}
+		postActions := &updateWorkflowAction{
+			createDecision: createDecisionTask,
+		}
+		return postActions, nil
+	}
```
wxing1292: let us get rid of this function and create 2 vars and let caller just return these 2 vars
```
var updateWorkflowWithNewDecision = updateWorkflowAction{
	createDecision: true,
}
var updateWorkflowWithoutDecision = updateWorkflowAction{
	createDecision: false,
}
```

wxing1292: one more thing, plz change *updateWorkflowAction to updateWorkflowAction, i.e. pointer to struct. since there is no case that updateWorkflowAction can be nil

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,24 @@ import (
 	"github.com/uber/cadence/common"
 )
 
+type workflowLockedContext struct {
```
wxing1292: plz name this workflowImpl and create a corresponding interface for it

service/history/historyEngine.go
```diff
@@ -2270,20 +2270,85 @@ type updateWorkflowAction struct {
...
 
+	// for execution with runID, just update with action
+	if execution.GetRunId() != "" {
+		workflowContext := e.getWorkflowLockedContext(ctx, domainID, execution)
+		return e.updateWorkflowExecutionWithContextAndAction(workflowContext, action)
+	}
+
+	// for execution without runID,
+	// reload current run in case continueAsNew caused unexpected workflow already completed error
+	for attempt := 0; attempt < conditionalRetryCount; attempt++ {
```
wxing1292: plz change this function to return the corresponding workflow for simplicity
plz also add another function in history engine which call this logic (getting the workflow) and do corresponding update

service/history/historyEngine.go
```diff
@@ -2265,39 +2265,61 @@ func (e *historyEngineImpl) DeleteExecutionFromVisibility(
 	return e.visibilityMgr.DeleteWorkflowExecution(request) // delete from db
 }
 
-type updateWorkflowAction struct {
-	noop           bool
-	createDecision bool
+func (e *historyEngineImpl) updateWorkflow(
```
vancexu: We can use `updateWorkflow` to replace existing `updateWorkflowExecutionWithAction` and `updateWorkflowExecution`, since we do want to reload if runID is empty. 

I can do a refactor in another PR once this get approved

service/history/workflowExecutionUtil.go
vancexu: Updated

service/history/historyEngine.go
vancexu: Updated

service/history/workflowExecutionUtil.go
vancexu: Done

service/history/historyEngine.go
vancexu: updated

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,74 @@ import (
...
+}
+
+func (w *workflowContextImpl) getRunID() string {
+	return w.getContext().getExecution().GetRunId()
+}
+
+func newWorkflowContext(
+	context workflowExecutionContext,
+	releaseFn releaseWorkflowExecutionFunc,
+	mutableState mutableState) workflowContext {
```
wxing1292:
```go
func newWorkflow(
	context workflowExecutionContext,
	releaseFn releaseWorkflowExecutionFunc,
	mutableState mutableState,
)*workflowImpl{
```

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,74 @@ import (
...
+	releaseFn    releaseWorkflowExecutionFunc
+}
+
+type updateWorkflowAction struct {
+	noop           bool
+	createDecision bool
+}
+
+var (
+	emptyUpdateAction             = updateWorkflowAction{}
```
wxing1292: emptyUpdateAction is the same as updateWorkflowWithoutDecision?

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,74 @@ import (
...
+}
+
+func (w *workflowContextImpl) reloadMutableState() (mutableState, error) {
+	mutableState, err := w.getContext().loadWorkflowExecution()
+	if err != nil {
+		return nil, err
+	}
+	w.mutableState = mutableState
+	return mutableState, nil
+}
```
wxing1292: why do we need such function?

service/history/workflowExecutionUtil.go
```diff
@@ -25,6 +25,74 @@ import (
 	"github.com/uber/cadence/common"
 )
 
+type workflowContext interface {
+	getContext() workflowExecutionContext
+	getMutableState() mutableState
+	reloadMutableState() (mutableState, error)
+	getReleaseFn() releaseWorkflowExecutionFunc
+	getRunID() string
```
wxing1292: maybe also provide `getWorkflowID() string`?

service/history/historyEngine.go
```diff
@@ -2713,9 +2750,69 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+	}
+
+	return newWorkflowContext(context, release, mutableState), nil
+}
+
+func (e *historyEngineImpl) loadCurrentWorkflow(
+	ctx ctx.Context,
+	domainID string,
+	workflowID string,
+	runID string,
```
wxing1292: if loading current workflow, then `runID string` is not needed

service/history/historyEngine.go
```diff
@@ -2713,9 +2750,69 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+	mutableState, err := context.loadWorkflowExecution()
+	if err != nil {
+		release(err)
+		return nil, err
+	}
+
+	return newWorkflowContext(context, release, mutableState), nil
+}
+
+func (e *historyEngineImpl) loadCurrentWorkflow(
```
wxing1292: this function should be called loadTargetWorkflow, not current workflow since run id is provided

service/history/historyEngine.go
```diff
@@ -2713,9 +2750,69 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+				return nil, err
+			}
+
+			if workflowContext.getRunID() == prevRunID {
+				workflowContext.getReleaseFn()(ErrWorkflowCompleted)
+				return nil, ErrWorkflowCompleted
+			}
+		}
+	}
+	return workflowContext, err
```
wxing1292: i would write it is this way
```go
if runID != "" {
	return e.loadWorkflow(ctx, domainID, workflowID, runID)
}

// loop to check & get the current workflow
for attempt := 0; attempt < conditionalRetryCount; attempt++ {
	
	workflowContext, err := e.loadWorkflow(ctx, domainID, workflowID, "")
	if err != nil {
		return nil, err
	}

	if workflowContext.getMutableState().IsWorkflowExecutionRunning() {
		return workflowContext, nil
	}

	// workflow not running, need to check current record
	resp, err := r.shard.GetExecutionManager().GetCurrentExecution(
		&persistence.GetCurrentExecutionRequest{
			DomainID:   domainID,
			WorkflowID: workflowID,
		},
	)
	if err != nil {
		workflowContext.getReleaseFn()(err)
		return nil, err
	}

	if resp.RunID == workflowContext.GetRunID() {
		return workflowContext, nil
	}
	workflowContext.getReleaseFn()(nil)
}
```

wxing1292: we need let caller make the judgement whether to return `ErrWorkflowCompleted` or not, not within this function

service/history/workflowExecutionUtil.go
wxing1292: can we delete this function in the next PR?

service/history/workflowExecutionUtil.go
vancexu: Why impl instead of interface? I think `newXXX` return interface is easier for future change (say change impl in future)

service/history/workflowExecutionUtil.go
vancexu: just for better reading.

service/history/workflowExecutionUtil.go
vancexu: In update functions UpdateHistoryLoop, each iteration requires a `loadWorkflowExecution` to  get new mutableState.  That's why I think we need this reload. 
Let me know if that's something we can also get rid of.

service/history/workflowExecutionUtil.go
vancexu: added

service/history/historyEngine.go
vancexu: I renamed this to `loadWorkflow,` and previous loadWorkflow to `loadWorkflowHelper`, since API most likely won't need to call `loadWorkflowHelper` with the new change.

service/history/historyEngine.go
vancexu: Accepted and updated.

service/history/workflowExecutionUtil.go
wxing1292: no this is just the return value, easier for writing unit test

service/history/workflowExecutionUtil.go
wxing1292: then can you rename it? emptyUpdateAction is misleading

service/history/historyEngine.go
```diff
@@ -2713,9 +2750,80 @@ func (e *historyEngineImpl) ReapplyEvents(
...
 				e.logger.Error("failed to re-apply stale events", tag.Error(err))
-				return nil, &workflow.InternalServiceError{Message: "unable to re-apply stale events"}
+				return emptyUpdateAction, &workflow.InternalServiceError{Message: "unable to re-apply stale events"}
 			}
 
 			return postActions, nil
 		})
 }
+
+func (e *historyEngineImpl) loadWorkflowHelper(
```
wxing1292: sorry but can you plz rename this function, e.g. helper does not mean anything

service/history/workflowExecutionUtil.go
vancexu: alright, updated

service/history/workflowExecutionUtil.go
vancexu: After discussion, I change it back to nil and update post action as pointer.

service/history/historyEngine.go
vancexu: Renamed to loadWorkflowOnce

PR #2751, opened 2019-10-31
Adding a simple command to get version and build info of a Cadence cluster

ntayyab:
This will help in trouble-shooting


ntayyab:
I signed the CLA but looks like somewhere in the world of github my id is duplicated and it needs to dedupe. I would resubmit this after figuring that out

PR #2755, opened 2019-11-01
Fix out of boundary of resend events

yux0:


service/history/nDCActivityReplicator.go
```diff
@@ -202,63 +202,70 @@ func (r *nDCActivityReplicatorImpl) shouldApplySyncActivity(
...
+				var endEventVersion *int64
+				if scheduleID < lastIncomingItem.GetEventID() {
+					// resend is exclusive-exclusive. Get the last event to be the schedule event
+					endEventID = common.Int64Ptr(scheduleID + 1)
+					endVersion, err := incomingVersionHistory.GetEventVersion(scheduleID + 1)
+					if err != nil {
+						return false, err
+					}
+					endEventVersion = common.Int64Ptr(endVersion)
+				}
```
wxing1292: plz remove this section, just resend to the end of history

service/history/nDCActivityReplicator.go
```diff
@@ -202,63 +202,70 @@ func (r *nDCActivityReplicatorImpl) shouldApplySyncActivity(
...
+					runID,
+					common.Int64Ptr(lcaItem.GetEventID()),
+					common.Int64Ptr(lcaItem.GetVersion()),
+					endEventID,
+					endEventVersion,
+				)
+			}
+		} else {
+			// case 2: lcaItem is not the last item of incoming version history nor local version history
+			// indicates incoming version <> local version
```
wxing1292: // case 2: incoming version history & local version history diverges

and plz move all description to the very top, like
```
// case 1. <desc>
// case 2. <desc>

if {
 // case 1
} else {
 // case 2
}

service/history/nDCActivityReplicator.go
yux0: removed

PR #2759, opened 2019-11-01
Add validation for add search attributes

vancexu:
https://github.com/uber/cadence/issues/2726


service/frontend/adminHandler.go
```diff
@@ -139,6 +140,9 @@ func (adh *AdminHandler) AddSearchAttribute(ctx context.Context, request *admin.
 	if len(request.GetSearchAttribute()) == 0 {
 		return &gen.BadRequestError{Message: "SearchAttributes are not provided"}
 	}
+	if err := adh.validateConfigForAdvanceVisibility(); err != nil {
+		return &gen.BadRequestError{Message: fmt.Sprintf("Config related to advancedVisibilityStore is not found")}
```
meiliang86: Technically this is not bad result, but NotSupported operation?
Also the message should be "AdvancedVisibilityStore is not supported for this Cadence Cluster"?

wxing1292: from my understanding bad request is ok, since cluster is not configured to support ES?

PR #2761, opened 2019-11-04
Use different argument for cassandra address in Admin CLI

longquanzheng:
https://github.com/uber/cadence/issues/2663


tools/cli/adminCommands.go
```diff
@@ -263,7 +263,7 @@ func readOneRow(query *gocql.Query) (map[string]interface{}, error) {
 }
 
 func connectToCassandra(c *cli.Context) *gocql.Session {
-	host := getRequiredOption(c, FlagAddress)
+	host := getRequiredOption(c, FlagCassandraAddress)
```
wxing1292: could you also add a `FlagCassandraPort`?

wxing1292: line 270 still use the `FlagPort `

longquanzheng: done

longquanzheng:
related to https://github.com/uber/cadence/issues/2777

PR #2762, opened 2019-11-04
Consistent query version check

andrewjdawson2016:
This diff does the following

- Refactors query state machine to make terminal state more generic
- Adds terminal state of failed to query
- Combine client version checker and feature checker into a single checker interface
- Add consistent query feature check
- Fail queries which are responded from worker which does not support consistent query


service/history/query.go
```diff
@@ -23,131 +23,125 @@
...
-
-	validFailed := queryResult.GetResultType().Equals(shared.QueryResultTypeFailed) &&
-		queryResult.Answer == nil &&
-		queryResult.ErrorMessage != nil
-
-	if !validAnswered && !validFailed {
-		return errQueryResultIsInvalid
+	switch terminationState.queryTerminationType {
+	case queryTerminationTypeCompleted:
+		if terminationState.queryResult == nil || terminationState.failure != nil {
```
vitarb: This changes the error for nil queryResult from errQueryResultIsNil to errTerminationStateInvalid. Is it intentional?

service/history/queryRegistry.go
```diff
@@ -192,6 +199,9 @@ func (r *queryRegistryImpl) getQuery(id string) (query, error) {
 	if q, ok := r.unblocked[id]; ok {
 		return q, nil
 	}
```
vitarb: Is this still needed since we are removing unblockQuery method?

service/history/query.go
```diff
@@ -23,131 +23,125 @@
...
+	errQueryNotInTerminalState = &shared.InternalServiceError{Message: "query not in terminal state"}
 )
 
 type (
-	queryState int
+	queryTerminationType int
 
 	query interface {
-		getQueryInternalState() *queryInternalState
+		getQueryID() string
```
meiliang86: This interface does not look very useful to me. You might want to just use the impl directly.

service/history/queryRegistry.go
```diff
@@ -40,14 +40,16 @@ type (
 		getCompletedIDs() []string
 		hasUnblockedQuery() bool
 		getUnblockedIDs() []string
+		hasFailedQuery() bool
```
meiliang86: There are quite a few methods in this interface that are for testing purpose only. Can we remove them and only leave the "real" ones there?
For testing you can use the assertion or test against the impl directly.

service/history/decisionHandler.go
```diff
@@ -637,12 +620,65 @@ func (handler *decisionHandlerImpl) createRecordDecisionTaskStartedResponse(
...
 		}
-		queries[state.id] = state.queryInput
+		queries[id] = input
 	}
 	response.Queries = queries
 	return response, nil
 }
+
+func (handler *decisionHandlerImpl) handleBufferedQueries(
+	msBuilder mutableState,
```
meiliang86: msBuilder -> mutableState
Same from the caller side.

service/history/decisionHandler.go
```diff
@@ -637,12 +620,65 @@ func (handler *decisionHandlerImpl) createRecordDecisionTaskStartedResponse(
...
+	clientImpl string,
+	clientFeatureVersion string,
+	queryResults map[string]*workflow.WorkflowQueryResult,
+	createNewDecisionTask bool,
+) {
+	queryRegistry := msBuilder.GetQueryRegistry()
+	if !queryRegistry.hasBufferedQuery() {
```
meiliang86: What can lead to this?

common/client/versionChecker.go
```diff
@@ -0,0 +1,165 @@
...
+	GoWorkerConsistentQueryVersion = "1.5.0"
+
+	stickyQuery     = "sticky-query"
+	consistentQuery = "consistent-query"
+)
+
+type (
+	// VersionChecker is used to check client/server compatibility and client's capabilities
+	VersionChecker interface {
+		ClientSupported(context.Context, bool) error
```
meiliang86: In general, can you use named params in the interface method so people know what these variables mean? (e.g. this bool...)

service/history/decisionHandler.go
```diff
@@ -637,12 +620,65 @@ func (handler *decisionHandlerImpl) createRecordDecisionTaskStartedResponse(
...
+		return
+	}
+	if versionErr := client.NewVersionChecker().SupportsConsistentQuery(clientImpl, clientFeatureVersion); versionErr != nil {
```
meiliang86: We don't need to create version checker on each call, right?

meiliang86: Also add a comment above this line to explain why we are doing this.

service/history/historyEngine.go
```diff
@@ -849,12 +835,18 @@ func (e *historyEngineImpl) QueryWorkflow(
 			default:
 				return nil, ErrQueryEnteredInvalidState
 			}
+		case queryTerminationTypeUnblocked:
+			msResp, err := e.getMutableState(ctx, request.GetDomainUUID(), *request.GetRequest().GetExecution())
```
meiliang86: Why do we need to load ms again?

common/client/versionChecker.go
```diff
@@ -0,0 +1,165 @@
...
+
+// SupportsConsistentQuery returns error if consistent query is not supported otherwise nil.
+// In case client version lookup fails assume the client does not support feature.
+func (vc *versionChecker) SupportsConsistentQuery(clientImpl string, clientFeatureVersion string) error {
+	return vc.featureSupported(clientImpl, clientFeatureVersion, consistentQuery)
+}
+
+func (vc *versionChecker) featureSupported(clientImpl string, clientFeatureVersion string, feature string) error {
+	if clientImpl == "" || clientFeatureVersion == "" {
+		return &shared.ClientVersionNotSupportedError{ClientImpl: clientImpl, FeatureVersion: clientFeatureVersion}
```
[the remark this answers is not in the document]
andrewjdawson2016: I don't think so for two reasons:

- The customer will get this error in response to a feature usage. For example they will call query and then get this error back so its clear that the error has to do with query.
- Second this error does not always relate to features, it can also just mean the client is too new for this server in which case there is no specific feature.

common/client/versionChecker.go
```diff
@@ -0,0 +1,165 @@
...
+	supportedVersions, ok := implMap[feature]
+	if !ok {
+		return &shared.ClientVersionNotSupportedError{ClientImpl: clientImpl, FeatureVersion: clientFeatureVersion}
+	}
+	version, err := version.NewVersion(clientFeatureVersion)
+	if err != nil {
+		return &shared.ClientVersionNotSupportedError{FeatureVersion: clientFeatureVersion, ClientImpl: clientImpl, SupportedVersions: supportedVersions.String()}
+	}
+	if !supportedVersions.Check(version) {
+		return &shared.ClientVersionNotSupportedError{ClientImpl: clientImpl, FeatureVersion: clientFeatureVersion, SupportedVersions: supportedVersions.String()}
```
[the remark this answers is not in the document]
andrewjdawson2016: Response above.

service/history/queryRegistry.go
```diff
@@ -64,122 +67,126 @@ func newQueryRegistry() queryRegistry {
...
 	r.Lock()
 	defer r.Unlock()
-
 	delete(r.buffered, id)
 	delete(r.completed, id)
 	delete(r.unblocked, id)
+	delete(r.failed, id)
 }
 
 func (r *queryRegistryImpl) getQuery(id string) (query, error) {
```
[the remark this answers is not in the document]
andrewjdawson2016: resolved

service/history/queryRegistry.go
andrewjdawson2016: It is still needed. Even though we removed unblockQuery method we are not removing the ability to unblock a query. Basically now query state machine just allows posting a terminal state. One of those terminal states is unblocked query.

andrewjdawson2016: It is still needed. Even though we removed unblockQuery method we are not removing the ability to unblock a query. Basically now query state machine just allows posting a terminal state. One of those terminal states is unblocked query.

service/history/queryRegistry.go
andrewjdawson2016: I don't feel that strongly but I did think about this and I prefer to leave it like this for the following reasons

- This is a registry API, so even though many of these APIs are not used they are reasonable methods to have on this primitive. 
- I find it confusing to read the interface if you can get completed queries and buffered queries only and yet the user needs to know that queries can also be failed, or unblocked.

service/history/query.go
andrewjdawson2016: I use an interface here because I want to make it clear to the reader that certain invariants of query are always return I cannot clearly indicate this if the rest of the system is working directly on references to query structs.

PR #2772, opened 2019-11-06
Add security protection for admin add search attributes API

vancexu:
Will make CLI change in another PR


idl/github.com/uber/cadence/admin.thrift
```diff
@@ -150,4 +150,5 @@ struct GetWorkflowExecutionRawHistoryV2Response {
 
 struct AddSearchAttributeRequest {
   10: optional map<string, shared.IndexedValueType> searchAttribute
+  20: optional string SecurityToken
```
wxing1292: securityToken, following the convention

vancexu: good catch, updated.

PR #2775, opened 2019-11-06
Visibility Archival logger tags and dynamic configs

yycptt:
- Add logger tags for visibility archival query
- Add dynamic config for max visibility archival query page size
- Use long poll timeout for `ListArchivedWorkflowExecution` API

- Add dynamic config for max visibility archival query RPS (used by internal impl)
- Add dynamic config for max visibility archival query range (used by internal impl)


client/frontend/client.go
```diff
@@ -152,7 +152,7 @@ func (c *clientImpl) ListArchivedWorkflowExecutions(
 	if err != nil {
 		return nil, err
 	}
-	ctx, cancel := c.createContext(ctx)
+	ctx, cancel := c.createLongPollContext(ctx)
```
yycptt: We need to think more about this. Long poll context is only three minutes, the query may take longer time.

andrewjdawson2016: Now that we enforce number of days you can query how long could a query really take?

yycptt: In most cases it shouldn't take very long. But it depends on the cluster status. I experienced the case where a query that fetches less than 1000 records takes around 2 mins.

We also need to monitor the query latency and adjust the max allowed query range.

PR #2782, opened 2019-11-06
DescribeWorkflowExecution should return whether a workflow is running…

anish531213:
… or not


service/history/historyEngine.go
```diff
@@ -1155,6 +1155,14 @@ func (e *historyEngineImpl) DescribeWorkflowExecution(
 		result.WorkflowExecutionInfo.CloseTime = common.Int64Ptr(completionEvent.GetTimestamp())
 	}
 
+	switch executionInfo.State {
+	case persistence.WorkflowStateRunning,
```
yux0: should WorkflowStateCreated alsos consider as open?

anish531213: I don't know about that. @wxing1292, @longquanzheng

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -368,6 +368,7 @@ struct WorkflowExecutionInfo {
   100: optional Memo memo
   101: optional SearchAttributes searchAttributes
   110: optional ResetPoints autoResetPoints
+  120: optional bool IsWorkflowOpen
```
longquanzheng: There are many other places that creating this WorkflowExecutionInfo struct. Make sure you change them all so that it won't be an issue when someone else needs to rely on this field.

service/history/historyEngine.go
longquanzheng: +1 created should be open as well

longquanzheng: Oh your code is correct. @yux do you have concern about considering created as open? From customer's perspective they need it open to make sure the workflow is not closed. 

On the other hand, I think it will be useful if we could expose the close statuses: terminated/failed/canceled/timeouted. I think we can just do it together.

anish531213: I think we are exposing the close statuses if exectutionInfo.State is WorkflowStateCompleted.

yycptt: @longquanzheng  Close status is already exposed. But we don't have a close status for open workflows so that field will be empty in that case. I think the point of this PR is to add a new field so that user doesn't need to determine if the workflow is running or not by checking if the closeStatus is nil.  

So I think the value for `isWorkflowRunning` can simply determined by `executionInfo.State == persistence.WorkflowStateCompleted`

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -368,6 +368,7 @@ struct WorkflowExecutionInfo {
   100: optional Memo memo
   101: optional SearchAttributes searchAttributes
   110: optional ResetPoints autoResetPoints
+  120: optional bool isWorkflowOpen
```
vancexu: Is there any customer asking for this? Can they just use `CloseStatus == Nil` for this?

I don't think we should add this field here, because:
first it is little bit redundant.
second WorkflowExecutionInfo is also used in visibility APIs (such as ListWorkflowExecutions), this change will be cause ListOpenWorkflows return workflow with `isWorkflowOpen` be false, which is confusing

vancexu: If possible, I would just ask customer to use `CloseStatus == Nil`

If not, I would add describe open workflow flag in DescribeWorkflowExecutionRequests to return empty resp if workflow already closed.

service/history/historyEngine.go
yux0: Missed the next line. It makes sense that WorkflowStateCreated is workflow open.

wxing1292:
make sure to also set the isWorkflowOpen field in other places

common/archiver/filestore/visibilityArchiver.go
```diff
@@ -340,5 +340,6 @@ func convertToExecutionInfo(record *visibilityRecord) *shared.WorkflowExecutionI
 		SearchAttributes: &shared.SearchAttributes{
 			IndexedFields: archiver.ConvertSearchAttrToBytes(record.SearchAttributes),
 		},
+		IsWorkflowOpen: common.BoolPtr(record.CloseStatus.Ptr() == nil),
```
yycptt: I don't think you need this check. Visibility records are archived after retention period so the workflow must be closed.

anish531213: removed.

PR #2794, opened 2019-11-07
Fix workflow status check on start of transaction

yux0:


wxing1292:
let us keep the workflow status check and discard the activity version thing for now.

PR #2795, opened 2019-11-07
New persistence client bean for DB clients management

wxing1292:


common/persistence/client/bean.go
```diff
@@ -0,0 +1,296 @@
...
+		historyManager          persistence.HistoryManager
+		executionManagerFactory persistence.ExecutionManagerFactory
+
+		sync.RWMutex
+		shardIDToExecutionManager map[int]persistence.ExecutionManager
+	}
+)
+
+// NewBeanByFactory crate a new store bean using factory
+func NewBeanByFactory(
```
longquanzheng: How about NewBeanFromFactory?

common/persistence/client/factory.go
```diff
@@ -44,8 +44,8 @@ type (
 		NewTaskManager() (p.TaskManager, error)
 		// NewShardManager returns a new shard manager
 		NewShardManager() (p.ShardManager, error)
-		// NewHistoryV2Manager returns a new historyV2 manager
-		NewHistoryV2Manager() (p.HistoryManager, error)
+		// NewHistoryManager returns a new historyV2 manager
```
longquanzheng: could you also fix the comment

common/persistence/client/bean.go
```diff
@@ -0,0 +1,296 @@
...
+
+import (
+	"sync"
+
+	"github.com/uber/cadence/common/persistence"
+)
+
+type (
+	// Bean in an collection of persistence manager
+	Bean interface {
```
longquanzheng: Is it being used anywhere in this PR? or just for future?

wxing1292: i have a really large commit doing service startup logic refactoring, this is part of that

common/persistence/client/factory.go
wxing1292: sure

common/persistence/client/bean.go
wxing1292: sure

common/persistence/client/factory.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package persistence
+package client
```
yycptt: Can we rename the `New()` function in this file to maybe `NewFactory()`? I think `client.New()` returning a persistence factory a little bit counter-intuitive.

wxing1292: done

PR #2803, opened 2019-11-08
Introduce resource package for dependence management

wxing1292:
* New resource implementation & test for better resource management
* Deprecate mockery generated RPC client bean


common/resource/resource.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/uber/cadence/common/membership"
+	"github.com/uber/cadence/common/messaging"
+	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/persistence"
+	persistenceClient "github.com/uber/cadence/common/persistence/client"
+)
+
+type (
+	// Resource is the interface which expose common resources
+	Resource interface {
```
meiliang86: The name is a little bit weird to me. But I don't have good suggestions either. Maybe you can ask others' opinion.

wxing1292: @longquanzheng @yux0 @yycptt @andrewjdawson2016 @BowenXu ?

wxing1292: how about this, in the very end, this `resource` struct is going to replace the existing `service` struct.
when that happen, i will just rename resource to service

PR #2806, opened 2019-11-08
Record last activity failure reason and detail when timeout

longquanzheng:
For https://github.com/uber/cadence/issues/2627
tested locally:
(I have to use adm show because need to update IDL in client)
```
{"eventId":7,"timestamp":1573261540891863000,"eventType":"ActivityTaskTimedOut","version":-24,"taskId":173015064,"activityTaskTimedOutEventAttributes":{"scheduledEventId":5,"startedEventId":6,"timeoutType":"START_TO_CLOSE","lastFailuireReason":"cadenceInternal:Generic","lastFailuireDetails":"bG9uZ2VyIGVycm9y"}}
```


service/history/historyBuilder.go
```diff
@@ -156,8 +156,10 @@ func (b *historyBuilder) AddActivityTaskFailedEvent(scheduleEventID, startedEven
 }
 
 func (b *historyBuilder) AddActivityTaskTimedOutEvent(scheduleEventID, startedEventID int64,
-	timeoutType workflow.TimeoutType, lastHeartBeatDetails []byte) *workflow.HistoryEvent {
-	event := b.newActivityTaskTimedOutEvent(scheduleEventID, startedEventID, timeoutType, lastHeartBeatDetails)
+	timeoutType workflow.TimeoutType, lastHeartBeatDetails []byte, lastFailureReason string,
+	lastFailureDetail []byte) *workflow.HistoryEvent {
```
meiliang86: Can you format as the following:
```
AddActivityTaskTimedOutEvent(
        scheduleEventID int64, 
        startedEventID int64,
        timeoutType workflow.TimeoutType,
        lastHeartBeatDetails []byte,
        lastFailureReason string,
        lastFailureDetail []byte) *workflow.HistoryEvent {
```

meiliang86: Same below.

longquanzheng: Sure

PR #2807, opened 2019-11-09
Adding a counter when RespondDecisionTaskCompleted returns with empty…

vitarb:
Adding a metric for RespondDecisionTaskCompleted when empty decision list is completed.


common/metrics/defs.go
```diff
@@ -1585,6 +1585,7 @@ const (
 	ConsistentQueryTimeoutCount
 
 	NumHistoryMetrics
+	DecisionTaskCompletedWithEmptyDecisionsCount
```
wxing1292: plz change this to line 1477, just after `MultipleCompletionDecisionsCounter` and rename it `EmptyCompletionDecisionsCounter `

wxing1292: could you plz also add the corresponding mapping at after https://github.com/uber/cadence/blob/a4561ec56beab9f4004327b92766b37a5ef66064/common/metrics/defs.go#L1789

vitarb: Done, however the list is not ordered alphabetically anyways.

PR #2811, opened 2019-11-11
Use new resource struct for matching / worker

wxing1292:
* Use new resource struct for matching initialization logic
* Use new resource struct for worker initialization logic
* Update worker scanner initialization logic
* Rewrite ringpop initialization logic to get rid of circular dependency


cmd/server/server.go
```diff
@@ -211,9 +211,12 @@ func (s *server) startService() common.Daemon {
 	case historyService:
 		daemon = history.NewService(&params)
```
yux0: What about history and frontend

service/matching/handler.go
```diff
@@ -37,38 +37,53 @@ import (
...
-	taskPersistence persistence.TaskManager
-	metadataMgr     persistence.MetadataManager
-	engine          Engine
-	config          *Config
-	metricsClient   metrics.Client
-	startWG         sync.WaitGroup
-	domainCache     cache.DomainCache
-	rateLimiter     quotas.Limiter
-	service.Service
+	resource.Resource
```
yux0: Is it necessary to keep the resource in the handler?

wxing1292: line 95 need to use it for rpc handler registration

service/matching/handler.go
```diff
@@ -37,38 +37,53 @@ import (
...
+
+	taskMgr       persistence.TaskManager
```
meiliang86: You don't need to keep taskMgr, metadataMgr, domainCache, config here.

service/matching/service.go
```diff
@@ -22,76 +22,71 @@ package matching
...
+	"github.com/uber/cadence/common/resource"
 	"github.com/uber/cadence/common/service"
 	"github.com/uber/cadence/common/service/dynamicconfig"
 )
 
 // Service represents the cadence-matching service
 type Service struct {
-	stopC  chan struct{}
-	params *service.BootstrapParams
+	resource.Resource
```
meiliang86: is ServiceCommon/ServiceBase a better name?

service/matching/service.go
```diff
@@ -22,76 +22,71 @@ package matching
...
+func NewService(
+	params *service.BootstrapParams,
+) (resource.Resource, error) {
+
+	serviceConfig := NewConfig(dynamicconfig.NewCollection(params.DynamicConfig, params.Logger))
+	serviceResource, err := resource.New(
+		params,
+		common.MatchingServiceName,
+		serviceConfig.ThrottledLogRPS,
+		func(
```
meiliang86: Do you really need this func? Seems like all you are doing is calling persistenceBean.GetVisibilityManager()

wxing1292: frontend / history will have some overrides

service/matching/service.go
wxing1292: i will change this resource thing to service after the moving everything, then delete service, then renaming it to service

service/matching/handler.go
wxing1292: done

service/matching/handler.go
wxing1292: making every component (high level) having access to resource is the idea here.

PR #2822, opened 2019-11-13
CLI improve list command usage hint on timestamp

vancexu:
https://github.com/uber/cadence/issues/2821

### DIFF
--- a/tools/cli/flags.go
+++ b/tools/cli/flags.go
@@ -402,11 +402,11 @@ func getFlagsForListAll() []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:  FlagEarliestTimeWithAlias,
-			Usage: "EarliestTime of start time, supported formats are '2006-01-02T15:04:05Z07:00' and raw UnixNano",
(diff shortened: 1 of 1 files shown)


tools/cli/flags.go
```diff
@@ -402,11 +402,11 @@ func getFlagsForListAll() []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:  FlagEarliestTimeWithAlias,
-			Usage: "EarliestTime of start time, supported formats are '2006-01-02T15:04:05Z07:00' and raw UnixNano",
```
yycptt: I think this is the time format defined by the `time` package? I understand that when actually writing the time it should be either `...Z` or `...+/-07:00`, otherwise it can't be parsed. If changed to `+07:00`, will it make users think `...Z` is not supported?

vancexu: True, that's why we originally put it there. 
But with several similar customer complains, I think most people believe `2006-01-02T15:04:05Z07:00` is misleading format name, and if they try `2006-01-02T15:04:05+07:00` and success, I think they will try `2006-01-02T15:04:05Z`

PR #2823, opened 2019-11-13
Fix create workflow error handling issue

wxing1292:
* Fix persistence layer create zombie workflow dedup error handling
* Fix nDC transaction manager creating zombie workflow error handling
* NDC integration test will duplicate events for better testing
* Improve replication fetcher test reliability
* Fix error message in MySQL
* Add more log to NDC history replicator


service/history/nDCTransactionMgrForNewWorkflow.go
```diff
@@ -239,14 +239,23 @@ func (r *nDCTransactionMgrForNewWorkflowImpl) createAsZombie(
...
 		targetWorkflowHistorySize,
 		now,
 		createMode,
 		prevRunID,
 		prevLastWriteVersion,
 	)
+	switch err.(type) {
+	case nil:
+		return nil
+	case *persistence.WorkflowExecutionAlreadyStartedError:
```
yux0: should createAsCurrent need to handle this error?

yux0: Synced offline. CurrentAsCurrent will not go into this

PR #2824, opened 2019-11-13
Update archival development guide for visibility archiver

yycptt:


common/archiver/README.md
```diff
@@ -28,7 +28,8 @@ type HistoryArchiver interface {
 	// the resource that histories should be archived into. The implementor gets to determine how to interpret the URI.
 	// The Archive method may or may not be automatically retried by the caller. The ArchiveOptions are used
 	// to interact with these retries including giving the implementor the ability to cancel retries and record progress
-	// between retry attempts. 
+  // between retry attempts. 
+  // This method will be invoked after the workflow retention period.
```
meiliang86: after the workflow passes its retention period

common/archiver/README.md
```diff
@@ -45,13 +46,22 @@ type HistoryArchiver interface {
 
 ```go
 type VisibilityArchiver interface {
+    // Archive is used to archive one workflow visibility record. Check the Archive() method of the HistoryArchiver interface for parameters' meaning and requirements for this method.
```
meiliang86: Does go doc support link to code?

common/archiver/README.md
```diff
@@ -45,13 +46,22 @@ type HistoryArchiver interface {
 
 ```go
 type VisibilityArchiver interface {
+    // Archive is used to archive one workflow visibility record. Check the Archive() method of the HistoryArchiver interface for parameters' meaning and requirements for this method. 
+    // The only difference is that the ArchiveOption parameter won't include an option for recording process. 
+    // Please make sure your implementation is lossless. If any in-memory batching mechanism is used, then those batched records will be lost during server restarts.
```
meiliang86: Does it mean that the implementation should never do batching?

common/archiver/README.md
```diff
@@ -45,13 +46,22 @@ type HistoryArchiver interface {
 
 ```go
 type VisibilityArchiver interface {
+    // Archive is used to archive one workflow visibility record. Check the Archive() method of the HistoryArchiver interface for parameters' meaning and requirements for this method. 
+    // The only difference is that the ArchiveOption parameter won't include an option for recording process. 
+    // Please make sure your implementation is lossless. If any in-memory batching mechanism is used, then those batched records will be lost during server restarts. 
+    // This method will be invoked when workflow closes. This means if conflict resolution happens after a workflow closes, this method will be called more than once for the same workflow execution.
```
meiliang86: // This method will be invoked when workflow closes. Note that because of conflict resolution, it is possible for a workflow to through the closing process multiple times, which means that this method can be invoked more than once after a workflow closes.

common/archiver/README.md
yycptt: It depends on how the batching is implemented. If something like our indexer is implemented, then it's fine. So I think the keyword here is in-memory.

common/archiver/README.md
yycptt: Oh, I am referring to the method and its comment starting from L26 in this file, not the actual code. I will change the wording. Sorry for the confusion.

PR #2825, opened 2019-11-14
Matching query refactor

andrewjdawson2016:
This diff does three things

- **Version check on consistent query on direct dispatch through matching:** Supporting consistent query requires an up to date workflow worker. We were doing checks on this workflow worker's version on query results from RespondDecisionTaskCompleted but we were not doing them on RespondQueryTaskCompleted. This diff adds this second check.
- **Move first decision wait logic for query to history:** A query cannot be run until the first decision task has been handled by the worker. This has always been the behavior of query. But currently if query occurs before first decision task then the query just times out with no useful error message. This diff introduces a more useful error message. It also moves these first decision task checks form matching to history. Logically they belong in history the only reason they were ever in matching is because query didn't use to go through history. Query will wait for up to one second for the first decision task to complete, if there is no first decision task completed within one second of calling query user will get describe error back.  The one second delay is to support the use case of starting a workflow and then querying immediately. In the old logic we waited up to ctx timeout for a first decision but doing this is useless because then even if first decision occurs there will be no time to respond to get query result and respond back.
- **Only return WorkflowQueryFailed when user query handler was run and failed:** Currently we return WorkflowQueryFailed error all over the place for random errors (for example timeouts are converted to WorkflowQueryFailed and failure to access mutable state are converted to this error). I think this makes this error more or less meaningless to users. I believe WorkflowQueryFailed should only be returned if user query handler was run and it returned an error. All other errors should be things like context timeout, InternalServiceError or BadRequestError. This matches the behavior of our other APIs.


service/history/historyEngine.go
```diff
@@ -745,36 +749,57 @@ func (e *historyEngineImpl) QueryWorkflow(
...
 	de, err := e.shard.GetDomainCache().GetDomainByID(request.GetDomainUUID())
 	if err != nil {
 		return nil, err
 	}
 
+	context, release, err := e.historyCache.getOrCreateWorkflowExecution(ctx, request.GetDomainUUID(), *request.GetRequest().GetExecution())
+	if err != nil {
+		return nil, err
+	}
+	defer func() { release(retErr) }()
```
longquanzheng: I think we can always release(nil) because whatever error returned, it won't mutate the workflow because this is just query.

PR #2828, opened 2019-11-14
Support Postgres in SQL CLI and make SQL pluggable for any database 

longquanzheng:
1. Create postgres version of schema, mostly follow this: https://en.wikibooks.org/wiki/Converting_MySQL_to_PostgreSQL

2. Create sql tool for postgres. While doing this, I move mysql in a different module so that both mysql/postgres are plugins for this CLI


tools/sql/conn.go
```diff
@@ -40,90 +39,99 @@ type (
...
-		`PRIMARY KEY (year, month, update_time));`
+	// Driver is the driver interface that each SQL database needs to implement
+	Driver interface {
+		GetDriverName() string
+		CreateDBConnection(driverName, host string, port int, user string, passwd string, database string) (*sqlx.DB, error)
+		GetReadSchemaVersionSQL() string
+		GetWriteSchemaVersionSQL() string
+		GetWriteSchemaUpdateHistorySQL() string
+		GetCreateSchemaVersionTableSQL() string
+		GetCreateSchemaUpdateHistoryTableSQL() string
```
venkat1109: [ NOT A BLOCKER ] Don't like the fact that there is an implicit assumption about the parameters of the SQL statement for the caller of this API. I think a ConnectionProvider interface and a ConnectionFactory as opposed to this driver interface would be more flexible & cleaner. However,  I understand it's unlikely that parameters would differ between SQL engines and this is something we can change later when we have a need for flexibility. So, up to you

tools/sql-extensions/postgres/driver.go
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
venkat1109: 2019

tools/sql-extensions/mysql/driver.go
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
venkat1109: 2019

tools/sql/conn.go
longquanzheng: yeah, let's keep that in mind. It's also easier to manage that we give less flexibility at the beginning. It's also possible that people go to implement https://github.com/uber/cadence/blob/master/tools/common/schema/types.go#L52 this interface which will have most flexibility.

schema/postgres/cadence/schema.sql
```diff
@@ -0,0 +1,255 @@
+CREATE TABLE domains(
+  shard_id INTEGER NOT NULL DEFAULT 54321,
+  id BYTEA NOT NULL,
+  name VARCHAR(255) UNIQUE NOT NULL,
+  --
+	data BYTEA NOT NULL,
+	data_encoding VARCHAR(16) NOT NULL,
+	is_global SMALLINT NOT NULL,
```
wxing1292: indentation?

schema/postgres/cadence/versioned/v0.1/base.sql
```diff
@@ -0,0 +1,245 @@
+CREATE TABLE domains(
+  shard_id INTEGER NOT NULL DEFAULT 54321,
+  id BYTEA NOT NULL,
+  name VARCHAR(255) UNIQUE NOT NULL,
+  --
+	data BYTEA NOT NULL,
+	data_encoding VARCHAR(16) NOT NULL,
+	is_global SMALLINT NOT NULL,
+  PRIMARY KEY(shard_id, id)
```
wxing1292: indentation plz

schema/postgres/cadence/versioned/v0.3/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
```
wxing1292: where is v0.2?

PR #2830, opened 2019-11-15
Adding auto-reset points to visibility record

vitarb:
This change:
* Introduces binary checksums into ElasticSearch by saving them in the mutableStateBuilder into the search attributes and adding a corresponding transfer task to upsert search attributes.
* Adds BinaryChecksums as a new indexed field in the elasticsearch.
* Whitelists this new property on the client side.

### DIFF
--- a/service/history/mutableStateBuilder.go
+++ b/service/history/mutableStateBuilder.go
@@ -202,7 +202,6 @@ func newMutableStateBuilder(
 		State:              persistence.WorkflowStateCreated,
 		CloseStatus:        persistence.WorkflowCloseStatusNone,
 		LastProcessedEvent: common.EmptyEventID,
-		SearchAttributes:   make(map[string][]byte),
 	}
 	s.hBuilder = newHistoryBuilder(s, logger)
 	s.taskGenerator = newMutableStateTaskGenerator(shard.GetDomainCache(), s.logger, s)
@@ -1934,6 +1933,9 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
 	bytes, err := json.Marshal(recentBinaryChecksums)
 	if err != nil {
 		return err
+	}
+	if exeInfo.SearchAttributes == nil {
+		exeInfo.SearchAttributes = make(map[string][]byte)
 	}
 	exeInfo.SearchAttributes[definition.BinaryChecksums] = bytes
 	if e.shard.GetConfig().AdvancedVisibilityWritingMode() != common.AdvancedVisibilityWritingModeOff {
--- a/service/history/mutableStateBuilder.go
+++ b/service/history/mutableStateBuilder.go
@@ -21,6 +21,7 @@
 package history
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -33,6 +34,7 @@ import (
 	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/clock"
 	"github.com/uber/cadence/common/cluster"
+	"github.com/uber/cadence/common/definition"
 	"github.com/uber/cadence/common/errors"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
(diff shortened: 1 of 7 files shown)


service/history/mutableStateBuilder.go
```diff
@@ -21,9 +21,12 @@
 package history
 
 import (
+	"encoding/json"
 	"fmt"
 	"time"
 
+	"github.com/uber/cadence/common/definition"
+
```
longquanzheng: Our team has pretty strict rule for import order: we usually have three groups: buitins, local repo and other reopos. Like this answer: https://stackoverflow.com/questions/38704769/golang-import-grouping-by-package 

Maybe we can try to let "make fmt or "make lint" to do this automatically @wxing1292 ?

service/history/mutableStateBuilder.go
```diff
@@ -1921,6 +1931,15 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
...
 	}
+	bytes, err := json.Marshal(recentBinaryChecksums)
+	if err != nil {
+		e.logger.Error(fmt.Sprintf("unable to serialize recent binary checksums %v", recentBinaryChecksums))
+	}
+	if exeInfo.SearchAttributes == nil {
+		exeInfo.SearchAttributes = make(map[string][]byte)
+	}
+	exeInfo.SearchAttributes[definition.BinaryChecksums] = bytes
+	e.AddTransferTasks(&persistence.UpsertWorkflowSearchAttributesTask{})
```
longquanzheng: Looks like it only works in active cluster in normal case. A few things need to look at:
1. Looks like AddTransferTasks should be called in taskGenerator so that it can obtain the version(for XDC), other wise it won't work with XDC.
2. Make sure that after reset, the search attrs are still correct. I believe you don't have to do anything because it's already the problem bowei has solved. But you need to verify.

service/history/mutableStateBuilder.go
wxing1292: i think make fmt will sort the imports, let me double verify

service/history/mutableStateBuilder.go
wxing1292: plz use the mutable state task generator

docker/docker-compose.yml
```diff
@@ -12,7 +12,9 @@ services:
       - "8125:8125"
       - "8126:8126"
   cadence:
-    image: ubercadence/server:master-auto-setup
+    build:
```
vancexu: remember to revert this manual

service/history/mutableStateBuilder.go
```diff
@@ -1921,6 +1931,15 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
 	exeInfo.AutoResetPoints = &workflow.ResetPoints{
 		Points: currResetPoints,
 	}
+	bytes, err := json.Marshal(recentBinaryChecksums)
```
vancexu: Can we add check: if ES is enabled then upsert binaryChecksums as search attributes?
Use `e.shard.GetConfig().AdvancedVisibilityWritingMode() != common.AdvancedVisibilityWritingModeOff` to check.

service/history/mutableStateBuilder.go
vitarb: Done

docker/docker-compose.yml
vitarb: Done

service/history/mutableStateBuilder.go
vitarb: Done

service/history/mutableStateBuilder.go
vitarb: Good call, done!

service/history/mutableStateBuilder.go
```diff
@@ -1921,6 +1930,18 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
...
+	if err != nil {
+		return err
+	}
+	if exeInfo.SearchAttributes == nil {
```
wxing1292: when will the search attributes be nil?

wxing1292: in general, we should avoid checking nil every in the code
plz add the initialization in the new function as well as load function

Issue #1482, opened 2019-02-21
Add attempts field to ActivityTaskCompleted event


mfateev:
It might be useful to understand why activity took so long to execute.

anish531213:
Hi @mfateev, attempts field is already added to ActivityTaskStarted event. Do we want to:
1. Move that to ActivityTaskCompleted  and other(ActivityTaskFailed, Canceled, Timeout) etc.
2. Keep in both ActivityTaskStarted and 1.
3. Keep just in ActivityTaskStarted.

meiliang86:
I think it should be enough to have this in ActivityTaskStarted. @mfateev you can reopen this if you think otherwise.

PR #2835, opened 2019-11-18
Add attempts field to ActivityTaskCompleted event

anish531213:
https://github.com/uber/cadence/issues/1482


yycptt:
I think we already have the `attempt` in the `ActivityTaskStarted` Event? Not sure if we want to add the same field to the completed event again (those two fields will always have the same value I think).  Probably what we want is move the field from start event to the close event, but then we also need to do this for other events like activity task failed, timed out, cancelled etc.

service/history/mutableStateBuilder.go
```diff
@@ -2212,7 +2213,7 @@ func (e *mutableStateBuilder) AddActivityTaskCompletedEvent(
 	if err := e.addTransientActivityStartedEvent(scheduleEventID); err != nil {
 		return nil, err
 	}
-	event := e.hBuilder.AddActivityTaskCompletedEvent(scheduleEventID, startedEventID, request)
+	event := e.hBuilder.AddActivityTaskCompletedEvent(scheduleEventID, startedEventID, ai.Attempt, request)
```
vitarb: In case of an error above you might not have valid ActivityInfo which could result in a panic.

meiliang86:
See comments in #1482.

PR #2836, opened 2019-11-18
Persistence refactoring to be pluggable for any database

longquanzheng:
This PR is to refactor the persistence layer to make it easier and plugable to support different databases, also refactor MySQL for the new interfaces
More specifically:
1. Add persistence interface for CLI/Testing 
2. Add interface for plugin.factory
3. Move MySQL to the sql-extensions and implement(/refactor to) factory
4. Refactor sql persistence test for MySQL

```
(postgres2)$./cadence-server --env development_postgres start
2019/11/27 16:59:40 Loading config; env=development_postgres,zone=,configDir=./config
2019/11/27 16:59:40 Loading configFiles=[./config/base.yaml ./config/development_postgres.yaml]
2019/11/27 16:59:40 error creating file based dynamic config client, use no-op config client instead. error: error checking dynamic config file at path , error: stat : no such file or directory
{"level":"info","ts":"2019-11-27T16:59:40.141-0800","msg":"Created RPC dispatcher and listening","service":"cadence-frontend","address":"127.0.0.1:7933","logging-call-at":"rpc.go:81"}
```

Some TODOs -- will create other small diffs for them
1. https://github.com/uber/cadence/issues/2863
2. https://github.com/uber/cadence/issues/2856
3. https://github.com/uber/cadence/issues/2854
4. https://github.com/uber/cadence/issues/2869
5. https://github.com/uber/cadence/issues/2870
6. https://github.com/uber/cadence/issues/2877


longquanzheng:
https://github.com/uber/cadence/issues/2536

meiliang86:
@venkat1109 Can you take a look at this as well?

common/persistence/persistence-tests/persistenceTestBase.go
```diff
@@ -1261,6 +1264,12 @@ func (s *TestBase) CompleteTask(domainID, taskList string, taskType int, taskID
 
 // TearDownWorkflowStore to cleanup
 func (s *TestBase) TearDownWorkflowStore() {
+	s.ExecutionMgrFactory.Close()
+	// TODO VisibilityMgr/Store is created with a separated code path, this is incorrect and may cause leaking connection
```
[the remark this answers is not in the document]
longquanzheng: This is an issue for testing in Postgres. Postgres requires all connection to be closed before dropping a database. I will add a comment to it here to make it more clear.

cmd/server/cadence.go
```diff
@@ -25,13 +25,14 @@ import (
 	"os"
 	"strings"
 
+	_ "github.com/uber/cadence/common/persistence/sql-extensions/mysql"    // needed to load mysql extensions
+	_ "github.com/uber/cadence/common/persistence/sql-extensions/postgres" // needed to load postgres extensions
+
 	"github.com/urfave/cli"
```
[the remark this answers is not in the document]
longquanzheng: OK

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -681,21 +688,51 @@ type (
...
 		tableCRUD
+		AdminCURD
+
 		BeginTx() (Tx, error)
 		DriverName() string
+		IsDupEntryError(err error) bool
 		Close() error
 	}
 
+	// DBAndTx is mainly for fixing the Lint warning and also simplify the type assertion
```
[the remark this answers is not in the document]
longquanzheng: Yes two assertion is fine. The actual problem is lint warning as we always want to clear them. So if the return type of Init() is db struct, then the warning will arise because db is a private member. If it is DB interface then it worn't work(compile error) for beginTx()...

I agree that this is a little bit confusing. So here are all the options:
1. Make db struct public, maybe call it MySQLDB/PostgresDB -- but all member fileds are private. This is against of design pattern because we want to hide the implementation details.
2. Returning db, ignore the lint warning. 
3. Suppress lint warning.

I can't find a way to do 3 yet. So maybe 2 is our best option.

common/persistence/sql/storage/sqldb/interfaces.go
```diff
@@ -681,21 +688,51 @@ type (
...
 		Commit() error
 		Rollback() error
 	}
 
-	// Interface defines the API for a SQL database
-	Interface interface {
+	// DB defines the API for SQL operations
+	DB interface {
 		tableCRUD
+		AdminCURD
```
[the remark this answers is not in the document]
longquanzheng: I think this won't work easily. We want a DB interface as a universal interface to perform all db operation. If you separate it, then we need to return and keep two different objects from the driver for CLI/testing which is more confusing.

common/persistence/sql/storage/sqldb/interfaces.go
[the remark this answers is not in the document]
longquanzheng: I have changed it to option 2. Let's see if @venkat1109 have any other opinions.

cmd/server/cadence.go
```diff
@@ -21,20 +21,19 @@
...
 	"log"
 	"os"
 	"os/signal"
 	"strings"
 	"syscall"
 
+	"github.com/urfave/cli"
+
+	"github.com/uber/cadence/common"
+	_ "github.com/uber/cadence/common/persistence/sql/plugins/mysql"    // needed to load mysql extensions
```
venkat1109: fix out of date comment

common/persistence/sql/plugins/interfaces.go
```diff
@@ -681,21 +688,43 @@ type (
 		GetAckLevels(queueType common.QueueType, forUpdate bool) (map[string]int, error)
 	}
 
+	// adminCURD defines admin operations for CLI and test suites
+	adminCURD interface {
```
venkat1109: fix typo - s/CURD/CRUD

common/persistence/sql/plugins/mysql/events.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
venkat1109: leave the copyright year as is since this file contains changes from 2017-till now

common/persistence/sql/plugins/mysql/plugin.go
```diff
@@ -47,12 +51,40 @@ var dsnAttrOverrides = map[string]string{
...
+type plugin struct{}
+
+var _ plugins.Plugin = (*plugin)(nil)
+
+func init() {
+	sql.RegisterPlugin(PluginName, &plugin{})
+}
+
+// InitDB initialize the db object
+func (p *plugin) InitDB(cfg *config.SQL) (plugins.DB, error) {
```
venkat1109: consider renaming these InitFoo* methods to NewFoo* methods - its more consistent with the usage here

longquanzheng:
> Looks good. I have one suggestion though
> 
> How do you feel about renaming the package "plugins" to "sqlplugin" ?
> i.e. common/persistence/sql/plugins -> common/persistence/sql/sqlplugin
> 
> This will make the code on the callers more readable. Currently, referencing plugins.DB, plugings.Tx etc looks weird - sqlplugin.DB / sqlplugin.Tx looks more readable to me.

Agreed. I had the same feeling as well. In case that we will have some other plugins like archival plugin. Thanks for pointing it out

common/persistence/sql/plugins/mysql/plugin.go
longquanzheng: Sounds good

common/persistence/sql/plugins/mysql/events.go
longquanzheng: Got it.

cmd/server/cadence.go
longquanzheng: ok

common/persistence/sql/plugins/interfaces.go
longquanzheng: OK

PR #2841, opened 2019-11-19
Add support for custom MySQL TLS configuration

bjacobso:
This is an attempt to add support for TLS MySQL (we use RDS/Aurora and require TLS support) using the recently added TLS support for Cassandra. In the process it reuses the config in persistence for use in tools and removes ConnectParams. 

Feedback is welcome!

### DIFF
--- a/common/persistence/sql/storage/store.go
+++ b/common/persistence/sql/storage/store.go
@@ -22,15 +22,18 @@ package storage
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"log"
 	"net/url"
+	"strconv"
 	"strings"
 
+	"github.com/go-sql-driver/mysql"
 	"github.com/iancoleman/strcase"
 	"github.com/jmoiron/sqlx"
-
-	"github.com/uber/cadence/common/persistence/sql/storage/mysql"
-	"github.com/uber/cadence/common/persistence/sql/storage/sqldb"
 	"github.com/uber/cadence/common/service/config"
 )
 
@@ -39,6 +42,8 @@ const (
 	isolationLevelAttrName       = "transaction_isolation"
 	isolationLevelAttrNameLegacy = "tx_isolation"
 	defaultIsolationLevel        = "'READ-COMMITTED'"
+	customTLSName                = "tls-custom"
+	defaultSQLPort               = 3306
 )
 
 var dsnAttrOverrides = map[string]string{
@@ -51,7 +56,14 @@ var dsnAttrOverrides = map[string]string{
 // underlying SQL database. The returned object is to tied to a single
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
-func NewSQLDB(cfg *config.SQL) (sqldb.Interface, error) {
+func NewSQLDB(cfg *config.SQL) (*sqlx.DB, error) {
+	if cfg.TLS != nil {
+		err := RegisterTLSConfig(cfg)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	db, err := sqlx.Connect(cfg.DriverName, buildDSN(cfg))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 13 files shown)


common/persistence/sql/storage/store.go
```diff
@@ -67,7 +79,83 @@ func NewSQLDB(cfg *config.SQL) (sqldb.Interface, error) {
...
+		return fmt.Errorf("failed to register tls config: %v", err)
+	}
+
+	if cfg.ConnectAttributes == nil {
+		cfg.ConnectAttributes = map[string]string{}
+	}
+
+	if cfg.ConnectAttributes["tls"] == customTLSName || cfg.ConnectAttributes["tls"] == "" {
+		// Use the new custom tls configuration
+		cfg.ConnectAttributes["tls"] = customTLSName
```
bjacobso: I'm modifying the config here to use the new registered tls config. Not sure if this is best practice or not?

common/persistence/sql/storage/store.go
```diff
@@ -67,7 +79,83 @@ func NewSQLDB(cfg *config.SQL) (sqldb.Interface, error) {
...
 	// Maps struct names in CamelCase to snake without need for db struct tags.
 	db.MapperFunc(strcase.ToSnake)
-	return mysql.NewDB(db, nil), nil
+	return db, nil
+}
+
+func RegisterTLSConfig(cfg *config.SQL) error {
+	host, _, err := getHostPortFromConnectAddr(cfg.ConnectAddr)
+	if err != nil {
+		log.Panicf("error getting host port from ConnectAddr: %v", err)
```
venkat1109: Return an error from here and let the upper layer decide what to do with the error

common/persistence/sql/storage/store.go
venkat1109: sorry, I am missing where this modified connectAttributes is being used ?

common/persistence/sql/storage/store.go
```diff
@@ -67,7 +79,83 @@ func NewSQLDB(cfg *config.SQL) (sqldb.Interface, error) {
...
+		// Use the new custom tls configuration
+		cfg.ConnectAttributes["tls"] = customTLSName
+	} else {
+		return fmt.Errorf("invalid tls config: %v, must be %v", cfg.ConnectAttributes["tls"], customTLSName)
+	}
+
+	return nil
+}
+
+func getHostPortFromConnectAddr(connectAdd string) (*string, *int, error) {
```
venkat1109: I believe you can use net.SplitHostPort(..)

tools/common/schema/types.go
```diff
@@ -138,6 +138,9 @@ const (
 	// CLIFlagQuiet is the cli flag for quiet mode
 	CLIFlagQuiet = CLIOptQuiet + ", q"
 
+	// CLIFlagConnectAttributesTxIsolation controls the transaction isolation version for mysql
+	CLIFlagConnectAttributesTxIsolation = "tx-isolation"
```
venkat1109: I think there are two flavors of this - tx_isolation and transaction_isolation.. one of them is deprecated in the recent mysql version. We should either support both or alternatively, simply accept key-value attributes to be passed along to the dsn. I prefer the latter but I am ok with convenience option for the frequently used ones (like tx_isolation) as long as we support both flavors

tools/sql/handler.go
```diff
@@ -99,7 +73,7 @@ func CheckCompatibleVersion(
 // using the given command line arguments
 // as input
 func setupSchema(cli *cli.Context) error {
-	params, err := parseConnectParams(cli)
+	params, err := parseConnectConfig(cli)
```
venkat1109: after the renaming, params should also be renamed to cfg or config

tools/sql/handler.go
```diff
@@ -177,31 +150,57 @@ func doDropDatabase(p ConnectParams, name string) {
...
 	}
-	return params, nil
+	return cfg, nil
 }
 
-// ValidateConnectParams validates params
-func ValidateConnectParams(params *ConnectParams, isDryRun bool) error {
-	if len(params.Host) == 0 {
+// ValidateConnectConfig validates params
+func ValidateConnectConfig(cfg *config.SQL, isDryRun bool) error {
```
venkat1109: can you also add validation to make sure when TLS is enabled, all required params are also set ?

tools/sql/main.go
```diff
@@ -90,10 +90,35 @@ func BuildCLIOptions() *cli.App {
...
+		cli.StringFlag{
+			Name:  schema.CLIFlagTLSCertFile,
+			Usage: "sql tls client cert path (tls must be enabled)",
+		},
+		cli.StringFlag{
+			Name:  schema.CLIFlagTLSKeyFile,
+			Usage: "sql tls client key path (tls must be enabled)",
+		},
+		cli.StringFlag{
+			Name:  schema.CLIFlagTLSCaFile,
```
venkat1109: s/CLIFlagTLSCaFile/CLIFlagTLSCAFile - abbrs are capitalized

tools/sql-extensions/mysql/handler_test.go
```diff
@@ -46,17 +47,17 @@ func (s *HandlerTestSuite) SetupTest() {
 	s.Assertions = require.New(s.T()) // Have to define our overridden assertions in the test setup. If we did it earlier, s.T() will return nil
 }
 
-func (s *HandlerTestSuite) TestValidateConnectParams() {
-	p := new(sql.ConnectParams)
-	s.NotNil(sql.ValidateConnectParams(p, false))
-	s.NotNil(sql.ValidateConnectParams(p, true))
+func (s *HandlerTestSuite) TestValidateConnectConfig() {
```
venkat1109: please update this test after you add validation for tls params

common/persistence/sql/storage/store.go
```diff
@@ -67,7 +79,83 @@ func NewSQLDB(cfg *config.SQL) (sqldb.Interface, error) {
...
+	}
+
+	if cfg.ConnectAttributes == nil {
+		cfg.ConnectAttributes = map[string]string{}
+	}
+
+	if cfg.ConnectAttributes["tls"] == customTLSName || cfg.ConnectAttributes["tls"] == "" {
+		// Use the new custom tls configuration
+		cfg.ConnectAttributes["tls"] = customTLSName
+	} else {
```
venkat1109: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

tools/sql-extensions/postgres/driver.go
```diff
@@ -83,11 +81,8 @@ func (d *driver) GetDriverName() string {
 	return driverName
 }
 
-func (d *driver) CreateDBConnection(driverName, host string, port int, user string, passwd string, database string) (*sqlx.DB, error) {
-	if database == "" {
-		database = "postgres"
-	}
-	db, err := sqlx.Connect(driverName, fmt.Sprintf(dataSourceNamePostgres, user, passwd, host, port, database))
+func (d *driver) CreateDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	db, err := storage.NewSQLDB(cfg)
```
longquanzheng: This will not work unless you do it after my PR: https://github.com/uber/cadence/pull/2836 
Also need to keep the logic to use "postgres" when database is empty to act as admin for creating/dropping database for postgres.

longquanzheng:
@bjacobso I think your change is easier if you can do it after I did this refactoring: https://github.com/uber/cadence/pull/2836

bjacobso:
@venkat1109 Thanks for the feedback! @longquanzheng I will wait until your PR lands to continue. 

Overall though does the direction make sense? We really need TLS in order to put Cadence in production.

venkat1109:
> Overall though does the direction make sense? 
The direction certainly looks reasonable to me. One thing that I don't like is the fact that the config takes a path to keyFile. Ideally, this should be some sort of TLSCertProvider interface which behind the scenes has many implementations one of which could be a keyFile stored on disk. But for most production use cases, the keyFile will likely be stored on a KeyServer and the implementation will fetch it and store it only in memory (never on disk). This is orthogonal to what you are trying to achieve though and should be a separate task item.

bjacobso:
@longquanzheng I see you have made good progress on your refactor. Should I wait for it to land in master or do you think it is in a stable spot for me to start making my changes?

longquanzheng:
> @longquanzheng I see you have made good progress on your refactor. Should I wait for it to land in master or do you think it is in a stable spot for me to start making my changes?

It should be in a good shape enough for just start working on top of it. But it's probably even easier for you if you wait for landing into master -- it shouldn't take a long time. As tomorrow is holiday, I think @venkat1109  will review my PR next week.

longquanzheng:
> @longquanzheng I see you have made good progress on your refactor. Should I wait for it to land in master or do you think it is in a stable spot for me to start making my changes?

Hi @bjacobso thanks for waiting. My PR is landed. It should be much easier for you to do it this.

bjacobso:
Closing in favor of https://github.com/uber/cadence/pull/2913

PR #2842, opened 2019-11-19
Dynamic config limits for consistent query

andrewjdawson2016:


service/history/decisionHandler.go
```diff
@@ -654,24 +662,66 @@ func (handler *decisionHandlerImpl) handleBufferedQueries(
...
+			if err := queryRegistry.setTerminationState(id, failedTerminationState); err != nil {
+				handler.logger.Error(
+					"failed to set query termination state to failed",
+					tag.WorkflowDomainName(domain),
+					tag.WorkflowID(workflowID),
+					tag.WorkflowRunID(runID),
+					tag.QueryID(id),
+					tag.Error(err))
+				scope.IncCounter(metrics.FailQueryFailedCount)
+			}
```
yycptt: nit: `continue` here and we don't need the `else` below? For me it's a little bit more readable.

service/history/decisionHandler_test.go
```diff
@@ -23,6 +23,12 @@ package history
 import (
 	"testing"
 
+	"github.com/uber-go/tally"
+
+	"github.com/uber/cadence/common/log/loggerimpl"
```
yycptt: nit: run `make fmt`?

PR #2843, opened 2019-11-19
Mitigation for timer double fire issues

yux0:


service/history/decisionHandler.go
```diff
@@ -237,6 +238,13 @@ func (handler *decisionHandlerImpl) handleDecisionTaskFailed(
 		RunId:      common.StringPtr(token.RunID),
 	}
 
+	if request.GetCause().Equals(workflow.DecisionTaskFailedCauseWorkflowWorkerUnhandledFailure) {
+		handler.logger.Info("Timer double fire issue",
```
venkat1109: make the log info timer_double_fire_issue - just so that it's easily searchable. Also, please emit a metric as well so that there can be a alarm on this

service/history/mutableStateBuilder.go
```diff
@@ -1366,6 +1366,16 @@ func (e *mutableStateBuilder) DeleteUserTimer(
 			fmt.Sprintf("Unable to find timer: %v in mutable state", timerID),
 			tag.ErrorTypeInvalidMutableStateAction,
 		)
+		// DeleteUserTimer only called by Replicate Timer Cancel and Timer Fired events
+		// Ignore timer double fired issue on domain cb026e4f-a3af-4a5a-8036-6b53f2f8cab5
+		if e.executionInfo != nil && e.executionInfo.DomainID == "cb026e4f-a3af-4a5a-8036-6b53f2f8cab5" {
+
+			e.logger.Warn("encounter timer double fire issue",
```
venkat1109: same comment for the error string - please also prefix with passive_ - so that its clear this is on the passive side

yux0:
Do not merge into release branch

PR #2858, opened 2019-11-22
Add --datacenter option to cadence-cassandra-tool

kelnos:
If passed, the tool will create the keyspace using NetworkTopologyStrategy instead of SimpleStrategy.

### DIFF
--- a/tools/cassandra/cqlclient.go
+++ b/tools/cassandra/cqlclient.go
@@ -38,6 +38,7 @@ import (
 
 type (
 	cqlClient struct {
+		datacenter    string
 		nReplicas     int
 		session       *gocql.Session
 		clusterConfig *gocql.ClusterConfig
@@ -50,6 +51,7 @@ type (
 		Password    string
 		Keyspace    string
 		Timeout     int
+		datacenter  string
 		numReplicas int
 		TLS         *auth.TLS
 	}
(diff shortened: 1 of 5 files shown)


tools/cassandra/cqlclient.go
```diff
@@ -88,8 +90,11 @@ const (
 		`old_version text, ` +
 		`PRIMARY KEY ((year, month), update_time));`
 
-	createKeyspaceCQL = `CREATE KEYSPACE IF NOT EXISTS %v ` +
+	createKeyspaceSimpleCQL = `CREATE KEYSPACE IF NOT EXISTS %v ` +
 		`WITH replication = { 'class' : 'SimpleStrategy', 'replication_factor' : %v};`
+
+	createKeyspaceNetworkTopologyCQL = `CREATE KEYSPACE IF NOT EXISTS %v ` +
```
venkat1109: If we are adding support for networkTopology, we should fully support it i.e. support multiple datacenters aka replication map  - https://docs.datastax.com/en/cql/3.3/cql/cql_reference/cqlCreateKeyspace.html

Following are two options, let me know what you think: 

Option#1
Let's add a new CLI option called 'replicaPlacementStrategy' which can be either 'simple' or 'networkTopology'. If its set to simple, --rf argument simply takes a number. If its set to 'networkTopology', --rf argument takes a string map - i.e. 'dc1:2,dc2:1' etc. From an implementation standpoint, --rf now needs to be a StringFlag instead of an IntFlag. I believe this change should be backwards compatible, though.

Option#2
We simply mark --rf as deprecated and add a new --replicationPlacementStrategy or simply --replication option. The value for this option needs to be one of  { 'class' : 'SimpleStrategy', 'replication_factor' : N }  or 
     { class' : 'NetworkTopologyStrategy', 'dc1': 3 }

kelnos:
@venkat1109 I'm fine making that change but probably won't get to it in the near term.  I made the existing change to make it easier to bring up cadence in my company's environment, and that was good enough for me.  I'll try to revisit once I've finished bringing things up here.

PR #2860, opened 2019-11-23
Set max result window for ES integration test

yycptt:


host/elasticsearch_test.go
```diff
@@ -1086,3 +1081,14 @@ func (s *elasticsearchIntegrationSuite) TestUpsertWorkflowExecution_InvalidKey()
 	s.Equal(workflow.DecisionTaskFailedCauseBadSearchAttributes, failedDecisionAttr.GetCause())
 	s.True(len(failedDecisionAttr.GetDetails()) > 0)
 }
+
+func (s *elasticsearchIntegrationSuite) verifyMaxResultWindowSize(indexName string, targetSize int) {
+	for {
```
vancexu: Can you add a max retry count to prevent dead loop?

host/elasticsearch_test.go
```diff
@@ -67,7 +67,15 @@ func (s *elasticsearchIntegrationSuite) SetupSuite() {
 	s.setupSuite("testdata/integration_elasticsearch_cluster.yaml")
 	s.esClient = CreateESClient(s.Suite, s.testClusterConfig.ESConfig.URL.String())
 	PutIndexTemplate(s.Suite, s.esClient, "testdata/es_index_template.json", "test-visibility-template")
-	CreateIndex(s.Suite, s.esClient, s.testClusterConfig.ESConfig.Indices[common.VisibilityAppName])
+
+	indexName := s.testClusterConfig.ESConfig.Indices[common.VisibilityAppName]
+	CreateIndex(s.Suite, s.esClient, indexName)
+
+	_, err := s.esClient.IndexPutSettings(indexName).
```
vancexu: Might be worth to make these a function, so that putIndex with maxWindowSize is guaranteed in effect after that call.

PR #2871, opened 2019-11-26
Fix race condition in integration test

yycptt:
This is just a temp fix for the race condition of persistence config. The race condition for archiver provider happens only in the integration test as in the test all services shares one archiver provider.

The race condition for persistence config happens because all the services are using the same persistence config (the datastore map in particular) and will retry to modify the maxQPS during start up and use the modified maxQPS value to create a persistence factory. 

We need to come up with a better way to resolve this race condition. One idea is to remove all the pointers in the config definition, which makes every copy of persistence config a deep copy.

### DIFF
--- a/host/onebox.go
+++ b/host/onebox.go
@@ -469,7 +469,7 @@ func (c *cadenceImpl) startHistory(
 		params.DynamicConfig = dynamicconfig.NewNopClient()
 		dispatcher, err := params.DispatcherProvider.Get(common.FrontendServiceName, c.FrontendAddress())
 		if err != nil {
-			c.logger.Fatal("Failed to get dispatcher for frontend", tag.Error(err))
+			c.logger.Fatal("Failed to get dispatcher for history", tag.Error(err))
 		}
 		params.PublicClient = cwsc.New(dispatcher.ClientConfig(common.FrontendServiceName))
 		params.ArchivalMetadata = c.archiverMetadata
@@ -477,7 +477,7 @@ func (c *cadenceImpl) startHistory(
 
 		params.PersistenceConfig, err = copyPersistenceConfig(c.persistenceConfig)
 		if err != nil {
-			c.logger.Fatal("Failed to copy persistence config for frontend", tag.Error(err))
+			c.logger.Fatal("Failed to copy persistence config for history", tag.Error(err))
 		}
 
 		service := service.New(params)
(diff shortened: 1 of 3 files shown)


host/onebox.go
```diff
@@ -468,7 +473,12 @@ func (c *cadenceImpl) startHistory(
 		}
 		params.PublicClient = cwsc.New(dispatcher.ClientConfig(common.FrontendServiceName))
 		params.ArchivalMetadata = c.archiverMetadata
-		params.ArchiverProvider = c.archiverProvider
+		params.ArchiverProvider = newTestArchiverProvider()
+
+		params.PersistenceConfig, err = copyPersistenceConfig(c.persistenceConfig)
+		if err != nil {
+			c.logger.Fatal("Failed to copy persistence config for frontend", tag.Error(err))
```
yux0: for history

host/onebox.go
```diff
@@ -557,11 +567,16 @@ func (c *cadenceImpl) startMatching(hosts map[string][]string, startWG *sync.Wai
...
 	params.MetricsClient = metrics.NewClient(params.MetricScope, service.GetMetricsServiceIdx(params.Name, c.logger))
 	params.DynamicConfig = newIntegrationConfigClient(dynamicconfig.NewNopClient())
 	params.ArchivalMetadata = c.archiverMetadata
-	params.ArchiverProvider = c.archiverProvider
+	params.ArchiverProvider = newTestArchiverProvider()
+
+	var err error
+	params.PersistenceConfig, err = copyPersistenceConfig(c.persistenceConfig)
+	if err != nil {
+		c.logger.Fatal("Failed to copy persistence config for frontend", tag.Error(err))
```
yux0: matching

host/onebox.go
```diff
@@ -594,11 +609,16 @@ func (c *cadenceImpl) startWorker(hosts map[string][]string, startWG *sync.WaitG
...
```
yux0: worker

host/onebox.go
```diff
@@ -742,6 +762,39 @@ func (c *cadenceImpl) GetExecutionManagerFactory() persistence.ExecutionManagerF
...
+		&config.HistoryArchiverProvider{
+			Filestore: cfg,
+		},
+		&config.VisibilityArchiverProvider{
+			Filestore: cfg,
+		},
+	)
+}
+
+func copyPersistenceConfig(pConfig config.Persistence) (config.Persistence, error) {
```
yux0: Add a comment why we are doing this

yycptt: Done.

PR #2872, opened 2019-11-26
Update words with spelling

yux0:
Update words with spelling


service/history/service.go
```diff
@@ -198,7 +198,7 @@ const (
 func NewConfig(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isAdvancedVisConfigExist bool) *Config {
 	cfg := &Config{
 		NumberOfShards:                                        numberOfShards,
-		EnableNDC:                                             dc.GetBoolPropertyFnWithDomainFilter(dynamicconfig.EnableNDC, false),
+		EnableNDC:                                             dc.GetBoolPropertyFnWithDomainFilter(dynamicconfig.EnableNDC, true),
```
yycptt: enable by default?

yux0: good catch. Reverted local changes

PR #2875, opened 2019-11-27
Use new resource struct for history

yycptt:
Use resource struct for history service, history handler, shard controller and shard context. 
We can also consider use the resource struct in history engine, but will leave this as a future work.


service/history/workflowResetter_test.go
```diff
@@ -31,36 +31,37 @@ import (
...
-		logger              log.Logger
-
 		controller         *gomock.Controller
+		mockResource       *resource.Test
 		mockTransactionMgr *MocknDCTransactionMgr
 
-		workflowResetter *workflowResetterImpl
+		mockShard          *shardContextImpl
+		mockHistoryV2Mgr   *mocks.HistoryV2Manager
+		mockStateRebuilder *MocknDCStateRebuilder
```
wxing1292: could you put all `*gomock.Controller` related vars in the section under `controller         *gomock.Controller`?

service/history/transferQueueStandbyProcessor_test.go
```diff
@@ -56,29 +54,24 @@ type (
...
 		mockReplicationProcessor *MockReplicatorQueueProcessor
 		mockTimerProcessor       *MocktimerQueueProcessor
 		mockDomainCache          *cache.MockDomainCache
+		mockClusterMetadata      *cluster.MockMetadata
+		mockNDCHistoryResender   *xdc.MockNDCHistoryResender
+		mockMatchingClient       *matchingservicetest.MockClient
 
-		mockShardManager        *mocks.ShardManager
-		mockHistoryEngine       *historyEngineImpl
+		mockShard               ShardContext
```
wxing1292: maybe it is a good idea to create a mock shard context to be used by all tests

service/history/shardController.go
```diff
@@ -42,24 +41,19 @@ const (
...
-		shutdownWG          sync.WaitGroup
-		shutdownCh          chan struct{}
-		logger              log.Logger
-		throttledLoggger    log.Logger
-		config              *Config
-		metricsClient       metrics.Client
+		resource.Resource
+
+		membershipUpdateCh chan *membership.ChangedEvent
+		engineFactory      EngineFactory
```
wxing1292: could we get rid of the engine factory?

wxing1292: history engine should be created by shard controller

service/history/shardController.go
```diff
@@ -42,24 +41,19 @@ const (
...
+		shardClosedCh      chan int
+		isStarted          int32
+		isStopped          int32
```
wxing1292: plz use the daemon status, commonly used in entities which have `Start` and `Stop` functions

service/history/workflowResetter_test.go
yycptt: Good catch. Missed this one.

service/history/transferQueueStandbyProcessor_test.go
yycptt: I thought about this when modifying the tests. I think there are advantages and disadvantages. A mock shard context will result in lots of duplicated code for creating mocks for things that are already available in `resourceTest` and for asserting expectations, etc. Using `ResourceTest` and the `shardContext` struct can avoid these duplications. However, using mock shard context can make shard context independent from resource during testing. We can create an issue for this and make the decision later.

service/history/shardController.go
yycptt: I will have another PR for refactoring shard controller, shard item, shard context and history engine construction. This will be done as part of that PR.

service/history/shardController.go
yycptt: Done. Also modified `DescribeHistoryHost` a little bit.

service/history/shardController.go
yycptt: Actually I think history engine should be created by shard context?

wxing1292: yes engine and shard context are the used together and should be coupled, in my opinion

service/history/transferQueueStandbyProcessor_test.go
wxing1292: let us do a mock shard, which will improve the code reusability. 
basically doing a initialization with all dependency generated by resourceTest

common/resource/resourceImpl.go
```diff
@@ -151,6 +152,11 @@ func New(
 		return nil, err
 	}
 
+	hostInfo, err := membershipMonitor.WhoAmI()
+	if err != nil {
+		return nil, err
+	}
+
```
wxing1292: does this works before starting `membershipMonitor`?

service/history/transferQueueStandbyProcessor_test.go
yycptt: Sure. Let me create a separate pr for this.

common/resource/resourceImpl.go
yycptt: Moved to resource.Start().

yycptt:
Everything looks good when running canary on my local machine.

PR #2879, opened 2019-12-02
Made config printable in debug log

h7kanna:
This is to log the config in debug mode. Currently, it is empty because the Config fails to Marshal to JSON.
Example:
2019/12/01 21:01:00 config=


h7kanna:
Needs further testing.

common/service/config/config.go
```diff
@@ -403,6 +403,6 @@ func (c *Config) Validate() error {
 
 // String converts the config object into a string
 func (c *Config) String() string {
-	out, _ := json.MarshalIndent(c, "", "    ")
+	out, _ := yaml.Marshal(c)
```
longquanzheng: I think it works in our local env but not in docker simply because this Marshal return JSON with mutiple lines. Maybe we just replace the "\n" to spaces can fix it.

h7kanna: Hi, this is not only for debugging. This helps to generate config YAML from service/config structs (Go to YAML). Marshaling is not working because of missing interfaces.

longquanzheng: What is your use case? We usually don't rely on log as interfaces.

h7kanna: Oh, I was saying about the rest of the changes. They make Config struct marshallable. I am not depending on the log as an interface. I was experimenting with generating YAML config from go structs.

PR #2881, opened 2019-12-03
initiate graceful shutdown after SIGTERM

vitarb:
This is the simplest possible implementation that initiates shutdown sequence on SIGTERM closing open resources, ring, shard manager, etc.


cmd/server/cadence.go
```diff
@@ -73,12 +75,24 @@ func startHandler(c *cli.Context) {
...
 		}
 		server := newServer(svc, &cfg)
 		server.Start()
+		go func() {
+			sig := <-sigc
+			switch sig {
+			case syscall.SIGTERM:
+				log.Println("Received SIGTERM signal, initiating shutdown.")
+				server.Stop()
+				os.Exit(0)
```
venkat1109: the first goroutine to finish will call Stop() and so, any other service which is in the midst of a Stop() would just terminate forcibly. For this to work correctly, you need to have exactly one sigterm callback (not one per service) and that callback should initiate shutdown of all services when it receives a signal.

cmd/server/cadence.go
```diff
@@ -73,12 +75,24 @@ func startHandler(c *cli.Context) {
...
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGTERM)
 	for _, svc := range services {
 		if _, ok := cfg.Services[svc]; !ok {
 			log.Fatalf("`%v` service missing config", svc)
 		}
 		server := newServer(svc, &cfg)
 		server.Start()
+		go func() {
+			sig := <-sigc
```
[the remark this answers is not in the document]
vitarb: You are right, fixed that.

cmd/server/cadence.go
```diff
@@ -72,18 +75,37 @@ func startHandler(c *cli.Context) {
...
 			log.Fatalf("`%v` service missing config", svc)
 		}
 		server := newServer(svc, &cfg)
+		daemons = append(daemons, server)
 		server.Start()
 	}
 
+	go shutdownDaemonsOnSignal(sigc, daemons)
+
 	select {}
```
venkat1109: Just a suggestion - instead of spinning up a new goroutine - consider renaming your shutdownDaemonsOnSignal to something like blockOnSigTerm - and remove the select{} on this method. The select{} exist only because we don't have something like a sig handler.

vitarb: Addressed

cmd/server/cadence.go
```diff
@@ -21,17 +21,20 @@
...
-	"github.com/urfave/cli"
+	"syscall"
 
 	"github.com/uber/cadence/common/service/config"
 	"github.com/uber/cadence/tools/cassandra"
 	"github.com/uber/cadence/tools/sql"
 	_ "github.com/uber/cadence/tools/sql-extensions/mysql"    // needed to load mysql extensions
 	_ "github.com/uber/cadence/tools/sql-extensions/postgres" // needed to load postgres extensions
+
+	"github.com/urfave/cli"
```
wxing1292: make fmt

PR #2883, opened 2019-12-03
Fix missing properties in continueasnew when workflow fail

vancexu:
https://github.com/uber/cadence/issues/2882


vancexu:
> Looks good to me. Do we have any unit test for `processWorkflowTimeout`? If so, maybe a good idea to also update those tests.

Good suggestion. Unfortunately we don't have such unit tests on timer tasks so I turned to integration tests. It would be great to revisit

PR #2888, opened 2019-12-05
Concurrent shard loading

vitarb:
This change introduces ability to acquire shards concurrently.
Although underlying shardController's operations like get/remove engine are synchronized there should still be noticeable speed up due to the increased parallelism.

Note that default value for the concurrent number of goroutines is 1 meaning that it would use one thread for loading (e.g. no change from the pervious behavior).
Users would need to change history.acquireShardConcurrency property in the dynamic config in order to use higher number of goroutines. 

We may consider changing this default in future once the change is tested.

### DIFF
--- a/service/history/shardController.go
+++ b/service/history/shardController.go
@@ -295,25 +295,39 @@ func (c *shardController) acquireShards() {
 	sw := c.metricsScope.StartTimer(metrics.AcquireShardsLatency)
 	defer sw.Stop()
 
-AcquireLoop:
-	for shardID := 0; shardID < c.config.NumberOfShards; shardID++ {
-		info, err := c.GetHistoryServiceResolver().Lookup(string(shardID))
-		if err != nil {
-			c.logger.Error("Error looking up host for shardID", tag.Error(err), tag.OperationFailed, tag.ShardID(shardID))
-			continue AcquireLoop
-		}
-
-		if info.Identity() == c.GetHostInfo().Identity() {
-			_, err1 := c.getEngineForShard(shardID)
-			if err1 != nil {
-				c.metricsScope.IncCounter(metrics.GetEngineForShardErrorCounter)
-				c.logger.Error("Unable to create history shard engine", tag.Error(err1), tag.OperationFailed, tag.ShardID(shardID))
-				continue AcquireLoop
+	concurrency := common.MaxInt(c.config.AcquireShardConcurrency(), 1)
+	shardActionCh := make(chan int, concurrency)
+	var wg sync.WaitGroup
+	wg.Add(concurrency)
+	// Spawn workers that would lookup and add/remove shards concurrently.
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			defer wg.Done()
+			for shardID := range shardActionCh {
+				info, err := c.GetHistoryServiceResolver().Lookup(string(shardID))
+				if err != nil {
+					c.logger.Error("Error looking up host for shardID", tag.Error(err), tag.OperationFailed, tag.ShardID(shardID))
+				} else {
+					if info.Identity() == c.GetHostInfo().Identity() {
+						_, err1 := c.getEngineForShard(shardID)
+						if err1 != nil {
+							c.metricsScope.IncCounter(metrics.GetEngineForShardErrorCounter)
+							c.logger.Error("Unable to create history shard engine", tag.Error(err1), tag.OperationFailed, tag.ShardID(shardID))
+						}
+					} else {
+						c.removeEngineForShard(shardID)
+					}
+				}
 			}
-		} else {
-			c.removeEngineForShard(shardID)
-		}
+		}()
 	}
+	// Submit tasks to the channel.
+	for shardID := 0; shardID < c.config.NumberOfShards; shardID++ {
+		shardActionCh <- shardID
+	}
+	close(shardActionCh)
+	// Wait until all shards are processed.
+	wg.Wait()
 
 	c.metricsScope.UpdateGauge(metrics.NumShardsGauge, float64(c.numShards()))
 }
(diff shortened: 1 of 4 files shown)


service/history/shardController.go
```diff
@@ -322,24 +322,31 @@ func (c *shardController) acquireShards() {
...
-			c.removeEngineForShard(shardID)
-		}
+			if info.Identity() == c.host.Identity() {
+				_, err1 := c.getEngineForShard(shardID)
+				if err1 != nil {
+					c.metricsClient.IncCounter(metrics.HistoryShardControllerScope, metrics.GetEngineForShardErrorCounter)
+					c.logger.Error("Unable to create history shard engine", tag.Error(err1), tag.OperationFailed, tag.ShardID(shardID))
+				}
+			} else {
+				c.removeEngineForShard(shardID)
```
venkat1109: what do you think about the idea of prioritizing loads over removes ? i.e. as you are iterating through the shardIDs, if a shard is no longer owned, don't call removeEngineForShard() immediately, but store the shardID in a slice. Once all the loading is complete, then you can go over the slice and remove the shards (if the shard still needs to be removed). 

Doesn't have to be in this same diff, but if you like the idea, let's do this in a follow up diff.

service/history/shardController.go
```diff
@@ -322,24 +322,31 @@ func (c *shardController) acquireShards() {
...
 
-AcquireLoop:
-	for shardID := 0; shardID < c.config.NumberOfShards; shardID++ {
-		info, err := c.hServiceResolver.Lookup(string(shardID))
-		if err != nil {
-			c.logger.Error("Error looking up host for shardID", tag.Error(err), tag.OperationFailed, tag.ShardID(shardID))
-			continue AcquireLoop
-		}
+	concurrency := c.config.AcquireShardConcurrency()
+	sem := make(chan bool, concurrency)
```
venkat1109: [not a blocker] From a style perspective, I prefer chan struct{} for this. Since the value is never used.

service/history/shardController.go
```diff
@@ -322,24 +322,31 @@ func (c *shardController) acquireShards() {
...
+				if err1 != nil {
+					c.metricsClient.IncCounter(metrics.HistoryShardControllerScope, metrics.GetEngineForShardErrorCounter)
+					c.logger.Error("Unable to create history shard engine", tag.Error(err1), tag.OperationFailed, tag.ShardID(shardID))
+				}
+			} else {
+				c.removeEngineForShard(shardID)
+			}
+		}(shardID)
+	}
+	for i := 0; i < concurrency; i++ {
```
venkat1109: It's not obvious that you have this loop to wait for all the goroutines to finish (since readers are typically used to seeing sync.WaitGroups). Consider adding a comment to help the future reader.

service/history/shardController.go
```diff
@@ -322,24 +322,31 @@ func (c *shardController) acquireShards() {
...
 	defer sw.Stop()
 
-AcquireLoop:
-	for shardID := 0; shardID < c.config.NumberOfShards; shardID++ {
-		info, err := c.hServiceResolver.Lookup(string(shardID))
-		if err != nil {
-			c.logger.Error("Error looking up host for shardID", tag.Error(err), tag.OperationFailed, tag.ShardID(shardID))
-			continue AcquireLoop
-		}
+	concurrency := c.config.AcquireShardConcurrency()
```
venkat1109: just for safety - enforce a lower bound of 1 i.e. concurrency = common.MaxInt(1, c.config.AcquireShards...())

service/history/shardController.go
```diff
@@ -322,24 +322,31 @@ func (c *shardController) acquireShards() {
...
 
-		if info.Identity() == c.host.Identity() {
-			_, err1 := c.getEngineForShard(shardID)
-			if err1 != nil {
-				c.metricsClient.IncCounter(metrics.HistoryShardControllerScope, metrics.GetEngineForShardErrorCounter)
-				c.logger.Error("Unable to create history shard engine", tag.Error(err1), tag.OperationFailed, tag.ShardID(shardID))
-				continue AcquireLoop
+	for shardID := 0; shardID < c.config.NumberOfShards; shardID++ {
+		sem <- true
+		go func(shardID int) {
```
venkat1109: consider using the [worker-pool](https://gobyexample.com/worker-pools) pattern here (you can ignore the results chan in that example - which is not needed here). I think code will be more straightforward and also efficient (doesn't create goroutine per shardID) with that pattern. In addition, you can use waitGroups to wait for all goroutines to finish, which go-readers are used to seeing.

service/history/shardController.go
vitarb: Makes sense, we can run this experiment too.

service/history/shardController.go
vitarb: Added a comment.

service/history/shardController.go
vitarb: Done.

service/history/shardController.go
vitarb: It is slightly more verbose, but arguably easier to read. Updated.

service/history/shardController.go
```diff
@@ -322,25 +322,38 @@ func (c *shardController) acquireShards() {
 	sw := c.metricsClient.StartTimer(metrics.HistoryShardControllerScope, metrics.AcquireShardsLatency)
 	defer sw.Stop()
 
-AcquireLoop:
+	concurrency := common.MaxInt(c.config.AcquireShardConcurrency(), 1)
+	shardActionCh := make(chan int, c.config.NumberOfShards)
```
venkat1109: I think you don't need a channel of length NumberOfShards if you move the loop that enqueues the shardIDs into the shardActionCh after the loop that spins up goroutines. The channel could be any size > 0 but channel of size=concurrency might be optimal.

service/history/shardController.go
```diff
@@ -322,25 +322,38 @@ func (c *shardController) acquireShards() {
...
+	var wg sync.WaitGroup
+	wg.Add(c.config.NumberOfShards)
```
venkat1109: same comment - you only need wg of size=concurrency. And goroutines just do - defer wg.Done()

service/history/shardController.go
vitarb: Done

service/history/shardController.go
vitarb: Addressed too.

service/history/shardController.go
vitarb: No longer relevant since code has been refactored.

Issue #2877, opened 2019-11-28
Flaky test in postgres persistence


longquanzheng:
/*
FAIL: TestSQLQueuePersistence/TestDomainReplicationQueue (0.26s)
        queuePersistenceTest.go:102:
            	Error Trace:	queuePersistenceTest.go:102
            	Error:      	Not equal:
            	            	expected: 99
            	            	actual  : 98
            	Test:       	TestSQLQueuePersistence/TestDomainReplicationQueue
 */
```
func TestSQLQueuePersistence(t *testing.T) {
	s := new(pt.QueuePersistenceSuite)
	s.TestBase = pt.NewTestBaseWithSQL(getTestClusterOption())
	s.TestBase.Setup()
	suite.Run(t, s)
}
```

longquanzheng:
This still needs some work. This is an open issue in the PR that I found but not solved yet.

Issue #2869, opened 2019-11-26
Refactor server main() to make it easier for different persistence


longquanzheng:
Make it same as our CLI

Issue #2863, opened 2019-11-25
Fix NameExec for bulk insert in Postgres library(sqlx) 


longquanzheng:
In sqlx, with NameExec, if there are more than one records to insert, we will see errors like:
```
InternalServiceError{Message: applyTasks failed. Failed to create replication tasks. Error: InternalServiceError{Message: createReplicationTasks failed. Error: pq: got 12 parameters but the statement requires 4}}
          
```

This is because of a bug in sqlx library that it doesn't handle the **$X** correctly when expanding the query template into real SQL.

For example, for 3 records,
It should be 
```
INSERT INTO ... $1,$2,$3,$4),($5,$6,$7,$8),($9,$10,$11,$12)
```
Instead of 
```
INSERT INTO ... $1,$2,$3,$4),($1,$2,$3,$4),($1,$2,$3,$4)
```

longquanzheng:
https://github.com/longquanzheng/sqlx/commit/9400aeba572ddb3f41494b4553d91fa4a0dc33c4

longquanzheng:
This still needs work. This is an open issue in the PR that I found but not solved yet.

joeandaverde:
I submitted a PR to sqlx that should resolve this issue: https://github.com/jmoiron/sqlx/pull/579

Issue #2854, opened 2019-11-22
Should use shared connection pool for creating VisibilityStore/Manager


longquanzheng:

longquanzheng:
This still needs work. This is an open issue in the PR that I found but not solved yet.

PR #2889, opened 2019-12-06
Implement Postgres SQL Plugin

longquanzheng:
This PR is implementing the defined interface of SQL Plugin and the testing.
It also fixes some issues with the Postgres schema we landed previously with CLI tool.

Running with samples locally looks good:

```
(postgres4)$./cadence-server --env development_postgres start
2019/11/27 16:59:40 Loading config; env=development_postgres,zone=,configDir=./config
2019/11/27 16:59:40 Loading configFiles=[./config/base.yaml ./config/development_postgres.yaml]
2019/11/27 16:59:40 error creating file based dynamic config client, use no-op config client instead. error: error checking dynamic config file at path , error: stat : no such file or directory
{"level":"info","ts":"2019-11-27T16:59:40.141-0800","msg":"Created RPC dispatcher and listening","service":"cadence-frontend","address":"127.0.0.1:7933","logging-call-at":"rpc.go:81"}
```

### DIFF
--- a/common/persistence/sql/sqlplugin/postgres/admin.go
+++ b/common/persistence/sql/sqlplugin/postgres/admin.go
@@ -1,0 +1,130 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package postgres
+
+import (
+	"fmt"
+	"time"
+)
+
+const (
+	readSchemaVersionQuery = `SELECT curr_version from schema_version where db_name=$1`
+
+	writeSchemaVersionQuery = `INSERT into schema_version(db_name, creation_time, curr_version, min_compatible_version) VALUES ($1,$2,$3,$4)
+										ON CONFLICT (db_name) DO UPDATE 
+										  SET creation_time = excluded.creation_time,
+										   	  curr_version = excluded.curr_version,
+										      min_compatible_version = excluded.min_compatible_version;`
+
+	writeSchemaUpdateHistoryQuery = `INSERT into schema_update_history(year, month, update_time, old_version, new_version, manifest_md5, description) VALUES($1,$2,$3,$4,$5,$6,$7)`
+
+	createSchemaVersionTableQuery = `CREATE TABLE schema_version(db_name VARCHAR(255) not null PRIMARY KEY, ` +
+		`creation_time TIMESTAMP, ` +
+		`curr_version VARCHAR(64), ` +
+		`min_compatible_version VARCHAR(64));`
+
+	createSchemaUpdateHistoryTableQuery = `CREATE TABLE schema_update_history(` +
+		`year int not null, ` +
+		`month int not null, ` +
+		`update_time TIMESTAMP not null, ` +
+		`description VARCHAR(255), ` +
+		`manifest_md5 VARCHAR(64), ` +
+		`new_version VARCHAR(64), ` +
+		`old_version VARCHAR(64), ` +
+		`PRIMARY KEY (year, month, update_time));`
+
+	//NOTE we have to use %v because somehow mysql doesn't work with ? here
+	createDatabaseQuery = "CREATE database %v"
+
+	dropDatabaseQuery = "Drop database %v"
+
+	listTablesQuery = "select table_name from information_schema.tables where table_schema='public'"
+
+	dropTableQuery = "DROP TABLE %v"
+)
+
+// CreateSchemaVersionTables sets up the schema version tables
+func (pdb *db) CreateSchemaVersionTables() error {
+	if err := pdb.Exec(createSchemaVersionTableQuery); err != nil {
+		return err
+	}
+	return pdb.Exec(createSchemaUpdateHistoryTableQuery)
+}
+
+// ReadSchemaVersion returns the current schema version for the keyspace
+func (pdb *db) ReadSchemaVersion(database string) (string, error) {
+	var version string
+	err := pdb.db.Get(&version, readSchemaVersionQuery, database)
+	return version, err
+}
+
+// UpdateSchemaVersion updates the schema version for the keyspace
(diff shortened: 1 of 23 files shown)


longquanzheng:
TODOs: 

#2863
#2856
#2854
#2869
#2870
#2877

common/persistence/sql/sqlplugin/postgres/admin.go
```diff
@@ -0,0 +1,130 @@
...
+		`year int not null, ` +
+		`month int not null, ` +
+		`update_time TIMESTAMP not null, ` +
+		`description VARCHAR(255), ` +
+		`manifest_md5 VARCHAR(64), ` +
+		`new_version VARCHAR(64), ` +
+		`old_version VARCHAR(64), ` +
+		`PRIMARY KEY (year, month, update_time));`
+
+	//NOTE we have to use %v because somehow mysql doesn't work with ? here
```
venkat1109: comment out of date. mysql is irrelevant here.

common/persistence/sql/sqlplugin/postgres/db.go
```diff
@@ -0,0 +1,94 @@
...
+	db        *sqlx.DB
+	tx        *sqlx.Tx
+	conn      sqlplugin.Conn
+	converter DataConverter
+}
+
+var _ sqlplugin.DB = (*db)(nil)
+var _ sqlplugin.Tx = (*db)(nil)
+
+func (pdb *db) GetDriverName() string {
```
venkat1109: where is this used ?

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -0,0 +1,100 @@
...
+// THE SOFTWARE.
+
+package postgres
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/uber/cadence/common/persistence/sql"
```
venkat1109: imports formatting

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/jmoiron/sqlx"
+
+	"github.com/uber/cadence/common/persistence/sql/sqlplugin"
+	"github.com/uber/cadence/common/service/config"
+)
+
+const (
+	// PluginName is the name of the plugin
+	PluginName             = "postgres"
+	dataSourceNamePostgres = "user=%v password=%v host=%v port=%v dbname=%v sslmode=disable "
```
venkat1109: why is sslmode always set to disable. I thought we support TLS via mysql already ? No ?

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -0,0 +1,100 @@
...
+	db := NewDB(conn, nil)
+	return db, nil
+}
+
+// CreateDBConnection creates a returns a reference to a logical connection to the
+// underlying SQL database. The returned object is to tied to a single
+// SQL database and the object can be used to perform CRUD operations on
+// the tables in the database
+func (d *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	ss := strings.Split(cfg.ConnectAddr, ":")
```
venkat1109: use net.SplitHostPort() instead

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	host := ss[0]
+	port, err := strconv.Atoi(ss[1])
+	if err != nil {
+		return nil, fmt.Errorf("invalid port number: %v, %v", ss[1], cfg.ConnectAddr)
+	}
+
+	dbName := cfg.DatabaseName
+	//NOTE: postgres doesn't allow to connect with empty dbName, the admin dbName is "postgres"
+	if dbName == "" {
```
venkat1109: I believe this is supposed to be an error condition. During config init, we should be validating that dbName is non-nil. Double check

common/persistence/sql/sqlplugin/postgres/typeconv.go
```diff
@@ -0,0 +1,57 @@
...
+	// DataConverter defines the API for conversions to/from
+	// go types to postgres datatypes
+	DataConverter interface {
+		ToPostgresDateTime(t time.Time) time.Time
+		FromPostgresDateTime(t time.Time) time.Time
+	}
+	converter struct{}
+)
+
+// ToMySQLDateTime converts to time to MySQL datetime
```
venkat1109: comment out of date

common/persistence/sql/sqlplugin/postgres/typeconv.go
```diff
@@ -0,0 +1,57 @@
...
+func (c *converter) ToPostgresDateTime(t time.Time) time.Time {
```
venkat1109: why are these conversions necessary in postgres ? can you add a comment explaining why this conversion needs to be done manually ? In the case of MySQL, there were limitations around date/time datatypes that doesn't quite work nicely with the server side use of date/time and so the conversion was necessary

common/persistence/sql/sqlplugin/postgres/typeconv.go
```diff
@@ -0,0 +1,57 @@
...
+	zn, _ := t.Zone()
+	if zn != localZone {
+		nano := t.UnixNano()
+		t := time.Unix(0, nano)
+		return t
+	}
+	return t
+}
+
+// FromMySQLDateTime converts mysql datetime and returns go time
```
venkat1109: comment out of date

go.mod
```diff
@@ -81,4 +81,4 @@ require (
 	gopkg.in/yaml.v2 v2.2.2
 )
 
-replace github.com/jmoiron/sqlx v1.2.0 => github.com/mfateev/sqlx v0.0.0-20180910213730-fa49b1cf03f7
+replace github.com/jmoiron/sqlx v1.2.0 => github.com/longquanzheng/sqlx v0.0.0-20191125235044-053e6130695c
```
venkat1109: please add a comment / readme somewhere explaining why this is needed and also create a github issue so that we don't forget about this

common/persistence/sql/sqlplugin/postgres/admin.go
longquanzheng: Yes, it's a mistype -- should be postgres

common/persistence/sql/sqlplugin/postgres/db.go
longquanzheng: It's used in "func (m *sqlStore) GetName() string {" in sql/common.go".

common/persistence/sql/sqlplugin/postgres/plugin.go
longquanzheng: No, not yet, his PR is waiting for my PR to land.

common/persistence/sql/sqlplugin/postgres/plugin.go
longquanzheng: Sounds good

common/persistence/sql/sqlplugin/postgres/plugin.go
longquanzheng: Yes, but that's only validated in server bootstrap. This case is only for CLI and testing suite.

common/persistence/sql/sqlplugin/postgres/typeconv.go
longquanzheng: Yes I forgot to add those comments. It took me quite a while to fix those conversion. Basically  our application layer is not consistent with timezone: for example, in some case we write timestamp with local timezone but when the time.Time is converted from "JSON"(from paging token), the timezone is missing... The simplest way is to disable timezone in postgres, I tried different ways but they don't work.

go.mod
longquanzheng: Yes I have created an issue. I will add a comment

common/persistence/sql/sqlplugin/postgres/db.go
longquanzheng: You are right. The one that is being used is GetPluginName(). This one is duplicated when I rename from Driver to Plugin.

sunfmin:
How about to make `ubercadence/server:master-auto-setup` work with postgres?

enmand:
@sunfmin https://github.com/uber/cadence/pull/2894 should make the auto-setup docker image work if you build it yourself with ` docker build . -t REPO/TAG_NAME --build-arg TARGET=auto-setup`

PR #2890, opened 2019-12-06
frontend.PollForDecisionTask: add validation to verify history is complete

venkat1109:
This patch adds validation logic within frontend::PollForDecisionTask API to make sure the history we send to the worker is complete. When this validation fails, we fail the API call and emit a metric. Purpose of this change is to find out if these types of errors do occur in production (to explain some of the odd behavior we see).

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -1368,6 +1368,7 @@ const (
 	CadenceErrRetryTaskCounter
 	CadenceErrBadBinaryCounter
 	CadenceErrClientVersionNotSupportedCounter
+	CadenceErrIncompleteHistoryCounter
 	PersistenceRequests
 	PersistenceFailures
 	PersistenceLatency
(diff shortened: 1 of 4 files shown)


common/metrics/defs.go
```diff
@@ -1437,6 +1437,7 @@ const (
 
 	MatchingClientForwardedCounter
 	MatchingClientInvalidTaskListName
+	CadenceErrIncompleteHistoryCounter
```
wxing1292: i think you can put it on line 1370? near CadenceErrClientVersionNotSupportedCounter

service/frontend/workflowHandler.go
```diff
@@ -3254,6 +3259,36 @@ func (wh *WorkflowHandler) createPollForDecisionTaskResponse(
 	return resp, nil
 }
 
+func verifyHistoryIsComplete(
+	history *gen.History,
+	expectedFirstEventID int64,
+	expectedLastEventID int64) error {
```
wxing1292:
```go
func verifyHistoryIsComplete(
	history *gen.History,
	expectedFirstEventID int64,
	expectedLastEventID int64,
) error {
```

venkat1109: done

PR #2894, opened 2019-12-09
Add Postgres configuration in docker configs

enmand:
This PR adds the required docker configuration for the `auto-setup` image on Postgres

### DIFF
--- a/docker/config_template.yaml
+++ b/docker/config_template.yaml
@@ -26,7 +26,7 @@ persistence:
         {{- else if eq $db "mysql" }}
         default:
             sql:
-                driverName: "mysql"
+                pluginName: "mysql"
                 databaseName: {{ default .Env.DBNAME "cadence" }}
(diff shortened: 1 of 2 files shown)


docker/config_template.yaml
```diff
@@ -48,6 +48,29 @@ persistence:
                 connectAttributes:
                     tx_isolation: 'READ-COMMITTED'
                 {{- end }}
+        {{- else if eq $db "postgres" }}
+        default:
+            sql:
+                pluginName: "postgres"
+                databaseName: {{ default .Env.DBNAME "cadence" }}
+                connectAddr: "{{ default .Env.POSTGRES_SEEDS "" }}:{{ default .Env.DB_PORT "5432" }}"
```
enmand: Maybe this should be `SQL_*` to match the env vars in the SQL tool?

longquanzheng: I think this is fine. We use POSTGRES_SEEDS in https://github.com/uber/cadence/blob/master/environment/env.go#L62

longquanzheng:
Thanks for doing this ! @enmand

PR #2899, opened 2019-12-10
Sticky query support

andrewjdawson2016:
Sticky query has many edge cases and small bugs it also lacks visibility and the dynamic domain controls. This diff fixes many of these issues. The code around sticky query is still messy and likely has some bugs in edge cases (it is tricky code). A long term fix that should be implemented is the worker should respond back to the server when stickiness is cleared this would greatly improve the quality of this bit of code. 

At a high level this code does the following:
- There was a sticky ttl for workflows which do not produce many tasks but it was not being used. Here I make sure it was used - here I ensure its used. 
- I added a dynamic config to disable sticky query per domain
- I added a lot of metrics around query, sticky query and various time out cases
- I fixed a bug in which a closed workflow attempts to reset stickiness and wrongly produces an error from QueryWorkflow.


service/history/historyEngine.go
```diff
@@ -897,29 +900,109 @@ func (e *historyEngineImpl) queryDirectlyThroughMatching(
...
+		stickyStopWatch := scope.StartTimer(metrics.DirectQueryDispatchStickyLatency)
+		matchingResp, err := e.rawMatchingClient.QueryWorkflow(stickyContext, stickyMatchingRequest)
+		stickyStopWatch.Stop()
+		cancel()
+		if err == nil {
+			scope.IncCounter(metrics.DirectQueryDispatchStickySuccessCount)
+			return &h.QueryWorkflowResponse{Response: matchingResp}, nil
+		}
+		if yarpcError, ok := err.(*yarpcerrors.Status); !ok || yarpcError.Code() != yarpcerrors.CodeDeadlineExceeded {
+			e.logger.Error("query directly though matching on sticky failed, will not attempt query on non-sticky",
```
venkat1109: prefer emitting a metric (possibly domain scoped) to a log message here

service/history/historyEngine.go
```diff
@@ -897,29 +900,109 @@ func (e *historyEngineImpl) queryDirectlyThroughMatching(
...
+				tag.WorkflowDomainName(queryRequest.GetDomain()),
+				tag.WorkflowID(queryRequest.Execution.GetWorkflowId()),
+				tag.WorkflowRunID(queryRequest.Execution.GetRunId()),
+				tag.WorkflowQueryType(queryRequest.Query.GetQueryType()),
+				tag.Error(err))
+			return nil, err
+		}
+		if msResp.GetIsWorkflowRunning() {
+			e.logger.Info("query direct through matching failed on sticky, clearing sticky before attempting on non-sticky",
```
venkat1109: same comment about logging

service/history/historyEngine.go
```diff
@@ -897,29 +900,109 @@ func (e *historyEngineImpl) queryDirectlyThroughMatching(
...
+			cancel()
+			if err != nil && err != ErrWorkflowCompleted {
+				return nil, err
+			}
+			scope.IncCounter(metrics.DirectQueryDispatchClearStickinessSuccessCount)
+		}
+	}
+
+	if err := common.IsValidContext(ctx); err != nil {
+		e.logger.Info("query context timed out before query on non-sticky task list could be attempted",
```
venkat1109: this can cause log noise - prefer metric to logging here

service/history/historyEngine.go
```diff
@@ -897,29 +900,109 @@ func (e *historyEngineImpl) queryDirectlyThroughMatching(
...
+			clearStickinessStopWatch.Stop()
+			cancel()
+			if err != nil && err != ErrWorkflowCompleted {
+				return nil, err
+			}
+			scope.IncCounter(metrics.DirectQueryDispatchClearStickinessSuccessCount)
+		}
+	}
+
+	if err := common.IsValidContext(ctx); err != nil {
```
venkat1109: Can you also check if there is any time left before the context deadline ? There is no point in making a call if the context will be valid only for another 2ms

service/history/historyEngine.go
andrewjdawson2016: this log already existed

service/history/historyEngine.go
andrewjdawson2016: this log will be run a subset of the times that the log below will be run and the log below was already present.

service/history/historyEngine.go
andrewjdawson2016: Updated the IsValidContext method to account for this generally.

PR #2900, opened 2019-12-11
frontend.getHistory: add validation / sanity checks

venkat1109:
This patch is an update to an earlier diff that added validation within pollForDecisionTask. The earlier patch breaks with pagination and in addition, doesn't do any validation within GetWFHistory. This patch fixes both of the above problems.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -1734,7 +1734,6 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 			tag.WorkflowID(getRequest.Execution.GetWorkflowId()),
 			tag.WorkflowRunID(getRequest.Execution.GetRunId()),
 			tag.WorkflowDomainID(domainID), tag.WorkflowSize(int64(getRequest.GetMaximumPageSize())))
-
 		getRequest.MaximumPageSize = common.Int32Ptr(common.GetHistoryMaxPageSize)
 	}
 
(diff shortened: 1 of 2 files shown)


service/frontend/workflowHandler.go
```diff
@@ -3024,7 +3024,31 @@ func (wh *WorkflowHandler) getHistory(
 
 	scope.RecordTimer(metrics.HistorySize, time.Duration(size))
 
+	if err := verifyHistoryIsComplete(
+		historyEvents,
+		firstEventID,
+		nextEventID-1,
+		isFirstPage,
+		len(nextPageToken) == 0,
```
longquanzheng: How about make a similar variable **isLastPage** to help it more readable?

service/frontend/workflowHandler.go
```diff
@@ -3274,35 +3298,63 @@ func (wh *WorkflowHandler) createPollForDecisionTaskResponse(
...
+		return fmt.Errorf("invalid history: contains zero events")
+	}
+
+	firstEventID := events[0].GetEventId()
+	lastEventID := events[nEvents-1].GetEventId()
+
+	if !isFirstPage { // atleast one page of history has been read previously
+		if firstEventID <= expectedFirstEventID {
+			// not first page and no events have been read in the previous pages - not possible
+			return fmt.Errorf(
```
longquanzheng: Nit: use internalServiceError as our convention. Today there is only special case that we want to use fmt.Errorf because of our internal SLO calculation. I am going to clean i up with https://github.com/uber/cadence/issues/1128

PR #2903, opened 2019-12-12
Cadence authorization

vancexu:
https://github.com/uber/cadence/issues/2833

Todo in separate PR:
- add dynamic config
- add metrics
- add tests
- refactor FE handler

### DIFF
--- a/common/auth/authority.go
+++ b/common/auth/authority.go
@@ -1,0 +1,57 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination authority_mock.go
+
+package auth
+
+import "context"
+
+const (
+	// DecisionDeny means auth decision is deny
+	DecisionDeny AuthorizationDecision = iota
+	// DecisionAllow means auth decision is allow
+	DecisionAllow
+	//DecisionNoOpinion means auth decision is unknown
+	DecisionNoOpinion
+)
+
+type (
+	// AuthorizationParams is input for authority to make decision.
+	// It can be extended in future if required auth on resources like WorkflowType and TaskList
+	AuthorizationParams struct {
+		Actor      string
+		Action     string
+		DomainName string
+	}
+
+	// AuthorizationResult is result from authority.
+	AuthorizationResult struct {
+		AuthorizationDecision AuthorizationDecision
+	}
+
+	// AuthorizationDecision is enum type for auth decision
+	AuthorizationDecision int
+)
+
+// Authority is an interface for authorization
+type Authority interface {
+	IsAuthorized(ctx context.Context, params *AuthorizationParams) (AuthorizationResult, error)
+}
--- a/common/auth/nopAuthority.go
+++ b/common/auth/nopAuthority.go
@@ -1,0 +1,37 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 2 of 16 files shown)


common/auth/authority.go
```diff
@@ -0,0 +1,57 @@
...
+		AuthorizationDecision AuthorizationDecision
+	}
+
+	// AuthorizationDecision is enum type for auth decision
+	AuthorizationDecision int
+)
+
+// Authority is an interface for authorization
+type Authority interface {
+	IsAuthorized(ctx context.Context, params *AuthorizationParams) (AuthorizationResult, error)
```
venkat1109: IsFoo methods should return bool - since you are returning a result in this case, I prefer calling this method "Authorize() (Result, error)"

common/auth/authority.go
```diff
@@ -0,0 +1,57 @@
...
+// THE SOFTWARE.
+
+//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination authority_mock.go
+
+package auth
+
+import "context"
+
+const (
+	// DecisionDeny means auth decision is deny
```
venkat1109: s/auth/authorization - same in other places

common/auth/nopAuthority.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
venkat1109: 2019

service/frontend/authHandler.go
```diff
@@ -0,0 +1,648 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
venkat1109: 2019

service/frontend/authHandler.go
```diff
@@ -0,0 +1,648 @@
...
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/auth"
+	"github.com/uber/cadence/common/resource"
+)
+
+// TODO(vancexu): add metrics
+
+const (
+	// ActionCommon is for authority implementation to handle auth for common operations
+	ActionCommon = "common"
```
venkat1109: see comment above - instead of passing action, we should consider passing in the APIName or APIGroup. Also, admin really is a role (actor) rather than action

service/frontend/authHandler.go
```diff
@@ -0,0 +1,648 @@
...
+	// ActionAdmin is for authority implementation to handle auth for admin only operations
+	ActionAdmin = "admin"
+)
+
+// AuthHandlerImpl frontend handler wrapper for authentication and authorization
+type AuthHandlerImpl struct {
```
venkat1109: nit: can we call this SecureWorkflowHandler or AccessControlledWorkflowHandler instead ? The workflowHandler naming convention is to simply be consistent with existing naming - esp. given the fact that this is  a decorator class

service/frontend/authHandler.go
```diff
@@ -0,0 +1,648 @@
...
+	return a.frontendHandler.CountWorkflowExecutions(ctx, request)
+}
+
+// DeprecateDomain API call
+func (a *AuthHandlerImpl) DeprecateDomain(
+	ctx context.Context,
+	request *shared.DeprecateDomainRequest,
+) error {
+
+	isAllowed, err := a.isAdmin(ctx)
```
venkat1109: this look weird. How about creating a helper something like 

```
isAuthorized(actor string, domainName string) (bool, error) {}
```
For this api, you can pass the actor as admin

service/frontend/workflowHandler.go
```diff
@@ -118,6 +118,7 @@ var (
 	errInvalidTaskStartToCloseTimeoutSeconds      = &gen.BadRequestError{Message: "A valid TaskStartToCloseTimeoutSeconds is not set on request."}
 	errClientVersionNotSet                        = &gen.BadRequestError{Message: "Client version is not set on request."}
 	errQueryDisallowedForDomain                   = &gen.BadRequestError{Message: "Domain is not allowed to query, please contact cadence team to re-enable queries."}
+	errUnauthorized                               = &gen.BadRequestError{Message: "Request unauthorized."}
```
venkat1109: this doesn't belong here. Please move this within your decorator (authHandlerImpl) handler file

vancexu:
Thank you @venkat1109 for the review, all your suggestions make sense to me, and I updated accordingly.

common/authorization/authorizer.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
venkat1109: 2019

common/authorization/nopAuthorizer.go
```diff
@@ -0,0 +1,37 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package authorization
+
+import "context"
+
+type nopAuthority struct{}
```
venkat1109: noopAuthorizer - to be consistent with naming

common/authorization/authorizer.go
```diff
@@ -0,0 +1,57 @@
...
+
+import "context"
+
+const (
+	// DecisionDeny means auth decision is deny
+	DecisionDeny Decision = iota + 1
+	// DecisionAllow means auth decision is allow
+	DecisionAllow
+	//DecisionNoOpinion means auth decision is unknown
+	DecisionNoOpinion
```
venkat1109: Do you currently have a use case where we need this ? If not, lets remove this and add it back when we have an actual need for it.

service/frontend/accessControlledWorkflowHandler.go
```diff
@@ -0,0 +1,725 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
venkat1109: [not a blocker ] filename seems rather long - I don't have a suggestion but if you think of a way to shorten it without compromising clarity, go for it.

service/frontend/accessControlledWorkflowHandler.go
```diff
@@ -0,0 +1,725 @@
...
+	"github.com/uber/cadence/.gen/go/cadence/workflowserviceserver"
+	"github.com/uber/cadence/.gen/go/health"
+	"github.com/uber/cadence/.gen/go/health/metaserver"
+	"github.com/uber/cadence/.gen/go/replicator"
+	"github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/resource"
+)
+
+// TODO(vancexu): add metrics
```
venkat1109: we should have a discussion of how you plan to add metrics i.e. who emits requests/error count esp. on a per domain level when you have this decorator pattern going i.e. ideally, I would like to avoid creating domain tagged scope twice - one here and another within workflowHandler API.  

Alternative approach would be to not use decorator pattern and instead have an authorizer wrapper that takes a "request interface{}" and uses type interference to find out the API - it will likely be way less lines of code. Let's get this PR in and discuss this separately.

PR #2906, opened 2019-12-12
Backward compatible feature version check when clientImpl is not set

andrewjdawson2016:


common/client/versionChecker.go
```diff
@@ -135,7 +139,27 @@ func (vc *versionChecker) SupportsConsistentQuery(clientImpl string, clientFeatu
...
+			version, err := version.NewVersion(clientFeatureVersion)
+			if err != nil {
+				return &shared.ClientVersionNotSupportedError{FeatureVersion: clientFeatureVersion}
+			}
+			if !vc.stickyQueryUnknownImplConstraints.Check(version) {
+				return &shared.ClientVersionNotSupportedError{FeatureVersion: clientFeatureVersion, SupportedVersions: vc.stickyQueryUnknownImplConstraints.String()}
+			}
+			return nil
+		default:
+			panic(fmt.Sprintf("unknown feature: %v", feature))
```
venkat1109: how about returning a NotSupportedError instead of panic ?

PR #2907, opened 2019-12-13
Release strict assertion for UpdateCurrentWorkflowByPass for marking workflow as corrupted state

longquanzheng:
This is a pre step for implementing a scanner to mark workflow as corrupted and then eventually delete it.


wxing1292:
assuming that corrupted state is rare, the solution is fine
plz also update `IsWorkflowExecutionRunning` in mutable state builder to return `false` in this case
plz also update `IsCurrentWorkflowGuaranteed` in mutable state builder to return `false` in this case

PR #2908, opened 2019-12-13
Consistent query bug fixes

andrewjdawson2016:
- BadRequest error needs to be returned if query enters invalid state or client is not supported this ensures client side does not keep retrying queries forever. 
- We must not unblock or answer queries if the decision is a heartbeat decision because that can violate consistency of query.

### DIFF
--- a/service/history/decisionHandler.go
+++ b/service/history/decisionHandler.go
@@ -550,7 +550,14 @@ Update_History_Loop:
 			return nil, updateErr
 		}
 
-		handler.handleBufferedQueries(msBuilder, clientImpl, clientFeatureVersion, req.GetCompleteRequest().GetQueryResults(), createNewDecisionTask, domainEntry)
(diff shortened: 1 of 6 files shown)


service/history/decisionHandler.go
```diff
@@ -682,6 +690,12 @@ func (handler *decisionHandlerImpl) handleBufferedQueries(
 		return
 	}
 
+	// if its a heartbeat decision it means local activities may still be running on the worker
+	// which were started by an external event which happened before the query
+	if decisionHeartbeating {
+		return
```
venkat1109: does this mean the queries are left to timeout or is it possible that after the local-activities are finished, these queries will be unblocked ?

andrewjdawson2016: Yes its possible they will become answered. Basically a query can be answered on any decision that is created during the lifetime of it living in query registry.

PR #2910, opened 2019-12-13
Added API for getting partiton information for scalable tasklist

anish531213:


client/matching/loadbalancer.go
```diff
@@ -56,6 +56,14 @@ type (
 			taskListType int,
 			forwardedFrom string,
 		) string
+
+		// GetAllPartitions returns the list of all the partitions for a given tasklist
+		// If the taskList is not scalable it will return a empty tasklist
```
venkat1109: if the taskList is not *partitioned*,

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,24 @@ struct DescribeTaskListResponse {
   20: optional TaskListStatus taskListStatus
 }
 
+struct GetTaskListPartitionInfoResponse {
+  10: optional PartitionInfo activityTaskPartition
+  20: optional PartitionInfo decisionTaskPartition
+}
+
+struct PartitionInfo {
+  10: optional bool isScalableTaskList
```
venkat1109: Do you need this field ? Every task list has atleast one partition i.e. the root partition.

venkat1109:
Do you plan to have a frontend API for this in a separate diff ? Without it, this change probably will not work.

client/matching/loadbalancer.go
anish531213: Sorry the comment is misleading, If the tasklist is not partitioned, I am returning array with tasklist name only. I am returning empty list if there is an error finding the domain name.

idl/github.com/uber/cadence/shared.thrift
anish531213: I think you are right. Is it true that all the task list are scalable but some can have only one partition? If so, there is no need for this field.

anish531213:
> Do you plan to have a frontend API for this in a separate diff ? Without it, this change probably will not work.

client/matching/loadbalancer.go
```diff
@@ -134,3 +142,34 @@ func (lb *defaultLoadBalancer) pickPartition(
...
+	}
+
+	partitionKeys = append(partitionKeys, taskListName)
+
+	n := lb.nWritePartitions(domainName, taskListName, taskListType)
+	if n <= 0 {
+		return partitionKeys, nil
+	}
+
+	for i := 1; i < n; i++ {
```
yycptt: Off by one error?

idl/github.com/uber/cadence/shared.thrift
venkat1109: Yes, from a mental model perspective, you can think of all task lists as partitioned task lists and by default, there is only one partition

client/matching/loadbalancer.go
venkat1109: That's weird, whether the task list has 1 or 100 partitions, you should return the same set of result params from this API on success

client/matching/loadbalancer.go
```diff
@@ -134,3 +142,34 @@ func (lb *defaultLoadBalancer) pickPartition(
...
+	taskListType int,
+) ([]string, error) {
+	domainName, err := lb.domainIDToName(domainID)
+	var partitionKeys []string
+	if err != nil {
+		return partitionKeys, err
+	}
+
+	taskListName := taskList.GetName()
+	if strings.HasPrefix(taskListName, taskListPartitionPrefix) {
```
venkat1109: I believe this is used in more than one place in this file, can you separate this logic into its own helper method for clarity ?

common/clientCache.go
```diff
@@ -57,6 +58,14 @@ func NewClientCache(
 	}
 }
 
+func (c *clientCacheImpl) GetHostInfoForClient(key string) (string, error) {
```
venkat1109: Should this be called GetHostNameForKey instead ? The input is key and the output is hostname

common/clientCache.go
```diff
@@ -57,6 +58,14 @@ func NewClientCache(
...
+	clientKey, err := c.keyResolver(key)
```
venkat1109: s/clientKey/hostname

client/matching/loadbalancer.go
anish531213: It is off by one because there is no prefix for the first partition.

client/matching/loadbalancer.go
anish531213: Yes. That's true. Will update.

idl/github.com/uber/cadence/matching.thrift
```diff
@@ -101,6 +101,11 @@ struct DescribeTaskListRequest {
   20: optional shared.DescribeTaskListRequest descRequest
 }
 
+struct GetTaskListPartitionInfoRequest {
```
venkat1109: where is the API signature defined ? I don't see it in this diff

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,24 @@ struct DescribeTaskListResponse {
   20: optional TaskListStatus taskListStatus
 }
 
+struct GetTaskListPartitionInfoResponse {
```
venkat1109: IMO, ListTaskListPartitions() is a better name for this API. GetXXX is typically used to read/return info about a specific entity.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,24 @@ struct DescribeTaskListResponse {
...
+  10: optional PartitionInfo activityTaskPartition
+  20: optional PartitionInfo decisionTaskPartition
+}
+
+struct PartitionInfo {
```
venkat1109: Do you really need this struct ? Why not declare the map<> inside your Response struct ? If you do want to keep this struct, I would suggest not calling it PartitionInfo - because the struct contains info about all partitions not just one.

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,24 @@ struct DescribeTaskListResponse {
...
+  10: optional bool isScalableTaskList
+  20: optional map<PartitionKey, HostInfo> partitionHostInfo
+}
+
+struct PartitionKey {
+  10: optional string partitionKey
+}
+
+struct HostInfo {
+  10: optional string hostAddress
```
venkat1109: is this address or name ? double check

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,23 @@ struct DescribeTaskListResponse {
   20: optional TaskListStatus taskListStatus
 }
 
+struct ListTaskListPartitionsResponse {
+  10: optional Partitions activityTaskPartitions
+  20: optional Partitions decisionTaskPartitions
+}
+
+struct Partitions {
+  20: optional map<PartitionKey, HostInfo> partitionHostInfo
```
yux0: 10:

client/matching/client.go
```diff
@@ -199,6 +199,53 @@ func (c *clientImpl) DescribeTaskList(ctx context.Context, request *m.DescribeTa
 	return client.DescribeTaskList(ctx, request, opts...)
 }
 
+func (c *clientImpl) ListTaskListPartitions(ctx context.Context, request *m.ListTaskListPartitionsRequest, opts ...yarpc.CallOption) (*workflow.ListTaskListPartitionsResponse, error) {
+	opts = common.AggregateYarpcOptions(ctx, opts...)
+
+	aTPartitionInfo, err := c.listTaskListPartitions(request, persistence.TaskListTypeActivity)
```
yux0: activityTasklistInfo & decisionTasklistInfo

idl/github.com/uber/cadence/cadence.thrift
```diff
@@ -581,4 +581,17 @@ service WorkflowService {
...
+   /**
+   * ReapplyEvents applies stale events to the current workflow and current run
+   **/
+  shared.ListTaskListPartitionsResponse ListTaskListPartitions(1: shared.ListTaskListPartitionsRequest request)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      3: shared.EntityNotExistsError entityNotExistError,
+      4: shared.LimitExceededError limitExceededError,
+      5: shared.ServiceBusyError serviceBusyError,
+      6: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
```
venkat1109: is this error expected from this API ? If not, just remove it

idl/github.com/uber/cadence/matching.thrift
```diff
@@ -211,4 +216,15 @@ service MatchingService {
         3: shared.EntityNotExistsError entityNotExistError,
         4: shared.ServiceBusyError serviceBusyError,
       )
+
+
+  /**
+  * ListTaskListPartitions returns a map of patitionKey and hostAddress for a taskList
```
venkat1109: typo on partitionKey

idl/github.com/uber/cadence/shared.thrift
```diff
@@ -1468,6 +1468,28 @@ struct DescribeTaskListResponse {
...
+  10: optional string domain
+  20: optional TaskList taskList
+}
+
+struct ListTaskListPartitionsResponse {
+  10: optional Partitions activityTaskPartitions
+  20: optional Partitions decisionTaskPartitions
+}
+
+struct Partitions {
```
venkat1109: This struct seems a bit weird to me from composability standpoint. Here is one proposal to simplify things for clarity, let me know what you think:

```
struct TaskListPartitionMetadata {
    string name or string key
    string ownerHostName or string owner
}

struct ListTaskListPartitionsResponse {
    list<TaskListPartitionMetadata> activityTaskListPartitions
    list<TaskListPartitionMetadata> decisionTaskListPartitions
}
```

```

File: service/matching/matchingEngine.go [hunk 16]
```diff
@@ -556,6 +562,81 @@ func (e *matchingEngineImpl) DescribeTaskList(ctx context.Context, request *m.De
...
+}
+
+func (e *matchingEngineImpl) getAllPartitions(
+	domain string,
+	taskList workflow.TaskList,
+	taskListType int,
+) ([]string, error) {
+	var partitionKeys []string
+	taskListName := taskList.GetName()
+	if strings.HasPrefix(taskListName, taskListPartitionPrefix) {
```
venkat1109: see newTaskListName(taskList.GetName()) in tasklist.go - it has an IsRoot() method. You can also add a GetRoot() or Root() method to return the name of the root partition.

idl/github.com/uber/cadence/shared.thrift
anish531213: fixed

client/matching/client.go
anish531213: fixed

idl/github.com/uber/cadence/cadence.thrift
anish531213: Yep, This API is not for the client. So, removed this error.

idl/github.com/uber/cadence/matching.thrift
anish531213: fixed.

idl/github.com/uber/cadence/shared.thrift
anish531213: Yeah, that seems more clear and possible good structure for my API information.

service/matching/matchingEngine.go
anish531213: Added GetRoot method to get the root partition.

service/matching/matchingEngine.go
```diff
@@ -105,6 +108,7 @@ func NewEngine(taskManager persistence.TaskManager,
 	logger log.Logger,
 	metricsClient metrics.Client,
 	domainCache cache.DomainCache,
+	monitor membership.Monitor,
```
venkat1109: can you take ServiceResolver as input instead ? Looks like top level resource object has a GetMatchingServiceResolver() method already

service/matching/handler.go
```diff
@@ -76,6 +76,7 @@ func NewHandler(
 			resource.GetLogger(),
 			resource.GetMetricsClient(),
 			resource.GetDomainCache(),
+			resource.GetMembershipMonitor(),
```
venkat1109: see comment below - consider passing resource.GetMatchingServiceResolver() here

service/matching/tasklist.go
```diff
@@ -82,6 +82,11 @@ func (tn *qualifiedTaskListName) IsRoot() bool {
 	return tn.partition == 0
 }
 
+// GetRoot returns the root name for a task list
+func (tn *qualifiedTaskListName) GetRoot() string {
```
venkat1109: can you also update tasklist_test.go to exercise this new method ?

PR #2912, opened 2019-12-16
Add acquire shard time metric

vitarb:
Record a metric that measures time since last shard update each time when ownership changes.

### DIFF
--- a/service/history/shardContext.go
+++ b/service/history/shardContext.go
@@ -54,6 +54,7 @@ type (
 		GetThrottledLogger() log.Logger
 		GetMetricsClient() metrics.Client
 		GetTimeSource() clock.TimeSource
+		ShardOwnershipChanged() bool
 
 		GetEngine() Engine
 		SetEngine(Engine)
@@ -66,6 +67,7 @@ type (
 
 		SetCurrentTime(cluster string, currentTime time.Time)
 		GetCurrentTime(cluster string) time.Time
+		GetLastUpdatedTime() time.Time
 		GetTimerMaxReadLevel(cluster string) time.Time
 
 		GetTransferAckLevel() int64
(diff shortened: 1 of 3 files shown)


service/history/shardContext.go
```diff
@@ -127,6 +129,7 @@ type (
 
 		// exist only in memory
 		standbyClusterCurrentTime map[string]time.Time
+		shardOwnershipChanged     bool
```
venkat1109: ownershipChanged is a bit confusing to me - (1) does it mean this shard is no longer valid because ownership has moved to another host or (2) does it mean this shard is loaded because of a previous ownership change. If I understand right, you want (2) ? Is that right ? If that's the case, how do you feel about storing "previousOwner" here instead of this flag.

service/history/shardController.go
```diff
@@ -391,6 +391,10 @@ func (i *historyShardsItem) getOrCreateEngine(shardClosedCh chan<- int) (Engine,
 		if err != nil {
 			return nil, err
 		}
+		if context.ShardOwnershipChanged() {
+			i.GetMetricsClient().RecordTimer(metrics.ShardInfoScope, metrics.ShardItemAcquisitionLatency,
+				context.GetCurrentTime(i.GetClusterMetadata().GetCurrentClusterName()).Sub(context.GetLastUpdatedTime()))
+		}
```
venkat1109: I am not sure if this can give you accurate value for shard_ownership_change time - UpdateShard is called only every minute unless we run out of rangeID space, in which case, it is immediately invoked.

vitarb: Yep this is an approximation, I plan to also add code to ensure that shards are updated during shutdown to ensure most up to date timestamps.

service/history/shardContext.go
vitarb: This flag means that while we were trying to acquire a shard, record that we've loaded from the database had a different owner than the current host's identity. Hence it means that shard ownership has changed and this is the first time that we've discovered it.

PR #2913, opened 2019-12-16
Add initial support for custom SQL TLS configuration

bjacobso:
This is an attempt to add support for TLS MySQL (we use RDS/Aurora and require TLS support) using the recently added TLS support for Cassandra. In the process it reuses the config in persistence for use in tools and removes ConnectParams.

For the cli you can use env:
```
SQL_TLS=1 SQL_TLS_CA_FILE=./my-ca-file.pem SQL_CONNECT_ATTRIBUTES="tx_isolation=READ-COMMITTED" ./cadence-sql-tool ...
```
or command flags:
```
./cadence-sql-tool ... --tls --tls-ca-file ./my-ca-file.pem --plugin mysql --connect-attributes "tx_isolation=READ-COMMITTED" create ...
```

And `tls:` can now be used in config.yaml for running `cadence-server`

Feedback is welcome!

_Closing previous attempt, which was before major refactor related to postgres support:
https://github.com/uber/cadence/pull/2841_

### DIFF
--- a/common/persistence/sql/sqlplugin/mysql/plugin.go
+++ b/common/persistence/sql/sqlplugin/mysql/plugin.go
@@ -22,10 +22,15 @@ package mysql
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
+	"net"
 	"net/url"
 	"strings"
 
+	"github.com/go-sql-driver/mysql"
 	"github.com/iancoleman/strcase"
 	"github.com/jmoiron/sqlx"
 
@@ -43,6 +48,8 @@ const (
 	isolationLevelAttrName       = "transaction_isolation"
 	isolationLevelAttrNameLegacy = "tx_isolation"
 	defaultIsolationLevel        = "'READ-COMMITTED'"
+	// customTLSName is the name used if a custom tls configuration is created
+	customTLSName = "tls-custom"
 )
 
 var dsnAttrOverrides = map[string]string{
@@ -84,6 +91,11 @@ func (p *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	err := registerTLSConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
 	db, err := sqlx.Connect(PluginName, buildDSN(cfg))
 	if err != nil {
 		return nil, err
@@ -97,9 +109,70 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
 	if cfg.MaxConnLifetime > 0 {
 		db.SetConnMaxLifetime(cfg.MaxConnLifetime)
 	}
+
 	// Maps struct names in CamelCase to snake without need for db struct tags.
 	db.MapperFunc(strcase.ToSnake)
 	return db, nil
+}
+
+func registerTLSConfig(cfg *config.SQL) error {
+	if cfg.TLS == nil || !cfg.TLS.Enabled {
+		return nil
+	}
+
+	host, _, err := net.SplitHostPort(cfg.ConnectAddr)
+	if err != nil {
+		return fmt.Errorf("error in host port from ConnectAddr: %v", err)
+	}
+
+	// TODO: create a way to set MinVersion and CipherSuites via cfg.
+	tlsConfig := &tls.Config{
+		ServerName: host,
+	}
+
+	if cfg.TLS.CaFile != "" {
+		rootCertPool := x509.NewCertPool()
+		pem, err := ioutil.ReadFile(cfg.TLS.CaFile)
+		if err != nil {
+			return fmt.Errorf("failed to load CA files: %v", err)
+		}
+		if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+			return fmt.Errorf("failed to append CA file")
+		}
+		tlsConfig.RootCAs = rootCertPool
+	}
+
+	if cfg.TLS.CertFile != "" && cfg.TLS.KeyFile != "" {
+		clientCert := make([]tls.Certificate, 0, 1)
+		certs, err := tls.LoadX509KeyPair(
+			cfg.TLS.CertFile,
+			cfg.TLS.KeyFile,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to load tls x509 key pair: %v", err)
+		}
+		clientCert = append(clientCert, certs)
+		tlsConfig.Certificates = clientCert
+	}
+
+	// In order to use the TLS configuration you need to register it. Once registered you use it by specifying
+	// `tls` in the connect attributes.
+	err = mysql.RegisterTLSConfig(customTLSName, tlsConfig)
+	if err != nil {
+		return fmt.Errorf("failed to register tls config: %v", err)
+	}
+
+	if cfg.ConnectAttributes == nil {
+		cfg.ConnectAttributes = map[string]string{}
+	}
+
+	// If no `tls` connect attribute is provided then we override it to our newly registered tls config automatically.
+	// This allows users to simply provide a tls config without needing to remember to also set the connect attribute
+	if cfg.ConnectAttributes["tls"] == "" {
+		cfg.ConnectAttributes["tls"] = customTLSName
+	}
+
+	return nil
 }
 
 func buildDSN(cfg *config.SQL) string {
--- a/common/persistence/sql/sqlplugin/postgres/plugin.go
+++ b/common/persistence/sql/sqlplugin/postgres/plugin.go
@@ -21,6 +21,7 @@
 package postgres
 
 import (
+	"errors"
 	"fmt"
 	"net"
 
@@ -37,6 +38,8 @@ const (
 	PluginName             = "postgres"
 	dataSourceNamePostgres = "user=%v password=%v host=%v port=%v dbname=%v sslmode=disable "
 )
+
+var errTLSNotImplemented = errors.New("tls for postgres has not been implemented")
 
 type plugin struct{}
 
@@ -71,6 +74,11 @@ func (d *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (d *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	err := registerTLSConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
 	host, port, err := net.SplitHostPort(cfg.ConnectAddr)
 	if err != nil {
 		return nil, fmt.Errorf("invalid connect address, it must be in host:port format, %v, err: %v", cfg.ConnectAddr, err)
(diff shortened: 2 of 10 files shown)


common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -97,11 +111,67 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
...
+		return fmt.Errorf("failed to register tls config: %v", err)
+	}
+
+	if cfg.ConnectAttributes == nil {
+		cfg.ConnectAttributes = map[string]string{}
+	}
+
+	if cfg.ConnectAttributes["tls"] == customTLSName || cfg.ConnectAttributes["tls"] == "" {
+		// Use the new custom tls configuration
+		cfg.ConnectAttributes["tls"] = customTLSName
```
bjacobso: This is needed because the way you tell mysql to use your newly registered global TLS config is by adding `tls=my_config_name` to you connection attributes. I add that automatically so you don't need to set that as well as the TLS options.

bjacobso:
@venkat1109 I believe I addressed all your comments from my last PR

@longquanzheng indeed this was a lot easier after your postgres changes, thanks!

tools/common/schema/types.go
```diff
@@ -84,6 +84,8 @@ const (
 	CLIOptDatabase = "database"
 	// CLIOptPluginName is the cli option for plugin name
 	CLIOptPluginName = "plugin"
+	// CLIOptConnectAttributes is the cli option for connect attributes (key/values via a url query string)
+	CLIOptConnectAttributes = "connect-attributes"
```
bjacobso: @venkat1109 I went with your previous suggestion of allowing arbitrary connect attributes: in this case it is via a string in the format of `key1=value1&key2=value2`

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -91,3 +98,8 @@ func (d *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
 	db.MapperFunc(strcase.ToSnake)
 	return db, nil
 }
+
+// TODO: implement postgres specific support for TLS
+func registerTLSConfig(cfg *config.SQL) error {
+	return fmt.Errorf("tls for postgres has not been implemented")
```
bjacobso: I'm not too familiar with how postgres tls works so I left this unimplemented, but I could research more

bjacobso:
Thoughts this change @venkat1109 and @longquanzheng?

common/persistence/sql/sqlplugin/postgres/plugin.go
samarabbas: nit: why using Errorf? Probably better to statically define this as a const in the file and return it here.  We have similar pattern in other parts of the code.

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -71,6 +71,13 @@ func (d *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (d *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	if cfg.TLS != nil && cfg.TLS.Enabled {
+		err := registerTLSConfig(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("error registering TLS config: %v", err)
```
samarabbas: nit: I don't see the point of wrapping the error we return in "registerTLSConfig" in a different error here.  Can you just return the same error to caller?

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -84,6 +91,13 @@ func (p *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	if cfg.TLS != nil && cfg.TLS.Enabled {
+		err := registerTLSConfig(cfg)
+		if err != nil {
+			return nil, fmt.Errorf("error registering TLS config: %v", err)
```
samarabbas: nit: do you think wrapping the 'registerTLSConfig' error adds any value here?

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -97,11 +111,67 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
...
+	err = mysql.RegisterTLSConfig(customTLSName, tlsConfig)
+	if err != nil {
+		return fmt.Errorf("failed to register tls config: %v", err)
+	}
+
+	if cfg.ConnectAttributes == nil {
+		cfg.ConnectAttributes = map[string]string{}
+	}
+
+	if cfg.ConnectAttributes["tls"] == customTLSName || cfg.ConnectAttributes["tls"] == "" {
```
samarabbas: I'm confused by this check.  Maybe you intended the following:
```
if cfg.ConnectAttributes["tls"] != customTLSName || cfg.ConnectAttributes["tls"] == "" {
```
or just maybe:
```
if cfg.ConnectAttributes["tls"] == "" {
```

samarabbas: Maybe it will help if you can include an example of config when someone specifies tls as ConnectAttributes.

common/persistence/sql/sqlplugin/postgres/plugin.go
bjacobso: agreed, removed

common/persistence/sql/sqlplugin/mysql/plugin.go
bjacobso: I removed the error and used the simpler check you suggested. 

The thinking was that in order to correctly use a tls config you need to register it and tell the connection to use it via a connect attribute. I'm automatically setting the connect attribute here so you don't need to also provide that when using a tls connection. The error originally was intended for a case where someone provided a _different_ tls connect attribute for some reason and also provided a tls config. In that case the tls config will silently not get used. Not sure if that really constitutes an error here.

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -84,6 +91,13 @@ func (p *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	if cfg.TLS != nil && cfg.TLS.Enabled {
```
venkat1109: do this check within the registerTLSConfig. If tls is not enabled, simply return nil from registerTLSConfig

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -97,11 +111,67 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
...
 }
 
+func registerTLSConfig(cfg *config.SQL) error {
+	host, _, err := net.SplitHostPort(cfg.ConnectAddr)
+	if err != nil {
+		return fmt.Errorf("error in host port from ConnectAddr: %v", err)
+	}
+
+	tlsConfig := &tls.Config{
+		ServerName: host, // hostname
```
venkat1109: ServerName is self-explanatory / comment is redundant

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -97,11 +111,67 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
...
+	}
+
+	if cfg.TLS.CertFile != "" && cfg.TLS.KeyFile != "" {
+		clientCert := make([]tls.Certificate, 0, 1)
+		certs, err := tls.LoadX509KeyPair(
+			cfg.TLS.CertFile,
+			cfg.TLS.KeyFile,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to load key pair: %v", err)
```
venkat1109: failed to load tls x509 key pair

common/persistence/sql/sqlplugin/mysql/plugin.go
```diff
@@ -97,11 +111,67 @@ func (p *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
...
 	return db, nil
 }
 
+func registerTLSConfig(cfg *config.SQL) error {
+	host, _, err := net.SplitHostPort(cfg.ConnectAddr)
+	if err != nil {
+		return fmt.Errorf("error in host port from ConnectAddr: %v", err)
+	}
+
+	tlsConfig := &tls.Config{
```
venkat1109: I would strongly suggest specifying the following parameters: 

- MinVersion for the minimum TLS version that's acceptable.  By default, the minVersion is 1.0, which pretty much should not be used. The default value for this should be 1.3. 
- CipherSuites - Ideally, this should also be set. TLS 1.3 already prevents you from choosing bad ciphersuites

I am ok if you punt on this change for this PR. But please create a github task to take these as config params

common/persistence/sql/sqlplugin/postgres/plugin.go
```diff
@@ -71,6 +74,13 @@ func (d *plugin) CreateAdminDB(cfg *config.SQL) (sqlplugin.AdminDB, error) {
 // SQL database and the object can be used to perform CRUD operations on
 // the tables in the database
 func (d *plugin) createDBConnection(cfg *config.SQL) (*sqlx.DB, error) {
+	if cfg.TLS != nil && cfg.TLS.Enabled {
```
venkat1109: move this inside registerTLSConfig

common/persistence/sql/sqlplugin/mysql/plugin.go
bjacobso: It looks like AWS Aurora supports up to 1.2 which would be the minimum for our use case. I can hard code 1.2 if you think that is ok for other users? I think the best way would be allowing the configuration of those options via the cfg but that would take more plumbing and would need to take in account TLS for Cassandra and Kafka because I am sharing the config object.

For now I left it unchanged (with a TODO comment) and will create an issue.

venkat1109: Let's tackle this separately via config. For now, I am ok with using the defaults

PankajWorks:
@bjacobso First of all thanks for this PR . I have 2 immediate questions here. 
- If the certificate is signed and its not same as hostname how the user can configure it.  
- If the user specifies the reserved config values i.e. one of ["skip-verify", "preferred", "true", "false"] what should be the behaviour?

bjacobso:
@PankajWorks 

> If the certificate is signed and its not same as hostname how the user can configure it.

It looks like my PR doesn't take this into account unfortunately. Are you referring to this:
https://github.com/uber/cadence/pull/2974/files#diff-460ccb997687fa1effc1307704bde981R215-R217

You would need to add a host option to the TLS struct here:
https://github.com/uber/cadence/pull/2913/files#diff-6dc12e8c80a8118bbc5213ff428682c3R38

> If the user specifies the reserved config values i.e. one of ["skip-verify", "preferred", "true", "false"] what should be the behaviour?

In this case that value will take precedent and the TLS configuration won't be used. Currently this doesn't return an error or anything it will just be registered but not used.

PankajWorks:
@bjacobso it will good to have an variable for ServerName to override with.

PR #2915, opened 2019-12-17
Expose NDC doc

wxing1292:


docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
+# Design doc: Cadence NDC
+
+Last updated: Dec 2019
+
+Reference: [#2290](https://github.com/uber/cadence/issues/2290)
+
+
+## Abstract
+
+Cadence Cross DC is a feature which asynchronously replicates workflows from active data center to other passive data centers, for reconstruction. When necessary, customer can failover to any of the data centers which has the backup for high availability.
```
yux0: the data centers which have

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+2. Version history
+3. Workflow history conflict resolution
+4. Zombie workflows
+5. Workflow task processing
+
+### Version
+
+Version is a newly introduced concept in Cadence Cross DC which describes the chronological order of events (per customer domain).
+
+Cadence Cross DC is AP, all domain change events & workflow history events are replicated asynchronously for high throughput. This means that data across data centers are not strongly consistent. To guarantee that domain data & workflow data will achieve eventual consistency (expecially when there is data conflict during a failover), version is introduced and attached to customers' domains. All workflow history events generated in a domain will also come with the version in that domain.
```
yux0: expecially -> especially

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+
+All participating data centers are pre-configured with a unique initial version, and a shared version increment:
+
+  `initial version < shared version increment`
+
+When performing failover for one domain from one data center to another data center, the version attched to the domain will be changed by the following rule:
```
yux0: attched -> attached

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+Below will show version history of the same workflow in 2 different data centers.
+
+T = 0:  existing version history in data center A & B
+```
+| -------- | ------------- | --------------- | ------- |
+| Events                   | Version History           |
+| -------- | ------------- | --------------- | ------- |
+| Event ID | Event Version | Event ID        | Version |
+| -------- | ------------- | --------------- | ------- |
+| 1        | 1             | 2               | 1       |
```
yux0: Might be easier to understand by use the real version. Something like data center A version: 1, data center B version: 2. version increment = 3.

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+  `(vesion in domain) % (shared version increment) == (this data centers' initial version)`
+  2. the version of this workflow's last event is not less then version in domain, i.e.
+  `(last event's version) <= (version in domain)`
+
+
+### Version History
+
+Version history is a newly introduced concept which provides high level summary about version information of workflow history.
+
+Whenever there is a new workflow history event generated, the version from domain will be attached. Workflow mutable state will keep track of all hisory events & corresponding version.
```
yux0: hisory -> history

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+
+1. Version
+2. Version history
+3. Workflow history conflict resolution
+4. Zombie workflows
+5. Workflow task processing
+
+### Version
+
+Version is a newly introduced concept in Cadence Cross DC which describes the chronological order of events (per customer domain).
```
meiliang86: Need to explicitly call out that version is attached to domain, not cluster, not workflows.
Adding an example would help as well.

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+  find the smallest version which has `version >= old version in domain`
+
+When there is a data conflict, comparison will be made and workflow history events with the highest version will win.
+
+When a data center is trying to mutate a workflow, version will be checked. A data center can mutate a workflow if and only if
+
+  1. version in the domain belongs to this data center, i.e.
+  `(vesion in domain) % (shared version increment) == (this data centers' initial version)`
+  2. the version of this workflow's last event is not less then version in domain, i.e.
+  `(last event's version) <= (version in domain)`
```
meiliang86: ```>=```

meiliang86: if 1 is satisfied, what is the case that can lead to last event's version > version in domain?

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+| -------- | ------------- | --------------- | ------- |
+| Event ID | Event Version | Event ID        | Version |
+| -------- | ------------- | --------------- | ------- |
+| 1        | 1             | 3               | 1       |
+| 2        | 1             |                 |         |
+| 3        | 1             |                 |         |
+| -------- | ------------- | --------------- | ------- |
+```
+
+T = 3:  domain failover triggered, domain version is now 2
```
meiliang86: Does it matter whether this is active or passive? Or are they the same (eventually)?
Maybe worth calling out explicitly.

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+```
+| -------- | ------------- | --------------- | ------- |
+| Events                   | Version History           |
+| -------- | ------------- | --------------- | ------- |
+| Event ID | Event Version | Event ID        | Version |
+| -------- | ------------- | --------------- | ------- |
+| 1        | 1             | 3               | 1       |
+| 2        | 1             | 5               | 2       |
+| 3        | 1             |                 |         |
+| 4        | 2             |                 |         |
```
meiliang86: Missing 5 in events.

docs/design/2290-cadence-ndc.md
```diff
@@ -0,0 +1,310 @@
...
+        |                          | -----------------------> |
+        |                          |                          |
+| ------------- |          | ------------- |          | ------------- |
+| Data Center A |          | Network Layer |          | Data Center B |
+| ------------- |          | ------------- |          | ------------- |
+```
+
+Since run 2 appears in data center B first, run 1 cannot be replicated as runnable due to rule `at most one run open` (see above), thus, `zombie` workflow state is introduced. Zombie state indicates a workflow which cannot be actively mutated by a data center (assuming corresponding domain is active in this data center). A zombie workflow can only be changed by replication task.
+
+Run 1 will be replicated similar as run 2, except run 1's workflow state will be zombie before run 1 reaches completion.
```
meiliang86: Just curious, where do we keep this state? In DB?

wxing1292: all persisted to database.

a workflow can potentially turn from a zombie workflow to current running workflow (in this data center), if say there is a data center which only has run 1 (no run 2) & the run 1 is running with a higher version.

docs/design/2290-cadence-ndc.md
wxing1292: all replication tasks are handled asynchronously, meaning that domain replication (change of version) can experience delay.

docs/design/2290-cadence-ndc.md
wxing1292: added an example

PR #2925, opened 2019-12-19
Implement mutableState scanner for corruption workflows

longquanzheng:


common/persistence/dataInterfaces.go
```diff
@@ -1436,6 +1460,7 @@ type (
 		CreateShard(request *CreateShardRequest) error
 		GetShard(request *GetShardRequest) (*GetShardResponse, error)
 		UpdateShard(request *UpdateShardRequest) error
+		ScanWorkflows(request *ScanWorkflowsRequest) (*ScanWorkflowsResponse, error)
```
venkat1109: How about calling this ListXXX() to stick with our naming convention of create/read-get/update/delete/list ?

common/persistence/dataInterfaces.go
```diff
@@ -1424,6 +1424,30 @@ type (
...
+		WorkflowExecutions []ScanWorkflowDetailInfo
+	}
+
+	// ScanWorkflowDetailInfo is detail of ScanWorkflowsResponse
+	ScanWorkflowDetailInfo struct{
+		DomainID string
+		WorkflowID string
+		RunID  string
+		State  int
+		IsCurrentWorkflow bool
```
venkat1109: this seems very cassandra specific i.e. in cassandra, both mutable state and current_record are stored in the executions table. But this will not be the case in SQL based stores. In mySQL for example, we have separate tables for execution and current_execution. Have the contract for this API return results from both is a bit confusing as well as will complicate pageToken implementation for other stores. We have two options:

 - Create separate APIs
 -  Add request filters to indicate what type of results is desired

service/worker/scanner/workflow/scavenger.go
```diff
@@ -0,0 +1,345 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package workflow
+
+import (
+	"context"
+	"github.com/uber/cadence/.gen/go/shared"
```
venkat1109: goimports formatting

service/worker/scanner/workflow/scavenger.go
```diff
@@ -0,0 +1,345 @@
...
+var persistenceOperationRetryPolicy = common.CreatePersistanceRetryPolicy()
+
+// Run runs the scavenger
+func (s *Scavenger) Run(ctx context.Context) (ScavengerHeartbeatDetails, error) {
+	taskCh := make(chan taskDetail, pageSize)
+	respCh := make(chan taskResult, pageSize)
+	concurrency := s.rps/rpsPerConcurrency + 1
+
+	for i := 0; i < concurrency; i++ {
+		go s.startTaskProcessor(ctx, taskCh, respCh)
```
venkat1109: consider using the Executor class in service/worker/executor package. If you use that package, you don't need to manually manage go routine pool / channels etc. However, the heartbeating mechanism used here needs to be modified a bit (which I think should be fine)

service/worker/scanner/workflow/scavenger.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	err = backoff.Retry(op, persistenceOperationRetryPolicy, common.IsPersistenceTransientError)
+	if err == nil{
+		// not corrupted
+		return false, nil
+	}
+	if _, ok := err.(*shared.InternalDataCorruptionError); !ok{
+		// some other error
+		return false, err
```
venkat1109: I believe this should be true, nil

service/worker/scanner/workflow/scavenger.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	// for corrupted, we need to
+	// 1. delete current record
+	// 2. covert into corrupted state
+	// TODO  https://github.com/uber/cadence/issues/2926
+	return true, nil
```
venkat1109: false,err ?

service/worker/scanner/workflow/scavenger.go
```diff
@@ -0,0 +1,345 @@
...
+				case <-ctx.Done():
+					return s.hbd, ctx.Err()
+				}
+			}
+		}
+
+		s.hbd.CurrentPage++
+		s.hbd.NextPageToken = resp.NextPageToken
+		s.hbd.ErrorCount += errorCount
+		s.hbd.ProcessedCount += len(resp.WorkflowExecutions)
```
venkat1109: Would be great if these are domain specific metrics. To achieve that, you need to emit these metrics from within the task handler

common/persistence/dataInterfaces.go
andrewjdawson2016: Let's chat about this one more. The hard part here is this API is meant to abstract out the underlying storage technology, but by virtue of different storage technologies potentially using different schemas it means the response is not uniform across technologies.

PR #2928, opened 2019-12-20
Update gocql to pull in https://github.com/gocql/gocql/pull/1321

AngerM:
Updates gocql to pull in https://github.com/gocql/gocql/pull/1321


yux0:
Looks good to me. Could you also update the failed test? The failed test is common/persistence/persistence-tests/historyV2PersistenceTest.go

Update the line 164 from "if resp.NextPageToken == nil {" to "if len(resp.NextPageToken) == 0 {"
should fix the test

PR #2931, opened 2019-12-20
Add more options to reset batch: skip_base_not_current, non_determini…

longquanzheng:
…stic_only

more improvements with dryrun and originalTimestamp

dry run logs


tools/cli/workflowCommands.go
```diff
@@ -1631,44 +1647,106 @@ func doReset(c *cli.Context, domain, wid, rid string, reason, resetType string,
...
-		Reason:                common.StringPtr(fmt.Sprintf("%v:%v", getCurrentUserFromEnv(), reason)),
-	})
+		MaximumPageSize: common.Int32Ptr(1000),
+		NextPageToken:   nil,
+	}
 
-	if err != nil {
-		return printErrorAndReturn("ResetWorkflowExecution failed", err)
+	var firstEvent, decisionFailed *shared.HistoryEvent
+	for {
```
yux0: Talked offline. Consider use paginator iterator

longquanzheng: The command is already implemented with raw for loop, changing all of them is a little bit risky(I don't have time to test it)
let us do it next PR: https://github.com/uber/cadence/issues/2945

PR #2933, opened 2019-12-21
Fix NDC workflow events reapplication issue

wxing1292:
* Allow NDC workflow events reapplication resetting current closed workflow.
* Workflow with no decision task started event still pending #2420

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2811,8 +2811,8 @@ func (e *historyEngineImpl) ReapplyEvents(
 				resetRunID := uuid.New()
 				baseRebuildLastEventID := mutableState.GetPreviousStartedEventID()
 
-				// TODO when https://github.com/uber/cadence/issues/2420 is finished
-				//  remove this block
+				// TODO when https://github.com/uber/cadence/issues/2420 is finished, remove this block,
+				//  since cannot reapply event to a finished workflow which had no decisions started
 				if baseRebuildLastEventID == common.EmptyEventID {
 					e.logger.Warn("cannot reapply event to a finished workflow",
 						tag.WorkflowDomainID(domainID),
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2814,6 +2814,11 @@ func (e *historyEngineImpl) ReapplyEvents(
 				// TODO when https://github.com/uber/cadence/issues/2420 is finished
 				//  remove this block
 				if baseRebuildLastEventID == common.EmptyEventID {
+					e.logger.Warn("cannot reapply event to a finished workflow",
+						tag.WorkflowDomainID(domainID),
+						tag.WorkflowID(currentExecution.GetWorkflowId()),
+					)
+					e.metricsClient.IncCounter(metrics.HistoryReapplyEventsScope, metrics.EventReapplySkippedCount)
 					return &updateWorkflowAction{noop: true}, nil
 				}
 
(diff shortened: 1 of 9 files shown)


service/history/nDCTransactionMgr.go
```diff
@@ -232,83 +244,119 @@ func (r *nDCTransactionMgrImpl) backfillWorkflow(
...
 
-	return targetWorkflow.getContext().updateWorkflowExecutionWithNew(
-		now,
-		mode,
-		nil,
-		nil,
-		transactionPolicy,
-		nil,
-	)
+	if err := targetWorkflow.getContext().reapplyEvents(
```
venkat1109: reading through the code, its not obvious to me as to what we do here i.e. at this point, either targetWorkflow is not current workflow or this is not the active cluster. What do we now ? Can you add a comment to help the reader ?

service/history/nDCTransactionMgr.go
```diff
@@ -232,83 +244,119 @@ func (r *nDCTransactionMgrImpl) backfillWorkflow(
...
-	// workflow events reapplication to self (self workflow being current workflow)
-	// we need to handle 2 cases
-	// 1. workflow still running -> just reapply
-	// 2. workflow closed -> TODO wait until https://github.com/uber/cadence/issues/2420
-	//  NOTE: also remember that workflow reset will acquire a lock on current workflow (this)
-
-	// simple case workflow still running (implies current workflow),
-	// no workflow reset necessary
-	if targetWorkflowActiveCluster == currentCluster {
-		// target workflow is active && target workflow is current workflow
```
venkat1109: I think this comment is useful, unless it is no longer relevant, please don't delete this.

service/history/historyEngine.go
```diff
@@ -2796,7 +2802,66 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+				// need to reset target workflow (which is also the current workflow)
+				// to accept events to be reapplied
+				baseRunID := mutableState.GetExecutionInfo().RunID
+				resetRunID := uuid.New()
+				baseRebuildLastEventID := mutableState.GetPreviousStartedEventID()
+
+				// TODO when https://github.com/uber/cadence/issues/2420 is finished
+				//  remove this block
+				if baseRebuildLastEventID == common.EmptyEventID {
+					e.logger.Warn("cannot reapply event to a finished workflow",
```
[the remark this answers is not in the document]
wxing1292: that is correct, how would you like to change the comments?

service/history/nDCTransactionMgr.go
wxing1292: sure

service/history/historyEngine.go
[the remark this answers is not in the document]
venkat1109: I just wanted to call out that its a finished workflow _which had no decisions scheduled at all_
One suggestion would be  "cannot reapply event to a finished workflow which had no decisions scheduled"

service/history/historyEngine.go
```diff
@@ -2796,7 +2802,66 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+						e.shard.GetDomainCache(),
+						e.shard.GetClusterMetadata(),
+						context,
+						mutableState,
+						noopReleaseFn,
+					),
+					eventsReapplicationResetWorkflowReason,
+					reapplyEvents,
+				); err != nil {
+					return nil, err
```
yux0: If the workflow cannot be reset due to no decision task, will the reapply retry forever?

wxing1292: line 2816 does the check and give up, until #2420 is finished

PR #2941, opened 2019-12-27
mutableState checkums part#1

venkat1109:
This patch contains part#1 of the changes needed for generating / verifying checksums on workflow mutable state. This patch generates a checksum everytime mutableStateBuilder.closeXXX() methods are called. The generated checksum will be stored in persistence along with mutable state. When workflow mutable state is retrieved from persistence, checksum verification is done within mutableStateBuilder.Load() method. This patch does not contain the following changes (will be done in a followup part#2 PR):

- Changes to persistence to store/retrieve checksum
- Implementation for InvalidateCheckumBefore dynamic config option
- Buffered events to be included as part of checksum payload

__Approach Overview__
On a high level, the proposed solution is to compute a crc checksum on a pseudo payload derived from workflow mutable state. This checksum would be recomputed everytime there is an update to the workflow mutable state. The checksum would be stored alongside the mutable state in persistence. When a workflow is first loaded from mutable state into the cache, pseudo payload will be reconstructed and derived checksum will be verified for correctness.  If there is a mismatch, appropriate remedial action, if possible, would be taken.

To compute mutable state checksums, a pseudo checksum payload is constructed using a subset of fields from mutable state. The reason for doing this to reduce overhead of cost of computing checksums (cpu / heapalloc / latency). Checksum will be computed on the serialized byte array of the payload struct. For verification, the same pseudo header will be reconstructed, checksum generated and compared against the stored checksum. Verification will be performed when workflow mutable state is loaded into cache. In the future, we can consider offline verification of checksum for all workflows in the background via the mutable state scanner.

An alternative to pseudo checksums would have been checksum on the whole serialized mutable state blob. However, the current implementation does not serialize mutable state i.e. its up to the persistence layer to store mutable state as columns across one or more tables or a single blob. Given this, this proposal avoids serializing / deserializing whole mutable state just for the sake for checksum verification. However, if after testing, the cost turns out to be low, we might consider doing this.
 
__Versioning__
Regardless of the serialization format, there needs to be a version associated with the checksum payload. This is to allow for future modifications to the set of fields that comprise the pseudo checksum payload. This version will be stored alongside the checksum in persistence. When a new field needs to be added to payload, a PayloadV2 struct will be created. Likewise, during verification, the payload_version field will be used to decide which struct to use.


idl/github.com/uber/cadence/private.thrift
```diff
@@ -0,0 +1,47 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+namespace java com.uber.cadence.private
```
wxing1292: can we use something other than `private.thrift`?
if we are going to have multiple versions of mac, (hmac, crc), then I would prefer to call this file `checksum.thrift`

idl/github.com/uber/cadence/private.thrift
```diff
@@ -0,0 +1,47 @@
...
+
+struct MutableStateChecksumPayload {
+    10: optional bool cancelRequested
+    15: optional i16 state
+
+    20: optional i64 (js.type = "Long") currentVersion
```
wxing1292: current version is not necessary, since this will be updated each time the mutable state is loaded in memory (updated by domain cache)

idl/github.com/uber/cadence/private.thrift
```diff
@@ -0,0 +1,47 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+namespace java com.uber.cadence.private
+
+struct MutableStateChecksumPayload {
+    10: optional bool cancelRequested
+    15: optional i16 state
```
wxing1292: plz also include the `close status`

service/history/checksum.go
```diff
@@ -0,0 +1,115 @@
...
+		SignalCount:          common.Int64Ptr(int64(executionInfo.SignalCount)),
+		DecisionAttempt:      common.Int32Ptr(int32(executionInfo.DecisionAttempt)),
+		DecisionScheduledID:  common.Int64Ptr(executionInfo.DecisionScheduleID),
+		DecisionStartedID:    common.Int64Ptr(executionInfo.DecisionStartedID),
+		DecisionVersion:      common.Int64Ptr(executionInfo.DecisionVersion),
+		StickyTaskListName:   common.StringPtr(executionInfo.StickyTaskList),
+	}
+
+	if replicationState != nil {
+		payload.CurrentVersion = common.Int64Ptr(replicationState.CurrentVersion)
```
wxing1292: current version is unnecessary

service/history/checksum.go
```diff
@@ -0,0 +1,115 @@
...
+		LastProcessedEventID: common.Int64Ptr(executionInfo.LastProcessedEvent),
+		SignalCount:          common.Int64Ptr(int64(executionInfo.SignalCount)),
+		DecisionAttempt:      common.Int32Ptr(int32(executionInfo.DecisionAttempt)),
+		DecisionScheduledID:  common.Int64Ptr(executionInfo.DecisionScheduleID),
+		DecisionStartedID:    common.Int64Ptr(executionInfo.DecisionStartedID),
+		DecisionVersion:      common.Int64Ptr(executionInfo.DecisionVersion),
+		StickyTaskListName:   common.StringPtr(executionInfo.StickyTaskList),
+	}
+
+	if replicationState != nil {
```
wxing1292: plz also add the support of version history

idl/github.com/uber/cadence/private.thrift
```diff
@@ -0,0 +1,47 @@
...
+    24: optional i64 (js.type = "Long") nextEventID
+    25: optional i64 (js.type = "Long") lastProcessedEventID
+    26: optional i64 (js.type = "Long") signalCount
+
+    35: optional i32 decisionAttempt
+    36: optional i64 (js.type = "Long") decisionVersion
+    37: optional i64 (js.type = "Long") decisionScheduledID
+    38: optional i64 (js.type = "Long") decisionStartedID
+
+    45: optional list<i64> pendingTimerStartedIDs
```
wxing1292: strictly speaking `ID + version` should be used here, but i guess ID should be enough for normal case.

idl/github.com/uber/cadence/private.thrift
venkat1109: done

idl/github.com/uber/cadence/private.thrift
venkat1109: Agreed, will punt on version for now.

PR #2954, opened 2020-01-03
Include branch token and new run branch token in the replication task

yux0:
Include branch token and new run branch token in the replication task to support DLQ merging


yux0:
No need to have branch token

PR #2962, opened 2020-01-06
Fix bench of issues with new lint tool

vancexu:
https://github.com/uber/cadence/issues/2961
When using better lint tool, it throws hundreds of violations.
This PR address half of them, and fixes some bugs majorly around:
1. error check
2. bugs caused by copy paste
3. remove unused and dead code

(canary is marked as nolint for now, since it is not fully functioning)
There will be following PR address other issues.

### DIFF
--- a/cmd/tools/cassandra/main.go
+++ b/cmd/tools/cassandra/main.go
@@ -27,5 +27,5 @@ import (
 )
 
 func main() {
-	cassandra.RunTool(os.Args)
+	cassandra.RunTool(os.Args) //nolint:errcheck
 }
--- a/common/persistence/domainReplicationQueue.go
+++ b/common/persistence/domainReplicationQueue.go
@@ -23,13 +23,11 @@ package persistence
 import (
 	"errors"
 	"fmt"
-	"math"
 	"time"
 
 	"github.com/uber/cadence/.gen/go/replicator"
 	"github.com/uber/cadence/common/codec"
 	"github.com/uber/cadence/common/log"
-	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/metrics"
 )
 
@@ -64,7 +62,6 @@ type (
 		metricsClient       metrics.Client
 		logger              log.Logger
 		encoder             codec.BinaryEncoder
-		ackLevelUpdated     bool
 		ackNotificationChan chan bool
 		done                chan bool
 	}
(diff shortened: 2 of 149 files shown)


service/history/timerQueueProcessorBase.go
```diff
@@ -590,26 +590,3 @@ func (t *timerQueueProcessorBase) deleteWorkflowVisibility(
 	}
 	return backoff.Retry(op, persistenceOperationRetryPolicy, common.IsPersistenceTransientError)
 }
-
-func (t *timerQueueProcessorBase) getTimerTaskType(
```
yycptt: I think this function is useful for logging & debugging purpose, can we keep it?

tools/sql/handler.go
```diff
@@ -170,10 +170,13 @@ func doDropDatabase(p ConnectParams, name string) {
 	p.Database = ""
 	conn, err := NewConnection(&p)
 	if err != nil {
-		handleErr(err)
+		logErr(err)
```
yycptt: I think the logic here should be 
```
if err = handler(err); err != nil {
    logErr(err)
}
```
and logging should be removed from handleErr().
Same below.

tools/cassandra/cqlclient.go
```diff
@@ -266,17 +266,26 @@ func (client *cqlClient) dropAllTablesTypes() error {
...
-			log.Printf("Error dropping type %v, err=%v\n", t, err1)
-			return err1
+	numOfTypes := len(types)
+	for i := 0; i < numOfTypes && len(types) > 0; i++ {
+		var tmp []string
+		for _, t := range types {
+			err = client.dropType(t)
+			if err != nil {
+				log.Printf("Error dropping type %v, err=%v\n", t, err)
+				tmp = append(tmp, t)
```
yycptt: Change `tmp` to a more meaning name?

service/history/timerQueueProcessorBase.go
vancexu: sure. if so I will keep it

tools/sql/handler.go
vancexu: That a good observation but `handleErr` here is:
```
func handleErr(err error) error {
	log.Println(err)
	return err
}
```

That's why I write that way in PR. Will keep it like that

tools/cassandra/cqlclient.go
vancexu: renamed

cmd/tools/cassandra/main.go
```diff
@@ -27,5 +27,5 @@ import (
 )
 
 func main() {
-	cassandra.RunTool(os.Args)
+	cassandra.RunTool(os.Args) //nolint:errcheck
```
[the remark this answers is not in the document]
vancexu: I think explicit add nolint comment might be better because:
1. it will avoid people guessing why you do `_ = cassandra.RunTool()`
2. it might save tiny assignment operation

Will keep it that way

common/persistence/domainReplicationQueue.go
```diff
@@ -125,58 +122,6 @@ func (q *domainReplicationQueueImpl) GetAckLevels() (map[string]int, error) {
...
-				if err != nil {
-					q.logger.Warn("Failed to purge acked domain replication messages.", tag.Error(err))
-				}
-			}
-		case <-q.ackNotificationChan:
-			q.ackLevelUpdated = true
-		}
-	}
-}
-
```
wxing1292: why deleting this 2 functions?
@yux0 are these 2 functions still useful or not?

yux0: Thanks for raising the concern. I will add those functions back

PR #2963, opened 2020-01-07
Wire up domain replication queue cleanup coroutine

yux0:
The domain replication queue has a function to cleanup the applied messages which is not being invoked. This change is to invoke the function from the domain replication queue. 
The queue maintains a map of the ack levels of each cluster. The domain replication queue will clean up all the messages before the minimum ack level among all clusters.


common/persistence/domainReplicationQueue.go
```diff
@@ -110,18 +116,63 @@ func (q *domainReplicationQueueImpl) UpdateAckLevel(lastProcessedMessageID int,
...
+		}
+	}
+
+	err = q.queue.DeleteMessagesBefore(minAckLevel)
+	if err != nil {
+		return fmt.Errorf("failed to purge messages: %v", err)
+	}
+
+	q.metricsClient.
+		Scope(metrics.FrontendDomainReplicationQueueScope).
```
venkat1109: Using frontendScope within the common/persistence layer seems weird. Is this a copy/paste error ?

common/persistence/domainReplicationQueue.go
```diff
@@ -45,25 +48,24 @@ func NewDomainReplicationQueue(
...
-		logger:              logger,
-		encoder:             codec.NewThriftRWEncoder(),
-		ackNotificationChan: make(chan bool),
-		done:                make(chan bool),
+		queue:           queue,
+		clusterName:     clusterName,
+		metricsClient:   metricsClient,
+		logger:          logger,
+		encoder:         codec.NewThriftRWEncoder(),
+		lastCleanupTime: time.Now().UTC(),
```
venkat1109: The requirement to call UTC() everywhere this field is accessed seems error prone and a bug waiting to happen. How about we have the API DeleteMessagesBefore(t time.Time) call t.UTC() so that the callers don't have to worry about this ?

common/persistence/domainReplicationQueue.go
```diff
@@ -110,18 +116,63 @@ func (q *domainReplicationQueueImpl) UpdateAckLevel(lastProcessedMessageID int,
 		return fmt.Errorf("failed to update ack level: %v", err)
 	}
 
-	select {
-	case q.ackNotificationChan <- true:
```
venkat1109: is this channel not consumed anywhere before ? I wonder what the original intent was

common/persistence/domainReplicationQueue.go
```diff
@@ -84,6 +86,10 @@ func (q *domainReplicationQueueImpl) GetReplicationMessages(
 	lastMessageID int,
 	maxCount int,
 ) ([]*replicator.ReplicationTask, int, error) {
+
+	// trigger a domain queue cleanup
+	go q.purgeProcessor()
```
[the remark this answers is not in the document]
yux0: The only reason I change to this approach is that I didn't find a good place to start this coroutine.

common/persistence/domainReplicationQueue.go
```diff
@@ -23,16 +23,19 @@ package persistence
...
 	"github.com/uber/cadence/.gen/go/replicator"
 	"github.com/uber/cadence/common/codec"
 	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/metrics"
 )
 
 const (
-	purgeInterval = time.Minute
+	maxMessagePurgeDelta = 30 * time.Minute
```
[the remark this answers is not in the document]
yux0: Domain replication won't happen often. A longer period should be fine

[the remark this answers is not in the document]
yux0: Revert the change to keep the original approach

common/persistence/domainReplicationQueue.go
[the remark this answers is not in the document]
yux0: Synced offline. Going to leverage sync.once

common/persistence/domainReplicationQueue.go
yux0: Human error. Update the name of the scope

common/persistence/domainReplicationQueue.go
yux0: With the origin approach, I don't need this time source

common/persistence/domainReplicationQueue.go
yux0: It used in the original design. I am adding it back

common/persistence/domainReplicationQueue.go
```diff
@@ -122,6 +131,57 @@ func (q *domainReplicationQueueImpl) GetAckLevels() (map[string]int, error) {
...
+	err = q.queue.DeleteMessagesBefore(minAckLevel)
+	if err != nil {
+		return fmt.Errorf("failed to purge messages: %v", err)
+	}
+
+	q.metricsClient.
+		Scope(metrics.HistoryDomainReplicationQueueScope).
+		UpdateGauge(metrics.DomainReplicationTaskAckLevel, float64(minAckLevel))
+
+	q.ackLevelUpdated = false
```
wxing1292: `q.ackLevelUpdated = false` should be moved to 
```go
if q.ackLevelUpdated {
	err := q.purgeAckedMessages()
	if err != nil {
		q.logger.Warn("Failed to purge acked domain replication messages.", tag.Error(err))
	} else {
		q.ackLevelUpdated = false
	}
}
```

common/persistence/domainReplicationQueue.go
yux0: The only trigger from frontend. Keep the frontendScope for now

PR #2964, opened 2020-01-08
Add Domain Replication DLQ

yux0:
1. Add domain DLQ 
2. Add enqueue dlq logic in worker
3. Update some DLQ persistence API (not fully integrated, require next PR with CLI changes)


common/persistence/cassandra/cassandraQueue.go
```diff
@@ -36,17 +36,19 @@ import (
...
-	templateDeleteMessagesQuery      = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
-	templateGetQueueMetadataQuery    = `SELECT cluster_ack_level, version FROM queue_metadata WHERE queue_type = ?`
-	templateInsertQueueMetadataQuery = `INSERT INTO queue_metadata (queue_type, cluster_ack_level, version) VALUES(?, ?, ?) IF NOT EXISTS`
-	templateUpdateQueueMetadataQuery = `UPDATE queue_metadata SET cluster_ack_level = ?, version = ? WHERE queue_type = ? IF version = ?`
+	templateEnqueueMessageQuery        = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
+	templateGetLastMessageIDQuery      = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
+	templateGetMessagesQuery           = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
+	templateGetMessagesFromDLQQuery    = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id <= ?`
+	templateDeleteMessagesQuery        = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
+	templateDeleteMessagesFromDLQQuery = `DELETE FROM queue WHERE queue_type = ? and message_id = ?`
```
venkat1109: rename to templateDeleteMessageFromDLQQuery - this query deletes just one message.

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -115,29 +117,43 @@ func (q *cassandraQueue) createQueueMetadataEntryIfNotExist() error {
 func (q *cassandraQueue) EnqueueMessage(
 	messagePayload []byte,
 ) error {
-	nextMessageID, err := q.getNextMessageID()
+	nextMessageID, err := q.getLastMessageID(q.queueType)
```
venkat1109: rename nextMessageID to lastMessageID now that you have changed the method name

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -115,29 +117,43 @@ func (q *cassandraQueue) createQueueMetadataEntryIfNotExist() error {
...
 
-	return q.tryEnqueue(nextMessageID, messagePayload)
+	return q.tryEnqueue(q.queueType, nextMessageID+1, messagePayload)
+}
+
+func (q *cassandraQueue) EnqueueMessageToDLQ(
+	messagePayload []byte,
+) error {
+	// Use negative queue type as the dlq type
+	nextMessageID, err := q.getLastMessageID(-q.queueType)
```
venkat1109: same renaming comment

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -225,6 +287,36 @@ func (q *cassandraQueue) DeleteMessagesBefore(messageID int) error {
...
+
+	return nil
+}
+
+func (q *cassandraQueue) RangeDeleteMessagesFromDLQ(
+	messageID int,
+) error {
+
+	// Use negative queue type as the dlq type
+	query := q.session.Query(templateDeleteMessagesQuery, -q.queueType, messageID)
```
venkat1109: this doesn't really do a range delete. Do you even need this method ?

common/persistence/sql/sqlplugin/mysql/queue.go
```diff
@@ -32,7 +32,9 @@ const (
 	templateEnqueueMessageQuery            = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(:queue_type, :message_id, :message_payload)`
 	templateGetLastMessageIDQuery          = `SELECT message_id FROM queue WHERE message_id >= (SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1) FOR UPDATE`
 	templateGetMessagesQuery               = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
-	templateDeleteMessagesQuery            = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
+	templateGetMessagesBeforeQuery         = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id <= ?`
```
venkat1109: do you need the messages to be ordered by message_id ? If yes, you need to add ORDER BY

common/persistence/sql/sqlplugin/postgres/queue.go
```diff
@@ -32,7 +32,9 @@ const (
 	templateEnqueueMessageQuery            = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(:queue_type, :message_id, :message_payload)`
 	templateGetLastMessageIDQuery          = `SELECT message_id FROM queue WHERE message_id >= (SELECT message_id FROM queue WHERE queue_type=$1 ORDER BY message_id DESC LIMIT 1) FOR UPDATE`
 	templateGetMessagesQuery               = `SELECT message_id, message_payload FROM queue WHERE queue_type = $1 and message_id > $2 LIMIT $3`
-	templateDeleteMessagesQuery            = `DELETE FROM queue WHERE queue_type = $1 and message_id < $2`
+	templateGetMessagesBeforeQuery         = `SELECT message_id, message_payload FROM queue WHERE queue_type = $1 and message_id <= $2`
```
venkat1109: Add ORDER BY if necessary

common/persistence/cassandra/cassandraQueue.go
yux0: Updated the name

common/persistence/cassandra/cassandraQueue.go
yux0: renamed

common/persistence/cassandra/cassandraQueue.go
yux0: renamed

common/persistence/cassandra/cassandraQueue.go
yux0: renamed to DeleteMessagesFromDLQ. No place is calling this method. The plan is to provide a CLI as purge all messages before <message ID> and it will use this method

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -36,17 +36,19 @@ import (
...
-	templateGetLastMessageIDQuery    = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
-	templateGetMessagesQuery         = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
-	templateDeleteMessagesQuery      = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
-	templateGetQueueMetadataQuery    = `SELECT cluster_ack_level, version FROM queue_metadata WHERE queue_type = ?`
-	templateInsertQueueMetadataQuery = `INSERT INTO queue_metadata (queue_type, cluster_ack_level, version) VALUES(?, ?, ?) IF NOT EXISTS`
-	templateUpdateQueueMetadataQuery = `UPDATE queue_metadata SET cluster_ack_level = ?, version = ? WHERE queue_type = ? IF version = ?`
+	templateEnqueueMessageQuery        = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
+	templateGetLastMessageIDQuery      = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
+	templateGetMessagesQuery           = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
```
venkat1109: [ not a blocker for this PR ] AFAIK, this is an anti-pattern in cassandra wrt performance. You always end to specify both end of the range i.e. you want something like message_id > ? and message_id < ?

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -36,17 +36,19 @@ import (
...
 const (
-	templateEnqueueMessageQuery      = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
-	templateGetLastMessageIDQuery    = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
-	templateGetMessagesQuery         = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
-	templateDeleteMessagesQuery      = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
-	templateGetQueueMetadataQuery    = `SELECT cluster_ack_level, version FROM queue_metadata WHERE queue_type = ?`
-	templateInsertQueueMetadataQuery = `INSERT INTO queue_metadata (queue_type, cluster_ack_level, version) VALUES(?, ?, ?) IF NOT EXISTS`
-	templateUpdateQueueMetadataQuery = `UPDATE queue_metadata SET cluster_ack_level = ?, version = ? WHERE queue_type = ? IF version = ?`
+	templateEnqueueMessageQuery        = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
+	templateGetLastMessageIDQuery      = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
```
venkat1109: see comment below, if I remember right, this can cause performance problems when you have lots of messages on the queue

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -206,15 +224,59 @@ func (q *cassandraQueue) ReadMessages(
...
+	lastMessageID int,
+) ([]*persistence.QueueMessage, error) {
+	// Reading replication tasks need to be quorum level consistent, otherwise we could loose task
+	// Use negative queue type as the dlq type
+	query := q.session.Query(templateGetMessagesFromDLQQuery,
+		-q.queueType,
+		lastMessageID,
+	)
+
+	iter := query.Iter()
```
venkat1109: what is the page size here ? Are you simply going to return all the messages from DLQ in one shot ?

common/persistence/cassandra/cassandraQueue.go
```diff
@@ -36,17 +36,19 @@ import (
...
-	templateGetMessagesQuery         = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
-	templateDeleteMessagesQuery      = `DELETE FROM queue WHERE queue_type = ? and message_id < ?`
-	templateGetQueueMetadataQuery    = `SELECT cluster_ack_level, version FROM queue_metadata WHERE queue_type = ?`
-	templateInsertQueueMetadataQuery = `INSERT INTO queue_metadata (queue_type, cluster_ack_level, version) VALUES(?, ?, ?) IF NOT EXISTS`
-	templateUpdateQueueMetadataQuery = `UPDATE queue_metadata SET cluster_ack_level = ?, version = ? WHERE queue_type = ? IF version = ?`
+	templateEnqueueMessageQuery        = `INSERT INTO queue (queue_type, message_id, message_payload) VALUES(?, ?, ?) IF NOT EXISTS`
+	templateGetLastMessageIDQuery      = `SELECT message_id FROM queue WHERE queue_type=? ORDER BY message_id DESC LIMIT 1`
+	templateGetMessagesQuery           = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id > ? LIMIT ?`
+	templateGetMessagesFromDLQQuery    = `SELECT message_id, message_payload FROM queue WHERE queue_type = ? and message_id <= ?`
```
venkat1109: this query will return all the messages in the DLQ. Don't you need batching and/or pagination ? In addition, I believe you will need ORDER BY here

common/persistence/sql/sqlplugin/mysql/queue.go
yux0: Yes. The ordering is needed. I checked Cassandra do ascending order by default. Do you think I need to add this to cassandra query?

common/persistence/sql/sqlplugin/postgres/queue.go
yux0: Added

common/persistence/sql/sqlplugin/mysql/queue.go
venkat1109: This will return all the messages, don't you need batching and/or LIMIT ?

yux0: One assumption is the DLQ won't have a large amount of messages. As this is not a valid assumption for a queue. I will add both boundaries in the query and limit the size

common/persistence/cassandra/cassandraQueue.go
yux0: Added the range of get query and the result limit count. This should not be an issue as this queue only used for domain replication. We can revisit the design of the queue when there is need for performance.

common/persistence/cassandra/cassandraQueue.go
yux0: Added the lower boundary

common/persistence/cassandra/cassandraQueue.go
yux0: Add the max count

common/persistence/cassandra/cassandraQueue.go
yux0: Added order by and limit the return number

service/worker/replicator/domainReplicationMessageProcessor.go
```diff
@@ -146,22 +150,40 @@ func (p *domainReplicationMessageProcessor) getAndHandleDomainReplicationTasks()
...
 		if err != nil {
 			p.metricsClient.IncCounter(metrics.DomainReplicationTaskScope, metrics.ReplicatorFailures)
-			// TODO: put task into DLQ
+			p.logger.Error("Failed to apply domain replication tasks", tag.Error(err))
+
+			dlqErr := backoff.Retry(func() error {
+				return p.putDomainReplicationTaskToDLQ(task)
+			}, p.retryPolicy, isTransientRetryableError)
+			if dlqErr != nil {
+				p.logger.Error("Failed to put replication tasks to DLQ", tag.Error(dlqErr))
```
venkat1109: can you also emit a metric here ?

PR #2967, opened 2020-01-09
mutable state checksums part#2: persistence changes

venkat1109:
This patch contains the cassandra persistence changes to support mutable state checksums. This is a follow up to #2941

### DIFF
--- a/common/persistence/cassandra/cassandraPersistenceUtil.go
+++ b/common/persistence/cassandra/cassandraPersistenceUtil.go
@@ -28,6 +28,7 @@ import (
 
 	workflow "github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/checksum"
 	p "github.com/uber/cadence/common/persistence"
 )
 
@@ -55,6 +56,7 @@ func applyWorkflowMutationBatch(
 		versionHistories,
 		cqlNowTimestampMillis,
 		condition,
+		workflowMutation.Checksum,
 	); err != nil {
 		return err
 	}
@@ -170,6 +172,7 @@ func applyWorkflowSnapshotBatchAsReset(
 		versionHistories,
 		cqlNowTimestampMillis,
 		condition,
+		workflowSnapshot.Checksum,
 	); err != nil {
 		return err
 	}
@@ -274,6 +277,7 @@ func applyWorkflowSnapshotBatchAsNew(
 		executionInfo,
 		replicationState,
 		versionHistories,
+		workflowSnapshot.Checksum,
 		cqlNowTimestampMillis,
 	); err != nil {
 		return err
@@ -362,6 +366,7 @@ func createExecution(
 	executionInfo *p.InternalWorkflowExecutionInfo,
 	replicationState *p.ReplicationState,
 	versionHistories *p.DataBlob,
+	checksum checksum.Checksum,
 	cqlNowTimestampMillis int64,
 ) error {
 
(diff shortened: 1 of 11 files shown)


schema/cassandra/cadence/versioned/v0.24/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "CurrVersion": "0.24",
+  "MinCompatibleVersion": "0.23",
```
wxing1292: 0.24?

common/persistence/persistenceInterface.go
```diff
@@ -24,6 +24,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/uber/cadence/common/checksum"
```
wxing1292: make fmt?

common/persistence/cassandra/cassandraPersistenceUtil.go
```diff
@@ -2500,6 +2524,24 @@ func createReplicationInfoMap(
 	return rInfoMap
 }
 
+func createChecksum(result map[string]interface{}) checksum.Checksum {
+	if len(result) == 0 {
+		return checksum.Checksum{}
+	}
+	csum := checksum.Checksum{}
```
wxing1292:
```go
csum := checksum.Checksum{}
if len(result) == 0 {
	return csum
}
```

common/persistence/persistenceInterface.go
```diff
@@ -244,6 +246,7 @@ type (
 		SignalInfos         map[int64]*SignalInfo
 		SignalRequestedIDs  map[string]struct{}
 		BufferedEvents      []*DataBlob
+		Checksum            checksum.Checksum
 	}
```
wxing1292: i would put checksum in a seperate group, e.g.

```go
InternalWorkflowMutableState struct {
		ExecutionInfo    *InternalWorkflowExecutionInfo
		ReplicationState *ReplicationState
		VersionHistories *DataBlob

		ActivityInfos       map[int64]*InternalActivityInfo
		TimerInfos          map[string]*TimerInfo
		ChildExecutionInfos map[int64]*InternalChildExecutionInfo
		RequestCancelInfos  map[int64]*RequestCancelInfo
		SignalInfos         map[int64]*SignalInfo
		SignalRequestedIDs  map[string]struct{}
		BufferedEvents      []*DataBlob

		Checksum            checksum.Checksum
}
```

common/persistence/persistenceInterface.go
```diff
@@ -380,6 +383,7 @@ type (
 		ReplicationTasks []Task
 
 		Condition int64
+		Checksum  checksum.Checksum
```
wxing1292: same as above

common/persistence/persistenceInterface.go
```diff
@@ -402,6 +406,7 @@ type (
...
```
wxing1292: same as above

wxing1292:
one more thing, where is the impl for SQL?

schema/cassandra/cadence/versioned/v0.24/manifest.json
venkat1109: This change is backwards compatible with 0.23 and so this is set to 0.23

common/persistence/persistenceInterface.go
venkat1109: I assume this is simply personal preference. In general, this is something that is not auto enforcable by gofmt etc. And so, over time, your personal preference for grouping this separately is not sustainable. I will group this separately for now. But philosophically, I am not a big fan these type of semi-enforced groupings (unless there is strong reasons like struct packing in C/C++)

venkat1109:
> one more thing, where is the impl for SQL?
SQL is currently not supported. I will consider adding support for it after rolling this out. There will be a github task created though for anyone to pick that up.

PR #2972, opened 2020-01-09
Add persistence ListExecutions API

andrewjdawson2016:


common/persistence/dataInterfaces.go
```diff
@@ -1000,6 +1000,18 @@ type (
...
+	// ListExecutionsRequest is request to ListExecutions
+	ListExecutionsRequest struct {
+		NextPageToken []byte
+		PageSize      int
+	}
+
+	// ListExecutionsResponse is response from ListExecutions
+	ListExecutionsResponse struct {
+		NextPageToken                []byte
+		GetWorkflowExecutionRequests []GetWorkflowExecutionRequest
```
venkat1109: this seems a bit weird. Any reason to reuse a request struct in a response struct ?

common/persistence/dataInterfaces.go
```diff
@@ -1477,6 +1489,8 @@ type (
 
 		// Remove Task due to corrupted data
 		DeleteTask(request *DeleteTaskRequest) error
+
+		ListExecutions(request *ListExecutionsRequest) (*ListExecutionsResponse, error)
```
venkat1109: how do you plan to use this API in the mutable state scanner ? Will the scanner first call this api to get a list of (wfID,runID) and then make another call to GetWorkflowExecution() for each one ? In Cassandra, executions table is weird in that it is multiple tables in one. Some entries will have "special" runIDs which have no meaning.

andrewjdawson2016: 1. How will this be used: yes two calls was my idea. Its possible for this API to return more information but not all persistence layers can do this with a single table scan, so I am returning Ids so that all persistence stores can be used the same way. 

2. When I run the scan I filter based on row type ```query := d.session.Query(templateListExecutionsQuery, d.shardID, rowTypeExecution)```. I understood this to me that I will only . get meaningful runIds. Do you know if this is not the case?

common/persistence/dataInterfaces.go
andrewjdawson2016: I agree it seems a little weird. But the response I want is exactly this request: domainID and an execution. I could create a type alias to this struct or create a new struct with the same fields?

andrewjdawson2016: Maybe one reason that its not weird is that after calling this List API the next call will be to get the workflow execution and you will then already have the requests (no need to convert between types).

common/persistence/dataInterfaces.go
venkat1109: ah.. i missed the rowTypeExecution part.

common/persistence/dataInterfaces.go
venkat1109: From an API contract perspective, it looks weird. In future, there could be other use cases where the next step is not GetWorkflowExecution(). I am ok with alias or new struct.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -2317,6 +2321,45 @@ func (d *cassandraPersistence) RangeCompleteTimerTask(request *p.RangeCompleteTi
 	return nil
 }
 
+func (d *cassandraPersistence) ListExecutions(request *p.ListExecutionsRequest) (*p.ListExecutionsResponse, error) {
+	query := d.session.Query(templateListExecutionsQuery, d.shardID, rowTypeExecution)
+
+	iter := query.PageSize(int(request.PageSize)).PageState(request.NextPageToken).Iter()
```
meiliang86: Might want to define a max page size here just for protection.

common/persistence/dataInterfaces.go
meiliang86: +1 on using a different name

common/persistence/dataInterfaces.go
venkat1109: Let's hold onto this PR until andrew is back from vacation. I want to have a brief sync around the API i.e. returning (wfID, runID) as opposed to more info related to mutable state

Issue #2681, opened 2019-10-17
Cadence does not support TLS for Cassandra/MySQL


vsabella:
Currently we run our MySQL and Cassandra clusters using TLS. Looking for TLS support in Cadence and in the toolchains (like cadence-cassandra-tool)

https://godoc.org/github.com/gocql/gocql#SslOptions

vsabella:
Additionally, it looks like there are multiple implementations of Cassandra wrappers - some for CLI tools and some for the persistence store. Is there any desire to unify these?

PankajWorks:
@mfateev since tls for cassandra is already done. wanted to check what is the plan around this.

mfateev:
@PankajWorks We don't have concrete plans and resources to work on it in the next couple of months. Any contributions in this area are welcome.

PankajWorks:
We are working on this issue for mysql. **We are proposing the approach here so that we can get some feedback. One key requirement we are trying to solve is the ability to specify TLS validation as well - which necessitates the use of the [mysql.RegisterTLSConfig](https://godoc.org/github.com/go-sql-driver/mysql#RegisterTLSConfig) and required parameters. The changes include:**

- Introduce a "enableTLS" flag to enable/disable custom TLS Configuration. It will set  tls="custom" as DSN attribute

- If enableTls flag is set to true , the user will have to provide different certificates as required by [mysql.RegisterTLSConfig](https://godoc.org/github.com/go-sql-driver/mysql#RegisterTLSConfig) to make a secure connection i.e.
    - cacert - The path name of the Certificate Authority (CA) certificate file.
    - tlscert - The path name of the client public key certificate file.
    - tlskey - The path name of the client private key file.
-  If user decides to add tls as part of connectAttributes and enableTls is not specified, then that value will be used while constructing DSN. 
if tls is used as part of connectAttributes along with enableTLS=true then tls value present in connectAttributes will get overridden by custom TLS configuration.
**We would like to work on this for MySQL first. Postgres TLS support can be added once MySQL support is in**.

bjacobso:
@PankajWorks @SachinHg 

I have a PR to add support TLS here:
https://github.com/uber/cadence/pull/2913

Does this accomplish what you are trying to achieve?

PR #2974, opened 2020-01-10
Add TLS support for MySQL plugin.

SachinHg:
This PR follows from #2681.

1. Adds TLS support for MySQL plugin
2. Introduce new flags for cadence-sql-tool
    - `tls` 
    - `tls-config`
    - `tls-ca-file`
    - `tls-server`
    - `tls-cert-file`
    - `tls-key-file`
3. Add TLS config values for SQL into new common `auth.SQLTLS`


bjacobso:
@SachinHg I believe this will do what you are trying to do:
https://github.com/uber/cadence/pull/2913

SachinHg:
@bjacobso thanks for the commit. I will close this PR but would like for your comments regarding https://github.com/uber/cadence/pull/2913#issuecomment-573287077 . 
Thanks

PR #2975, opened 2020-01-10
Task Processing Job Implementation

yycptt:
**Will prioritize work on task processor side instead of the queue management. So I will address the comments later. Thanks.**

This pr implements the task processing job, the smallest building block for the multi-tenant task processing logic. On top of the processing job, another component will be implemented to manage a list of jobs with disjoint range (can be viewed as a queue) and responsible for managing their life cycles, read tasks, etc.

Logging and metrics will be added later.


service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination taskProcessingJob_mock.go -self_package github.com/uber/cadence/service/history
+
+package history
```
meiliang86: how easy it is to put task processing logic in its own package?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	incomingJobImpl, ok := incomingJob.(*taskProcessingJobImpl)
+	if !ok {
+		panic("unknown task processsing job type")
+	}
+
+	if t.jobLevel != incomingJobImpl.jobLevel {
+		panic("trying to merge job with different job level")
+	}
+
+	if !compareTaskKeyLess(&incomingJobImpl.ackLevel, &t.maxReadLevel) {
```
yux0: Does ack level always <= max read level? Will this always return nil?

service/history/taskProcessingJob.go
wxing1292: this file should be in its own package

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/task"
+)
+
+type (
+	taskProcessingJob interface {
+		Level() int
+		AckLevel() taskKey
+		ReadLevel() taskKey
+		MaxReadLevel() taskKey
```
wxing1292: how about
```go
MinLevel() // level where the job begins
MaxLevel() // level where the job ends
AckLevel() // level where the tasks before are acked
LoadLevel() // level where the tasks before are loaded
QueueID() int // the (previous) level
```

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+
+	taskKey struct {
+		taskID              int64
+		visibilityTimestamp time.Time
+	}
+
+	domainFilter struct {
+		domainIDs    map[string]struct{}
+		invertResult bool
+	}
```
wxing1292: i would do 2 policies
1. everything except some domains
2. only domain specified

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+
+		config *taskProcessingJobConfig
+
+		logger        log.Logger
+		metricsClient metrics.Client
+	}
+)
+
+func newTaskProcessingJob(
+	jobLevel int,
```
wxing1292: should name this queue id

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		if _, ok := pendingTaskStats[taskDomainID]; !ok {
+			pendingTaskStats[taskDomainID] = 1
+		} else {
+			pendingTaskStats[taskDomainID]++
+		}
+	}
+
+	// step 2: evaluate split policy
+	domainIDToNewJobLevel := splitPolicy.Evaluate(pendingTaskStats, t.Level())
+	if domainIDToNewJobLevel == nil || len(domainIDToNewJobLevel) == 0 {
```
wxing1292: `len(domainIDToNewJobLevel) == 0` is enough

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		// not need to split
+		return nil
```
wxing1292: return self to avoid special handling?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:generate mockgen -copyright_file ../../LICENSE -package $GOPACKAGE -source $GOFILE -destination taskProcessingJob_mock.go -self_package github.com/uber/cadence/service/history
+
+package history
+
+import (
```
wxing1292: plz write some description about the term used here
my preference
1. use min / max query level indicating the min / max bound for a job
2. use ack / load level indicate what tasks are acked / loaded in memory

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		}
+
+		t.outstandingTasks[taskKey] = task
+		if compareTaskKeyLess(&t.readLevel, &taskKey) {
+			t.readLevel = taskKey
+		}
+	}
+}
+
+func (t *taskProcessingJobImpl) Split(splitPolicy taskProcessingJobSplitPolicy) []taskProcessingJob {
```
wxing1292: break this function down to several smaller / clearer functions

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+				newJobsOutstandingTasks[newLevel] = make(map[taskKey]queueTask)
+			}
+			newJobsOutstandingTasks[newLevel][key] = task
+		} else {
+			remainingTasks[key] = task
+		}
+	}
+
+	var newJobs []taskProcessingJob
+	newMaxReadLevel := taskKey{
```
wxing1292: newMaxQueryLevel or newMaxLevel

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	}
+
+	return domainIDToNewLevel
+}
+
+func isStandbyDomain(
+	domainCache cache.DomainCache,
+	domainID string,
+	currentClusterName string,
+) bool {
```
wxing1292: return error and bool

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	currentLevel int,
+) map[string]int {
+	domainIDToNewLevel := make(map[string]int)
+
+	for _, policy := range p.policies {
+		result := policy.Evaluate(pendingTaskStats, currentLevel)
+		for domainID, level := range result {
+			if _, ok := domainIDToNewLevel[domainID]; !ok {
+				domainIDToNewLevel[domainID] = level
+			}
```
wxing1292: what if ok? first come first serve?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		standbyJobLevel:    standbyJobLevel,
+		currentClusterName: currentClusterName,
+	}
+}
+
+func (p *standbyDomainSplitPolicy) Evaluate(
+	pendingTaskStats map[string]int,
+	currentLevel int,
+) map[string]int {
+	if currentLevel == p.standbyJobLevel {
```
wxing1292: what is a `standbyJobLevel`?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		}
+	}
+
+	if totalPendingTaskNum < p.totalPendingTaskThreshold {
+		return nil
+	}
+	return domainIDToNewLevel
+}
+
+func newStandbyDomainSplitPolicy(
```
wxing1292: i would put the policies in a separate file

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		totalPendingTaskThreshold:     totalPendingTaskThreshold,
+		perDomainPendingTaskThreshold: perDomainPendingTaskThreshold,
+	}
+}
+
+func (p *baseSplitPolicy) Evaluate(
+	pendingTaskStats map[string]int,
+	currentLevel int,
+) map[string]int {
+	if currentLevel >= p.maxJobLevel {
```
wxing1292: what is a `maxJobLevel`? why a policy has a job level?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+		totalPendingTaskNum += numPendingTasks
+		if numPendingTasks > p.perDomainPendingTaskThreshold {
+			domainIDToNewLevel[domainID] = currentLevel + 1
+		}
+	}
+
+	if totalPendingTaskNum < p.totalPendingTaskThreshold {
+		return nil
+	}
+	return domainIDToNewLevel
```
wxing1292: so if `if totalPendingTaskNum < p.totalPendingTaskThreshold` then even if # of domain tasks exceeds `perDomainPendingTaskThreshold` nothing is done? then why having `totalPendingTaskThreshold`?

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+			continue
+		}
+
+		if compareTaskKeyLess(&taskKey, &t.ackLevel) {
+			continue
+		}
+
+		if compareTaskKeyLess(&t.maxReadLevel, &taskKey) {
+			continue
+		}
```
wxing1292: for impossible case, just panic

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	return newJobs
+}
+
+func (t *taskProcessingJobImpl) Merge(incomingJob taskProcessingJob) []taskProcessingJob {
+	incomingJobImpl, ok := incomingJob.(*taskProcessingJobImpl)
+	if !ok {
+		panic("unknown task processsing job type")
+	}
+
+	if t.jobLevel != incomingJobImpl.jobLevel {
```
wxing1292: my understanding, job itself should be level / queue ID (i would prefer queue ID, since there are so many levels) independent.
each task processing queue controller, should have the queue ID.
each task processing queue, can have 0-many jobs.

service/history/taskProcessingJob.go
```diff
@@ -0,0 +1,646 @@
...
+	}
+
+	generationFunc := []func(*taskProcessingJobImpl, *taskProcessingJobImpl) *taskProcessingJobImpl{
+		// if exists, generate the job for the first non-overlapping part (whose range is before the overlapping part)
+		generateMergedPrefixJob,
+		// generate the job for the overlapping part
+		generateMergedJob,
+		// if exists, generate the job for the second non-overlapping part (whose ranger is after the overlapping part)
+		generateMergedSuffixJob,
+	}
```
wxing1292: if the content of this anonymous function does not change, make it a real function?

Issue #825, opened 2018-06-07
CLI: Improve listworkflow with friendly date range 


vancexu:
Currently CLI `workflow list` takes -et -lt as UTC and UnixNano, which is not human friendly. 
Need to improve it to list last 1 hour, last week etc.

We can also think about replace existing datetime parser to something like https://github.com/araddon/dateparse, which relax on datetime format.

kaustubhmallik:
@wxing1292 Is it still open to grab? If so, I would be more than happy to help.

vancexu:
@kaustubhmallik Yes it is. It would be great if you can take it.

kaustubhmallik:
@vancexu I would love to. Have you given any thought about it already? Or should I come up with my approach, and maybe you can guide me further.

kaustubhmallik:
Hi @vancexu, I have thought of doing it in the following manner.

The argument would look like - ```last <num><duration>```

Users are supposed to put quotes (single/double) around the argument just like any other *nix argument to be treated as one argument.

- num  - positive integral value (1e6 > num > 0 ). ZERO and negatives are considered invalid value. If num is skipped, a default value of ONE is considered.

- duration - possible values for duration will be (full name/short name):
        - second/s
        - minute/m
        - hour/h
        - day/d
        - week/w
        - month/M
        - year/y

The maximum duration can go as back as Unix epoch time (1 Jan 1970 00:00:00 UTC), as we are already using Unix timestamp, so it will be compatible with it.

For example, some possible arguments can be (each row have equivalent arguments):
* last 1second, last 1s
* last 10 minute, last 10m
* last hour, last h, last 1h
* last 4 week, last 4w
* last month, last M, last 1M
* last 1year, last 1y, last y

Let me know if this sounds good to you, or if there are some improvements/suggestions, I would love to hear it.

vancexu:
@kaustubhmallik  The proposal sounds good.

kaustubhmallik:
Now that would be awesome! Do you want to add that in the same proposal, or you want to do that separately? I guess it might be better if we can handle it separately, but please suggest what you feel like.

vancexu:
> Now that would be awesome! Do you want to add that in the same proposal, or you want to do that separately? I guess it might be better if we can handle it separately, but please suggest what you feel like.

Yes, agree it's better to do separately. Just share some idea for next.

kaustubhmallik:
In order to enter the time range, I can think of two ways:

[...]

PR #2990, opened 2020-01-19
Improve list workflow date range args (#825)

kaustubhmallik:
Allow date range like last "X duration" where X is a positive integer (0 < X < 1000000)
and duration is a time duration like second(s)/minute(m)/hour(h)/day(d)/month(M)/year(y)
For example, a few possible values are:
  - workflow list --et 1m: list workflows with earliest time in last 1 minute
  - workflow list --lt 3h: list workflows with latest time in last 3 hours
  - workflow list --et 2day: list workflows with earliest time in last 2 days

Also, note that oldest possible time is unix epoch time, 01 Jan 1970, 00:00:00 UTC


tools/cli/app_test.go
```diff
@@ -558,9 +558,125 @@ func (s *cliAppSuite) TestObserveWorkflowWithID() {
...
 
 func (s *cliAppSuite) TestParseTime() {
-	s.Equal(int64(100), parseTime("", 100))
-	s.Equal(int64(1528383845000000000), parseTime("2018-06-07T15:04:05+00:00", 0))
-	s.Equal(int64(1528383845000000000), parseTime("1528383845000000000", 0))
+	tests := []struct {
+		timeStr       string // input
+		defVal        int64  // input
+		lowerUnixNano int64  // lower range of expected output
+		upperUnixNano int64  // upper range of expected output
```
vancexu: How about removing `lowerUnixNano` and `upperUnixNano`,  but only had one `expected`, for all the new tests, you can assert `expected <= eval(text) <= expected + delta`. This will reduce some lines of code

kaustubhmallik: Yeah, even I thought of doing it, The only reason I did it, as the previously written tests had exact values, and I didn't want to change the passing criteria for those. So, I did it with a `lower` and `upper` bound, where the bounds coincide for previously written tests.

One other way I thought of was, I can have an `expected` value and a `delta` as part of the test cases, where I set `delta` value as `0` for previously written tests, and `5 ms` for new tests.

Or if we are not too strict with little change in passing criteria, we can just go ahead with what you suggested above. It surely improves the readability of the tests.

Let me know what you think is the best way. I will refactor the tests accordingly.

vancexu: I see you concerns. Let's have two tests, one for existing test cases using equal; one for new cases using `expected <= eval(text) <= expected + delta`

kaustubhmallik: Restored the original and added a second one for new changes, as you suggested 😃

kaustubhmallik:
@vancexu Thanks for accepting it 👍

PR #2991, opened 2020-01-20
Implementation of s3 using new archival interface

jontro:


common/archiver/s3store/historyArchiver.go
```diff
@@ -0,0 +1,333 @@
...
+		return nil, err
+	}
+
+	return &historyArchiver{
+		container:       container,
+		s3cli:           s3.New(sess),
+		historyIterator: historyIterator,
+	}, nil
+}
+func (h *historyArchiver) Archive(
```
yycptt: Please emit metrics in this method. 

MetricScope: `HistoryArchiverScope`
MetricNames:
```
        CadenceLatency
	HistoryArchiverArchiveNonRetryableErrorCount
	HistoryArchiverArchiveTransientErrorCount
	HistoryArchiverArchiveSuccessCount
	HistoryArchiverTotalUploadSize
	HistoryArchiverHistorySize
```

In your case upload size and history size will be the same since it's always a full upload.

common/archiver/s3store/historyArchiver.go
```diff
@@ -0,0 +1,333 @@
...
+		if err != nil {
+			return nil, &shared.InternalServiceError{Message: err.Error()}
+		}
+		response.NextPageToken = nextToken
+	}
+
+	return response, nil
+}
+
+func (h *historyArchiver) ValidateURI(URI archiver.URI) error {
```
yycptt: Is it possible to also add a test for if the bucket exists in S3 to the function? Same for the `ValidateURI` function in visibility archiver.

common/archiver/s3store/historyArchiver.go
```diff
@@ -0,0 +1,333 @@
...
+		if err != nil && !common.IsPersistenceTransientError(err) && !isRetryableError(err) && featureCatalog.NonRetriableError != nil {
+			err = featureCatalog.NonRetriableError()
+		}
+	}()
+
+	logger := archiver.TagLoggerWithArchiveHistoryRequestAndURI(h.container.Logger, request, URI.String())
+
+	if err := h.ValidateURI(URI); err != nil {
+		logger.Error(archiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason(archiver.ErrReasonInvalidURI), tag.Error(err))
+		return err
```
yycptt: Should be an non-retryable error

common/archiver/s3store/historyArchiver.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	if err := archiver.ValidateHistoryArchiveRequest(request); err != nil {
+		logger.Error(archiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason(archiver.ErrReasonInvalidArchiveRequest), tag.Error(err))
+		return err
```
yycptt: Should be an non-retryable error

common/archiver/s3store/queryParser.go
```diff
@@ -0,0 +1,241 @@
...
+	PrecisionMinute = "Minute"
+	PrecisionSecond = "Second"
+)
+const (
+	queryTemplate         = "select * from dummy where %s"
+	defaultDateTimeFormat = time.RFC3339
+)
+
+// NewQueryParser creates a new query parser for filestore
+func NewQueryParser() QueryParser {
```
yycptt: Can you add some explanation for the query syntax to the `README` file? E.g. Which filter is required? What's the limitation? Supported operators etc.

common/archiver/s3store/visibilityArchiver.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, err
+	}
+	return &visibilityArchiver{
+		container:   container,
+		s3cli:       s3.New(sess),
+		queryParser: NewQueryParser(),
+	}, nil
+}
+
+func (v *visibilityArchiver) Archive(
```
yycptt: Please emit metrics inside this method.

MetricScope: `VisibilityArchiverScope`
MetricsNames:
```
        CadenceLatency
	VisibilityArchiverArchiveNonRetryableErrorCount
	VisibilityArchiverArchiveTransientErrorCount
	VisibilityArchiveSuccessCount
```

common/archiver/s3store/visibilityArchiver.go
```diff
@@ -0,0 +1,252 @@
...
+		return err
+	}
+
+	encodedVisibilityRecord, err := encode(request)
+	if err != nil {
+		logger.Error(archiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason(errEncodeVisibilityRecord), tag.Error(err))
+		return err
+	}
+
+	// Upload archive to all indexes
```
yycptt: Probably worth explaining how data is stored and indexed in README.

common/archiver/provider/provider.go
```diff
@@ -172,6 +180,23 @@ func (p *archiverProvider) GetVisibilityArchiver(scheme, serviceName string) (ar
 			return nil, err
 		}
 
+		p.Lock()
+		defer p.Unlock()
+		if existingVisibilityArchiver, ok := p.visibilityArchivers[archiverKey]; ok {
+			return existingVisibilityArchiver, nil
+		}
+		p.visibilityArchivers[archiverKey] = visibilityArchiver
+		return visibilityArchiver, nil
```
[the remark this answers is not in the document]
jontro: Done!

common/archiver/s3store/historyArchiver.go
```diff
@@ -0,0 +1,333 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
jontro: Done

common/archiver/s3store/historyArchiver.go
jontro: Done!

common/archiver/s3store/historyArchiver.go
jontro: This would mean an extra round trip to s3 for each archive call. The error will be caught later on if misconfigured. What do you think?

common/archiver/s3store/historyArchiver.go
jontro: I wrapped it in a BadRequestError, not sure if this is the correct pattern?

common/archiver/s3store/historyArchiver.go
jontro: See previous comment

common/archiver/s3store/queryParser.go
jontro: I've started a section. English isn't my first language, will improve on it some more.

common/archiver/s3store/visibilityArchiver.go
jontro: Done!

common/archiver/s3store/visibilityArchiver.go
jontro: Done!

jontro:
I also added some missing ensureContextTimeout to Archive and Get in both stores

common/archiver/s3store/historyArchiver.go
yycptt: Currently, once a history/visibility URI is set for a domain it can not be updated later, because otherwise data can be archived to multiple locations and we don't be able to retrieve it upon request. That's why it's important when setting the URI, we need to get it right at the first time. 

Having the existence check help catch some typos when setting URI. I am ok one extra call as it should be cheap because it's not uploading/downloading any data. Thoughts?

common/archiver/s3store/queryParser.go
yycptt: 👍

common/archiver/s3store/historyArchiver.go
yycptt: We typically use BadRequestError when the input comes from the user. How about return some special error and check those errors in the `isRetryableError` function? 

I think there's a predefined error for invalid URI in the `archiver` package (constants.go). You can add another one for invalid archive request in that file.

jontro: The error returned by ValidateURI would be archiver.ErrURISchemeMismatch or errNoBucketSpecified. isRetryableError would return false for these errors.  So I am not relying on that fact. So how the code looked originally would not have been retried either afaik. Maybe we can have a chat about this to clarify how I should do this properly :)

common/archiver/s3store/historyArchiver.go
jontro: Agreed, didn't think about that case. However it might not be necessary to use ValidateURI in the Archive() and Get() methods

common/archiver/s3store/historyArchiver.go
yycptt: Sorry, My bad. Please ignore all my comments regarding this. I had some misunderstanding for the `isRetryableError` function. Your first version (return err directly) is correct. My apologies.

common/archiver/s3store/historyArchiver.go
yycptt: I agree. Maybe creating a util function to validate only the format? Archive() and Get() only call that util function while ValidateURI() will do an additional existence check?

PR #2994, opened 2020-01-22
Parallel Task Processor

yycptt:


common/task/priorityTaskProcessor.go
```diff
@@ -0,0 +1,155 @@
...
+func (p *priorityTaskProcessorImpl) Start() {
+	if !atomic.CompareAndSwapInt32(&p.status, common.DaemonStatusInitialized, common.DaemonStatusStarted) {
+		return
+	}
+
+	p.workerWG.Add(p.options.WorkerCount)
+	for i := 0; i < p.options.WorkerCount; i++ {
+		go p.taskWorker()
+	}
+	p.scheduler.Start()
```
yux0: The worker will call schedule.Consume(). Do you want to start the scheduler before the worker?

common/task/priorityTaskProcessor.go
```diff
@@ -0,0 +1,155 @@
...
+			return
+		}
+
+		if err := backoff.Retry(op, p.options.RetryPolicy, isRetryable); err != nil {
+			if !task.RetryErr(err) {
+				// encounter an non-retryable error
+				task.Nack()
+				return
+			}
+			// retryable error, continue the loop
```
yux0: is will do forever retry?

common/task/priorityTaskProcessor.go
```diff
@@ -0,0 +1,155 @@
...
+	op := func() error {
+		if err := task.Execute(); err != nil {
+			return task.HandleErr(err)
+		}
+		return nil
+	}
+
+	isRetryable := func(err error) bool {
+		if p.isStopped() {
+			return false
```
frozeNinK: i would argue the following:

1. a task should not have the ability to occupy a coroutine / thread and execute forever (i.e. retry for ever)
2. a task can have additional function `ScheduleRetry()` which tells the component which load the task to do a re-dispatching
3. there shall be no retry within this component

basically here, the only things to be done:
1. poll a task
2. execute (check shutdown)
3. 
* if no err -> ack
* if err -> check retry err 
** retry -> schedule retry
** no retry -> nack

common/task/priorityTaskProcessor.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// NewPriorityTaskProcessor creates a new PriorityTaskProcessor
+func NewPriorityTaskProcessor(
+	scheduler PriorityScheduler,
+	logger log.Logger,
+	metricsClient metrics.Client,
+	options *PriorityTaskProcessorOptions,
+) PriorityTaskProcessor {
+	return &priorityTaskProcessorImpl{
```
venkat1109: This relationship looks backwards from a mental model perspective i.e. scheduler should own a workerPool to schedule things and not the other way around. How about making workerPool (i.e. your taskProcessor) completely independent i.e. Something like the following:

```
Scheduler interface {
   Submit(Task task)
}

Task interface {
    ..
    Priority() int
}

TaskWorkerPool(poolSize, taskQ, metrics, logger)
 -- where taskQ is a small buffered input queue that feeds in tasks to be processed

```

What do you think ?

common/task/interface.go
```diff
@@ -18,22 +18,42 @@
...
+	// PriorityTaskProcessor is the generic coroutine pool interface
+	// which process tasks with priority
+	PriorityTaskProcessor interface {
+		common.Daemon
+		Submit(task Task, priority int) error
+	}
+
+	// PriorityScheduler is the generic interface for scheduling
+	// tasks with priority
+	PriorityScheduler interface {
```
venkat1109: - Can we call this TaskScheduler ? Scheduler name is too broad
 - From an interface design perspective, I would suggest to keep the Scheduler interface simple and move the priority to Task interface. Also Consume don't belong inside the Scheduler interface (see comments later). Something along the lines of :

```
TaskScheduler interface {
   Submit(Task task) error
}

Task interface {
    ...
   Priority() int
}
```
This way, priority scheduler simply becomes a type of Scheduler implementation which makes use of task_priority. There could be other scheduler implementations which ignore priority (like today).

common/task/priorityTaskProcessor.go
yycptt: Synced offline. Will implement priority task processor on top of parallel task processor (this pr) so that the logic for worker pool and processing tasks independent and can be reused.

Will send out another pr for the priority task processor (scheduler).

common/task/priorityTaskProcessor.go
yycptt: I think here we rely on the implementation of the task interface to determine if retry is needed or not. The retry is only for transient errors like service busy error. For other errors like domain not active or shard ownership lost, the task should give up retry so that it can be Nacked. Inside `Nack()`, the task implementation can determine if the task should be discarded or re-dispatched.

common/task/parallelTaskProcessor.go
```diff
@@ -0,0 +1,166 @@
...
+			return
+		}
+
+		if err := backoff.Retry(op, p.options.RetryPolicy, isRetryable); err != nil {
+			if !task.RetryErr(err) {
+				// encounter an non-retryable error
+				task.Nack()
+				return
+			}
+			// retryable error, continue the loop
```
yux0: If the error is a retryable error, it will still retry forever.

common/task/parallelTaskProcessor.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	for {
+		if p.isStopped() {
+			// neither ack or nack here
+			return
+		}
+
+		if err := backoff.Retry(op, p.options.RetryPolicy, isRetryable); err != nil {
+			if !task.RetryErr(err) {
```
yux0: Is this check needed? Can it nack and re-dispatch after retry?

yycptt: Both works. The answer depends on the behavior you want. If you only want to re-dispatch on task-defined nonretryable error then this check is needed. Otherwise, it can be removed. In my opinion, `RetryErr()` should only return `true` when the task should be retried immediately, and the returned value should be honored by the processor.

yycptt: Will update.

yycptt: Done.

common/task/parallelTaskProcessor.go
```diff
@@ -0,0 +1,160 @@
...
+			return task.HandleErr(err)
+		}
+		return nil
+	}
+
+	isRetryable := func(err error) bool {
+		if p.isStopped() {
+			return false
+		}
+		return task.RetryErr(err)
```
vancexu: Please make sure all tasks had a configurable max_retry count in RetryErr

common/task/interface.go
```diff
@@ -44,24 +49,35 @@ type (
...
 		// IsEmpty indicate if the task set is empty
 		IsEmpty() bool
 		// Len return the size of the queue
 		Len() int
 	}
 )
+
+const (
+	// TaskStatePending is the state for a task when it's waiting to be processed or currently being processed
+	TaskStatePending State = iota
```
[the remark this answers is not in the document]
yycptt: Done.

common/task/parallelTaskProcessor.go
yycptt: Sure.

PR #2996, opened 2020-01-23
Implement GetRawHistoryAPI for long poll

anish531213:


.gen/go/cadence/workflowserviceclient/client.go
```diff
@@ -126,6 +126,12 @@ type Interface interface {
 		opts ...yarpc.CallOption,
 	) (*shared.ListWorkflowExecutionsResponse, error)
 
+	LongPollWorkflowExecutionRawHistory(
```
yux0: PollForWorkflowExecutionRawHistory?

anish531213: Updated.

vancexu:
Could you explain why we need another API for long poll raw history? Can we do similar thing as existing GetWorkflowExecutionHistory to support long poll?

Issue #2929, opened 2019-12-20
Log domain/workflow/run IDs for non-deterministic error from RespondDecisionTaksFail


longquanzheng:
for troubleshooting issues for customers

venkat1109:
Also emit domain specific metrics on non-deterministic errors

PR #3012, opened 2020-01-28
Log domain/workflow/run ids for non-deterministic error from RespondD…

luowenking1:
…ecisionTaskFail (#2929)


service/history/handler.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/uber/cadence/.gen/go/history/historyserviceserver"
 	r "github.com/uber/cadence/.gen/go/replicator"
 	gen "github.com/uber/cadence/.gen/go/shared"
+	workflow "github.com/uber/cadence/.gen/go/shared"
```
yycptt: L37 already imported the package.

common/metrics/defs.go
```diff
@@ -1516,6 +1516,8 @@ const (
 	MatchingClientForwardedCounter
 	MatchingClientInvalidTaskListName
 
+	NonDeterministicErrorCount
```
yycptt: Please add the metric to the `MetricDefs` map.

yycptt: Please also group this metric with other error metrics starting from L1431.

service/history/handler.go
```diff
@@ -550,6 +551,14 @@ func (h *Handler) RespondDecisionTaskFailed(
 		token.RunID,
 		token.ScheduleID))
 
+	if failedRequest != nil && failedRequest.GetCause() == workflow.DecisionTaskFailedCauseUnhandledDecision {
+		h.GetLogger().Info("Non-Deterministic Error", tag.WorkflowRunID(token.RunID))
```
yycptt: Please also tag the domainID and workflowID

service/history/handler.go
```diff
@@ -550,6 +551,14 @@ func (h *Handler) RespondDecisionTaskFailed(
...
+		domainName, err := h.GetDomainCache().GetDomainName(token.DomainID)
+		if err == nil {
+			nonDeterministic := h.GetMetricsClient().Scope(metrics.HistoryRespondDecisionTaskFailedScope, metrics.DomainTag(domainName))
+			nonDeterministic.IncCounter(metrics.NonDeterministicErrorCount)
```
yycptt: `scope` is already defined on L527, you can reuse its value here. Also it's not necessary to create a temp variable `nonDeterministic`.

service/history/handler.go
```diff
@@ -550,6 +551,14 @@ func (h *Handler) RespondDecisionTaskFailed(
 		token.RunID,
 		token.ScheduleID))
 
+	if failedRequest != nil && failedRequest.GetCause() == workflow.DecisionTaskFailedCauseUnhandledDecision {
+		h.GetLogger().Info("Non-Deterministic Error", tag.WorkflowRunID(token.RunID))
+		domainName, err := h.GetDomainCache().GetDomainName(token.DomainID)
+		if err == nil {
```
yycptt: If err is not nil, you can emit the metric with `DomainUnknownTag`

common/metrics/defs.go
```diff
@@ -1799,6 +1798,7 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		CadenceErrBadBinaryCounter:                          {metricName: "cadence_errors_bad_binary", metricType: Counter},
 		CadenceErrClientVersionNotSupportedCounter:          {metricName: "cadence_errors_client_version_not_supported", metricType: Counter},
 		CadenceErrIncompleteHistoryCounter:                  {metricName: "cadence_errors_incomplete_history", metricType: Counter},
+		CadenceErrNonDeterministicCounter:                   {metricName: "cadence_NonDeterministic_Error", metricType: Counter},
```
vancexu: Let's follow the pattern for metricName: cadence_errors_nondeterministic

service/history/handler.go
```diff
@@ -551,12 +550,14 @@ func (h *Handler) RespondDecisionTaskFailed(
 		token.RunID,
 		token.ScheduleID))
 
-	if failedRequest != nil && failedRequest.GetCause() == workflow.DecisionTaskFailedCauseUnhandledDecision {
-		h.GetLogger().Info("Non-Deterministic Error", tag.WorkflowRunID(token.RunID))
+	if failedRequest != nil && failedRequest.GetCause() == gen.DecisionTaskFailedCauseUnhandledDecision {
+		h.GetLogger().Info("Non-Deterministic Error", tag.WorkflowDomainID(token.DomainID), tag.WorkflowID(token.WorkflowID), tag.WorkflowRunID(token.RunID))
 		domainName, err := h.GetDomainCache().GetDomainName(token.DomainID)
+		counter := metrics.CadenceErrNonDeterministicCounter
```
vancexu: Maybe metricName is more proper than counter here? since `counter` make people think it's a number.

yycptt: I would suggest create a variable for domainTag and set different for it based on the error returned. Then you can write  `h.GetMetricsClient().Scope(scope, tag).IncCounter(metrics.CadenceErrNonDeterministicCounter)` only once. Just my personal opinion. Your call.

service/history/handler.go
```diff
@@ -550,6 +550,19 @@ func (h *Handler) RespondDecisionTaskFailed(
 		token.RunID,
 		token.ScheduleID))
 
+	if failedRequest != nil && failedRequest.GetCause() == gen.DecisionTaskFailedCauseUnhandledDecision {
```
yycptt: Can you add a comment here which basically says we need a better way to determine if decision task failed due to non-deterministic error?

service/history/handler.go
```diff
@@ -550,6 +550,19 @@ func (h *Handler) RespondDecisionTaskFailed(
...
+		h.GetLogger().Info("Non-Deterministic Error", tag.WorkflowDomainID(token.DomainID), tag.WorkflowID(token.WorkflowID), tag.WorkflowRunID(token.RunID))
```
yycptt: Sorry, I forgot this in my previous comments. It's possible to also log the error details here?

Issue #3004, opened 2020-01-24
Emit domain specific metrics for DLQ


venkat1109:
Today, a dump of the DLQ is needed to figure out the domains impacted. Domain specific metrics should be of great help here.

PR #3014, opened 2020-01-29
Dlq metric emitting

luowenking1:
The pull request is created for the issue #3004

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -1659,7 +1659,7 @@ const (
 	ReplicationTasksFetched
 	ReplicationTasksReturned
 	ReplicationDLQFailed
-	ReplicationDLQDumpCount
+	ReplicationTaskNackCount
 	GetReplicationMessagesForShardLatency
 	GetDLQReplicationMessagesLatency
 	EventReapplySkippedCount
@@ -1998,7 +1998,7 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		ReplicationTasksFetched:                           {metricName: "replication_tasks_fetched", metricType: Timer},
 		ReplicationTasksReturned:                          {metricName: "replication_tasks_returned", metricType: Timer},
 		ReplicationDLQFailed:                              {metricName: "replication_dlq_enqueue_failed", metricType: Counter},
-		ReplicationDLQDumpCount:                           {metricName: "replication_dlq_dump_count", metricType: Counter},
+		ReplicationTaskNackCount:                          {metricName: "replication_dlq_dump_count", metricType: Counter},
 		GetReplicationMessagesForShardLatency:             {metricName: "get_replication_messages_for_shard", metricType: Timer},
 		GetDLQReplicationMessagesLatency:                  {metricName: "get_dlq_replication_messages", metricType: Timer},
 		EventReapplySkippedCount:                          {metricName: "event_reapply_skipped_count", metricType: Counter},
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -1659,6 +1659,7 @@ const (
 	ReplicationTasksFetched
 	ReplicationTasksReturned
 	ReplicationDLQFailed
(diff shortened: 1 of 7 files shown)


common/metrics/defs.go
```diff
@@ -1658,6 +1658,7 @@ const (
 	ReplicationTasksFetched
 	ReplicationTasksReturned
 	ReplicationDLQFailed
+	ReplicationDLQDumpCount
```
yycptt: `ReplicationTaskNackCount` as nack not necessarily means the task will be put on DLQ?

service/worker/replicator/replicationTask.go
```diff
@@ -515,6 +554,11 @@ func (t *workflowReplicationTask) Nack() {
 	t.metricsClient.RecordTimer(t.metricsScope, metrics.ReplicatorLatency, t.timeSource.Now().Sub(t.startTime))
 
 	t.state = task.TaskStateNacked
+
+	// TODO: need to discuss whether it is the good place to place the counter increment for the metrics because
```
vancexu: Sorry Wen,
Instead of emitting metrics, can we just add a log here with debugging info such as domainID, workflowID, runID, taskID?
It will still helpful for quick troubleshooting, but avoid passing domainCache everywhere and tag every replication tasks?
What your option @venkat1109 @meiliang86 ?

common/metrics/defs.go
```diff
@@ -1997,6 +1998,7 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		ReplicationTasksFetched:                           {metricName: "replication_tasks_fetched", metricType: Timer},
 		ReplicationTasksReturned:                          {metricName: "replication_tasks_returned", metricType: Timer},
 		ReplicationDLQFailed:                              {metricName: "replication_dlq_enqueue_failed", metricType: Counter},
+		ReplicationTaskNackCount:                          {metricName: "replication_dlq_dump_count", metricType: Counter},
```
yux0: Also update the metric name replication_dlq_nack_count

service/worker/replicator/replicationTask.go
```diff
@@ -128,6 +131,7 @@ func newActivityReplicationTask(
 			timeSource:    timeSource,
 			historyClient: historyClient,
 			metricsClient: metricsClient,
+			domainTag:     common.GetDomainTagByName(domainCache.GetDomainName(attr.GetDomainId())),
```
yux0: Separate it to two lines?
domainName := domainCache.GetDomainName(attr.GetDomainId())
common.GetDomainTagByName(domainName)

service/worker/replicator/replicationTask.go
```diff
@@ -515,6 +528,11 @@ func (t *workflowReplicationTask) Nack() {
 	t.metricsClient.RecordTimer(t.metricsScope, metrics.ReplicatorLatency, t.timeSource.Now().Sub(t.startTime))
 
 	t.state = task.TaskStateNacked
+
+	// TODO: need to discuss whether it is the good place to place the counter increment for the metrics because
```
yux0: We can remove this TODO as we handle the retry ourselves and NAck means put it into DLQ

common/util.go
```diff
@@ -525,3 +525,12 @@ func GetDefaultAdvancedVisibilityWritingMode(isAdvancedVisConfigExist bool) stri
 	}
 	return AdvancedVisibilityWritingModeOff
 }
+
+// GetDomainTagByName returns the corresponding tag if the domain name could be found, otherwise return un-known tag
+func GetDomainTagByName(domainName string, err error) metrics.Tag {
```
yycptt: This function signature looks kind of weird to me. How about `GetDomainTagByName(domainName string) metrics.Tag` and if domainName is empty return the unknown tag.

Actually I would suggest move the entire function to history package to avoid the cycle dependency issue. Otherwise, you still need to have an if/else statement every place you call the updated function.

PR #3016, opened 2020-01-30
Update logging tag with the correct component

yux0:
Update logging tag with the correct component

### DIFF
--- a/service/history/taskProcessor.go
+++ b/service/history/taskProcessor.go
@@ -123,15 +123,15 @@ func (t *taskProcessor) start() {
 		notificationChan := t.workerNotificationChans[i]
 		go t.taskWorker(notificationChan)
 	}
-	t.logger.Info("Queue task processor started.")
+	t.logger.Info("Task processor started.")
 }
 
 func (t *taskProcessor) stop() {
 	close(t.shutdownCh)
 	if success := common.AwaitWaitGroup(&t.workerWG, time.Minute); !success {
-		t.logger.Warn("Queue task processor timed out on shutdown.")
+		t.logger.Warn("Task processor timed out on shutdown.")
 	}
-	t.logger.Info("Queue task processor shutdown.")
+	t.logger.Info("Task processor shutdown.")
 }
 
 func (t *taskProcessor) taskWorker(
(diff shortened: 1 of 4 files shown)


service/history/taskProcessor.go
```diff
@@ -125,15 +123,15 @@ func (t *taskProcessor) start() {
 		notificationChan := t.workerNotificationChans[i]
 		go t.taskWorker(notificationChan)
 	}
-	t.logger.Info("Timer queue task processor started.")
+	t.logger.Info("Queue task processor started.")
```
vancexu: Maybe remove "Queue" and just leave it as "task processor ..."? It sounds redundant.

service/history/transferQueueActiveProcessor.go
```diff
@@ -87,7 +87,7 @@ func newTransferQueueActiveProcessor(
 		MetricScope:                        metrics.TransferActiveQueueProcessorScope,
 	}
 	currentClusterName := shard.GetService().GetClusterMetadata().GetCurrentClusterName()
-	logger = logger.WithTags(tag.ClusterName(currentClusterName))
+	logger = logger.WithTags(tag.ClusterName(currentClusterName), tag.ComponentTransferQueue)
```
vancexu: I think this is not needed, because it is already tagged at `newTransferQueueProcessor`

service/history/transferQueueStandbyProcessor.go
```diff
@@ -78,7 +78,7 @@ func newTransferQueueStandbyProcessor(
 		MaxRetryCount:                      config.TransferTaskMaxRetryCount,
 		MetricScope:                        metrics.TransferStandbyQueueProcessorScope,
 	}
-	logger = logger.WithTags(tag.ClusterName(clusterName))
+	logger = logger.WithTags(tag.ClusterName(clusterName), tag.ComponentTransferQueue)
```
vancexu: ditto

service/history/taskProcessor.go
yux0: Upated

service/history/transferQueueStandbyProcessor.go
yux0: Removed the tag

service/history/transferQueueActiveProcessor.go
yux0: Good catch. Removed the tag

PR #3017, opened 2020-01-30
FIFO and WRR Task Scheduler

yycptt:
FIFO scheduler is just an no-op implementation which forwards the call to the underlying parallel task processor. This is for testing and development purpose.

WRR scheduler goes through all the queues and dispatch tasks from that queue if exists. The number of tasks dispatched from each queue is determined by the weights.


common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -0,0 +1,195 @@
...
+	WeightedRoundRobinTaskSchedulerOptions struct {
+		Weights     []int
+		QueueSize   int
+		WorkerCount int
+		RetryPolicy backoff.RetryPolicy
+	}
+
+	weightedRoundRobinTaskSchedulerImpl struct {
+		status        int32
+		numPriorities int
```
yux0: Feels like this don't mean priority.  It is the number of queues in WRR.

common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -0,0 +1,195 @@
...
+		for priority := 0; priority != w.numPriorities; priority++ {
+			for i := 0; i != w.options.Weights[priority]; i++ {
+				select {
+				case task := <-w.taskChs[priority]:
+					// dispatched at least one task in this round
+					outstandingTasks = true
+
+					// submit will fail only when the processor has shut down,
+					// in which case we can just drop the task
+					_ = w.processor.Submit(task)
```
yux0: do you need to handle the error?

common/metrics/defs.go
```diff
@@ -1223,6 +1225,7 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		ElasticsearchDeleteWorkflowExecutionsScope:                 {operation: "DeleteWorkflowExecution"},
 		SequentialTaskProcessingScope:                              {operation: "SequentialTaskProcessing"},
 		ParallelTaskProcessingScope:                                {operation: "ParallelTaskProcessing"},
+		TaskSchedulerScope:                                         {operation: "TaskSchedulerScope"},
```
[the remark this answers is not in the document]
yycptt: Updated.

common/task/weightedRoundRobinTaskScheduler.go
yycptt: Renamed to `numQueues`

common/task/weightedRoundRobinTaskScheduler.go
yycptt: I think ignore the error and drop the task is the best option here.  The only error submit can and should return by processor is when it has been shutdown and failed to accept new tasks. In this case scheduler itself will also shutdown. We can Nack() the task and do a re-dispatch, but it will be a no-op as scheduler won't be able to accept it.

Actually I'd like to change the returned value for Submit from error to bool. Thoughts?

yycptt: Nack the task on error and added a unit test

common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -0,0 +1,197 @@
...
+		shutdownCh:   make(chan struct{}),
+		notifyCh:     make(chan struct{}, 1),
+		logger:       logger,
+		metricsScope: metricsScope,
+		options:      options,
+		processor: NewParallelTaskProcessor(
+			logger,
+			metricsScope,
+			&ParallelTaskProcessorOptions{
+				QueueSize:   wRRTaskProcessorQueueSize,
```
vancexu: I am confused on QueueSize, should it be options.QueueSize instead of `wRRTaskProcessorQueueSize`?

vancexu: If not, you probably want to renamed it taskChSize or something else

common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -0,0 +1,197 @@
...
+			case <-w.notifyCh:
+				// block until there's a new task
+			case <-w.shutdownCh:
+				return
+			}
+		}
+
+		outstandingTasks = false
+		for priority := 0; priority != w.numQueues; priority++ {
+			for i := 0; i != w.options.Weights[priority]; i++ {
```
vancexu: It would be dead loop if someone mistakenly had negative Weight. Maybe safer to use `i < w.options.Weights[priority]`

yycptt: Updated.

common/task/weightedRoundRobinTaskScheduler.go
yycptt: Synced offline. Agreed to keep the existing name.

PR #3021, opened 2020-02-03
Add config option to add prefix to emitted metrics

bbassingthwaite:
This becomes useful when two different cadence clusters need to emit metrics to the same metrics backend, or even just improves visibility into the cadence metrics since some are prefixed with cadence and others are not.

I've tested this out on the prometheus and it works as expected. My assumption is that it should work with the other metric types, but i'm not too familiar with tally and how its integration works with them. 
The good news is that unless someone specifies prefix, the behavior doesn't change.

### DIFF
--- a/common/service/config/config.go
+++ b/common/service/config/config.go
@@ -292,8 +292,8 @@ type (
 		// Tags is the set of key-value pairs to be reported
 		// as part of every metric
 		Tags map[string]string `yaml:"tags"`
-		// Prefixes all outgoing metrics with the provided string
-		Prefix string
+		// Prefix sets the prefix to all outgoing metrics
+		Prefix string `yaml:"prefix"`
 	}
 
 	// Statsd contains the config items for statsd metrics reporter
--- a/common/service/config/config.go
+++ b/common/service/config/config.go
@@ -292,6 +292,8 @@ type (
 		// Tags is the set of key-value pairs to be reported
 		// as part of every metric
 		Tags map[string]string `yaml:"tags"`
+		// Prefixes all outgoing metrics with the provided string
+		Prefix string
 	}
 
 	// Statsd contains the config items for statsd metrics reporter
--- a/common/service/config/metrics.go
+++ b/common/service/config/metrics.go
@@ -92,6 +92,7 @@ func (c *Metrics) newM3Scope(logger log.Logger) tally.Scope {
 	scopeOpts := tally.ScopeOptions{
 		Tags:           c.Tags,
 		CachedReporter: reporter,
+		Prefix:         c.Prefix,
 	}
 	scope, _ := tally.NewRootScope(scopeOpts, time.Second)
 	return scope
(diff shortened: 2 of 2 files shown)


mfateev:
I think we want a more generic abstraction of the "clusterID". For example it could be used not only for metrics, but also to prefix DB schema name, etc.

common/service/config/config.go
```diff
@@ -292,6 +292,8 @@ type (
 		// Tags is the set of key-value pairs to be reported
 		// as part of every metric
 		Tags map[string]string `yaml:"tags"`
+		// Prefixes all outgoing metrics with the provided string
```
vancexu: I think it will make golint complains because comment not start with `Prefix`

common/service/config/config.go
```diff
@@ -292,6 +292,8 @@ type (
...
+		Prefix string
```
vancexu: Let's add `yaml:", omitempty"` to this property.

vancexu: Are `Tags` not enough for Prometheus use case?  Uber use M3 and we tag "deployment: cluster_x"

bbassingthwaite: Since we have multiple different services pushing to a common prometheus backend, it makes it cleaner to namespace metrics with a prefix. Metrics like `restarts`, `build_age` make it likely that other services may clash with what cadence is exposing. 

It also makes discovery easier when you can type in `cadence_` if you provided a prefix `cadence` and you are presented a list of all the cadence metrics.

vancexu: According to this doc https://prometheus.io/docs/concepts/data_model/ , can you pass tags say `service:cadence` as prometheus label? So that different services with same metric name have different label. It should still be easy to read with a filter on `service` label.

I feel adding `Prefix` is not be a good solution because:
Imagine one have multiple cadence clusters/deployments for different tier of customers, each clusters have multiple datacenter, with prefix solution metrics name will look like `cadence_tier1_us_west_`, with more clusters and features, it may eventually like `cadence_tier1_us_west_zone1_mysqlDB_advancedVisibilityEnabled_..._`, it looks like abusing metric name.

bbassingthwaite: The prefixes in this case are used to distinguish services/teams. This cadence cluster is running for a service named `foo`, so we prefix our metrics with `foo_` to separate them from other running services. It becomes complex to manage multiple tenants running in prometheus without prefixing since there can be clashes on names and labels. 

There may be a separate cadence cluster run by a separate team for a different service and they would prefix their metrics with `bar_`.

It's quite common, if not the standard for services to expose prefix for users to decide how they want to best structure their metrics.

vancexu: Could you explain why not use labels? 
You should be able to have labels (tags) `service_name`, `cluster` ... (as many as you need) to solve the multiple tenants issues right?

bbassingthwaite: It's what our organization does as a standard.

In addition, from the prometheus [docs](https://prometheus.io/docs/practices/naming/) for best practices for naming metrics:
> ...should have a (single-word) application prefix relevant to the domain the metric belongs to.

vancexu: Ok, I synced with team and it sounds fair. 
Looks like there are many users like prefix and even tally had a prefix in scope option. 

Could you address my other small comment, then we can move forward with this PR?

vancexu: Can you also add `yaml:"prefix"`

bbassingthwaite: Thanks @vancexu, I have addressed your feedback.

PR #3022, opened 2020-02-03
Add logging when dumping in dlq #3004

luowenking1:


service/worker/replicator/replicationTask.go
```diff
@@ -514,6 +514,8 @@ func (t *workflowReplicationTask) Nack() {
 	t.metricsClient.IncCounter(t.metricsScope, metrics.ReplicatorMessages)
 	t.metricsClient.RecordTimer(t.metricsScope, metrics.ReplicatorLatency, t.timeSource.Now().Sub(t.startTime))
 
+	t.logger.Info("Request dump put in DLQ", tag.WorkflowDomainID(t.queueID.DomainID), tag.WorkflowID(t.queueID.WorkflowID), tag.WorkflowRunID(t.queueID.RunID))
```
vancexu: Please update log info to something like "Replication task went to DLQ"

vancexu: Could you also add tag taskID?

venkat1109: 1. Info logs should be meaningful to the user reading it. Over here, your log should call that a replicationTask is being moved to DLQ (rather than a request dump)

2. For readability, following style is preferred for multi-line logging with tags:

```
t.logger.Info("replicationTask moved to DLQ",
     tag.WorkflowDomainID(t.queueID.DomainID),
     tag.WorkflowID(t.queueID.WorkflowID),
     tag.WorkflowRunID(t.queueID.RunID),
)
```

3. Also print the task.TaskID() if its available. There should be a corresponding tag.TaskID() as well

PR #3023, opened 2020-02-03
Task Priority Assigner

yycptt:


service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+
+		sync.RWMutex
+		rateLimiters map[string]quotas.Limiter
+	}
+)
+
+var _ taskPriorityAssigner = (*taskPriorityAssignerImpl)(nil)
+
+const (
+	taskPriorityActiveTransfer int = iota
```
yux0: to follow the best practice, let's start from iota +1

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (a *taskPriorityAssignerImpl) Assign(
+	task queueTask,
+) error {
+	if task.GetQueueType() == common.ReplicationQueueType {
+		task.SetPriority(taskPriorityReplicationStandby)
+		return nil
+	}
+
+	// timer of transfer task, first check if domain is active or not
```
yux0: or?

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+)
+
+var _ taskPriorityAssigner = (*taskPriorityAssignerImpl)(nil)
+
+const (
+	taskPriorityActiveTransfer int = iota
+	taskPriorityActiveTimer
+	taskPriorityActiveThrottled
+	taskPriorityReplicationStandby
```
yux0: From the implementation, it looks like the taskPriorityReplicationStandby used for active replication task or standy timer/ transfer task? shoule it name taskPriorityReplication?

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil
+	}
+
+	// timer of transfer task, first check if domain is active or not
+	domainName, active, err := a.getDomainInfo(task.GetDomainID())
+	if err != nil {
+		return err
+	}
+
+	if !active {
```
yux0: do you think use switch would make those if statements more clear?

common/constants.go
```diff
@@ -129,11 +129,13 @@ const (
 	ArchivalPaused = "paused"
 )
 
-// Queue types used in queue table
-// Use positive numbers for queue type
-// Negative numbers are reserved for DLQ
 const (
-	DomainReplicationQueueType QueueType = 1
+	// TransferQueueType is the type for transfer queue
```
venkat1109: common/constants seems like a weird place for these consts. Where are they being used outside of history / persistence ? If these are only used in history/persistence, consider moving these to either history or persistence datainterfaces

common/persistence/sql/sqlQueue.go
```diff
@@ -26,15 +26,14 @@ import (
...
-	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/persistence/sql/sqlplugin"
 )
 
 type (
 	sqlQueue struct {
-		queueType common.QueueType
+		queueType persistence.QueueType
```
venkat1109: consider doing renaming in a separate PR in the future. Its easier to review as well as to revert etc

common/service/dynamicconfig/constants.go
```diff
@@ -140,6 +140,7 @@ var keys = map[Key]string{
 	StandbyClusterDelay:                                   "history.standbyClusterDelay",
 	StandbyTaskMissingEventsResendDelay:                   "history.standbyTaskMissingEventsResendDelay",
 	StandbyTaskMissingEventsDiscardDelay:                  "history.standbyTaskMissingEventsDiscardDelay",
+	TaskProcessRPS:                                        "histroy.taskProcessRPS",
```
venkat1109: fix typo - history

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	taskPriorityAssignerImpl struct {
+		currentClusterName string
+		domainCache        cache.DomainCache
+		config             *Config
+		logger             log.Logger
+		scope              metrics.Scope
+
+		sync.RWMutex
```
venkat1109: It's a common practice to declare anonymous mutex as the first member of the struct.

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return err
+	}
+
+	if !active {
+		task.SetPriority(taskPriorityReplicationStandby)
+		return nil
+	}
+
+	if !a.getRateLimiter(domainName).Allow() {
```
venkat1109: a bit confused. Why does the priority assigner also do throttling ? Doesn't look like this should belong here

service/history/taskPriorityAssigner.go
yycptt: Sure I can update. But I think in our code base we are following two different conventions: one is but the mutex as the first member of the struct as you suggested. The other one is put it just above the field the lock is protecting. Examples are `queueAckMgr.go`, `shardContext.go` etc.

service/history/taskPriorityAssigner.go
yycptt: It's not doing throttling. This is a just non-blocking call for checking if the number of tasks from a certain domain exceeds limit. If `Allow()` returns false, then it means that domain has generated too many tasks, and laters tasks will be assigned a low priority (`taskPriorityActiveThrottled`). Tasks will still be sent for processing though.

common/persistence/sql/sqlQueue.go
yycptt: Sure. Will follow this in the future.

common/service/dynamicconfig/constants.go
yycptt: Done.

service/history/taskPriorityAssigner.go
yycptt: Updated.

service/history/taskPriorityAssigner.go
yycptt: Synced offline. Will keep the current way it's written.

service/history/taskPriorityAssigner.go
```diff
@@ -0,0 +1,164 @@
...
+		scope              metrics.Scope
+		rateLimiters       map[string]quotas.Limiter
+	}
+)
+
+var _ taskPriorityAssigner = (*taskPriorityAssignerImpl)(nil)
+
+const (
+	taskPriorityActiveTransfer int = iota
+	taskPriorityActiveTimer
```
venkat1109: Thought about this some more yesterday. I believe what you need here is some sort of priority_class and map the incoming tasks to one of those. Right now, based on your declaration, the task_type and priority are coupled with each other which gives you less flexibility. My proposal is something along the following lines:

- You define some fixed priority_classes and may be also sub-classes. For example, let's say there is a priority_class and a priority_subclass. You can allocate 3 bits for class and another 3 bits for sub-class. To start with, something simple like 3 priority classes - high, default, low would suffice. We can add more later if needed. As for the necessity for sub-classes, it allows you to differentiate between two tasks within same priority class. 

- You assign tasks to priority_class/sub_class i.e. one single integer priority value. For example, both active_transfer_tasks and active_timer_tasks can be high priority but workflow_timeout task can be low_priority. As another example, some domains may be latency sensitive, so, the active_transfer_tasks for these domain can be a high priority sub-class while other active_transfer_tasks can be a low-priority sub-class.

service/history/taskPriorityAssigner.go
venkat1109: This is fine for now. But ideally, rather than using a fixed throttle per domain, you should consider allocating ratios within the available total quota i.e. something like - no single domain can consume more than 10% of available task_processing_capacity on a host (or) lets throttle domainFoo to consume only 1% total available processing capacity on this host.

service/history/taskPriorityAssigner.go
venkat1109: For the latter, anonymous mutexes don't make sense since the mutex is only protecting a sub-resource within the struct. For these cases, the pattern should to be use a named mutex variable. For example, if mutex protects some resourceMap, then mutex should be named resourceMapMutex.

yycptt: Got it! Thanks for explaining.

service/history/taskPriorityAssigner.go
yycptt: Makes sense. Let's sync offline regarding the implementation details.

yycptt: Synced offline. Now using 3 bits for priority_class and another 3 for priority_subclass. For now each level only has three values: high, default and low. We can add more later if necessary.

The mapping from task -> priority is: 
active timer/transfer -> high (priority_class), default (priority_subclass)
throttled (based on domain) timer/transfer -> default (priority_class), default (priority_subclass)
replication, standby timer/transfer -> low (priority_class), default (priority_subclass)

We can change the mapping later. For example, use different subclass based on the exact type of timer/transfer task.

PR #3026, opened 2020-02-05
Add metrics for profiling activity timeout task latency

yycptt:
Revert this change after root causing the issue.


service/history/timerQueueActiveProcessor.go
```diff
@@ -369,6 +371,11 @@ func (t *timerQueueActiveProcessorImpl) processActivityTimeout(
 		return err
 	}
 	if mutableState == nil || !mutableState.IsWorkflowExecutionRunning() {
+		t.metricsClient.RecordTimer(
```
yux0: Why record the timer here?

yycptt: It's possible the function will return here without any error. so emit a metric to see the latency.

PR #3028, opened 2020-02-05
Implement GetRawHistoryAPI for long poll

anish531213:


service/frontend/accessControlledHandler.go
```diff
@@ -243,6 +243,27 @@ func (a *AccessControlledWorkflowHandler) GetWorkflowExecutionRawHistory(
 	return a.frontendHandler.GetWorkflowExecutionRawHistory(ctx, request)
 }
 
+// GetRawHistoryLongPull API call
```
yux0: update PollForWorkflowExecutionRawHistory

service/frontend/dcRedirectionHandler.go
```diff
@@ -303,6 +303,36 @@ func (handler *DCRedirectionHandlerImpl) GetWorkflowExecutionRawHistory(
 	return resp, err
 }
 
+// GetRawHistory API call
```
yux0: update to PollForWorkflowExecutionRawHistory

service/frontend/workflowHandler.go
```diff
@@ -1833,6 +1833,207 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
...
+				nextEventID,
+				getRequest.GetMaximumPageSize(),
+				nil,
+				token.TransientDecision,
+				token.BranchToken,
+			)
+			if err != nil {
+				return nil, wh.error(err, scope)
+			}
+			// since getHistory func will not return empty history, so the below is safe
```
yux0: Make sure this is still true for getRawHistory

service/frontend/workflowHandler.go
```diff
@@ -1833,6 +1833,207 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
...
+				queryHistory(domainID, execution, queryNextEventID, token.BranchToken)
+			if err != nil {
+				return nil, wh.error(err, scope)
+			}
+			token.FirstEventID = token.NextEventID
+			token.NextEventID = nextEventID
+			token.IsWorkflowRunning = isWorkflowRunning
+		}
+	} else {
+		if !isCloseEventOnly {
```
yux0: The isCloseEventOnly in the original logic also control the get history or poll for history. The logic inside this else is for get history. I don't think you need to have those in this long poll API

service/frontend/workflowHandler.go
```diff
@@ -1834,6 +1834,192 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
...
+
+		return response.CurrentBranchToken,
+			response.Execution.GetRunId(),
+			response.GetLastFirstEventId(),
+			response.GetNextEventId(),
+			isWorkflowRunning,
+			nil
+	}
+
+	isCloseEventOnly := getRequest.GetHistoryEventFilterType() == gen.HistoryEventFilterTypeCloseEvent
```
yux0: As discussed, we can remove this filter type. The logic would be simpler

service/frontend/workflowHandler.go
```diff
@@ -1834,6 +1834,165 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
 	}, nil
 }
 
+// PollForWorkflowExecutionRawHistory get Raw History for long poll request
+func (wh *WorkflowHandler) PollForWorkflowExecutionRawHistory(
```
yux0: It would be good if you can add some tests for this API.

service/frontend/workflowHandler.go
```diff
@@ -1834,6 +1834,137 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
...
+	if err != nil {
+		return nil, wh.error(err, scope)
+	}
+
+	execution.RunId = &runID
+
+	token.RunID = runID
+	token.FirstEventID = common.FirstEventID
+	token.NextEventID = nextEventID
+	token.IsWorkflowRunning = isWorkflowRunning
```
yux0: Those are not used

service/frontend/workflowHandler.go
```diff
@@ -1834,6 +1834,137 @@ func (wh *WorkflowHandler) GetWorkflowExecutionRawHistory(
...
+	)
+	if err != nil {
+		return nil, wh.error(err, scope)
+	}
+	// since getHistory func will not return empty history, so the below is safe
+	history = history[len(history)-1 : len(history)]
+
+	return &gen.PollForWorkflowExecutionRawHistoryResponse{
+		RawHistory:    history,
+		NextPageToken: nil,
```
yux0: add the token here to return

Issue #2380, opened 2019-08-13
Replace EmitMetrics flag with SkipMetrics


mfateev:
The general pattern should be to have flags that by default provide the most meaningful experience. This flag had to be SkipMetrics to comply.

PR #3029, opened 2020-02-06
Remove EmitMetric option from the CLI #2380

luowenking1:
Instead of renaming the property which may call the back-end breaking, I removed it from CLI and mark the property as depreciated.


common/persistence/persistenceInterface.go
```diff
@@ -599,7 +599,7 @@ type (
 	InternalDomainConfig struct {
 		// NOTE: this retention is in days, not in seconds
 		Retention                int32
-		EmitMetric               bool
+		EmitMetric               bool                    // deprecated
```
venkat1109: you can completely get rid of this. There is no concern with backward compatibility within the internal server code

tools/cli/flags.go
```diff
@@ -123,8 +123,8 @@ const (
 	FlagOwnerEmailWithAlias               = FlagOwnerEmail + ", oe"
 	FlagRetentionDays                     = "retention"
 	FlagRetentionDaysWithAlias            = FlagRetentionDays + ", rd"
-	FlagEmitMetric                        = "emit_metric"
-	FlagEmitMetricWithAlias               = FlagEmitMetric + ", em"
+	FlagEmitMetric                        = "emit_metric"           // deprecated
```
venkat1109: completely delete these. This is just internal constant

luowenking1:
Pull request for task #2380

.gen/go/shared/shared.go
```diff
@@ -33318,7 +33319,10 @@ func (v *RegisterDomainRequest) ToWire() (wire.Value, error) {
 		fields[i] = wire.Field{ID: 40, Value: w}
 		i++
 	}
-	if v.EmitMetric != nil {
+	if v.EmitMetric == nil {
```
venkat1109: I don't see the cadence-idl submodule update corresponding to this. Is that included ?

tools/cli/domainUtils.go
```diff
@@ -119,10 +115,6 @@ var (
 			Name:  FlagRetentionDaysWithAlias,
 			Usage: "Workflow execution retention in days",
 		},
-		cli.StringFlag{
```
venkat1109: also please double check if this flag is currently being accessed in things like docker/scripts/buildkite etc. This is a breaking change i.e. will break scripts which already have this flag

.gen/go/shared/shared.go
luowenking1: The idl update was committed to repo idl with this pr https://github.com/uber/cadence-idl/pull/10

luowenking1: > I don't see the cadence-idl submodule update corresponding to this. Is that included ?

The idl update was committed to repo idl with this pr uber/cadence-idl#10

tools/cli/app_test.go
```diff
@@ -178,7 +178,7 @@ func (s *cliAppSuite) TestDomainUpdate() {
 	s.serverFrontendClient.EXPECT().UpdateDomain(gomock.Any(), gomock.Any()).Return(nil, nil).Times(2)
 	err := s.app.Run([]string{"", "--do", domainName, "domain", "update"})
 	s.Nil(err)
-	err = s.app.Run([]string{"", "--do", domainName, "domain", "update", "--desc", "another desc", "--oe", "another@uber.com", "--rd", "1", "--em", "f"})
+	err = s.app.Run([]string{"", "--do", domainName, "domain", "update", "--desc", "another desc", "--oe", "another@uber.com", "--rd", "1", "f"})
```
yycptt: also remove "f"?

luowenking1: I have removed it. Thanks.

tools/cli/domainUtils.go
luowenking1: > also please double check if this flag is currently being accessed in things like docker/scripts/buildkite etc. This is a breaking change i.e. will break scripts which already have this flag

I have checked and resolved them in the unit tests.

PR #3034, opened 2020-02-08
Add database API to list current executions

andrewjdawson2016:


common/persistence/persistenceMetricClients.go
```diff
@@ -330,6 +330,20 @@ func (p *workflowExecutionPersistenceClient) GetCurrentExecution(request *GetCur
 	return response, err
 }
 
+func (p *workflowExecutionPersistenceClient) ListCurrentExecutions(request *ListCurrentExecutionsRequest) (*ListCurrentExecutionsResponse, error) {
+	p.metricClient.IncCounter(metrics.PersistenceListCurrentExecutionsScope, metrics.PersistenceRequests)
+
+	sw := p.metricClient.StartTimer(metrics.PersistenceGetCurrentExecutionScope, metrics.PersistenceLatency)
```
venkat1109: typo - fix Scope name

common/persistence/persistenceMetricClients.go
```diff
@@ -330,6 +330,20 @@ func (p *workflowExecutionPersistenceClient) GetCurrentExecution(request *GetCur
...
+	response, err := p.persistence.ListCurrentExecutions(request)
+	sw.Stop()
+
+	if err != nil {
+		p.updateErrorMetric(metrics.PersistenceGetCurrentExecutionScope, err)
```
venkat1109: fix scope name

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -437,6 +437,12 @@ workflow_state = ? ` +
 		`and visibility_ts = ? ` +
 		`and task_id = ?`
 
+	templateListCurrentExecutionsQuery = `SELECT domain_id, workflow_id, current_run_id ` +
+		`FROM executions ` +
+		`WHERE shard_id = ? ` +
+		`and type = ? ` +
+		`and run_id = ?`
```
yycptt: Tried this query on my local machine and got 
```
InvalidRequest: Error from server: code=2200 [Invalid query] message="PRIMARY KEY column "run_id" cannot be restricted as preceding column "domain_id" is not restricted"
```
Add a TODO here?

venkat1109: good catch. we actually need to filter on domainID - we have a cookie cutter domainID for currentExecutions

yycptt: Please correct me if I am wrong. The primary key for the table is `(shard_id, type, domain_id, workflow_id, run_id, visibility_ts, task_id)`. If we have filter on `run_id`, then I think all previous columns need to be filtered, which means we also need to add a filter on `workflow_id`.

andrewjdawson2016:
Decided to implement this with visibility API instead of direct database scan.

PR #3035, opened 2020-02-08
Add ability to cache scopes

andrewjdawson2016:
This adds the ability in our metricsClient to cache scopes. We have seen that scope construction contributes significantly to GC time. 

This code turns on caching for all metrics scopes with no config to disable. This was done this way as a first pass because it produces the simplest API, we should discuss further if we need to enable or disable this feature in different places in code.

### DIFF
--- a/common/metrics/client.go
+++ b/common/metrics/client.go
@@ -21,6 +21,7 @@
 package metrics
 
 import (
+	"strings"
 	"time"
 
 	"github.com/uber-go/tally"
@@ -29,10 +30,11 @@ import (
 // ClientImpl is used for reporting metrics by various Cadence services
 type ClientImpl struct {
 	//parentReporter is the parent scope for the metrics
-	parentScope tally.Scope
-	childScopes map[int]tally.Scope
-	metricDefs  map[int]metricDefinition
-	serviceIdx  ServiceIdx
+	parentScope  tally.Scope
+	childScopes  map[int]tally.Scope
+	metricDefs   map[int]metricDefinition
+	serviceIdx   ServiceIdx
+	cachedScopes map[string]Scope
 }
 
 // NewClient creates and returns a new instance of
@@ -42,10 +44,11 @@ type ClientImpl struct {
 func NewClient(scope tally.Scope, serviceIdx ServiceIdx) Client {
 	totalScopes := len(ScopeDefs[Common]) + len(ScopeDefs[serviceIdx])
 	metricsClient := &ClientImpl{
-		parentScope: scope,
-		childScopes: make(map[int]tally.Scope, totalScopes),
-		metricDefs:  getMetricDefs(serviceIdx),
-		serviceIdx:  serviceIdx,
+		parentScope:  scope,
+		childScopes:  make(map[int]tally.Scope, totalScopes),
+		metricDefs:   getMetricDefs(serviceIdx),
+		serviceIdx:   serviceIdx,
+		cachedScopes: make(map[string]Scope),
 	}
 
 	for idx, def := range ScopeDefs[Common] {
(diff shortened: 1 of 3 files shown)


common/metrics/client.go
```diff
@@ -125,3 +135,13 @@ func mergeMapToRight(src map[string]string, dest map[string]string) {
 		dest[k] = v
 	}
 }
+
+func stringifyScope(scopeIdx int, tags ...Tag) string {
+	var b strings.Builder
+	b.WriteString(string(scopeIdx))
```
venkat1109: this looks like atleast as expensive (if not more) as creating a sub-scope itself. Try writing a benchmark test to compare allocations with the baseline

common/metrics/client.go
```diff
@@ -29,10 +30,11 @@ import (
...
 	//parentReporter is the parent scope for the metrics
-	parentScope tally.Scope
-	childScopes map[int]tally.Scope
-	metricDefs  map[int]metricDefinition
-	serviceIdx  ServiceIdx
+	parentScope  tally.Scope
+	childScopes  map[int]tally.Scope
+	metricDefs   map[int]metricDefinition
+	serviceIdx   ServiceIdx
+	cachedScopes map[string]Scope
```
venkat1109: how is this thread safe ?

common/metrics/client.go
```diff
@@ -104,7 +107,14 @@ func (m *ClientImpl) UpdateGauge(scopeIdx int, gaugeIdx int, value float64) {
...
 // information to the metrics emitted
 func (m *ClientImpl) Scope(scopeIdx int, tags ...Tag) Scope {
-	return newMetricsScope(m.childScopes[scopeIdx], m.metricDefs, false).Tagged(tags...)
+	strScope := stringifyScope(scopeIdx, tags...)
+	scope, ok := m.cachedScopes[strScope]
+	if ok {
+		return scope
+	}
+	scope = newMetricsScope(m.childScopes[scopeIdx], m.metricDefs, false).Tagged(tags...)
+	m.cachedScopes[strScope] = scope
```
venkat1109: You might want to use a bounded size map here just to be defensive - eventhough the size itself can be rather large. When the size is exceeded, some simple strategy such as evicting at random is prolly fine to start with. Iff the size is chosen carefully, no eviction is likely expected.

common/metrics/client.go
venkat1109: does this work ie. type casting int to string ?

common/metrics/scope.go
```diff
@@ -30,15 +31,26 @@ type metricsScope struct {
 	scope          tally.Scope
 	defs           map[int]metricDefinition
 	isDomainTagged bool
+	cachedScopes   map[string]Scope
```
venkat1109: this also needs to be bounded and the bound prolly needs to be way smaller than the one in the metricsClient.

common/metrics/client.go
andrewjdawson2016: Its not right now, I need to think through this thread safety still.

common/metrics/scope.go
```diff
@@ -30,15 +31,26 @@ type metricsScope struct {
...
 }
 
 func newMetricsScope(scope tally.Scope, defs map[int]metricDefinition, isDomain bool) Scope {
-	return &metricsScope{scope, defs, isDomain}
+	return &metricsScope{
+		scope:          scope,
+		defs:           defs,
+		isDomainTagged: isDomain,
+		cachedScopes:   make(map[string]Scope),
```
yycptt: Is it possible to share the cache between `metricsClient` and `metricsScope`?

I think people use `metricsClient.Scope(idx, tags)` and `metricsScope.Tag(tags)` (if any existing scope is available) interchangeably. If the underlying cache are separate, then one will reuse the cached scope while the other one will create a new scope and I think it's a little bit confusing.

andrewjdawson2016:
We actually don't need to cache scopes we just needed to reduce duplication of domain records on read.

PR #3043, opened 2020-02-12
dynamic config: optimize to reduce heap allocs

venkat1109:
Recently, I was running some heap traces and noticed that the dynamic config code path was doing unexpected number of allocations. The root cause was a method called logValue() which attempts to log dynamic config (key, value) whenever there is a change in value. It uses a map to store old key/value - but the key for this map is a string which is constructed using fmt.Sprintf() - this is the cause for un-necessary allocations. This patch fixes the problem by getting rid of the fmt.Sprintf()

**Before:**
BenchmarkLogValue-8         	 1000000	      1226 ns/op	     243 B/op	      10 allocs/op

**After:**
BenchmarkLogValue-8         	 6978836	       172 ns/op	      24 B/op	       3 allocs/op

heap_profile_logvalue


common/service/dynamicconfig/config.go
```diff
@@ -21,40 +21,54 @@
...
 // can be directly accessed by calling the function without propagating the client everywhere in
 // code
 type Collection struct {
-	client Client
-	logger log.Logger
-	keys   *sync.Map
+	client   Client
+	logger   log.Logger
+	keys     *sync.Map // map of config Key to strongly typed value
+	errCount int64
```
yux0: int or int32 should be enough?

venkat1109: I can change this to int32 - but I am not sure if there is actually any benefit of doing so. Most modern processors are 64-bit and have 64 bit registers. For the same reason, int will be same as int64 on any modern server.

PR #3049, opened 2020-02-14
Add an entry point for custom datastore implementation

vitarb:
This PR provides an extension point for adding a new datastore outside of cadence.

Users can implement AbstractDataStoreFactory interface, which needs to return a datastore factory, which in its turn needs to be able to create implementations of all datastore gateways. 

AbstractDataStoreFactory can be configured using a separate section in the config, which currently consists of a data store name and a map of parameters.

Future work:
* CLI - some plumbing needs to be done to allow users to extend CLI and provide AbstractDataStoreFactory during the startup.
* Testing - we need to define approach for testing of custom datastores.


tools/cli/domainUtils.go
```diff
@@ -298,6 +298,7 @@ func initializeMetadataMgr(
 	}
 	pFactory := client.NewFactory(
 		&pConfig,
+		nil, // TODO propagate abstract datastore factory from the CLI.
```
vancexu: How do you plan to use CLI to propagate the factory? 
I thought the factory implementation will be needed before/during cadence cluster setup, but no CLI command will work before Cadence are setup successfully.

Maybe what we need is a global CustomDataStoreFactory, and user will SetCustomDataStoreFactory when they start Cadence.

common/persistence/client/factory.go
```diff
@@ -73,18 +73,25 @@ type (
 		NewVisibilityStore() (p.VisibilityStore, error)
 		NewQueue(queueType p.QueueType) (p.Queue, error)
 	}
+	// AbstractDataStoreFactory creates a DataStoreFactory, can be used to implement custom datastore support outside
+	// of the cadence core.
+	AbstractDataStoreFactory interface {
```
vancexu: Maybe rename as CustomDatastoreFactory? (to be consistent with CustomDatastoreConfig)

common/service/service.go
```diff
@@ -58,24 +60,25 @@ type (
...
-		DispatcherProvider  client.DispatcherProvider
-		DCRedirectionPolicy config.DCRedirectionPolicy
-		PublicClient        workflowserviceclient.Interface
-		ArchivalMetadata    archiver.ArchivalMetadata
-		ArchiverProvider    provider.ArchiverProvider
-		Authorizer          authorization.Authorizer
+		MetricScope              tally.Scope
+		MembershipFactory        MembershipMonitorFactory
+		RPCFactory               common.RPCFactory
+		AbstractDatastoreFactory persistenceClient.AbstractDataStoreFactory
```
venkat1109: Summarizing what we discussed offline, there are two possible approaches here:

(1) Allow passing AbstractFactory here i.e. what this PR currently does
(2) Allow passing an override for the persistence factory as a whole. With this approach, the internal repo will have its own implementation of persistence factory - which knows how to build internal data store as well as other data stores that persistence.factoryImpl() knows how to build i.e. it will use composition 

The disadvantage of (1) is that you have to pass a map[string]string connectAttributes in CustomDataStoreConfig. 
The disadvantage of (2) is that there is some additional code to re-implement factory interface within the internal repo.

After looking at this PR, I am getting convinced that the amount of additional code is going to be smallish if we go with (2) i.e. both (1) and (2) are simple enough to implement. Given this, I prefer (2) because it looks cleaner in that the public repo won't have these customXXX / opaque maps  that are basically pass through. 

Let me know what you think.

venkat1109: nvm - your current approach can also be made to work without having to pass map[string]string connectAttributes. i.e. both approaches will need to pass CustomDataStoreConfig - I don't see a way around it. So, your current approach looks just fine to me.

common/persistence/client/factory.go
```diff
@@ -119,15 +126,17 @@ var storeTypes = []storeType{
 // given configuration. In addition, all objects will emit metrics automatically
 func NewFactory(
 	cfg *config.Persistence,
+	abstractDataStoreFactory AbstractDataStoreFactory,
```
venkat1109: instead of adding another arg to this constructor, can we add a member to config.Persistence called `CustomDataStoreFactory` for which no value can be specified via config i.e. yaml: "-"

common/persistence/client/factory.go
vitarb: What this thing does falls exactly under the definition of an abstract factory pattern, hence IMO this is the least confusing name for it.

common/service/service.go
vitarb: That's right.

common/persistence/client/factory.go
vitarb: That's an option, I didn't want to add it to the config because logically it's not a config but rather an implementation, so although code will be a bit cleaner it would also be a bit more confusing doing it that way.

venkat1109: that's fine by me.. in that case, consider making the AbstractDataStoreFactory an optional arg to the constructor in persistence.NewFactory.. [not a blocker ]

tools/cli/domainUtils.go
vitarb: CLI would have to be invoked through internal binary, which would populate custom datastore factory, similarly to how main binary works.

PR #3052, opened 2020-02-14
Make persistence config dynamic for system scanner

andrewjdawson2016:
In order to safely implement executions scanner we need the persistence RPS to be controllable through dynamic config. This diff addresses the two issues that had prevented us from doing this:

1. Persistence RPS was not truly dynamic. The dynamic config value was read once when service components booted up and then never read again. I changed this to plumb the dynamic config through everywhere. This is safe because we were never relying on the value in static yaml config.
2. The worker only had one resource.Resource (with one persistence RPS) shared between all features of the worker. This made it impossible to set a persistence RPS for one worker feature and a different RPS for another worker feature. In order to address this I added a new field on worker for the resources for scanner. Before my change we attempted to this by setting the PersistenceMaxRPS in the scanner before we start the scanner, but this wasn't working because the resources (including the rate limited persistence resources) had already been created.

### DIFF
--- a/common/persistence/client/factory.go
+++ b/common/persistence/client/factory.go
@@ -23,6 +23,8 @@ package client
 import (
 	"sync"
 
+	"github.com/uber/cadence/common/service/dynamicconfig"
+
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/metrics"
 	p "github.com/uber/cadence/common/persistence"
@@ -138,7 +140,7 @@ func NewFactory(
 		logger:                   logger,
 		clusterName:              clusterName,
 	}
-	limiters := buildRatelimiters(cfg)
+	limiters := buildRateLimiters(cfg)
 	factory.init(clusterName, limiters)
 	return factory
 }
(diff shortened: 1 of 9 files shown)


common/persistence/client/factory.go
```diff
@@ -23,6 +23,8 @@ package client
 import (
 	"sync"
 
+	"github.com/uber/cadence/common/service/dynamicconfig"
+
```
venkat1109: Looks like unexpected newline - goimports ?

common/service/config/config.go
```diff
@@ -209,7 +209,7 @@ type (
 		// ConnectAttributes is a set of key-value attributes to be sent as part of connect data_source_name url
 		ConnectAttributes map[string]string `yaml:"connectAttributes"`
 		// MaxQPS the max request rate on this datastore
-		MaxQPS int `yaml:"maxQPS"`
+		MaxQPS dynamicconfig.IntPropertyFn `yaml:"-" json:"-"`
```
venkat1109: this PR seems to have some overlap with https://github.com/uber/cadence/pull/3062

andrewjdawson2016: Sync'ed offline about this.

PR #3055, opened 2020-02-17
Skeleton of executions scanner

andrewjdawson2016:
This diff includes the skeleton of a new system scanner which will go over workflow executions. This scanner is meant to be used to assert system invariants, generate reports and someday could also automatically perform repair actions. 

This code contains all the plumbing needed to actually bootstrap this workflow and run it. This diff does not contain the actual implementation of the validations or unit tests (these will come in follow up diffs).

### DIFF
--- a/service/worker/scanner/executions/scavenger.go
+++ b/service/worker/scanner/executions/scavenger.go
@@ -1,17 +1,17 @@
 // The MIT License (MIT)
-// 
+//
 // Copyright (c) 2020 Uber Technologies, Inc.
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
+//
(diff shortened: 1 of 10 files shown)


service/worker/scanner/executions/scavenger.go
```diff
@@ -0,0 +1,193 @@
...
+	// 5. emit metrics on the run of scavenger
+}
+
+func (s *Scavenger) awaitExecutor() {
+	outstanding := s.executor.TaskCount()
+	for outstanding > 0 {
+		select {
+		case <-time.After(executorPollInterval):
+			outstanding = s.executor.TaskCount()
+			s.metrics.UpdateGauge(metrics.ExecutionsScavengerScope, metrics.TaskListOutstandingCount, float64(outstanding))
```
venkat1109: fix metrics name

service/worker/scanner/executions/scavenger.go
```diff
@@ -0,0 +1,193 @@
...
+var (
+	executionsBatchSize      = 32   // maximum number of executions we process concurrently
+	executionsPageSize       = 1000 // page size of executions read from visibility manager
+	executorPollInterval     = time.Minute
+	executorMaxDeferredTasks = 10000
+)
+
+// NewScavenger returns an instance of executions scavenger daemon
+// The Scavenger can be started by calling the Start() method on the
+// returned object. Calling the Start() method will result in one
```
venkat1109: Would be nice if we are able to resume from where we left off previously. Not a blocker though

andrewjdawson2016: I agree - I filed a ticket https://github.com/uber/cadence/issues/3064

PR #3057, opened 2020-02-18
Fix SignalExternalWorkflow input should be optional

vancexu:
https://github.com/uber/cadence/issues/3056


service/history/decisionChecker.go
```diff
@@ -456,9 +456,6 @@ func (v *decisionAttrValidator) validateSignalExternalWorkflowExecutionAttribute
 	if attributes.SignalName == nil {
 		return &workflow.BadRequestError{Message: "SignalName is not set on decision."}
 	}
-	if attributes.Input == nil {
```
venkat1109: do we have integration tests that make sure our clients (go/java) don't break during replay when they encounter signal with no input ?

vancexu: Thanks for pointing out, this PR will actually be a breaking change.

mfateev: I don't think it is breaking as it cannot break already running workflows.

vancexu: @venkat1109 I did some manual tests with both go and java client, this change shouldn't break customer.
@mfateev yes you are right, I confused myself previously, it is not a breaking change.

PR #3059, opened 2020-02-19
Add canary test for visibility archival

yycptt:


canary/archival.go
```diff
@@ -158,6 +190,85 @@ func archivalExternalWorkflow(ctx workflow.Context, scheduledTimeNanos int64) er
...
+	scope, sw := recordActivityStart(scope, activityTypeVisibilityArchival, scheduledTimeNanos)
+	defer recordActivityEnd(scope, sw, err)
+
+	client := getActivityArchivalContext(ctx).cadence
+	resp, err := client.Describe(ctx, archivalDomain)
+	if err != nil {
+		return err
+	}
+
+	if resp.Configuration != nil &&
```
andrewjdawson2016: Why would this occur?

canary/archival.go
```diff
@@ -158,6 +190,85 @@ func archivalExternalWorkflow(ctx workflow.Context, scheduledTimeNanos int64) er
...
+		resp.Configuration.GetVisibilityArchivalStatus() == shared.ArchivalStatusDisabled {
+		return errors.New("domain not configured for visibility archival")
+	}
+
+	visArchivalURI := ""
+	if resp.Configuration != nil {
```
andrewjdawson2016: Will Configuration ever not be set even if we get a response?

canary/archival.go
```diff
@@ -158,6 +190,85 @@ func archivalExternalWorkflow(ctx workflow.Context, scheduledTimeNanos int64) er
...
+		visArchivalURI = resp.Configuration.GetVisibilityArchivalURI()
+	}
+
+	scheme := getURIScheme(visArchivalURI)
```
andrewjdawson2016: Users don't actually have do this right? You aren't using scheme anywhere in the request? 

I think unless we are really testing something by verifying this scheme we should just have canary interact with this feature like a user would.

canary/archival.go
```diff
@@ -158,6 +190,85 @@ func archivalExternalWorkflow(ctx workflow.Context, scheduledTimeNanos int64) er
...
+	if scheme != visArchivalScheme {
+		return fmt.Errorf("unknown visibility archival scheme: %s, expecting %s", scheme, visArchivalScheme)
+	}
+
+	listReq := &shared.ListArchivedWorkflowExecutionsRequest{
+		Domain:   stringPtr(archivalDomain),
+		PageSize: int32Ptr(visArchivalPageSize),
+		Query: stringPtr(fmt.Sprintf(visArchivalQueryTemplate,
+			wfTypeArchival,
+			time.Now().Add(-time.Hour*12).UnixNano(),
```
andrewjdawson2016: Factor out into constants.

canary/archival.go
yycptt: I don't think this will occur. Including this check for safety so that the code won't panic. I can remove if you have strong opinion on this.

Same below.

canary/archival.go
```diff
@@ -158,6 +190,85 @@ func archivalExternalWorkflow(ctx workflow.Context, scheduledTimeNanos int64) er
...
+		}
+
+		if listResp.NextPageToken == nil {
+			break
+		}
+
+		listReq.NextPageToken = listResp.NextPageToken
+	}
+
+	if len(listResp.Executions) == 0 {
```
andrewjdawson2016: Roughly how many do we expect to have here? Seems like if we can assert something stronger than more than 0 it would be good.

canary/archival.go
yycptt: User won't do this as we ensure the right URI is set for them. But for canary, I think it makes sense to perform this check so can we can identify the problem faster when the test fails (either due to misconfiguration or there's something wrong with visibility archival).

canary/archival.go
yycptt: Will do.

canary/archival.go
yycptt: I put more thought on this last night and made the following two changes:
1. Separate the canary test for history archival and visibility archival so they can enabled and disabled independently.
2. Created an interface for validating visibility archival (a template) which each visibility archival scheme should implement. It basically tells canary if the test should be run, what is the query and how to validate the result. With this interface I believe the code is general enough for all archiver implementations and each implementation can have more control over the test. 

I made those changes because I think when open source customer uses canary in production, it's unlikely they set the scheme for canary domain to filestore, it will be either s3 or gcp or other implementation. So we should make the code more general and flexible. 

What's your thought on this?

andrewjdawson2016: This makes a lot of sense :)

PR #3066, opened 2020-02-21
Add ServerName to TLS config.

aoby:
This is a follow-up to https://github.com/uber/cadence/pull/2769

I didn't test that earlier PR with host verification enabled. Now when I deploy the [0.11.0 release](https://github.com/uber/cadence/releases/tag/v0.11.0) with host verification enabled, it fails with:

```
Incompatible cassandra versions: unable to create CQL Client: gocql: unable to create session: control: unable to connect to initial hosts: x509: cannot validate certificate for <ip address> because it doesn't contain any IP SANs
```

This PR adds a server name setting, which is passed into `tls.Config.ServerName` to bypass the error. I have verified this connects successfully with both `enableHostVerification: true` and `serverName: ` set.


aoby:
Thx for the review @meiliang86. Can we get this merged soon? Do I need anything else?

yycptt:
@aoby Sorry for the delay. I just merged the pull request.

Issue #2874, opened 2019-11-27
Race condition in server start up logic


yycptt:
Currently all services uses the same dataStore map 

When service starts up, it will modify the maxQPS field of the dataStore and use the modified value to create a persistence factory. This introduces a race condition on the maxQPS field.

To fix it, we need to make a deep copy of persistence config for each service so that services won't affect each other.

luowenking1:
The issue was addressed by the pr https://github.com/uber/cadence/pull/3062

PR #3068, opened 2020-02-21
Race condition in server start up logic #2874

luowenking1:
This pr moves the MaxQPS to service level so that the race condition could be removed. The problem I observed here is that we lose the flexibility of setting max qps value to a specific datastore. However, look like so far always the default datastore will be set up with this value. Please let me know your idea about it.


common/service/config/config.go
```diff
@@ -210,7 +210,7 @@ type (
 		ConnectProtocol string `yaml:"connectProtocol" validate:"nonzero"`
 		// ConnectAttributes is a set of key-value attributes to be sent as part of connect data_source_name url
 		ConnectAttributes map[string]string `yaml:"connectAttributes"`
-		// MaxQPS the max request rate on this datastore
+		// MaxQPS is the max request rate to this datastore
```
andrewjdawson2016: I think you need to rebase - current master this is static config. 

If you are up for fixing the bug with double creation of resource bean on startup we can include the diff to make this dynamic config in this diff.

common/service/config/config.go
```diff
@@ -47,8 +47,6 @@ type (
 	Config struct {
 		// Ringpop is the ringpop related configuration
 		Ringpop Ringpop `yaml:"ringpop"`
-		// Persistence contains the configuration for cadence datastores
-		Persistence Persistence `yaml:"persistence"`
```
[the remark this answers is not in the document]
andrewjdawson2016: If people feel there is a good reason to have per service persistence config or if we can implement this in such a way that if I update config I don't need to touch 4 places each time then I am game to move this whole struct to the service level.

common/service/config/persistence.go
```diff
@@ -34,16 +34,8 @@ const (
...
-	if !ok {
-		return
-	}
-	if ds.Cassandra != nil {
-		ds.Cassandra.MaxQPS = qps
-		return
-	}
-	ds.SQL.MaxQPS = qps
+func (c *Persistence) SetMaxQPS(service Service, qps dynamicconfig.IntPropertyFn) {
+	service.MaxQPS = qps
```
andrewjdawson2016: We don't need this function anymore because its just one line - more clear to just set it per service.

common/service/config/config.go
```diff
@@ -78,6 +78,8 @@ type (
 		Metrics Metrics `yaml:"metrics"`
 		// PProf is the PProf configuration
 		PProf PProf `yaml:"pprof"`
+		// MaxQPS is the max request rate to this datastore
+		MaxQPS dynamicconfig.IntPropertyFn `yaml:"-" json:"-"`
```
andrewjdawson2016: Lets call this MaxPersistenceQPS

common/persistence/client/factory.go
```diff
@@ -323,15 +324,15 @@ func (f *factoryImpl) init(clusterName string, limiters map[string]quotas.Limite
...
 		if ds.Cassandra != nil {
-			qps = ds.Cassandra.MaxQPS
+			qps = service.MaxQPS
 		}
 		if ds.SQL != nil {
-			qps = ds.SQL.MaxQPS
+			qps = service.MaxQPS
 		}
 		if qps != nil {
 			result[dsName] = quotas.NewDynamicRateLimiter(func() float64 { return float64(qps()) })
```
andrewjdawson2016: You have to rebase here.

common/service/service.go
```diff
@@ -79,6 +79,8 @@ type (
 		ArchivalMetadata         archiver.ArchivalMetadata
 		ArchiverProvider         provider.ArchiverProvider
 		Authorizer               authorization.Authorizer
+		MaxQPS                   dynamicconfig.IntPropertyFn
```
andrewjdawson2016: Why are both of these added here? You added MaxQPS to Service config already?

cmd/server/cadence/server.go
```diff
@@ -119,6 +119,7 @@ func (s *server) startService() common.Daemon {
 	dc := dynamicconfig.NewCollection(params.DynamicConfig, params.Logger)
 
 	svcCfg := s.cfg.Services[s.name]
+	params.Service = svcCfg
```
andrewjdawson2016: I find this confusing. I believe what is happening here is in this file we set svcConfig. Then in each service (frontend, history, matching, worker). We update a field of this service config (maxQPS). Then we access this reference when we initialize the bean when we initialize the persistence factory? 

Another option is don't add anything to bootstrap params and just pass the dynamic config value to the bean creation when each service bootsup.

PR #3070, opened 2020-02-22
add missing tpath tag

vitarb:


yycptt:
I believe the code using this tag was already deleted in our internal release branch so I don't think we need to add this tag.

PR #3073, opened 2020-02-23
readme: Update cd for docker-compose up.

jaykrell:


docker/README.md
```diff
@@ -8,7 +8,7 @@ along with all its dependencies (cassandra, statsd, graphite). Exposes cadence
 frontend on port 7933 and grafana metrics frontend on port 8080.
 
 ```
-cd $GOPATH/src/github.com/uber/cadence/docker
+cd $GOPATH/src/go.uber.org/cadence/docker/buildkite
```
yux0: What is the reason to have this change?

jaykrell: Isn’t this a correction?
It seemed like it to me.

jaykrell: I mean, I thought I followed the directions & the first path didn’t exist or had no dockerfile. The second seems to be what was intended?

yycptt: @jaykrell I am sorry for the confusion. The path you specified is the docker compose file used by our go-client test. That file is in a different repo and it's for testing purpose, so I don't think we should use that. 

We do have a `docker` folder under the repo root and that folder also has a `docker-compose` file. That's the one we should use. Previously the path is not correct because after we moved to Gopkg the repo doesn't need to be under gopath, but we forget to change the path here. I think in your case, the repo is also not in the gopath right? Can you update the path to something like `path/to/cadence/docker`? Thank you.

jaykrell: I am new to go, docker, and Cadence.
And git-weak.
I am guessing.
One thing I noticed is “go get” seemed to do more than “git clone” so I assumed more was useful and went with go get.

That might be more to clarify in docs, but unsolvable what to teach vs. what to assume.

Anyway I’ll check again later.

PR #3076, opened 2020-02-23
Gcloud storage visibility query

pjgg:


common/archiver/gcloud/queryParser.go
```diff
@@ -0,0 +1,306 @@
...
+		return nil, err
+	}
+
+	if parsedQuery.workflowType == nil {
+		parsedQuery.workflowType = new(string)
+	}
+
+	if parsedQuery.workflowID == nil {
+		parsedQuery.workflowID = new(string)
+	}
```
yycptt: This looks a little bit weird to me, how do you tell the difference between `workflowType is not specified` and `workflowType should be empty string` (if user has a typo)? I think a better way is move those checks to the `query()` function. Check if the fields are nil or not. If not, create a preCondition based on the value and pass the condition to query client.

Same for the `runID` field below.

common/archiver/gcloud/queryParser.go
```diff
@@ -0,0 +1,306 @@
...
+
+	if (parsedQuery.closeTime != 0 || parsedQuery.startTime != 0) && parsedQuery.searchPrecision == nil {
```
yycptt: add a check for if both start and close time are specified?

yycptt:
Multiple errors failed with the following error. Seems like a config problem for unit testing.
```
--- FAIL: TestVisibilityArchiverSuiteSuite/TestArchive_Fail_InvalidVisibilityURI (0.00s) --
  | visibilityArchiver_test.go:140:
  | Error Trace:	visibilityArchiver_test.go:140
  | Error:      	Received unexpected error:
  | dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

common/archiver/gcloud/util.go
```diff
@@ -27,10 +27,15 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
+
+	"github.com/uber/cadence/common"
```
yycptt: Please group it when the cadence imports below.

yycptt:
There are still two tests failing
```

--- FAIL: TestVisibilityArchiverSuiteSuite/TestArchive_Fail_InvalidVisibilityURI (0.01s) --
  | visibilityArchiver_test.go:139:
  | Error Trace:	visibilityArchiver_test.go:139
  | Error:      	Received unexpected error:
  | dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
  | Test:       	TestVisibilityArchiverSuiteSuite/TestArchive_Fail_InvalidVisibilityURI

--- FAIL: TestVisibilityArchiverSuiteSuite/TestQuery_Fail_InvalidVisibilityURI (0.00s) --
  | visibilityArchiver_test.go:165:
  | Error Trace:	visibilityArchiver_test.go:165
  | Error:      	Received unexpected error:
  | dialing: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
  | Test:       	TestVisibilityArchiverSuiteSuite/TestQuery_Fail_InvalidVisibilityURI
```

common/archiver/gcloud/visibilityArchiver.go
```diff
@@ -0,0 +1,259 @@
...
+		prefix = constructTimeBasedSearchKey(request.domainID, indexKeyCloseTimeout, request.parsedQuery.closeTime, *request.parsedQuery.searchPrecision)
+	}
+	if request.parsedQuery.startTime != 0 {
+		prefix = constructTimeBasedSearchKey(request.domainID, indexKeyStartTimeout, request.parsedQuery.startTime, *request.parsedQuery.searchPrecision)
+	}
+
+	filters := []connector.Precondition{
+		newWorkflowIDPrecondition(hash(common.StringDefault(request.parsedQuery.workflowID))),
+		newRunIDPrecondition(hash(common.StringDefault(request.parsedQuery.runID))),
+		newWorkflowTypeNamePrecondition(hash(common.StringDefault(request.parsedQuery.workflowType))),
```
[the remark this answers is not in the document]
pjgg: The logic is implemented as part of `hash` function. Inbound string is going to be hashed if it is not empty, Otherwise it will return the given string.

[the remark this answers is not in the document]
yycptt: I see. But `StringDefault` is called before hash and convert nil pointer to empty string. So basically your assumption here is that if user specified an empty string using query like `workflowID = ""`, it will treated as this filter is not specified right? Is that your desired behavior? If that's the case, can we mention this behavior in README, as it's probably not want people anticipated.  

My suggestion is separate the case for `workflowID = ""` and not specifying workflowID so that the behavior is easier to understand.

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/fortytw2/leaktest v1.3.0 // indirect
 	github.com/go-sql-driver/mysql v1.4.1
 	github.com/gocql/gocql v0.0.0-20191126110522-1982a06ad6b9
-	github.com/golang/mock v1.3.1
+	github.com/golang/mock v1.4.0
```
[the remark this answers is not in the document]
yycptt: Would you please revert this change if possible?

common/archiver/gcloud/visibilityArchiver.go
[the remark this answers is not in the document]
pjgg: I agree!. The code looks too much simpler in this way! (easy to understand and read).

go.sum
```diff
@@ -88,10 +88,12 @@ github.com/go-stack/stack v1.8.0/go.mod h1:v0f6uXyyMGvRgIKkXu+yp6POWl0qKG85gN/me
 github.com/gocql/gocql v0.0.0-20191126110522-1982a06ad6b9 h1:90F3aCM6RNWVC05yJUGbgxgNoDdXyy4mQ8S893sSEm8=
 github.com/gocql/gocql v0.0.0-20191126110522-1982a06ad6b9/go.mod h1:DL0ekTmBSTdlNF25Orwt/JMzqIq3EJ4MVa/J/uK64OY=
 github.com/gogo/googleapis v0.0.0-20180223154316-0cd9801be74a/go.mod h1:gf4bu3Q80BeJ6H1S1vYPm8/ELATdvryBaNFGgqEef3s=
+github.com/gogo/googleapis v1.2.0/go.mod h1:Njal3psf3qN6dwBtQfUmBZh2ybovJ0tlu3o/AC7HYjU=
```
yycptt: Please run `go mod tidy`. Since there's no changes to `go.mod`, I don't think `go.sum` will change.

common/archiver/gcloud/connector/client.go
```diff
@@ -153,19 +157,52 @@ func (s *storageWrapper) Query(ctx context.Context, URI archiver.URI, fileNamePr
...
+		attrs, err = it.Next()
+		if err == iterator.Done {
+			return resultSet, true, currentPos, nil
+		}
+
+		if completed := isPageCompleted(pageSize, len(resultSet)); completed {
+			return resultSet, completed, currentPos, err
+		}
+
+		valid := len(filters) == 0
```
yycptt: nit: `valid` can be default to true.

PR #3080, opened 2020-02-26
Implement multipart s3 history archiver.

### LINKED ISSUE #3002: Multipart Version of S3 History Archiver
Currently the S3 implementation of history archiver interface will always load the entire workflow history into memory and upload only one file. This approach has two drawbacks for large workflows:
1. Load full history into memory may cause a memory issue.
2. If the any error happens during the upload, the entire history needs to be uploaded again.

A better idea is split workflow history into multiple part, upload them piece by piece and checkpoint the progress.

jontro:
Fixes #3002.
Remove MaxRetries, this was a left over from the old interface


common/archiver/s3store/historyArchiver.go
```diff
@@ -223,38 +267,61 @@ func (h *historyArchiver) Get(
...
 
-	key := constructHistoryKey(URI.Path(), request.DomainID, request.WorkflowID, request.RunID, token.CloseFailoverVersion)
-	encodedHistoryBatches, err := download(ctx, h.s3cli, URI, key)
+	prefix := constructHistoryKeyPrefixWithVersion(URI.Path(), request.DomainID, request.WorkflowID, request.RunID, token.CloseFailoverVersion)
+
+	results, err := h.s3cli.ListObjectsV2WithContext(ctx, &s3.ListObjectsV2Input{
+		Bucket:     aws.String(URI.Hostname()),
+		Prefix:     aws.String(prefix),
+		MaxKeys:    aws.Int64(int64(request.PageSize)), // Upper bound of history events
+		StartAfter: (*token).StartAfterKey,
```
yycptt: `token.StartAfterKey` should be enough?

Just to confirm: the key specified here is exclusive right?

common/archiver/s3store/util.go
```diff
@@ -113,9 +113,16 @@ func bucketExists(ctx context.Context, s3cli s3iface.S3API, URI archiver.URI) er
 }
 
 // Key construction
-func constructHistoryKey(path, domainID, workflowID, runID string, version int64) string {
+func constructHistoryKey(path, domainID, workflowID, runID string, version int64, batchIdx int) string {
+	prefix := constructHistoryKeyPrefixWithVersion(path, domainID, workflowID, runID, version)
+	// Note batchIdx has an upper limit of 200 and we rely on them being sorted alphabetically.
```
yycptt: May I ask what's the reason behind this upper limit?

common/archiver/s3store/historyArchiver.go
```diff
@@ -223,38 +267,61 @@ func (h *historyArchiver) Get(
...
+		if isRetryableError(err) {
+			return nil, &shared.InternalServiceError{Message: err.Error()}
+		}
+		return nil, &shared.BadRequestError{Message: err.Error()}
 	}
-	historyBatches, err := decodeHistoryBatches(encodedHistoryBatches)
-	if err != nil {
-		return nil, &shared.InternalServiceError{Message: err.Error()}
+	if len(results.Contents) == 0 {
+		return nil, &shared.BadRequestError{Message: errNoHistoryBatchesFound}
```
yycptt: Please return `&shared.EntityNotExistError{Message: XXX}` in this case.

common/archiver/s3store/historyArchiver.go
```diff
@@ -223,38 +267,61 @@ func (h *historyArchiver) Get(
...
 		if numOfEvents >= request.PageSize {
+			isTruncated = true
 			break
 		}
+		encodedRecord, err := download(ctx, h.s3cli, URI, *item.Key)
+		if err != nil {
+			if isRetryableError(err) {
+				return nil, &shared.InternalServiceError{Message: err.Error()}
+			}
+			return nil, &shared.BadRequestError{Message: err.Error()}
```
yycptt: At this stage, I don't think the error will be a badReqeustError.

common/archiver/s3store/historyArchiver.go
```diff
@@ -223,38 +267,61 @@ func (h *historyArchiver) Get(
...
 	}
 
-	if numOfBatches < len(historyBatches) {
-		token.NextBatchIdx += numOfBatches
+	if isTruncated {
+		token.StartAfterKey = lastKey
 		nextToken, err := serializeToken(token)
 		if err != nil {
-			return nil, &shared.InternalServiceError{Message: err.Error()}
+			return nil, &shared.BadRequestError{Message: err.Error()}
```
yycptt: InternalServiceError?

common/archiver/s3store/historyArchiver.go
```diff
@@ -168,26 +178,61 @@ func (h *historyArchiver) Archive(
...
+			if isRetryableError(err) {
+				logger.Error(archiver.ArchiveTransientErrorMsg)
+			} else {
+				logger.Error(archiver.ArchiveNonRetriableErrorMsg)
+			}
+			return err
+		}
+		uploadedSize := int64(binary.Size(encodedHistoryPart))
+		scope.AddCounter(metrics.HistoryArchiverTotalUploadSize, uploadedSize)
+		scope.AddCounter(metrics.HistoryArchiverHistorySize, uploadedSize)
```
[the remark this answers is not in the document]
yycptt: It's actually not easy to emit metrics for historySize and total upload size, consider the following two situations:

1. After having uploaded two blobs, the worker crash. Later when worker is restarted, it will start from the third blob, but the history size counter will start from 0. To solve this, the size should also be added to the progress.
2. If you added existence check before uploading, then the size of skipped blob counts to the history size but not the upload size.

Feel free to just leave TODO here without emitting those two metrics if it's too much work.

yycptt:
Following unit test are failing, can you please fix them?

```

--- FAIL: TestVisibilityArchiverSuite/TestArchiveAndQuery (0.00s) --
  | visibilityArchiver_test.go:508:
  | Error Trace:	visibilityArchiver_test.go:508
  | Error:      	"[]" should have 2 item(s), but has 0
  | Test:       	TestVisibilityArchiverSuite/TestArchiveAndQuery

--- FAIL: TestVisibilityArchiverSuite/TestArchiveAndQueryPrecisions (0.01s) --
  | visibilityArchiver_test.go:461:
  | Error Trace:	visibilityArchiver_test.go:461
  | Error:      	"[]" should have 2 item(s), but has 0
  | Test:       	TestVisibilityArchiverSuite/TestArchiveAndQueryPrecisions
  | Messages:   	Iteration %!(EXTRA int=0)

--- FAIL: TestVisibilityArchiverSuite/TestQuery_Success_NoNextPageToken (0.00s) --
  | visibilityArchiver_test.go:305:
  | Error Trace:	visibilityArchiver_test.go:305
  | Error:      	"[]" should have 2 item(s), but has 0
  | Test:       	TestVisibilityArchiverSuite/TestQuery_Success_NoNextPageToken

--- FAIL: TestVisibilityArchiverSuite/TestQuery_Success_SmallPageSize (0.00s) --
  | visibilityArchiver_test.go:328:
  | Error Trace:	visibilityArchiver_test.go:328
  | Error:      	Expected value not to be nil.
  | Test:       	TestVisibilityArchiverSuite/TestQuery_Success_SmallPageSize



```

common/archiver/s3store/util.go
jontro: Sorry, 200 is way too low. I want to make sure the first version is returned first, so I would need to zero pad the number so 10 is sorted after 9. What would be a good upper bound for number of history batches? 9999?

common/archiver/s3store/historyArchiver.go
jontro: The reason I've been using BadRequestErrors here is that the client will retry otherwise. I'm of course open to suggestions on how you would like it to be handled

common/archiver/s3store/historyArchiver.go
jontro: See above comment

common/archiver/s3store/historyArchiver.go
jontro: Yes, the key is exclusive

common/archiver/s3store/historyArchiver.go
jontro: See next comment. I'm not sure if it's the intended behaviour to retry. This specific error seems pretty hard to trigger though

common/archiver/s3store/util.go
yycptt: I understand you rely on the result being sorted when listing and you need a fixed length of the page number. But I guess my question is why do we need listing in the first place. We should be able to construct the entire key given the page number right? Then by keeping the page number in the next page token, we can also do pagination. Maybe I missed some context here?

common/archiver/s3store/historyArchiver.go
yycptt: Yeah, we didn't handle retry very well on our client side. It will retry on internal service error, but not bad request error or entity not exist error. 

But when returning an error here, I think the most important thing is the error should make sense to our user. Returning a badRequestError will make user confuse as their request is totally valid. So I think whether or not the request should be retried should be handled by some other logic at the client side, not here. 

Ideally we should have a nonRetryableInternalServiceError type, but for now, I guess retrying a few time is ok.

common/archiver/s3store/util.go
jontro: Yes, but then we would need to "test" if there is another page by calling exists on the next page item for each iteration of the loop. This would make it a non issue but less efficient

yycptt: I see. So for each page, we need to first call exist and then download? Is it possible to just call download and stop when encounter a not exist error?

Another way might be include some metadata information for each blob. For example, the upload dc, time, whether or not this is the last blob etc. This is what we do for our internal implementation.

The third option is just have a upper limit with 4 or 5 9s. But please document this limitation in the README.

I am ok with any approaches here, your choice.

common/archiver/s3store/historyArchiver.go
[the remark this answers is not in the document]
jontro: Nice, am I doing it right now?

common/archiver/s3store/util.go
jontro: Went with using the metadata info that appears in HistoryBlobHeader, feels like the cleanest solution

common/archiver/s3store/historyArchiver.go
jontro: Thanks, however now after the refactor EnityNotExists could appear here as is confirmed by the TestGet_Fail_KeyNotExist unit test. Otherwise it will return InternalServiceError

jontro:
Fixed the failing unit tests as well

common/archiver/s3store/historyArchiver.go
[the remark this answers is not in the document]
yycptt: 👍 Yes, it's correct!

common/archiver/s3store/historyArchiver.go
```diff
@@ -168,26 +178,61 @@ func (h *historyArchiver) Archive(
...
 
-	if err := upload(ctx, h.s3cli, URI, key, encodedHistoryBatches); err != nil {
-		logger.Error(archiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason(errWriteKey), tag.Error(err))
-		return err
+func loadHistoryIterator(ctx context.Context, request *archiver.ArchiveHistoryRequest, historyManager persistence.HistoryManager, featureCatalog *archiver.ArchiveFeatureCatalog, progress *uploadProgress) (historyIterator archiver.HistoryIterator, err error) {
+	if featureCatalog.ProgressManager != nil {
+		if featureCatalog.ProgressManager.HasProgress(ctx) {
+			err = featureCatalog.ProgressManager.LoadProgress(ctx, progress)
+			if err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
yycptt: Since the function will never return an error, I think we can remove err from the function signature and the logic at the caller side for checking the error.

If we failed to `LoadProgress`, you can just call `NewHistoryIterator` which won't an error.

common/archiver/s3store/util.go
yycptt: Awesome! Thanks for implementing this.

common/archiver/s3store/historyArchiver.go
yycptt: Yeah, if it's entity not exist, then return `shared.EntityNotExistError{}`.

common/archiver/s3store/historyArchiver.go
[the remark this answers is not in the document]
jontro: Great, fixed!

common/archiver/s3store/historyArchiver.go
jontro: Good. I think all comments should be adressed now

common/archiver/s3store/historyArchiver.go
```diff
@@ -168,29 +173,86 @@ func (h *historyArchiver) Archive(
...
+			if err == nil {
+				historyIterator, err := archiver.NewHistoryIteratorFromState(request, historyManager, targetHistoryBlobSize, progress.IteratorState)
+				if err == nil {
+					return historyIterator
+				}
+			}
+			progress.IteratorState = nil
+			progress.BatchIdx = 0
+			progress.historySize = 0
+			progress.uploadedSize = 0
```
yycptt: Those four lines can be remove right?

jontro: Well if NewHistoryIteratorFromState fails we need to start from the beginning. Otherwise BatchIdx might be a different number for instance

jontro: I was also unsure if LoadProgress could potentially modify progress and fail

yycptt: Makes sense, you are totally right. I had some misunderstanding. Sorry.

PR #3082, opened 2020-02-26
Race condition in server start up logic #2874

luowenking1:


common/persistence/client/factory.go
```diff
@@ -23,6 +23,8 @@ package client
 import (
 	"sync"
 
+	"github.com/uber/cadence/common/service/dynamicconfig"
+
```
yycptt: remove the empty line and `make fmt`

common/persistence/client/factory.go
```diff
@@ -125,6 +127,7 @@ var storeTypes = []storeType{
 // The objects returned by this factory enforce ratelimit and maxconns according to
 // given configuration. In addition, all objects will emit metrics automatically
 func NewFactory(
+	persistenceMaxQPS dynamicconfig.IntPropertyFn,
 	cfg *config.Persistence,
```
yycptt: nit: probably move maxQPS after cfg? 

Just my personal preference, feel free to ignore.

common/persistence/client/factory.go
```diff
@@ -322,18 +325,11 @@ func (f *factoryImpl) init(clusterName string, limiters map[string]quotas.Limite
...
-			qps = ds.Cassandra.MaxQPS
-		}
-		if ds.SQL != nil {
-			qps = ds.SQL.MaxQPS
-		}
-		if qps > 0 {
-			result[dsName] = quotas.NewSimpleRateLimiter(qps)
+	for dsName, _ := range cfg.DataStores {
+		if persistenceMaxQPS != nil {
+			result[dsName] = quotas.NewDynamicRateLimiter(func() float64 { return float64(persistenceMaxQPS()) })
```
yycptt: Check if the returned value from dynamic config is 0 before creating the rate limiter?

common/persistence/client/factory.go
andrewjdawson2016: I agree

common/persistence/client/factory.go
andrewjdawson2016: What do you want to do if it is 0? I don't think we have these types of dynamic config checks for values in other places?

common/persistence/client/factory.go
```diff
@@ -322,18 +325,11 @@ func (f *factoryImpl) init(clusterName string, limiters map[string]quotas.Limite
 	f.datastores[storeTypeVisibility] = visibilityDataStore
 }
 
-func buildRatelimiters(cfg *config.Persistence) map[string]quotas.Limiter {
+func buildRatelimiters(cfg *config.Persistence, persistenceMaxQPS dynamicconfig.IntPropertyFn) map[string]quotas.Limiter {
```
andrewjdawson2016: nit: here since you are in the persistence package you can just refer to this as MaxQPS in this context.

common/persistence/client/factory.go
yycptt: Previous code did have a check on if the value is 0 right? I think if it's equal to 0, then no rate limiter should be created.

yycptt: Sorry, please ignore me. It was already checked in the if statement and I missed that part.

PR #3084, opened 2020-02-28
Host Level Queue Task Processor

yycptt:


service/history/queueTaskProcessor.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/task"
+)
+
+type (
+	queueTaskProcessorOptions struct {
+		fifoSchedulerOptions *task.FIFOTaskSchedulerOptions
```
yux0: What about having a type here and use switch to replace those if statement?

service/history/queueTaskProcessor.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return false, err
+	}
+	return scheduler.TrySubmit(task)
+}
+
+func (p *queueTaskProcessorImpl) prepareSubmit(
+	task queueTask,
+) (task.Scheduler, error) {
+	if !p.isRunning() {
```
yux0: As discussed, this check could move to getOrCreateTaskScheduler

service/history/queueTaskProcessor.go
yycptt: Done

service/history/queueTaskProcessor.go
yycptt: Done.

Issue #3087, opened 2020-03-02
PR/commit hygiene concerns/suggestions


antstorm:
Hey guys,

Really appreciate all the effort that is being put in to this project. It feels like Cadence is still super underrated and has enormous future potential.

My company is already using it for some use cases and we're planning on leveraging it much more. One important concern that our Security team has raised is around PR hygiene. It looks like lots of changes are being shipped without any description/context. I do understand that there's probably an internal process in place at Uber that coordinates these changes and there's probably plenty of discussions to support them, but I think exposing some of the context publicly would be very beneficial to Cadence users.

The use cases for this involve assessing whether upgrade is needed, figuring out how much potential risk it brings or finding a production issue related to a version upgrade.

Don't wanna impose too much bureaucracy here, but I think simple PR template would go a long way and will be greatly appreciated by other adopters of Cadence. It will also help open source devs to jump on board quicker and start contributing.

Basically the intention is to understand "What" has changed, "Why" it got changed and "How" it was tested/verified, with some risk scale assessment from the person making the change.

Hope that makes sense and please keep up the good work!

PR #3088, opened 2020-03-02
Add PR template

antstorm:
**What changed? Why?**
Adds a template that will be used by default in all PRs. Following this will increase visibility into the changes that are being added to the project and should make it easier to find a culprit PR if there's an issue or get more context on what is going on. More info in #3087 

**How has it been tested?**
This PR message should prove that it works

**Potential risk level (on a scale from 1 to 5)**
1


antstorm:
For some reason @claassistantio won't verify that I've already signed the agreement ^

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+<!-- Describe the future self what has changed and most importanly why -->
+**What changed? Why?**
```
yux0: Can we separate this to two questions?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+
+
+<!-- How have you verified this change? Tested locally? Added a unit test? Checked in staging env? -->
+**How has it been tested?**
+
+
+<!-- How risky is it to deploy this straight to production? -->
+**Potential risk level (on a scale from 1 to 5)**
```
yux0: Could it be more clear about the scale? 1 is the highest risk or 5 is the highest risk?

yux0: In addition, the number could be arbitrary if there is no criteria about risk level. What about changing the question to: What could be broken (What is the worst case) when deploying this change to production?

yycptt:
> For some reason @claassistantio won't verify that I've already signed the agreement ^

Try close and reopen this pull request. It usually resolves the issue.

PULL_REQUEST_TEMPLATE.md
antstorm: Good idea 👍

PULL_REQUEST_TEMPLATE.md
antstorm: I'll clarify this, thanks for pointing out!

antstorm:
@yux0 I've added suggested changes

antstorm:
@yux0 I've made suggested changes

antstorm:
@yux0 thanks for approving, let me close it and re-open a new one to get through CLA

PR #3089, opened 2020-03-02
Add detail messages for retry error v2

yux0:
Add detail messages to retry error v2 for debugging.


service/history/nDCBranchMgr.go
```diff
@@ -204,6 +204,7 @@ func (r *nDCBranchMgrImpl) verifyEventsOrder(
 	if incomingFirstEventID > nextEventID {
 		executionInfo := r.mutableState.GetExecutionInfo()
 		return false, newNDCRetryTaskErrorWithHint(
+			"Resend events due to out of order delivery",
```
yycptt: nit: Also create constants for the string? Same for others.

yux0: Updated

PR #3102, opened 2020-03-09
matching_scalable_tasklist: bug fixes

venkat1109:
This patch fixes multiple problems with the scalable task list implementation:

**Problem#1:**  Tasks can be persisted twice, once at the child partition and once at the root partition
Currently, the parent partition is missing the necessary check to not persist a task when it's forwarded from a child partition (for remote sync match). This leads to tasks being persisted and processed twice.

Solution: Added check with AddTask() to not persist the task when forwardedFrom is true

**Problem#2:** ACK levels can get stuck for child partitions. Following is the scenario that leads to this problem:
  - Child partition receives a task, there is no poller and so persists the task to db
  - Child partition taskReader reads backlog from db and attempts to match against a poller
  - Child partition forwards the task to parent for remote sync match
  - Task is remote sync matched and child partition receives success
  - Child partition fails to delete the task from db / mark the task as completed

Solution: ACK the task even in the case of remote sync match

**Problem#3:** The impact of this problem could be increased task dispatch latency when there is backlog and insufficient number of pollers. Following is the scenario:

  - Child partition taskReader reads backlog from db and attempts to match against a poller
  - Child partition forwards the task to parent for remote sync match after local sync match fails
  - The call to the parent has a context timeout of 2s with the intention that the call will block on the parent partition for up to 2s trying to remote sync match. 
 - Parent partition overrides the context timeout from 2s to 200ms because sync matches typically are non-blocking.
- Child partition spends the remained 1.5s trying to do local sync match before it makes another attempt at remote sync match.

Solution: Add logic within the parent partition to allow sync match to block for this scenario.

### DIFF
--- a/service/matching/matcher.go
+++ b/service/matching/matcher.go
@@ -158,8 +158,12 @@ func (tm *TaskMatcher) offerOrTimeout(ctx context.Context, task *internalTask) (
 	select {
 	case tm.taskC <- task: // poller picked up the task
 		if task.responseC != nil {
-			err := <-task.responseC
-			return true, err
+			select {
+			case err := <-task.responseC:
+				return true, err
+			case <-ctx.Done():
+				return false, nil
+			}
 		}
 		return false, nil
 	case <-ctx.Done():
--- a/service/matching/matcher.go
+++ b/service/matching/matcher.go
@@ -25,10 +25,10 @@ import (
 	"errors"
 	"time"
 
-	"github.com/uber/cadence/.gen/go/shared"
-
 	"golang.org/x/time/rate"
 
+	"github.com/uber/cadence/.gen/go/matching"
+	"github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/quotas"
 )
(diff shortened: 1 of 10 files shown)


service/matching/matcher.go
```diff
@@ -79,9 +79,27 @@ func newTaskMatcher(config *taskListConfig, fwdr *Forwarder, scopeFunc func() me
...
+// task. This method should ONLY be used for sync match.
+//
+// When a local poller is not available and forwarding to a parent
+// task list partition is possible, this method will attempt forwarding
+// to the parent partition.
+//
+// Cases when this method will block:
+//
+// Ratelimit:
+// When a ratelimit is token is not available, this method might block
```
vancexu: typo "is token"

venkat1109: fixed

service/matching/matcher.go
```diff
@@ -129,6 +154,19 @@ func (tm *TaskMatcher) Offer(ctx context.Context, task *internalTask) (bool, err
 	}
 }
 
+func (tm *TaskMatcher) offerOrTimeout(ctx context.Context, task *internalTask) (bool, error) {
+	select {
+	case tm.taskC <- task: // poller picked up the task
+		if task.responseC != nil {
+			err := <-task.responseC
```
yux0: Do you need to honor the ctx deadline while waiting on response?

venkat1109: Good question. From a behavior standpoint, if it takes too long for the response, the caller sees timeout anyway (irrespective of whether we return from here or not). This can happen when RecordDecisionTaskStarted() simply times out and is in a retry loop.  The only reason to do this would be for system protection i.e. when history is down and too may goroutines wait here for the response (for the duration of the retry/backoff expiration).  Since it doesn't change behavior for the caller (one who calls Offer()), I think it's worth honoring the deadline.

PR #3105, opened 2020-03-10
Update switch statement to prevent unsupported task type

yux0:
Update switch statement to prevent unsupported task type or newly added type

**What changed?**
Update Kafka producer get key for replication task to handle default case.

**Why?**
To prevent newly added replication type not added in the kafka producer

**How did you test it?**
Locally

**Potential risks**
This is to prevent coding issues. This should not impact production.

### DIFF
--- a/common/messaging/kafkaProducer.go
+++ b/common/messaging/kafkaProducer.go
@@ -113,6 +113,12 @@ func (p *kafkaProducer) getKeyForReplicationTask(task *replicator.ReplicationTas
 		// the messaging layer perspective
 		attributes := task.SyncActivityTaskAttributes
 		return sarama.StringEncoder(attributes.GetWorkflowId())
+	case replicator.ReplicationTaskTypeHistoryMetadata,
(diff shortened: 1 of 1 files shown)


common/messaging/kafkaProducer.go
```diff
@@ -113,6 +113,12 @@ func (p *kafkaProducer) getKeyForReplicationTask(task *replicator.ReplicationTas
 		// the messaging layer perspective
 		attributes := task.SyncActivityTaskAttributes
 		return sarama.StringEncoder(attributes.GetWorkflowId())
+	case replicator.ReplicationTaskTypeHistoryMetadata,
+		replicator.ReplicationTaskTypeDomain,
+		replicator.ReplicationTaskTypeSyncShardStatus:
+		return nil
+	default:
+		panic("encounter unsupported replication task type")
```
venkat1109: - can you also log the task type and any other debug info as part of panic ?

yux0: Added the task type in the log

PR #3108, opened 2020-03-11
Add Activity E2E latency metrics

vancexu:
**What changed?**
Add new metric `activity_end_to_end_latency` for activity 

**Why?**
Currently, there is no cadence metrics on activity e2e latency available for customer. The cadence client (both go and java) had activity execution and e2e latency metrics, but they are not correct for async activity completion (and no easy way to get it correctly on client side). 

**How did you test it?**
Will be tested in staging before release.

**Potential risks**
No risk. 
Only adding load on metrics dependency.

### DIFF
--- a/common/taskTokenSerializerInterfaces.go
+++ b/common/taskTokenSerializerInterfaces.go
@@ -33,10 +33,12 @@ type (
 	TaskToken struct {
 		DomainID        string `json:"domainId"`
 		WorkflowID      string `json:"workflowId"`
+		WorkflowType    string `json:"workflowType"`
 		RunID           string `json:"runId"`
 		ScheduleID      int64  `json:"scheduleId"`
 		ScheduleAttempt int64  `json:"scheduleAttempt"`
 		ActivityID      string `json:"activityId"`
+		ActivityType    string `json:"activityType"`
 	}
 
 	// QueryTaskToken identifies a query task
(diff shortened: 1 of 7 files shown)


service/history/historyEngine.go
```diff
@@ -1492,8 +1493,13 @@ func (e *historyEngineImpl) RespondActivityTaskCompleted(
 				// Unable to add ActivityTaskCompleted event to history
 				return &workflow.InternalServiceError{Message: "Unable to add ActivityTaskCompleted event to history."}
 			}
+			activityStartedTime = ai.StartedTime
 			return nil
 		})
+	if err == nil && !activityStartedTime.IsZero() {
+		e.metricsClient.RecordTimer(metrics.HistoryRespondActivityTaskCompletedScope, metrics.ActivityE2ELatency, time.Since(activityStartedTime))
```
vancexu: Pending review, let me try also add activity and workflow type

meiliang86: and task list.

vancexu: resolved

common/taskTokenSerializerInterfaces.go
```diff
@@ -33,10 +33,13 @@ type (
 	TaskToken struct {
 		DomainID        string `json:"domainId"`
 		WorkflowID      string `json:"workflowId"`
+		WorkflowType    string `json:"workflowType"`
 		RunID           string `json:"runId"`
 		ScheduleID      int64  `json:"scheduleId"`
 		ScheduleAttempt int64  `json:"scheduleAttempt"`
 		ActivityID      string `json:"activityId"`
+		ActivityType    string `json:"activityType"`
+		TaskList        string `json:"taskList"`
```
meiliang86: ActivityInfo already has task list. So you don't need to add it here.

vancexu: Make sense, fixed.

PR #3109, opened 2020-03-12
Support for abstract datastore in persistence tests

vitarb:
This change:

- Adds AbstractDatastoreFactory to the TestBase
- Makes logger public so it can be set outside of the package
- Adds/renames couple fields in the CustomDatastoreConfig

### DIFF
--- a/common/service/config/config.go
+++ b/common/service/config/config.go
@@ -225,8 +225,8 @@ type (
 	CustomDatastoreConfig struct {
 		// Name of the custom datastore
 		Name string `yaml:"name"`
-		// ConnectAttributes is a set of key-value attributes that can be used by AbstractDatastoreFactory implementation
-		ConnectAttributes map[string]string `yaml:"connectAttributes"`
+		// Options is a set of key-value attributes that can be used by AbstractDatastoreFactory implementation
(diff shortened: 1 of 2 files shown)


common/service/config/config.go
```diff
@@ -225,8 +225,12 @@ type (
 	CustomDatastoreConfig struct {
 		// Name of the custom datastore
 		Name string `yaml:"name"`
-		// ConnectAttributes is a set of key-value attributes that can be used by AbstractDatastoreFactory implementation
-		ConnectAttributes map[string]string `yaml:"connectAttributes"`
+		// ApplicationID describes the name of the application using this datastore. (e.g. cadence)
+		ApplicationID string `yaml:"applicationId"`
+		// Instance is the name of the datastore instance being used.
+		Instance string `yaml:"instance"`
```
yycptt: What's the difference between `Name` and `Instance`?

vitarb: Name is the name of the configuration, while instance is an actual instance that connection will be made to.

In case of docstore we'll have something like:
```
  datastores:
    docstore-default:
      customDatastore:
        name: "docstore"
        applicationId: "cadence"
        instance: "minimez"
```
in the test config.

PR #3111, opened 2020-03-12
Get raw history compatibility

luowenking1:


service/frontend/workflowHandler.go
```diff
@@ -2176,22 +1827,38 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+					nextEventID,
+					getRequest.GetMaximumPageSize(),
+					nil,
+					token.TransientDecision,
+					token.BranchToken,
+				)
+			}
 			if err != nil {
 				return nil, wh.error(err, scope)
 			}
```
anish531213: Also, check line 1866, `history.Events = history.Events[len(history.Events)-1 : len(history.Events)]`. This won't be valid. You need to handle this for both case.

luowenking1: > Also, check line 1866, `history.Events = history.Events[len(history.Events)-1 : len(history.Events)]`. This won't be valid. You need to handle this for both case.

Done

common/service/dynamicconfig/constants.go
```diff
@@ -93,6 +93,8 @@ var keys = map[Key]string{
 	DisableListVisibilityByFilter:         "frontend.disableListVisibilityByFilter",
 	FrontendThrottledLogRPS:               "frontend.throttledLogRPS",
 	EnableClientVersionCheck:              "frontend.enableClientVersionCheck",
+	EnableRawHistoryQuery:                 "frontend.enableRawHistoryQuery",
+	EnableRawHistoryQueryByDomain:         "frontend.enableRawHistoryQueryByDomain",
```
venkat1109: nit: suffixing with "Query" can be confusing for the reader because Query has other meaning within workflows in cadence. Can we call this something else - one suggestion would be frontend.sendRawWorkflowHistory

common/service/dynamicconfig/constants.go
```diff
@@ -339,6 +341,11 @@ const (
 	FrontendThrottledLogRPS
 	// EnableClientVersionCheck enables client version check for frontend
 	EnableClientVersionCheck
+	// EnableRawHistoryQuery is whether to enable raw history retrieving
+	EnableRawHistoryQuery
+	// EnableRawHistoryQuery is whether to enable raw history retrieving by domain
+	EnableRawHistoryQueryByDomain
```
venkat1109: Why do we need two config knobs ? If you create one byDomain you can always use a default value that applies to all domains.

service/frontend/workflowHandler.go
```diff
@@ -2176,27 +1827,47 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		token.IsWorkflowRunning = isWorkflowRunning
 		token.PersistenceToken = nil
 	}
+	rawHistoryQueryEnabled := wh.config.EnableRawHistoryQuery() && wh.config.EnableRawHistoryQueryByDomain(getRequest.GetDomain())
```
venkat1109: see comment above - just ByDomain config is sufficient

service/frontend/workflowHandler.go
```diff
@@ -2176,27 +1827,47 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
 
 	history := &gen.History{}
+	historyBlob := []*gen.DataBlob{}
```
venkat1109: if both are non-nil, how does the client know which history is valid ?

service/frontend/workflowHandler.go
```diff
@@ -2213,17 +1884,31 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+					domainID,
+					*execution,
+					token.FirstEventID,
+					token.NextEventID,
+					getRequest.GetMaximumPageSize(),
+					token.PersistenceToken,
+					token.TransientDecision,
+					token.BranchToken,
+				)
+			}
```
venkat1109: consider creating a helper method that takes the enabledFlag as param and returns 3 results - history, rawHistory, error.

service/frontend/workflowHandler.go
```diff
@@ -1686,355 +1686,6 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
...
-		return nil, wh.error(err, scope)
-	}
-	return &gen.GetWorkflowExecutionRawHistoryResponse{
-		RawHistory:    history,
-		NextPageToken: nextToken,
-	}, nil
-}
-
-// PollForWorkflowExecutionRawHistory get Raw History for long poll request
-func (wh *WorkflowHandler) PollForWorkflowExecutionRawHistory(
```
venkat1109: why was there a PollForWorkflowExecutionRawHistory but no corresponding PollForWorkflowExecutionHistory ? Why did we create this API ?

service/frontend/workflowHandler.go
```diff
@@ -2176,27 +1827,47 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
-				nextEventID,
-				getRequest.GetMaximumPageSize(),
-				nil,
-				token.TransientDecision,
-				token.BranchToken,
-			)
-			if err != nil {
-				return nil, wh.error(err, scope)
+			if rawHistoryQueryEnabled {
+				historyBlob, _, err = wh.getRawHistory(
```
venkat1109: I didn't see the tests being updated to exercise this new code path. Can you port the rawHistory tests (which are deleted) to exercise this code instead ?

service/frontend/workflowHandler.go
luowenking1: > consider creating a helper method that takes the enabledFlag as param and returns 3 results - history, rawHistory, error.

Seems not only this flag as argument, also need to take token and getRequest as well.

common/service/dynamicconfig/constants.go
luowenking1: > nit: suffixing with "Query" can be confusing for the reader because Query has other meaning within workflows in cadence. Can we call this something else - one suggestion would be frontend.sendRawWorkflowHistory

done

common/service/dynamicconfig/constants.go
luowenking1: > Why do we need two config knobs ? If you create one byDomain you can always use a default value that applies to all domains.

I saw consistent query is set up with two knobs for some reason. Just made it work in the same way. Anyway, I have removed the extra one.

service/frontend/workflowHandler.go
luowenking1: > see comment above - just ByDomain config is sufficient

done

service/frontend/workflowHandler.go
luowenking1: > if both are non-nil, how does the client know which history is valid ?

fixed

service/frontend/workflowHandler.go
luowenking1: > why was there a PollForWorkflowExecutionRawHistory but no corresponding PollForWorkflowExecutionHistory ? Why did we create this API ?

PollForWorkflowExecutionRawHistory was added by Anish for long pull of raw history purposed. Since we put all logic in one common API, we do not need it any more. We do not have PollForWorkflowExecutionHistory, the long pull implementation also resides in the GetWorkflowExecutionHistory API.

service/frontend/workflowHandler.go
luowenking1: > I didn't see the tests being updated to exercise this new code path. Can you port the rawHistory tests (which are deleted) to exercise this code instead ?

We did not have any unit test for GetRawHistoryExecution API on server side. Now I am trying to add integration test but still not find a way to turn on/off the knobs. Yesterday you mentioned we can add something to RPC header for that purpose, however, I am still investigating how to make it.

venkat1109: the rpc header idea was for the client side integration test. For the server side integration test, we can tweak dynamic config to enable raw history for the test domain.

service/frontend/workflowHandler.go
```diff
@@ -2176,27 +1827,47 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+					domainID,
+					*execution,
+					lastFirstEventID,
+					nextEventID,
+					getRequest.GetMaximumPageSize(),
+					nil,
+					token.TransientDecision,
+					token.BranchToken,
+				)
+				if err != nil {
```
anish531213: This is for close event only, so you need to return only the last blob similar to line 1869.

service/frontend/workflowHandler.go
anish531213: @luowenking1 you can use the previous integration test that I wrote for GetWorkflowRawHistory and PollForWorkflowExecutionHistory.

service/frontend/workflowHandler.go
anish531213: After an initial conversation with @yycptt @yux0, we wanted to separate GetWorkflowExecution API into two API, long pull and short pull, to make the code simpler and also to have different metrics for long pull and short pull.

service/frontend/workflowHandler.go
luowenking1: > @luowenking1 you can reuse the previous integration test that I wrote for GetWorkflowRawHistory and PollForWorkflowExecutionHistory.

Yes I have added the tests in the gethistory_test.go file.

common/service/dynamicconfig/constants.go
```diff
@@ -94,6 +94,7 @@ var keys = map[Key]string{
 	FrontendThrottledLogRPS:               "frontend.throttledLogRPS",
 	EnableClientVersionCheck:              "frontend.enableClientVersionCheck",
 	ValidSearchAttributes:                 "frontend.validSearchAttributes",
+	SendRawWorkflowHistoryByDomain:        "frontend.sendRawWorkflowHistoryByDomain",
```
venkat1109: please remove the ByDomain suffix just to be consistent with rest of the config i.e. we don't follow the pattern of suffixing with filter criteria

host/integrationbase.go
```diff
@@ -94,11 +95,12 @@ func (s *IntegrationBase) setupSuite(defaultClusterConfigFile string) {
 		s.engine = s.testCluster.GetFrontendClient()
 		s.adminClient = s.testCluster.GetAdminClient()
 	}
-
+	s.testRawHistoryDomainName = "TestRawHistoryDomain"
```
venkat1109: where is this domain created ? I don't see it

service/frontend/workflowHandler.go
```diff
@@ -2176,27 +1827,50 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
...
+					return nil, wh.error(err, scope)
+				}
+				// since getHistory func will not return empty history, so the below is safe
+				history.Events = history.Events[len(history.Events)-1 : len(history.Events)]
 			}
-			// since getHistory func will not return empty history, so the below is safe
-			history.Events = history.Events[len(history.Events)-1 : len(history.Events)]
 			token = nil
 		} else if isLongPoll {
```
venkat1109: previously, the else if and else code paths had history and history.Events non-nil - with your code change, they will be nil. Can you fix this to not change existing behavior ?

PR #3114, opened 2020-03-13
Add filter to not reapply event with the same version

yux0:
**What changed?**
During NDC event reapplication, add a filter to not apply event with the same version as the current workflow last write version. 

**Why?**
If the reapply event with the same version, it means the event is from the same cluster and no failover happened during this time. It means the event is already applied in the current cluster and there is no need to reapply this event.

**How did you test it?**
Reapply events with version A to the current workflow with version A. The events should not show in history.

**Potential risks**
In the worst case, the event will not reapply to the current branch.

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2904,8 +2904,13 @@ func (e *historyEngineImpl) ReapplyEvents(
 				return nil, err
 			}
 			for _, event := range reapplyEvents {
+				if event.GetVersion() == lastWriteVersion {
+					// The reapply is from the same cluster. Ignoring.
+					continue
+				}
 				dedupResource := definition.NewEventReappliedID(runID, event.GetEventId(), event.GetVersion())
-				if event.GetVersion() == lastWriteVersion || mutableState.IsResourceDuplicated(dedupResource) {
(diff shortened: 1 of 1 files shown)


wxing1292:
LGTM

service/history/historyEngine.go
```diff
@@ -2897,6 +2897,24 @@ func (e *historyEngineImpl) ReapplyEvents(
...
+			toReapplyEvents := make([]*workflow.HistoryEvent, len(reapplyEvents))
+			lastWriteVersion, err := mutableState.GetLastWriteVersion()
+			if err != nil {
+				return nil, err
+			}
+			for _, event := range reapplyEvents {
+				dedupResource := definition.NewEventReappliedID(runID, event.GetEventId(), event.GetVersion())
+				if event.GetVersion() == lastWriteVersion || mutableState.IsResourceDuplicated(dedupResource) {
+					continue
+				}
```
wxing1292:
```go
if event.GetVersion() == lastWriteVersion {
    // events to be reapplied is from main event timeline, should not reapply
    continue
} else if mutableState.IsResourceDuplicated(dedupResource) {
    // already re-apply the event
    continue
 }
```

yux0: Done

Issue #1873, opened 2019-05-17
Record more started time/errors for retried activities


longquanzheng:
Right now we only record the last started time when writing down the started event. This is super confusing for customers even for us.

We don't have to record all of them, but it will be nice to have a limit like at most 5 of them can be recorded down.

dizUber:
During using the Cadence, we found one activity will fail to schedule on all the machines, the last time it fails by timeout. But the failure records in the process will help debugging the issue.

martinxsliu:
Hi all, is there any progress on this issue? We're looking at adopting Cadence, in part as a potential replacement for our existing background workers. Currently, if an activity fails and is retried, there is no way to see what the error message is that caused the failure which makes it difficult to debug. This feature is something that comes out of the box with traditional job workers like Sidekiq or Celery.

PR #3116, opened 2020-03-13
Record error info for retried activities (#1873)

emrahs:
Exposes the lastFailureReason and lastFailureDetails, which exist in the mutable state, through
the ActivityTaskStartedEvent in the history. The fields were added to ActivityTaskStartedEvent
as part of https://github.com/uber/cadence-idl/pull/15

Resolves: #1873


service/history/historyBuilder.go
```diff
@@ -591,14 +598,22 @@ func (b *historyBuilder) newActivityTaskScheduledEvent(decisionTaskCompletedEven
...
+	lastFailureDetails []byte,
+) *workflow.HistoryEvent {
 	historyEvent := b.msBuilder.CreateNewHistoryEvent(workflow.EventTypeActivityTaskStarted)
 	attributes := &workflow.ActivityTaskStartedEventAttributes{}
 	attributes.ScheduledEventId = common.Int64Ptr(scheduledEventID)
 	attributes.Attempt = common.Int32Ptr(attempt)
 	attributes.Identity = common.StringPtr(identity)
 	attributes.RequestId = common.StringPtr(requestID)
+	attributes.LastFailureReason = common.StringPtr(lastFailureReason)
+	attributes.LastFailureDetails = lastFailureDetails
```
venkat1109: is it possible to update any existing test to validate that these values are now set when activity fails ?

meiliang86: +1

PR #3117, opened 2020-03-16
Fix archival handled requests not match pumped requests

yycptt:
**What changed?**
Use json instead of gob when hashing archival requests

**Why?**
gob is non-deterministic when encoding map objects

**How did you test it?**
Unit test

**Potential risks**
No, this fixes a sanity check itself.

### DIFF
--- a/service/worker/archiver/util.go
+++ b/service/worker/archiver/util.go
@@ -22,7 +22,7 @@ package archiver
 
 import (
 	"bytes"
-	"encoding/gob"
+	"encoding/json"
 	"time"
 
 	"github.com/dgryski/go-farm"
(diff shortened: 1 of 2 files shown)


vancexu:
It looks like gob non-deterministic encoding issue was fixed at 2012 https://github.com/golang/go/issues/3026, If there is still such issue we probably can submit an issue to them

yycptt:
> It looks like gob non-deterministic encoding issue was fixed at 2012 [golang/go#3026](https://github.com/golang/go/issues/3026), If there is still such issue we probably can submit an issue to them

I am aware of the issue you linked. However if I understand that issue correctly, it's talking about decoding not encoding. The encoding for map is still non-deterministic. https://stackoverflow.com/a/50733233

PR #3121, opened 2020-03-18
Change error type on query before first decision task

andrewjdawson2016:
Change error type on query before first decision task

Because badRequest is not an accurate error. 

No

None

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -183,7 +183,7 @@ var (
 	// ErrQueryEnteredInvalidState is error indicating query entered invalid state
 	ErrQueryEnteredInvalidState = &workflow.BadRequestError{Message: "query entered invalid state, this should be impossible"}
 	// ErrQueryWorkflowBeforeFirstDecision is error indicating that query was attempted before first decision task completed
-	ErrQueryWorkflowBeforeFirstDecision = errors.New("workflow must handle at least one decision task before it can be queried")
+	ErrQueryWorkflowBeforeFirstDecision = &workflow.QueryFailedError{Message: "workflow must handle at least one decision task before it can be queried"}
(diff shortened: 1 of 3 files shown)


service/history/historyEngine.go
```diff
@@ -183,7 +183,7 @@ var (
 	// ErrQueryEnteredInvalidState is error indicating query entered invalid state
 	ErrQueryEnteredInvalidState = &workflow.BadRequestError{Message: "query entered invalid state, this should be impossible"}
 	// ErrQueryWorkflowBeforeFirstDecision is error indicating that query was attempted before first decision task completed
-	ErrQueryWorkflowBeforeFirstDecision = &workflow.BadRequestError{Message: "workflow must handle at least one decision task before it can be queried"}
+	ErrQueryWorkflowBeforeFirstDecision = errors.New("workflow must handle at least one decision task before it can be queried")
```
meiliang86: How about QueryFailedError?

andrewjdawson2016: We can make it queryFailedError but there is a work item here in that we over use this error type. We need an error type that indicates the users code is the reason the query failed. I was planning to use queryFailedError to indicate this. But if not we should consider introducing a new error type like "QueryWorkflowFailedError" (the naming here is hard).

andrewjdawson2016: Right now just by looking at the error type its hard to determine if the queryFailure was because of users code or not.

meiliang86: I'm fine with creating a new error type. The concern here is that user is not able to assert on this general error and do backoff/retry.

andrewjdawson2016: Sounds good I will create a new error type.

yux0: Maybe this should not belong here. But we should create two generic types of errors. Retryable and NonRetryable.

PR #3124, opened 2020-03-20
Task redispatch queue

yycptt:
**What changed?**
Add a re-dispatch queue to transfer/timer queue processor

**Why?**
- When using the new priority task processor, it's possible that there are lots of tasks with low priority and the channel for that priority becomes full. Instead of blocking tasks with other priority, those tasks should be buffered elsewhere, so that it can be re-dispatched later. 
- If the processing logic for a task keeps failing for a long time, instead of infinity retry, that task should also be put in to the re-dispatch queue and be retried later.

**How did you test it?**
Unit test

**Potential risks**
No risk right now as the implementation for priority task processor has not been wired up yet.


common/collection/concurrentQueue.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (q *concurrentQueueImpl) Peek() interface{} {
+	q.Lock()
+	defer q.Unlock()
+
+	if q.isEmptyLocked() {
+		panic("Cannot peek items because queue is empty")
+	}
```
wxing1292: this should return nil, as special value

common/collection/concurrentQueue.go
```diff
@@ -0,0 +1,87 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package collection
+
+import (
+	"sync"
+)
+
+type (
```
wxing1292: out of curiosity, why not use a chan or a wrapper on top of chan?

yycptt: The requirement for the implementation is that it should potentially have infinite size instead of a fixed size, otherwise the Add() method will be blocking. And if Add() is blocking, then the following situation will occur:
- Assume there is only one worker goroutine and the queue is currently full. The worker goroutine executed a task and decides to redispatch this task. It calls Add(), but since the queue is full, it got blocked. Now basically there's no goroutine for processing tasks and result in a deadlock.
- If the queue has a fixed size and tasks from one domain occupies the entire queue. This blocks tasks from other domains (with a high priority) being processed and defeat the purpose of multi-tenant task processing. 
Right now the assumption here is that host memory is large enough to hold all the tasks and we can use an infinite size queue as the re-dispatch queue. Of course this is not true and this problem will be addressed by the task queue split & merge work (basically offload some tasks under high memory pressure).

Yes there are ways around the situation I mentioned above, for example, use a new goroutine from adding tasks to the re-dispatch queue and probably having one re-dispatch queue for each task priority. But right now those are not necessary and using a list solves all the problems. This, combined with the later task queue split & merge work, should not result in any memory issue.

common/collection/interface.go
```diff
@@ -28,6 +28,8 @@ type (
 		// Add push an item to the queue
 		Add(item interface{})
 		// Remove pop an item from the queue
+		// TODO: consider return an error here to indicate no more item in the Queue
```
yux0: If you add a nil check when adding new items, then remove can also return nil

service/history/queueProcessor.go
```diff
@@ -104,14 +115,16 @@ func newQueueProcessorBase(
...
+		status:             common.DaemonStatusInitialized,
+		notifyCh:           make(chan struct{}, 1),
+		shutdownCh:         make(chan struct{}),
+		metricsClient:      shard.GetMetricsClient(),
+		logger:             logger,
+		ackMgr:             queueAckMgr,
+		lastPollTime:       time.Time{},
+		taskProcessor:      taskProcessor,
+		queueTaskProcessor: queueTaskProcessor,
+		redispatchQueue:    collection.NewConcurrentQueue(),
```
yux0: What about reDispatchQueue?

common/collection/concurrentQueue.go
wxing1292: Add panic on nil is ok.

common/collection/concurrentQueue.go
wxing1292: if the idea is to have unlimited space 
image

then why not using a list? (rather than a slice)

common/collection/concurrentQueue.go
yycptt: Done. Removed panic from `Peek()` and `Remove()`. Added a check in `Add()` and panic if the item being added is `nil`.

common/collection/interface.go
yycptt: Makes sense. Added the check when adding new items. Removed the TODO here.

common/collection/concurrentQueue.go
yycptt: I tried both implementations actually, and benchmark shows slice has better performance. 1083 ns/op when using slice, while 1329 ns/op when using list. Slice also incurs less allocs per op. I think that's because we are storing small object in the queue (`interface{}`) so the overhead for maintaining a linked list is too high.

wxing1292: just keep in mind that using slice may result in more mem consumption.
to me 1000+ vs 1300+ ns is not a huge difference.

yycptt: Yes, I am aware of that fact that doing `items[1:]` will not free the memory for the first item. But it will eventually be garbage collected during slice reallocation, so the memory will not grow unbounded. 

I will monitor the mem usage in prod and if it turns out to be a problem, I will switch to list implementation. If there's no mem issue, then I prefer the one with better performance in terms of time.

PR #3129, opened 2020-03-24
Adding design doc for graceful failover

yux0:
**What changed?**
Design Doc

**Why?**
N/A

**How did you test it?**
N/A

**Potential risks**
N/A


docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,230 @@
...
+The coordinator persists the state in memory and updates this state to database periodically. The state can be stored in the shard table. The state struct looks like:
+map[string][]*int32
+The key contains the domain and the failover version.
+The value is a slice of shard ID.
+Failover timeout
+Currently, each history host has a component domain cache. Each shard on the same host gets domain failover notification from the domain cache. Domain cache periodically checks the database and updates the domain in memory. The failover timeout can leverage this component.
+
+During graceful failover domain update, we record the timeout in the domain data. Domain cache reads all domain data periodically and checks if any of the graceful failover should be timed out. If the domain cache get a graceful failover to be timed out. It can sends a notification to shard to update the domain from pending_active to active.
+
+### Buffer queue
```
mfateev: I would advise against adding this feature. The reason there is no buffering of these calls as they provide consistency guarantees. For example Signal should fail if workflow is already closed. 
I would reject requests during failover and rely on clients to retry. 
If someone really wants to skip on consistency they can always buffer requests in Kafka.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,230 @@
...
+1. Workflow progress could be lost.
+2. No causal consistency guarantee on domain level.
+
+The graceful domain failover uses to solve those two problems.
+
+## Use cases
+
+Users can trigger graceful domain failover via Cadence CLI:
+
+`cadence --domain cadence-global domain failover graceful -active_cluster XYZ -failover_timeout 120s`
```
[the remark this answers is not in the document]
yux0: Yes. After the feature complete, we will have some data about the baseline of graceful failover. Then we can provide a default timeout.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Cluster X
+![cross cluster X sequence diagram](3051-clusterX.png)
+
+1. Frontend receives a domain replication message.
+2. Frontend updates the domain data in Database with activeCluster set to Cluster Y and a higher failover version.
+3. Domain cache fetch domain updates in a refresh loop.
+4. Database returns the domain data.
+5. After the domain updates, domain cache sends a domain failover notification to each shard.
+6. After the domain updates, domain cache sends a domain failover notification to each shard.
```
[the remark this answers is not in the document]
yux0: Combined step 5,6. The intention is to show the cache notifies all shards.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,230 @@
...
+7. Shard 1 updates the shard info with a pending failover marker to insert.
+8. Shard 1 try to insert the failover marker and remove the pending failover marker from shard info after successful insertion.
```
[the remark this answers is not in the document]
yux0: Yes. Added a statement to explicitly mentioned in the task processor section that it pauses on processing tasks during the Pending_Active. This pause function relies on the multi-tenancy design so that it won't block other domains.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
yux0: Good suggestion. I think this is not the MVP and the feature can be configurable by domains. However, it depends on the user's requirements. We might still think about this feature.

In terms of consistency, it calls to the source cluster to get workflow state before buffering the request. If the workflow is closed, then Signal will fail.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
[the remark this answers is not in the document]
meiliang86: I think the point is that if user does not specify, it will use graceful failover up to the default timeout.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Use cases
+
+Users can trigger graceful domain failover via Cadence CLI:
+
+`cadence --domain cadence-global domain failover graceful -active_cluster XYZ -failover_timeout 120s`
+
+Users can force complete domain failover via Cadence CLI:
+
+`cadence --domain cadence-global domain update force -active_cluster XYZ`
```
meiliang86: It is a little bit confusing that one is called domain failover and the other is called domain update. We should make it consistent.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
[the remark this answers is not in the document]
yux0: yes. will specify the behavior

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
yux0: update the CLI to be consistent

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
emrahs: Agreed that buffer queue isn't part of the core requirements at this stage. Leaving it out also simplifies the design and creates crisp boundaries.

On the other hand, my understanding is that there are customers who want to have graceful failovers without sacrificing the availability of the synchronous APIs (Start, Signal, etc.). Asking them to create their own buffer is a viable option, but this will probably become a common enough pattern that it might make sense to solve at our end eventually.  Besides, if we do it for them, we can do more than merely buffering the requests in a simple queue such as checking the workflow status from the source cluster as @yux0 mentioned or providing some more guarantees around not buffering conflicting requests such as terminating workflow and then signaling it. 

In a later phase, I believe we should make the destination cluster immediately available for newly started workflows during graceful failovers, such that the cluster simply acts as "Active" for new workflows, rather than acting like "PendingActive". That would limit the impact of downtime and/or latency caused by graceful failover only to the workflows that were in-flight before the failover was triggered. This "graceful failover with immediate availability for new workflows" enhancement is currently just an idea that we haven't explored yet, but I think the availability and latency gains will be good enough to justify exploring it. If we do that, some sort of request buffering will make even more sense in order to provide a consistent experience for all in-flight workflows regardless of when exactly they were started.

Let's leave out request buffering in this phase, but keep it on the table for the next iteration. In the meantime, I'd love to hear your thoughts on the points I mentioned above.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,239 @@
...
+
+Cadence supports domain failover with multi-cluster setup. However, the problems with the current failover are:
+1. Workflow progress could be lost.
+2. No causal consistency guarantee on domain level.
+
+The graceful domain failover uses to solve those two problems.
+
+## Use cases
+
+Users can trigger graceful domain failover via Cadence CLI (timeout is optional):
```
andrewjdawson2016: I am confused why timeout is optional - it seems like the below design depends on a timeout? Do we provide a default if one is not given?

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,239 @@
...
+`cadence --domain cadence-global domain update -failover_type force -active_cluster XYZ`
+
+## Prerequisites
+There are conditions before starting a graceful domain failover.
+1. All clusters are available.
+2. The networking between clusters are available.
+3. The domain state has to be stable prior to graceful failover.
+
+## Limitation
+No concurrent graceful domain failover will be supported.
```
andrewjdawson2016: Do we actually reject requests for concurrent graceful failovers or just provide no guarantees if the user issues concurrent graceful failovers?

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
```diff
@@ -0,0 +1,239 @@
...
+Due to the complexity of the protocol, it will go through the architecture from cross-cluster level, single cluster level to host level.
+
+The basic protocol is to insert markers in the active-to-passive cluster to indicate the boundary when the domain switches to passive. On the other side, the passive-to-active cluster listens to those markers and switches domain to active after receiving all the markers.
+
+### Cross cluster level
+
+![cross clusters sequence diagram](3051-cross-clusters.png)
+1, The operator issues a graceful failover to the passive-to-active cluster.
+
+2 - 3, The passive-to-active cluster gets domain data from all other clusters for two purposes: 1. Make sure the network and clusters are available prior to start the graceful failover. 2. Make sure there is no ongoing failover.
```
andrewjdawson2016: If we want to be able to give the user an error message if they try to initiation concurrent graceful failovers then the before cluster X can respond back to cluster Y it needs to persist some data saying that it has responded yes to being able to handle a graceful failover. If it gets another attempt for graceful failover it needs to read this value and reject the request. Or we just say that users can initiate concurrent graceful failover but they lose all guarantees.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
yux0: Yes. We provides a default max support failover timeout

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
yux0: In race condition, users could issue concurrent graceful failover and if this happens then we don't guarantee anything

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
yux0: Yes. We can do optimization to block concurrent graceful failover. However, if the concurrent graceful failover happens, we provide no guarantees.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
mfateev: I think the problem with buffering is that you have to deal with invalid buffered requests. Then you start inventing DLQs and similar concepts. 
I believe this is premature optimization. Under normal situation graceful failover should be under 3 seconds, well within latency spike due to all sort of other types of failures.

docs/design/graceful-domain-failover/3051-graceful-domain-failover.md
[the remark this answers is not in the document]
mfateev: It is not just task processor. All start/signal, etc requests should be rejected if ever reach this shard after the failover marker is pending.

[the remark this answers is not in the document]
yux0: Yes. The requests will be rejected after the marker inserted.

PR #3130, opened 2020-03-24
ringpop: update hashring immediately on ring change

venkat1109:
**What changed?**
When a cadence host is added/removed/restarted, every other host in the cluster will receive a change notification via ringpop. This is the primary mechanism by which discovery / failure detection works today. When such a change notification is received, every node updates its consistent hash ring to route future requests to the correct owner. Its critical that the hashring update happens as soon as possible during deployments / restarts etc to avoid downtime / availability drops. Currently, there is an optimization to avoid too many updates to hashring within a short span of time. But this is hurting availability. 

This patch adds a fix by updating ring as soon as notification is received. In addition, a dedup map is added to resolver to avoid updating the ring when (a) nothing changes on an event (b) the host added or removed is for a different role. This should mitigate the too many updates within a short span of time problem.

**Why?**
To reduce availability dips during deployments and host restarts.

**How did you test it?**
Localhost as well as in a staging environment.

**Potential risks**
In the worst case, discovery / failure detection can be broken. This would mean unavailability or host stealing shards from each other continuously.

### DIFF
--- a/common/membership/rpServiceResolver.go
+++ b/common/membership/rpServiceResolver.go
@@ -81,8 +81,12 @@ func newRingpopServiceResolver(
 		membersMap:  make(map[string]struct{}),
 		listeners:   make(map[string]chan<- *ChangedEvent),
 	}
-	resolver.ringValue.Store(hashring.New(farm.Fingerprint32, replicaPoints))
+	resolver.ringValue.Store(newHashRing())
 	return resolver
+}
+
+func newHashRing() *hashring.HashRing {
+	return hashring.New(farm.Fingerprint32, replicaPoints)
 }
 
 // Start starts the oracle
@@ -117,7 +121,7 @@ func (r *ringpopServiceResolver) Stop() {
 	r.listenerLock.Lock()
 	defer r.listenerLock.Unlock()
 	r.rp.RemoveListener(r)
-	r.ringValue.Store(hashring.New(farm.Fingerprint32, replicaPoints))
+	r.ringValue.Store(newHashRing())
 	r.listeners = make(map[string]chan<- *ChangedEvent)
 	close(r.shutdownCh)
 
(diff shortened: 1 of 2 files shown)


common/membership/rpServiceResolver.go
```diff
@@ -271,19 +283,40 @@ func (r *ringpopServiceResolver) refreshRingWorker() {
...
+	newMembersMap := make(map[string]struct{}, len(addrs))
+	for _, addr := range addrs {
+		newMembersMap[addr] = struct{}{}
+		if _, ok := r.membersMap[addr]; !ok {
+			changed = true
+		}
+	}
+	for addr := range r.membersMap {
+		if _, ok := newMembersMap[addr]; !ok {
+			changed = true
```
wxing1292: missing a `break`?

venkat1109: added

common/membership/rpServiceResolver.go
```diff
@@ -69,18 +71,18 @@ func newRingpopServiceResolver(
...
 		shutdownCh:  make(chan struct{}),
 		logger:      logger.WithTags(tag.ComponentServiceResolver, tag.Service(service)),
-
-		ring: hashring.New(farm.Fingerprint32, replicaPoints),
-
-		listeners: make(map[string]chan<- *ChangedEvent),
+		membersMap:  make(map[string]struct{}),
+		listeners:   make(map[string]chan<- *ChangedEvent),
 	}
+	resolver.ringValue.Store(hashring.New(farm.Fingerprint32, replicaPoints))
```
vancexu: nit: consider making `hashring.New(farm.Fingerprint32, replicaPoints)` a method.

venkat1109: Done

PR #3132, opened 2020-03-24
ringpop: add method to selfEvict from ring

venkat1109:
**What changed?**
This patch adds a method to the MembershipMonitor interface to self evict a member from the ring. This method will be used in subsequent diffs to achieve graceful shutdown during deployments.

**Why?**
The reason for this change is to support graceful shutdown during deployments.

**How did you test it?**
This patch doesn't change any existing behavior. It simply exposes a new method.

**Potential risks**
No risk, the new method is not invoked by any piece of code.


common/membership/rpMonitor.go
```diff
@@ -115,6 +115,10 @@ func (rpo *ringpopMonitor) WhoAmI() (*HostInfo, error) {
 	return NewHostInfo(address, labels.AsMap()), nil
 }
 
+func (rpo *ringpopMonitor) EvictSelf() error {
+	return rpo.rp.SelfEvict()
```
vancexu: maybe keep using name EvictSelf?

venkat1109: the underlying ringpop library uses the name SelfEvict() which cannot be changed. Yes, we can use the same name. I renamed it because method names usually start with "verb" which makes them very readable. And this underlying library is only invoked in this place.. so nobody cares about this code.. if you feel strongly, I can use the same name for consistency.

yycptt: +1 for making method name start with verb.

Issue #1454, opened 2019-02-12
Implement clean host shutdown to avoid errors during deployments


mfateev:
Currently deployments lead to small number of intermittent request failures. While it is not catastrophic it still affects availability metrics and might alert customers. 
This issue is to implement clean service hosts shutdown to avoid any errors in case of scheduled deployments.

vitarb:
Looking at the data, there is indeed a strong correlation between deployments and availability drops. We currently observe availability drops down to 80-90% during the deployment time. To address this issue we should handle SIGTERM signal that is sent a few seconds before SIGKILL by our deployment infrastructure to notify the ring about shard ownership changes ensuring that old host is still online while ownership change is inflight.

demirkayaender:
This is done

PR #3134, opened 2020-03-25
First pass at graceful shutdown handlers

venkat1109:
**What changed?**
This patch implements mechanism to gracefully drain traffic during deployments to avoid availability drops. The mechanism varies somewhat based on the role.

The shutdown protocol on some host H looks like the following:

**History**
1. Remove H from the membership ring. This will make other members believe that H is unhealthy
2. Wait for other members to discover that H is unhealthy
3. Stop acquiring new shards (periodically or based on other membership changes) on H
4. Wait for shard ownership to transfer (and inflight requests to drain) while still accepting new requests on H
5. Reject all requests arriving at rpc handler on H to avoid taking on more work
6. Wait for grace period
7. Force stop the whole world and exit

**Frontend**
1. Fail rpc health check, this will cause client side load balancer to stop forwarding requests to this node
2. Wait for failure detection time, typically this will be around 5s or less
3. Stop taking new rpc requests by returning InternalServiceError
4. Wait for a second
5. Stop the world and exit

**Matching**
1. Remove H from the membership ring. This will make other members believe that H is unhealthy
2. Wait for others to discover that H is unhealthy and traffic to drain
3. Stop the world and exit

**Why?**
Grace shutdown mechanism is needed to reduce availability dips during deployments

**How did you test it?**
Tested locally. Also tested on a staging environment.

**Potential risks**
Graceful shutdown fails to work and there continues to be availability dips during deployment. 
A bug in the code causes process to crash during shutdown.

First pass at fixing #1454

### DIFF
--- a/service/frontend/accessControlledHandler.go
+++ b/service/frontend/accessControlledHandler.go
@@ -41,14 +41,14 @@ var errUnauthorized = &shared.BadRequestError{Message: "Request unauthorized."}
 type AccessControlledWorkflowHandler struct {
 	resource.Resource
 
-	frontendHandler workflowserviceserver.Interface
+	frontendHandler ServerHandler
 	authorizer      authorization.Authorizer
 
 	startFn func()
 	stopFn  func()
 }
 
-var _ workflowserviceserver.Interface = (*AccessControlledWorkflowHandler)(nil)
+var _ ServerHandler = (*AccessControlledWorkflowHandler)(nil)
 
 // NewAccessControlledHandlerImpl creates frontend handler with authentication support
 func NewAccessControlledHandlerImpl(wfHandler *DCRedirectionHandlerImpl, authorizer authorization.Authorizer) *AccessControlledWorkflowHandler {
@@ -60,8 +60,6 @@ func NewAccessControlledHandlerImpl(wfHandler *DCRedirectionHandlerImpl, authori
 		Resource:        wfHandler.Resource,
 		frontendHandler: wfHandler,
 		authorizer:      authorizer,
-		startFn:         func() { wfHandler.Start() },
-		stopFn:          func() { wfHandler.Stop() },
 	}
 }
 
@@ -73,6 +71,12 @@ func (a *AccessControlledWorkflowHandler) RegisterHandler() {
 	a.GetDispatcher().Register(metaserver.New(a))
 }
 
+// UpdateHealthStatus sets the health status for this rpc handler.
+// This health status will be used within the rpc health check handler
+func (a *AccessControlledWorkflowHandler) UpdateHealthStatus(status HealthStatus) {
+	a.frontendHandler.UpdateHealthStatus(status)
+}
+
 // Health callback for for health check
 func (a *AccessControlledWorkflowHandler) Health(ctx context.Context) (*health.HealthStatus, error) {
 	hs := &health.HealthStatus{Ok: true, Msg: common.StringPtr("auth is good")}
@@ -81,12 +85,12 @@ func (a *AccessControlledWorkflowHandler) Health(ctx context.Context) (*health.H
 
 // Start starts the handler
 func (a *AccessControlledWorkflowHandler) Start() {
-	a.startFn()
+	a.frontendHandler.Start()
 }
 
 // Stop stops the handler
 func (a *AccessControlledWorkflowHandler) Stop() {
-	a.stopFn()
+	a.frontendHandler.Stop()
 }
 
 // CountWorkflowExecutions API call
--- a/service/frontend/accessControlledHandler.go
+++ b/service/frontend/accessControlledHandler.go
@@ -43,9 +43,6 @@ type AccessControlledWorkflowHandler struct {
 
 	frontendHandler ServerHandler
 	authorizer      authorization.Authorizer
-
-	startFn func()
-	stopFn  func()
 }
 
 var _ ServerHandler = (*AccessControlledWorkflowHandler)(nil)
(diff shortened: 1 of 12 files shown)


service/history/handler.go
```diff
@@ -165,7 +177,7 @@ func (h *Handler) CreateEngine(
 func (h *Handler) Health(ctx context.Context) (*health.HealthStatus, error) {
 	h.startWG.Wait()
 	h.GetLogger().Debug("History health check endpoint reached.")
-	hs := &health.HealthStatus{Ok: true, Msg: common.StringPtr("history good")}
+	hs := &health.HealthStatus{Ok: true, Msg: common.StringPtr("OK")}
```
emrahs: nit: I'd consider setting the "Msg" to "Shutting down" when the host is in that state.

emrahs: Also, shouldn't we be reporting "unhealthy" status here? It seems like all endpoints are returning errShuttingDown even though the reported state is "healthy".

service/frontend/accessControlledHandler.go
```diff
@@ -41,14 +41,14 @@ var errUnauthorized = &shared.BadRequestError{Message: "Request unauthorized."}
 type AccessControlledWorkflowHandler struct {
 	resource.Resource
 
-	frontendHandler workflowserviceserver.Interface
+	frontendHandler ServerHandler
 	authorizer      authorization.Authorizer
 
 	startFn func()
```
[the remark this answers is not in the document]
venkat1109: No, forgot to remove these

service/history/handler.go
venkat1109: Setting healthStatus to Ok=false has no effect for history - because we don't use rpc routing / load balancing mechanisms for forwarding requests to history. But I can still set it to ShuttingDown (even though it has no effect) - I didn't do it only because I didn't test that change in staging.

service/frontend/accessControlledHandler.go
```diff
@@ -73,6 +71,12 @@ func (a *AccessControlledWorkflowHandler) RegisterHandler() {
...
 }
 
+// UpdateHealthStatus sets the health status for this rpc handler.
+// This health status will be used within the rpc health check handler
+func (a *AccessControlledWorkflowHandler) UpdateHealthStatus(status HealthStatus) {
+	a.frontendHandler.UpdateHealthStatus(status)
+}
+
 // Health callback for for health check
 func (a *AccessControlledWorkflowHandler) Health(ctx context.Context) (*health.HealthStatus, error) {
```
yycptt: Do we need to forward this call to the underlying frontendHandler?

service/history/handler.go
venkat1109: Oh another reason I wasn't marking it unhealthy was because I wasn't rejecting calls to RespondDecision/ActivityTaskCompleted, RecordDecison/ActivityTaskStarted - because for these, most of the work is already done and rejecting it at last stage is probably not desirable. If shards have migrated, these calls will fail anyways - but I decided to accept them anyways till the end.

service/frontend/accessControlledHandler.go
venkat1109: ofcourse, thanks for catching this.

venkat1109: done

service/frontend/accessControlledHandler.go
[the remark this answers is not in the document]
venkat1109: done

service/frontend/workflowHandler.go
```diff
@@ -59,13 +60,32 @@ const (
 	defaultLastMessageID                 = -1
 )
 
-var _ workflowserviceserver.Interface = (*WorkflowHandler)(nil)
+const (
+	// HealthStatusOK is used when this node is healthy and rpc requests are allowed
+	HealthStatusOK HealthStatus = iota
```
yycptt: nit: relying on the default value to be 0 and not explicitly set the healthState to `HealthStatusOK` in `NewWorkflowHandler()` is a little bit confusing. Probably follow uber's best practice for enum and use  `HealthStatusOK HealthStatus = iota + 1`

service/history/service.go
```diff
@@ -432,10 +434,50 @@ func (s *Service) Stop() {
 		return
 	}
 
+	// initiate graceful shutdown :
+	// 1. remove self from the membership ring
+	// 2. wait for other members to discover we are going down
+	// 3. stop acquiring new shards (periodically or based on other membership changes)
+	// 4. wait for shard ownership to transfer (and inflight requests to drain) while still accepting new requests
+	// 5. reject all requests arriving at rpc handler to avoid taking on more work
```
yycptt: nit: probably worth documenting which requests will still be accepted at this stage.

service/frontend/workflowHandler.go
venkat1109: that's a good idea. Will do that.

service/history/service.go
venkat1109: 👍

service/history/handler.go
emrahs: Thanks for the info! That gives me another question though: why don't we continue accepting calls to all of the APIs as long as we still have the shard ownership, just like we do for Respond* and Record*?

venkat1109: We could and I agree the benefit of this last step is questionable. But the reasoning is the following - for things like StartWF, SignalWF etc - we are about to start new work which can potentially take up to 200ms or more, when we know for sure that we are going down in another 500ms or so.. since we already waited ~5s or so, the shard should have most likely moved by now.. if not, we force the caller to go to the current shard owner.. everytime we reject, it triggers a backoff/retry on the caller (frontend) - where each retry will recompute the shard_owner.. so by rejecting, we are increasing the chance of request not failing and going to right owner.. 

Reason I say benefit is questionable is because - since we wait ~5s, its usually enough time for everyone on the cluster to learn that this instance is going down.. so its unlikely frontend will still forward calls to this node..

emrahs: OK, now I see that the time window is actually much shorter than I thought at first and with that context your thought process makes sense.

PR #3135, opened 2020-03-25
Validate search attribute value

yycptt:
**What changed?**
Add validation logic to make sure search attribute value matches its type.

Also fixed deserializing logic for search attribute in describe workflow CLI command. Previously it doesn't support list of values (used by `CadenceChangeVersion` and `BinaryChecksums`). 

**Why?**
https://github.com/uber/cadence/issues/3061

**How did you test it?**
Unit test and local canary (searchAttribute test)

**Potential risks**
In the worst case, all workflows using search attributes will fail.


common/elasticsearch/validator/searchAttrValidator.go
```diff
@@ -73,12 +74,19 @@ func (sv *SearchAttributesValidator) ValidateSearchAttributes(input *gen.SearchA
...
+			sv.logger.WithTags(tag.ESKey(key), tag.WorkflowDomainName(domain)).
+				Error("invalid search attribute key")
+			return &gen.BadRequestError{Message: fmt.Sprintf("%s is not a valid search attribute key", key)}
+		}
+		// verify: value has the correct type
+		if !sv.isValidSearchAttributesValue(validAttr, key, val) {
 			sv.logger.WithTags(tag.ESKey(key), tag.WorkflowDomainName(domain)).
-				Error("invalid search attribute")
-			return &gen.BadRequestError{Message: fmt.Sprintf("%s is not valid search attribute", key)}
+				Error(fmt.Sprintf("%s is not a valid search attribute value for key %s", val, key))
```
vancexu: let's use tags instead of formatted message.

yycptt: Done.

PR #3136, opened 2020-03-25
shardController: improvements for graceful shutdown

venkat1109:
**What changed?**
This patch contains a set of changes to reduce availability dips during shard load / unloads. The changes specifically are:

- On ring membership changes, the current code used to remove shards no longer owned by the instance. However pro-actively removing the shard doesn't buy much but causes availability blip. 
Shards no longer owned will be automatically closed when the new owner steals the shard from this instance. So, this patch gets rid of removeShard on membership change and also implements mechanism to unload queue processors on shardOwnershipLost errors.

- shardContext is the class that implements operations that can be executed only by a shard owner. Pretty much any workflow related operation needs to go through this class. During deployments, when shard ownership moves to another instance (and the host which is going down is draining traffic),  the shardContext operations will receive a ShardOwnershipLost error on the instance that is going down. However, the retry loop in shardContext is dumb and continues to loop through retries even after this error. This can hold requests for much longer, increasing latency, decreasing availability and creating additional DB load. This patch fails the requests as soon as shard is closed due to shardownershipLost errors.

- When a shard is closed due to ownershipLost errors, previously, a channel is used to send an async notification to the shardController, which in turn will delete the shard entry from its map and stop the processors. However, this async mechanism has a race condition during shutdown when shards are constantly moving around. This patch eliminates the race by using a callback function instead of async mechanism.

**Why?**
The goal here is to reduce availability dips during deployment (shard unload  / load)

**How did you test it?**
Tested locally and in staging.

**Potential risks**
This patch modifies the shutdown path of shard / shardController etc which are key components involved in all workflow operations. A bug could cause shards to constantly move between hosts reducing availability or increasing latency.  

Related to #1454

### DIFF
--- a/service/history/historyEngineInterfaces.go
+++ b/service/history/historyEngineInterfaces.go
@@ -119,7 +119,7 @@ type (
 		completeTimerTask(timerTask *persistence.TimerTaskInfo)
 		getAckLevel() timerKey
 		getReadLevel() timerKey
-		updateAckLevel()
+		updateAckLevel() error
 	}
 
 	historyEventNotifier interface {
--- a/service/history/historyEngineInterfaces.go
+++ b/service/history/historyEngineInterfaces.go
@@ -62,7 +62,7 @@ type (
 		completeQueueTask(taskID int64)
 		getQueueAckLevel() int64
 		getQueueReadLevel() int64
-		updateQueueAckLevel()
+		updateQueueAckLevel() error
 	}
 
 	queueTaskInfo interface {
--- a/service/history/shardContext.go
+++ b/service/history/shardContext.go
@@ -879,7 +879,6 @@ func (s *shardContextImpl) closeShard() {
 
 	go func() {
 		s.closeCallback(s.shardID, s.shardItem)
-		s.shardItem.stopEngine()
 	}()
 
 	// fails any writes that may start after this point.
--- a/service/history/shardContext.go
+++ b/service/history/shardContext.go
@@ -21,6 +21,7 @@
 package history
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"sync"
@@ -115,8 +116,8 @@ type (
 		rangeID          int64
 		executionManager persistence.ExecutionManager
 		eventsCache      eventsCache
-		closeCh          chan<- int
-		isClosed         bool
+		closeCallback    func(int, *historyShardsItem)
+		closed           int32
 		config           *Config
 		logger           log.Logger
 		throttledLogger  log.Logger
@@ -139,6 +140,9 @@ type (
 )
 
 var _ ShardContext = (*shardContextImpl)(nil)
+
+// ErrShardClosed is returned when shard is closed and a req cannot be processed
+var ErrShardClosed = errors.New("shard closed")
 
 const (
 	logWarnTransferLevelDiff = 3000000 // 3 million
@@ -489,6 +493,7 @@ Create_Loop:
 					} else {
 						// Shard is stolen, trigger shutdown of history engine
 						s.closeShard()
+						break Create_Loop
 					}
 				}
 			default:
@@ -503,6 +508,7 @@ Create_Loop:
 						// At this point we have no choice but to unload the shard, so that it
 						// gets a new RangeID when it's reloaded.
 						s.closeShard()
+						break Create_Loop
 					}
 				}
 			}
@@ -580,6 +586,7 @@ Update_Loop:
 					} else {
 						// Shard is stolen, trigger shutdown of history engine
 						s.closeShard()
+						break Update_Loop
 					}
 				}
 			default:
@@ -594,6 +601,7 @@ Update_Loop:
 						// At this point we have no choice but to unload the shard, so that it
 						// gets a new RangeID when it's reloaded.
 						s.closeShard()
+						break Update_Loop
 					}
 				}
 			}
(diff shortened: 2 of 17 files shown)


service/history/shardContext.go
```diff
@@ -856,24 +868,23 @@ func (s *shardContextImpl) getRangeID() int64 {
...
+	if !atomic.CompareAndSwapInt32(&s.closed, 0, 1) {
 		return
 	}
 
-	s.isClosed = true
-
-	go s.shardItem.stopEngine()
+	go func() {
+		s.closeCallback(s.shardID, s.shardItem)
+		s.shardItem.stopEngine()
```
yux0: does closeCallback call stopEngine?

service/history/shardController.go
```diff
@@ -167,18 +165,24 @@ func (c *shardController) getEngineForShard(shardID int) (Engine, error) {
...
+	return item.getOrCreateEngine(c.shardClosedCallback)
 }
 
-func (c *shardController) removeEngineForShard(shardID int) {
+func (c *shardController) removeEngineForShard(shardID int, shardItem *historyShardsItem) {
 	sw := c.metricsScope.StartTimer(metrics.RemoveEngineForShardLatency)
 	defer sw.Stop()
 	item, _ := c.removeHistoryShardItem(shardID)
-	if item != nil {
+	if item != nil && (item == shardItem || shardItem == nil) {
```
yux0: What is the case to compare the shardItem and item?

service/history/shardContext.go
venkat1109: Yes - forgot to remove this additional call to stopEngine. Will remove it.

service/history/shardController.go
venkat1109: It is just a defensive check to make sure we are deleting what we intend to delete. As for the specific scenario, consider the following:

(1) ShardContext gets an OwnershipLostError and decides to close shard - so it is about to call this method
(2) In the meanwhile, some other piece of code also calls removeEngineForShard - say, the frontend API for CloseShard()
(3) After (2) is done but before (1) is executed, acquireShards loads the same shard once again because now, we are supposed to own that shard (because of some other host going down)
(4) Now, (1) executes and without the defensive check, it will remove the wrong shardItem. Its not the end of the world, since we will recreate the shard on the next request or so  - but it doesn't hurt to have this additional check. I will add comment to clarify this.

service/history/shardContext.go
venkat1109: done

service/history/shardController.go
yux0: make sense to me now

service/history/historyEngineInterfaces.go
```diff
@@ -62,7 +62,7 @@ type (
 		completeQueueTask(taskID int64)
 		getQueueAckLevel() int64
 		getQueueReadLevel() int64
-		updateQueueAckLevel()
+		updateQueueAckLevel() error
```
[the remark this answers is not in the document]
venkat1109: let me check. Show all implementations for interface showed only this file in my editor.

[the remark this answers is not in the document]
venkat1109: Yes, it makes sense to add to other processors as well. I will update the diff.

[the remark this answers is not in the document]
venkat1109: Just updated the diff to stop timerQProcessor. I though queueProcessor is the base class for everything previously, turns out I was wrong :)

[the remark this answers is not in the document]
yycptt: Actually I mean there's another interface called `timerQueueAckMgr` in this file. It also has a `updateAckLevel()` method. Do we need to add a return value for error to that method and check the error in `timerQueueProcessorBase`?

service/history/timerQueueProcessor.go
```diff
@@ -261,6 +261,11 @@ func (t *timerQueueProcessorImpl) completeTimersLoop() {
 				err := t.completeTimers()
 				if err != nil {
 					t.logger.Info("Failed to complete timers.", tag.Error(err))
+					if err == ErrShardClosed {
```
vancexu: should this be added to all callers of completeTimers, such as L256?

vancexu: should this be added to all callers of completeTimers, such as L256?

service/history/historyEngineInterfaces.go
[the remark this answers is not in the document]
venkat1109: I see.. there actually are two updateAckLevels for timers, I think I updated the code path that's calling UpdateTimerAckLevel but there is still UpdateTimerClusterAckLevel which is the one you are talking about.

UpdateTimerAckLevel(ackLevel time.Time) error
UpdateTimerClusterAckLevel(cluster string, ackLevel time.Time) error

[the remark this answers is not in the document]
yycptt: I see. Yeah, it's quite confusing. `UpdateTimerAckLevel` is for updating ack level for all cluster, while `UpdateTimerClusterAckLevel` is only for a specific cluster. I think we also have the same things for transfer queue. Maybe you also want to update that.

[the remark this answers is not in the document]
venkat1109: Yup, thanks !

service/history/timerQueueProcessor.go
yycptt: I think at L256, the shard is already being shut down, meaning someone already called `t.Stop()`. Check the error and call `Stop()` again will be a no-op.

venkat1109: not necessary on L256 because that code path is exiting the loop anyways (its the shutdown path).

PR #3137, opened 2020-03-25
Add failover end timeout in domain data

yux0:
**What changed?**
Add failover config in domain data

**Why?**
The failover config is for graceful domain failover. The config 

**How did you test it?**
Database tests

**Potential risks**
Fail to update DV schema prior to new deployment will cause service startup failure


common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -408,6 +426,14 @@ func (m *cassandraMetadataPersistenceV2) GetDomain(request *p.GetDomainRequest)
...
 	replicationConfig.Clusters = p.GetOrUseDefaultClusters(m.currentClusterName, replicationConfig.Clusters)
 
+	var failoverConfig *p.DomainFailoverConfig
+	if failoverStartTime > 0 && failoverTimeout > 0 {
+		failoverConfig = &p.DomainFailoverConfig{
+			StartTime: failoverStartTime,
+			Timeout:   failoverTimeout,
+		}
+	}
+
```
vancexu: Looks if is not necessary. maybe just 
```
&p.DomainFailoverConfig{
			StartTime: failoverStartTime,
			Timeout:   failoverTimeout,
		}
```
in line 445

schema/cassandra/cadence/schema.cql
```diff
@@ -404,6 +409,7 @@ CREATE TABLE domains_by_name_v2 (
   config_version                bigint, -- indicating the version of domain config, excluding the failover / change of active cluster name
   failover_version              bigint, -- indicating the version of active domain only, used for domain failover
   failover_notification_version bigint, -- indicating the last change related to domain failover
+  failover_config               frozen<domain_failover_config>, -- indicating domain failover state
```
vancexu: On the naming things. I feel failover_state might be more accurate in this case?
Why not just get rid of this struct and unwrap failover_start_time and failover_timeout?

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -291,6 +303,8 @@ func (m *cassandraMetadataPersistenceV2) UpdateDomain(request *p.InternalUpdateD
 		request.ConfigVersion,
 		request.FailoverVersion,
 		request.FailoverNotificationVersion,
+		failoverConfig.StartTime,
```
andrewjdawson2016: Is there validation we need to do on StartTime and and Timeout. A few thoughts 

- They should either both be set or neither should be set any other combination is invalid?
- Do we want protections on how long this timeout should be?

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
```diff
@@ -573,3 +600,15 @@ func (m *cassandraMetadataPersistenceV2) deleteDomain(name, ID string) error {
...
 }
+
+func (m *cassandraMetadataPersistenceV2) getFailoverConfig(
+	request *p.InternalUpdateDomainRequest,
+) p.DomainFailoverConfig {
+
+	config := p.DomainFailoverConfig{}
+	if request.FailoverConfig != nil {
+		config.StartTime = request.FailoverConfig.StartTime
+		config.Timeout = request.FailoverConfig.Timeout
```
andrewjdawson2016: On failover config do we want some id to identify which failover this is? I am wondering if I call update domain twice with failover configs if if we want to include some uuid to identify these failover requests in logs. This is mainly for debugability - I don't think its needed for correctness, from a correctness point of view we only care about the latest request.

andrewjdawson2016: What was the idea of having a timeout here instead of just specifying an end time? By giving a start time and a timeout I am wondering if the application logic is going to have to do startTime + timeout >= something in a bunch a places?

yux0: You are right. I will use end time

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
yux0: Updated to use failover end time. Here if it is nil, it means there is no graceful failover happening

schema/cassandra/cadence/schema.cql
yux0: Removed the struct and only use failover end time

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
yux0: Updated

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
yux0: Yes. I changed them to failover end time and only update one field now. The protection logic will be in business logic.

common/persistence/cassandra/cassandraMetadataPersistenceV2.go
yux0: Failover version should be able to tell which failover it is.

.gen/go/sqlblobs/sqlblobs.go
```diff
@@ -3380,6 +3409,21 @@ func (v *DomainInfo) IsSetVisibilityArchivalURI() bool {
 	return v != nil && v.VisibilityArchivalURI != nil
 }
 
+// GetFailoverEndTime returns the value of FailoverEndTime if it is set or its
+// zero value if it is unset.
+func (v *DomainInfo) GetFailoverEndTime() (o int64) {
+	if v != nil && v.FailoverEndTime != nil {
```
vancexu: reuse `IsSetFailoverEndTime`?

yux0: This is generated code

.gen/go/sqlblobs/sqlblobs.go
```diff
@@ -2289,6 +2289,7 @@ type DomainInfo struct {
 	HistoryArchivalURI          *string           `json:"historyArchivalURI,omitempty"`
 	VisibilityArchivalStatus    *int16            `json:"visibilityArchivalStatus,omitempty"`
 	VisibilityArchivalURI       *string           `json:"visibilityArchivalURI,omitempty"`
+	FailoverEndTime             *int64            `json:"failoverEndTime,omitempty"`
```
yycptt: Please also check in the change for idl repo.

schema/cassandra/cadence/versioned/v0.27/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "CurrVersion": "0.27",
```
yycptt: Why 0.27? I think we only increase version once per release. 0.12.x is using 0.25, so I guess we can keeping 0.26 here. You can have multiple cql file in one version.

yux0: I am thinking about having the 0.26 in 0.12 release. Let's discuss offline.

yycptt: I see. Make sense.

PR #3138, opened 2020-03-25
Log workflow info also when visibility data goes to DLQ

mkolodezny:
**What changed?**
Log workflow info also when visibility data goes to DLQ

**Why?**
We need those tags to filter messages which are going to DLQ (failed to be handled by Kafka consumer )

**How did you test it?**
unit tests

**Potential risks**
messages wont be delivered into DLQ


service/worker/indexer/esProcessor.go
```diff
@@ -23,6 +23,8 @@ package indexer
 import (
 	"context"
 	"encoding/json"
+	"github.com/uber/cadence/.gen/go/indexer"
+	"github.com/uber/cadence/common/codec"
```
yycptt: Please group all the import from cadence together. Same for the other file.

service/worker/indexer/esProcessor.go
```diff
@@ -65,6 +67,7 @@ type (
 		config        *Config
 		logger        log.Logger
 		metricsClient metrics.Client
+		msgEncoder    codec.BinaryEncoder
```
yycptt: Making `esProcessorImpl` to have its own msgEncoder makes the code a little bit inflexible because if we need to change the message encoding in the further, we will have to modify both `esProcessor` and `indexerProcessor`. It also introduces some duplicated logic for message decoding. 

Given that `indexerProcessor` already has a msgEncoder and a method for decoding kafka message, can we somehow reuse that? For example, pass the encoder as a parameter when initializing `esProcessor`. For the decoding logic, maybe we can create a new util function for that.

service/worker/indexer/esProcessor.go
```diff
@@ -191,6 +195,13 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
 	}
 
 	if nack {
+		wid, rid, domainID, err := p.getMsgInfo(kafkaMsg.message)
+		if err != nil {
+			p.logger.Error("Unable to process message.", tag.WorkflowID(wid), tag.WorkflowRunID(rid),
```
yycptt: How about merging this log with the log on L168, so that in the one log we can have both workflow information and the response from es? I think it will make investigation a little bit easier.

service/worker/indexer/esProcessor.go
```diff
@@ -89,6 +92,7 @@ func NewESProcessorAndStart(config *Config, client es.Client, processorName stri
 		config:        config,
 		logger:        logger.WithTags(tag.ComponentIndexerESProcessor),
 		metricsClient: metricsClient,
+		msgEncoder:    codec.NewThriftRWEncoder(),
```
vancexu: This should be passed in to be same as indexer

service/worker/indexer/esProcessor.go
```diff
@@ -191,6 +195,13 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
 	}
 
 	if nack {
+		wid, rid, domainID, err := p.getMsgInfo(kafkaMsg.message)
+		if err != nil {
```
vancexu: if - else logic looks not right. if err != nil means decode error right?

service/worker/indexer/esProcessor.go
```diff
@@ -191,6 +195,13 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
...
+			p.logger.Error("Unable to process message.", tag.WorkflowID(wid), tag.WorkflowRunID(rid),
+				tag.WorkflowDomainID(domainID))
+		} else {
+			p.logger.Error("Unable to process message with failed decoding.")
```
vancexu: Looks duplicate log from getMsgInfo, could remove.

service/worker/indexer/esProcessor.go
```diff
@@ -199,6 +210,23 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
...
+func (p *esProcessorImpl) getMsgInfo(kafkaMsg messaging.Message) (wid string, rid string, domainID string, err error) {
+	var msg indexer.Message
+	if err := p.msgEncoder.Decode(kafkaMsg.Value(), &msg); err != nil {
+		p.logger.Error("Failed to deserialize kafka message.", tag.Error(err))
+		return "", "", "", err
+	}
+	return *msg.WorkflowID, *msg.RunID, *msg.DomainID, nil
+}
+
+func (p *esProcessorImpl) deserialize(payload []byte) (*indexer.Message, error) {
```
vancexu: looks not used, could remove if true

service/worker/indexer/esProcessor.go
```diff
@@ -23,12 +23,15 @@ package indexer
 import (
 	"context"
 	"encoding/json"
+
```
vancexu: nit: remove

service/worker/indexer/esProcessor_test.go
```diff
@@ -23,6 +23,9 @@ package indexer
 import (
 	"encoding/json"
 	"errors"
+	"github.com/uber/cadence/.gen/go/indexer"
```
vancexu: nit: separate standard dep from others.

service/worker/indexer/esProcessor.go
```diff
@@ -199,6 +202,34 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
...
 
+func (p *esProcessorImpl) getKafkaMsg(key string) (kafkaMsg *kafkaMessageWithMetrics, ok bool) {
+	msg, ok := p.mapToKafkaMsg.Get(key)
+	if !ok {
+		return // duplicate kafka message
+	}
+	kafkaMsg, ok = msg.(*kafkaMessageWithMetrics)
+	if !ok { // must be bug in code and bad deployment
+		p.logger.Fatal("Message is not kafka message.", tag.ESKey(key))
+		return
```
vancexu: nit: remove `return` since it is not reachable because of above `Fatal`

service/worker/indexer/esProcessor.go
```diff
@@ -181,13 +187,9 @@ func (p *esProcessorImpl) nackKafkaMsg(key string) {
 }
 
 func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
-	msg, ok := p.mapToKafkaMsg.Get(key)
+	var kafkaMsg, ok = p.getKafkaMsg(key)
```
venkat1109: Style: Prefixing with var is neither necessary nor a common practice in Go.

service/worker/indexer/esProcessor.go
```diff
@@ -199,6 +201,33 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
...
+		return // duplicate kafka message
+	}
+	kafkaMsg, ok = msg.(*kafkaMessageWithMetrics)
+	if !ok { // must be bug in code and bad deployment
+		p.logger.Fatal("Message is not kafka message.", tag.ESKey(key))
+	}
+	return kafkaMsg, ok
+}
+
+func (p *esProcessorImpl) getMsgWithInfo(key string) (wid string,
```
venkat1109: style: stick to everything on one line (if the line length is within 120)
```
func (p *esProcessorImpl) getMsgWithInfo(key string) (string,string, string)

(or)

func (p *esProcessorImpl) getMsgWithInfo(
 key string,
) (string, string, string) {
}
```

service/worker/indexer/esProcessor.go
```diff
@@ -199,6 +201,33 @@ func (p *esProcessorImpl) ackKafkaMsgHelper(key string, nack bool) {
...
+	rid string, domainID string) {
+	kafkaMsg, ok := p.getKafkaMsg(key)
+	if !ok {
+		return
+	}
+
+	var msg indexer.Message
+	if err := p.msgEncoder.Decode(kafkaMsg.message.Value(), &msg); err != nil {
+		p.logger.Error("Failed to deserialize kafka message.", tag.Error(err))
```
venkat1109: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

service/worker/indexer/esProcessor_test.go
```diff
@@ -32,6 +32,9 @@ import (
 	"github.com/stretchr/testify/suite"
 	"go.uber.org/zap"
 
+	"github.com/uber/cadence/.gen/go/indexer"
```
venkat1109: This import ordering looks non-standard - make sure your editor is setup to run goimports on save file

PR #3139, opened 2020-03-25
Gracefully handle polling history on standby

andrewjdawson2016:
Return more detailed error on standby side when polling for history. Also enable skip archival on getHistory. 

We do autoforwarding of startWorkflow requests but not of getHistory this makes a bad experience for customers that want to start a workflow on then block on the result from the standby side. This error message and not querying archival will improve the experience here. 

Low risk


service/frontend/workflowHandler.go
```diff
@@ -1734,10 +1734,12 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		getRequest.MaximumPageSize = common.Int32Ptr(common.GetHistoryMaxPageSize)
 	}
 
-	enableArchivalRead := wh.GetArchivalMetadata().GetHistoryConfig().ReadEnabled()
-	historyArchived := wh.historyArchived(ctx, getRequest, domainID)
-	if enableArchivalRead && historyArchived {
-		return wh.getArchivedHistory(ctx, getRequest, domainID, scope)
+	if !getRequest.GetSkipArchival() {
+		enableArchivalRead := wh.GetArchivalMetadata().GetHistoryConfig().ReadEnabled()
+		historyArchived := wh.historyArchived(ctx, getRequest, domainID)
```
yycptt: Not related to your change but I think if `historyArchived` is true and either `skipArchival` or `enabledArchivalRead` is false, we can return `EntityNotExistError` directly. There's no need to `PollMutableState` from history host again.

PR #3140, opened 2020-03-25
Add autofowarding of consistent queries

andrewjdawson2016:
Autoforward of consistent queries. This is done because some customers do not control which datacenter they call query out of. If they want a consistent query result then the only option is to autoforward these queries.


service/frontend/dcRedirectionHandler.go
```diff
@@ -548,8 +548,15 @@ func (handler *DCRedirectionHandlerImpl) QueryWorkflow(
...
-			resp, err = remoteClient.QueryWorkflow(ctx, request)
+			// Only autofoward consistent queries, this is done for two reasons:
+			// 1. Query is meant to be fast, autoforwarding all queries will increase latency.
+			// 2. If eventual consistency was requested then the results from running out of local dc will be fine.
+			if request.GetQueryConsistencyLevel() == shared.QueryConsistencyLevelStrong {
+				remoteClient := handler.GetRemoteFrontendClient(targetDC)
+				resp, err = remoteClient.QueryWorkflow(ctx, request)
+			} else {
+				resp, err = handler.frontendHandler.QueryWorkflow(ctx, request)
+			}
```
meiliang86: Is there a unit test for this?

andrewjdawson2016: Yes service/frontend/dcRedirectionHandler_test.go

PR #3145, opened 2020-03-27
Add metrics to AccessControlledHandler

vancexu:
**What changed?**
Add metrics to AccessControlledHandler

**Why?**
For monitoring authorization status

**How did you test it?**
Unit test

**Potential risks**
No risk

### DIFF
--- a/service/frontend/accessControlledHandler.go
+++ b/service/frontend/accessControlledHandler.go
@@ -23,16 +23,14 @@ package frontend
 import (
 	"context"
 
-	"github.com/uber/cadence/common/authorization"
-
 	"github.com/uber/cadence/.gen/go/cadence/workflowserviceserver"
 	"github.com/uber/cadence/.gen/go/health"
 	"github.com/uber/cadence/.gen/go/health/metaserver"
 	"github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common/authorization"
+	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/resource"
 )
-
-// TODO(vancexu): add metrics
 
 var errUnauthorized = &shared.BadRequestError{Message: "Request unauthorized."}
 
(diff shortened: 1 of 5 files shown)


service/frontend/accessControlledHandler.go
```diff
@@ -121,11 +121,15 @@ func (a *AccessControlledWorkflowHandler) DeprecateDomain(
 	request *shared.DeprecateDomainRequest,
 ) error {
 
+	scope := a.GetResource().GetMetricsClient().
```
andrewjdawson2016: Can we use `getMetricsScopeWithDomain` here?

service/frontend/accessControlledHandler.go
```diff
@@ -724,10 +780,34 @@ func (a *AccessControlledWorkflowHandler) UpdateDomain(
...
-	return result.Decision == authorization.DecisionAllow, nil
+	isAuth := result.Decision == authorization.DecisionAllow
+	if !isAuth {
+		scope.IncCounter(metrics.CadenceErrUnauthorizedCounter)
+	}
+	return isAuth, nil
+}
+
+// getMetricsScopeWithDomain return metrics scope with domain tag
+func (a *AccessControlledWorkflowHandler) getMetricsScopeWithDomain(
```
andrewjdawson2016: How about just making the following method a utility method that takes in a metrics client instead of duplicating the code 

```
// startRequestProfileWithDomain initiates recording of request metrics and returns a domain tagged scope
func (wh *WorkflowHandler) startRequestProfileWithDomain(scope int, d domainGetter) (metrics.Scope, metrics.Stopwatch) {
	var metricsScope metrics.Scope
	if d != nil {
		metricsScope = wh.GetMetricsClient().Scope(scope).Tagged(metrics.DomainTag(d.GetDomain()))
	} else {
		metricsScope = wh.GetMetricsClient().Scope(scope).Tagged(metrics.DomainUnknownTag())
	}
	sw := metricsScope.StartTimer(metrics.CadenceLatency)
	metricsScope.IncCounter(metrics.CadenceRequests)
	return metricsScope, sw
}
```

service/frontend/accessControlledHandler.go
vancexu: Request in XXXDomain APIs are not domainGetter, so hard to reuse. But I make it a function to some some lines.

service/frontend/accessControlledHandler.go
vancexu: updated.

PR #3146, opened 2020-03-27
Wire up priority task processor implementation

yycptt:
**What changed?**
Wire up implementations for priority task processor

**Why?**
Priority task processor is part of the multi-tenancy effort for throttling task processing for bursty domains

**How did you test it?**
Tested locally with feature flag turned both on and off through unit and integration test. 

**Potential risks**
No risk when the feature flag is turned off.
If the flag is turned on, may impact performance.


common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -252,3 +265,43 @@ func (w *weightedRoundRobinTaskSchedulerImpl) notifyDispatcher() {
...
+		}
+	}
+}
+
+func convertWeightsFromDynamicConfig(
+	weightsFromDC map[string]interface{},
+) (map[int]int, error) {
+	weights := make(map[int]int)
+	for key, value := range weightsFromDC {
+		priority, err := strconv.Atoi(key)
```
vancexu: Maybe add TrimSpace to be defensive?

common/task/weightedRoundRobinTaskScheduler.go
```diff
@@ -252,3 +265,43 @@ func (w *weightedRoundRobinTaskSchedulerImpl) notifyDispatcher() {
...
+				w.weights.Store(weights)
+			}
+		case <-w.shutdownCh:
+			ticker.Stop()
+			return
+		}
+	}
+}
+
+func convertWeightsFromDynamicConfig(
```
vancexu: It looks like a misconfig will cause task processing errors. 
Shall we eat the convertConfig error, and only log the error and return default config?

service/history/handler.go
```diff
@@ -121,10 +123,52 @@ func (h *Handler) Start() {
...
+			h.GetClusterMetadata().GetCurrentClusterName(),
+			h.GetDomainCache(),
+			h.GetLogger(),
+			h.GetMetricsClient(),
+			h.config,
+		)
+		queueTaskProcessorOptions := &queueTaskProcessorOptions{}
+		switch schedulerType := task.SchedulerType(h.config.TaskSchedulerType()); schedulerType {
+		case task.SchedulerTypeFIFO:
+			queueTaskProcessorOptions.schedulerType = schedulerType
```
vancexu: move out common code

service/history/queueProcessor.go
```diff
@@ -286,6 +284,31 @@ func (p *queueProcessorBase) processBatch() {
 	return
 }
 
+func (p *queueProcessorBase) submitTask(
```
vancexu: nit: add comment on return value, or change it.
I think when people read code like `submitTask() bool`, people usually thought it mean submit success (true) or fail (false). But this one is a little bit surprising.

service/history/timerQueueProcessorBase.go
```diff
@@ -355,6 +355,31 @@ func (t *timerQueueProcessorBase) readAndFanoutTimerTasks() (*persistence.TimerT
 	return nil, nil
 }
 
+func (t *timerQueueProcessorBase) submitTask(
```
vancexu: ditto

common/task/weightedRoundRobinTaskScheduler.go
yycptt: If the weight is invalid during server start up, then the start up logic will fail and server will panic.
If server has already been started and the weight is updated to an invalid value, then then update weight logic in the scheduler will be a no-op (keep using the existing weight) and print an error log. 
Does the above behavior make sense to you? Thoughts?

service/history/queueProcessor.go
yycptt: Makes sense. Reverted the meaning for the return value. Also reverted the return value from `p.taskProcessor.addTask()`

service/history/handler.go
yycptt: Done.

service/history/timerQueueProcessorBase.go
yycptt: Done.

common/task/weightedRoundRobinTaskScheduler.go
vancexu: ok, make sense.

PR #3147, opened 2020-03-27
Add additional logging and metrics around size limits

andrewjdawson2016:
Add additional logging for blob size errors. Also add tags to metrics. This will enable us to identify customers which are in violation.


common/util.go
```diff
@@ -436,13 +436,28 @@ func CreateHistoryStartWorkflowRequest(
...
+func CheckEventBlobSizeLimit(
+	actualSize int,
+	warnLimit int,
+	errorLimit int,
+	domainID string,
+	workflowID string,
+	runID string,
+	scope metrics.Scope,
+	logger log.Logger,
+	blobSizeViolationOperationTag tag.Tag,
```
vancexu: Maybe passed in `operationName` instead of tag?

PR #3154, opened 2020-03-30
Add admin failover for managed domains

vancexu:
**What changed?**
Add admin CLI command to failover managed domains

**Why?**
Currently domain failover is by design managed by cadence user. Some users dislike to failover themselves because they don't know how or when to failover. 
By adding this managed failover support, cadence now have the ability to failover users domain on their behalf. 

**How did you test it?**
Unit test.
Local test.
Will do staging test before release.

**Potential risks**
No risk.

### DIFF
--- a/tools/cli/admin.go
+++ b/tools/cli/admin.go
@@ -811,6 +811,20 @@ func newAdminClusterCommands() []cli.Command {
 				AdminDescribeCluster(c)
 			},
 		},
+		{
+			Name:    "failover",
+			Aliases: []string{"fo"},
+			Usage:   "Failover domains with domain data IsManagedByCadence=true to target cluster",
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  FlagActiveClusterNameWithAlias,
+					Usage: "Target active cluster name",
+				},
+			},
+			Action: func(c *cli.Context) {
+				newDomainCLI(c, false).FailoverDomains(c)
+			},
+		},
 	}
 }
 
--- a/tools/cli/defs.go
+++ b/tools/cli/defs.go
@@ -65,6 +65,8 @@ const (
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
 
 	searchAttrInputSeparator = "|"
+
+	managedFailoverDomainData = "IsManagedByCadence"
 )
 
 var envKeysForUserName = []string{
--- a/tools/cli/defs.go
+++ b/tools/cli/defs.go
@@ -65,8 +65,6 @@ const (
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
 
 	searchAttrInputSeparator = "|"
-
-	managedFailoverDomainData = "IsManagedByCadence"
 )
 
 var envKeysForUserName = []string{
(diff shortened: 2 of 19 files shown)


tools/cli/domainCommands.go
```diff
@@ -282,6 +283,72 @@ func (d *domainCLIImpl) UpdateDomain(c *cli.Context) {
...
+func (d *domainCLIImpl) getAllDomains() []*shared.DescribeDomainResponse {
+	var res []*shared.DescribeDomainResponse
+	pagesize := int32(200)
+	var token []byte
+	for more := true; more; more = len(token) > 0 {
+		listRequest := &shared.ListDomainsRequest{
+			PageSize:      common.Int32Ptr(pagesize),
+			NextPageToken: token,
+		}
+		listResp, err := d.listDomains(context.Background(), listRequest)
```
yycptt: use context with some default timeout?

tools/cli/defs.go
```diff
@@ -65,6 +65,8 @@ const (
 	showErrorStackEnv    = `CADENCE_CLI_SHOW_STACKS`
 
 	searchAttrInputSeparator = "|"
+
+	managedFailoverDomainData = "IsManagedByCadence"
```
yycptt: This const probably should be defined by the server (in the common package for example) not by the CLI. CLI can then import and use this constant.

tools/cli/domainCommands.go
```diff
@@ -282,6 +283,72 @@ func (d *domainCLIImpl) UpdateDomain(c *cli.Context) {
...
+	targetCluster := getRequiredOption(c, FlagActiveClusterName)
+
+	domains := d.getAllDomains()
+	shouldFailover := func(domain *shared.DescribeDomainResponse) bool {
+		isDomainNotActiveInTargetCluster := domain.ReplicationConfiguration.GetActiveClusterName() != targetCluster
+		return isDomainNotActiveInTargetCluster && isDomainFailoverManagedByCadence(domain)
+	}
+	var succeedDomains []string
+	var failedDomains []string
+	for _, domain := range domains {
```
yycptt: This seems to be a very dangerous operation. How about ask for confirmation from operator before proceeding. For example, print the list of domains that are about to failover so that the operator can at least do some verification?

tools/cli/admin.go
```diff
@@ -811,6 +811,20 @@ func newAdminClusterCommands() []cli.Command {
 				AdminDescribeCluster(c)
 			},
 		},
+		{
+			Name:    "failover",
+			Aliases: []string{"fo"},
+			Usage:   "Failover domains with domain data IsManagedByCadence=true to target cluster",
+			Flags: []cli.Flag{
```
yycptt: It seems like this command should only be executed by cadence team. Should the command also takes in a security token and verify its content before actually execution?

Just an idea to explore. I understand this is not easy as the underlying domain failover command doesn't require security token.

vancexu: Good point. But security token on open sourced repo is merely support on server API. 
Internally we will be protected by the auth feature.

tools/cli/defs.go
vancexu: updated.

tools/cli/domainCommands.go
vancexu: updated. add a simple prompt

tools/cli/domainCommands.go
vancexu: updated

PR #3155, opened 2020-03-30
matching: per task list metrics

venkat1109:
**What changed?**
As of today, cadence-matching doesn't emit metrics with taskList name as one of the tags. But this is becoming a necessity operationally to get visibility into things like tasks_in, tasks_out, tasks_buffered at task list level. There is some visibility today from the client side SDK, but that doesn't always provide a complete picture. 

In order to enable taskList specific metrics, the tags {  'domain': 'domainName', 'taskList: taskListName' } has to be associated along with the metric name. Once we do this, we get visibility at task list level. However, we don't want to lose visibility into global aggregated view across all domain / all task-lists. Not all metrics servers can support aggregation across thousands of time-series data at the time of querying. To avoid having to do aggregation at the query time, this patch adds support for dual emitting metrics once for per_tasklist and another for aggregate. This is done in a backwards compatible manner so that existing dashboards / metrics will not be broken.

To achieve this, support for a new "metricRollupName" is added to the metricScope class. The idea behind this is the following - let's say, matching has a per-tasklist metric called "sync_matches_per_tl".
This metric also is needed at an aggregate level. So, when defining metric constants, we specify metricRollupName to be "sync_matches". Now, the scope will emit the metric twice - once with "sync_matches_per_tl" with tags "domain: and tasklist:" and another with metric name "sync_matches" without these tags.

**Why?**
To enable per task list metrics inside of cadence-matching.

**How did you test it?**
Tested on localhost (using a localhost m3 docker as metrics server)

**Potential risks**
If there is a bug or typo in this patch, existing metrics and dashboards can be broken for cadence-matching.

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -35,9 +35,8 @@ const (
 	activityType  = "activityType"
 	decisionType  = "decisionType"
 
-	domainAllValue   = "all"
-	taskListAllValue = "all"
-	unknownValue     = "_unknown_"
+	domainAllValue = "all"
+	unknownValue   = "_unknown_"
 )
 
 // Tag is an interface to define metrics tags
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -35,8 +35,9 @@ const (
 	activityType  = "activityType"
 	decisionType  = "decisionType"
 
-	domainAllValue = "all"
-	unknownValue   = "_unknown_"
+	domainAllValue   = "all"
+	taskListAllValue = "all"
+	unknownValue     = "_unknown_"
 )
 
 // Tag is an interface to define metrics tags
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -53,6 +53,8 @@ type (
 
 	domainUnknownTag struct{}
 
+	taskListUnknownTag struct{}
+
 	instanceTag struct {
 		value string
 	}
@@ -104,6 +106,21 @@ func DomainUnknownTag() Tag {
 }
 
 // Key returns the key of the domain unknown tag
+func (d taskListUnknownTag) Key() string {
+	return domain
+}
+
+// Value returns the value of the domain unknown tag
+func (d taskListUnknownTag) Value() string {
+	return unknownValue
+}
+
+// TaskListUnknownTag returns a new tasklist:unknown tag-value
+func TaskListUnknownTag() Tag {
+	return taskListUnknownTag{}
+}
+
+// Key returns the key of the domain unknown tag
 func (d domainUnknownTag) Key() string {
 	return domain
 }
@@ -151,7 +168,7 @@ func TaskListTag(value string) Tag {
 	if len(value) == 0 {
 		value = unknownValue
 	}
-	return taskListTag{value}
+	return taskListTag{sanitizer.Value(value)}
 }
 
 // Key returns the key of the task list tag
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -211,17 +211,17 @@ func (e *matchingEngineImpl) removeTaskListManager(id *taskListID) {
 }
 
 // AddDecisionTask either delivers task directly to waiting poller or save it into task list persistence.
-func (e *matchingEngineImpl) AddDecisionTask(ctx context.Context, addRequest *m.AddDecisionTaskRequest) (bool, error) {
-	domainID := addRequest.GetDomainUUID()
-	taskListName := addRequest.TaskList.GetName()
-	taskListKind := common.TaskListKindPtr(addRequest.TaskList.GetKind())
+func (e *matchingEngineImpl) AddDecisionTask(hCtx *handlerContext, request *m.AddDecisionTaskRequest) (bool, error) {
+	domainID := request.GetDomainUUID()
+	taskListName := request.TaskList.GetName()
+	taskListKind := common.TaskListKindPtr(request.TaskList.GetKind())
 
 	e.logger.Debug(
 		fmt.Sprintf("Received AddDecisionTask for taskList=%v, WorkflowID=%v, RunID=%v, ScheduleToStartTimeout=%v",
-			addRequest.TaskList.GetName(),
-			addRequest.Execution.GetWorkflowId(),
-			addRequest.Execution.GetRunId(),
-			addRequest.GetScheduleToStartTimeoutSeconds()))
+			request.TaskList.GetName(),
+			request.Execution.GetWorkflowId(),
+			request.Execution.GetRunId(),
+			request.GetScheduleToStartTimeoutSeconds()))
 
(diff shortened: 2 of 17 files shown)


yux0:
LGTM. One question is how to deal with the task list with the same name? Does it mean that it requires domain tag when emitting tasklist tag?

venkat1109:
> LGTM. One question is how to deal with the task list with the same name? Does it mean that it requires domain tag when emitting tasklist tag?

Yes, tasklist always need to be qualified with domainName. Otherwise, its not unique.

samarabbas:
This would significantly impact the number of metric emitted by Cadence server as number of tasklist could be huge.  I would highly recommend to check first with metric backend team before enabling this change.  Or at the bare minimum have a dynamic config knob to disable it incase it cause issues.
Another thing is, it's cleaner to use a different metric name instead of a tag "all" for aggregated metric.

venkat1109:
> This would significantly impact the number of metric emitted by Cadence server as number of tasklist could be huge. I would highly recommend to check first with metric backend team before enabling this change. 

The change in number of metrics emitted by cadence-matching will be 2x ( not unbounded ). For every metric we emit in cadence-matching today, we will emit two metrics going forward - one with tasklist=all and another for tasklist=foo. As for the metrics backend scaling, metrics backend can scale much higher than this scale today at UBER. The limitation that internal UBER has is around "how many timeseries data can be aggregated at query time (i.e. at graphing time)" - this number is 50 to 100k timeseries today. WIth the proposal in this diff, we are always looking at "exactly one time series" - there is no aggregation involved at query time. So, this is something that metrics system is designed to do within UBER. Also for reference, we already do this kind of dual emitting for frontend (domain level metrics and aggregate metrics). 

Another option is to only emit task list specific metric and not emit the aggregate metric ("all" metric). Internal UBER backend has a way to specify a config out of band to do aggregations. This will solve need within UBER - but without an aggregate emitted within code, OSS customers will get impacted (if their metrics backend doesn't scale).

>  Or at the bare minimum have a dynamic config knob to disable it incase it cause issues.
This is a good idea. I will add this when I actually send out a diff for emitting metrics within matching service. The config knob doesn't quite belong in this diff - because there are places where we emit task list specific metric even today - I don't want to have one kill switch for the whole of cadence-server.

> it's cleaner to use a different metric name instead of a tag "all" for aggregated metric.
Yes and no. Yes, because its cleaner. No, because using a different metric name would clutter the code all over the place. Because everywhere we emit metrics, we will need two lines, one for emitCounter and one for emitCounterAggregate (with a different friendly name). 

Another option is to not use "all" but auto suffix something like ".aggregated" to metric name everytime there is a task list specific metric emitted. But this would involve string concat for each and every metric that gets emitted.

I went with the "all" option because it is the cheapest option that solves the problem at hand. I can change this diff to suffix ".aggregated" instead and benchmark it. If it causes issues, will revert back to "all".

venkat1109:
One more point reg. scale - client SDKs have been emitting task list specific metrics for a while now within UBER with no issues

samarabbas:
> The change in number of metrics emitted by cadence-matching will be 2x ( not unbounded ). For every metric we emit in cadence-matching today, we will emit two metrics going forward - one with tasklist=all and another for tasklist=foo. As for the metrics backend scaling, metrics backend can scale much higher than this scale today at UBER. The limitation that internal UBER has is around "how many timeseries data can be aggregated at query time (i.e. at graphing time)" - this number is 50 to 100k timeseries today. WIth the proposal in this diff, we are always looking at "exactly one time series" - there is no aggregation involved at query time. So, this is something that metrics system is designed to do within UBER. Also for reference, we already do this kind of dual emitting for frontend (domain level metrics and aggregate metrics).

I'm less concerned about dual emit and more concerned about the explosion in cardinality.  Emitting this metric for sticky tasklist seems like an overkill.  Most of the metric backend usually do not prefer high cardinality in tag values even if they are never queries with an aggregation query.  I'm definitely open to emitting tasklist tag but we should have this as an optional feature which can be disabled through config knob.

> Another option is to not use "all" but auto suffix something like ".aggregated" to metric name everytime there is a task list specific metric emitted. But this would involve string concat for each and every metric that gets emitted.

Why can't we just dual emit with different metric names?

venkat1109:
> I'm less concerned about dual emit and more concerned about the explosion in cardinality. Emitting this metric for sticky tasklist seems like an overkill. Most of the metric backend usually do not prefer high cardinality in tag values even if they are never queries with an aggregation query. I'm definitely open to emitting tasklist tag but we should have this as an optional feature which can be disabled through config knob.

I think there is some misunderstanding. The intention here is not to add task list specific metrics for sticky. Just for the user specified task list and the associated partitions. Talked offline and clarified the intent. 

> Why can't we just dual emit with different metric names?
Like I said before, we can do it but will cause code clutter - every metric needs to be emitted twice now.

Example:

```
func AddDecisionTask() {
    scope.IncCounter(aggregateCounterName)
    taskListScope.IncCounter(taskListSpecificCounterName)
}
```

It just becomes a maintenance nightmare if we go this route. I would prefer suffixing with ".aggregated" instead of having this all over the code.

service/matching/matchingEngine.go
```diff
@@ -211,17 +211,17 @@ func (e *matchingEngineImpl) removeTaskListManager(id *taskListID) {
 }
 
 // AddDecisionTask either delivers task directly to waiting poller or save it into task list persistence.
-func (e *matchingEngineImpl) AddDecisionTask(ctx context.Context, addRequest *m.AddDecisionTaskRequest) (bool, error) {
-	domainID := addRequest.GetDomainUUID()
-	taskListName := addRequest.TaskList.GetName()
-	taskListKind := common.TaskListKindPtr(addRequest.TaskList.GetKind())
+func (e *matchingEngineImpl) AddDecisionTask(hCtx *handlerContext, request *m.AddDecisionTaskRequest) (bool, error) {
```
meiliang86, line 214: Seems like only the regular context is used in these method. Do you foresee that handlerContext can be used in future, maybe for us to add more metrics inside engine?

common/metrics/tags.go
```diff
@@ -35,8 +35,9 @@ const (
 	activityType  = "activityType"
 	decisionType  = "decisionType"
 
-	domainAllValue = "all"
-	unknownValue   = "_unknown_"
+	domainAllValue   = "all"
+	taskListAllValue = "all"
```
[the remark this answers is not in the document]
venkat1109: Yup. Removed it.

service/matching/matchingEngine.go
venkat1109: I assume you are talking about the taskListManager.AddTask() calls ? The reason these don't take handlerContext is because the lifetime of taskListManager is beyond the lifetime of request. For this reason, they don't use the metric scope in the handlerContext. They have their own scope (which is per_task_list )

PR #3159, opened 2020-04-01
Improve filestore query parser for treating more values as valid close status

yycptt:
**What changed?**
Improve the parse function for close status in visibility archival filestore query parse to accept more values.

**Why?**
Previously this parse function doesn't handle terminated status. It also requires the close status to be a string (not a number), which doesn't match our internal implementation.

**How did you test it?**
Unit test

**Potential risks**
N/A, filestore is only used in local development, not in production.


just-at-uber:
Tested locally and works when requesting for a TimedOut workflow with CloseStatus = 5

PR #3160, opened 2020-04-01
Add back pressure logic for task loading

yycptt:
**What changed?**
* Skip task loading when there are too many pending tasks in re-dispatch queue. 
Note that task loading is not completely blocked. There's a timer for `maxPollInterval`, when that timer fires, task queue will still load some tasks from persistence. This is timer has a quite large value (1 min) so it shouldn't be a problem.
* Emit metrics for re-dispatch queue size

**Why?**
For multi-tenant task processing, the queue processor will first load tasks from persistence and try to submit the task to the task processor. For bursty domain, the submit will fail as the task queue for that domain (priority, actually) is full and the task will be buffered in the re-dispatch queue. It's possible that one domain keeps generating large number of tasks which results in the re-dispatch queue holding large number of tasks. This may cause the host to go out of memory. 

This PR adds some back pressure logic to block task loading when the re-dispatch queue contains too many tasks. Basically when the number of tasks in the re-dispatch queue reaches a threshold, only do re-dispatch and stop loading new tasks. (Of course the loading is not completed blocked, new tasks will be eventually loaded after some time period).

Note that the value for the threshold represents a tradeoff between multi-tenancy and system protection.
- Large threshold means more tasks can be buffered so it's better at handling busty domains while not affecting other domains.
- Small threshold means the queue processor will use less resource on the host and provides better system protection.

**How did you test it?**
Unit test, integration test, local canary.

**Potential risks**
In the worst case, new tasks can't be loaded from persistence and processing is blocked, if the feature flag is enabled.

### DIFF
--- a/service/history/queueProcessor.go
+++ b/service/history/queueProcessor.go
@@ -51,6 +51,7 @@ type (
 		MaxRetryCount                       dynamicconfig.IntPropertyFn
 		RedispatchInterval                  dynamicconfig.DurationPropertyFn
 		RedispatchIntervalJitterCoefficient dynamicconfig.FloatPropertyFn
+		MaxRedispatchQueueSize              dynamicconfig.IntPropertyFn
 		EnablePriorityTaskProcessor         dynamicconfig.BoolPropertyFn
 		MetricScope                         int
 	}
@@ -64,7 +65,7 @@ type (
 		options              *QueueProcessorOptions
 		processor            processor
 		logger               log.Logger
-		metricsClient        metrics.Client
+		metricsScope         metrics.Scope
 		rateLimiter          quotas.Limiter // Read rate limiter
 		ackMgr               queueAckMgr
 		taskProcessor        *taskProcessor // TODO: deprecate task processor, in favor of queueTaskProcessor
@@ -98,6 +99,7 @@ func newQueueProcessorBase(
 	historyCache *historyCache,
 	queueTaskInitializer queueTaskInitializer,
 	logger log.Logger,
+	metricsScope metrics.Scope,
 ) *queueProcessorBase {
 
 	var taskProcessor *taskProcessor
@@ -123,8 +125,8 @@ func newQueueProcessorBase(
 		status:               common.DaemonStatusInitialized,
 		notifyCh:             make(chan struct{}, 1),
 		shutdownCh:           make(chan struct{}),
-		metricsClient:        shard.GetMetricsClient(),
 		logger:               logger,
+		metricsScope:         metricsScope,
 		ackMgr:               queueAckMgr,
 		lastPollTime:         time.Time{},
 		taskProcessor:        taskProcessor,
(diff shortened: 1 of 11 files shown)


service/history/queueProcessor.go
```diff
@@ -210,7 +212,15 @@ processorPumpLoop:
 			// use a separate gorouting since the caller hold the shutdownWG
 			go p.Stop()
 		case <-p.notifyCh:
-			p.processBatch()
+			if !p.isPriorityTaskProcessorEnabled() || p.redispatchQueue.Len() <= p.options.MaxRedispatchQueueSize() {
```
yux0, line 215: What is isPriorityTaskProcessorEnabled for?

yycptt: This is the feature flag for controlling whether the new task processor based on priority will be used or not. If it's not enabled, then basically there's no re-dispatch queue and task loading won't be skipped.

service/history/queueProcessor.go
```diff
@@ -343,9 +354,11 @@ func redispatchQueueTasks(
 	redispatchQueue collection.Queue,
 	queueTaskProcessor queueTaskProcessor,
 	logger log.Logger,
+	metricsScope metrics.Scope,
 	shutdownCh <-chan struct{},
 ) {
 	queueLength := redispatchQueue.Len()
+	metricsScope.RecordTimer(metrics.TaskRedispatchQueuePendingTasksTimer, time.Duration(queueLength))
```
venkat1109, line 361: consider using a per-shard gauge metric instead of a timer. Per shard metrics are ok as long as no rollups are needed at visualization time (which we won't need in this case).

service/history/service.go
```diff
@@ -309,6 +314,7 @@ func NewConfig(dc *dynamicconfig.Collection, numberOfShards int, storeType strin
 		ReplicatorProcessorUpdateAckIntervalJitterCoefficient:  dc.GetFloat64Property(dynamicconfig.ReplicatorProcessorUpdateAckIntervalJitterCoefficient, 0.15),
 		ReplicatorProcessorRedispatchInterval:                  dc.GetDurationProperty(dynamicconfig.ReplicatorProcessorRedispatchInterval, 5*time.Second),
 		ReplicatorProcessorRedispatchIntervalJitterCoefficient: dc.GetFloat64Property(dynamicconfig.ReplicatorProcessorRedispatchIntervalJitterCoefficient, 0.15),
+		ReplicatorProcessorMaxRedispatchQueueSize:              dc.GetIntProperty(dynamicconfig.ReplicatorProcessorMaxRedispatchQueueSize, 10000),
```
venkat1109, line 317: How much memory does a single task take ? Are all tasks equal ? If not, consider using a size based limit instead of a count based limit. Same comment for other queue processors.

yycptt: A single task takes less than 1KB (actually much smaller, but we can use this as an upper bound). All tasks will have similar size as they only contains some metadata information like domainID, workflowID, timestamp, task type, task state etc. So I think count based limit is fine here.

The number 10k here is a guess based on our current number of pending tasks number in staging/prod. I will need to tune this number when doing testing and after deployment.

PR #3161, opened 2020-04-02
Add global ratelimiter for frontend API

mkolodezny:
**What changed?**

Calculate average quota for current number of nodes in the ring. And use it for frontend rate limiter per domain.

**Why?**

Cadence is almost always Cassandra-bound, since we can scale up Cadence nodes on demand much faster than adding nodes to our CaaS cluster. It was also observed that Cassandra’s performance breaks down if Cadence is pushing more load than the CaaS can handle. 

**How did you test it?**

unit tests
ran test workflows locally.

**Potential risks**

Rate limiting to the frontend API will be under(over)limited.

### DIFF
--- a/service/frontend/service.go
+++ b/service/frontend/service.go
@@ -55,7 +55,7 @@ type Config struct {
 	ESIndexMaxResultWindow          dynamicconfig.IntPropertyFn
 	HistoryMaxPageSize              dynamicconfig.IntPropertyFnWithDomainFilter
 	RPS                             dynamicconfig.IntPropertyFn
-	DomainRPS                       dynamicconfig.IntPropertyFnWithDomainFilter
+	MaxDomainRPSPerInstance         dynamicconfig.IntPropertyFnWithDomainFilter
 	GlobalDomainRPS                 dynamicconfig.IntPropertyFnWithDomainFilter
 	MaxIDLengthLimit                dynamicconfig.IntPropertyFn
 	EnableClientVersionCheck        dynamicconfig.BoolPropertyFn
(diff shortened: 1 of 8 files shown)


service/frontend/workflowHandler.go
```diff
@@ -160,6 +160,13 @@ func NewWorkflowHandler(
 				return float64(config.RPS())
 			},
 			func(domain string) float64 {
+				if monitor := resource.GetMembershipMonitor(); monitor != nil {
+					ringSize, err := monitor.GetMemberCountWithRole(common.FrontendServiceName)
+					if err == nil && ringSize > 0 {
```
meiliang86, line 165: Is this always the case? If true we should consider removing the domainRPS logic completely.

mkolodezny:
> We have also talked about having a similar global rate limiter for cassandra in history. Will it be in a separate PR?

Yes, the cassandra part will be in a separate PR for safety sake.

service/frontend/service.go
```diff
@@ -108,6 +109,7 @@ func NewConfig(dc *dynamicconfig.Collection, numHistoryShards int, enableReadFro
 		HistoryMaxPageSize:                  dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendHistoryMaxPageSize, common.GetHistoryMaxPageSize),
 		RPS:                                 dc.GetIntProperty(dynamicconfig.FrontendRPS, 1200),
 		DomainRPS:                           dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendDomainRPS, 1200),
+		GlobalDomainRPS:                     dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendGlobalDomainRPS, 7000),
```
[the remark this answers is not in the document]
mkolodezny: lets set it to be same as FrontendRPS default value - because no domain can go over FrontendRPS anyways.

service/frontend/service.go
```diff
@@ -107,7 +108,8 @@ func NewConfig(dc *dynamicconfig.Collection, numHistoryShards int, enableReadFro
 		ESIndexMaxResultWindow:              dc.GetIntProperty(dynamicconfig.FrontendESIndexMaxResultWindow, 10000),
 		HistoryMaxPageSize:                  dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendHistoryMaxPageSize, common.GetHistoryMaxPageSize),
 		RPS:                                 dc.GetIntProperty(dynamicconfig.FrontendRPS, 1200),
-		DomainRPS:                           dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendDomainRPS, 1200),
+		MaxDomainRPSPerInstance:             dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendMaxDomainRPSPerInstance, 1200),
+		GlobalDomainRPS:                     dc.GetIntPropertyFilteredByDomain(dynamicconfig.FrontendGlobalDomainRPS, 1200),
```
venkat1109, line 112: A good default value for GlobalDomainRPS is actually 0 or infinite. Someone spinning up cluster with defaults in open source community would run into odd rate limit of 1200 per cluster even though they had multiple frontends. It does complicate your functor that returns rps value - IMO I think its worth it.

mkolodezny: Agree, set it to zero and return MaxDomainRPSPerInstance if in that case.

PR #3163, opened 2020-04-02
Use reader in admin kafka with 3mb buffer size

yux0:
**What changed?**
Kafka admin CLI on merge DLQ

**Why?**
Some replication message is too large to be parsed in admin CLI

**How did you test it?**
Local and test environment

**Potential risks**
In test worst case, merge DLQ Admin CLI fail


tools/cli/adminKafkaCommands.go
```diff
@@ -840,28 +840,32 @@ func parseReplicationTask(in string) (tasks []*replicator.ReplicationTask, err e
 	}
 	defer file.Close()
 
-	scanner := bufio.NewScanner(file)
+	reader := bufio.NewReaderSize(file, 1024*1024*500)
```
yycptt, line 843: 500MB? I think the number is the size for each line in the file right, not the entire file?

yux0: Consider we have 2mb batch limit. Update to 3MB

PR #3165, opened 2020-04-03
Add database admin scan command

andrewjdawson2016:
Added admin db scan command. And ListConcreteExecutions API to persistence. 

This script is useful to detect data inconsistency. 

If the rates used are too high it can interfere with production database access.


common/cassandra/cassandraCluster.go
```diff
@@ -64,9 +64,6 @@ func NewCassandraCluster(cfg config.Cassandra) *gocql.ClusterConfig {
 	if cfg.MaxConns > 0 {
 		cluster.NumConns = cfg.MaxConns
 	}
-
-	cluster.PoolConfig.HostSelectionPolicy = gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy())
```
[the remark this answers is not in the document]
andrewjdawson2016: Is this used in other places besides CLI: Yes
Reason for delete: https://gitmemory.com/issue/gocql/gocql/1274/471217774
Confirmed locally that deleting fixes issue.
Checked with gocql team to confirm that its safe to delete this and yes said yes. Without setting it will will use round robin anyways.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -2033,6 +2038,41 @@ func (d *cassandraPersistence) GetCurrentExecution(request *p.GetCurrentExecutio
 	}, nil
 }
 
+func (d *cassandraPersistence) ListConcreteExecutions(
```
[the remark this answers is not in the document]
andrewjdawson2016: ListWorkflowExecutions already used for visibility API

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/uber/cadence/.gen/go/shared"
+	"github.com/uber/cadence/common/codec"
+	"github.com/uber/cadence/common/log/loggerimpl"
+	"github.com/uber/cadence/common/persistence"
+	cassp "github.com/uber/cadence/common/persistence/cassandra"
+)
+
+const (
+	scanWorkerCount            = 40
+	numberOfShards             = 16384
```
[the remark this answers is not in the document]
andrewjdawson2016: good point.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+		DomainID:     info.DomainID,
+		WorkflowID:   info.WorkflowID,
+		RunID:        info.RunID,
+		NextEventID:  info.NextEventID,
+		TreeID:       treeID,
+		BranchID:     branchID,
+		CloseStatus:  info.CloseStatus,
+		ScanMetadata: metadata,
+	}
+	data, err := json.Marshal(exec)
```
yycptt, line 206: Probably use `MarshalIndent` to make it more readable. Same for other write function.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+
+	shardReports := make(chan *ScanShardReport)
+	for i := 0; i < scanWorkerCount; i++ {
+		go func(workerIdx int) {
+			for shardID := 0; shardID < numberOfShards; shardID++ {
+				if shardID%scanWorkerCount == workerIdx {
+					shardReports <- scanShard(session, shardID, scanFiles)
+				}
+			}
+		}(i)
```
yycptt, line 103: Is it possible to stop/terminate the scan after it's started?

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	combinedShardReport := &ScanShardReport{}
+	for i := 0; i < numberOfShards; i++ {
```
yycptt, line 107: Probably print some progress message to std out. It's hard to tell the progress by looking at the progress file (it's too large to read if it contains report for each shard).

common/persistence/persistenceMetricClients.go
```diff
@@ -330,6 +330,20 @@ func (p *workflowExecutionPersistenceClient) GetCurrentExecution(request *GetCur
 	return response, err
 }
 
+func (p *workflowExecutionPersistenceClient) ListConcreteExecutions(request *ListConcreteExecutionsRequest) (*ListConcreteExecutionsResponse, error) {
+	p.metricClient.IncCounter(metrics.PersistenceListConcreteExecutionsScope, metrics.PersistenceRequests)
+
+	sw := p.metricClient.StartTimer(metrics.PersistenceGetCurrentExecutionScope, metrics.PersistenceLatency)
```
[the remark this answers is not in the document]
andrewjdawson2016: Thank you

tools/cli/admin.go
```diff
@@ -910,3 +910,61 @@ func newAdminDLQCommands() []cli.Command {
...
 	}
 }
+
+func newDBCommands() []cli.Command {
+	return []cli.Command{
+		{
+			Name:    "scan",
+			Aliases: []string{"scan"},
+			Usage:   "scan all concrete executions in database and detect corruptions",
+			Flags: []cli.Flag{
```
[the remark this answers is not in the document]
andrewjdawson2016: Good point. I will make this change

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/uber/cadence/common/codec"
+	"github.com/uber/cadence/common/log/loggerimpl"
+	"github.com/uber/cadence/common/persistence"
+	cassp "github.com/uber/cadence/common/persistence/cassandra"
+)
+
+const (
+	scanWorkerCount            = 40
+	numberOfShards             = 16384
+	delayBetweenExecutionPages = 50 * time.Millisecond
```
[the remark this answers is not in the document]
andrewjdawson2016: Will make this change. Will use a dynamic rate limiter which starts at a low rps and over the course of X seconds moves to target rps.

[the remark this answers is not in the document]
andrewjdawson2016: Sounds good - will do.

tools/cli/adminDBCommands.go
andrewjdawson2016: Just kill -9 the process

tools/cli/adminDBCommands.go
andrewjdawson2016: Okay I will think about this in combination with Venkats comment about perShard vs combining progress being confusing.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+					shardReports <- scanShard(session, shardID, scanFiles)
+				}
+			}
+		}(i)
+	}
+
+	combinedShardReport := &ScanShardReport{}
+	for i := 0; i < numberOfShards; i++ {
+		report := <-shardReports
+		combineShardReport(report, combinedShardReport)
```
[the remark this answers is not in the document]
andrewjdawson2016: Seems like this is confusing to multiple people. I will rethink how progress is reported.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,264 @@
...
+		failedToRunCheckFile:  failedToRunCheckFile,
+		startEventCorruptFile: startEventCorruptFile,
+		progressFile:          progressFile,
+	}, deferFn
+}
+
+func combineShardReport(report *ScanShardReport, combined *ScanShardReport) {
+	if combined.ShardID < report.ShardID {
+		combined.ShardID = report.ShardID
+	}
```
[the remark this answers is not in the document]
andrewjdawson2016: Will address.

common/cassandra/cassandraCluster.go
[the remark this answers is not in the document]
andrewjdawson2016: Added it back and then unset it only in CLI

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,354 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/gocql/gocql"
+	"github.com/urfave/cli"
+
+	"github.com/uber/cadence/common"
+
```
yycptt, line 36: nit: remove empty line

tools/cli/flags.go
```diff
@@ -213,6 +215,8 @@ const (
 	FlagMaxMessageCountWithAlias          = FlagMaxMessageCount + ", mmc"
 	FlagLastMessageID                     = "last_message_id"
 	FlagLastMessageIDWithAlias            = FlagLastMessageID + ", lm"
+	FlagGoRoutineCount                    = "go_routine_count"
```
yycptt, line 218: nit: Probably `FlagCurrency`? I don't think we need to reveal the fact that CLI is implemented with Golang.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,354 @@
...
+				branch.GetTreeID(),
+				branch.GetBranchID(),
+				"got not found error",
+				"",
+				execution)
+		} else {
+			report.NumberOfFailedChecks++
+			recordScanFailure(scanFiles.failedToRunCheckFile, "got error from read history other than not exists error", err)
+		}
+	} else if history == nil || len(history.History) == 0 {
```
yycptt, line 223: I guess this "if-else structure" will quickly become unmanageable as we add more checks (e.g. check current mutable state record etc.). 

Probably define some generic interface for those checks and corresponding actions (write a report etc.). Then we use a for loop to check all the cases. With this we can also specify what type of checks are needed when starting the script. For example if we only need to check if history exists, then we can skip the check for current mutable state execution and speed up the script.

We can create an issue for this and do the refactoring when adding more checks later.

andrewjdawson2016: Yup - was working on this as you posted this review. Let me know if you prefer a different structure to what I ended up with.

tools/cli/adminDBCommands.go
```diff
@@ -0,0 +1,631 @@
...
+			}
+			return report
+		}
+		token = resp.NextPageToken
+		for _, e := range resp.ExecutionInfos {
+			if report.Scanned == nil {
+				report.Scanned = &ShardScanReportExecutionsScanned{}
+			}
+			report.Scanned.TotalExecutionsCount++
+			historyVerificationResult, history, historyBranch := verifyHistoryExists(
```
yux0, line 243: Doesn't have to be in this PR. But we can have something like plugin mode. Easy to add new checks and from the CLI, we can specify which check to execute. What do you think?

andrewjdawson2016: yes totally agree I will do this as part of scanner though. CLI can consume this once we have it in workflow.

PR #3167, opened 2020-04-03
Add feature flag to not fail in flight decision

yux0:
**What changed?**
Check if there is buffered event prior to add a decision task fail. 

**Why?**
We fail the decision to make sure the versions of events inside one batch are the same. However, this blocks replicating missing workflow from another cluster.

For example:
1. Event 1 - 10 are generated in cluster A with version X
2. Event 11-20 are generated in cluster B with version X+10 
3. Cluster A gets deconstructed and lost all data
4. Cluster A should recover this workflow from cluster B

**How did you test it?**
local two clusters test

**Potential risks**
In the worst case, the versions inside one batch will be different.


yycptt:
Please add some comments in the code explaining the idea of this pr. I guess we may need to re-visit this solution in the future.

PR #3169, opened 2020-04-06
Add global ratelimiter for persistence

mkolodezny:
**What changed?**

Calculate average quota for current number of nodes in the ring. And use it for persistence API rate limiters of frontend, history, matching and worker.

**Why?**

Follow up for https://github.com/uber/cadence/pull/3161: cadence is almost always Cassandra-bound, since we can scale up Cadence nodes on demand much faster than adding nodes to our CaaS cluster. It was also observed that Cassandra’s performance breaks down if Cadence is pushing more load than the CaaS can handle. 

**How did you test it?**

unit tests
ran test workflows locally.

**Potential risks**

Rate limiting to the persistence API will be under(over)limited.

### DIFF
--- a/common/resource/resourceImpl.go
+++ b/common/resource/resourceImpl.go
@@ -133,6 +133,7 @@ func New(
 	params *service.BootstrapParams,
 	serviceName string,
 	persistenceMaxQPS dynamicconfig.IntPropertyFn,
+	persistenceGlobalMaxQPS dynamicconfig.IntPropertyFn,
 	throttledLoggerMaxRPS dynamicconfig.IntPropertyFn,
 	visibilityManagerInitializer VisibilityManagerInitializer,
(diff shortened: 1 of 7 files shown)


common/resource/resourceImpl.go
```diff
@@ -192,7 +193,16 @@ func New(
 
 	persistenceBean, err := persistenceClient.NewBeanFromFactory(persistenceClient.NewFactory(
 		&params.PersistenceConfig,
-		persistenceMaxQPS,
+		func(...dynamicconfig.FilterOption) int {
+			if persistenceGlobalMaxQPS() > 0 {
```
venkat1109, line 197: Would be awesome if we are able to exercise this code path in integration test. Checkout host/dynamicconfig.go

PR #3171, opened 2020-04-07
Add tasklist to visibility 

mkolodezny:
**What changed?**

Add tasklist to visibility record. 

**Why?**

To provide possibility to search workflows by task list with ES

**How did you test it?**

unit tests
cadence cli 

**Potential risks**

Search queries to ES


yycptt:
I thought the original issue means include task list as one of the default search attributes (just like `BinaryCheckSums` and `CadenceChangeVersion`), not adding a new field to the visibility record.  Am I missing something here?

I am concerned about this because if we add a new field to the visibility record, it will impact visibility archival, especially our internal implementation.

vancexu:
> I thought the original issue means include task list as one of the default search attributes (just like `BinaryCheckSums` and `CadenceChangeVersion`), not adding a new field to the visibility record. Am I missing something here?
> 
> I am concerned about this because if we add a new field to the visibility record, it will impact visibility archival, especially our internal implementation.

The major reason is to also benefit users without ES.
We will need to continue adding this to visibility archival before release.

yycptt:
> > I thought the original issue means include task list as one of the default search attributes (just like `BinaryCheckSums` and `CadenceChangeVersion`), not adding a new field to the visibility record. Am I missing something here?
> > I am concerned about this because if we add a new field to the visibility record, it will impact visibility archival, especially our internal implementation.
> 
> The major reason is to also benefit users without ES.
> We will need to continue adding this to visibility archival before release.

I see. Let's take the discussion offline and probably explore if there's an alternative. Making changes to visibility archival, especially our internal implementation will be a little bit tricky.

common/persistence/elasticsearch/decodeBench_test.go
```diff
@@ -43,6 +43,7 @@ var (
          "WorkflowID": "6bfbc1e5-6ce4-4e22-bbfb-e0faa9a7a604-1-2256",
          "WorkflowType": "TestWorkflowExecute",
  		 "Encoding" : "thriftrw",
+         "TaskList"" : "taskList","
```
vancexu, line 46: typo?

common/persistence/elasticsearch/decodeBench_test.go
```diff
@@ -43,6 +43,7 @@ var (
          "WorkflowID": "6bfbc1e5-6ce4-4e22-bbfb-e0faa9a7a604-1-2256",
          "WorkflowType": "TestWorkflowExecute",
  		 "Encoding" : "thriftrw",
```
vancexu, line 45: nit: indent?

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -910,13 +915,14 @@ func (v *esVisibilityStore) convertSearchResultToVisibilityRecord(hit *elastic.S
 
 func getVisibilityMessage(domainID string, wid, rid string, workflowTypeName string,
 	startTimeUnixNano, executionTimeUnixNano int64, taskID int64, memo []byte, encoding common.EncodingType,
-	searchAttributes map[string][]byte) *indexer.Message {
+	taskList string, searchAttributes map[string][]byte) *indexer.Message {
```
vancexu, line 918: nit: either put taskList to the end, or after workflowTypeName/taskID

yycptt:
> > > I thought the original issue means include task list as one of the default search attributes (just like `BinaryCheckSums` and `CadenceChangeVersion`), not adding a new field to the visibility record. Am I missing something here?
> > > I am concerned about this because if we add a new field to the visibility record, it will impact visibility archival, especially our internal implementation.
> > 
> > 
> > The major reason is to also benefit users without ES.
> > We will need to continue adding this to visibility archival before release.
> 
> I see. Let's take the discussion offline and probably explore if there's an alternative. Making changes to visibility archival, especially our internal implementation will be a little bit tricky.

Synced offline. Will keep the current approach as tasklist is logically part of the workflow definition and should not be part of search attributes. Changes for visibility archival will be made in later PRs.

@mkolodezny Probably update the pr title to avoid any confusion. Probably `Add tasklist to visibility record`.

common/persistence/elasticsearch/decodeBench_test.go
mkolodezny: I do not see this in my editor for some reason.

config/dynamicconfig/development_es.yaml
```diff
@@ -15,6 +15,7 @@ frontend.validSearchAttributes:
       CloseTime: 2
       CloseStatus: 2
       HistoryLength: 2
+      TaskList: 0
```
vancexu, line 18: 1 because its keyword type

PR #3172, opened 2020-04-07
Db scanner additions

andrewjdawson2016:
- Add fileWriter buffer for db admin scanner to cut down on fileIO
- Reduce amount of history read 
- Print pid to stop scan with

### DIFF
--- a/tools/cli/adminDBCommands.go
+++ b/tools/cli/adminDBCommands.go
@@ -69,7 +69,7 @@ const (
 )
 
 const (
-	historyPageSize = 10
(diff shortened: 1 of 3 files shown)


tools/cli/adminDBCommands.go
```diff
@@ -69,7 +69,7 @@ const (
 )
 
 const (
-	historyPageSize = 50
+	historyPageSize = 10
```
yycptt, line 72: Discussed offline. Seems like 1 is enough here.

PR #3174, opened 2020-04-08
Add admin db clean command

andrewjdawson2016:
Added admin db clean command

It is useful to be able to clean up database corruptions


tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+		report.Failure = &ShardCleanReportFailure{
+			Note:    "failed to create execution store",
+			Details: err.Error(),
+		}
+		return report
+	}
+
+	scanner := bufio.NewScanner(shardCorruptedFile)
+	for scanner.Scan() {
+		if report.Handled == nil {
```
andrewjdawson2016, line 187: If we find this is too slow we can create a buffered reader.

tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+		successfullyCleanWriter.Add(&ce)
+		if ce.CorruptedExceptionMetadata.CorruptionType != OpenExecutionInvalidCurrentExecution {
+			deleteCurrentReq := &persistence.DeleteCurrentWorkflowExecutionRequest{
+				DomainID:   ce.DomainID,
+				WorkflowID: ce.WorkflowID,
+				RunID:      ce.RunID,
+			}
+			// deleting current execution is best effort, the success or failure of the cleanup
+			// is determined above based on if the concrete execution could be deleted
+			execStore.DeleteCurrentWorkflowExecution(deleteCurrentReq)
```
andrewjdawson2016, line 224: need to add databasePreCondition call here

tools/cli/adminDBScanCommand.go
```diff
@@ -165,19 +165,18 @@ func AdminDBScan(c *cli.Context) {
 	scaleUpSeconds := c.Int(FlagRPSScaleUpSeconds)
 	scanWorkerCount := c.Int(FlagConcurrency)
 	executionsPageSize := c.Int(FlagPageSize)
-	scanReportRate := c.Int(FlagScanReportRate)
+	scanReportRate := c.Int(FlagReportRate)
 	if numShards < scanWorkerCount {
 		scanWorkerCount = numShards
 	}
 
-	printStopMessage()
```
yycptt, line 173: Why remove this?

tools/cli/adminDBCommandsFileWriter.go
```diff
@@ -29,75 +29,92 @@ import (
...
+	flushThreshold = 50
 )
 
 type (
-	// AdminDBCommandFileWriter is used to batch writes to db scan
-	AdminDBCommandFileWriter interface {
-		AddCorruptedExecution(*CorruptedExecution)
-		AddExecutionCheckFailure(*ExecutionCheckFailure)
+	// AdminDBCorruptedExecutionBufferedWriter is used to buffer writes to file for entities of type CorruptedExecution
+	AdminDBCorruptedExecutionBufferedWriter interface {
```
yycptt, line 37: Why do we need two different interfaces and implementations? Does the following one work?
```
BufferedWriter interface {
        Add(interface{})
        Flush()
}
```

tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+		}
+		return report
+	}
+
+	scanner := bufio.NewScanner(shardCorruptedFile)
+	for scanner.Scan() {
+		if report.Handled == nil {
+			report.Handled = &ShardCleanReportHandled{}
+		}
+		report.Handled.TotalExecutionsCount++
```
yycptt, line 190: move this after L194

tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+			deleteCurrentReq := &persistence.DeleteCurrentWorkflowExecutionRequest{
+				DomainID:   ce.DomainID,
+				WorkflowID: ce.WorkflowID,
+				RunID:      ce.RunID,
+			}
+			// deleting current execution is best effort, the success or failure of the cleanup
+			// is determined above based on if the concrete execution could be deleted
+			execStore.DeleteCurrentWorkflowExecution(deleteCurrentReq)
+		}
+		// TODO: we will want to also cleanup history for corrupted workflows, this will be punted on until this is converted to a workflow
```
yycptt, line 226: Will the scanner workflow also scan history table and delete histories without mutable state record?

tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+	shardCorruptedFile, err := getShardCorruptedFile(inputDirectory, shardID)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			report.Failure = &ShardCleanReportFailure{
+				Note:    "failed to get corruption file",
+				Details: err.Error(),
+			}
+		}
+		return report
+	}
```
yycptt, line 175: Add `defer shardCorruptedFile.Close()` after this `if` statement

tools/cli/adminDBCleanCommand.go
```diff
@@ -0,0 +1,309 @@
...
+		SuccessfullyCleanedFile *os.File
+		FailedCleanedFile       *os.File
+	}
+)
+
+// AdminDBClean is the command to clean up executions
+func AdminDBClean(c *cli.Context) {
+	lowerShardBound := c.Int(FlagLowerShardBound)
+	upperShardBound := c.Int(FlagUpperShardBound)
+	numShards := upperShardBound - lowerShardBound
```
yycptt, line 99: I don't quite understand this shard range. From my understanding, shardID is quite random for different workflows. It's unlikely the shard we want to clean up happens to fall in a small range. 

I can see two use cases here:
1. User want to clean up all the shards. In this case we should ask the user to specify the total number of shards.
2. User want to clean up some specific shard in question. In this case we should accept a list of shardIDs. 

Those two flags should be mutual exclusive, meaning only one of them can be specified at a time.

andrewjdawson2016: The use case is this: scan runs over ranges and produces directories which contain different shards. Now lets say you want to do a a full clean. The way you will do this is give an input directory and a shard range. If you want to clean just one shard then you will do a scan of that one shard and then do a clean of that one shard by giving a range of size 1.

andrewjdawson2016: I skip shards which don't have scan files as part of the clean up. Maybe if I specified this it would be more clear? Thoughts?

tools/cli/adminDBCleanCommand.go
andrewjdawson2016: Good catch thank you

tools/cli/adminDBCleanCommand.go
andrewjdawson2016: Thank you :)

tools/cli/adminDBCleanCommand.go
andrewjdawson2016: Followed up on this offline.

tools/cli/adminDBCommandsFileWriter.go
andrewjdawson2016: Good catch :)

tools/cli/adminDBScanCommand.go
andrewjdawson2016: Because it doesn't produce the correct pid to stop the command.

yycptt: Do you want to print a correct stop message then?

PR #3176, opened 2020-04-08
Update isDomainActive condition to honor pending active statue

yux0:
**What changed?**
Update isDomainActive condition to honor pending active statue

**Why?**
Support graceful failover

**How did you test it?**
unit test

**Potential risks**


common/cache/domainCache.go
```diff
@@ -675,6 +677,7 @@ func (c *domainCache) buildEntryFromRecord(
 	newEntry.failoverVersion = record.FailoverVersion
 	newEntry.isGlobalDomain = record.IsGlobalDomain
 	newEntry.failoverNotificationVersion = record.FailoverNotificationVersion
+	newEntry.failoverEndTime = record.FailoverEndTime
```
vancexu, line 680: record.F or .f?

common/cache/domainCache.go
```diff
@@ -131,6 +131,7 @@ type (
 		failoverVersion             int64
 		isGlobalDomain              bool
 		failoverNotificationVersion int64
+		failoverEndTime             *int64
```
vancexu, line 134: do we need 3 states for failoverEndTime? or make it int64 is enough?

common/cache/domainCache.go
yux0: This GetDomainResponse is defined in a different package. Have to use F

common/cache/domainCache.go
yux0: Yes. I don't want to use a special value to indicate the end time is not set. Just use a pointer here.

PR #3179, opened 2020-04-08
Add execution per shard stats

andrewjdawson2016:


tools/cli/adminDBScanCommand.go
```diff
@@ -739,6 +747,15 @@ func includeShardInProgressReport(report *ShardScanReport, progressReport *Progr
...
 		progressReport.CorruptionTypeBreakdown.TotalInvalidFirstEvent += report.Scanned.CorruptionTypeBreakdown.TotalInvalidFirstEvent
+		if progressReport.ShardExecutionCountsDistribution.MinExecutions == nil ||
+			*progressReport.ShardExecutionCountsDistribution.MinExecutions > report.Scanned.TotalExecutionsCount {
+			progressReport.ShardExecutionCountsDistribution.MinExecutions = &report.Scanned.TotalExecutionsCount
+		}
+		if progressReport.ShardExecutionCountsDistribution.MaxExecutions == nil ||
+			*progressReport.ShardExecutionCountsDistribution.MaxExecutions < report.Scanned.TotalExecutionsCount {
+			progressReport.ShardExecutionCountsDistribution.MaxExecutions = &report.Scanned.TotalExecutionsCount
+		}
+		progressReport.ShardExecutionCountsDistribution.AverageExecutions = progressReport.TotalExecutionsCount / int64(progressReport.NumberOfShardsFinished)
```
meiliang86, line 758: nit: AverageExecutions can be calculated only once outside of the per shard loop?

andrewjdawson2016: We emit this periodically though. So this is saying "based on the shards we have handled so far" this is the average.

PR #3180, opened 2020-04-09
Move history service logic into separate packages (part 1) 

yycptt:
**What changed?**
Move history service config and events cache into separate packages

**Why?**
https://github.com/uber/cadence/issues/3181

`service/history` package currently contains a large number of files. The goal is to break down the logic into smaller pieces and create separate packages for them. This also makes it easier to test and reason about.

Move config and eventsCache first as they are commonly used by other components in the history service. If we don't create separate packages for them first, when moving other components cycle dependency will be introduced.

**How did you test it?**
Build successfully, not lint warning. Unit test passed.

**Potential risks**
N/A.


service/history/events/cache.go
```diff
@@ -18,54 +18,55 @@
...
 	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/service/history/config"
 )
 
 type (
-	eventsCache interface {
-		getEvent(
+	// Cache caches workflow history event
+	Cache interface {
+		GetEvent(
```
yux0, line 40: comment on the public method?

yycptt: Sure I can add comments if you have strong opinion on this. But I think in this case the method name is self-explanatory so even if I add some comments it will just be some dummy words. Plus the linter doesn't complain about this. More generally, I don't think our code base has comments for all exported method, for example `ReplicationTaskFetcher`, `ShardContext`, etc.

But let me know if you still think it's necessary, I will add them.

yux0: As long as the lint check doesn't complain, then it is fine.

PR #3184, opened 2020-04-10
DB scan admin command retry db operations

andrewjdawson2016:

### DIFF
--- a/tools/cli/adminDBCleanCommand.go
+++ b/tools/cli/adminDBCleanCommand.go
@@ -289,13 +289,13 @@ func createCleanOutputDirectories() *CleanOutputDirectories {
 		SuccessfullyCleanedDirectoryPath: fmt.Sprintf("./clean_%v/successfully_cleaned", now),
 		FailedCleanedDirectoryPath:       fmt.Sprintf("./clean_%v/failed_cleaned", now),
 	}
-	if err := os.MkdirAll(cod.ShardCleanReportDirectoryPath, 0766); err != nil {
+	if err := os.MkdirAll(cod.ShardCleanReportDirectoryPath, 0777); err != nil {
 		ErrorAndExit("failed to create ShardCleanReportDirectoryPath", err)
 	}
-	if err := os.MkdirAll(cod.SuccessfullyCleanedDirectoryPath, 0766); err != nil {
+	if err := os.MkdirAll(cod.SuccessfullyCleanedDirectoryPath, 0777); err != nil {
 		ErrorAndExit("failed to create SuccessfullyCleanedDirectoryPath", err)
 	}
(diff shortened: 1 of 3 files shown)


tools/cli/adminDBCleanCommand.go
```diff
@@ -205,26 +208,23 @@ func cleanShard(
...
-		err = execStore.DeleteWorkflowExecution(deleteConcreteReq)
+		err = retryDeleteWorkflowExecution(limiter, &report.TotalDBRequests, execStore, deleteConcreteReq)
 		if err != nil {
 			report.Handled.FailedCleanedCount++
 			failedCleanWriter.Add(&ce)
 			continue
 		}
 		report.Handled.SuccessfullyCleanedCount++
 		successfullyCleanWriter.Add(&ce)
-		if ce.CorruptedExceptionMetadata.CorruptionType != OpenExecutionInvalidCurrentExecution {
```
yycptt, line 217: Deleted because you want to rely on the fact that `DeleteCurrentWorkflowExecution` is a conditional delete based on the runID? Just want to confirm.

tools/cli/adminDBScanCommand.go
```diff
@@ -141,11 +147,12 @@ type (
 	// ProgressReport contains metadata about the scan for all shards which have been finished
 	// This is periodically printed to stdout
 	ProgressReport struct {
-		NumberOfShardsFinished           int
+		NumberOfShardsFinished           int64
```
yycptt, line 150: `int64` seems like an overkill for # of shards.

tools/cli/adminDBScanCommand.go
```diff
@@ -141,11 +147,12 @@ type (
...
 		TotalExecutionsCount             int64
 		CorruptedExecutionsCount         int64
 		ExecutionCheckFailureCount       int64
 		NumberOfShardScanFailures        int64
+		ShardsFailed                     []int
```
yycptt, line 155: If # of shards has type `int64`, then this one should have type `[]int64`. Need to be consistent.

tools/cli/adminDBCleanCommand.go
andrewjdawson2016: Yes

tools/cli/adminDBScanCommand.go
andrewjdawson2016: Agree but I will need to cast to int64 anyways below.

tools/cli/adminDBCleanCommand.go
```diff
@@ -30,6 +30,9 @@ import (
 	"os"
 	"time"
 
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/backoff"
```
yycptt, line 34: Group with other cadence imports.

PR #3186, opened 2020-04-10
Decompose history service logic into separate packages (part 2)

yycptt:
**What changed?**
Move historyEventNotifier -> `events` package, history engine interface -> `engine` package, shardContext, shard Controller -> `shard` package

**Why?**
#3181

**How did you test it?**
Build successfully, not lint warning. Unit/integration test passed.

**Potential risks**
N/A


service/history/config/config.go
```diff
@@ -379,6 +380,16 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
 	return cfg
 }
 
+// NewForTest create new history service config for test
+func NewForTest() *Config {
```
vancexu, line 384: Maybe rename to NewTestConfig?

service/history/historyBuilder_test.go
```diff
@@ -44,7 +45,7 @@ type (
 		*require.Assertions
 
 		controller      *gomock.Controller
-		mockShard       *shardContextTest
+		mockShard       *shard.ContextTest
```
vancexu, line 48: Maybe rename to TestContext?

service/history/config/config.go
yycptt: I didn't include the word config in the function name because the package name is already config, so caller will write something like `config.NewForTest()`. Changing to `config.NewTestConfig()` sounds redundant to me.

service/history/historyBuilder_test.go
yycptt: Sounds good. Updated.

PR #3188, opened 2020-04-11
Fix panic on get history API

vancexu:
**What changed?**
Fix panic on get history API

**Why?**
when get history for non-exist workflow, currently it will cause server panic, and leading client side get unexpected context deadline exceed error.  

**How did you test it?**
local test, 
unit test

**Potential risks**
no risk. fix bug that are not in release yet.


meiliang86:
This needs to go into the 0.12 release. @anish531213

PR #3192, opened 2020-04-14
Add rawHistorySupported flag for backward compatability

anish531213:
Added rawHistorySupported flag on GetWorkflowExecutionHistoryRequest to make backward compatibility with the client.

**Why?**

**How did you test it?**

**Potential risks**


service/frontend/workflowHandler.go
```diff
@@ -1964,7 +1964,7 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		token.IsWorkflowRunning = isWorkflowRunning
 		token.PersistenceToken = nil
 	}
-	rawHistoryQueryEnabled := wh.config.SendRawWorkflowHistory(getRequest.GetDomain())
+	rawHistoryQueryEnabled := wh.config.SendRawWorkflowHistory(getRequest.GetDomain()) && getRequest.GetRawHistorySupported()
```
mkolodezny, line 1967: Should we provide the option for client which history to return?

mkolodezny:
Discussed offline: lets use the server version checker to check client instead.

PR #3193, opened 2020-04-15
Refactor db scan checks and include in delete

andrewjdawson2016:
This factors out db scan checks into there own interface. This makes it easier to add more checks in the future. It also enables re-verifying that an execution is still corrupted in the delete code path.

This makes the code more extensible. 

I have tested locally. This code is missing unit tests, unit tests will be added once logic is converted to workflow. 

If there is any bug in scan or delete it could lead an operator to delete non-corrupted production data.

### DIFF
--- a/tools/cli/admin.go
+++ b/tools/cli/admin.go
@@ -774,6 +774,10 @@ func newDBCommands() []cli.Command {
 					Name:  FlagReportRate,
 					Usage: "the number of shards which get handled between each emitting of progress",
 					Value: 10,
+				},
+				cli.BoolFlag{
+					Name:  FlagSkipHistoryChecks,
+					Usage: "skip over history check invariants",
 				}),
 			Action: func(c *cli.Context) {
 				AdminDBScan(c)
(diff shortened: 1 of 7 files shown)


tools/cli/admin.go
```diff
@@ -774,6 +774,10 @@ func newDBCommands() []cli.Command {
 					Name:  FlagReportRate,
 					Usage: "the number of shards which get handled between each emitting of progress",
 					Value: 10,
+				},
+				cli.BoolFlag{
```
andrewjdawson2016, line 778: In the future we can make this more flexible. I didn't want to handle this as part of this diff.

tools/cli/adminDBChecks.go
```diff
@@ -0,0 +1,400 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package cli
```
andrewjdawson2016, line 23: The number of files in this cli package around scanning is becoming overwhelming. In a follow up diff I will move these to worker package even before converting this to a workflow.

PR #3194, opened 2020-04-15
Add interfaces for multi-cursor queue implementation

yycptt:
**What changed?**
- Add interfaces for multi-cursor task queue implementation

**Why?**
- Prepare for multi-cursor task queue implementation

**How did you test it?**
Local build passed. No lint error. Unit test passed 

**Potential risks**
N/A


service/history/queueTask.go
```diff
@@ -33,17 +33,18 @@ import (
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/service/dynamicconfig"
-	"github.com/uber/cadence/common/task"
+	ctask "github.com/uber/cadence/common/task"
```
yycptt, line 36: This is temporary. Ideally, only files in `history/task` package will import `common/task` package. I will move files that are currently in `history` package and imports `common/task` to `history/task`

service/history/processing/interface.go
```diff
@@ -0,0 +1,89 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+//go:generate mockgen -copyright_file ../../../LICENSE -package $GOPACKAGE -source $GOFILE -destination interface_mock.go -self_package github.com/uber/cadence/service/history/processing
+
+package processing
```
yycptt, line 23: Any suggestion on the package name? This package will contain all the logic for managing tasks queues (loading, tracking/persisting ack level, sending task for processing etc.)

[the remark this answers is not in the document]
yycptt: Thanks for the suggestion on the package name. I will definitely consider that. 

I still want to separate the logic for task itself (detailed processing logic, priority assignment  etc.) and the logic for managing them (basically the logic for splitting/merging queues, loading tasks, managing ack levels) into two package as I think they are dealing with different things. If we dump them into one package, that package will become huge (>20 files I guess) and defeat the purpose of #3181. 

Thoughts?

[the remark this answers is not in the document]
vancexu: I see, make sense to have separate package.

vancexu: Maybe [task|job][Manager|Controller]?

emrahs: How about queueManager? I think it makes sense since this package is largely responsible for creating and managing virtual queues.

I also don't like the Job word as it is too generic. What would you think about renaming it to something like VirtualQueue, TaskQueue or [Scoped|Filtered]Queue?

service/history/processing/interface.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/service/history/task"
+)
+
+type (
```
emrahs, line 31: I think the "Job" name is a bit too generic and doesn't imply its actual responsibilities. What would you think about renaming it to something like VirtualQueue, TaskQueue or [Scoped|Filtered]Queue? I like VirtualQueue the most.

If we use "VirtualQueue" name instead of "Job", I'd suggest renaming the interfaces as follows: 

- JobInfo: VirtualQueueScope
- Job: VirtualQueue
- JobSplitPolicy: VirtualQueueSplitPolicy
- JobQueue: VirtualQueueCollection maybe? (I'm less confident about this one..)
- JobQueueManager: VirtualQueueManager

service/history/processing/interface.go
```diff
@@ -0,0 +1,89 @@
...
+	// JobSplitPolicy determins if a Job should be split
+	// into multiple Jobs
+	JobSplitPolicy interface {
+		Evaluate(Job) []JobInfo
+	}
+
+	// JobQueue manages a list of non-overlapping Jobs
+	// and keep track of the current active Job
+	JobQueue interface {
+		Info() []JobInfo
```
emrahs, line 73: Why does this return a list of JobInfos instead of a list of Jobs? 

Also, I'd consider renaming this function. It should probably be the plural form of the types of objects returned (ie. Jobs for [|Job, JobInfos for []JobInfo, etc.)

service/history/processing/interface.go
yycptt: Ideally I would like to use a single word for the package name, which is Golang convention and we follow that throughout the code base. So how about just using `queue` as the package name? As @emrahs mentioned, this package is largely responsible for creating and managing queues.

service/history/processing/interface.go
yycptt: I think returning `JobInfo` covers all the use cases I can think of:
- JobManager wants to persist job metadata to DB. In this case, all it needs can be retrieved from JobInfo
- For debugging purpose, we want to "describe the queue". Again in this case all we need is the metadata.

Maybe you have another use case in your mind that requires other information for Job and is not covered by JobInfo?

In terms of renaming, sure I can create type alias like Jobs for []Job, JobInfos for []JobInfo.

service/history/processing/interface.go
yycptt: I still want to keep `Info` in the type name as I think it contains more than just a scope. If the metadata only has min/max level and domain filter, then I think `Scope` is a perfect name. But in our case, it also contains information like read level, ack level, ID, etc. So it more like a status, instead of just scope. So maybe `xxInfo` or `xxStatus`?

In terms of the name of `Job`, I am not sure either. I am a little bit hesitate to rename it to `Queue` because then we don't have a good name for the current `JobQueue`. From my perspective, Job is one level below the concept of queue, more like a sub-task. 

If possible, can you elaborate a little bit on why you think `Job` doesn't imply its responsibilities? For example, which part is missing. Maybe based on that we can came up with a better name.

service/history/processing/interface.go
emrahs: Hmm, as a Golang newbie, I missed that convention. Found this write-up which has thorough recommendations about naming packages, and it's in line with @yycptt 's point about brevity: https://blog.golang.org/package-names

With that context, +1 from me for `queue`.

service/history/processing/interface.go
emrahs: In my view, JobInfo is essentially a data model object that represents the state/scope of a Job with getters/setters to access and manipulate the fields contained in it. There is a 1:1 relationship between a Job and JobInfo. Due to the nature of the relationship between them, it makes sense to use composition pattern to embed a JobInfo in the Job. I think Job should be the entity that's responsible for updating/persisting its JobInfo when needed, otherwise, bugs can arise due to lack of coordination between the state of Job and JobInfo objects. 

JobQueue is the abstraction that keeps track of a set of Jobs, and each Job object has its JobInfo as discussed above. What I find a bit unusual about the current proposal is that a JobQueue implementation would either need to keep track of JobInfo objects in addition to the Job objects, or it'd need to aggregate them from the Job objects at the time of Info() []JobInfo function call. I think it should simply return those Job objects so that the caller of this function can get the JobInfo from the Job for read-only purposes or interact with the Job interface to modify it (ie. UpdateAckLevel)

service/history/processing/interface.go
yycptt: Synced offline. Will use the following names:
- ProcessingQueue
- ProcessingQueueState
- ProcessingQueueSplitPolicy
- ProcessingQueueCollection
- ProcessingQueueManager

service/history/processing/interface.go
yycptt: Synced offline. Idea is make Job itself ignorant of persistence layer, but will change the method signature to `Queues() []ProcessingQueue` to be more flexible.

service/history/queue/interface.go
```diff
@@ -38,50 +38,56 @@ type (
 		ReverseMatch bool
 	}
 
-	// JobInfo indicates the scope of a Job and its progress
-	JobInfo interface {
-		QueueID() int
+	// ProcessingQueueStates is a list of ProcessingQueueState
+	ProcessingQueueStates []ProcessingQueueState
```
emrahs, line 42: Now that we have` Queues() ProcessingQueues` do we still need this alias?

service/history/queue/interface.go
```diff
@@ -38,50 +38,56 @@ type (
...
-		Merge([]Job)
+	// ProcessingQueueCollection manages a list of non-overlapping processing queues
+	// and keep track of the current active ProcessingQueue
+	ProcessingQueueCollection interface {
+		Queues() ProcessingQueues
+		Split(ProcessingQueueSplitPolicy) []ProcessingQueue
+		Merge(ProcessingQueues)
 		AddTasks(map[task.Key]task.Task)
-		ActiveJob() Job
+		ActiveJob() ProcessingQueue
```
emrahs, line 83: needs to be renamed

service/history/queue/interface.go
```diff
@@ -38,50 +38,56 @@ type (
...
 		AckLevel() task.Key
 		ReadLevel() task.Key
 		DomainFilter() DomainFilter
 	}
 
-	// Job is responsible for keeping track of the state of tasks
+	// ProcessingQueues is a list of ProcessingQueue
+	ProcessingQueues []ProcessingQueue
+
+	// ProcessingQueue is responsible for keeping track of the state of tasks
```
emrahs, line 57: This comment needs to be updated with new names

yycptt: Done.

service/history/queue/interface.go
yycptt: Renamed to `ActiveQueue()`

service/history/queue/interface.go
yycptt: Removed. I don't think they are necessary, but I thought you want me to add those alias?

emrahs: Sorry for the confusion. I wanted to say it's good to have alias if we use It :) Thanks!

PR #3195, opened 2020-04-15
Rate limit domain cache refresh

yycptt:
**What changed?**
Skip domain cache refresh if the time duration is too short since last refresh

**Why?**
If a domain doesn't exist and client keeping calling API for that domain, domain cache will be refreshed for every call, causing a large load on persistence layer and other APIs to fail due to service busy error.

**How did you test it?**
Unit test

**Potential risks**
N/A

### DIFF
--- a/common/cache/domainCache.go
+++ b/common/cache/domainCache.go
@@ -483,9 +483,6 @@ UpdateLoop:
 		}
 	}
 
-	// only update last refresh time when refresh succeeded
-	c.lastRefreshTime = now
-
 	// NOTE: READ REF BEFORE MODIFICATION
 	// ref: historyEngine.go registerDomainFailoverCallback function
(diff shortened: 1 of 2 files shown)


yycptt:
@wxing1292

common/cache/domainCache.go
```diff
@@ -476,6 +483,9 @@ UpdateLoop:
 		}
 	}
 
+	// only update last refresh time when refresh succeeded
+	c.lastRefreshTime = now
```
wxing1292, line 487: could you put this at the very end?

wxing1292:
LGTM

common/cache/domainCache.go
yycptt: Done.

PR #3201, opened 2020-04-17
Refactor start and signalwithstart logic

vancexu:
**What changed?**
Refactor start and signalwithstart logic in history

**Why?**
reduce duplicate, prepare for terminateIfRunning 

**How did you test it?**
existing tests 

**Potential risks**
if bug, start/signalwithstart api may encounter unexpected behavior

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -543,7 +543,8 @@ func (e *historyEngineImpl) startWorkflowHelper(
 		return nil, err
 	}
 
-	// for signalWithStart
+	// for signalWithStart, WorkflowIDReusePolicy is default to WorkflowIDReusePolicyAllowDuplicate
+	// while for startWorkflow it is default to WorkflowIdReusePolicyAllowDuplicateFailedOnly.
 	isSignalWithStart := signalWithStartArg != nil
 	var prevMutableState execution.MutableState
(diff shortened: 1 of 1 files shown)


service/history/historyEngine.go
```diff
@@ -518,12 +538,42 @@ func (e *historyEngineImpl) StartWorkflowExecution(
...
 	}
 
-	startEvent, err := mutableState.AddWorkflowExecutionStartedEvent(
+	// for signalWithStart
+	isSignalWithStart := signalWithStartArg != nil
+	var prevMutableState execution.MutableState
+	if isSignalWithStart && signalWithStartArg.prevMutableState != nil {
+		prevMutableState = signalWithStartArg.prevMutableState
+	}
+	if prevMutableState != nil {
```
yux0, line 552: Could we add some comment about the behavior of signalWithStart? The case I am not clear is when the workflow is closed and no current open workflow. When calling signalWithStart, what is going to happen?

vancexu: It will start workflow according to WorkflowIDReusePolicy, default policy for signalwithstart is WorkflowIDReusePolicyAllowDuplicate (which is different from startWorkflow)

vancexu: updated with comment

PR #3203, opened 2020-04-18
Replace context in historyEngine

vancexu:
**What changed?**
Rename context package, replace golang.org/x/net/context

**Why?**
As of Go 1.7 this package(golang.org/x/net/context) is available in the standard library under the name context. https://golang.org/pkg/context.

for renaming:
Renaming standard package is usually not a good practice, it confuses/mislead people. 
In this case, team unintentionally import "golang.org/x/net/context" for common context operation, because they are not aware of "context" is renamed. Same time, renamed to ctx and having local var ctx just make code confusing and less flexible.
  

**How did you test it?**
existing tests

**Potential risks**
no


vancexu:
I saw `ctx "context"` existing in some other history files also, and I think that is for saving local var name context to be used for workflowContext. 
I personally want to remove the package alias "ctx", and use `wfContext` for workflowContext. Let me know if team has any objection on this.

PR #3206, opened 2020-04-18
Check on-going failover across clusters

yux0:
**What changed?**
Check if there is ongoing failover when a graceful failover initiated. This is a best-effort operation to prevent concurrent failover. But in extreme case, this check could invalid.

**Why?**
Best effort to prevent concurrent failvoer.

**How did you test it?**
Local test.

**Potential risks**
This is a in-progress feature


service/frontend/workflowHandler.go
```diff
@@ -3706,6 +3717,69 @@ func isFailoverRequest(updateRequest *gen.UpdateDomainRequest) bool {
...
+		}
+		frontendClient := wh.GetRemoteFrontendClient(clusterName)
+		wg.Add(1)
+		go decribeDomain(
+			ctx,
+			frontendClient,
+			domainName,
+		)
+	}
+	wg.Wait()
```
yycptt, line 3760: `respChan` has 0 capacity and `wg.Done()` is called after resp has been sent. If `wg.Wait()` is called here, I think it will result in a deadlock because there's no receiver for the channel -> resp can't be sent -> `wg.Done()` won't be called.

I would suggest get rid of the wait group entirely. You already know how many responses will be sent when creating the go routine, just use a for loop to receive responses from `respChan`.

service/frontend/workflowHandler.go
```diff
@@ -3706,6 +3717,69 @@ func isFailoverRequest(updateRequest *gen.UpdateDomainRequest) bool {
 	return updateRequest.ReplicationConfiguration != nil && updateRequest.ReplicationConfiguration.ActiveClusterName != nil
 }
 
+func isGraceFailoverRequest(updateRequest *gen.UpdateDomainRequest) bool {
+	return updateRequest.IsSetFailoverTimeoutInSeconds()
```
yycptt, line 3721: What if someone calls the API with failover timeout set but there's no change in the active cluster field or leaves it as empty?

service/frontend/workflowHandler.go
```diff
@@ -3706,6 +3717,69 @@ func isFailoverRequest(updateRequest *gen.UpdateDomainRequest) bool {
...
+		if resp == nil {
+			return &gen.InternalServiceError{
+				Message: "Failed to verify failover version from all clusters",
+			}
+		}
+		if failoverVersion == nil {
+			failoverVersion = resp.FailoverVersion
+		}
+		if failoverVersion != resp.FailoverVersion {
+			prevCluster := clusterMetadata.ClusterNameForFailoverVersion(*resp.FailoverVersion)
```
yycptt, line 3774: I don't quite understand this. What does `previous` mean here?

yux0: This only used for log message. I remove this part.

service/frontend/workflowHandler.go
yux0: Those will handle in the domain handler. 
First scenario: added in handler to ban this case.
Second scenario: It already ban in the handler

service/frontend/workflowHandler.go
yux0: Good catch. Set the channel size and still use the wait group.

PR #3207, opened 2020-04-20
Implement domain filter for multi-cursor queue

yycptt:
**What changed?**
Implement domain filter for multi-cursor queue

**Why?**
Each processing queue is responsible for only a subset of domains. Domain filter is component for representing the a set of domains. It has util methods including `Filter`, `Include`, `Exclude`, etc. for creating new filters.

**How did you test it?**
Unit test.

**Potential risks**
N/A. It's currently not used by the system.


service/history/queue/domain_filter.go
```diff
@@ -0,0 +1,132 @@
...
+package queue
+
+// NewDomainFilter creates a new domain filter
+func NewDomainFilter(
+	domainIDs map[string]struct{},
+	reverseMatch bool,
+) DomainFilter {
+	return DomainFilter{
+		DomainIDs:    domainIDs,
+		ReverseMatch: reverseMatch,
```
yux0, line 30: I think I asked this before. But could you give me an example of this reverse match use case?

yycptt: This is basically for the default queue which is responsible for all the domains in the beginning. Instead of listing all domainIDs here and keep refreshing to see if there's new domains, I think using an empty list and do a reverse match is a much better approach. 

Later if a split is performed on the default queue and a domain no longer belongs to the default queue, reverse match will still be true and domainID list will contain the one that got split out. This means now the default queue is responsible for all domains except the one that has been split out.

yux0: It is weird to me because the Filter/Include/Exclude have different behavior depends on the domain filter initialization. If it is reverse matche is true, then exclude is include and include is exclude.

yycptt: Synced offline. `Include` will always add domains to the set, `Exclude` will always remove domains from the set and `Merge` always output a union of two sets.

PR #3208, opened 2020-04-20
Not Ready For Review: Add blobstore interface

andrewjdawson2016:
Add an interface to blobstore. This has several potentially interesting usecases, but the one that this is being added for is the scan/clean workflow.  This interface is meant to start very simple and as we find reason to add to it - we can make it more complex. End users do not depend on this interface so we can safely make changes in the future. 

If the interface is defined poorly its possible that we will need to change it and potentially also make some changes to implementations of it.

### DIFF
--- a/common/resource/resource.go
+++ b/common/resource/resource.go
@@ -23,8 +23,6 @@ package resource
 import (
 	"go.uber.org/yarpc"
 
-	"github.com/uber/cadence/common/blobstore"
-
 	"go.uber.org/cadence/.gen/go/cadence/workflowserviceclient"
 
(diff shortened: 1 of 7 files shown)


common/resource/resource.go
```diff
@@ -65,6 +66,7 @@ type (
 		GetMetricsClient() metrics.Client
 		GetArchiverProvider() provider.ArchiverProvider
 		GetMessagingClient() messaging.Client
+		GetBlobstoreClient() blobstore.Client
```
yycptt, line 69: Please also add this method to `resourceImpl.go` and `resourceTest.go`

PR #3210, opened 2020-04-22
Add blobstore interface to bootstrap params and implement filestore

andrewjdawson2016:
Having a blobstore as a resource is a useful primitive in many contexts. It is being added here to be used in db scan. Having a local file system implementation allows testing this locally.

In order to have a blobstore for scanner. 

There are unit tests and I verified locally that blobstore client works. 

The interface might have been defined badly so that when external developers implement there own version of blobstore it might need to be changed later.


cmd/server/cadence/server.go
```diff
@@ -203,10 +205,13 @@ func (s *server) startService() common.Daemon {
 	)
 
 	params.ArchiverProvider = provider.NewArchiverProvider(s.cfg.Archival.History.Provider, s.cfg.Archival.Visibility.Provider)
-
 	params.PersistenceConfig.TransactionSizeLimit = dc.GetIntProperty(dynamicconfig.TransactionSizeLimit, common.DefaultTransactionSizeLimit)
-
 	params.Authorizer = authorization.NewNopAuthorizer()
+	params.BlobstoreClient, err = filestore.NewFilestoreClient(s.cfg.Blobstore.Filestore)
```
yux0, line 210: should we use file store / file client for everything here?

yux0, line 210: could we move this to worker/server.go? The other three servers don't need this blob store client.

andrewjdawson2016: They might. I think this is a common utility just like archiver.

yycptt: I agree that blobstore can be a common utility but I think we need to think more about how to config/initialize it. Right now it seems like there's only one global config for it. I mean the entire system can use only one blobstore implementation with a certain config right now. But maybe in the future there will be use cases where different components need different blobstore implementations. Take a step back, even within an implementation, I think different configs might be needed. For example, archiver want to upload history to a path(bucket) with very long retention period, while the bucket for scanner only need a very short retention time.

cmd/server/cadence/server.go
```diff
@@ -24,6 +24,8 @@ import (
 	"log"
 	"time"
 
+	"github.com/uber/cadence/common/blobstore/filestore"
```
yycptt, line 27: Please group with other cadence imports.

common/file_util.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package common
```
yycptt, line 23: How about creating a separate package for all file related util functions?

common/blobstore/filestore/client.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	if len(cfg.OutputDirectory) == 0 {
+		return nil, errors.New("output directory not given for file blobstore")
+	}
+	outputDirectory := cfg.OutputDirectory
+	exists, err := common.DirectoryExists(outputDirectory)
+	if err != nil {
+		return nil, err
+	}
+	if exists {
```
yycptt, line 56: nit: I would prefer check `if !exists { // create the dir and return error}`, then `return &client{ outputDirectory: outputDirectory}, nil`

common/blobstore/filestore/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Delete deletes a blob
+func (c *client) Delete(_ context.Context, request *blobstore.DeleteRequest) (*blobstore.DeleteResponse, error) {
+	if err := os.Remove(c.filepath(request.Key)); err != nil {
+		return nil, err
+	}
+	return &blobstore.DeleteResponse{}, nil
+}
+
+func (c *client) deserializeBlob(data []byte) (*blobstore.Blob, error) {
```
yycptt, line 115: Seems like all other places use Blob directly instead of using pointer, probably also return `blobstore.Blob` here?

common/service/config/config.go
```diff
@@ -68,6 +68,8 @@ type (
 		DynamicConfigClient dynamicconfig.FileBasedClientConfig `yaml:"dynamicConfigClient"`
 		// DomainDefaults is the default config for every domain
 		DomainDefaults DomainDefaults `yaml:"domainDefaults"`
+		// Blobstore is the config for setting up blobstore
+		Blobstore Blobstore
```
yycptt, line 72: Add a `yaml` tag?

config/development_standby.yaml
```diff
@@ -152,3 +152,7 @@ domainDefaults:
 
 publicClient:
   hostPort: "localhost:8933"
+
+blobstore:
+  filestore:
+    outputDirectory: "/tmp/blobstore"
```
yycptt, line 158: nit: Please add a newline at the end

cmd/server/cadence/server.go
andrewjdawson2016: Followed up offline on this - will create ticket around this.

common/resource/resourceTest.go
```diff
@@ -69,6 +70,7 @@ type (
 		MetricsClient     metrics.Client
 		ArchivalMetadata  *archiver.MockArchivalMetadata
 		ArchiverProvider  *provider.MockArchiverProvider
+		BlobstoreClient   blobstore.Client
```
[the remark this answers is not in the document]
yycptt: I think we need to use `blobstore.MockClient` type here, otherwise in the test you can't call methods like `EXCEPT()`

PR #3214, opened 2020-04-23
Multi-cursor processing queue implementation

yycptt:
**What changed?**
Implement processing queue and processing queue state interface

**Why?**
Part of the multi-cursor task processing effort. Processing queue the is lowest level component for the entire implementation.

**How did you test it?**
Unit test.

**Potential risks**
N/A.


service/history/queue/processing_queue.go
```diff
@@ -0,0 +1,366 @@
...
+				key, task.GetDomainID(), task.GetWorkflowID(), task.GetRunID(), task.GetTaskType()))
+			continue
+		}
+
+		if !taskBelongsToProcessQueue(q.state, key, task) {
+			errMsg := "Processing queue encountered a task doesn't belong to its scope"
+			q.logger.Error(errMsg, tag.Error(
+				fmt.Errorf("Processing queue state: %+v, task: %+v", q.state, key),
+			))
+			panic(errMsg)
```
yux0, line 250: Is this a sanity check and should panic?

yycptt: Yes this is just a sanity check. If this happens, it means there's a bug in the implementation for queue collection. and yes, the code will panic on L250 if this happens.

PR #3215, opened 2020-04-23
Add TerminateIfRunning to WorkflowIDReusePolicy

vancexu:
**What changed?**
Add TerminateIfRunning to WorkflowIDReusePolicy

**Why?**
https://github.com/uber/cadence/issues/2653

**How did you test it?**
unit / integration test
local test

**Potential risks**
no


service/frontend/workflowHandler.go
```diff
@@ -2221,6 +2221,11 @@ func (wh *WorkflowHandler) SignalWithStartWorkflowExecution(
 			Message: "A valid TaskStartToCloseTimeoutSeconds is not set on request."}, scope)
 	}
 
+	if signalWithStartRequest.GetWorkflowIdReusePolicy() == gen.WorkflowIdReusePolicyTerminateIfRunning {
```
yux0, line 2224: Any reason why this is not allow?

service/history/historyEngine.go
```diff
@@ -2700,30 +2717,66 @@ func (e *historyEngineImpl) applyWorkflowIDReusePolicyForSigWithStart(
...
+	domainID string,
 ) error {
 
 	// here we know some information about the prev workflow, i.e. either running right now
 	// or has history check if the workflow is finished
 	switch prevState {
 	case persistence.WorkflowStateCreated,
 		persistence.WorkflowStateRunning:
+		if wfIDReusePolicy == workflow.WorkflowIdReusePolicyTerminateIfRunning {
+			err := e.requestTerminateWorkflow(ctx, domainID, execution, prevRunID)
```
yycptt, line 2774: I think I may missed something here. How do we ensure terminate and start are done in one transaction?

service/frontend/workflowHandler.go
vancexu: Originally I think we shouldn't allow it because of 1. user requirements 2. simplicity
But thinking more on that I feel it's better to support it. Will add.
Thanks for bring it up

service/history/historyEngine.go
vancexu: The problem to solve is: when user terminate workflow then start, they may see WorkflowAlreadyExist. 
With this change, the `requestTerminateWorkflow` is called with current execution lock so others are block to start. Then it terminate the current run, and start a new one in one "semantic" transaction (not a db transaction).  

This part is indeed not-straightforward and may confuse people, let me know if that make sense

vancexu: I will also add these comments to code

yux0: Two questions:
1. If the lock is acquired in the start workflow helper, then when it calls terminate, how can terminate get the lock?
2. For any reason, if the call stops after it terminates the workflow. Then could user see the behavior like call startWorkflowExecution and received a context timeout then the current open workflow gets terminated?

vancexu: Synced offline, but the short answers are:
1. the lock are different. lock in startworkflowhelper is for current execution, terminate is using concrete execution lock
2. yes, user will see current get terminated but start failed.

The 2nd question is super and thanks for bring it up. Let me think more on best option.

vancexu: I will do rewrite this and make it in one db transaction

service/history/historyEngine.go
```diff
@@ -696,6 +713,138 @@ func (e *historyEngineImpl) startWorkflowHelper(
...
+	signalWithStartRequest *h.SignalWithStartWorkflowExecutionRequest,
+) (*workflow.StartWorkflowExecutionResponse, error) {
+	runningMutableState := runningWFCtx.getMutableState()
+UpdateWorkflowLoop:
+	for attempt := 0; attempt < conditionalRetryCount; attempt++ {
+		if !runningMutableState.IsWorkflowExecutionRunning() {
+			return nil, ErrWorkflowCompleted
+		}
+
+		if err := execution.TerminateWorkflow(
```
yux0, line 740: I think we want to do the terminate then start at the db level?

service/history/historyEngine.go
```diff
@@ -696,6 +713,138 @@ func (e *historyEngineImpl) startWorkflowHelper(
...
+			newMutableState,
+			workflowExecution,
+			startRequest,
+			signalWithStartRequest,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		updateErr := runningWFCtx.getContext().UpdateWorkflowExecutionWithNewAsActive(
```
yux0, line 782: LGTM. This update will call updateWorkflowExecutionEventReapply and reapply signals. Just double check it compatible with signal with start

service/history/historyEngine.go
yux0: Sync offline. It makes sense now

service/history/historyEngine.go
```diff
@@ -72,6 +72,13 @@ const (
 	defaultQueryFirstDecisionTaskWaitTime     = time.Second
 	queryFirstDecisionTaskCheckInterval       = 200 * time.Millisecond
 	replicationTimeout                        = 30 * time.Second
+
+	// TerminateIfRunningReason reason for terminateIfRunning
+	TerminateIfRunningReason = "TerminateIfRunning Policy"
+	// TerminateIfRunningDetailsTemplate details template for terminateIfRunning
+	TerminateIfRunningDetailsTemplate = "New runID: %s"
+	// TerminateIfRunningIdentity identity for terminateIfRunning
+	TerminateIfRunningIdentity = "Cadence history"
```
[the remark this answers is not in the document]
yycptt: Let's use the constant defined in the new execution package, `execution.IdentityHistoryService`

PR #3216, opened 2020-04-23
Add feature flags for RPC replication migration

yux0:
**What changed?**
Add feature flag for RPC replication migration

**Why?**
The migration plan looks like:
If we have cluster A and cluster B and want to migrate to have A use RPC replication from B.

1. B front end switches from publishing domain replication from Kafka to RPC
2. A history starts RPC replication
3. B history stops publishing history replication to Kafka
4. A worker switches from reading from Kafka to RPC replication. 

**How did you test it?**
Test environment.

**Potential risks**
The replication stack will be broken until rolled back the feature flag.


common/service/dynamicconfig/constants.go
```diff
@@ -99,6 +99,7 @@ var keys = map[Key]string{
 	EnableClientVersionCheck:              "frontend.enableClientVersionCheck",
 	ValidSearchAttributes:                 "frontend.validSearchAttributes",
 	SendRawWorkflowHistory:                "frontend.sendRawWorkflowHistory",
+	EnableRPCReplication:                  "frontend.rpcReplicationEnabled",
```
yycptt, line 102: nit: `frontend.enableRPCReplication`?

common/service/dynamicconfig/constants.go
```diff
@@ -232,6 +233,8 @@ var keys = map[Key]string{
 	ReplicationTaskProcessorNoTaskInitialWait:              "history.ReplicationTaskProcessorNoTaskInitialWait",
 	ReplicationTaskProcessorCleanupInterval:                "history.ReplicationTaskProcessorCleanupInterval",
 	ReplicationTaskProcessorCleanupJitterCoefficient:       "history.ReplicationTaskProcessorCleanupJitterCoefficient",
+	RPCReplicationEnabled:                                  "history.RPCReplicationEnabled",
+	KafkaReplicationEnabled:                                "history.KafkaReplicationEnabled",
```
yycptt, line 237: We named it `XXXEnabled` instead of `EnableXXX` to avoid conflict with the one for frontend? If that's the case probably append the flag name with history/frontend/worker.

common/service/dynamicconfig/constants.go
yux0: updated

common/service/dynamicconfig/constants.go
yux0: Updated

PR #3218, opened 2020-04-25
Fix metadata replication task with NDC

yux0:
**What changed?**
generate metadata replication task when ndc replication task rejected by queue because of size too large.

**Why?**
ndc replication task size could be too large and fails to put it into queue

**How did you test it?**
Local test

**Potential risks**
The metadata replication task fails to be applied and ends up in DLQ. The replication stack relies on resend and keep running.


yycptt:
Please also check in IDL submodule change before landing.

PR #3224, opened 2020-04-29
Add iterator for buffered writer

andrewjdawson2016:
This adds an iterator for buffered writer. This will be used in the clean workflow.


common/util/iterator.go
```diff
@@ -0,0 +1,165 @@
...
+	copy(separatorTokenCopy, separatorToken)
+	itr := &iterator{
+		currentPage: -1,
+
+		separatorToken: separatorTokenCopy,
+		getFn:          getFn,
+		minPage:        minPage,
+		maxPage:        maxPage,
+	}
+	itr.advance(true)
```
vancexu, line 86: maybe just itr.init() instead of advance()?

andrewjdawson2016: Its also called in next method though.

PR #3228, opened 2020-04-30
Fix NPE for task when global domain disabled

vancexu:
**What changed?**
Fix NPE 

**Why?**
Came into this NPE while testing on staging. 
This NPE happens when config clusterMetadata set enableGlobalDomain to false. 

**How did you test it?**
staging test

**Potential risks**
no

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2695,7 +2695,7 @@ func (e *historyEngineImpl) NotifyNewReplicationTasks(
 	tasks []persistence.Task,
 ) {
 
-	if len(tasks) > 0 {
+	if len(tasks) > 0 && e.replicatorProcessor != nil {
 		e.replicatorProcessor.notifyNewTask()
 	}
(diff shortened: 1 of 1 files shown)


service/history/historyEngine.go
```diff
@@ -2695,7 +2695,7 @@ func (e *historyEngineImpl) NotifyNewReplicationTasks(
 	tasks []persistence.Task,
 ) {
 
-	if len(tasks) > 0 {
+	if len(tasks) > 0 && e.replicatorProcessor != nil {
```
yux0, line 2698: Should it be a bug is the processor is nil?

yux0: My understanding is this processor always set when engine initializes

vancexu: only set when global domain is enabled. notify replication for non-global cluster is the bug, processor being nil is not a bug.

PR #3232, opened 2020-05-04
Implement Processing Queue Split Policy

yycptt:
**What changed?**
Implement task processing queue split policy

**Why?**
Part of multi-tenant solution

**How did you test it?**
Unit test

**Potential risks**
N/A


service/history/queue/split_policy.go
```diff
@@ -0,0 +1,276 @@
...
+) []ProcessingQueueState {
+	queueImpl := queue.(*processingQueueImpl)
+
+	threshold, ok := p.pendingTaskThreshold[queueImpl.state.level]
+	if !ok {
+		// no threshold specified for the level, skip spliting
+		return nil
+	}
+
+	if queueImpl.state.level == p.maxNewQueueLevel {
```
yux0, line 112: Is the maxNewQueueLevel inclusive or exclusive?

yycptt: It's inclusive. For this split policy, if the current level is `x`, then the level for the split(new) queue will be `x+1`. So if the current level already equals to `maxNewQueueLevel` and you do split, level for new queue will exceed the threshold. Therefore, we need to skip splitting here.

yycptt: Actually, I can move this check before the check for `pendingTaskThreshold` to avoid confusion. But the result will be the same when current level = maxNewQueueLevel, either the split is rejected due to no threshold defined or even if there's one, it will be rejected by the maxNewQueueLevel.

PR #3234, opened 2020-05-05
Add iterators and writers for persistence and blobstore for scanner workflow

andrewjdawson2016:
**What changed?**

These iterators and writes will be used in scanner and fixer workflow. 

There are unit tests for generic write and iterator but I will write more for the concrete impls.


service/worker/scanner/executions/common/persistenceIterator.go
```diff
@@ -0,0 +1,90 @@
...
+		resp, err := pr.ListConcreteExecutions(req)
+		if err != nil {
+			return nil, nil, err
+		}
+		executions := make([]pagination.Entity, len(resp.Executions), len(resp.Executions))
+		for i, e := range resp.Executions {
+			// TODO: check with Yu that this is correct
+			branchToken := e.ExecutionInfo.BranchToken
+			if e.VersionHistories != nil {
+				branchToken = e.VersionHistories.Histories[e.VersionHistories.CurrentVersionHistoryIndex].BranchToken
```
yux0, line 77: I think you may do versionHistories.GetCurrentVersionHistory().BranchToken

common/pagination/interface.go
```diff
@@ -0,0 +1,69 @@
...
+	// FetchFn fetches the entities and next PageToken for the provided PageToken or error on failure.
+	FetchFn func(PageToken) ([]Entity, PageToken, error)
+)
+
+type (
+	// Iterator is used to get entities from a collection of pages.
+	// When HasNext returns true it is guaranteed that Next will not return an error.
+	// Once iterator returns an error it will never make progress again and will always return that same error.
+	// Iterator is not thread safe and does not make defensive in or out copies.
+	Iterator interface {
```
yycptt, line 57: How's the `Iterator` different from the one defined in the `common/collection` package? https://github.com/uber/cadence/blob/master/common/collection/iterator.go#L25

PR #3237, opened 2020-05-05
Add tag and dynamic config for authz

vancexu:
**What changed?**
Add tag and dynamic config for authz

**Why?**
Needed for authz implementation (internally)

**How did you test it?**
staging test

**Potential risks**
no


common/log/tag/tags.go
```diff
@@ -336,6 +336,11 @@ func Address(ad string) Tag {
 	return newStringTag("address", ad)
 }
 
+// Env return tag for runtime environment
+func Env(env string) Tag {
```
meiliang86, line 340: Where do you use this?

common/service/dynamicconfig/constants.go
```diff
@@ -70,6 +70,7 @@ var keys = map[Key]string{
 	EnableParentClosePolicyWorker:       "system.enableParentClosePolicyWorker",
 	EnableStickyQuery:                   "system.enableStickyQuery",
 	EnablePriorityTaskProcessor:         "system.enablePriorityTaskProcessor",
+	EnableAuthorization:                 "system.enableAuthorization",
```
meiliang86, line 73: Missing something? You need to use it somewhere, right?

common/log/tag/tags.go
vancexu: This will be used in authorizer implementation. In this case, it's uber internal server.

common/service/dynamicconfig/constants.go
vancexu: This will be used in authorizer implementation. In this case, it's uber internal server.

PR #3238, opened 2020-05-05
Refactor: remove redundant event cache from mutableStateBuilder

vancexu:
**What changed?**
Remove eventcache from newMutableStateBuilder to simplify its signature

**Why?**
eventcache is redundant 

**How did you test it?**
unit test

**Potential risks**
no


service/history/execution/mutable_state_builder.go
```diff
@@ -260,26 +257,24 @@ func NewMutableStateBuilderWithReplicationState(
...
 ) MutableState {
 
-	s := newMutableStateBuilder(shard, eventsCache, logger, domainEntry)
+	s := newMutableStateBuilder(shard, logger, domainEntry)
 	s.versionHistories = persistence.NewVersionHistories(&persistence.VersionHistory{})
 	return s
 }
 
 // NewMutableStateBuilderWithEventV2 is used only in test
 func NewMutableStateBuilderWithEventV2(
```
mkolodezny, line 270: nit: do we still need v1 anywhere?

service/history/historyReplicator.go
```diff
@@ -146,7 +146,6 @@ func newHistoryReplicator(
 		getNewMutableState: func(domainEntry *cache.DomainCacheEntry, logger log.Logger) execution.MutableState {
 			return execution.NewMutableStateBuilderWithReplicationState(
 				shard,
-				shard.GetEventsCache(),
```
mkolodezny, line 149: nit: does replicator work with v1 mutable state?

service/history/execution/mutable_state_builder.go
```diff
@@ -288,14 +283,13 @@ func NewMutableStateBuilderWithEventV2(
 // NewMutableStateBuilderWithReplicationStateWithEventV2 is used only in test
```
mkolodezny, line 283: nit: is the only difference between v1 and v2 in the persistence Cassandra layer?

PR #3239, opened 2020-05-06
Update license file and some small code cleanup

meiliang86:
**What changed?**
License and some small code clean up.

**Why?**

**How did you test it?**
Existing test should cover this.

**Potential risks**
na


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2020 Uber Technologies, Inc.
+Copyright (c) 2017-2020 Uber Technologies Inc.
```
sheerun, line 3: I like how Facebook handles it. They just write 2017-present and finally forget about updating this

meiliang86: Good suggestion :) let me see if that's possible.

PR #3241, opened 2020-05-07
Add flag for replication cleanup process

yux0:
**What changed?**
Add a feature flag to config rpc replication queue clean up process

**Why?**
Disable this cleanup so that we don't have backward compatibility if we want to rollback

**How did you test it?**
Local test

**Potential risks**
The task deletes by the process and after rolled back, the original logic cannot find the task and throws internal service error

### DIFF
--- a/common/persistence/domainReplicationQueue.go
+++ b/common/persistence/domainReplicationQueue.go
@@ -101,7 +101,7 @@ func (q *domainReplicationQueueImpl) Start() {
 }
 
 func (q *domainReplicationQueueImpl) Stop() {
-	if !atomic.CompareAndSwapInt32(&q.status, common.DaemonStatusInitialized, common.DaemonStatusStarted) {
+	if !atomic.CompareAndSwapInt32(&q.status, common.DaemonStatusInitialized, common.DaemonStatusStopped) {
 		return
 	}
 	close(q.done)
--- a/service/frontend/adminHandler.go
+++ b/service/frontend/adminHandler.go
@@ -124,9 +124,7 @@ func (adh *AdminHandler) Start() {
 
 // Stop stops the handler
 func (adh *AdminHandler) Stop() {
-	if adh.config.EnableCleanupReplicationTask() {
-		adh.Resource.GetDomainReplicationQueue().Stop()
-	}
+	adh.Resource.GetDomainReplicationQueue().Stop()
 }
 
 // AddSearchAttribute add search attribute to whitelist
--- a/service/frontend/adminHandler.go
+++ b/service/frontend/adminHandler.go
@@ -117,12 +117,16 @@ func (adh *AdminHandler) RegisterHandler() {
 // Start starts the handler
 func (adh *AdminHandler) Start() {
 	// Start domain replication queue cleanup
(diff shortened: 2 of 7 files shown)


service/frontend/service.go
```diff
@@ -135,6 +136,7 @@ func NewConfig(dc *dynamicconfig.Collection, numHistoryShards int, enableReadFro
 		DisallowQuery:                       dc.GetBoolPropertyFnWithDomainFilter(dynamicconfig.DisallowQuery, false),
 		SendRawWorkflowHistory:              dc.GetBoolPropertyFnWithDomainFilter(dynamicconfig.SendRawWorkflowHistory, false),
 		EnableRPCReplication:                dc.GetBoolProperty(dynamicconfig.FrontendEnableRPCReplication, false),
+		EnableCleanupReplicationTask:        dc.GetBoolProperty(dynamicconfig.FrontendEnableCleanupReplicationTask, false),
```
yycptt, line 139: Do we need to default to `true` as currently the code will always do clean up? Same for the flag for history service.

service/frontend/adminHandler.go
```diff
@@ -117,12 +117,16 @@ func (adh *AdminHandler) RegisterHandler() {
...
+		adh.Resource.GetDomainReplicationQueue().Start()
+	}
 }
 
 // Stop stops the handler
 func (adh *AdminHandler) Stop() {
-	adh.Resource.GetDomainReplicationQueue().Stop()
+	if adh.config.EnableCleanupReplicationTask() {
+		adh.Resource.GetDomainReplicationQueue().Stop()
+	}
```
yycptt, line 129: The value for this dynamicconfig may and will change when the service is running, so it's possible only `Start()` is called or only `Stop()` is called. Ideally the config value should be get once and stored somewhere on service startup.

I checked the implementation for domain replication queue, it seems to be ok if only `Stop()` is called, not sure about what will happen if only `Start()` is called. But if you're relying on this fact, probably add some comments in the code.

BTW, there's a copy & paste error on `domainReplicationQueue.go` L104.

yux0: Good point. So I updated it to always call stop() and check the flag before calling start()

service/frontend/service.go
yux0: update fault to true

common/persistence/domainReplicationQueue.go
```diff
@@ -101,7 +101,7 @@ func (q *domainReplicationQueueImpl) Start() {
 }
 
 func (q *domainReplicationQueueImpl) Stop() {
-	if !atomic.CompareAndSwapInt32(&q.status, common.DaemonStatusInitialized, common.DaemonStatusStarted) {
+	if !atomic.CompareAndSwapInt32(&q.status, common.DaemonStatusInitialized, common.DaemonStatusStopped) {
```
yycptt, line 104: `common.DaemonStatusInitialized` -> `common.DaemonStarted`

service/frontend/adminHandler.go
yycptt: Please also add some comments explaining that it's ok to call `Stop()` without calling `Start()`.

PR #3242, opened 2020-05-07
Update contributing instructions

sheerun:
**What changed?**

- Remove "Dependency management" section (Cadance is no longer using godep for managing dependencies. Instead go mod is used which doesn't require any commands)
- Update installation instructions of of cassandra, kafka, and zookeeper on OSX (you cannot use default zookeeper formula because it's not compatible with java that ships with OSX)


CONTRIBUTING.md
```diff
@@ -93,5 +93,5 @@ go test -v
 # run single test
 go test -v <path> -run <TestSuite> -testify.m <TestSpercificTaskName>
 # example:
-go test -v github.com/uber/cadence/common/persistence -run TestCassandraPersistenceSuite -testify.m TestPersistenceStartWorkflow
+go test -v github.com/uber/cadence/common/persistence/persistence-tests -run TestVisibilitySamplingSuite -testify.m TestListClosedWorkflowExecutions
```
meiliang86, line 96: Can we keep this unchanged?

sheerun: I guess, but old one isn't working, there is no TestCassandraPersistenceSuite in this directory. Do you still want me to revert?

sheerun: @meiliang86 ping

PR #3244, opened 2020-05-07
Better postgresql test defaults on OSX

sheerun:
**What changed?**
- on OSX sets postgresql username / password to "$USER" and empty string (defaults for homebrew)
- fixes some related comments

**Why?**
- better out-of-the-box testing experience on osx

**How did you test it?**
- I've run postgresql tests with `go test -v github.com/uber/cadence/common/persistence/sql/sqlplugin/postgres`

**Potential risks**

Can fail if someone has already configured postgresql in non-standard way with postgres / cadence username / password on OSX

### DIFF
--- a/common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
+++ b/common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
@@ -21,6 +21,8 @@
 package postgres
 
 import (
+	"os"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
@@ -30,12 +32,26 @@ import (
 )
 
 const (
-	testUser      = "postgres"
-	testPassword  = "cadence"
 	testSchemaDir = "schema/postgres"
 )
 
 func getTestClusterOption() *pt.TestBaseOptions {
+	testUser := "postgres"
+	testPassword := "cadence"
+
+	if runtime.GOOS == "darwin" {
+		testUser = os.Getenv("USER")
+		testPassword = ""
+	}
+
+	if os.Getenv("POSTGRES_USER") != "" {
+		testUser = os.Getenv("POSTGRES_USER")
+	}
+
+	if os.Getenv("POSTGRES_PASSWORD") != "" {
+		testUser = os.Getenv("POSTGRES_PASSWORD")
+	}
+
 	return &pt.TestBaseOptions{
 		SQLDBPluginName: PluginName,
 		DBUsername:      testUser,
--- a/environment/env.go
+++ b/environment/env.go
@@ -161,7 +161,7 @@ func GetCassandraPort() int {
 	return p
 }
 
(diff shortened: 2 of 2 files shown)


environment/env.go
```diff
@@ -66,7 +66,6 @@ const (
 	PostgresDefaultPort = "5432"
 )
 
-// SetupEnv setup the necessary env
```
yux0, line 69: What is the reason to remove those? It will cause lint warning

sheerun: They were extraneous and wrong in few places. I guess I'll fix them if they are needed by linter..

common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
```diff
@@ -30,12 +32,17 @@ import (
...
-	testPassword  = "cadence"
 	testSchemaDir = "schema/postgres"
 )
 
 func getTestClusterOption() *pt.TestBaseOptions {
+	testUser := "postgres"
+	testPassword := "cadence"
+
+	if runtime.GOOS == "darwin" {
+		testUser = os.Getenv("USER")
```
yux0, line 43: Could it use POSTGRES_USER?

common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
```diff
@@ -30,12 +32,17 @@ import (
...
+		testPassword = ""
```
yux0, line 44: Provide an option to set the password, like POSTGRES_PASSWORD?

common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
```diff
@@ -30,12 +32,17 @@ import (
...
-	testUser      = "postgres"
-	testPassword  = "cadence"
 	testSchemaDir = "schema/postgres"
 )
 
 func getTestClusterOption() *pt.TestBaseOptions {
+	testUser := "postgres"
+	testPassword := "cadence"
+
+	if runtime.GOOS == "darwin" {
```
yux0, line 42: Does it have to specify the os?

common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
sheerun: On OSX the default for postgresql user is USER, POSTGRES_USER is unset. Maybe you want me to use POSTGRES_USER with fallback to USER?

common/persistence/sql/sqlplugin/postgres/postgres_server_test.go
sheerun: Yes, because on OSX the default for postgresql user is USER which is not the case on other OSes, eg. on Ubuntu the default is "postgres" regardless of user.

sheerun:
I hope it's acceptable now

PR #3250, opened 2020-05-11
Add scanner invariants

andrewjdawson2016:
In this change I added three invariants to executions scanner. The first invariant checks that history exists, the next checks that the first event in history is valid and the last checks that if the workflow execution is open there is a current mutable state which points at it. In addition to defining these invariants I also fixed an off by one error in historyVersions. In addition this adds TreeID and BranchID to the executions written by scanner this will aid in debugging. Finally this diff makes sure the persistence iterator and blobstore iterator always return a valid execution so that way invariants checks can assume execution is valid form.


service/worker/scanner/executions/common/util.go
```diff
@@ -0,0 +1,169 @@
...
+	switch err.(type) {
+	case *shared.EntityNotExistsError:
+		return false, nil
+	default:
+		return false, err
+	}
+}
+
+// Open returns true if workflow state is open false if workflow is closed
+func Open(state int) bool {
```
yycptt, line 134: nit: `isOpen()`?

service/worker/scanner/executions/invariants/openCurrentExecution.go
```diff
@@ -0,0 +1,105 @@
...
+			InfoDetails:     fmt.Sprintf("current points at %v", currentExecResp.RunID),
+		}
+	}
+	return common.CheckResult{
+		CheckResultType: common.CheckResultTypeHealthy,
+	}
+}
+
+func (o *openCurrentExecution) Fix(execution common.Execution) common.FixResult {
+	return common.DeleteExecution(&execution, o.pr)
```
yycptt, line 100: Do we need to also delete history? Same question for `validFirstEvent` check. Probably add a TODO if we will do this in the future.

PR #3252, opened 2020-05-11
Fix reset reapply

yux0:
**What changed?**
1. Fix check slice length bug
2. Check max signal count on re-application

**Why?**
This is to fix the feedback loop on reset and re-apply

**How did you test it?**
Local test

**Potential risks**
This feedback loop issue is not fixed and causes backlog in replication stack

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2986,7 +2986,7 @@ func (e *historyEngineImpl) ReapplyEvents(
 		currentExecution,
 		func(wfContext execution.Context, mutableState execution.MutableState) (*updateWorkflowAction, error) {
 			// Filter out reapply event from the same cluster
-			var toReapplyEvents []*workflow.HistoryEvent
+			toReapplyEvents := make([]*workflow.HistoryEvent, 0, len(reapplyEvents))
 			lastWriteVersion, err := mutableState.GetLastWriteVersion()
 			if err != nil {
 				return nil, err
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2986,7 +2986,7 @@ func (e *historyEngineImpl) ReapplyEvents(
 		currentExecution,
 		func(wfContext execution.Context, mutableState execution.MutableState) (*updateWorkflowAction, error) {
 			// Filter out reapply event from the same cluster
-			toReapplyEvents := make([]*workflow.HistoryEvent, len(reapplyEvents))
+			var toReapplyEvents []*workflow.HistoryEvent
 			lastWriteVersion, err := mutableState.GetLastWriteVersion()
 			if err != nil {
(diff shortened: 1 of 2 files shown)


service/history/historyEngine.go
```diff
@@ -3009,6 +3009,21 @@ func (e *historyEngineImpl) ReapplyEvents(
 				}, nil
 			}
 
+			// Check if max signal count reaches
+			// Currently, we only re apply signal events
+			executionInfo := mutableState.GetExecutionInfo()
+			maxAllowedSignals := e.config.MaximumSignalsPerExecution(domainEntry.GetInfo().Name)
+			if maxAllowedSignals > 0 && (int(executionInfo.SignalCount)+len(toReapplyEvents)) >= maxAllowedSignals {
+				e.logger.Info("Execution limit reached for maximum signals during signal re-application",
```
yycptt, line 3017: This seems like a quite critical error, as it's non-retryable and the passive side doesn't even know the reapply has failed. Shall we use `Error` level log and probably emit some metrics?

service/history/historyEngine_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2017 - 2020 Uber Technologies, Inc.
```
yycptt, line 1: Let's follow the format in the license file: `Copyright (c) 2017-2020 Uber Technologies Inc.`

service/history/historyEngine.go
wxing1292: signal reapplication should ignore this limit.
or at least return error

service/history/historyEngine.go
```diff
@@ -2986,7 +2986,7 @@ func (e *historyEngineImpl) ReapplyEvents(
 		currentExecution,
 		func(wfContext execution.Context, mutableState execution.MutableState) (*updateWorkflowAction, error) {
 			// Filter out reapply event from the same cluster
-			toReapplyEvents := make([]*workflow.HistoryEvent, len(reapplyEvents))
+			var toReapplyEvents []*workflow.HistoryEvent
```
wxing1292, line 2989: add a `0` in the middle, for less mem allocation
```go
toReapplyEvents := make([]*workflow.HistoryEvent, 0, len(reapplyEvents))
```

yux0: updated

service/history/historyEngine_test.go
yux0: updated

service/history/historyEngine.go
yux0: Remove this check in this PR. If it returns an error, the caller side has. to handle this error or it will retry forever. I will include it in a different PR.

PR #3255, opened 2020-05-11
Populate history response to fix NPE on client side

mkolodezny:
https://github.com/uber/cadence/issues/3115


service/history/historyEngine.go
```diff
@@ -1580,11 +1580,16 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(
 			response.ScheduledEvent = scheduledEvent
 			response.ScheduledTimestampOfThisAttempt = common.Int64Ptr(ai.ScheduledTime.UnixNano())
 
+			response.StartedTimestamp = common.Int64Ptr(ai.StartedTime.UnixNano())
```
meiliang86, line 1583: StartedTimestamp can be updated inside AddActivityTaskStartedEvent. Other fields seems to be fine.

PR #3256, opened 2020-05-12
Fix list task list partition issue in matching

meiliang86:
**What changed?**
Fixing issue for  in matching. thrift has issue marshaling empty lists.

**Why?**
List tasklist partition was broken hence the fix.

**How did you test it?**
On local env.

**Potential risks**
Risk is low as this method is only used by CLI for debugging purpose.


service/matching/matchingEngine.go
```diff
@@ -609,11 +612,8 @@ func (e *matchingEngineImpl) listTaskListPartitions(request *m.ListTaskListParti
...
 		return nil, err
 	}
-	partitionHostInfo := make([]*workflow.TaskListPartitionMetadata, len(partitions))
 
-	if err != nil {
-		return nil, err
-	}
+	var partitionHostInfo []*workflow.TaskListPartitionMetadata
 	for _, partition := range partitions {
 		if host, err := e.getHostInfo(partition); err != nil {
```
markmark206, line 618: @anish531213 , @meiliang86 – I am curious about the **` !=`** in `err != nil`.

This blocks adds a new `TaskListPartitionMetadata` object when `getHostInfo()` *fails*, and I am unclear about the intention here.

Thank you for any insights you might have here!

PR #3260, opened 2020-05-13
Implement task processing queue collection

yycptt:
**What changed?**
Implement task processing queue collection, which manages a set of non-overlapping processing queues

**Why?**
Part of multi-cursor queue implementation

**How did you test it?**
Unit test

**Potential risks**
N/A, implementation not used yet.


service/history/queue/processing_queue.go
```diff
@@ -221,19 +221,25 @@ func (q *processingQueueImpl) Merge(
...
-	newQueueStates = append(newQueueStates, newProcessingQueueState(
+	overlappingQueueState := newProcessingQueueState(
 		q1.state.level,
 		maxTaskKey(q1.state.ackLevel, q2.state.ackLevel),
 		minTaskKey(q1.state.readLevel, q2.state.readLevel),
 		minTaskKey(q1.state.maxLevel, q2.state.maxLevel),
 		q1.state.domainFilter.Merge(q2.state.domainFilter),
-	))
+	)
+	if overlappingQueueState.readLevel.Less(overlappingQueueState.ackLevel) {
```
yux0, line 231: should this handle during initializing the queue state?

service/history/queue/processing_queue_collection.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+// NewProcessingQueueCollection creates a new collection for non-overlapping queues
+func NewProcessingQueueCollection(
+	level int,
+	queues []ProcessingQueue,
+) ProcessingQueueCollection {
+	sortProcessingQueue(queues)
+	queueCollection := &processingQueueCollection{
+		level:  level,
```
yux0, line 44: What does this level mean?

service/history/queue/processing_queue.go
yycptt: Updated. Moved the logic into queue state initialization.

service/history/queue/processing_queue_collection.go
yycptt: Consider it as the collectionID. All the queues inside a queue collection will have the same `level` which equals to the level of the collection (collectionID).

PR #3270, opened 2020-05-20
Adding failover marker replication task in thrift and persistence struct

yux0:
**What changed?**
Adding failover marker thrift struct and persistence struct
IDL PR: https://github.com/uber/cadence-idl/pull/31

**Why?**
To support graceful failover

**How did you test it?**
Tests will be added when there is business logic

**Potential risks**
N/A


service/history/historyEngine.go
```diff
@@ -317,7 +317,7 @@ func (e *historyEngineImpl) Start() {
 	clusterMetadata := e.shard.GetClusterMetadata()
 	if e.replicatorProcessor != nil &&
 		(clusterMetadata.GetReplicationConsumerConfig().Type != sconfig.ReplicationConsumerTypeRPC ||
-		e.config.EnableKafkaReplication()) {
+			e.config.EnableKafkaReplication()) {
```
yycptt, line 320: Revert the change?

common/persistence/dataInterfaces.go
```diff
@@ -605,6 +606,15 @@ type (
 		ScheduledID         int64
 	}
 
+	// FailoverMarkerTask is the marker for graceful failover
+	FailoverMarkerTask struct {
+		TaskID              int64
+		VisibilityTimestamp time.Time
+		Version             int64
+		SourceCluster       string
```
yycptt, line 614: Does the underlying DB schema has a field for the `sourceCluster`?

yux0: No. We don't have. Currently, I just re-use a string field.

service/history/historyEngine.go
yux0: It is done by the formater

common/persistence/dataInterfaces.go
yycptt: I see. Probably add some comments then?

yux0: I will update the schema in the following PR.

PR #3271, opened 2020-05-20
Fix shard instability

yycptt:
**What changed?**
1. Currently, acquire shard happens after a shard item is created. If acquire shard fails, the same shard item can be used again to acquire the shard. This means, one shard item can be associated with multiple shard context. In the case where the first acquire shard failed but the second one succeed, the close shard callback can potentially remove the shard item from shard controller. This diff invalidate the shard item if acquire shard fails and force a new shard item to be created for the next acquire shard call.

2. Current `removeEngineForShard` will delete the shard item from shard controller even if the current shard item is not the one we intend to delete. This diff fixes that.

**Why?**
Fix the continuous shard movement problem

**How did you test it?**
Unit test and tested locally

**Potential risks**
In the worst case, the shard movement issue will keep happening.


service/history/shard/controller.go
```diff
@@ -234,13 +234,17 @@ func (c *controller) ShardIDs() []int32 {
...
 	sw := c.metricsScope.StartTimer(metrics.RemoveEngineForShardLatency)
 	defer sw.Stop()
-	item, _ := c.removeHistoryShardItem(shardID)
-	// the shardItem comparison is just a defensive check to make sure we are deleting
-	// what we intend to delete. In the event that multiple callers call removeEngine / getEngine
-	// concurrently, it is possible to reorder a delete/delete/add sequence into a delete/add/delete
-	// sequence. This check is to protect against those scenarios.
-	if item != nil && (item == shardItem || shardItem == nil) {
-		item.stopEngine()
+	currentShardItem, _ := c.removeHistoryShardItem(shardID, shardItem)
```
vancexu, line 237: could we simplify this as 
```
currentShardItem, err := c.removeHistoryShardItem(shardID, shardItem)
if err != nil {
  if shardItem != nil {shardItem.stopEngine}
} else {
if currentShardItem != nil {
 currentShardItem.stopEngine()
}
}
```

service/history/shard/controller.go
```diff
@@ -299,22 +303,28 @@ func (c *controller) getOrCreateHistoryShardItem(shardID int) (*historyShardsIte
...
+
 	delete(c.historyShards, shardID)
 	nShards = len(c.historyShards)
-	c.Unlock()
 
 	c.metricsScope.IncCounter(metrics.ShardItemRemovedCounter)
 
-	shardItem.logger.Info("", tag.LifeCycleStopped, tag.ComponentShardItem, tag.Number(int64(nShards)))
-	return shardItem, nil
+	currentShardItem.logger.Info("", tag.LifeCycleStopped, tag.ComponentShardItem, tag.Number(int64(nShards)))
```
vancexu, line 326: is that possible currentShardItem be nil? I saw check on L246

yycptt: It won't be nil here. The check on L246 is for the cases where this function returns an error (L313 and L318)

PR #3273, opened 2020-05-20
Implement simple RWMutex cache to be used in domainCache

andrewjdawson2016:
We are facing an issue in which domain specific metrics cannot be emitted in task processing. We suspect that this is due the lock contention associated with using a LRU cache. LRU cache grabs a write lock every time Get is called so that it can move elements to the front of the cache. This is not actually needed in the case of domain cache because we never expect to evict entries from domain cache.

### DIFF
--- a/common/cache/simple.go
+++ b/common/cache/simple.go
@@ -143,7 +143,15 @@ func (c *simple) Put(key interface{}, value interface{}) interface{} {
 func (c *simple) PutIfNotExist(key interface{}, value interface{}) (interface{}, error) {
 	c.Lock()
 	defer c.Unlock()
-	return c.putInternal(key, value, false)
+	existing, err := c.putInternal(key, value, false)
+	if err != nil {
+		return nil, err
+	}
+	if existing == nil {
+		// This is a new value
+		return value, nil
+	}
+	return existing, nil
 }
 
 // Delete deletes a key, value pair associated with a key
--- a/common/cache/simple.go
+++ b/common/cache/simple.go
@@ -96,8 +96,6 @@ func (e *simpleEntry) CreateTime() time.Time {
 // Simple cache also does not have the concept of pinning that LRU cache has.
 // Internally simple cache uses a RWMutex instead of the exclusive Mutex that LRU cache uses.
 // The RWMutex makes simple cache readable by many threads without introducing lock contention.
-// The maxSize provided to simple cache should be larger than the possible largest size because
-// simple cache will simply panic if max size is exceeded.
 func NewSimple(opts *SimpleOptions) Cache {
 	if opts == nil {
 		opts = &SimpleOptions{}
@@ -122,7 +120,6 @@ func (c *simple) Get(key interface{}) interface{} {
 }
 
 // Put puts a new value associated with a given key, returning the existing value (if present).
-// If Put is called on a cache which is already full it will panic.
 func (c *simple) Put(key interface{}, value interface{}) interface{} {
 	c.Lock()
 	defer c.Unlock()
--- a/common/cache/simple.go
+++ b/common/cache/simple.go
@@ -24,17 +24,13 @@ package cache
 
 import (
 	"container/list"
-	"errors"
 	"sync"
 	"time"
 )
 
 var (
-	// SimpleCacheFull is the error that is returned from PutIfNotExists if cache is full
-	SimpleCacheFull = errors.New("simple cache is full")
-
 	// DummyCreateTime is the create time used by all entries in the cache.
-	DummyCreateTime = time.Now()
+	DummyCreateTime = time.Time{}
 )
 
 type (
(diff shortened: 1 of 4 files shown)


common/cache/simple.go
```diff
@@ -0,0 +1,212 @@
...
+	"sync"
+	"time"
+)
+
+var (
+	// SimpleCacheFull is the error that is returned from PutIfNotExists if cache is full
+	SimpleCacheFull = errors.New("simple cache is full")
+
+	// DummyCreateTime is the create time used by all entries in the cache.
+	DummyCreateTime = time.Now()
```
yux0, line 37: dummyCreateTime? what about use a fixed time

common/cache/simple.go
```diff
@@ -0,0 +1,212 @@
...
+	return e.value
+}
+
+// CreateTime is not implemented for simple cache entries
+func (e *simpleEntry) CreateTime() time.Time {
+	return DummyCreateTime
+}
+
+// NewSimple creates a new simple cache with given options.
+// Simple cache will never evict entries and it will never reorder the elements.
```
yux0, line 100: So if it hits the upper limit of the cache, does it mean all other domains(not load into the cache) will be blocked?

andrewjdawson2016: If cache fills up we will just start panicing.

yux0: Yes. Panic makes sense to me.

common/cache/simple.go
```diff
@@ -0,0 +1,212 @@
...
+
+// Size returns the number of entries currently in the cache
+func (c *simple) Size() int {
+	c.RLock()
+	defer c.RUnlock()
+
+	return len(c.accessMap)
+}
+
+func (c *simple) Iterator() Iterator {
```
meiliang86, line 184: The current way could work. Could also just return a channel. So we are not leaking the lock to the caller. Also can get rid of the list.

```
func (c *simple) Iterator() chan<- Entry {
	iterChan := make(chan Entry, 1000)

	go func() {
		c.RLock()
		defer c.RUnlock()
		for key, value := range c.accessMap {
			iterChan <- &simpleEntry{
				key:   key,
				value: value,
			}
		}
		close(iterChan)
	}()

	return iterChan
}
```

common/cache/simple.go
meiliang86: Why not just make this unbounded? This is for the expected unbounded use case so let it grow and hit potential oom is better than just panicing imo, as usually caller has no control on # of items to put to the cache in this case.

meiliang86: Also, worth adding a gauge metrics on actual cache size. Need to be careful on where to emit the metrics though.

common/cache/simple.go
andrewjdawson2016: We cannot unless we change the interface for Cache which will impact LRU cache as well - pretty big change.

common/cache/simple.go
yycptt: +1 for removing the size limit.

wxing1292:
According to my memory the domains in the cache are immutable and caches are swapped during update so just use a hashmap plus TTL should be fine?

common/cache/simple.go
andrewjdawson2016: okay changed this.

andrewjdawson2016:
@wxing1292 yes you are right - ideally it should just be a hashmap. The reason it is not is because domainCache is currently a messy abstraction its both a cache for id->name mapping and a method for registering callbacks for replication. Ideally domain cache should be separated out into two pieces one for replication and one that is simply a hashmap with TTL as you mentioned. But until this separation is done it is hard to simply pull out the cache into a hashmap. @yux0 has it on his plate at some point to refactor domain cache into these pieces at which point this simple cache is no longer needed for domain cache - but it might still be useful in other contents.

yux0:
> @wxing1292 yes you are right - ideally it should just be a hashmap. The reason it is not is because domainCache is currently a messy abstraction its both a cache for id->name mapping and a method for registering callbacks for replication. Ideally domain cache should be separated out into two pieces one for replication and one that is simply a hashmap with TTL as you mentioned. But until this separation is done it is hard to simply pull out the cache into a hashmap. @yux0 has it on his plate at some point to refactor domain cache into these pieces at which point this simple cache is no longer needed for domain cache - but it might still be useful in other contents.

After I took a closer look at the failover callback, I realized that the callback has to make sure domain cache does not update during the callback execution. I think it is the reason that the callback lives in the domain cache. I need more time to see how should we separate the them.

wxing1292:
According to my memory, you cannot separate the callback with the domain cache, which has to guarantee the domain cache visibility.

Using a hashmap should be the simplest way to solve potential nested locking issue

wxing1292:
I was going to use the hashmap instead of the LRU cache,  but did not found much time  to do it

common/cache/simple.go
meiliang86: I know. Not suggesting you to change it in this PR.

PR #3279, opened 2020-05-22
Add filter for pending active task to redispatch

yux0:
**What changed?**
Add a task filter to retry tasks when the domain is pending active

**Why?**
During graceful failover, the cluster should not process task as active

**How did you test it?**

**Potential risks**
N/A


service/history/taskAllocator.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/service/history/shard"
+	ctask "github.com/uber/cadence/service/history/task"
```
yycptt, line 31: drop the import alias?

yux0: the task is used as variable below

yycptt: I see. I use `ctask` in other places for the `common/task` package as `c` stands for `common`.

yux0: I see. Then I will try to find another alias

PR #3281, opened 2020-05-22
Bump feature version for client

vancexu:


yycptt:
My understanding is that feature version 1.6 (https://github.com/uber/cadence/pull/3199) is for the get raw history change which is not part of 0.12.x server release. It seems like there's a mistake when releasing the go-client 0.12.x and we mistakenly included more commits.

vancexu:
As discussed offline, releasing 0.12 from 0.12.x branch now will not impact open source users who updated go client to 0.12. 

(But one thing worth to mention is: when user use raw history feature from 0.12 client (which shouldn't really be included in 0.12), it is actually does nothing, since cadence server 0.12 is not really using enable raw history.)

PR #3284, opened 2020-05-27
Priority Task Processor Improvements

yycptt:
**What changed?**
1. Fix nil pointer exception when changing dynamic config for enabling priority task processor. This only happens when a shard is reloaded after the config is changed but before the host got restart.
2. Update task processing retry policy to have a max attempts of 3 before putting into redispatch queue. This is for preventing standby tasks from occupying a worker goroutine for too long. For active tasks, in most cases the task will be completed before its third attempt.
3. Update the logic for task priority assignment. Idea is to assign high priority to no-op tasks (tasks for standby domain submitted by active processor or vice versa) so that they won't get a low priority and block the ack level (especially for active processor) from advancing.

**Why?**
Bug fix and performance improvements.

**How did you test it?**
Unit test. Local canary.

**Potential risks**
N/A, feature flag disabled


service/history/task/priority_assigner.go
```diff
@@ -69,26 +69,32 @@ func NewPriorityAssigner(
...
+	// timer or transfer task, first check if task is active or not and if domain is active or not
+	isActiveTask := queueType == QueueTypeActiveTimer || queueType == QueueTypeActiveTransfer
 	domainName, active, err := a.getDomainInfo(queueTask.GetDomainID())
 	if err != nil {
 		return err
 	}
 
-	if !active {
+	if !isActiveTask && !active {
+		// only assign low priority to tasks submitted by standby processor for standby domains
```
yux0, line 87: nit: list out the four cases and maybe mention which cases are no-op?

yycptt: Done.

PR #3286, opened 2020-05-27
Scanner impl

andrewjdawson2016:
This adds a component to scanner which knows how to scan over all executions in a shard. It uses the primitives written in common and invariants package to do this. In addition to this there are a few other small changes 
- If flush is called on writer and writer is empty then do not trigger flush. This is needed so we can blindly call flush at the end of scan without getting an empty page. 
- Additionally Keys is converted to a pointer. This is done because if there are no corruptions or failures keys will be empty in this case we want keys to be nil.
- Additionally I include invariant type on the CheckResult this is useful in updating the InvariantCorruptionType map.


common/pagination/interface.go
```diff
@@ -66,6 +66,7 @@ type (
 	Writer interface {
 		Add(Entity) error
 		Flush() error
+		FlushIfNotEmpty() error
```
yycptt, line 69: Is there any situation where we want to `Flush` when it's empty?

service/worker/scanner/executions/common/types.go
```diff
@@ -147,20 +160,21 @@ type (
...
 	ShardFixReport struct {
 		ShardID int
-		Handled ShardFixHandled
+		Stats   ShardFixStats
 		Result  ShardFixResult
 	}
 
-	// ShardFixHandled indicates the executions which were handled by fix.
-	ShardFixHandled struct {
+	// ShardFixStats indicates the stats of executions where where handled by shard Fix.
```
yycptt, line 176: nit: Fix the typo?

service/worker/scanner/executions/shard/scanner.go
```diff
@@ -0,0 +1,141 @@
...
+		shardID:          shardID,
+		itr:              common.NewPersistenceIterator(pr, persistencePageSize, shardID),
+		failedWriter:     common.NewBlobstoreWriter(id, common.FailedExtension, blobstoreClient, blobstoreFlushThreshold),
+		corruptedWriter:  common.NewBlobstoreWriter(id, common.CorruptedExtension, blobstoreClient, blobstoreFlushThreshold),
+		invariantManager: invariants.NewInvariantManager(invariantCollections, pr),
+	}
+}
+
+// Scan scans over all executions in shard and runs invariant checks per execution.
+func (s *scanner) Scan() common.ShardScanReport {
```
yycptt, line 65: Is there any plan for adding logs & metrics for the scanner? or it will be done later in the workflow code?

andrewjdawson2016: All metrics and logs will exist at the workflow level.

common/pagination/interface.go
andrewjdawson2016: You potentially can use this pagination writer to flush a page which body is empty but has no empty tags.

service/worker/scanner/executions/common/types.go
andrewjdawson2016: fixed thank you

common/pagination/interface.go
andrewjdawson2016: Or said more generally the actual page contents might only be one input what is actually written.

PR #3290, opened 2020-05-28
Insert failover when domain moves from active to passive

yux0:
**What changed?**
Insert failover when domain moves from active to passive

**Why?**
Insert failover marker into replication queue and let the new active cluster consumes the markers.

**How did you test it?**
TODO

**Potential risks**
N/A


service/history/historyEngine.go
```diff
@@ -381,6 +381,28 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
...
+		if prevDomain != nil {
+			previousActiveCluster := prevDomain.GetReplicationConfig().ActiveClusterName
+			if prevDomain.IsGlobalDomain() &&
+				nextDomain.IsGlobalDomain() &&
+				domainFailoverNotificationVersion >= shardNotificationVersion &&
+				previousActiveCluster != domainActiveCluster &&
+				previousActiveCluster == e.currentClusterName {
+				return e.shard.InsertFailoverMarker(nextDomain.GetInfo().ID, nextDomain.GetFailoverVersion())
+			}
+		}
```
wxing1292, line 402: if during the failover host encounters restart, the `prevDomain` will be nil since there is no prev value in cache.
if the above situation happens, there should be handling of this host handling over the domain to the to-be active host / cluster.

service/history/historyEngine.go
```diff
@@ -402,10 +424,24 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
...
+				if err := failoverMarkerInsertion(
+					shardNotificationVersion,
+					prevDomains[idx],
+					nextDomain,
+				); err != nil {
+					e.logger.Error("Failed to insert failover marker to replication queue.",
+						tag.Error(err),
+						tag.WorkflowDomainName(nextDomain.GetInfo().Name))
+					e.metricsClient.IncCounter(metrics.HistoryFailoverMarkerScope, metrics.HistroyFailoverMarkerInsertFailure)
+					// fail this failover callback and it retries on next domain cache refresh
```
wxing1292, line 442: you may need to change the domain cache callback function to return err so domain cache can revert the changes and redo this callback.

one more thing, the hand over seems to relying on edge triggering, maybe it is necessary to have the change history of the domain?

service/history/historyEngine.go
yux0: Good point. So I am not going to use the previous domain entity. By checking the new domain entity failover end time should be able to tell if it should insert a marker

service/history/historyEngine.go
yux0: Return error is one approach. I updated the position to insert this marker. So it will insert the marker after it calls the failover domain function in task processor. In this way, the domain can be updated even if the insert operation fails.

service/history/historyEngine.go
wxing1292: Does the domain record contains failover from which cluster to which cluster?

yux0: No. Currently, I don't store this data. What is the case to store a history of failovers

service/history/historyEngine.go
```diff
@@ -436,6 +454,30 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
 				e.timerProcessor.NotifyNewTimers(e.currentClusterName, fakeDecisionTimeoutTask)
 			}
 
+			// handle graceful failover on active to passive
+			// make sure task processor failover the domain before inserting the failover marker
+			failoverMarkerTasks := []*persistence.FailoverMarkerTask{}
+			for _, nextDomain := range nextDomains {
+				failoverMarkerInsertion(shardNotificationVersion, nextDomain, func() {
```
yycptt, line 461: Do we still need this function? seems like the checks can be done inline.

service/history/shard/context.go
```diff
@@ -1170,6 +1172,35 @@ func (s *contextImpl) GetLastUpdatedTime() time.Time {
...
+	transferMaxReadLevel := int64(0)
+	if err := s.allocateTransferIDsLocked(
+		tasks,
+		&transferMaxReadLevel,
+	); err != nil {
+		return err
+	}
+	defer s.updateMaxReadLevelLocked(transferMaxReadLevel)
+
+	if err := s.executionManager.CreateFailoverMarkerTasks(&persistence.CreateFailoverMarkersRequest{
```
yycptt, line 1196: 1. Maybe add some retry logic here? 
2. I am not sure about this but do we need to ensure the shard ownership is not lost when inserting the marker tasks?

common/metrics/defs.go
```diff
@@ -170,7 +170,7 @@ const (
 	PersistenceDeleteReplicationTaskFromDLQScope
 	// PersistenceRangeDeleteReplicationTaskFromDLQScope tracks PersistenceRangeDeleteReplicationTaskFromDLQScope calls made by service to persistence layer
 	PersistenceRangeDeleteReplicationTaskFromDLQScope
-	// PersistenceCreateFailoverMakerTaskScope tracks CreateFailoverMakerTasks calls made by service to persistence layer
+	// PersistenceCreateFailoverMakerTasksScope tracks CreateFailoverMakerTasks calls made by service to persistence layer
```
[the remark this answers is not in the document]
yux0: Updated

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -5283,14 +5283,13 @@ func (s *ExecutionManagerSuite) TestReplicationDLQ() {
...
-			DomainID:            domainID,
-			Version:             1,
-		},
+	marker := &p.FailoverMarkerTask{
+		TaskID:              1,
+		VisibilityTimestamp: time.Now(),
+		DomainID:            domainID,
+		Version:             1,
 	}
+	markers := []*p.FailoverMarkerTask{marker}
```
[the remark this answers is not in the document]
yux0: updated

service/history/historyEngine.go
yux0: Updated

service/history/replicatorQueueProcessor.go
```diff
@@ -555,6 +555,9 @@ func (p *replicatorQueueProcessorImpl) toReplicationTask(
 			task.SourceTaskId = common.Int64Ptr(qTask.GetTaskID())
 		}
 		return task, err
+	case persistence.ReplicationTaskTypeFailoverMarker:
+		//TODO: add function to generate failover marker task
+		return nil, nil
```
yycptt, line 560: nit: return an error?

yux0: currently, just ignore the task or the tests will fail. I will wire this in the next PR

PR #3294, opened 2020-05-29
Make event cache size based

mkolodezny:
**What changed?**
Make event cache size based

**Why?**
events part of https://github.com/uber/cadence/issues/2350

**How did you test it?**
performance tests

**Potential risks**
history events related API latency


common/cache/lru.go
```diff
@@ -301,8 +324,28 @@ func (c *lru) deleteInternal(element *list.Element) {
...
 	}
 	delete(c.byKey, entry.key)
+	c.updateSizeInBytesOnDelete(entry.key)
 }
 
 func (c *lru) isEntryExpired(entry *entryImpl, currentTime time.Time) bool {
 	return entry.refCount == 0 && !entry.createTime.IsZero() && currentTime.After(entry.createTime.Add(c.ttl))
 }
+
+func (c *lru) checkSizeBytes() bool {
```
meiliang86, line 334: nit: rename checkSizeBytes -> isCacheWithinSizeLimit

common/cache/lru.go
```diff
@@ -131,14 +136,28 @@ func New(maxSize int, opts *Options) Cache {
 		opts = &Options{}
 	}
 
-	return &lru{
+	cache := &lru{
```
meiliang86, line 139: To me, size and count is either/or, meaning that if it's size based, we don't care about the count, as long as it stays within size limit.

common/cache/lru.go
```diff
@@ -131,14 +136,28 @@ func New(maxSize int, opts *Options) Cache {
 		opts = &Options{}
```
meiliang86, line 136: move maxSize into option as well?

meiliang86: Also add a check to make sure either count limit or size limit is specified.

common/service/dynamicconfig/constants.go
```diff
@@ -146,6 +146,7 @@ var keys = map[Key]string{
 	HistoryShutdownDrainDuration:                           "history.shutdownDrainDuration",
 	EventsCacheInitialSize:                                 "history.eventsCacheInitialSize",
 	EventsCacheMaxSize:                                     "history.eventsCacheMaxSize",
```
meiliang86, line 148: See if you can change this (and corresponding stuffs) to EventsCacheMaxCount.

common/service/dynamicconfig/constants.go
```diff
@@ -487,6 +488,8 @@ const (
 	EventsCacheInitialSize
 	// EventsCacheMaxSize is max size of events cache
 	EventsCacheMaxSize
+	// EventsCacheMaxSizeInKb is max size of events cache in kilobyte
```
meiliang86, line 491: nit: Comment and code mismatch...

service/history/config/config.go
```diff
@@ -61,6 +61,7 @@ type Config struct {
 	// Change of these configs require shard restart
 	EventsCacheInitialSize       dynamicconfig.IntPropertyFn
 	EventsCacheMaxSize           dynamicconfig.IntPropertyFn
+	EventsCacheMaxSizeInKb       dynamicconfig.IntPropertyFn
```
meiliang86, line 64: Byes or kb?

service/history/events/cache.go
```diff
@@ -93,16 +93,18 @@ func NewGlobalCache(
...
+		nil,
 		initialSize,
 		maxSize,
 		ttl,
 		historyManager,
 		false,
 		logger,
 		metricsClient,
-		nil,
+		uint64(maxSizeInKb*1024),
```
meiliang86, line 107: Can we unify and use either bytes or kb across all places?

common/util.go
```diff
@@ -35,6 +35,7 @@ import (
 
 	h "github.com/uber/cadence/.gen/go/history"
 	m "github.com/uber/cadence/.gen/go/matching"
+	"github.com/uber/cadence/.gen/go/shared"
 	workflow "github.com/uber/cadence/.gen/go/shared"
```
yycptt, line 39: `shared` package has already been imported.

common/util.go
```diff
@@ -537,6 +538,211 @@ func GetSizeOfMapStringToByteArray(input map[string][]byte) int {
 	return res + golandMapReserverNumberOfBytes
 }
 
+// GetSizeOfMapStringToByteArray returns approximate size of the history event taking into account byte arrays only now
+func GetSizeOfHistoryEvent(event *shared.HistoryEvent) uint32 {
```
yycptt, line 542: nit: please update comment to match the function name

service/history/events/cache.go
```diff
@@ -93,16 +93,18 @@ func NewGlobalCache(
 	historyManager persistence.HistoryManager,
 	logger log.Logger,
 	metricsClient metrics.Client,
+	maxSizeInKb uint,
```
yycptt, line 96: Is it possible to make the value type for size consistent? I think I've seen three types: `uint`, `uint32` and `uint64`.

common/util.go
```diff
@@ -537,6 +538,211 @@ func GetSizeOfMapStringToByteArray(input map[string][]byte) int {
 	return res + golandMapReserverNumberOfBytes
 }
 
+// GetSizeOfMapStringToByteArray returns approximate size of the history event taking into account byte arrays only now
```
vancexu, line 541: nit: comment name mismatch

common/util.go
```diff
@@ -537,6 +538,211 @@ func GetSizeOfMapStringToByteArray(input map[string][]byte) int {
...
+func GetSizeOfHistoryEvent(event *shared.HistoryEvent) uint32 {
+	if event == nil {
+		return 0
+	}
+
+	res := 0
+	switch *event.EventType {
+	case shared.EventTypeWorkflowExecutionStarted:
+		if event.WorkflowExecutionStartedEventAttributes == nil {
```
vancexu, line 550: we can assuming event type and attribute are guaranteed matched, so no need for such check. 
If this rule is violated, lots of place will NPE, just let it expose itself so team can quick fix.

service/history/resource/resource.go
```diff
@@ -72,6 +72,7 @@ func New(
 			serviceResource.GetHistoryManager(),
 			params.Logger,
 			params.MetricsClient,
+			uint(config.EventsCacheMaxSizeInKb()),
```
vancexu, line 75: would be easier just uint64 everywhere for cache size.
For config, it also might be better to also use bytes to be consistent with rest

common/cache/lru.go
mkolodezny: Agree, if maxSize is provided then it is size based only and count based otherwise. And if size based, lets just check for an upper limit for count to prevent the infinite growing, for safety sake.

service/history/config/config.go
```diff
@@ -265,12 +266,13 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
...
-		EventsCacheInitialSize:               dc.GetIntProperty(dynamicconfig.EventsCacheInitialSize, 128),
-		EventsCacheMaxSize:                   dc.GetIntProperty(dynamicconfig.EventsCacheMaxSize, 512),
+		EventsCacheInitialCount:              dc.GetIntProperty(dynamicconfig.EventsCacheInitialCount, 128),
+		EventsCacheMaxCount:                  dc.GetIntProperty(dynamicconfig.EventsCacheMaxCount, 512),
+		EventsCacheMaxSize:                   dc.GetIntProperty(dynamicconfig.EventsCacheMaxSize, 0),
 		EventsCacheTTL:                       dc.GetDurationProperty(dynamicconfig.EventsCacheTTL, time.Hour),
 		EventsCacheGlobalEnable:              dc.GetBoolProperty(dynamicconfig.EventsCacheGlobalEnable, false),
-		EventsCacheGlobalInitialSize:         dc.GetIntProperty(dynamicconfig.EventsCacheInitialSize, 4096),
-		EventsCacheGlobalMaxSize:             dc.GetIntProperty(dynamicconfig.EventsCacheMaxSize, 65536),
+		EventsCacheGlobalInitialCount:        dc.GetIntProperty(dynamicconfig.EventsCacheGlobalInitialCount, 1<<12),
```
meiliang86, line 274: I feel that 4096 is easier to read :)

PR #3299, opened 2020-05-29
Added metricScope cache implementation

anish531213:
**What changed?**
Added generic metricsScopeCache component that keeps a record of domain to metricsScope

**Why?**
Creating scope with domainTag is an expensive operation and caused latency in taskProcessing.

Have not verified yet, this is a draft for review. Will test this in single history host in pord.

This change might increase latency on task processing.

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -124,7 +124,7 @@ type Entry interface {
 type GetCacheItemSizeFunc func(interface{}) uint64
 type MetricsCache interface {
 	// Get retrieves metrics scope based on me
-	Get(string) metrics.Scope
+	Get(domainID string, taskType int) metrics.Scope
 	// Put adds metrics scope for a domainID
-	Put(string, metrics.Scope)
+	Put(domainID string, taskType int, metricsScope metrics.Scope)
 }
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -21,6 +21,7 @@
 package cache
 
 import (
+	"github.com/uber/cadence/common/metrics"
 	"time"
 )
 
@@ -121,3 +122,9 @@ type Entry interface {
 
 // GetCacheItemSizeFunc returns the cache item size in bytes
 type GetCacheItemSizeFunc func(interface{}) uint64
+type MetricsCache interface {
+	// Get retrieves metrics scope based on me
+	Get(string) metrics.Scope
+	// Put adds metrics scope for a domainID
+	Put(string, metrics.Scope)
+}
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -21,8 +21,9 @@
 package cache
 
 import (
-	"github.com/uber/cadence/common/metrics"
 	"time"
+
+	"github.com/uber/cadence/common/metrics"
 )
 
 // A Cache is a generalized interface to a cache.  See cache.LRU for a specific
@@ -122,9 +123,10 @@ type Entry interface {
 
 // GetCacheItemSizeFunc returns the cache item size in bytes
 type GetCacheItemSizeFunc func(interface{}) uint64
-type MetricsCache interface {
-	// Get retrieves metrics scope based on me
-	Get(domainID string, taskType int) metrics.Scope
-	// Put adds metrics scope for a domainID
-	Put(domainID string, taskType int, metricsScope metrics.Scope)
+
+type MetricsScopeCache interface {
+	// Get retrieves metrics scope for a domainID and scopeID
+	Get(domainID string, scopeID int) (metrics.Scope, bool)
+	// Put adds metrics scope for a domainID and scopeID
+	Put(domainID string, scopeID int, metricsScope metrics.Scope)
 }
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -125,8 +125,8 @@ type Entry interface {
 type GetCacheItemSizeFunc func(interface{}) uint64
 
 type MetricsScopeCache interface {
-	// Get retrieves metrics scope for a domainID and scopeID
-	Get(domainID string, scopeID int) (metrics.Scope, bool)
-	// Put adds metrics scope for a domainID and scopeID
-	Put(domainID string, scopeID int, metricsScope metrics.Scope)
+	// Get retrieves metrics scope for a domainID and scopeIdx
+	Get(domainID string, scopeIdx int) (metrics.Scope, bool)
+	// Put adds metrics scope for a domainID and scopeIdx
+	Put(domainID string, scopeIdx int, metricsScope metrics.Scope)
 }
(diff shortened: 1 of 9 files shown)


common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,60 @@
...
+
+package cache
+
+import (
+	"sync"
+
+	"github.com/uber/cadence/common/metrics"
+)
+
+type MetricsCache struct {
```
andrewjdawson2016, line 31: Personally I prefer to always use interfaces rather than raw structs.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,60 @@
...
+	sync.RWMutex
+	scopeMap map[string]metrics.Scope
+}
+
+// NewMetricsCache creates a new metricsCache struct
```
andrewjdawson2016, line 36: "NewMetricCache constructs a new MetricsCache

service/history/taskProcessor.go
```diff
@@ -246,9 +249,20 @@ func (t *taskProcessor) processTaskOnce(
...
+	domainID := task.task.GetDomainID()
+	scope := t.metricsScopeCache.Get(domainID)
+
 	startTime := t.timeSource.Now()
-	scopeIdx, err := task.processor.process(task)
-	scope := t.metricsClient.Scope(scopeIdx).Tagged(t.getDomainTagByID(task.task.GetDomainID()))
+
+	if scope == nil {
+		scopeIdx, err = task.processor.process(task)
+		scope := t.metricsClient.Scope(scopeIdx).Tagged(t.getDomainTagByID(domainID))
```
andrewjdawson2016, line 262: I think you mean `scope = ...` rather than `scope := ...`

service/history/taskProcessor.go
```diff
@@ -246,9 +249,20 @@ func (t *taskProcessor) processTaskOnce(
...
+
+	domainID := task.task.GetDomainID()
+	scope := t.metricsScopeCache.Get(domainID)
+
 	startTime := t.timeSource.Now()
-	scopeIdx, err := task.processor.process(task)
-	scope := t.metricsClient.Scope(scopeIdx).Tagged(t.getDomainTagByID(task.task.GetDomainID()))
+
+	if scope == nil {
+		scopeIdx, err = task.processor.process(task)
```
andrewjdawson2016, line 261: So we only process if scope is nil?

yycptt: The code is a little bit confusing here, `process()` does two things: 1. process the task, 2. return a metric scope ID that should be used the for all the metrics for this task. 

So we always need to process the task, regardless of if the scope exists in the cache or not.

service/history/taskProcessor.go
```diff
@@ -246,9 +249,20 @@ func (t *taskProcessor) processTaskOnce(
 	default:
 	}
 
+	var scopeIdx int
+	var err error
+
+	domainID := task.task.GetDomainID()
+	scope := t.metricsScopeCache.Get(domainID)
```
yycptt, line 256: We have different scopes for the different task types, so `domainID` alone can't uniquely identify a metric scope.

service/history/taskProcessor.go
```diff
@@ -246,9 +248,22 @@ func (t *taskProcessor) processTaskOnce(
...
 	}
 
+	var scopeIdx int
+	var err error
+
+	domainID := task.task.GetDomainID()
+	taskType := task.task.GetTaskType()
+
+	scopeIdx, err = task.processor.process(task)
+	scope, found := t.metricsScopeCache.Get(domainID, scopeIdx)
```
andrewjdawson2016, line 257: Looks better :)

andrewjdawson2016: But lets call this `scope, ok`

andrewjdawson2016: More standard go.

common/resource/resourceTest.go
```diff
@@ -176,6 +177,7 @@ func NewTest(
 		// other common resources
 
 		DomainCache:       cache.NewMockDomainCache(controller),
+		MetricsScopeCache: cache.NewMetricsScopeCache(),
```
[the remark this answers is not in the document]
yycptt: I think it's probably better to use a real implementation rather than a mock here. Otherwise we will have to specify many mock calls for this cache in the test.

common/cache/cache.go
```diff
@@ -104,3 +106,10 @@ type Entry interface {
 	// CreateTime represents the time when the entry is created
 	CreateTime() time.Time
 }
+
+type MetricsScopeCache interface {
```
yycptt, line 110: nit: exported types should have comments

service/history/taskProcessor.go
```diff
@@ -246,9 +248,21 @@ func (t *taskProcessor) processTaskOnce(
...
+	domainID := task.task.GetDomainID()
+
+	scopeIdx, err = task.processor.process(task)
+	scope, found := t.metricsScopeCache.Get(domainID, scopeIdx)
+
 	startTime := t.timeSource.Now()
-	scopeIdx, err := task.processor.process(task)
-	scope := t.metricsClient.Scope(scopeIdx).Tagged(t.getDomainTagByID(task.task.GetDomainID()))
+
+	if !found {
```
yycptt, line 261: nit: move this to L258

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Get retrieves scope for domainID and scopeIdx
+func (c *domainMetricsScopeCache) Get(domainID string, scopeIdx int) (metrics.Scope, bool) {
+	c.RLock()
+	defer c.RUnlock()
+
+	var buffer bytes.Buffer
+	buffer.WriteString(domainID)
+	buffer.WriteString("_")
+	buffer.WriteString(strconv.Itoa(scopeIdx))
```
yycptt, line 53: nit: create to util function to avoid code duplication.

service/history/taskProcessor.go
```diff
@@ -246,9 +248,21 @@ func (t *taskProcessor) processTaskOnce(
...
+	var scopeIdx int
+	var err error
+
+	domainID := task.task.GetDomainID()
+
+	scopeIdx, err = task.processor.process(task)
+	scope, found := t.domainMetricsScopeCache.Get(domainID, scopeIdx)
+
+	if !found {
+		scope = t.metricsClient.Scope(scopeIdx).Tagged(t.getDomainTagByID(domainID))
```
yycptt, line 260: We should also modify `getDomainTagByID()` to return an error if it can't get domain name. And don't cache the scope in that case.

meiliang86:
Be careful this cache can go unbounded. And given the # of scopes and # of domains here it will be quite big in production.

PR #3304, opened 2020-06-01
Persist failover marker in shard info from replication queue

yux0:
**What changed?**
Persist failover marker in shard info from replication queue

IDL change: https://github.com/uber/cadence-idl/pull/35

**Why?**
This change is for passive -> active cluster (domain level). After the cluster received the failover marker, it update the shard info. The shard probes the marker to the coordinator within X seconds. After X seconds, it will remove the markers from the shard.
The probe part will be included in the coordinator PR.

The pending failover marker will store in the DB as a data blob.

**How did you test it?**
unit tests

**Potential risks**
The major concern is to see if the data will slow down update shard operation.


service/history/shard/context.go
```diff
@@ -160,6 +164,7 @@ const (
 	logWarnTransferLevelDiff = 3000000 // 3 million
 	logWarnTimerLevelDiff    = time.Duration(30 * time.Minute)
 	historySizeLogThreshold  = 10 * 1024 * 1024
+	failoverMarkerRetention  = 2 * time.Minute
```
yycptt, line 167: Is this used?

service/history/failover/marker_notifier.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+
+			if len(markers) > 0 {
+				respCh := m.failoverCoordinator.NotifyFailoverMarkers(int32(m.shard.GetShardID()), markers)
+				select {
+				case <-m.shutdownCh:
+					return
+				case <-respCh:
+					// continue the next round to notify failover markers
+					continue
```
yycptt, line 117: Do we need to sleep in this case?

service/history/replication/task_executor.go
```diff
@@ -103,6 +106,9 @@ func (e *taskExecutorImpl) execute(
 	case r.ReplicationTaskTypeHistoryV2:
 		scope = metrics.HistoryReplicationV2TaskScope
 		err = e.handleHistoryReplicationTaskV2(replicationTask, forceApply)
+	case r.ReplicationTaskTypeFailoverMarker:
+		scope = metrics.HistoryReplicationV2TaskScope
```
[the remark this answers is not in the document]
yux0: Updated

service/history/failover/coordinator.go
```diff
@@ -122,30 +122,38 @@ func NewCoordinator(
...
+		common.DaemonStatusStarted,
+	) {
 		return
 	}
 
 	go c.receiveFailoverMarkersLoop()
 	go c.notifyFailoverMarkerLoop()
 
-	c.logger.Info("Failover coordinator started.")
+	c.logger.Info("Failover coordinator started.", tag.LifeCycleStarted)
```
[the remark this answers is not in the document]
yux0: Updated all

service/history/shard/context.go
yux0: Deleted

service/history/failover/marker_notifier.go
yux0: No. The coordinator controls when to return the response and it will schedule next notification

PR #3306, opened 2020-06-02
Timer queue processor base v2

yycptt:
**What changed?**
* Rewrite timer queue processor base with multi-cursor queue components
* Create util functions for common parts in timer and transfer queue processor base
* Use struct instead of pointer as timer/transfer task key
* Move timerGate to queue package

**Why?**
Part of multi-cursor queue implementation

**How did you test it?**
Unite test

**Potential risks**
N/A, implementation not wired up yet


service/history/queue/queue_processor_util.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return processingQueueCollections
+}
+
+// RedispatchTasks should be un-exported after the queue processing logic
+// in history package is deprecated.
+func RedispatchTasks(
+	redispatchQueue collection.Queue,
+	taskProcessor task.Processor,
+	logger log.Logger,
```
yux0, line 89: why it needs to pass in the logger

service/history/queue/timer_queue_processor_base.go
```diff
@@ -0,0 +1,590 @@
...
+			RedispatchTasks(
+				t.redispatchQueue,
+				t.taskProcessor,
+				t.logger,
+				t.metricsScope,
+				t.shutdownCh,
+			)
+			t.notifyNewTimer(time.Time{})
+		case <-pollTimer.C:
+			pollTimer.Reset(backoff.JitDuration(
```
yux0, line 224: should those timer resets happen after the actions?

service/history/queue/queue_processor_util.go
yycptt: Because I'd like to emit a log when re-dispatch failed and different queue processors are using different loggers (log tags are different).

service/history/queue/timer_queue_processor_base.go
yycptt: Yes, makes sense. I will also update the transfer queue processor.

yycptt: Updated.

PR #3307, opened 2020-06-02
Scanner and fixer workflow implementations

andrewjdawson2016:
This diff does all the wiring up of the scanner and fixer workflow using lower level primitives written earlier. 

I have verified scanner locally and everything is unit tested. 

Worst case is this consumes persistence resources and slows down worker (the replicator role is most important in worker). If this happens we can just use dynamic config to disable.


service/worker/scanner/scanner.go
```diff
@@ -123,25 +118,42 @@ func New(
...
+	backgroundActivityContext := context.WithValue(context.Background(), scannerContextKey, s.context)
+	backgroundActivityContext = context.WithValue(backgroundActivityContext, executions.ScannerContextKey, executions.ScannerContext{
+		Resource:                     s.context.Resource,
+		Scope:                        s.context.Resource.GetMetricsClient().Scope(metrics.ExecutionsScannerScope),
+		ScannerWorkflowDynamicConfig: s.context.cfg.ExecutionScannerConfig,
+	})
+	backgroundActivityContext = context.WithValue(backgroundActivityContext, executions.FixerContextKey, executions.FixerContext{
+		Resource: s.context.Resource,
+		Scope:    s.context.Resource.GetMetricsClient().Scope(metrics.ExecutionsFixerScope),
+	})
```
yycptt, line 130: nit: only set the scanner context when scanner is enabled? `ExecutionsScannerEnabled` flag requires restart anyway I think?

service/worker/scanner/executions/util.go
```diff
@@ -0,0 +1,89 @@
...
+			ExpirationInterval: 5 * time.Minute,
+		},
+	}
+	return workflow.WithActivityOptions(ctx, activityOptions)
+}
+
+func getLongActivityContext(ctx workflow.Context) workflow.Context {
+	activityOptions := workflow.ActivityOptions{
+		ScheduleToStartTimeout: time.Minute,
+		StartToCloseTimeout:    time.Minute,
```
yycptt, line 80: one minute for long running activity?

service/worker/scanner/executions/workflows.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	resolvedConfig := resolveFixerConfig(params.FixerWorkflowConfigOverwrites)
+	shardReportChan := workflow.GetSignalChannel(ctx, fixShardReportChan)
+	for i := 0; i < resolvedConfig.Concurrency; i++ {
+		idx := i
+		workflow.Go(ctx, func(ctx workflow.Context) {
+			for j, corrupted := range corruptedKeysResult.CorruptedKeys {
+				if j%resolvedConfig.Concurrency == idx {
+					activityCtx = getLongActivityContext(ctx)
```
yycptt, line 272: I think if the scanner workflow is still running, that error should be treated as non-retryable?

service/worker/scanner/executions/aggregators.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (a *shardFixResultAggregator) adjustAggregation(stats common.ShardFixStats, fn func(a, b int64) int64) {
+	a.aggregation.ExecutionCount = fn(a.aggregation.ExecutionCount, stats.ExecutionCount)
+	a.aggregation.SkippedCount = fn(a.aggregation.SkippedCount, stats.SkippedCount)
+	a.aggregation.FailedCount = fn(a.aggregation.FailedCount, stats.FailedCount)
+	a.aggregation.FixedCount = fn(a.aggregation.FixedCount, stats.FixedCount)
+}
+
+type shardScanResultAggregator struct {
```
yycptt, line 91: nit: define all the types at the beginning?

common/metrics/defs.go
```diff
@@ -1642,6 +1645,9 @@ const (
 	CadenceErrAuthorizeFailedPerTaskListCounter
 	CadenceErrRemoteSyncMatchFailedPerTaskListCounter
 
+	CadenceShardSuccessGauge
+	CadenceShardFailureGauge
```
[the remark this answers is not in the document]
andrewjdawson2016: I don't think these metrics need to be specific to scanner though? In theory shardManager might want to use these.

service/worker/scanner/executions/activities.go
```diff
@@ -0,0 +1,290 @@
...
+		ScannerWorkflowRunID      string
+	}
+
+	// FixShardActivityParams is the parameter for FixShardActivity
+	FixShardActivityParams struct {
+		CorruptedKeysEntry          CorruptedKeysEntry
+		ResolvedFixerWorkflowConfig ResolvedFixerWorkflowConfig
+	}
+
+	FixerCorruptedKeysActivityResult struct {
```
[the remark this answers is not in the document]
andrewjdawson2016: Added in follow up diff.

service/worker/scanner/executions/activities.go
```diff
@@ -0,0 +1,290 @@
...
+		CorruptedKeys []CorruptedKeysEntry
+		Shards        []int
```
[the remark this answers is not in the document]
andrewjdawson2016: You are right, it's not needed. It's currently being used to constructor aggregator for fixer but I can just pass corruptedKeys. Will remove in follow up diff.

service/worker/service.go
```diff
@@ -133,6 +134,16 @@ func NewConfig(params *service.BootstrapParams) *Config {
 			TaskListScannerEnabled:   dc.GetBoolProperty(dynamicconfig.TaskListScannerEnabled, true),
 			HistoryScannerEnabled:    dc.GetBoolProperty(dynamicconfig.HistoryScannerEnabled, true),
 			ExecutionsScannerEnabled: dc.GetBoolProperty(dynamicconfig.ExecutionsScannerEnabled, false),
+			ExecutionScannerConfig: &executions.ScannerWorkflowDynamicConfig{
+				Enabled:                 dc.GetBoolProperty(dynamicconfig.ExecutionsScannerEnabled, false),
```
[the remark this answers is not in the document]
andrewjdawson2016: Will address in next diff.

service/worker/scanner/scanner.go
andrewjdawson2016: Will address in next diff.

service/worker/scanner/executions/util.go
andrewjdawson2016: Retry policy overwrites this value.

service/worker/scanner/executions/workflows.go
andrewjdawson2016: Which line is this referring to? If this is referring to line 272 and beyond then at this point we already are sure the scan workflow is closed that is checked in FixerCorruptedKeysActivity (which is short running so if the workflow is still running it will return quickly).

yycptt: I should have commented in `activities.go` L219. There's a check for the workflow close status and return error if close status == nil. I think we should not retry in that case?

service/worker/scanner/executions/util.go
yycptt: Synced offline, this will be fixed in next pr.

service/worker/scanner/executions/workflows.go
yycptt: Synced offline. Will be addressed in next pr.

PR #3308, opened 2020-06-03
Add previous failover version in domain v2 table

yux0:
**What changed?**
Add previous failover version in domain v2 table

**Why?**
A history of failover is needed for graceful failover. The failover checks the history to insert failover marker on active->passive cluster

**How did you test it?**
Integration test

**Potential risks**
In the worst case, create domain, update domain and list domains operation fail


wxing1292:
@yux0 adding a prev failover version to domain indicates there can be at most one failover per domain at any given time, or otherwise the data will be overwritten.

maybe worth thinking about adding a table tracking domain changes history.

yux0:
> @yux0 adding a prev failover version to domain indicates there can be at most one failover per domain at any given time, or otherwise the data will be overwritten.
> 
> maybe worth thinking about adding a table tracking domain changes history.

This previous version is to let the active -> passive cluster insert failover marker. We have the assumption of no concurrent graceful failover. But I am not following the idea, how to maintain the history of domain changes if there are concurrent failover?

wxing1292:
> > @yux0 adding a prev failover version to domain indicates there can be at most one failover per domain at any given time, or otherwise the data will be overwritten.
> > maybe worth thinking about adding a table tracking domain changes history.
> 
> This previous version is to let the active -> passive cluster insert failover marker. We have the assumption of no concurrent graceful failover. But I am not following the idea, how to maintain the history of domain changes if there are concurrent failover?

1. 
> This previous version is to let the active -> passive cluster insert failover marker. 
I know, I mentioned this in the domain change cache notification change PR.

2. 
> We have the assumption of no concurrent graceful failover. 
What is the behavior of force failover during a "graceful" failover?

3. 
> But I am not following the idea, how to maintain the history of domain changes if there are concurrent failover?
Never thought this in details, just an idea.
Store the domain change history in each DC, like workflow history events, or kafka topics.
Each change to domain is an domain change history item, including domain update, domain failover (by API call, by replication).
The above can guarantee edge triggering functionalities (one example is handover).
Not sure whether this is an overkill for this case.

yux0:
> > > @yux0 adding a prev failover version to domain indicates there can be at most one failover per domain at any given time, or otherwise the data will be overwritten.
> > > maybe worth thinking about adding a table tracking domain changes history.
> > 
> > 
> > This previous version is to let the active -> passive cluster insert failover marker. We have the assumption of no concurrent graceful failover. But I am not following the idea, how to maintain the history of domain changes if there are concurrent failover?
> 
> 
> > This previous version is to let the active -> passive cluster insert failover marker.
> 
> I know, I mentioned this in the domain change cache notification change PR.
> 
> 
> > We have the assumption of no concurrent graceful failover.
> 
> What is the behavior of force failover during a "graceful" failover?
> 
The force failover will be honored and the "graceful" failover marker will be ignored.
> 
> > But I am not following the idea, how to maintain the history of domain changes if there are concurrent failover?
> 
> Never thought this in details, just an idea.
> Store the domain change history in each DC, like workflow history events, or kafka topics.
> Each change to domain is an domain change history item, including domain update, domain failover (by API call, by replication).
> The above can guarantee edge triggering functionalities (one example is handover).
> Not sure whether this is an overkill for this case.

At this point, I think it is an overkill for this case. But we can definitely revisit.

PR #3313, opened 2020-06-04
Size reduction of execution scanner workflow

andrewjdawson2016:
This diff reduces the sizes of the execution scanner. There are three sizes which were reduced: query sizes, activity sizes and history sizes. Query sizes were reduced by implementing pagination for query results. Activity sizes were reduced by using this query pagination. History sizes were reduced by patching multiple shards together in an activity. This additionally start heartbeating progress in scan and fix activities (because these are now handling multiple shards).


service/worker/scanner/executions/workflows.go
```diff
@@ -120,16 +123,47 @@ type (
...
+		LimitShards *int
+	}
+
+	// ShardQueryPaginationToken is used to return information used to make the next query
+	ShardQueryPaginationToken struct {
+		// NextShardID is one greater than the highest shard returned in the current query.
+		// NextShardID is nil if IsDone is true.
+		// It is possible to get NextShardID != nil and on the next call to get an empty result with IsDone = true.
+		NextShardID *int
+		IsDone      bool
```
yycptt, line 153: It seems like I only need to check `NextShardID != nil` to see if I need to keep calling? Then do we still need the `IsDone` field?

service/worker/scanner/executions/util.go
```diff
@@ -77,13 +110,84 @@ func getShortActivityContext(ctx workflow.Context) workflow.Context {
...
+	workerIdx int,
+) [][]int {
+	var batches [][]int
+	var currBatch []int
+	for i := 0; i < sliceLength; i++ {
+		if i%concurrency == workerIdx {
+			currBatch = append(currBatch, i)
+			if len(currBatch) == batchSize {
+				copyCurrBatch := make([]int, batchSize, batchSize)
+				copy(copyCurrBatch, currBatch)
```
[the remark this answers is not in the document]
andrewjdawson2016: not needed - removed.

service/worker/scanner/executions/activities.go
```diff
@@ -123,17 +136,46 @@ func ScannerEmitMetricsActivity(
...
+	}
+	for i := heartbeatDetails.LastShardIndexHandled + 1; i < len(params.Shards); i++ {
+		currentShardID := params.Shards[i]
+		shardReport, err := scanShard(activityCtx, params, currentShardID, heartbeatDetails)
+		if err != nil {
+			return nil, err
+		}
+		heartbeatDetails = ScanShardHeartbeatDetails{
+			LastShardIndexHandled: i,
+			Reports:               append(heartbeatDetails.Reports, *shardReport),
```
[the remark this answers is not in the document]
andrewjdawson2016: Reports are small - less than 1KB. So at a batch amount of 200 we are fine. If it proves to be too big we can just reduce the batch size.

service/worker/scanner/executions/workflows.go
andrewjdawson2016: This code doesn't use IsDone but I liked having IsDone = true for users manually querying the workflow.

yycptt: I see. Makes sense for end users.

PR #3321, opened 2020-06-06
Wire up update previous failover version

yux0:
**What changed?**
Wire up update previous failover version

**Why?**
This previous failover helps when inserting the failover marker from active -> passive side

**How did you test it?**
Unit test

**Potential risks**
N/A

### DIFF
--- a/common/domain/handler.go
+++ b/common/domain/handler.go
@@ -258,6 +258,7 @@ func (d *HandlerImpl) RegisterDomain(
 			domainRequest.ReplicationConfig,
 			domainRequest.ConfigVersion,
 			domainRequest.FailoverVersion,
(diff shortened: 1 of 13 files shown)


common/domain/handler.go
```diff
@@ -258,6 +258,7 @@ func (d *HandlerImpl) RegisterDomain(
 			domainRequest.ReplicationConfig,
 			domainRequest.ConfigVersion,
 			domainRequest.FailoverVersion,
+			persistence.InitialPreviousFailoverVersion,
```
yycptt, line 261: If the constant will be used by the application layer logic then I think it's probably better to define it in the common package, just like `EmptyVersion`.

yux0: Updated

PR #3323, opened 2020-06-08
Implement DomainMetricsScope cache

anish531213:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/common/cache/metricsScopeCache.go
+++ b/common/cache/metricsScopeCache.go
@@ -1,0 +1,72 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package cache
+
+import (
+	"bytes"
+	"strconv"
+	"sync"
+
+	"github.com/uber/cadence/common/metrics"
+)
+
+type domainMetricsScopeCache struct {
+	sync.RWMutex
+	scopeMap map[string]metrics.Scope
+}
+
+// NewDomainMetricsScopeCache constructs a new domainMetricsScopeCache
+func NewDomainMetricsScopeCache() DomainMetricsScopeCache {
+	return &domainMetricsScopeCache{
+		scopeMap: make(map[string]metrics.Scope),
+	}
+}
+
+// Get retrieves scope for domainID and scopeIdx
+func (c *domainMetricsScopeCache) Get(domainID string, scopeIdx int) (metrics.Scope, bool) {
+	c.RLock()
+	defer c.RUnlock()
+
+	var buffer bytes.Buffer
+	buffer.WriteString(domainID)
+	buffer.WriteString("_")
+	buffer.WriteString(strconv.Itoa(scopeIdx))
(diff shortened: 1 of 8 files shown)


common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,74 @@
...
+	data := cache.Load().(domainMetricsScopeCache)
+
+	var buffer bytes.Buffer
+	buffer.WriteString(domainID)
+	buffer.WriteString("_")
+	buffer.WriteString(strconv.Itoa(scopeIdx))
+	key := buffer.String()
+
+	data.scopeMap[key] = scope
+	cache.Store(data)
```
andrewjdawson2016, line 73: I don't think you will want to store on every put. I think you want to keep some buffer map that you periodically flush (e.g. start a flush go routine in the constructor). For now we can do something very simple where we try to flush every 30 seconds AND as long as there is at least one thing to flush.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,74 @@
...
+	key := buffer.String()
+
+	metricsScope, ok := data.scopeMap[key]
+	return metricsScope, ok
+}
+
+// Put puts map of domainID and scopeIdx to metricsScope
+func (c *domainMetricsScopeCache) Put(domainID string, scopeIdx int, scope metrics.Scope) {
+
+	data := cache.Load().(domainMetricsScopeCache)
```
andrewjdawson2016, line 64: There is a race condition here :)

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+		scopeMap: make(map[string]metrics.Scope),
+		buffer: &buffer{
+			bufferMap: make(map[string]metrics.Scope),
+		},
+	}
+
+	ticker := time.NewTicker(30 * time.Second)
+	quit := make(chan struct{})
+
+	go func() {
```
andrewjdawson2016, line 60: Lets factor this into its own method.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+		for {
+			select {
+			case <- ticker.C:
```
andrewjdawson2016, line 63: Favor <-time.After(time.Second * 30) that way you don't need to worry about stopping ticker.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+				go func() {
```
andrewjdawson2016, line 64: What is the goal of this nested go routine?

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+// NewDomainMetricsScopeCache constructs a new domainMetricsScopeCache
+func NewDomainMetricsScopeCache() DomainMetricsScopeCache {
+	mc := &domainMetricsScopeCache{
+		scopeMap: make(map[string]metrics.Scope),
+		buffer: &buffer{
+			bufferMap: make(map[string]metrics.Scope),
+		},
+	}
+
+	ticker := time.NewTicker(30 * time.Second)
```
andrewjdawson2016, line 57: Please factor this 30 seconds out into a constant with a meaningful name.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+	quit := make(chan struct{})
```
andrewjdawson2016, line 58: Looks like we never close this quit channel?

andrewjdawson2016: I guess you probably want this to extend the common.Daemon by having a start and stop method. On start method you will start a go routine which will periodically flush and you will enter the started state. Then in stop you will write to a channel which will stop this periodic flush method and enter the stopped state.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+	go func() {
+		for {
+			select {
+			case <- ticker.C:
+				go func() {
+					mc.buffer.Lock()
+					defer mc.buffer.Unlock()
+
+					if len(mc.buffer.bufferMap) > 0 {
+						data := cache.Load().(map[string]metrics.Scope)
```
andrewjdawson2016, line 69: I don't think you can update the underlying map directly. I think this will actually update the map used in Get. I think you need to create a new map and copy everything from buffer and existing map.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *domainMetricsScopeCache) Get(domainID string, scopeIdx int) (metrics.Scope, bool) {
+	var buffer bytes.Buffer
+	buffer.WriteString(domainID)
+	buffer.WriteString("_")
+	buffer.WriteString(strconv.Itoa(scopeIdx))
+	key := buffer.String()
+
+	data := cache.Load().(map[string]metrics.Scope)
+
+	metricsScope, ok := data[key]
```
andrewjdawson2016, line 97: No need to assign to variables - can just return.

common/cache/metricsScopeCache.go
```diff
@@ -0,0 +1,113 @@
...
+	return metricsScope, ok
+}
+
+// Put puts map of domainID and scopeIdx to metricsScope
+func (c *domainMetricsScopeCache) Put(domainID string, scopeIdx int, scope metrics.Scope) {
+	c.buffer.Lock()
+	defer c.buffer.Unlock()
+
+	var buffer bytes.Buffer
```
andrewjdawson2016, line 106: Can you write a method to generate this key?

andrewjdawson2016: Its used in two places it would be good to factor this out.

PR #3324, opened 2020-06-08
Add persistence layer for multi-cursor queue processing

emrahs:
**What changed?**
Adding the persistence layer for multi-cursor task processing to persist the Processing Queue State for transfer and timer queues for each shard. For Cassandra, we are adding two new fields to the Shard type to persist these and making the required code changes to utilize those. On the SQL side, we don't need to make a schema change since the thrift structures are stored in the database so we only make the changes needed for the backfill logic. 

**Why?**
Multi-cursor task processing feature is part of our multi-tenancy improvements and this PR is only focused on adding the persistence layer to store the queue states. Integration of these new fields with the multi-cursor task processing logic will be in a separate PR.

**How did you test it?**
Integration tests

**Potential risks**
These changes touch how shards are created, updated and read from the datastore both in Cassandra and SQL. In case of a severe bug, task processing ack levels may be persisted incorrectly or the entire shard may become unreadable.


schema/cassandra/version.go
```diff
@@ -23,7 +23,7 @@ package cassandra
 // NOTE: whenever there is a new data base schema update, plz update the following versions
 
 // Version is the Cassandra database release version
-const Version = "0.28"
+const Version = "0.29"
```
yycptt, line 26: I think we can keep using the current version 0.28 as it's not released yet.

schema/cassandra/cadence/versioned/v0.29/processing_queue_states.cql
```diff
@@ -0,0 +1,14 @@
+CREATE TYPE domain_filter (
+    domain_ids    set<uuid>,
+    reverse_match boolean,
+    );
```
yycptt, line 4: nit: fix format?

schema/cassandra/cadence/versioned/v0.29/processing_queue_states.cql
```diff
@@ -0,0 +1,14 @@
...
+
+CREATE TYPE processing_queue_state (
+    processing_queue_level  int,
+    ack_level               bigint,
+    max_level               bigint,
+    domain_filter           frozen<domain_filter>,
+);
```
yycptt, line 11: We will need another type for timer queue as the type of ack_level and max_level field should be timestamp for timer queue.

schema/cassandra/cadence/versioned/v0.29/processing_queue_states.cql
```diff
@@ -0,0 +1,14 @@
...
+
+ALTER TYPE shard ADD cluster_transfer_processing_queue_state map<text, frozen<processing_queue_state>>;
```
yycptt, line 13: The might be more than one processing queue per shard per cluster, so I think the type should be a map of list? Another idea is that we can just store a blob here and perform encoding & decoding in persistence layer.

schema/cassandra/cadence/schema.cql
```diff
@@ -11,14 +11,30 @@ CREATE TYPE shard (
...
+  cluster_timer_ack_level     map<text, timestamp>, -- TO BE DEPRECATED, IN FAVOR OF cluster_timer_processing_queue_state
   domain_notification_version bigint, -- the global domain change version this shard is aware of
   -- Mapping of (remote) cluster to corresponding replication level (last replicated task_id)
   cluster_replication_level   map<text, bigint>,
   -- Mapping of (remote) cluster to corresponding replication DLQ ack level (last replicated task_id)
   replication_dlq_ack_level   map<text, bigint>,
+  -- Mapping of cluster to corresponding transfer
+    cluster_transfer_processing_queue_state map<text, frozen<processing_queue_state>>,
+  -- Mapping of
+    cluster_timer_processing_queue_state map<text, frozen<processing_queue_state>>,
```
yycptt, line 25: nit: group the new fields with other (old) ack level fields?

common/persistence/cassandra/cassandraPersistenceUtil.go
```diff
@@ -1817,9 +1825,40 @@ func createShardInfo(
...
+		}
+	}
+
+	if info.ClusterTimerProcessingQueueState == nil {
+		info.ClusterTimerProcessingQueueState = map[string]p.ProcessingQueueState{}
+		for cluster, ackLevel := range info.ClusterTimerAckLevel {
+			processingQueueState := createProcessingQueueState(ackLevel.UnixNano())
+			info.ClusterTimerProcessingQueueState[cluster] = processingQueueState
+		}
+	}
```
yycptt, line 1842: I think it's fine to let the application layer logic to handle this situation? If transfer/timer queue encounters an empty map on start, it knows this is a new shard, and can create a default queue.

My idea is that persistency layer shouldn't know anything related to default queue level or how queues are created. All it needs to do is persist and load.

schema/cassandra/version.go
emrahs: Makes sense, I'll do it.

schema/cassandra/cadence/versioned/v0.29/processing_queue_states.cql
emrahs: Yup, will fix this.

common/persistence/cassandra/cassandraPersistenceUtil.go
emrahs: I was on the fence about this one. I agree that this isn't the perfect place to have this backfill logic but I favored consistency with similar backfill logic that was added to this layer earlier and wanted to avoid updating multiple places. After thinking about it again, I'm convinced that eliminating the backfill logic from this layer is the better option. I'll do it.

schema/cassandra/cadence/versioned/v0.29/processing_queue_states.cql
emrahs: Chatted offline about this. I'll keep the field as bigint here to avoid using the timestamp type, but will create a new data interface struct for timer queue state and handle the conversion between time and bigint in the persistence.

yycptt: Another idea it that we can store a blob for all processing queue states in a shard, check here: https://github.com/uber/cadence/pull/3304/files?file-filters%5B%5D=.cql&file-filters%5B%5D=.go#diff-6f7954af286de899bc2234fbbbd083fbR23

common/persistence/cassandra/cassandraPersistenceUtil.go
```diff
@@ -1817,9 +1825,40 @@ func createShardInfo(
 		info.ClusterReplicationLevel = make(map[string]int64)
 	}
 
+	if info.ClusterTransferProcessingQueueState == nil {
+		info.ClusterTransferProcessingQueueState = map[string]p.ProcessingQueueState{}
+		for cluster, ackLevel := range info.ClusterTransferAckLevel {
+			processingQueueState := createProcessingQueueState(ackLevel)
+			info.ClusterTransferProcessingQueueState[cluster] = processingQueueState
+		}
```
gogitops, line 1833: warning: cyclomatic complexity 61 of function createWorkflowExecutionInfo() is high (> 15) (gocyclo)

PR #3331, opened 2020-06-11
Add host level task worker pool

yycptt:
**What changed?**
Add host level task worker pool + some code refactoring

**Why?**
Share resources across different shards on the same hosts

**How did you test it?**
Unit test

**Potential risks**
N/A, feature flag disabled


common/task/fifoTaskScheduler.go
```diff
@@ -35,15 +35,17 @@ import (
...
+		WorkerCount     int
+		DispatcherCount int
+		RetryPolicy     backoff.RetryPolicy
 	}
 
 	fifoTaskSchedulerImpl struct {
 		status       int32
 		logger       log.Logger
 		metricsScope metrics.Scope
+		options      *FIFOTaskSchedulerOptions
```
yux0, line 48: will lint complain this public struct inside a private struct?

service/history/task/processor.go
```diff
@@ -22,70 +22,86 @@ package task
...
+		wrrSchedulerOptions  *task.WeightedRoundRobinTaskSchedulerOptions
 	}
 
 	processorImpl struct {
 		sync.RWMutex
 
 		priorityAssigner PriorityAssigner
-		schedulers       map[shard.Context]task.Scheduler
+		scheduler        task.Scheduler
+		shardSchedulers  map[shard.Context]task.Scheduler
```
yux0, line 50: What about rename to schedulerByHost and shecudlersByShards? Would it be easier to understand?

service/history/config/config.go
```diff
@@ -284,8 +287,11 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
 		TaskProcessRPS:                 dc.GetIntPropertyFilteredByDomain(dynamicconfig.TaskProcessRPS, 1000),
 		EnablePriorityTaskProcessor:    dc.GetBoolProperty(dynamicconfig.EnablePriorityTaskProcessor, false),
 		TaskSchedulerType:              dc.GetIntProperty(dynamicconfig.TaskSchedulerType, int(task.SchedulerTypeWRR)),
-		TaskSchedulerWorkerCount:       dc.GetIntProperty(dynamicconfig.TaskSchedulerWorkerCount, 20),
-		TaskSchedulerQueueSize:         dc.GetIntProperty(dynamicconfig.TaskSchedulerQueueSize, 2000),
+		TaskSchedulerWorkerCount:       dc.GetIntProperty(dynamicconfig.TaskSchedulerWorkerCount, 400),
+		TaskSchedulerShardWorkerCount:  dc.GetIntProperty(dynamicconfig.TaskSchedulerShardWorkerCount, 2),
```
yux0, line 291: Do you think it makes sense to have a shard id filter on those shard level config?

service/history/task/processor.go
```diff
@@ -135,45 +155,63 @@ func (p *processorImpl) StopShardProcessor(
...
+	submitted, err := p.scheduler.TrySubmit(task)
+	if err != nil {
+		return err
+	}
+
+	if submitted {
+		return nil
+	}
+
+	shardScheduler, err := p.getOrCreateShardTaskScheduler(task.GetShard())
```
yux0, line 171: If we have shard level config, then we can submit to the shard scheduler first then try host-level scheduler. In this case, if there is a hot shard, we can give this shard more resources without impact other shards.

common/task/fifoTaskScheduler.go
yux0: Talked offline. This makes sense.

service/history/task/processor.go
yycptt: Synced offline. Basic idea is that if we have a hot shard, we should throttled it instead of giving it more resources as that will other low throughput domains. By submitting to host level pool, tasks from the hot shard will be assigned a low priority and won't affect other domains.

service/history/task/processor.go
yycptt: Updated `scheduler` -> `hostScheduler`

PR #3338, opened 2020-06-12
Add back pressure mechanism for multi-cursor queue

yycptt:
**What changed?**
Implement back pressure mechanism for multi-cursor queue. Basic idea is to slow down (backoff 5 seconds) task loading for non-default queue if it can't submit tasks to task processor (meaning task chan is full). 

**Why?**
Part of multi-cursor implementation

**How did you test it?**
Unit test, integration test, local canary

**Potential risks**
N/A, feature flag disabled


service/history/queue/transfer_queue_processor_base.go
```diff
@@ -66,12 +68,14 @@ type (
...
 		metricsScope  metrics.Scope
 
-		rateLimiter  quotas.Limiter
-		lastPollTime time.Time
-		notifyCh     chan struct{}
-		status       int32
-		shutdownWG   sync.WaitGroup
-		shutdownCh   chan struct{}
+		rateLimiter   quotas.Limiter
+		notifyCh      chan struct{}
```
yux0, line 72: With the poll time mechanism, do we still need this notifyCh?

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -200,29 +210,34 @@ processorPumpLoop:
 		case <-t.shutdownCh:
 			break processorPumpLoop
 		case <-t.notifyCh:
-			if t.redispatchQueue.Len() <= t.options.MaxRedispatchQueueSize() {
-				t.processBatch()
-				continue
+			t.upsertPollTime(defaultProcessingQueueLevel, time.Time{})
```
yux0, line 213: Is the defaultProcessingQueueLevel in this base?

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -255,67 +270,83 @@ processorPumpLoop:
...
 		if err != nil {
 			t.logger.Error("Processor unable to retrieve tasks", tag.Error(err))
-			t.notifyNewTask() // re-enqueue the event
-			return
+			t.upsertPollTime(level, time.Time{}) // re-enqueue the event
+			continue
 		}
 
+		pollTime := t.shard.GetTimeSource().Now()
+		t.upsertPollTime(level, pollTime.Add(backoff.JitDuration(
```
yux0, line 329: do you need to check if it is default level here?

service/history/queue/transfer_queue_processor_base.go
yycptt: I think we still need this notifyCh for two reasons:
- Without it, each notifyNewTask call need to grab a lock on `nextPollTime` which may be blocking
- Without the notifyCh, we will need to update the poll timer for every new task, which involves starting/stopping/resetting timer and can be expensive.

service/history/queue/transfer_queue_processor_base.go
yycptt: I don't quite understand your question, can you maybe elaborate? Are you asking where is the `defaultProcessingQueueLevel` define? Right now it's defined in `transfer_queue_processor.go`, I don't think that's a good place, I will move it to `queue_processor_util.go`

service/history/queue/transfer_queue_processor_base.go
yycptt: I think it makes sense to set a max poll interval for queues at all level just as a safety mechanism. Let me know your thoughts on this.

yux0: I see. LGTM

Issue #2862, opened 2019-11-25
CLI start workflow needs header support


meiliang86:
Add header support for CLI start workflow.

PR #3341, opened 2020-06-15
Add header options for starting workflow via CLI (#2862)

vytautas-karpavicius:
**What changed?**
This change exposes three new optional parameters for starting a workflow:
- header_key
- header_value
- header_file

They allow setting header values in JSON format which are propogated
via workflow context. Values can be provides either directly or with a
file. If multiple keys and values are used, they separated by a space.
Count of keys and values must be equal, otherwise error is returned.

**Why?**
It was not possible to set header values via CLI when starting a workflow.

**How did you test it?**
Tested locally with:
```
./cadence --domain mickey workflow start --tasklist mini --workflow_type pluto --execution_timeout 1 --header_key='aaa bbb ccc ddd eee fff ggg hhh' --header_value='1 3.3 "text" true [] {} [1,2,3] {"inner":"a"}'
```

Verified that headers were set on local server:
```
header.fields.ggg.0     1
header.fields.ggg.1     2
header.fields.ggg.2     3
header.fields.hhh.inner a
header.fields.aaa       1
header.fields.bbb       3.3
header.fields.ccc       text
header.fields.ddd       true
```

**Potential risks**


tools/cli/workflowCommands.go
```diff
@@ -311,6 +316,35 @@ func processSearchAttr(c *cli.Context) map[string][]byte {
 	return fields
 }
 
+func processHeader(c *cli.Context) map[string][]byte {
```
vancexu, line 319: This looks similar to processMemo, can we try refactor to reduce duplicates?

vytautas-karpavicius: Sure, refactored them.

PR #3342, opened 2020-06-15
Fix ListTaskListPartition command

anish531213:
**What changed?**
Minor fixes in task list partition

**Why?**

**How did you test it?**
Tested locally using cli.

**Potential risks**
No risk.


service/matching/matchingEngine.go
```diff
@@ -615,13 +615,12 @@ func (e *matchingEngineImpl) listTaskListPartitions(
...
 	var partitionHostInfo []*workflow.TaskListPartitionMetadata
 	for _, partition := range partitions {
-		if host, err := e.getHostInfo(partition); err != nil {
-			partitionHostInfo = append(partitionHostInfo,
-				&workflow.TaskListPartitionMetadata{
-					Key:           common.StringPtr(partition),
-					OwnerHostName: common.StringPtr(host),
-				})
-		}
+		host, _ := e.getHostInfo(partition)
```
vancexu, line 618: Do we no longer need the the err if?

anish531213: If there is no error we return empty string for host info. So, we still can get partition name even if host information for that partition name cannot be found.

PR #3345, opened 2020-06-17
[Sketch] Add current execution check for scanner - one scan

mkolodezny:


service/worker/scanner/executions/common/interfaces.go
```diff
@@ -27,7 +27,7 @@ import "github.com/uber/cadence/common/persistence"
 type (
 	// PersistenceRetryer is used to retry requests to persistence
 	PersistenceRetryer interface {
-		ListConcreteExecutions(*persistence.ListConcreteExecutionsRequest) (*persistence.ListConcreteExecutionsResponse, error)
+		ListAllExecutions(*persistence.ListConcreteExecutionsRequest) (*persistence.ListConcreteExecutionsResponse, error)
```
vancexu, line 30: Maybe just add new API for listAll, otherwise the request and response is a little inconsistent with API name.

service/worker/scanner/executions/invariants/openCurrentExecution.go
```diff
@@ -114,6 +114,11 @@ func (o *openCurrentExecution) Fix(execution common.Execution) common.FixResult
 	return *fixResult
 }
 
+func (o *openCurrentExecution) CanApply(execution common.Execution) bool {
+	// TODO return true for concrete execution
+	return true
```
yux0, line 119: Does this use as a feature flag?

PR #3349, opened 2020-06-18
Fix handling of CurrentWorkflowConditionFailedError when create wf

vancexu:
**What changed?**
Fix handling of CurrentWorkflowConditionFailedError when create workflow

**Why?**
In create workflow, when CurrentWorkflowConditionFailedError happens, it doesn't make sense for shard to renewRange, because it doesn't help such error case and likely to cause more errors. 

**How did you test it?**
existing tests.
bench test on staging

**Potential risks**
Limited. in worst case, create workflow will encounter unexpected error.


service/history/shard/context.go
```diff
@@ -496,6 +496,7 @@ Create_Loop:
 			switch err.(type) {
 			case *shared.WorkflowExecutionAlreadyStartedError,
 				*persistence.WorkflowExecutionAlreadyStartedError,
+				*persistence.CurrentWorkflowConditionFailedError,
```
yux0, line 499: If the range ID changes, it will throw this CurrentWorkflowConditionFailedError? If it is the case, why no update the range id?

vancexu: CurrentWorkflowConditionFailedError in StartWorkflow only happened when concurrent record messed up. RangeID changes will not lead to this error. That's my understanding based on reading of https://github.com/uber/cadence/blob/master/common/persistence/cassandra/cassandraPersistence.go#L1229 - L1242

yux0: I think @yycptt makes a point. should we retry on this error

vancexu: this error will be retry by client; it should be non-retriable from history service point of view currently.

vancexu:
> Whats the metric to check once it gets landed?

1. Error log on CurrentWorkflowConditionFailedError should no longer see same failing signalwithstart request repeated happening with increasing rangeID 
2. History service signalwithstart latency should drop.

PR #3356, opened 2020-06-22
Update replication ack level if response has no task

yux0:
**What changed?**
Update replication ack level if response has no task

**Why?**
if the workflow passed the retention day and got delete, the replication task ack level should be updated.

**How did you test it?**
Unit test and local tests

**Potential risks**
N/A

### DIFF
--- a/service/history/replicatorQueueProcessor.go
+++ b/service/history/replicatorQueueProcessor.go
@@ -25,6 +25,7 @@ package history
 import (
 	ctx "context"
 	"errors"
+	"strconv"
 	"time"
 
 	"github.com/uber/cadence/.gen/go/replicator"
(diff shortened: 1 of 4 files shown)


service/history/replicatorQueueProcessor.go
```diff
@@ -461,21 +462,22 @@ func (p *replicatorQueueProcessorImpl) getTasks(
 		}
 	}
 
-	// Note this is a very rough indicator of how much the remote DC is behind on this shard.
-	p.metricsClient.RecordTimer(
+	replicationScope := p.metricsClient.Scope(
 		metrics.ReplicatorQueueProcessorScope,
+		metrics.InstanceTag(strconv.Itoa(p.shard.GetShardID())),
```
yycptt, line 467: We may need to check gc and cpu usage for this shardID tag. Not sure if it will cause the same thing as adding domain tag to task processing.

yux0: I don't think replication is this time-sensitive. But we can verify

PR #3361, opened 2020-06-25
Replication metrics

yux0:
**What changed?**
1. Add latency on applying replication task and generate replication task
2. Update the task retry policy to be less aggressive 

**Why?**
Fix replication issue

**How did you test it?**
Unit tests

**Potential risks**
N/A


service/history/replicatorQueueProcessor.go
```diff
@@ -463,10 +468,7 @@ func (p *replicatorQueueProcessorImpl) getTasks(
 		}
 	}
 
-	replicationScope := p.metricsClient.Scope(
-		metrics.ReplicatorQueueProcessorScope,
-		metrics.InstanceTag(strconv.Itoa(p.shard.GetShardID())),
-	)
+	timer.Stop()
```
yycptt, line 471: nit: use a more informative/descriptive name?

yux0: updated

PR #3363, opened 2020-06-26
Add region filter for cassandra config

anish531213:
**What changed?**
Added region filter while initializing Cassandra cluster as recommended from comments in this diff:
https://code.uberinternal.com/D4542141

**Why?**
Portability of cadence service asks to use region instead of the datacenter as a config.

**How did you test it?**
Will test this in staging before merging.

**Potential risks**
Service might be unable to connect to Cassandra host.


common/cassandra/cassandraCluster.go
```diff
@@ -70,6 +77,16 @@ func NewCassandraCluster(cfg config.Cassandra) *gocql.ClusterConfig {
 	return cluster
 }
 
+func RegionHostFilter() gocql.HostFilter {
```
dorianperkins, line 80: This leaks some proprietary implementation details into open-source. Should we instead use the Option pattern to allow the caller to provide a custom `gocql.HostFilter` which overrides the default?

PR #3365, opened 2020-06-30
Fix history resender source cluster

yux0:
**What changed?**
Replication task executor generates per source cluster

**Why?**
The replication resender needs the context of which remote cluster to request tasks.

**How did you test it?**
Local tests

**Potential risks**
Task replication will be broken for rpc replication

### DIFF
--- a/service/history/replication/dlq_handler.go
+++ b/service/history/replication/dlq_handler.go
@@ -26,11 +26,16 @@ import (
 	"context"
 
 	"github.com/uber/cadence/.gen/go/replicator"
+	workflow "github.com/uber/cadence/.gen/go/shared"
 	"github.com/uber/cadence/common"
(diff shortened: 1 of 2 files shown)


service/history/replication/dlq_handler.go
```diff
@@ -189,7 +202,7 @@ func (r *dlqHandlerImpl) MergeMessages(
 	)
 
 	for _, task := range tasks {
-		if _, err := r.taskExecutor.execute(
+		if _, err := r.taskExecutors[sourceCluster].execute(
 			sourceCluster,
```
yycptt, line 206: nit: seems like we can move the `sourceCluster` to the constructor of replication task executor.

yux0: moved source cluster to constructor

PR #3369, opened 2020-07-01
Fix task agressive retry with TwoPhaseRetryPolicy

vancexu:
**What changed?**
Change retry policy for task processor with a new TwoPhaseRetryPolicy.

**Why?**
Currently, task retry is using retry policy:
initial interval 50ms; expire interval 30s; max internal 10s
such retry in task processor lead to 12 attempts in 30s, and during a overloaded task outage, it will intensively repeat such policy (for example, 5 min outage will cause 100+ retries which make things worse)

We have to retry forever to not lose task, but during overloaded task outage, we don't want meaningless intensive retry. 
So this PR add a TwoPhaseRetryPolicy, that support retry 3 time really quick, then slowly retry in second phase. 
By this way, failed task retry will be limited during outage (same 5 min outage will now have 11 retries)

The parameter is based on observed metrics:
p99 task latency 200ms; we almost never seen task failures unless outage. 
We can consider make it configurable later. 

**How did you test it?**
unit test

**Potential risks**
task latency increase in rare failure cases.

### DIFF
--- a/service/history/taskProcessor.go
+++ b/service/history/taskProcessor.go
@@ -116,7 +116,7 @@ func newTaskProcessor(
 		domainMetricsScopeCache: shard.GetService().GetDomainMetricsScopeCache(),
 		timeSource:              shard.GetTimeSource(),
 		workerNotificationChans: workerNotificationChans,
-		retryPolicy:             common.CreatePersistenceRetryPolicy(),
+		retryPolicy:             backoff.NewTwoPhaseRetryPolicy(),
 		numOfWorker:             options.workerCount,
 	}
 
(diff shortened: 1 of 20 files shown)


service/history/taskProcessor.go
```diff
@@ -116,7 +116,7 @@ func newTaskProcessor(
 		domainMetricsScopeCache: shard.GetService().GetDomainMetricsScopeCache(),
 		timeSource:              shard.GetTimeSource(),
 		workerNotificationChans: workerNotificationChans,
-		retryPolicy:             common.CreatePersistanceRetryPolicy(),
+		retryPolicy:             backoff.NewTwoPhaseRetryPolicy(),
```
yycptt, line 119: Do you plan to use this two phase retry policy for other components? This task processor will be deprecated soon as we switch to the priority task processor, which is using a very different retry policy for tasks. You can find the new retry policy in `common/util.go`

vancexu: For now, major created it for this processor to alleviate incident. No plan to put it to priority task processor before new processor goes alive.
It can be used in other components where retry forever happens, but would be better to have some metrics before switching to this policy.

PR #3373, opened 2020-07-06
Update worker replication config

yux0:
**What changed?**
Update worker replication config for enabling/disable replication

**Why?**
Provide the ability to disable features in the worker role

**How did you test it?**
local tests

**Potential risks**
N/A


common/service/dynamicconfig/constants.go
```diff
@@ -760,8 +760,8 @@ const (
 	WorkerReReplicationContextTimeout
 	// WorkerEnableRPCReplication is the feature flag for RPC replication
 	WorkerEnableRPCReplication
-	// WorkerEnableHistoryReplication is the feature flag for history replication
-	WorkerEnableHistoryReplication
+	// WorkerEnableReplication is the feature flag for general replication
+	WorkerEnableReplication
```
yycptt, line 764: nit: I can understand the difference between those two configs but it may cause some confusion for the others. Is it possible to update the name to be more specific?

yux0: Updated

PR #3377, opened 2020-07-09
Add shard distribution metrics and query

andrewjdawson2016:
In this PR I add query endpoints to executions scanner to get the shard progress of the scanner. I also add metrics on the distribution of executions in shards and expose a query endpoint to get this information.

### DIFF
--- a/service/worker/scanner/executions/aggregators.go
+++ b/service/worker/scanner/executions/aggregators.go
@@ -33,21 +33,22 @@ type (
 		minShard int
 		maxShard int
 
-		reports     map[int]common.ShardFixReport
-		status      ShardStatusResult
-		aggregation AggregateFixReportResult
+		reports       map[int]common.ShardFixReport
+		status        ShardStatusResult
+		statusSummary ShardStatusSummaryResult
+		aggregation   AggregateFixReportResult
 	}
 
 	shardScanResultAggregator struct {
 		minShard int
 		maxShard int
 
-		reports                 map[int]common.ShardScanReport
-		status                  ShardStatusResult
-		aggregation             AggregateScanReportResult
-		corruptionKeys          map[int]common.Keys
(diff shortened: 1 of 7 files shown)


service/worker/scanner/executions/aggregators.go
```diff
@@ -186,14 +198,15 @@ func (a *shardScanResultAggregator) getStatusResult(req PaginatedShardQueryReque
 }
 
 func (a *shardScanResultAggregator) addReport(report common.ShardScanReport) {
-	a.removeReport(report.ShardID)
+	a.insertReportIntoSizes(report)
```
andrewjdawson2016, line 201: TODO: should only insert if report is for a successful shard scan.

service/worker/scanner/executions/workflows.go
```diff
@@ -190,6 +216,11 @@ func ScannerWorkflow(
 	}); err != nil {
 		return err
 	}
+	if err := workflow.SetQueryHandler(ctx, ShardStatusSummaryQuery, func() (ShardStatusSummaryResult, error) {
+		return aggregator.statusSummary, nil
+	}); err != nil {
+		return nil
```
yux0, line 222: return err?

service/worker/scanner/executions/workflows.go
```diff
@@ -288,6 +325,14 @@ func FixerWorkflow(
 	}); err != nil {
 		return err
 	}
+	if err := workflow.SetQueryHandler(ctx, ShardStatusSummaryQuery, func() (ShardStatusSummaryResult, error) {
+		if aggregator == nil {
+			return nil, errQueryNotReady
+		}
+		return aggregator.statusSummary, nil
+	}); err != nil {
+		return nil
```
yux0, line 334: return err?

PR #3378, opened 2020-07-09
Add cron timer jitter between 0-3 seconds

yux0:
**What changed?**
Add cron timer jitter between 0-3 seconds

**Why?**
The cron all fires around the clock time and could trigger many timers

**How did you test it?**
local

**Potential risks**
In worst case, the cron schedule will be off


yux0:
Not urgent to have this change

PR #3379, opened 2020-07-10
Improve domain tagged metrics in priority task processor

yycptt:
**What changed?**
* Add a dynamic config for controlling if task processing metrics should contain domain tag
* Delay metrics scope initialization to task execution phase so that it can be done in parallel

**Why?**
Try to improve performance for domain tagged metrics

**How did you test it?**
Unit test

**Potential risks**
N/A feature flag disabled.

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -417,7 +417,7 @@ func newTimerQueueActiveProcessor(
 	redispatchQueue := collection.NewConcurrentQueue()
 
 	// read dynamic config only once on startup to avoid gc pressure caused by keeping reading dynamic config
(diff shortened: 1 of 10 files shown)


service/history/queue/timer_queue_processor.go
```diff
@@ -416,6 +416,8 @@ func newTimerQueueActiveProcessor(
 
 	redispatchQueue := collection.NewConcurrentQueue()
 
+	// read dynamic config only once on startup to avoid gc pressure caused by keeping reading dynamic config
+	emitDomainTag := shard.GetConfig().QueueProcessorEnableDomainTaggedMetrics()
```
yux0, line 420: just use config from above?

yycptt: Updated.

PR #3382, opened 2020-07-14
Add workflowID runID to frontend API error log

vancexu:
**What changed?**
Add workflowID runID to frontend API error log

**Why?**
Without workflow info, it is very hard to troubleshoot, like why latency for small portion of GetHistory call increase.  

**How did you test it?**
Existing tests.

**Potential risks**
pressure on logging.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -1828,16 +1828,6 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 	return resp, nil
 }
 
-func getWfIDRunIDTags(wf *gen.WorkflowExecution) []tag.Tag {
-	if wf == nil {
-		return nil
-	}
-	return []tag.Tag{
-		tag.WorkflowID(wf.GetWorkflowId()),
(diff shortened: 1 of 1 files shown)


anish531213:
Is it possible to add these workflow and runID tag in the scope itself? The changes might be small if that is possible.

vancexu:
> Is it possible to add these workflow and runID tag in the scope itself? The changes might be small if that is possible.

This is for adding log tag, so we can not make use of metrics scope here.

service/frontend/workflowHandler.go
```diff
@@ -2023,7 +2024,7 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 	if isCloseEventOnly {
 		if !isWorkflowRunning {
 			if err := getHistory(lastFirstEventID, nextEventID, nil); err != nil {
-				return nil, wh.error(err, scope)
+				return nil, wh.error(err, scope, getWfIDRunIDTags(wfExecution)...)
```
mkolodezny, line 2027: thats gonna make tracing much clear, thanks!

PR #3383, opened 2020-07-14
Only emit domain tag metrics for active task processor

anish531213:
Emit domain metrics only for active task processor in old task processor

**What changed?**
Added a boolean config to be passed in task processor in order to emit domainTag metrics for only active task processor.

**Why?**
Made this change because we were seeing latency increase only for standby transfer and timer queue and not active transfer and timer queue.

**How did you test it?**
Haven't verified. Need to test this staging2.

**Potential risks**
No major risk or impact.


anish531213:
Not needed.

PR #3386, opened 2020-07-14
 Add timers subcommand to admin CLI

mantas-sidlauskas:
**What changed?**
Adding helper command under admin/db section in CLI.
This command lists timers scheduled. 
You can use number of filters:
* time range
* domain UUID
* shard range
* number of goroutines to use for querying shards 

**Why?**

**How did you test it?**

**Potential risks**


tools/cli/admin.go
```diff
@@ -799,6 +803,55 @@ func newDBCommands() []cli.Command {
 				AdminDBScan(c)
 			},
 		},
+		{
+			Name:    "timers",
+			Aliases: []string{"timers"},
+			Usage:   "list timers",
+			Flags: append(getDBFlags(),
+				cli.IntFlag{
```
meiliang86, line 811: nit: taking a list of shardIDs seems to be better/more versatile?

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+
```
meiliang86, line 31: nit: remove these empty lines between imports

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+	session := connectToCassandra(c)
+	defer session.Close()
+
+	ctx, cancel := newContextForLongPoll(c)
+	defer cancel()
+
+	st, err := parseSingleTs(startDate)
+	if err != nil {
+		ErrorAndExit("wrong date format for "+FlagStartDate, err)
```
meiliang86, line 61: nit: show accepted data formats in error?

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+func printTimers(
+	session *gocql.Session,
+	shardID, batchSize int,
+	startDate, endDate time.Time,
+	domainId string,
+	skipErrMode bool,
+) error {
+	execStore, err := cassp.NewWorkflowExecutionPersistence(shardID, session, loggerimpl.NewNopLogger())
+
+	if err != nil {
```
meiliang86, line 96: check skipErrMode here as well?

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+			NextPageToken: token,
+		}
+		resp, err := execStore.GetTimerIndexTasks(&req)
+
+		if err != nil {
+			return errors.Wrapf(err, "cannot get timer tasks for shardId %d", shardID)
+		}
+
+		token = resp.NextPageToken
+		if len(resp.Timers) < 1 {
```
meiliang86, line 117: This is unnecessary, i.e. the for loop below will be skipped anyway if Timers array is empty?

tools/cli/admin.go
```diff
@@ -799,6 +803,55 @@ func newDBCommands() []cli.Command {
...
+				},
+				cli.StringFlag{
+					Name:  FlagStartDate,
+					Usage: "start date",
+					Value: time.Now().UTC().Format(time.RFC3339),
+				},
+				cli.StringFlag{
+					Name:  FlagEndDate,
+					Usage: "end date",
+					Value: time.Now().UTC().Add(24 * time.Hour).Format(time.RFC3339),
```
vytautas-karpavicius, line 839: Just a suggestion, as it would require additional implementation effort.

In such occasions it usually much more convenient to deal with relative timestamps instead. For example `now + 1d` or `now + 2w`. There are libraries available for such support, e.g. https://github.com/karrick/tparse or you may choose to implement simplified version yourself.

tools/cli/adminDBTimers.go
mantas-sidlauskas: Error message returned is pretty verbose and provides example of date expected. Leaving as is for now

tools/cli/adminDBTimers.go
mantas-sidlauskas: good point

tools/cli/admin.go
mantas-sidlauskas: Thanks Vytautas, 

Will leave it as it is now, but your suggestion can be implemented in another PR. Let's figure it out, what will be usage pattern of this tool, first.

tools/cli/admin.go
```diff
@@ -799,6 +803,55 @@ func newDBCommands() []cli.Command {
...
+				},
+				cli.IntFlag{
+					Name:  FlagPageSize,
+					Usage: "page size used to query db executions table",
+					Value: 500,
+				},
+				cli.IntFlag{
+					Name:  FlagConcurrency,
+					Usage: "number of threads to handle scan",
+					Value: 1000,
```
vancexu, line 829: Will 1k too big for default? maybe change to 200?

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+	domainID := c.String(FlagDomainID)
+	skipErrMode := c.Bool(FlagSkipErrorMode)
+
+	session := connectToCassandra(c)
+	defer session.Close()
+
+	ctx, cancel := newContextForLongPoll(c)
+	defer cancel()
+
+	st, err := parseSingleTs(startDate)
```
vancexu, line 57: nit: maybe move to front L50, so save connection to db when invalid input.

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,140 @@
...
+
+			data, err := json.Marshal(t)
+			if err != nil {
+				err = errors.Wrap(err, "cannot marshal timer to json")
+				if !skipErrMode {
+					return err
+				}
+				fmt.Println(err.Error())
+			} else {
+				fmt.Println(string(data))
```
andrewjdawson2016, line 134: This looks like it just prints timers data? What if I want output like "give me the RPS for timers for the next day grouped by hour"?

Do users need to do some fancy grep and awk in order to answer these queries?

tools/cli/admin.go
meiliang86: you should easily generate a list of shards using some sub bash command and then fed into this command I assume?

tools/cli/adminDBTimers.go
meiliang86: I assume there will be part 2 that includes this?

meiliang86:
Just found out this commit from temporal https://github.com/temporalio/temporal/commit/1786350114b7519ae07f6fb35274aeb132f1d719
Seems to me that adding this inside the shard command is also reasonable.

PR #3388, opened 2020-07-15
Add retries to range renewal

emrahs:
**What changed?**
Added retries to rangeID renewal code to make it more robust.

**Why?**
When rangeID renewal fails, we close the shard and that affects availability if the shard closure was unnecessary. With the retries, we aim to reduce the number of unnecessary shard closures.

**How did you test it?**
Added unit tests. 

**Potential risks**
Retries may increase the load on the DB but that shouldn't be too much given the small size of retries and how often that code path would be utilized.

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -992,7 +992,7 @@ Retry_Loop:
 				tag.Attempt(attempt),
 			)
 			s.closeShard()
-			break
+			break Retry_Loop
 		default:
 			s.logger.Warn("UpdateShard failed with an unknown error.",
 				tag.Error(err),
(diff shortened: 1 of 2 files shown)


service/history/shard/context.go
```diff
@@ -973,26 +973,42 @@ func (s *contextImpl) renewRangeLocked(isStealing bool) error {
...
 				tag.ShardID(s.GetShardID()),
 				tag.Error(err),
+				tag.Attempt(attempt),
 			)
 			s.closeShard()
-		} else {
-			// Failure in updating shard to grab new RangeID
-			s.logger.Error("renewRangeLocked failed due to unknown error.",
-				tag.StoreOperationUpdateShard,
+			break
```
yux0, line 995: break Retry_Loop?

emrahs: good catch! :)

PR #3389, opened 2020-07-15
Emitting metrics if DLQ is not empty

yux0:
**What changed?**
Adding a loop to emit metrics if DLQ is not empty

**Why?**
The approach to calculating the delta of max DLQ message ID and acked DLQ message ID is not practical due to the metrics measure window. So we have to query and emit these metrics.

**How did you test it?**
integration test

**Potential risks**


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -731,6 +731,15 @@ workflow_state = ? ` +
...
 		`and task_id <= ?`
 
+	templateGetDLQReplicationTaskQuery = `SELECT replication ` +
+		`FROM executions ` +
+		`WHERE shard_id = ? ` +
+		`and type = ? ` +
+		`and domain_id = ? ` +
+		`and workflow_id = ? ` +
+		`and run_id = ? ` +
+		`limit 1`
```
vancexu, line 741: As discussed, If possible we would prefer to show the real backlog for easier view of num of messages in DLQ.

yux0: Updated

PR #3394, opened 2020-07-16
Fix Get DLQ if the workflow is archived

yux0:
**What changed?**
Check if the task is null or not

**Why?**
If the workflow pass retention day, the task will be null

**How did you test it?**
local test

**Potential risks**

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -3045,7 +3045,9 @@ func (e *historyEngineImpl) GetDLQReplicationMessages(
 			e.logger.Error("Failed to fetch DLQ replication messages.", tag.Error(err))
 			return nil, err
 		}
-		tasks = append(tasks, task)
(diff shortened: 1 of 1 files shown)


service/history/historyEngine.go
```diff
@@ -3045,7 +3045,9 @@ func (e *historyEngineImpl) GetDLQReplicationMessages(
 			e.logger.Error("Failed to fetch DLQ replication messages.", tag.Error(err))
 			return nil, err
 		}
-		tasks = append(tasks, task)
+		if task != nil {
+			tasks = append(tasks, task)
```
meiliang86, line 3049: Worth adding some log/metrics?

yux0: it is a common case if the workflow passes the retention day. I don't we need to have log or metrics.

PR #3398, opened 2020-07-17
Cancel context properly in matching task MustOffer

meiliang86:
**What changed?**
Cancel context in all cases.

**Why?**
https://github.com/temporalio/temporal/pull/438

**How did you test it?**

**Potential risks**


service/matching/matcher.go
```diff
@@ -234,6 +234,7 @@ forLoop:
 		case token := <-tm.fwdrAddReqTokenC():
 			childCtx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*2))
 			err := tm.fwdr.ForwardTask(childCtx, task)
+			cancel()
```
vancexu, line 237: Adding cancel here will causing code execution always goes to L247 instead of L245, won't that cause problem (since previously it try add task to chan in err case)?

meiliang86: Yeah good point. Somehow I misread the code.
I took the fix from temporal directly.

PR #3402, opened 2020-07-21
Extend execution scanner framework with an entity type param

mkolodezny:
**What changed?**

This is the first step to have a separate scanner workflow for current executions (and for other types in the future).
The next steps are:

- introduce current execution type APIs
- implement concrete execution check
- register the scanner workflow configured for current executions
 

**How did you test it?**

unit tests


common/reconciliation/common/interfaces.go
```diff
@@ -38,25 +38,25 @@ type (
 	// InvariantManager represents a manager of several invariants.
 	// It can be used to run a group of invariant checks or fixes.
 	InvariantManager interface {
```
meiliang86, line 40: Should we invert the dependency here? meaning putting Check() and Fix() as execution methods.

meiliang86: Or you can continue with the current approach and we can rethink about this when Mantas starts to break this workflow into pieces.

common/reconciliation/common/interfaces.go
```diff
@@ -38,25 +38,25 @@ type (
...
-		RunChecks(Execution) ManagerCheckResult
-		RunFixes(Execution) ManagerFixResult
+		RunChecks(interface{}) ManagerCheckResult
```
andrewjdawson2016, line 41: Can we either define these as a union type such that they have a type field and pointer fields where one is specified or do what Liang said above?

common/reconciliation/common/types.go
```diff
@@ -83,16 +83,21 @@ const (
...
-		RunID       string
+		ShardID    int
+		DomainID   string
+		WorkflowID string
+		RunID      string
+		State      int
+	}
+
+	// ConcreteExecution is a concrete execution.
+	ConcreteExecution struct {
```
andrewjdawson2016, line 96: I would expect to see a CurrentExecution type defined here?

common/reconciliation/common/types.go
```diff
@@ -83,16 +83,21 @@ const (
...
 	Execution struct {
-		ShardID     int
-		DomainID    string
-		WorkflowID  string
-		RunID       string
+		ShardID    int
+		DomainID   string
+		WorkflowID string
+		RunID      string
+		State      int
```
andrewjdawson2016, line 92: Does Execution serve as a base type for both concrete and current if so do all these fields apply to current?  I don't think current execution specifies State?

common/reconciliation/common/util.go
```diff
@@ -45,18 +45,32 @@ func ValidateExecution(execution *Execution) error {
...
 	}
-	if len(execution.BranchToken) == 0 {
+	if execution.State < persistence.WorkflowStateCreated || execution.State > persistence.WorkflowStateCorrupted {
+		return fmt.Errorf("unknown workflow state: %v", execution.State)
+	}
+	return nil
+}
+
+// ValidateConcreteExecution returns an error if ConcreteExecution is not valid, nil otherwise.
+func ValidateConcreteExecution(exec interface{}) error {
```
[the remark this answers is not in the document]
andrewjdawson2016: For all these files in common/reconcilation it seems to me like the pattern should be to define a union type with a type field and a bunch of pointers where one is non-nil. Then for each type we add we will have the same methods (e.g. Validate)

common/reconciliation/common/util.go
```diff
@@ -135,16 +149,34 @@ func Open(state int) bool {
...
 
+// Open returns true if execution state is open false if workflow is closed
+func ExecutionOpen(execution interface{}) bool {
+	return Open(getExecution(execution).State)
+}
+
+// getExecution returns base Execution
+func getExecution(execution interface{}) *Execution {
+	switch e := execution.(type) {
+	case *Execution:
```
andrewjdawson2016, line 160: Seems here we are treating Execution to mean current? Cut concrete is called ConcreteExecution?

common/reconciliation/invariants/historyExists.go
```diff
@@ -49,17 +49,25 @@ func NewHistoryExists(
 	}
 }
 
-func (h *historyExists) Check(execution common.Execution) common.CheckResult {
+func (h *historyExists) Check(execution interface{}) common.CheckResult {
```
andrewjdawson2016, line 52: This method should take in common.ConcreteExectuion. Invariants for concrete execution cannot be applied on current execution (and the reverse is also true) so at this level in code I don't think it makes sense to still be dealing with interfaces (or even dealing with union types at this level if we go with that path).

common/reconciliation/invariants/historyExists.go
```diff
@@ -106,7 +114,7 @@ func (h *historyExists) Check(execution common.Execution) common.CheckResult {
 	}
 }
 
-func (h *historyExists) Fix(execution common.Execution) common.FixResult {
+func (h *historyExists) Fix(execution interface{}) common.FixResult {
```
andrewjdawson2016, line 117: Same comment here.

service/worker/scanner/executions/activities.go
```diff
@@ -55,6 +55,11 @@ const (
 	ErrSerialization = "encountered serialization error"
 )
 
+var scanTypePrefixMap = map[common.ScanType]string{
+	common.ConcreteExecutionType: "", // leave it empty for now to be backwards compatible
```
andrewjdawson2016, line 59: I don't think there should be any backwards compatible concerns. We can also just terminate the current run.

andrewjdawson2016: Or update the dashboards.

common/reconciliation/common/util.go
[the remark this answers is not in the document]
mkolodezny: Changed to ConcreteExecution type and made private thanks!

common/reconciliation/common/interfaces.go
mkolodezny: Yep, we definitely need that! Will try to do in the second "implement concrete execution check" step.

common/reconciliation/common/interfaces.go
mkolodezny: replied above

common/reconciliation/common/types.go
mkolodezny: Exactly, that will go with the next update "introduce current execution type APIs" (step 1 in the description)

common/reconciliation/common/types.go
mkolodezny: Yep, according to DB records  It looks like we do update workflow state in current execution.

common/reconciliation/common/util.go
mkolodezny: Yes, you are right. the current type will be introduced as part of "introduce current execution type APIs" step

service/worker/scanner/executions/activities.go
mkolodezny: Lets do it safe and update it in "register the scanner workflow configured for current executions" step respectively.

common/reconciliation/invariants/historyExists.go
mkolodezny: here and below: could you please elaborate your proposal with a small design example. Otherwise, let's not overcomplicate the current one.

service/worker/scanner/executions/activities.go
yycptt: I think `scanTypePrefixMap` is only used inside activity implementation so there's no backward compatibility issue here? Can you elaborate a little bit on your concern here? 

The only backward compatibility issue I can think of is the newly added `scanType` field in WorkflowParams and ActivityParams. But it seems like you rely on the fact that `ConcreteExecutionType` is the default value for that field to resolve the issue?

common/reconciliation/common/types.go
```diff
@@ -83,16 +83,21 @@ const (
 
 // The following are types related to Invariant.
 type (
-	// Execution is an execution which should be checked or fixed.
+	// Execution is a base type for executions which should be checked or fixed.
 	Execution struct {
```
yycptt, line 87: Another idea is turn `Execution` into an interface with Getter method for each field and let concrete and current records extend this interface to include more methods. Then we can be more specific when writing function signatures and don't need to use `interface{}` everywhere. I think this can also address @andrewjdawson2016 's comments 
> This method should take in common.ConcreteExectuion. Invariants for concrete execution cannot be applied on current execution (and the reverse is also true) so at this level in code I don't think it makes sense to still be dealing with interfaces (or even dealing with union types at this level if we go with that path).

service/worker/scanner/executions/activities.go
mkolodezny: Its about the metric scope. Discussed offline.

common/reconciliation/common/types.go
mkolodezny: Yeah, I like that proposal cc @meiliang86 Lets try to do it in the next change for step "introduce current execution type APIs".

common/reconciliation/common/interfaces.go
andrewjdawson2016: Okay sounds good

common/reconciliation/invariants/historyExists.go
andrewjdawson2016: My only point here is that at this level in code you should not be dealing with an interface. Invariant checks for concrete executions do not have to do with current execution checks so at the level of invariants we should not be taking interfaces.

PR #3404, opened 2020-07-22
Add timers subcommand to admin CLI

mantas-sidlauskas:
**What changed?**
Adding helper command under admin/db section in CLI.

This command returns timers scheduled.
You can use number of filters:
* time range
* domain UUID
* shard range
* timer types to retrieve

**Why?**
This tool can be used for operational needs. It helps inspect what timers are scheduled for specific time range. For now, additional, external, tooling is needed to visualize results. 

**How did you test it?**

**Potential risks**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	github.com/otiai10/copy v1.1.1
 	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709
 	github.com/pierrec/lz4 v0.0.0-20190701081048-057d66e894a4 // indirect
+	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v1.2.1 // indirect
 	github.com/robfig/cron v1.2.0
 	github.com/sirupsen/logrus v1.4.2
@@ -55,7 +56,7 @@ require (
 	github.com/uber-go/tally v3.3.15+incompatible
 	github.com/uber/ringpop-go v0.8.5
 	github.com/uber/tchannel-go v1.16.0
-	github.com/urfave/cli v1.20.0
+	github.com/urfave/cli v1.22.4
 	github.com/valyala/fastjson v1.4.1
 	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
 	go.opencensus.io v0.22.2 // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -55,7 +56,7 @@ require (
 	github.com/uber-go/tally v3.3.15+incompatible
 	github.com/uber/ringpop-go v0.8.5
 	github.com/uber/tchannel-go v1.16.0
-	github.com/urfave/cli v1.20.0
+	github.com/urfave/cli v1.22.4
```
mantas-sidlauskas, line 59: Updating to include get https://github.com/urfave/cli/issues/790 resolved

tools/cli/admin.go
```diff
@@ -815,6 +819,59 @@ func newDBCommands() []cli.Command {
...
+		{
+			Name:  "timers",
+			Usage: "get timers from multiple shards for given time range",
+			Flags: append(getDBFlags(),
+				cli.IntFlag{
+					Name:  FlagLowerShardBound,
+					Usage: "lower bound of shard to scan (inclusive)",
+					Value: 0,
+				},
+				cli.IntFlag{
```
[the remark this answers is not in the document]
mantas-sidlauskas: No, not really, will update accordingly 

Thanks

tools/cli/admin.go
```diff
@@ -815,6 +819,59 @@ func newDBCommands() []cli.Command {
...
+					Name:  FlagDomainID,
+					Usage: "filter tasks by DomainID",
+				},
+				cli.BoolFlag{
+					Name:  FlagSkipErrorMode,
+					Usage: "skip errors",
+				},
+				cli.IntSliceFlag{
+					Name:  FlagTimerType,
+					Usage: "filter by timer types, default - 2 (User Timer). Can be provided multiple times",
```
vancexu, line 866: Could you add all the timerType mappings in the usage? 
Also may also consider adding a all type (-1) and make it default, so that we don't need -ttype 0 1 2 3 4 ...

tools/cli/adminDBTimers.go
```diff
@@ -0,0 +1,169 @@
...
+
+		resp := &persistence.GetTimerIndexTasksResponse{}
+
+		op := func() error {
+			var err error
+			resp, err = ratelimitedClient.GetTimerIndexTasks(&req)
+			return err
+		}
+
+		for attempt := 0; attempt < maxDBRetries; attempt++ {
```
yycptt, line 127: Can we rely on the retry policy itself (or may be create a new retry policy) here, instead of adding an extra layer of retry?

tools/cli/admin.go
yycptt: +1 for adding a all type

tools/cli/util.go
```diff
@@ -511,6 +511,14 @@ func mapKeysToArray(m map[string]string) []string {
 	return out
 }
 
+func intSliceToMap(s []int) map[int]struct{} {
```
vytautas-karpavicius, line 514: IMO, you should call this a set.

PR #3406, opened 2020-07-22
Task Redispatcher

yycptt:
**What changed?**
* create a new component for all task redispatching logic
* assign task priority only once
* stop redispatching tasks for a certain priority if task ch for that priority is full
* redispatch only part of the tasks when redispatch queue is full
* throttle/backoff task loading when redispatch queue still has too many tasks after redispatching

**Why?**
Code refactoring + improve performance for task redispatching when system is under high load

**How did you test it?**
Tested in test environment

**Potential risks**
In the worst case task may get stuck in the redispatch queue


service/history/queue/timer_queue_processor_base.go
```diff
@@ -203,8 +206,18 @@ processorPumpLoop:
 		case <-t.shutdownCh:
 			break processorPumpLoop
 		case <-t.timerGate.FireChan():
-			if t.redispatchQueue.Len() > t.options.MaxRedispatchQueueSize() {
-				t.redispatchTasks()
+			maxRedispatchQueueSize := t.options.MaxRedispatchQueueSize()
+			if t.redispatcher.Size() > maxRedispatchQueueSize {
+				t.redispatcher.Redispatch(maxRedispatchQueueSize)
+				t.redispatcher.Redispatch(maxRedispatchQueueSize)
```
yux0, line 212: Why do we need to do redispatch twice here?

yycptt: Thanks for catching this! I think it's added by mistake when resolving merge conflicts. Removed.

PR #3419, opened 2020-07-30
[Scanner] Use constant for current execution run id

mkolodezny:


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -2094,7 +2094,7 @@ func (d *cassandraPersistence) ListCurrentExecutions(
 		response.Executions = append(response.Executions, &p.CurrentWorkflowExecution{
 			DomainID:     result["domain_id"].(gocql.UUID).String(),
 			WorkflowID:   result["workflow_id"].(string),
-			RunID:        result["run_id"].(gocql.UUID).String(),
+			RunID:        permanentRunID,
```
yux0, line 2097: May be remove the run id from templateListCurrentExecutionsQuery

PR #3420, opened 2020-07-30
Fix multi-cursor queue polling logic

yycptt:
**What changed?**
* Fix an NPE exception in updateAckLevel() when queueCollection is empty
* Fix the relationship between look ahead and task notification logic for timer processor
* Wait for new task notification instead of periodically polling DB when there's no more task for non-default queue
* Add metrics for throttle task polling
* Make polling backoff interval a dynamic config

**Why?**
Fix multi-cursor queue polling logic 

**How did you test it?**
Integration test, local canary

**Potential risks**
N/A, feature flag disabled.


service/history/queue/processor_base.go
```diff
@@ -169,6 +171,11 @@ func (p *processorBase) updateAckLevel() (bool, error) {
 	totalPengingTasks := 0
 	for _, queueCollection := range p.processingQueueCollections {
 		ackLevel, numPendingTasks := queueCollection.UpdateAckLevels()
+		if ackLevel == nil {
```
vancexu, line 174: Looks like there are other places maybe need such check, can you double check and add if needed.
Another tiny improvement suggestion is on `queueCollection.UpdateAckLevels()`, check empty case at the beginning will make edge case easier to identify and save bits of ops when empty.

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -196,7 +211,11 @@ processorPumpLoop:
 		case <-t.shutdownCh:
 			break processorPumpLoop
 		case <-t.notifyCh:
-			t.upsertPollTime(defaultProcessingQueueLevel, time.Time{})
+			t.queueCollectionsLock.RLock()
+			for _, queueCollection := range t.processingQueueCollections {
```
yux0, line 215: Maybe mention why do you need to insert poll time to each queue?

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -206,7 +225,10 @@ processorPumpLoop:
 					// if redispatcher still has a large number of tasks
 					// this only happens when system is under very high load
 					// we should backoff here instead of keeping submitting tasks to task processor
-					time.Sleep(loadQueueTaskThrottleRetryDelay)
+					time.Sleep(backoff.JitDuration(
```
yux0, line 228: do you need a jitter here?

service/history/queue/processor_base.go
yycptt: Thanks for the suggestion. Added the same check when reseting queue states. Added the check for empty case in both `queueCollection.UpdateAckLevels()` and `queueCollection.Split()`.

service/history/queue/transfer_queue_processor_base.go
yycptt: Added comments in both tx and timer processor base

service/history/queue/transfer_queue_processor_base.go
yycptt: Just following the pattern for other time intervals. I think jitter might still be necessary as when the host level task chan is full, multiple queue processor will encounter the problem at almost the same time.

PR #3423, opened 2020-07-31
Fix sticky decision keeps failing on ConditionFailedError

vancexu:
**What changed?**
Fix a rare issue for sticky task keeps failing on ConditionFailedError

**Why?**
When rangeID changed and sticky task ConditionFailedError happened, it is possible for the sticky tasklist to go into a bad state that none task can be added to it because of endless ConditionFailedError. This will cause latency increase for workflow, because all sticky decision task timeout (and workflow can only progress by using normal tasklist)
Once it happened, the only possible way to recover is for user to restart workers (with those affected tasklists)

We want to stablize the tasklist rangeID so that no need for worker to restart to recover using the sticky tasklist. 
This PR use a relatively conservative approach to stop infinite retrying sticky failures on ConditionFailedError. By reducing concurrent access to tasklist, it will be more likely to break the endless ConditionFailedError. 

Another approach would be clear stickness on worker side, but there would be bigger changes and risks.

**How did you test it?**
Existing tests

**Potential risks**
n/a


service/history/taskProcessor.go
```diff
@@ -211,6 +211,12 @@ FilterLoop:
 				scope.RecordTimer(metrics.TaskAttemptTimer, time.Duration(task.attempt))
 				task.logger.Error("Critical error processing task, retrying.",
 					tag.Error(err), tag.OperationCritical, tag.TaskType(task.task.GetTaskType()))
+				if isStickyTaskConditionError(err) {
+					// sticky task could end up into endless loop in rare cases and
+					// cause worker keep getting decision timeout unless restart.
+					// return nil here to break the endless loop
+					return nil
+				}
```
yycptt, line 219: Seems like the logic belongs to `handleTaskError` func? task.attempt should be available there. 

Can you also update the same logic for new task processing logic in `service/history/task/task.go`?

vancexu: Moved and added

service/history/task/task.go
```diff
@@ -320,6 +321,13 @@ func (t *taskBase) HandleErr(
 		return nil
 	}
 
+	if t.attempt > t.maxRetryCount() && common.IsStickyTaskConditionError(err) {
```
yycptt, line 324: Please grab a lock when accessing the `attempt` field or call `GetAttempt()` which does it for you. 

The `attempt` field will also be accessed by the split policy to determine the if the task got stuck and split it to a new queue, that's why I added a lock for it.

yycptt: Hmmm. Actually it's probably fine reading that field without a lock in this function as there's won't be any concurrent write to this field (write only happens in the defer func). Thoughts?

vancexu: I think since we have `GetAttempt` it more correct and consistent to just use it (and avoid mislead other reader of related code).
If we want to optimize, we could try RWlock on this later.

PR #3438, opened 2020-08-06
Disable concrete execution scanner by default

anish531213:
**What changed?**
Changed the default value for concrete execution scanner to false.

**Why?**
This was causing persistence error to increase for our open-source customers. As it looks up the dynamic config before running the scanner, leaving it disabled by default looks like a good option.

**How did you test it?**

**Potential risks**


vancexu:
https://github.com/uber/cadence/issues/3433

PR #3441, opened 2020-08-07
Update the config to be able to set the value by shard

yux0:
**What changed?**
Update the config to be able to set the value by shard

**Why?**
replication speed should be configurable by shard

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/history/replicatorQueueProcessor.go
+++ b/service/history/replicatorQueueProcessor.go
@@ -38,6 +38,7 @@ import (
 	"github.com/uber/cadence/common/messaging"
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/common/service/dynamicconfig"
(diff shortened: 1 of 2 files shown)


service/history/replicatorQueueProcessor.go
```diff
@@ -441,7 +442,8 @@ func (p *replicatorQueueProcessorImpl) getTasks(
 		metrics.InstanceTag(strconv.Itoa(p.shard.GetShardID())),
 	)
 	taskGeneratedTimer := replicationScope.StartTimer(metrics.TaskLatency)
-	taskInfoList, hasMore, err := p.readTasksWithBatchSize(lastReadTaskID, p.fetchTasksBatchSize)
+	shardID := p.shard.GetShardID()
+	taskInfoList, hasMore, err := p.readTasksWithBatchSize(lastReadTaskID, p.fetchTasksBatchSize(shardID))
```
yycptt, line 446: Please keep an eye on the GC activity caused by this function is invoked very often

PR #3443, opened 2020-08-10
Update metrics cache to be unblocking

anish531213:
**What changed?**
Update metrics cache to be non-blocking on Get().

**Why?**
There might be task processing latency increase if metrics cache is blocked on Get(). This change will help to amortize the task processing latency by using a non-blocking cache and doing flush of all put requests every 10s.

**How did you test it?**
Unit test.

**Potential risks**
Might increase task processing latency.

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -23,6 +23,8 @@ package cache
 import (
 	"time"
 
+	"github.com/uber/cadence/common"
+
 	"github.com/uber/cadence/common/metrics"
 )
 
@@ -130,8 +132,6 @@ type DomainMetricsScopeCache interface {
 	Get(domainID string, scopeIdx int) (metrics.Scope, bool)
 	// Put adds metrics scope for a domainID and scopeIdx
 	Put(domainID string, scopeIdx int, metricsScope metrics.Scope)
-	// Start the metrics scope cache
-	Start()
-	// Stops the metrics scope cache
-	Stop()
+
+	common.Daemon
 }
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -130,4 +130,6 @@ type DomainMetricsScopeCache interface {
 	Get(domainID string, scopeIdx int) (metrics.Scope, bool)
 	// Put adds metrics scope for a domainID and scopeIdx
 	Put(domainID string, scopeIdx int, metricsScope metrics.Scope)
+	// Stops the metrics scope cache
+	Stop()
 }
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -130,6 +130,8 @@ type DomainMetricsScopeCache interface {
 	Get(domainID string, scopeIdx int) (metrics.Scope, bool)
 	// Put adds metrics scope for a domainID and scopeIdx
 	Put(domainID string, scopeIdx int, metricsScope metrics.Scope)
+	// Start the metrics scope cache
+	Start()
 	// Stops the metrics scope cache
 	Stop()
 }
--- a/common/cache/metricsScopeCache.go
+++ b/common/cache/metricsScopeCache.go
@@ -23,11 +23,11 @@
 package cache
 
 import (
-	"bytes"
-	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/uber/cadence/common"
 
 	"github.com/uber/cadence/common/metrics"
 )
@@ -35,7 +35,7 @@ import (
 const flushBufferedMetricsScopeDuration = 10 * time.Second
 
 type (
-	metricsScopeMap map[string]metrics.Scope
+	metricsScopeMap map[string]map[int]metrics.Scope
 
 	buffer struct {
 		sync.RWMutex
@@ -43,6 +43,7 @@ type (
 	}
 
 	domainMetricsScopeCache struct {
+		status        int32
 		buffer        *buffer
 		cache         atomic.Value
 		closeCh       chan struct{}
@@ -51,7 +52,7 @@ type (
 )
 
 // NewDomainMetricsScopeCache constructs a new domainMetricsScopeCache
-func NewDomainMetricsScopeCache() *domainMetricsScopeCache {
+func NewDomainMetricsScopeCache() DomainMetricsScopeCache {
 
 	mc := &domainMetricsScopeCache{
 		buffer: &buffer{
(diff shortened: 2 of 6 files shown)


common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+	"time"
 
 	"github.com/uber/cadence/common/metrics"
 )
 
+const flushBufferedMetricsScopeDuration = 10 * time.Second
+
 type domainMetricsScopeCache struct {
-	sync.RWMutex
 	scopeMap map[string]metrics.Scope
```
vancexu, line 38: maybe use `metricsScopeMap` instead of `map[string]metrics.Scope` to be consistent?

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+
+var cache atomic.Value
+
 // NewDomainMetricsScopeCache constructs a new domainMetricsScopeCache
 func NewDomainMetricsScopeCache() DomainMetricsScopeCache {
-	return &domainMetricsScopeCache{
+	mc := &domainMetricsScopeCache{
 		scopeMap: make(map[string]metrics.Scope),
+		buffer: &buffer{
+			bufferMap: make(map[string]metrics.Scope),
```
vancexu, line 56: maybe use `metricsScopeMap` instead of `map[string]metrics.Scope` to be consistent?

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+
 type domainMetricsScopeCache struct {
-	sync.RWMutex
 	scopeMap map[string]metrics.Scope
+	buffer   *buffer
+}
+
+type buffer struct {
+	sync.RWMutex
+	bufferMap map[string]metrics.Scope
```
vancexu, line 44: maybe use `metricsScopeMap` instead of `map[string]metrics.Scope` to be consistent?

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+			bufferMap: make(map[string]metrics.Scope),
+		},
+	}
+	cache.Store(make(metricsScopeMap))
+	go mc.flushBufferedMetricsScope()
+	return mc
+}
+
+func (c *domainMetricsScopeCache) flushBufferedMetricsScope() {
+	for {
```
vancexu, line 65: seems this thread need some exit logic to prevent leaking

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+}
+
+type buffer struct {
+	sync.RWMutex
+	bufferMap map[string]metrics.Scope
 }
 
+type metricsScopeMap map[string]metrics.Scope
+
+var cache atomic.Value
```
vancexu, line 49: is this cache shared by multiple DomainMetricsScopeCache? if not, better to attach it to DomainMetricsScopeCache, instead of global

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+	defer c.buffer.Unlock()
 
-	var buffer bytes.Buffer
-	buffer.WriteString(domainID)
-	buffer.WriteString("_")
-	buffer.WriteString(strconv.Itoa(scopeIdx))
-	key := buffer.String()
-
-	c.scopeMap[key] = scope
+	key := joinStrings(domainID, "_", strconv.Itoa(scopeIdx))
```
vancexu, line 111: may worth a small improvement to get key at the beginning then lock.

common/cache/metricsScopeCache_test.go
```diff
@@ -49,6 +50,8 @@ func TestGetMetricsScope(t *testing.T) {
 		metricsCache.Put(t.domainID, t.scopeID, mockMetricsScope)
 	}
 
+	time.Sleep(11 * time.Second)
```
yycptt, line 53: Please don't sleep that long time in unit tests. This plus the test below will cause the entire test file take 20 more seconds. You can consider change the flush duration to a parameter when creating the scope cache, and use a very small duration for test, like 100 milliseconds.

another nit: probably consider move those two tests into a single test suite.

common/cache/metricsScopeCache.go
yycptt: +1 for attaching it to `DomainMetricsScopeCache`. We may have multiple scope caches for different parts of the system in the future.

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/uber/cadence/common/metrics"
 )
 
+const flushBufferedMetricsScopeDuration = 10 * time.Second
+
 type domainMetricsScopeCache struct {
```
yycptt, line 37: nit: please group multiple type definitions with 
```
type (
    domainMetricsScopeCache struct { ...  }

    ... // other defs
)
```

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
 func NewDomainMetricsScopeCache() DomainMetricsScopeCache {
-	return &domainMetricsScopeCache{
+	mc := &domainMetricsScopeCache{
 		scopeMap: make(map[string]metrics.Scope),
+		buffer: &buffer{
+			bufferMap: make(map[string]metrics.Scope),
+		},
+	}
+	cache.Store(make(metricsScopeMap))
+	go mc.flushBufferedMetricsScope()
```
yycptt, line 60: You'll need a way/signal to stop this goroutine.

common/cache/metricsScopeCache.go
yycptt, line 38: Is this field used anywhere? Seems like we can get rid of it?

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,96 @@ import (
...
+	return mc
+}
+
+func (c *domainMetricsScopeCache) flushBufferedMetricsScope() {
+	for {
+		select {
+		case <-time.After(flushBufferedMetricsScopeDuration):
```
yycptt, line 67: May worth adding a unit test for the case where scopes are added to the cache in different iterations of this flush loop (to make sure previously flushed scope are still there)

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,113 @@ import (
...
+					scopeMap[key] = val
+				}
+
+				c.cache.Store(scopeMap)
+				c.buffer.bufferMap = make(metricsScopeMap)
+			}
+			c.buffer.Unlock()
+
+		case <- c.closeCh:
+			break
```
yycptt, line 98: I don't think this will work. `break` here will exit the `select` statement, not the for loop. https://golang.org/ref/spec#Break_statements

common/cache/metricsScopeCache.go
```diff
@@ -26,47 +26,112 @@ import (
...
+			return
+		}
 	}
 }
 
 // Get retrieves scope for domainID and scopeIdx
 func (c *domainMetricsScopeCache) Get(domainID string, scopeIdx int) (metrics.Scope, bool) {
-	c.RLock()
-	defer c.RUnlock()
+	key := joinStrings(domainID, "_", strconv.Itoa(scopeIdx))
```
meiliang86, line 100: Have we considered using map[string]map[int]metrics.Scope to avoid this step all together?

common/cache/cache.go
```diff
@@ -130,4 +132,6 @@ type DomainMetricsScopeCache interface {
 	Get(domainID string, scopeIdx int) (metrics.Scope, bool)
 	// Put adds metrics scope for a domainID and scopeIdx
 	Put(domainID string, scopeIdx int, metricsScope metrics.Scope)
+
+	common.Daemon
```
yycptt, line 136: nit: put embedded interface in the beginning.

common/cache/cache.go
```diff
@@ -23,6 +23,8 @@ package cache
 import (
 	"time"
 
+	"github.com/uber/cadence/common"
+
```
yycptt, line 27: nit: remove empty line

common/cache/metricsScopeCache.go
```diff
@@ -23,50 +23,122 @@
...
+					scopeMap[key] = val
+				}
+
+				// Copy from buffered array
+				for key, val := range c.buffer.bufferMap {
+					if _, ok := scopeMap[key]; !ok {
+						scopeMap[key] = map[int]metrics.Scope{}
+					}
+					for k, v := range val {
+						scopeMap[key][k] = v
```
yycptt, line 88: Will this result in race condition? I mean the value got copied on L79 is a map reference, this map will be read by the `Get()` method and be updated here concurrently I think? My understanding is that in this case whatever value got stored to the `atomic.Value` should be read only.

I am not sure why the concurrent test doesn't report data race though.

anish531213: oh, i see, then copy everything over will fix this.

anish531213: Updated.

PR #3444, opened 2020-08-11
Fix update both replication queue ack level per Kafka queue and RPC replicaiton

yux0:
**What changed?**
Fix update both replication queue ack level per Kafka queue and RPC replication

**Why?**
During the feature rollback, the Kafka queue can skip the messages which are processed by RPC replicaiton

**How did you test it?**

**Potential risks**


yux0:
> Revert IDL changes?

Removed

PR #3446, opened 2020-08-11
DB2 persistence support

yandooo:
https://github.com/uber/cadence/issues/3445

DB2 persistence support

https://github.com/uber/cadence/issues/3445

Tested manually. Please show examples of good MySQL/PostgreSQL tests to adopt for DB2 execution.

DB2 support should not affect any other persistence layers and optional for compilation.


cmd/server/config/base.yaml
```diff
@@ -0,0 +1,138 @@
+persistence:
```
andrewjdawson2016, line 1: We should not a config here. Config files should live under config directory.

cmd/server/main.go
```diff
@@ -24,6 +24,7 @@ import (
 	"os"
 
 	"github.com/uber/cadence/cmd/server/cadence"
+	//_ "github.com/uber/cadence/common/persistence/sql/sqlplugin/db2"      // needed to load db2 plugin
```
andrewjdawson2016, line 27: Please don't check in commented out code.

cmd/tools/sql/main.go
```diff
@@ -23,6 +23,7 @@ package main
 import (
 	"os"
 
+	//_ "github.com/uber/cadence/common/persistence/sql/sqlplugin/db2"      // needed to load db2 plugin
```
andrewjdawson2016, line 26: Please don't check in commented out code.

common/persistence/sql/sqlplugin/db2/domain.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+const (
+	createDomainQuery = `INSERT INTO 
+	cadence.domains (id, name, is_global, data, data_encoding)
+ VALUES(?, ?, ?, ?, ?)`
+
+	updateDomainQuery = `UPDATE cadence.domains 
+ SET name = ?, data = ?, data_encoding = ?
+ WHERE shard_id=54321 AND id = ?`
```
andrewjdawson2016, line 37: Use the constant you defined instead of hardcoding here.

common/persistence/sql/sqlplugin/db2/execution.go
```diff
@@ -0,0 +1,402 @@
...
+(shard_id, domain_id, workflow_id, run_id, create_request_id, state, close_status, start_version, last_write_version) VALUES
+(:SHARD_ID, :DOMAIN_ID, :WORKFLOW_ID, :RUN_ID, :CREATE_REQUEST_ID, :STATE, :CLOSE_STATUS, :START_VERSION, :LAST_WRITE_VERSION)`
+
+	deleteCurrentExecutionQuery = "DELETE FROM cadence.current_executions WHERE shard_id=? AND domain_id=? AND workflow_id=? AND run_id=?"
+
+	getCurrentExecutionQuery = `SELECT
+shard_id, domain_id, workflow_id, run_id, create_request_id, state, close_status, start_version, last_write_version
+FROM cadence.current_executions WHERE shard_id = ? AND domain_id = ? AND workflow_id = ?`
+
+	// significantly updated!
```
andrewjdawson2016, line 62: not a useful comment.

schema/db2/cadence/schema.sql
```diff
@@ -0,0 +1,257 @@
...
+  workflow_id VARCHAR(255) NOT NULL,
+  run_id CHAR(16) NOT NULL,
+  initiated_id BIGINT NOT NULL,
+--
+  data BLOB NOT NULL,
+  data_encoding VARCHAR(16),
+  PRIMARY KEY (shard_id, domain_id, workflow_id, run_id, initiated_id)
+);
+
+CREATE TABLE cadence.buffered_replication_task_maps (
```
andrewjdawson2016, line 194: This table is not being used.

Dockerfile.db2
```diff
@@ -0,0 +1,134 @@
+ARG TARGET=server
```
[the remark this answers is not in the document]
yandooo: Need clarification here. There is an original Dockerfile in the root of the project. It's the same Dockerfile but for DB2. The way how to build is the same as for the original. 
Unfortunately, there is no easy way to use alpine base images as need for gnu compile libs. 

docker-compose files are in docker directory and they are the same as for MySQL and PostgreSQL

cmd/server/config/base.yaml
yandooo: deleted

common/persistence/sql/sqlplugin/db2/domain.go
yandooo: copied from MySQL and PostgreSQL code. Also, it's hardcoded into schema.sql files

schema/db2/cadence/schema.sql
yandooo: it's in sync with MySQL and PostgreSQL schemas

common/persistence/sql/sqlplugin/db2/execution.go
yandooo: deleted

cmd/tools/sql/main.go
yandooo: the hope is `docs/persistence.md` is read first before trying to run on db2

cmd/server/main.go
yandooo: deleted

Dockerfile.db2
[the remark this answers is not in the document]
yandooo: the doc is in first commit `docs/persistence.md`. Section `DB2`

meiliang86:
@yandoo We have a plugin model for persistence so this piece of code can live outside of the Cadence repo.

yandooo:
@meiliang86 how is it different from MySQL/PostgreSQL code? How are we going to keep code between plugins in sync? In particular, when core logic/plugin contract/schema changes? 
Also, it's not clear how the plugin system addresses Makefile schema updates, a compilation that requires modification of cmd/* files, docker-compose yaml files, dynamic config generation, environment vars propagation. These files are in Cadence repo.

common/persistence/sql/sqlplugin/db2/domain.go
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yux0, line 1: Run a "make copyright" to get the latest header

longquanzheng:
> @meiliang86 how is it different from MySQL/PostgreSQL code? How are we going to keep code between plugins in sync? In particular, when core logic/plugin contract/schema changes?
> Also, it's not clear how the plugin system addresses Makefile schema updates, a compilation that requires modification of cmd/* files, docker-compose yaml files, dynamic config generation, environment vars propagation. These files are in Cadence repo.

Those are great questions. Let me see if I can help. 
First of all I generally agree that core Cadence cannot maintain all the database implementation, as there are so many of them and it will be too much burden and impossible for Cadence team to maintain.

But what you are worrying is important too. We can work together to keep some standard of the persistence interface, whenever there is some changes, we should make it clear in the announcement like in release notes. We should have a better format for those releasing notes.

Also it should be required that in-compatible changes should be proposed in a way that any database can transition to --- for example, there should be a middle version that allow migrating to new breaking version. 

I implemented the plugin with SQLite. As the consumer of the interface, we have to look close with the version changes anyway, if we are upgrading the server to newer version. 

I am also outside of Uber and am currently implementing MongoDB plugin, and I don't think these should be checked in as well. If you are interested, we can work together to address those issues.

yandooo:
@longquanzheng of course, we can keep guessing why things are done one way or another. As things stand the only way to properly integrate plugin code is to get it into Cadence codebase. Otherwise, it's a separate piece of code that is hard to consume in a reliable way. It's, perhaps, one of the reasons PostgreSQL is merged into core rather than delivered into the separate repo in DIY mode. 

In general, I have nothing against core committers to come here and say "we are not interested in DB2 plugin, keep it outside of the core". It still raises the question of why MySQL, PostgreSQL is in the core but at least this PR can be closed and we tick the box for unsuccessful open source contribution.

longquanzheng:
> @longquanzheng of course, we can keep guessing why things are done one way or another. As things stand the only way to properly integrate plugin code is to get it into Cadence codebase. Otherwise, it's a separate piece of code that is hard to consume in a reliable way. It's, perhaps, one of the reasons PostgreSQL is merged into core rather than delivered into the separate repo in DIY mode.
> 
> In general, I have nothing against core committers to come here and say "we are not interested in DB2 plugin, keep it outside of the core". It still raises the question of why MySQL, PostgreSQL is in the core but at least this PR can be closed and we tick the box for unsuccessful open source contribution.

Ideally yes they should be in this repo, however it's not scalable to maintain all database plugins. There are tens of or even hundreds of them, I believe you also agree if we put them here it will be a nightmare. 

For why MySQL/Postgres are different or important. It's just historical reason that more people asking for it at easier than others. Or in other words they are more popular open source DB. 

However, you do raise a very good point that there shouldn't be difference. I think we could ask Cadence team make another repo and put all other plugins there, and let us all maintain the plugins together as community effort. After all it's still community product. And then we can maintain a wiki like this https://github.com/golang/go/wiki/SQLDrivershttps://github.com/golang/go/wiki/SQLDrivers

But this core repo still have to put one plugin for SQL one for NoSQL, because the testing needs it. 

How does that sound?

yandooo:
@longquanzheng  If committers think it's better to extract PostgreSQL code into separate repo - let's do this first (with proper updates to docker-compose files, start shell scripts for development and avoid overriding go source files to include the plugin, enable plugin build/upload images to docker hub, etc.), so we'll contribute to the effort to get our stuff there.  Something like `uber/cadence-sql-plugins` should work just fine. Currently, it's inconsistent by example.

Important note. The reason moving code from private single contributor repo is to get wider community visibility and support under the original core project company. Especially, if the plugin(s) in any way looks strategic to extend Cadence production-level adoption. 

> I think we have another repo and put all other plugins there(If you agree we can do it in this repo: https://github.com/longquanzheng/cadence-extensions ) Let us all maintain the plugins together as community effort. After all it's still community product. And then we can maintain a wiki like this https://github.com/golang/go/wiki/SQLDrivers

Not many people want/can contribute to the wild single person repo with no license and CLA and also it does not make sense - nothing stops you from creating the wiki and reference other plugins implementation there. The same is done in the link you mentioned.

longquanzheng:
> Not many people want/can contribute to the wild single person repo
Yes, I agree, that's why it's up to you. It's also the fact the Golang SQL driver has reference to personal repo for some database plugins. I personally believe it would be nice if you could do all the plugin work (docker/makefil/etc) in your repo, and let Cadence refers to it.

andrewjdawson2016:
@longquanzheng I agree with you that we should have these implementations live outside of Cadence. I also buy @yandooo argument that the support for this today needs to be improved. I am wondering if as part of your efforts to build a new cadence-sql extension you will also do the following
- Work with the cadence team to create a repo to move all these plugins into 
- Ensure we have a standard practice around make files for these plugins

I think if these concerns were addressed in addition to having a solid production quality example of a cadence sql extension it would be much easier for other contributors to add extensions. 

Sadly my work needs to stay within Uber because it depends on internal Uber dependencies otherwise I would also add to this extensions repo. 

Let me know your thoughts. Also happy to setup a zoom call to chat about these options. For now I will close this PR because it seems @longquanzheng and I agree that this should live outside of cadence.

PR #3447, opened 2020-08-11
Slow down decision tasks after failure

vancexu:
**What changed?**
Better handling of decision task failure.

**Why?**
Right now decision task will be retried immediately if it failed. This will created a lot of load on cadence if large number of decision tasks are failing (e.g. customer rolls out a bad deployment causing workflow to panic).

**How did you test it?**
unit test
manual tests

**Potential risks**
Worker will wait longer than expected (up to 5min) for new decision in some failure cases.

### DIFF
--- a/service/history/execution/mutable_state_task_generator.go
+++ b/service/history/execution/mutable_state_task_generator.go
@@ -108,6 +108,7 @@ const (
 	defaultWorkflowRetentionInDays      int32 = 1
 	defaultInitIntervalForDecisionRetry       = 1 * time.Minute
 	defaultMaxIntervalForDecisionRetry        = 5 * time.Minute
+	defaultJitterCoefficient                  = 0.2
 )
 
 var _ MutableStateTaskGenerator = (*mutableStateTaskGeneratorImpl)(nil)
@@ -576,10 +577,10 @@ func getNextDecisionTimeout(attempt int64, defaultStartToCloseTimeout time.Durat
 
 	nextInterval := float64(defaultInitIntervalForDecisionRetry) * math.Pow(2, float64(attempt-2))
 	nextInterval = math.Min(nextInterval, float64(defaultMaxIntervalForDecisionRetry))
-	jitterPortion := int(0.2 * nextInterval)
+	jitterPortion := int(defaultJitterCoefficient * nextInterval)
 	if jitterPortion < 1 {
 		jitterPortion = 1
 	}
-	nextInterval = nextInterval*0.8 + float64(rand.Intn(jitterPortion))
+	nextInterval = nextInterval*(1-defaultJitterCoefficient) + float64(rand.Intn(jitterPortion))
 	return time.Duration(nextInterval)
 }
--- a/service/history/execution/mutable_state_task_generator.go
+++ b/service/history/execution/mutable_state_task_generator.go
@@ -24,6 +24,8 @@ package execution
 
 import (
 	"fmt"
+	"math"
+	"math/rand"
 	"time"
 
 	"github.com/uber/cadence/.gen/go/shared"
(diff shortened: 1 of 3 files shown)


service/worker/service.go
```diff
@@ -133,7 +133,7 @@ func NewConfig(params *service.BootstrapParams) *Config {
 			Persistence:            &params.PersistenceConfig,
 			ClusterMetadata:        params.ClusterMetadata,
 			TaskListScannerEnabled: dc.GetBoolProperty(dynamicconfig.TaskListScannerEnabled, true),
-			HistoryScannerEnabled:  dc.GetBoolProperty(dynamicconfig.HistoryScannerEnabled, true),
+			HistoryScannerEnabled:  dc.GetBoolProperty(dynamicconfig.HistoryScannerEnabled, false),
```
yycptt, line 136: Why do we need to disable this one by default? Concrete execution history has already been disabled I think.

service/history/execution/mutable_state_task_generator_test.go
```diff
@@ -0,0 +1,56 @@
...
+		defaultStartToCloseTimeout,
+		defaultStartToCloseTimeout,
+		defaultInitIntervalForDecisionRetry,
+		defaultInitIntervalForDecisionRetry * 2,
+		defaultInitIntervalForDecisionRetry * 4,
+		defaultMaxIntervalForDecisionRetry,
+		defaultMaxIntervalForDecisionRetry,
+		defaultMaxIntervalForDecisionRetry,
+	}
+	for i := 0; i < 8; i++ {
```
yycptt, line 43: nit: `i < len(expectedResult)`

service/history/execution/mutable_state_task_generator.go
```diff
@@ -554,3 +568,18 @@ func (r *mutableStateTaskGeneratorImpl) getTargetDomainID(
...
 }
+
+func getNextDecisionTimeout(attempt int64, defaultStartToCloseTimeout time.Duration) time.Duration {
+	if attempt <= 1 {
+		return defaultStartToCloseTimeout
+	}
+
+	nextInterval := float64(defaultInitIntervalForDecisionRetry) * math.Pow(2, float64(attempt-2))
+	nextInterval = math.Min(nextInterval, float64(defaultMaxIntervalForDecisionRetry))
+	jitterPortion := int(0.2 * nextInterval)
```
yycptt, line 579: nit: define the jitter coefficient as a const at the beginning of the file and use `(1 - coefficient)` instead of 0.8 below?

service/history/execution/mutable_state_task_generator.go
```diff
@@ -554,3 +568,18 @@ func (r *mutableStateTaskGeneratorImpl) getTargetDomainID(
 
 	return targetDomainID, nil
 }
+
+func getNextDecisionTimeout(attempt int64, defaultStartToCloseTimeout time.Duration) time.Duration {
+	if attempt <= 1 {
+		return defaultStartToCloseTimeout
+	}
+
+	nextInterval := float64(defaultInitIntervalForDecisionRetry) * math.Pow(2, float64(attempt-2))
```
andrewjdawson2016, line 577: Don't we have an exponential backoff library we can use here?

andrewjdawson2016: backoff.Retrier I think

service/history/execution/mutable_state_task_generator_test.go
vancexu: addressed

service/history/execution/mutable_state_task_generator.go
vancexu: addressed

service/history/execution/mutable_state_task_generator.go
vancexu: backoff.Retrier is kind of stateful with attempt recorded. 
Here we need a "stateless" backoff just to getNextInterval with given attempt. (we cannot attach one retrier to taskGenerator to achieve this because retrier doesn't know which task it is retrying)

service/worker/service.go
vancexu: I think team agreed to disable scanner by default.

yycptt: I think team agreed to stop execution scanner by default? This one is history scanner for deleting invalid first workflow history batch?

PR #3462, opened 2020-08-19
Admin describe queue state command

yycptt:
**What changed?**
* Add an admin CLI command for describing the current multi-cursor queue state
* Fix some bugs in the reset queue command
* Fix the estimation for transfer task queue look ahead during split
* Update some dynamic config names used by queue splitting

**Why?**
* Useful as an oncall and debug tool

**How did you test it?**
Staging 2 test

**Potential risks**
N/A, feature flag disabled


service/history/queue/processor_base.go
```diff
@@ -374,7 +374,7 @@ func (p *processorBase) resetProcessingQueueStates() (*ActionResult, error) {
 				defaultProcessingQueueLevel,
 				minAckLevel,
 				maxReadLevel,
-				NewDomainFilter(nil, false),
+				NewDomainFilter(nil, true),
```
yux0, line 377: What is this change for?

yycptt: It's a bug fix. The boolean here means whether this domain filter is a reverse match or not. When there's only one queue, the reverse match should always be true (match all domains). Otherwise all tasks will be skipped.

PR #3463, opened 2020-08-20
fix transient decision serialization for getRawHistory feature

mkolodezny:
**What changed?**
subj

**Why?**
In case of DecisionTaskTimedOut the cadence generates additional ScheduledEvent and StartedEvent to let the client to replay and retry the failed decision task. 
The client expects batched events blob instead of single ones. Hence it failed to deserialize the history going into infinite timeout loop.
The actual error was swallowed on client: https://github.com/uber-go/cadence-client/pull/1019

**How did you test it?**
tested locally

**Potential risks**
getRawHistory feature


meiliang86:
Can this be unit tested?

anish531213:
Does the client get an error when it is deserializing just one event in a batch. I guess it is fine if the batch has >= 1 history events.

mkolodezny:
> Does the client get an error when it is deserializing just one event in a batch. I guess it is fine if the batch has >= 1 history events.

  the transient decision events were not batched into blob so the client gets into the error https://github.com/uber-go/cadence-client/blob/888e8c403f34979f08f819a556879a52d476fdbe/internal/common/serializer/history_serializer.go#L231

mkolodezny:
> Can this be unit tested?

sure, will follow up on this.

PR #3465, opened 2020-08-20
Replication task generation delay

yux0:
**What changed?**
Add delay between replication tasks generation

**Why?**
Reduce the burst load

**How did you test it?**
N/A

**Potential risks**
N/A

### DIFF
--- a/service/history/replicatorQueueProcessor.go
+++ b/service/history/replicatorQueueProcessor.go
@@ -63,6 +63,7 @@ type (
 		queueAckMgr
 
 		lastShardSyncTimestamp time.Time
+		taskGenerationWait     dynamicconfig.DurationPropertyFn
 	}
 )
 
@@ -142,7 +143,7 @@ func newReplicatorQueueProcessor(
 	)
 	processor.queueAckMgr = queueAckMgr
 	processor.queueProcessorBase = queueProcessorBase
-
+	processor.taskGenerationWait = config.ReplicationTaskGenerationWait
 	return processor
 }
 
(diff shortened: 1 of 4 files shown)


service/history/replicatorQueueProcessor.go
```diff
@@ -467,6 +468,7 @@ func (p *replicatorQueueProcessorImpl) getTasks(
 		readLevel = taskInfo.GetTaskID()
 		if replicationTask != nil {
 			replicationTasks = append(replicationTasks, replicationTask)
+			time.Sleep(p.taskGenerationWait())
```
yycptt, line 471: I am not sure how much timer scheduling resource it will cost on the CPU. From that point, I would prefer using a ratelimiter.

Also seems like it will try to load dynamic config every 10 milliseconds, that will also cause lots of GC activity. Probably use the same value across a single getTasks() call.

yux0: If there is concern on CPU, I updated to a rate limiter.

service/history/replicatorQueueProcessor.go
```diff
@@ -451,6 +455,7 @@ func (p *replicatorQueueProcessorImpl) getTasks(
 	var replicationTasks []*replicator.ReplicationTask
 	readLevel := lastReadTaskID
 	for _, taskInfo := range taskInfoList {
+		_ = p.rateLimiter.Wait(ctx)
```
yycptt, line 458: if context timedout during waiting for token, should we just abort and return an error?

yux0: It will return error from the below calls

PR #3466, opened 2020-08-20
Integrate current execution check with replication resender

yux0:
**What changed?**
Integrate current execution check with replication resender

**Why?**
For corrupted workflow, we don't need to put it into dlq.

**How did you test it?**

**Potential risks**
N/A

### DIFF
--- a/common/xdc/nDCHistoryResender.go
+++ b/common/xdc/nDCHistoryResender.go
@@ -36,6 +36,7 @@ import (
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/persistence"
+	checks "github.com/uber/cadence/common/reconciliation/common"
 	"github.com/uber/cadence/common/service/dynamicconfig"
 )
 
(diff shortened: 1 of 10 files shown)


common/xdc/nDCHistoryResender.go
```diff
@@ -131,7 +135,20 @@ func (n *NDCHistoryResenderImpl) SendSingleWorkflowHistory(
...
 		result, err := historyIterator.Next()
-		if err != nil {
+		switch err.(type) {
+		case nil:
+			// continue to process the events
+			break
+		case *shared.EntityNotExistsError:
+			// Case 1: the workflow pass the retention period
+			// Case 2: the workflow is corrupted
+			n.checkCurrentExecution(
```
yycptt, line 145: If the check here finds that the workflow is not corrupted, then does it mean the workflow has either passed retention period on the source cluster or the workflow/history is corrupted in the source cluster (for example a timer is fired on the corrupted workflow and generates some replication task)? and in this case do we still want to retry the replication task? Seems like we can just drop the task?

yux0: Yes. we can drop the task

PR #3467, opened 2020-08-20
Differentiate metrics for domainTagged metrics and task operation metrics

anish531213:
**What changed?**
Right now when metrics is emitted with domain tag, our current dashboard and alerts will look at the metrics (p99/mean) only for the one domain due to maxSeries. So, in order to solve this problem, I introduced emitting both domainTagged metrics and operation metrics separately with different metics name. 

**Why?**

**How did you test it?**
Haven't tested it yet.

**Potential risks**
No risk.


common/metrics/defs.go
```diff
@@ -1723,6 +1723,19 @@ const (
...
+	TaskLatencyAllDomains
+	TaskFailuresAllDomains
+	TaskDiscardedAllDomains
+	TaskAttemptTimerAllDomains
+	TaskStandbyRetryCounterAllDomains
+	TaskNotActiveCounterAllDomains
+	TaskLimitExceededCounterAllDomains
+	TaskProcessingLatencyAllDomains
+	TaskQueueLatencyAllDomains
+	TransferTaskMissingEventCounterAllDomains
```
yycptt, line 1737: Can we follow convention in matching metrics names and rename them to `XXXPerDomain`?

service/history/task/task.go
```diff
@@ -67,18 +67,19 @@ type (
...
-		maxRetryCount dynamicconfig.IntPropertyFn
+		shard             shard.Context
+		state             ctask.State
+		priority          int
+		attempt           int
+		timeSource        clock.TimeSource
+		submitTime        time.Time
+		logger            log.Logger
+		scopeIdx          int
+		emitDomainTag     bool
```
yycptt, line 78: Let's remove this flag and it's corresponding dynamic config given that it won't affect performance.

PR #3470, opened 2020-08-21
Cap activity ScheduleToStart timeout when large retry expiration time is specified

yycptt:
**What changed?**
* Allow user to specify if activity ScheduleToStart timeout is retryable or not
* If it's retryable, then cap the maximum ScheduleToStart timeout to be max(origin timeout, min(10mins, retry expiration time))

**Why?**
* To discover ScheduleToStart timeout sooner when both retry policy and host specific tasklist is used (https://github.com/uber/cadence/issues/1346)
* If activity task somehow got lost and a large expiration time is specified, cap the timeout to 10mins can ensure a new task got dispatched after 10mins and won't block workflow for too long.

**How did you test it?**
Unit test, integration test.

**Potential risks**
In the worst case, the activity retry may not work correctly when schedule to start timeout happens.


service/history/decisionChecker.go
```diff
@@ -66,6 +66,8 @@ type (
 
 const (
 	reservedTaskListPrefix = "/__cadence_sys/"
+
+	maximumScheduleToStartTimeoutForRetryInSeconds = 600 // 10 minutes
```
meiliang86, line 70: 30min?

yycptt: Updated to 30mins

PR #3478, opened 2020-08-25
Reduce log verbosity and add message to logs that didn't have one

emrahs:
**What changed?**
Changed Info logs with high volume and low value to Debug level and added a message to logs that didn't have a message. 

**Why?**
Better use of logging pipeline and reducing the noise, and making lifecycle event logs more searchable. 

**How did you test it?**
Existing unit and integration tests should be sufficient

**Potential risks**
None

### DIFF
--- a/service/history/queue/timer_queue_processor_base.go
+++ b/service/history/queue/timer_queue_processor_base.go
@@ -148,8 +148,8 @@ func (t *timerQueueProcessorBase) Start() {
 		return
 	}
 
-	t.logger.Info("", tag.LifeCycleStarting)
-	defer t.logger.Info("", tag.LifeCycleStarted)
+	t.logger.Info("Timer queue processor state changed", tag.LifeCycleStarting)
+	defer t.logger.Info("Timer queue processor state changed", tag.LifeCycleStarted)
 
 	t.redispatcher.Start()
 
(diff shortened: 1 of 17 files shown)


service/history/queue/timer_queue_processor_base.go
```diff
@@ -148,8 +148,8 @@ func (t *timerQueueProcessorBase) Start() {
 		return
 	}
 
-	t.logger.Info("", tag.LifeCycleStarting)
-	defer t.logger.Info("", tag.LifeCycleStarted)
+	t.logger.Info("Timer queue processor state changed", tag.LifeCycleStarting)
+	defer t.logger.Info("Timer queue processor state changed", tag.LifeCycleStarted)
```
yycptt, line 152: I think in most cases we use empty log message because the logger is already tagged with component tag.

emrahs: Exactly. I realized that the intent was to rely on the tag, but there are several types of logs with the same lifecycle event tag values although they are unrelated and it is cumbersome to narrow down the logs to a specific component. 

The message field is one of the most frequently used fields to filter logs during an investigation, and I don't think it's a good experience to have a good chunk of logs that have the same message (null) even though they are unrelated.

PR #3480, opened 2020-08-26
Persist and load multi-cursor processing queue states

yycptt:
**What changed?**
* Wire multi-cursor queue processing logic to persist & load processing queue states.
* Add feature flag for both persist and load operation
* On load, check if ack level and processing queue states are compatible with each other

**Why?**
Part of multi-cursor queue impl

**How did you test it?**
Local canary, verified processing queue state can be persisted to and load from DB.

**Potential risks**
N/A, feature flag disabled.


service/history/shard/context.go
```diff
@@ -252,6 +260,69 @@ func (s *contextImpl) UpdateTransferClusterAckLevel(cluster string, ackLevel int
...
+
+func (s *contextImpl) UpdateTransferProcessingQueueStates(cluster string, states []*history.ProcessingQueueState) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if len(states) == 0 {
+		return errors.New("Empty transfer processing queue states")
+	}
+
+	if s.transferProcessingQueueStates.StatesByCluster == nil {
```
yux0, line 303: when it will be nil?

yux0 (approved):

service/history/shard/context.go
yycptt: It should not be nil, but adding the check for safety. I can remove if you have strong opinion.

PR #3487, opened 2020-08-29
Integrate current execution check with replication resender

yux0:
**What changed?**
1. Integrate current execution check with replication resender
2. Update the current execution check to make sure current execution exists. 

**Why?**
Reduce the number of messages go into DLQ

**How did you test it?**
Unit tests and local tests

**Potential risks**
N/A

### DIFF
--- a/common/reconciliation/invariants/concreteExecutionExists.go
+++ b/common/reconciliation/invariants/concreteExecutionExists.go
@@ -58,11 +58,15 @@ func (c *concreteExecutionExists) Check(
 		}
 	}
 
-	var runIDCheckResult *common.CheckResult
-	currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
-	if runIDCheckResult != nil {
-		return *runIDCheckResult
+	if len(currentExecution.CurrentRunID) == 0 {
+		// set the current run id
+		var runIDCheckResult *common.CheckResult
+		currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
+		if runIDCheckResult != nil {
+			return *runIDCheckResult
+		}
 	}
+
 	concreteExecResp, concreteExecErr := c.pr.IsWorkflowExecutionExists(&persistence.IsWorkflowExecutionExistsRequest{
 		DomainID:   currentExecution.DomainID,
 		WorkflowID: currentExecution.WorkflowID,
@@ -77,6 +81,11 @@ func (c *concreteExecutionExists) Check(
 		}
 	}
 	if !concreteExecResp.Exists {
+		//verify if the current execution exists
+		_, checkResult := c.validateCurrentRunID(currentExecution)
+		if checkResult != nil {
+			return *checkResult
+		}
 		return common.CheckResult{
 			CheckResultType: common.CheckResultTypeCorrupted,
 			InvariantType:   c.InvariantType(),
@@ -97,13 +106,15 @@ func (c *concreteExecutionExists) Fix(
 
 	currentExecution, _ := execution.(*common.CurrentExecution)
 	var runIDCheckResult *common.CheckResult
-	// this is to set the current run ID prior to the check and fix operations
-	currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
-	if runIDCheckResult != nil {
-		return common.FixResult{
-			FixResultType: common.FixResultTypeSkipped,
-			CheckResult:   *runIDCheckResult,
-			InvariantType: c.InvariantType(),
+	if len(currentExecution.CurrentRunID) == 0 {
+		// this is to set the current run ID prior to the check and fix operations
+		currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
+		if runIDCheckResult != nil {
+			return common.FixResult{
+				FixResultType: common.FixResultTypeSkipped,
+				CheckResult:   *runIDCheckResult,
+				InvariantType: c.InvariantType(),
+			}
 		}
 	}
 	fixResult, checkResult := checkBeforeFix(c, currentExecution)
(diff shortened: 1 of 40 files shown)


common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -59,6 +60,11 @@ func (c *concreteExecutionExists) Check(execution interface{}) common.CheckResul
 			InvariantType:   c.InvariantType(),
 		}
 	}
```
yycptt, line 62: Not related to your change but is the check from L57-62 correct? Even if the state for current execution is closed, as long as corresponding concrete execution doesn't exist, it's still a corruption I think?

common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -59,6 +60,11 @@ func (c *concreteExecutionExists) Check(execution interface{}) common.CheckResul
...
+	var runIDCheckResult *common.CheckResult
+	currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
+	if runIDCheckResult != nil {
+		return *runIDCheckResult
+	}
 	// by this point the corresponding concrete execution must exist and can be already closed
```
yycptt, line 68: Can we remove this comment? I think it's kind of misleading. If we are sure that corresponding concrete execution must exist here, then we don't need the `IsWorkflowExecutionExists` call below and can return `CheckResultTypeHealthy` directly I think?

common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -59,6 +60,11 @@ func (c *concreteExecutionExists) Check(execution interface{}) common.CheckResul
...
 	currentExecResp, currentExecErr := c.pr.IsWorkflowExecutionExists(&persistence.IsWorkflowExecutionExistsRequest{
```
yycptt, line 69: Also not related to your change, but I think there's a race condition here. The workflow retention period can kick in before we make this call, so even if the response says the concrete execution doesn't exist, we can't be 100% sure it's a corruption, we need to check if the current execution with the same runID is still there.

another nit: can we rename the var from `currentExecResp, currentExecErr` to `concreteExceResp, concreteExecErr`?

common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -89,11 +95,20 @@ func (c *concreteExecutionExists) Check(execution interface{}) common.CheckResul
 }
 
 func (c *concreteExecutionExists) Fix(execution interface{}) common.FixResult {
-	fixResult, checkResult := checkBeforeFix(c, execution)
+	currentExecution, _ := execution.(*common.CurrentExecution)
+	var runIDCheckResult *common.CheckResult
+	currentExecution, runIDCheckResult = c.validateCurrentRunID(currentExecution)
```
yycptt, line 100: Is this needed? The deletion will be a no-op when `runIDCheckResult != nil` I think?

common/reconciliation/invariants/concreteExecutionExists.go
yux0: Yes. You are correct

common/reconciliation/invariants/concreteExecutionExists.go
yux0: Removed

common/reconciliation/invariants/concreteExecutionExists.go
yux0: Yes. You are right. The deletion of the current execution will check the current run ID.

I will update the namings.

common/reconciliation/invariants/concreteExecutionExists.go
yux0: Yes. For scanner, the current run ID is set in the execution. For resender, it passes in the domain and workflow ID. This is to set the current run ID in this case.

common/reconciliation/invariants/concreteExecutionExists.go
mkolodezny: the retention kicks in after the workflow gets closed only but there is a check for the open state above.

yycptt: Synced offline. We will also cover the case where state is closed.

common/reconciliation/invariants/concreteExecutionExists.go
yux0: Updated the code after offline sync

common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -115,3 +142,44 @@ func (c *concreteExecutionExists) Fix(execution interface{}) common.FixResult {
...
+		}
+	}
+
+	if len(currentExecution.CurrentRunID) == 0 {
+		currentExecution.CurrentRunID = resp.RunID
+	}
+
+	if currentExecution.CurrentRunID != resp.RunID {
+		return nil, &common.CheckResult{
+			CheckResultType: common.CheckResultTypeFailed,
```
yycptt, line 179: actual I think we can return healthy here? Seems like it's the same thing as current record with this runID doesn't exist.

yux0: Yes. Updated

common/reconciliation/invariants/concreteExecutionExists.go
```diff
@@ -53,27 +57,35 @@ func (c *concreteExecutionExists) Check(execution interface{}) common.CheckResul
...
-			InfoDetails:     currentExecErr.Error(),
+			InfoDetails:     concreteExecErr.Error(),
 		}
 	}
-	if !currentExecResp.Exists {
+	if !concreteExecResp.Exists {
+		//verify if the current execution exists
+		_, checkResult := c.validateCurrentRunID(currentExecution)
+		if checkResult != nil {
+			return *checkResult
```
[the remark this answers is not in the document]
yux0: Updated in line 179

PR #3489, opened 2020-08-31
Do not extend activity expiration time

yycptt:
**What changed?**
Do not extend activity expiration time to be as long as activity schedule to close timeout.

**How did you test it?**
Integration test. Tested locally.

**Potential risks**
Customer relies on extending expiration time will be break.

### DIFF
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -2220,14 +2220,14 @@ func (e *mutableStateBuilder) ReplicateActivityTaskScheduledEvent(
 		TaskList:                 attributes.TaskList.GetName(),
 		HasRetryPolicy:           attributes.RetryPolicy != nil,
 	}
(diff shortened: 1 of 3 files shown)


service/history/execution/mutable_state_builder.go
```diff
@@ -2220,14 +2220,14 @@ func (e *mutableStateBuilder) ReplicateActivityTaskScheduledEvent(
 		TaskList:                 attributes.TaskList.GetName(),
 		HasRetryPolicy:           attributes.RetryPolicy != nil,
 	}
-	ai.ExpirationTime = ai.ScheduledTime.Add(time.Duration(scheduleToCloseTimeout) * time.Second)
+
```
anish531213, line 2223: Where is this activity info used? I couldn't find in the code where this is used.

yycptt: Check `RetryActivity()` method in this file

anish531213: Oh, I see. Thanks.

PR #3492, opened 2020-09-01
Changes to support SQL plugins through vendoring

anish531213:
**What changed?**
Made some persistence constructors public and changed visibility store to take database object.

**Why?**
These changes are made in order to support SQL plugin implemented outside of this repo.

**How did you test it?**
Need to test this locally.

**Potential risks**
No risk.

### DIFF
--- a/common/persistence/sql/factory.go
+++ b/common/persistence/sql/factory.go
@@ -108,7 +108,11 @@ func (f *Factory) NewExecutionStore(shardID int) (p.ExecutionStore, error) {
 
 // NewVisibilityStore returns a visibility store
 func (f *Factory) NewVisibilityStore() (p.VisibilityStore, error) {
-	return NewSQLVisibilityStore(f.cfg, f.logger)
+	conn, err := f.dbConn.get()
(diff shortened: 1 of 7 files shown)


longquanzheng:
>These changes are made in order to support SQL plugin implemented outside of this repo.

I thought it's already supported, isn't it correct?
like https://github.com/longquanzheng/cadence-extensions/tree/master/cadence-sqlite
I think I must miss something.

anish531213:
> > These changes are made in order to support SQL plugin implemented outside of this repo.
> 
> I thought it's already supported, isn't it correct?
> like https://github.com/longquanzheng/cadence-extensions/tree/master/cadence-sqlite
> I think I must miss something.

Oh, I see. I think we won't need this change as well. @andrewjdawson2016

andrewjdawson2016 (approved):

andrewjdawson2016:
Looks to me like Long is right this should be supported. Lets abandon this diff. If we find that for some reason we cannot use the existing method to Register our plugin we can revisit this.

PR #3494, opened 2020-09-02
Start queue processor before failover callback registration

yycptt:
**What changed?**
Start queue processor before failover callback registration

**Why?**
Failover callback will try to create a failover queue processor for domains that need to be failover. On host start, the process can happen before tx or timer queue is started. 

In new queue processor logic, queue readlevel(which will be used as max level for failover queue) can't be retrieved before queue processor is started, and failover callback registration will be blocked. 

When tx or timer queue is not started, we don't need failover processor at all because we don't need to scan any tasks. Also we can start the queue processor before failover callback is registered to avoid the deadlock.

**How did you test it?**
Tested in staging2

**Potential risks**
Tasks get dropped when domain tries to failover while there are shard movements.

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -387,10 +387,10 @@ func (e *historyEngineImpl) Start() {
 	e.logger.Info("History engine state changed", tag.LifeCycleStarting)
 	defer e.logger.Info("History engine state changed", tag.LifeCycleStarted)
 
-	e.registerDomainFailoverCallback()
-
 	e.txProcessor.Start()
 	e.timerProcessor.Start()
+
+	e.registerDomainFailoverCallback()
 
 	clusterMetadata := e.shard.GetClusterMetadata()
 	if e.replicatorProcessor != nil && clusterMetadata.GetReplicationConsumerConfig().Type != sconfig.ReplicationConsumerTypeRPC {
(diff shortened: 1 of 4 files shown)


service/history/historyEngine.go
```diff
@@ -387,11 +387,11 @@ func (e *historyEngineImpl) Start() {
 	e.logger.Info("History engine state changed", tag.LifeCycleStarting)
 	defer e.logger.Info("History engine state changed", tag.LifeCycleStarted)
 
-	e.registerDomainFailoverCallback()
-
 	e.txProcessor.Start()
 	e.timerProcessor.Start()
 
+	e.registerDomainFailoverCallback()
```
yux0, line 393: Add some comment that the callback has to be registered after the task processor starts

service/history/queue/timer_queue_processor.go
```diff
@@ -241,6 +241,10 @@ func (t *timerQueueProcessor) NotifyNewTask(
 func (t *timerQueueProcessor) FailoverDomain(
 	domainIDs map[string]struct{},
 ) {
+	if atomic.LoadInt32(&t.status) != common.DaemonStatusStarted {
```
yux0, line 244: Same. Have a comment here

yux0 (approved):

service/history/historyEngine.go
yycptt: Done.

service/history/queue/timer_queue_processor.go
yycptt: Done.

PR #3495, opened 2020-09-02
Improve developer contribution guide for Postgres development

longquanzheng:
**What**
1. Improve contribution guide to help setup environment of running test
2. Fix a line that skip running tests for SQL

**Why**
1. To help other developer 
2. To run the test for SQL 

**How did you test it?**
1. Run unit test 

**Potential risks**
No risk

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,11 +2,12 @@
 
 This doc is intended for contributors to `cadence` server (hopefully that's you!)
 
-**Note:** All contributors also need to fill out the [Uber Contributor License Agreement](http://t.uber.com/cla) before we can merge in any of your changes
+>Note: All contributors also need to fill out the [Uber Contributor License Agreement](http://t.uber.com/cla) before we can merge in any of your changes
 
 ## Development Environment
 
-* Go. Install on OS X with `brew install go`.
+* Golang. Install on OS X with `brew install go`. 
+>Note: If running into any compiling issue, make sure you are on a too old version of Golang. 
 
 ## Checking out the code
 
@@ -58,6 +59,9 @@ make bins
 ```
 
 ## Testing
+
+>Note: The default setup of Cadence depends on Cassandra and Kafka(is being deprecated). 
+This section assumes you are testing with them, too. Please refer to [persistence documentation](https://github.com/uber/cadence/blob/master/docs/persistence.md) if you want to test with others like MySQL/Postgres. 
 
 Before running the tests you must have `cassandra`, `kafka`, and its `zookeeper` dependency:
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,11 +2,12 @@
...
 This doc is intended for contributors to `cadence` server (hopefully that's you!)
 
-**Note:** All contributors also need to fill out the [Uber Contributor License Agreement](http://t.uber.com/cla) before we can merge in any of your changes
+>Note: All contributors also need to fill out the [Uber Contributor License Agreement](http://t.uber.com/cla) before we can merge in any of your changes
 
 ## Development Environment
 
-* Go. Install on OS X with `brew install go`.
+* Golang. Install on OS X with `brew install go`. 
+>Note: If running into any compiling issue, make sure you are on a too old version of Golang.
```
yux0, line 10: Do you mean to use the latest version of Golang?

longquanzheng: No my intention is to let the developer to upgrade to newer versions if they see compiling issues. The background is when I run the command, because I already installed go so it doesn't do anything, But I still got error because the version on may laptop is 1.11.

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -1671,7 +1671,7 @@ func (s *ExecutionManagerSuite) TestDeleteWorkflow() {
 // TestDeleteCurrentWorkflow test
 func (s *ExecutionManagerSuite) TestDeleteCurrentWorkflow() {
 	if s.ExecutionManager.GetName() != "cassandra" {
-		s.T().Skip("this test is only applicable for cassandra (uses TTL based deletes)")
+		return
```
meiliang86, line 1674: I think skip is the correct behavior here. Does it cause any issue for you?

CONTRIBUTING.md
meiliang86: "make sure you are on a too old version of Golang" -> "make sure you upgrade to the latest stable version of Golang"?

longquanzheng: I was hesitating not to say "latest". Yeah I am fine with that too, as they have encountered compiling issues, getting the latest makes sense.

common/persistence/persistence-tests/executionManagerTest.go
longquanzheng: Yes it causes issue that the whole TestSQLExecutionManagerSuite will skip, when I run postgres test locally.
Screen Shot 2020-09-10 at 2 43 51 PM

Anyway it's already committed in a previous PR when I fixed the Postgres upsert issue.

CONTRIBUTING.md
longquanzheng: Done :D

meiliang86 (approved):

meiliang86:
@longquanzheng Thanks for your contribution. @yux0 We should take a second pass with these setup steps once we deprecate kafka, and remove the kafka related stuffs there.

Issue #3482, opened 2020-08-28
Postgres plugin has a faulty SQL statement


chris-rcn:
**Describe the bug**
Some data does not get persisted when it should.

**To Reproduce**
Is the issue reproducible? 
- Yes

**Steps to reproduce the behavior:**
1. Poll for an activity.
2. Bounce the history node.
3. Attempt to complete the activity.  Receive "activity not found".

**Expected behavior**
The activity completes.

**Screenshots**
None

**Additional context**
Here is the fix:

	diff --git a/common/persistence/sql/sqlplugin/postgres/execution_maps.go b/common/persistence/sql/sqlplugin/postgres/execution_maps.go
	index dc518896..d0044bbf 100644
	--- a/common/persistence/sql/sqlplugin/postgres/execution_maps.go
	+++ b/common/persistence/sql/sqlplugin/postgres/execution_maps.go
	@@ -45,15 +45,12 @@ run_id = $4`
	        setKeyInMapQueryTemplate = `INSERT INTO %[1]v
	 (shard_id, domain_id, workflow_id, run_id, %[4]v, %[2]v)
	 VALUES
	 (:shard_id, :domain_id, :workflow_id, :run_id, :%[4]v, %[3]v)
	 ON CONFLICT (shard_id, domain_id, workflow_id, run_id, %[4]v) DO UPDATE
	-  SET shard_id = excluded.shard_id,
	-      domain_id = excluded.domain_id,
	-      workflow_id = excluded.workflow_id,
	-         run_id = excluded.run_id,
	-      %[4]v = excluded.%[4]v `
	+  SET (shard_id, domain_id, workflow_id, run_id, %[4]v, %[2]v)
	+  = (:shard_id, :domain_id, :workflow_id, :run_id, :%[4]v, %[3]v)`

longquanzheng:
Hi @chris-rcn , how is the the Cadence running with Postgres for you? would you mind share some experience? There are folks in the community that are very interested in Postgres too. :D

chris-rcn:
Working well.  I'm not aware of any other issues.

longquanzheng:
Closed as the issue has be resolved.

PR #3498, opened 2020-09-04
[SQL]Fix upsert SQL template for Postgres plugin

longquanzheng:
**What changed?**
Fixes issue #3482 and replace https://github.com/uber/cadence/pull/3484

**Why?**

**How did you test it?**
Integration test with local Postgres.
New test is added

**Potential risks**
Replace on conflict is already broken, so no more risks.


common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -1671,7 +1671,8 @@ func (s *ExecutionManagerSuite) TestDeleteWorkflow() {
 // TestDeleteCurrentWorkflow test
 func (s *ExecutionManagerSuite) TestDeleteCurrentWorkflow() {
 	if s.ExecutionManager.GetName() != "cassandra" {
-		s.T().Skip("this test is only applicable for cassandra (uses TTL based deletes)")
+		return
+		//s.T().Skip("this test is only applicable for cassandra (uses TTL based deletes)")
```
longquanzheng, line 1675: Skip here is wrong. this will skip more tests than it thought.

chris-rcn (approved):
This is working in my real-world usage.

common/persistence/persistence-tests/executionManagerTest.go
vancexu: maybe rephrase to comment to be just `"this test is only applicable for cassandra (uses TTL based deletes)"`

vancexu (commented):
Do you plan to add more unit test before merge it?

longquanzheng:
> Do you plan to add more unit test before merge it?

Yep, a new test is added.

vancexu (approved):
Thank you Long!

PR #3500, opened 2020-09-07
Fix NDC resetter persistence bugs

longquanzheng:
**What changed?**
1. Resetter call `PersistNonFirstWorkflowEvents` instead of `PersistFirstWorkflowEvents`
2. SqlExecutionManager allows CreateWorkflowModeContinueAsNew in `createWorkflowExecutionTx`

**Why?**
Without the change reset won't work in SQLin NDC path

**How did you test it?**
TODO

**Potential risks**
No risk as they are bug fix.

### DIFF
--- a/service/history/reset/resetter.go
+++ b/service/history/reset/resetter.go
@@ -285,7 +285,7 @@ func (r *workflowResetterImpl) persistToDB(
 	if err != nil {
 		return err
 	}
-	resetHistorySize, err := resetWorkflow.GetContext().PersistFirstWorkflowEvents(resetWorkflowEventsSeq[0])
+	resetHistorySize, err := resetWorkflow.GetContext().PersistNonFirstWorkflowEvents(resetWorkflowEventsSeq[0])
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


yux0 (approved):

service/history/reset/resetter.go
```diff
@@ -285,7 +285,7 @@ func (r *workflowResetterImpl) persistToDB(
 	if err != nil {
 		return err
 	}
-	resetHistorySize, err := resetWorkflow.GetContext().PersistFirstWorkflowEvents(resetWorkflowEventsSeq[0])
+	resetHistorySize, err := resetWorkflow.GetContext().PersistNonFirstWorkflowEvents(resetWorkflowEventsSeq[0])
```
yux0, line 288: I am not following here. Why do we want to use persist non first workflow events for the reset sequence[0]?

longquanzheng: If you are asking why PersistNonFirstWorkflowEvents instead of PersistFirstWorkflowEvents -- because PersistFirstWorkflowEvents will try to insert another record for "history_tree", which is not supposed to do. Because "ForkBranch" already done that. Redoing it here will cause potential issues. It works with Cassandra because it's using upsert. Theoretically we should use insert, but inset is too expensive for Cassandra.

If you are asking why resetWorkflowEventsSeq[0] not resetWorkflowEventsSeq[1 or others], I think that's some details for the workflow interface. We should do an assertion here to make sure len(resetWorkflowEventsSeq) == 1.

longquanzheng: Added an assertion

Issue #2694, opened 2019-10-20
Server exited with error when trying to create domain during Cassandra initialization


mingfang:
I’m using the command until cadence --domain $CADENCE_CLI_DOMAIN domain describe || cadence --domain $CADENCE_CLI_DOMAIN domain register; as a test for when cadence has started and ready. It looks like this command actually halts cadence if it is in the middle of initializing cassandra. This is error
```
2019/10/20 02:37:03 Schema updated from 0.21 to 0.22                                              
2019/10/20 02:37:03 UpdateSchemeTask done                                                         2019/10/20 02:37:03 gocql: unable to dial control conn 172.27.215.236: dial tcp 172.27.215.236:9042: connect: connection refused                                                                   
2019/10/20 02:37:03 error creating Keyspace:gocql: unable to create session: control: unable to connect to initial hosts: dial tcp 172.27.215.236:9042: connect: connection refused                
2019/10/20 02:37:03 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:03 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:03 no connections were made when creating the session                            
2019/10/20 02:37:03 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:04 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:04 no connections were made when creating the session                            
2019/10/20 02:37:04 Loading config; env=docker,zone=,configDir=/etc/cadence/config                
2019/10/20 02:37:04 Loading configFiles=[/etc/cadence/config/docker.yaml]                         
2019/10/20 02:37:04 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:04 error: failed to connect to 250.2.162.9:9042 due to error: Keyspace 'cadence_ visibility' does not exist                                                                        
2019/10/20 02:37:04 Incompatible versionsunable to create CQL Client: no connections were made when creating the session                                                                           
2019/10/20 02:37:04 Command exited with error: exit status 1 
```

The problem is in this line from my Terraform automation
[...]

PR #3505, opened 2020-09-10
k8s: fix cassandra-tool env key conflict with k8s

pip1998:
**What changed?**
1.here I just raplace the CASSANDRA_PORT to CASSANDRA_DB_PORT, to avoid
use the same key with k8s. Or is there any key more fit?

**Why?**
1.the default CASSANDRA_PORT key is used by k8s when you have an
service pod named cassandra(k8s will gen a env key which is name_port).
And the CASSANDRA_PORT has valid value which is an url, cannot be i
parsed to an int value. Finally, the cadence-cassandra-tool can not be
executed. Candence cannot work well when need cadence-canssandra.

### DIFF
--- a/cmd/tools/cassandra/main.go
+++ b/cmd/tools/cassandra/main.go
@@ -27,5 +27,8 @@ import (
 )
 
 func main() {
-	cassandra.RunTool(os.Args) //nolint:errcheck
+	err := cassandra.RunTool(os.Args)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pip1998:
fix bugs like #2694

vancexu (approved):
Thank you for the contribution!

yycptt (approved):

PR #3508, opened 2020-09-10
Multicursor Queue Processor Bug Fix

yycptt:
**What changed?**
* Fix bug on merging processing queue and processing queue collection
* Prevent ack level to move beyond read level

**Why?**
Bug fix.

**How did you test it?**
Tested in staging2

**Potential risks**
Feature flag disabled


service/history/queue/processing_queue.go
```diff
@@ -120,6 +120,12 @@ func newProcessingQueue(
 		queue.state = copyQueueState(state)
 	}
 
+	if queue.state.readLevel.Less(queue.state.ackLevel) {
+		logger.Fatal("ack level larger than readlevel when creating processing queue", tag.Error(
```
vancexu, line 124: What is the intention of Fatal here? Will recreate queue auto fix it?

service/history/queue/processing_queue.go
```diff
@@ -224,15 +230,31 @@ func (q *processingQueueImpl) Merge(
 		q1.state.domainFilter.Merge(q2.state.domainFilter),
 	))
 
+	for _, state := range newQueueStates {
+		if state.ReadLevel().Less(state.AckLevel()) || state.MaxLevel().Less(state.ReadLevel()) {
+			q.logger.Fatal("invalid processing queue merge result", tag.Error(
```
vancexu, line 235: I just wonder if queue state is bad, what would be the action to mitigate? 
Asking because I think we should be cautious on using Fatal because of the potential damage. We are fine to use fatal/panic If we don't expect it to happen (whenever happen it is code bug and definitely can not recover); or we know panic is the best way to limit the impact. 
Otherwise, we may prefer to error and then compensate.

yycptt: For all those sanity check, if it fails, it must be a code bug in the processing logic and I don't see an obvious way to recover. If we try to proceed with the wrong state, the consequence can be more severe. For example ack level may move beyond the point it's supposed to be and we delete tasks that haven't been processed from DB. By panic immediately, we can at least prevent that from happening.

vancexu (approved):

PR #3510, opened 2020-09-10
Add multicursor processing queue related metrics

yycptt:
**What changed?**
* Add multicursor processing queue related metrics

**Why?**
* To get better visibility into how processing queue is doing

**How did you test it?**
Tested in staging2.

**Potential risks**
Feature flag disabled

### DIFF
--- a/service/history/queue/processor_base.go
+++ b/service/history/queue/processor_base.go
@@ -308,7 +308,7 @@ func (p *processorBase) splitProcessingQueueCollection(
 
 func (p *processorBase) emitProcessingQueueMetrics() {
 	numProcessingQueues := 0
-	maxProcessingQueueLevel := -1
+	maxProcessingQueueLevel := 0
 	for _, queueCollection := range p.processingQueueCollections {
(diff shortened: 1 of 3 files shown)


service/history/queue/processor_base.go
```diff
@@ -304,6 +306,20 @@ func (p *processorBase) splitProcessingQueueCollection(
 	}
 }
 
+func (p *processorBase) emitProcessingQueueMetrics() {
+	numProcessingQueues := 0
+	maxProcessingQueueLevel := -1
```
vancexu, line 311: Maybe just init maxProcessingQueueLevel to 0? will scope.recordTimer(-1) causing issue?

yycptt: That's a good point. I will update. 

Theoretically, it can never happen as `processingQueueCollections` will never be empty; otherwise, it must be a code bug and we should panic or alert on that.

yycptt: Updated.

vancexu (approved):

PR #3512, opened 2020-09-11
Define domains protos

andrewjdawson2016:


proto/persistenceblobs/v1/message.proto
```diff
@@ -0,0 +1,64 @@
+syntax = "proto3";
```
vancexu, line 1: what is the v1 here?
Should we keep using `idls` folder for these defs?

andrewjdawson2016: The v1 is based on what temporal did. 

Yes eventually this should be in the idls directory. The end state will be as follows 
- We have an external repo that keeps track of all the idls (both in thrift and proto) used by frontend
- This repo gets added as a submodule in the current repo (under the idls directory)  
- In addition to adding this repo as a submodule we also define additional protos which are used only internally within cadence (e.g. history service protos, persistence, archival protos, matching protos etc...) 
- The internal protos will have a dependency on these submodule protos. 

While this is the desired end state we don't need to get to this state to unblock progress on persistence proto migration. Once we introduce proto for our frontend to use as well then we can move towards this end state.

vancexu: I see, thanks. I am ok with the diff, looping in others for visibility @meiliang86 @emrahs @vytautas-karpavicius

vancexu (approved):
LGTM

PR #3513, opened 2020-09-14
Split reconciliation/common

mantas-sidlauskas:
**What changed?**
* Package reconciliation/common is now split to multiple packages.
* Scan type is not passed as an argument to iterators/invariant manager
* Blobstore iterator can deserialise to interface, no deserialisation helpers needed

**Why?**
These changes allows to reduce scope for changes when adding a new type of scanner. 

**How did you test it?**
Unit test

**Potential risks**


andrewjdawson2016 (approved):
This refactor looks good to me.

PR #3515, opened 2020-09-14
Use a limited context timeout to get current exec lock

yux0:
**What changed?**
Use a limited context timeout to get current exec lock

**Why?**
If a customer floods the start workflow API with the same wf id, return a service busy instead of timing out the request.

**How did you test it?**
Local test

**Potential risks**
N/A


service/history/historyEngine.go
```diff
@@ -151,6 +151,8 @@ var (
 	ErrConsistentQueryNotEnabled = &workflow.BadRequestError{Message: "cluster or domain does not enable strongly consistent query but strongly consistent query was requested"}
 	// ErrConsistentQueryBufferExceeded is error indicating that too many consistent queries have been buffered and until buffered queries are finished new consistent queries cannot be buffered
 	ErrConsistentQueryBufferExceeded = &workflow.InternalServiceError{Message: "consistent query buffer is full, cannot accept new consistent queries"}
+	// ErrConcurrentStartRequest is error indicating there is an outstanding start workflow request. The incoming request fails to acquires the lock before the outstanding request finishes.
+	ErrConcurrentStartRequest = &workflow.ServiceBusyError{Message: "an outstanding start workflow request is in-progress. Failed to acquire the resource."}
```
yycptt, line 155: The error message is not very accurate I think. It's not necessarily that there's another outstanding start workflow request when the timeout happens. The timeout can also happen if customer starts a workflow and schedules a large number of activities, then the lock will be held by all the RecordActivityStarted API calls, results in timeout for the StartWorkflowAPI call.

service/history/historyEngine.go
```diff
@@ -645,14 +647,26 @@ func (e *historyEngineImpl) startWorkflowHelper(
...
 	workflowID := request.GetWorkflowId()
 	domainID := domainEntry.GetInfo().ID
+
 	// grab the current context as a lock, nothing more
+	// use a smaller context timeout to get the lock
+	getLockContext := ctx
+	var cancel context.CancelFunc
+	if deadline, ok := ctx.Deadline(); ok {
+		now := e.shard.GetTimeSource().Now()
+		ctxTimeout := now.Sub(deadline) / 2
```
yycptt, line 657: nit: will / 2 be too much? Probably leave around 1s left should be enough? 

May be create/reuse a util function similar to: https://github.com/uber/cadence/blob/master/service/matching/taskListManager.go#L508?

service/history/historyEngine.go
```diff
@@ -645,14 +647,26 @@ func (e *historyEngineImpl) startWorkflowHelper(
...
+	}
 	_, currentRelease, err := e.executionCache.GetOrCreateCurrentWorkflowExecution(
-		ctx,
+		getLockContext,
 		domainID,
 		workflowID,
 	)
 	if err != nil {
-		return nil, err
+		return nil, ErrConcurrentStartRequest
```
yycptt, line 669: I think I lack the context. So we decided to ignore the situation where the timeout is actually due to some DB issue instead of too many requests?

yycptt: Synced offline, we should enforce context timeout in DB layer so that the lock will be released on the context timedout.

yycptt (approved):

service/history/historyEngine.go
andrewjdawson2016: +1

andrewjdawson2016: Seems to me its safe to have a very short time out here. The worst case is we return a reasonable error to the user that the workflow is currently locked and they retry. I am thinking like 100ms is fine?

service/history/historyEngine.go
andrewjdawson2016: I think I am confused as well here. It looks like to me we try to get the in memory lock first and only if we successfully do that do we go on to update the database. So why do we have to enforce this context timeout at the db level?

andrewjdawson2016: Also the error could be non-nil here because the execution cache is full technically. I think we should check that the error is of type context timeout here.

service/history/historyEngine.go
yux0: Both calculate from the input context or use a fix timeout work here. The only concern with a fix timeout is, the fix timeout could be larger than the input context timeout. (If a customer define a context timeout as 10ms but we use 100ms for this context)

service/history/historyEngine.go
yux0: So the release lock will be released after this function returns. Currently, the DB operations do not take in the context timeout. So in the worst case, it will hold this lock for 10 seconds. Within these 10 seconds, if the customers call this API with the same workflow ID, those requests could fail as a service busy error which will impact the availability metrics. So in the long run, we need the DB operations to take in context timeout

service/history/historyEngine.go
meiliang86: Did I read the code correctly? Shouldn't it be deadline.Sub(now)?

service/history/historyEngine.go
```diff
@@ -645,13 +647,29 @@ func (e *historyEngineImpl) startWorkflowHelper(
...
+			defer cancel()
+		}
+	}
 	_, currentRelease, err := e.executionCache.GetOrCreateCurrentWorkflowExecution(
-		ctx,
+		getLockContext,
 		domainID,
 		workflowID,
 	)
+	getLockContext.Err()
```
meiliang86, line 668: What is this used for?

meiliang86:
Can you add a unit test for this?

service/history/historyEngine.go
yux0: Good catch. Added a test for it

service/history/historyEngine.go
yux0: Removed

PR #3524, opened 2020-09-17
Add context parameter to persistence interfaces

yycptt:
**What changed?**
Add context parameter to persistence interfaces.
 - For all implementations of persistence interface, `_ context.Context` is used as a placeholder in the method definition.
 - For `shardManager` and `TaskManger`, since `shardStore` and `TaskStore` are just aliases for those two data interface, context parameter has also been added to those to interfaces.
 - For all related methods calls that need a context parameter, `context.TODO()` is used as a placeholder.

**Why?**
First step for https://github.com/uber/cadence/issues/3519

**How did you test it?**
Local test

**Potential risks**
This diff only add the parameter to the interface, they are not used anywhere.


meiliang86 (approved):

common/persistence/cassandra/cassandraHelpers.go
```diff
@@ -26,11 +26,10 @@ import (
 	"os"
 	"strings"
 
-	"github.com/uber/cadence/common/auth"
-
 	"github.com/gocql/gocql"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/uber/cadence/common/auth"
```
longquanzheng, line 32: Curious is there any IDE plugin sorting imports automatically?

common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -350,7 +354,7 @@ func (h *cassandraHistoryV2Persistence) DeleteHistoryBranch(
 		BeginNodeID: common.Int64Ptr(beginNodeID),
 	})
 
-	rsp, err := h.GetHistoryTree(&p.GetHistoryTreeRequest{
+	rsp, err := h.GetHistoryTree(context.TODO(), &p.GetHistoryTreeRequest{
```
longquanzheng, line 357: why this one uses TODO, not the pass-in parameter?

longquanzheng:
LGTM as the first part. I just checked that gocql and sqlx both support context but for some reasons we never used them. Now adding context means we have to add this into the persistence plugin interfaces as well.

andrewjdawson2016 (approved):
Looks good to me

andrewjdawson2016:
@longquanzheng Yes I will take care of the follow up diff of plumbing this through to the plugin interface and the MySQL + Postgres implementations.

common/persistence/cassandra/cassandraHelpers.go
yycptt: I think current if there's no empty line between those imports, `make fmt` will automatically group them in a standard way.

common/persistence/cassandra/cassandraHistoryPersistence.go
yycptt: Yes it should use the pass-in parameter and I will change it to use the pass-in parameter in a later diff. For now, given that there're a large number of methods in persistence interface, I'd like to keep all methods implementing the persistence interface to use `_ context.Context` in the method definition, so that I can easily find all those methods later and won't miss any one.

common/persistence/elasticsearch/esVisibilityStore_test.go
```diff
@@ -140,7 +141,7 @@ func (s *ESVisibilitySuite) TestRecordWorkflowExecutionStarted() {
 		s.Equal(string(common.EncodingTypeThriftRW), fields[es.Encoding].GetStringData())
 		return true
 	})).Return(nil).Once()
-	err := s.visibilityStore.RecordWorkflowExecutionStarted(request)
+	err := s.visibilityStore.RecordWorkflowExecutionStarted(context.Background(), request)
```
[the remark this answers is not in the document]
yycptt: I am not entirely sure for elastic search, seems like we can't enforce timeout when publishing message to kafka, so mostly likely ctx parameter will be ignored in es implementation.

If the concern is for other implementations I think the query is still subject to connection timeout in addition to the context timeout passed in?

But yeah, I think using a real context make sense. I will update the test.

longquanzheng:
> @longquanzheng Yes I will take care of the follow up diff of plumbing this through to the plugin interface and the MySQL + Postgres implementations.

Sounds great. Thanks to you both!
Would you mind I take care of Cassandra part as part of the NoSQL support effort ? @yycptt , so that there will be less conflict between us. We can sync up for detailed work boundary

yycptt:
> > @longquanzheng Yes I will take care of the follow up diff of plumbing this through to the plugin interface and the MySQL + Postgres implementations.
> 
> Sounds great. Thanks to you both!
> Would you mind I take care of Cassandra part as part of the NoSQL support effort ? @yycptt , so that there will be less conflict between us. We can sync up for detailed work boundary

Synced offline. @longquanzheng will take care of plumbing this through cassandra persistence implementation while doing the NoSQL support. I will make changes to the part above persistence interface layer, basically data interface and application layer code.

common/persistence/elasticsearch/esVisibilityStore_test.go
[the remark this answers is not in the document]
yycptt: Updated to use real context in tests.

yycptt:
I will land the diff next Monday (09/21) after 0.15 release is cut.

andrewjdawson2016:
Thanks @yycptt and @longquanzheng  :)

PR #3525, opened 2020-09-18
[NoSQL] Refactor Cassandra history persistence for later NoSQL support

longquanzheng:
**What changed?**
Separate code in CassandraHistoryPersistence with interface 

**Why?**
To support https://github.com/uber/cadence/issues/3514
Proposal already synced with @meiliang86 and @andrewjdawson2016 

**How did you test it?**
Existing tests should do it.

**Potential risks**
Mostly no behavior change, just refactoring code.
But there is only potential behavior change. Previously when reading branch forkTime, we directly use time.Time. Here do an extra step: read by time.Time->convert to unixTimestamp->convert back to time.Time. It should be minimal because there is no critical usage of forkTime, and logically there shouldn't be a behavior change. See more details in the code comments.


common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -22,46 +22,29 @@ package cassandra
 
 import (
 	"fmt"
-	"sort"
```
longquanzheng, line 25: I will fix import later. Also this folder will be moved to "nosql" pkg later. Right now I keep in this folder so that the code review doesn't look crazy :D

common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -130,63 +113,70 @@ func (h *cassandraHistoryV2Persistence) AppendHistoryNodes(
...
-			branchInfo.TreeID, branchInfo.BranchID, ancs, cqlNowTimestamp, request.Info)
-		batch.Query(v2templateUpsertData,
-			branchInfo.TreeID, branchInfo.BranchID, request.NodeID, request.TransactionID, request.Events.Data, request.Events.Encoding)
-		err = h.session.ExecuteBatch(batch)
-	} else {
-		query := h.session.Query(v2templateUpsertData,
-			branchInfo.TreeID, branchInfo.BranchID, request.NodeID, request.TransactionID, request.Events.Data, request.Events.Encoding)
-		err = query.Exec()
 	}
+	nodeRow := &nosqlPlugin.HistoryNodeRow{
```
meiliang86, line 131: These types will be different between sql and nosql?

meiliang86 (commented):
I like this change. It makes the code more readable.

meiliang86 (approved):

common/persistence/cassandra/cassandraHistoryPersistence.go
longquanzheng: Yeah. There are some common pieces, we could move to some common pkg later if necessary. Right now they have to be different.

In SQL, we don't need `NextPageToken` in HistoryNodeRow. Because most SQL databse doesn't support paging token, so the sqlHistoryManager worked around it by using primary key.

But in Cassandra, we already done that. Ideally, we could do the same. Ideally we can do the samething, by using primary key, I guess. For the safty reason, I want to pass it here. It will add more flexibility to the database layer if they want to use paging token supported by underlying NoSQL DBs.

Also in SQL we decided later to use blob to represent the whole business logic including things like branchRanges for ancestors:
```	
HistoryTreeRow struct {
		ShardID      int
		TreeID       UUID
		BranchID     UUID
		Data         []byte
		DataEncoding string
	}
```
It turns out to be a better choice. Because using blob will help preventing non-compatibility changes to the interfaces, if we want to add some new fields.

However in Cassandra we already used specific fields just for `Ancestors` ,`CreateTime`, `Info`, etc. For backward compatibility, we have to keep passing them separately, if we don't want to waste CPU/memory for encoidng/decoding blobs. 

For new NoSQL DBs that we/community will implement later, ideally we should use blob as well. But I don't see a way to do that without avoiding compatibility issues yet. We could do that later if anyone has suggestion. 



So ideally

common/persistence/nosql/nosqlPlugin/cassandra/plugin.go
```diff
@@ -0,0 +1,28 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package cassandra
+
+const (
+	// PluginName is the name of the plugin
+	PluginName                   = "cassandra"
+)
+
```
yux0, line 27: nit: empty line

common/persistence/nosql/nosqlPlugin/cassandra/db.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yux0, line 1: nit: 2020

yux0 (approved):

andrewjdawson2016 (approved):
LGTM

common/persistence/nosql/nosqlPlugin/cassandra/plugin.go
longquanzheng: I tried to remove it but my IDE put it back. Would you mind leaving it?

common/persistence/nosql/nosqlPlugin/cassandra/db.go
longquanzheng: yeah sure. Good catch :D

longquanzheng:
@yux0 @meiliang86 @andrewjdawson2016 I have fixed the conflicts. Do you want to take another look?

PR #3534, opened 2020-09-22
Add persistence encoding and decoder interfaces

andrewjdawson2016:
This PR introduces the plumbing needed to serialize persistence structs as proto. It introduces a new encoder and decoder interface. Based on config the cluster can either use the proto encoder or the thrift encoder. The decoder will always be able to operate on both proto and thrift blobs. The proto encoder implementation will take as input a thrift strcut, convert it to a proto struct and then serialize it using proto.


common/persistence/serialization/interfaces.go
```diff
@@ -0,0 +1,66 @@
...
+		TimerInfoToBlob(*sqlblobs.TimerInfo) (p.DataBlob, error)
+		TaskInfoToBlob(*sqlblobs.TaskInfo) (p.DataBlob, error)
+		TaskListInfoToBlob(*sqlblobs.TaskListInfo) (p.DataBlob, error)
+		TransferTaskInfoToBlob(*sqlblobs.TransferTaskInfo) (p.DataBlob, error)
+		TimerTaskInfoToBlob(*sqlblobs.TimerTaskInfo) (p.DataBlob, error)
+		ReplicationTaskInfoToBlob(*sqlblobs.ReplicationTaskInfo) (p.DataBlob, error)
+	}
+
+	// Decoder is used to deserialize from persistence data blobs
+	Decoder interface {
```
longquanzheng, line 50: Nit: not a blocker. I personally think we save one interface if combing them as Encoder. As people usually understand encoder can do decoding.

longquanzheng (approved):
LGTM except a nit comment, but not a blocker.

common/persistence/serialization/interfaces.go
emrahs, line 50: Encoder & Decoder interfaces should be symmetric. While there is an encoder type per format and totally separate implementations, there is only one decoder type that requires an encoding format parameter for each decode function. We should make these interfaces symmetric. I like the factory pattern that you used for the encoder.

I'd also consider keeping encoder and decoder for a format in the same type instead of defining two types per format.

common/persistence/serialization/util.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func decodeErr(err error) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("error deserializing blob to struct: %v", err)
+}
+
+func thriftRWEncode(t thriftRWType) (p.DataBlob, error) {
```
emrahs, line 64: I'd move all thrift-specific functions to the thrift encoder/decoder type and keep this file only for common utils.

common/persistence/sql/common.go
```diff
@@ -35,8 +37,10 @@ import (
 
 // TODO: Rename all SQL Managers to Stores
 type sqlStore struct {
-	db     sqlplugin.DB
-	logger log.Logger
+	db      sqlplugin.DB
+	logger  log.Logger
+	encoder serialization.Encoder
```
emrahs, line 42: Related to my comment above: Is there any reason for keeping encoder and decoder interfaces separate? I think this separation bloats the code a little bit and leaves the door open for using incompatible encoder & decoder pairs in the future. It seems like you can just have one _parser_ that handles both encoding and decoding.

common/persistence/serialization/interfaces.go
andrewjdawson2016: Updated to do this.

andrewjdawson2016: Spoke offline and addressed comment in a slightly different way.

common/persistence/serialization/util.go
andrewjdawson2016: Done

PR #3537, opened 2020-09-22
Fix link to CLI docs

JacobCrofts:
**What changed?**

The link to the CLI docs was ever so slightly broken, so I fixed it.

**Why?**

self-evident

**How did you test it?**

n/a

**Potential risks**

no risks


yux0 (approved):
Thank you for fixing the broken link

PR #3539, opened 2020-09-22
Move sql connection interface to plugin level

andrewjdawson2016:
The connection interface being defined at the level of sql rather than plugin is confusing because it makes it seems like a plugin would have to be able to implement this interface. When in actuality as long as a plugin can implement the adminCrud interface it will work. The connection interface is simply an implementation detail of how MySQL and Postgres happen to implement the crud interface.


longquanzheng (approved):
LGTM.

yycptt (approved):

PR #3542, opened 2020-09-23
Remove unnecessary coverage calculation

longquanzheng:
**What changed?**
Remove unnecessary coverage calculation

**Why?**
https://github.com/uber/cadence/issues/3541
This will save some test running time for us. 

**How did you test it?**
We expect coverage doesn't change too much

**Potential risks**
No


yux0 (approved):

longquanzheng:
Seems like it does increase the coverage :P

Issue #3519, opened 2020-09-15
Enforce Context Timeout in Persistence Layer


yycptt:

longquanzheng:
Is there any background we are doing this? I am guessing there are some issues like goroutine leakage /memory?

yycptt:
In general, as those persistence calls are just RPC calls, we think it makes sense to set different context timeout for each call based on needs, instead of using a background context for a pre-defined timeout.

More specifically, for sync frontend API calls (like start workflow), we should release the current execution lock after the ctx for the API call exceeds deadline. While for now, we will hold the lock until DB call returns and we don't have control over when DB call will return. (For more context, check comments in this diff: https://github.com/uber/cadence/pull/3515). For task processing, a task shouldn't take a long time to process, if a DB call takes a long time during task processing, that task should back off and retry, instead of keep wait the for the response from DB and blocking other tasks from being processed.

PR #3547, opened 2020-09-24
Add context parameter to data interfaces

yycptt:
**What changed?**
- Add context parameter to all data interfaces and persistence retryer.
- context.TODO() is used as a placeholder in the caller of data interfaces. Note in this diff even if the caller function already has a context parameter as input, context.TODO() is still used to call data interface as wiring up the use of context parameter for persistence will be done in a later diff. 

**Why?**
#3519

**How did you test it?**
Local test

**Potential risks**
1. Those context parameter are not actually enforced in persistence implementation
2. context.TODO() is used as placeholder for all places where a context is needed, even if the context parameter is used by persistence implementation the behavior will be the same as before as when we don't specify a context in persistence implementation, context.Background() will be used.


host/testcluster.go
```diff
@@ -203,10 +208,13 @@ func NewCluster(options *TestClusterConfig, logger log.Logger) (*TestCluster, er
 func setupShards(testBase persistencetests.TestBase, numHistoryShards int, logger log.Logger) {
 	// shard 0 is always created, we create additional shards if needed
 	for shardID := 1; shardID < numHistoryShards; shardID++ {
-		err := testBase.CreateShard(shardID, "", 0)
+		ctx, cancel := context.WithTimeout(context.Background(), defaultTestPersistenceTimeout)
```
andrewjdawson2016, line 211: defer?

andrewjdawson2016 (approved):

host/testcluster.go
yycptt: It's a for loop, I can't use defer. `cancel()` is called on L214 and L217.

common/messaging/kafkaProducer.go
```diff
@@ -53,7 +54,7 @@ func NewKafkaProducer(topic string, producer sarama.SyncProducer, logger log.Log
 }
 
 // Publish is used to send messages to other clusters through Kafka topic
-func (p *kafkaProducer) Publish(msg interface{}) error {
+func (p *kafkaProducer) Publish(_ context.Context, msg interface{}) error {
```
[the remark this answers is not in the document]
yycptt: Synced offline. Basically domainReplicationQueue need to implement the Producer interface, so I have to change it accordingly.

PR #3551, opened 2020-09-25
Remove abstract data factory

andrewjdawson2016:
The plugin model is the correct way to plumb through a new persistence layer not using this abstract data factory. Therefore this will be removed.


longquanzheng (approved):
lgtm

PR #3555, opened 2020-09-27
Remove kafka dependency except for advanced visibility

longquanzheng:
**What changed?**
 Remove kafka dependency except for advanced visibility

**Why?**
Need to run without Kafka when not using advanced visibility 

**How did you test it?**
will change unit test.
Local run passing

**Potential risks**
Should be no risk

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -48,7 +48,6 @@ import (
 	ce "github.com/uber/cadence/common/errors"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
-	"github.com/uber/cadence/common/messaging"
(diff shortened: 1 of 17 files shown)


longquanzheng:
There are some TODO in the code I will fix in this PR.
Just want to get some early feedback to see if I miss anything @yux0

service/history/historyEngine.go
```diff
@@ -173,7 +173,7 @@ func NewEngineWithShardContext(
 	historyClient hc.Client,
 	publicClient workflowserviceclient.Interface,
 	historyEventNotifier events.Notifier,
-	publisher messaging.Producer,
+	_ messaging.Producer,
```
yux0, line 176: you can remove this argument

yux0 (approved):

PR #3558, opened 2020-09-29
Add Managed Failover 

vancexu:
**What changed?**

Implement managed failover as a cadence workflow, so that the experience would be:
* Start failover. 
  * `Cadence adm cluster failover start --source_cluster <dc> --target_cluster <dc2>`
  * Addition --domains flag to failover only provided domains.
* Show failover progress. 
  * `Cadence adm cluster failover query --r <rid>`
* Pause and resume failover.  
  * `Cadence adm cluster failover pause `
  * `Cadence adm cluster failover resume `
* Abort failover 
   * `Cadence adm cluster failover abort `
* Rollback a failover with retention.
   * `Cadence adm cluster failover rollback --r <rid>`
* Show failover history
  * List workflow 

**Why?**

Cadence failover is designed and expected to be executed by Cadence users, but users find it is hard to failover:
* They don’t know how. Failover requires knowledge on Lisa and CLI with multiple manual steps.
* They don’t know when. During active data center outages, they need to reach out to the Cadence team to figure out when is recommended to failover to where.  

As a result, it would take a long time for Cadence users to mitigate issues by failover domains. 

We need to take over failover for eligible users, so that we can mitigate outage quicker, and reduce operation overhead for users.

**How did you test it?**
Local test
Will also do staging test

**Potential risks**
No risk. 
But aggressive failover all domains may cause target dc under high load.


tools/cli/adminFailoverCommands.go
```diff
@@ -0,0 +1,275 @@
...
+// AdminFailoverList list failover runs
+func AdminFailoverList(c *cli.Context) {
+	c.Set(FlagWorkflowID, failoverManager.WorkflowID)
+	ListWorkflow(c)
+}
+
+func getCadenceClient(c *cli.Context) cclient.Client {
+	svcClient := cFactory.ClientFrontendClient(c)
+	return cclient.NewClient(svcClient, common.SystemLocalDomainName, &cclient.Options{})
+}
```
yycptt, line 256: nit: can we reuse the `getWorkflowClientWithOptionalDomain()` func?

tools/cli/adminFailoverCommands.go
```diff
@@ -0,0 +1,275 @@
...
+	var err error
+	if c.IsSet(FlagRunID) {
+		runID = c.String(FlagRunID)
+	} else {
+		descResp, err = client.DescribeWorkflowExecution(tcCtx, failoverManager.WorkflowID, "")
+		if err != nil {
+			ErrorAndExit("Failed to get failover workflow runID", err)
+		}
+		runID = descResp.WorkflowExecutionInfo.Execution.GetRunId()
+	}
```
yycptt, line 173: If empty runID is used for `QueryWorkflow`, it will automatically use the current runID I believe.

tools/cli/adminFailoverCommands.go
```diff
@@ -0,0 +1,275 @@
...
+
+	params := &startParams{
+		targetCluster:                  queryResult.SourceCluster,
+		sourceCluster:                  queryResult.TargetCluster,
+		domains:                        queryResult.SuccessDomains,
+		batchFailoverSize:              c.Int(FlagFailoverBatchSize),
+		batchFailoverWaitTimeInSeconds: c.Int(FlagFailoverWaitTime),
+		failoverTimeout:                c.Int(FlagFailoverTimeout),
+	}
+	failoverStart(c, params)
```
yycptt, line 244: Do we need to make sure the workflow is closed before making the query and start the rollback workflow?

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+	resumeCh := workflow.GetSignalChannel(ctx, ResumeSignal)
+	var shouldPause bool
+
+	for i := 0; i < times; i++ {
+		// check if need to pause
+		shouldPause = pauseCh.ReceiveAsync(nil)
+		if shouldPause {
+			wfState = wfPaused
+			resumeCh.Receive(ctx, nil)
+		}
```
yycptt, line 186: so we don't plan to handle the case where resume signal is sent before pause or multiple pause signal comes in before a resume signal?

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+func getFailoverActivityOptions() workflow.ActivityOptions {
+	return workflow.ActivityOptions{
+		ScheduleToStartTimeout: 10 * time.Second,
+		StartToCloseTimeout:    10 * time.Second,
+	}
+}
+
+func min(a, b int) int {
```
yycptt, line 233: nit: there's already a function `MinInt()` defined in the common package.

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+			resumeCh.Receive(ctx, nil)
+		}
+		wfState = wfRunning
+
+		failoverActivityParams := &FailoverActivityParams{
+			Domains:       domains[i*batchSize : min((i+1)*batchSize, totalNumOfDomains)],
+			TargetCluster: params.TargetCluster,
+		}
+		var actResult FailoverActivityResult
+		workflow.ExecuteActivity(ao, FailoverActivity, failoverActivityParams).Get(ctx, &actResult)
```
yycptt, line 194: I think we need to properly handle the error here. If worker dies while executing the failover activity, we don't even know which domains have succeeded and which have failed. This will also be a problem for rollback as we don't have the accurate information for successDomains.

yycptt (commented):
Shall we also add some unit test for the workflow implementation?

tools/cli/adminFailoverCommands.go
vancexu: may not worth it, because c.Set(domain) will require domain flag exist, which make it not a simple switch. will leave as it. But thanks for bring it up

service/worker/failoverManager/starter.go
```diff
@@ -0,0 +1,106 @@
...
+
+// Start starts the worker
+func (s *FailoverManager) Start() error {
+	ctx := context.WithValue(context.Background(), failoverManagerContextKey, s)
+	workerOpts := worker.Options{
+		MetricsScope:              s.tallyScope,
+		BackgroundActivityContext: ctx,
+		Tracer:                    opentracing.GlobalTracer(),
+	}
+	// TODO: change to global domain
```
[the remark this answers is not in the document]
vancexu: addressed

tools/cli/adminFailoverCommands.go
vancexu: I improved this part

tools/cli/adminFailoverCommands.go
vancexu: Yes we need to abort first. Added

service/worker/failoverManager/workflow.go
vancexu: My current thinking with this simple pause&resume implementation is operator can always find a way out if they mistakenly pause or resume. 
Another benefit arguable is you can achieve pause twice then resume once to achieve some rate control

I am open to other suggestions.

service/worker/failoverManager/workflow.go
vancexu: cool, removed

service/worker/failoverManager/workflow.go
vancexu: good catch. updated.

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+		SuccessDomains []string
+		FailedDomains  []string
+	}
+)
+
+func init() {
+	workflow.RegisterWithOptions(FailoverWorkflow, workflow.RegisterOptions{Name: WorkflowTypeName})
+	activity.RegisterWithOptions(FailoverActivity, activity.RegisterOptions{Name: failoverActivityName})
+	activity.RegisterWithOptions(GetDomainsActivity, activity.RegisterOptions{Name: getDomainsActivityName})
+}
```
yux0, line 127: Those are deprecated? Should we use the new register?

service/worker/failoverManager/workflow.go
yux0: From the CLI, it can check the workflow state before sending a resume signal

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+	if len(sourceCluster) == 0 {
+		return errors.New(errMsgSourceClusterIsEmpty)
+	}
+	if sourceCluster == targetCluster {
+		return errors.New(errMsgTargetClusterIsSameAsSource)
+	}
+	return nil
+}
+
+func shouldFailover(domain *shared.DescribeDomainResponse, sourceCluster string) bool {
```
yux0, line 294: also check if it is global domain?

service/worker/failoverManager/workflow.go
```diff
@@ -0,0 +1,376 @@
...
+	targetDomainsSet := make(map[string]struct{})
+	if isTargetDomainsProvided {
+		for _, domain := range targetDomains {
+			targetDomainsSet[domain] = struct{}{}
+		}
+	}
+
+	pagesize := int32(200)
+	var token []byte
+	for more := true; more; more = len(token) > 0 {
```
yux0, line 325: nit: we do have a pagination iterator in common package

service/worker/failoverManager/workflow.go
vancexu: cool, added

service/worker/failoverManager/workflow.go
vancexu: we should, I have plan to update all worker to use the new one later.

service/worker/failoverManager/workflow.go
vancexu: I feel this is simple enough so will keep it, thanks for the info though.

tools/cli/adminFailoverCommands.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	var queryResult failovermanager.QueryResult
+	queryResp.Get(&queryResult)
+
+	if isWorkflowRunning(&queryResult) {
+		err := client.TerminateWorkflow(tcCtx, failovermanager.WorkflowID, runID, "Rollback", nil)
+		if err != nil {
+			ErrorAndExit("Failed to terminate failover workflow", err)
+		}
+	}
```
yycptt, line 228: Shall we query the workflow again after terminating it as the `SuccessDomains` may change in the middle?

service/worker/failovermanager/workflow.go
```diff
@@ -0,0 +1,380 @@
...
+		wfState = WorkflowRunning
+
+		failoverActivityParams := &FailoverActivityParams{
+			Domains:       domains[i*batchSize : common.MinInt((i+1)*batchSize, totalNumOfDomains)],
+			TargetCluster: params.TargetCluster,
+		}
+		var actResult FailoverActivityResult
+		err = workflow.ExecuteActivity(ao, FailoverActivity, failoverActivityParams).Get(ctx, &actResult)
+		if err != nil {
+			failedDomains = append(failedDomains, failoverActivityParams.Domains...)
```
yycptt, line 201: I think even if the activity has failed, part of the domains may still be successfully failovered to the target cluster. Then if we do a rollback, not all domains will be rollbacked. We probably need a third unknown state for the domains. And also rollback those unknown domains, which will either be a no-op or be failovered to the source cluster.

tools/cli/adminFailoverCommands.go
vancexu: that's true, added.

service/worker/failovermanager/workflow.go
vancexu: Good point, you are right the current rollback may falsely left domains during rollback.  
I want to keep it simple, so I didn't update here with more state, but let rollback handle all domains (both success and failed), the new start GetDomainActivity will do correct things. 
I also updated comments on query result FailedDomains to let caller know it may contains False positive. I suppose operator will always redo failover for FailedDomains so it may acceptable to contain false positive results.

yycptt (approved):
LGTM

PR #3561, opened 2020-09-29
Add cluster name filter

yux0:
**What changed?**
Add cluster name filter to dynamic config

**Why?**
Add cluster name filter to dynamic config.

**How did you test it?**
staging test

**Potential risks**


common/service/dynamicconfig/config.go
```diff
@@ -44,14 +48,31 @@ func NewCollection(client Client, logger log.Logger) *Collection {
 	}
 }
 
+// NewCollection creates a new collection
+func NewCollection(
+	client Client,
+	clusterName string,
+	logger log.Logger,
+) *Collection {
```
yycptt, line 56: How about changing the signature to `NewCollection(Client, log.Logger, ...Filter)`? I think it will make the code more generic and we don't need the separate `NewCollectionForTest()` function.

yycptt (commented):
Can we also check if the fileBased dynamic config client support this or not?

common/service/dynamicconfig/config.go
yux0: Good point. Changed it

yux0:
> Can we also check if the fileBased dynamic config client support this or not?

Updated

common/service/dynamicconfig/file_based_client.go
```diff
@@ -280,14 +280,14 @@ func (fc *fileBasedClient) getValueWithFilters(key Key, filters map[Filter]inter
...
 func match(v *constrainedValue, filters map[Filter]interface{}) bool {
-	if len(v.Constraints) != len(filters) {
+	if len(v.Constraints) > len(filters) {
 		return false
 	}
 
-	for filter, filterValue := range filters {
-		if v.Constraints[filter.String()] != filterValue {
+	for constrain, constrainedValue := range v.Constraints {
+		if filters[parseFilter(constrain)] != constrainedValue {
```
yycptt, line 290: Shall we check if `parseFilter(constrain)` exists in `filters` first?

yycptt (approved):

PR #3577, opened 2020-10-03
3559 default rpc name

swayamraina:
fixes https://github.com/uber/cadence/issues/3559

### DIFF
--- a/client/clientBean.go
+++ b/client/clientBean.go
@@ -117,13 +117,14 @@ func NewClientBean(factory Factory, dispatcherProvider DispatcherProvider, clust
 			continue
 		}
 
-		dispatcher, err := dispatcherProvider.Get(info.RPCName, info.RPCAddress)
+		name := string(info.RPCName)
+		dispatcher, err := dispatcherProvider.Get(name, info.RPCAddress)
 		if err != nil {
 			return nil, err
 		}
 
 		adminClient, err := factory.NewAdminClientWithTimeoutAndDispatcher(
-			info.RPCName,
+			name,
 			admin.DefaultTimeout,
 			admin.DefaultLargeTimeout,
 			dispatcher,
@@ -133,7 +134,7 @@ func NewClientBean(factory Factory, dispatcherProvider DispatcherProvider, clust
 		}
 
 		frontendClient, err := factory.NewFrontendClientWithTimeoutAndDispatcher(
-			info.RPCName,
+			name,
 			frontend.DefaultTimeout,
 			frontend.DefaultLongPollTimeout,
 			dispatcher,
--- a/cmd/server/cadence/server.go
+++ b/cmd/server/cadence/server.go
@@ -107,7 +107,7 @@ func (s *server) startService() common.Daemon {
 	var err error
 
 	params := service.BootstrapParams{}
-	params.Name = "cadence-" + s.name
+	params.Name = s.name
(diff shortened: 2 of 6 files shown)


client/clientBean.go
```diff
@@ -117,13 +117,14 @@ func NewClientBean(factory Factory, dispatcherProvider DispatcherProvider, clust
 			continue
 		}
 
-		dispatcher, err := dispatcherProvider.Get(info.RPCName, info.RPCAddress)
+		name := string(info.RPCName)
```
longquanzheng, line 120: I am not seeing this change is necessary. Could you explain?

cmd/server/cadence/server.go
```diff
@@ -107,7 +107,7 @@ func (s *server) startService() common.Daemon {
 	var err error
 
 	params := service.BootstrapParams{}
-	params.Name = "cadence-" + s.name
```
longquanzheng, line 110: I think this will not work. We intended to put "cadence-" as prefix, so that in the command line people just use "frontend,history,matching". It's already in a lot of README and scripts, this will be a big change but no necessary.

common/service/config/config.go
```diff
@@ -271,7 +271,7 @@ type (
 		Enabled                bool  `yaml:"enabled"`
 		InitialFailoverVersion int64 `yaml:"initialFailoverVersion"`
 		// RPCName indicate the remote service name
-		RPCName string `yaml:"rpcName"`
+		RPCName RPCName `yaml:"rpcName"`
```
longquanzheng, line 274: I thought you wanted to make a default value here, correct?

common/service/config/config.go
```diff
@@ -419,6 +419,8 @@ type (
 
 	// BootstrapMode is an enum type for ringpop bootstrap mode
 	BootstrapMode int
+
+	RPCName string
```
longquanzheng, line 423: Generally we want to avoid making a type like this, it will make coding harder. We make types like this only when they are enums.

config/development.yaml
```diff
@@ -77,8 +77,13 @@ clusterMetadata:
     standby:
```
longquanzheng, line 77: Also don't forget to update other yamls

longquanzheng:
@swayamraina thank you for your contribution.

common/service/config/config.go
swayamraina: I am trying to put in the default value while doing the unmarshalling, I guess this is what I had hinted on. 
I do agree on you comment on enums, do you have any particular approach which tackles both of these things?

longquanzheng: I thought it can be something like 
````
RPCName string `yaml:"rpcName", default:"true`
```

swayamraina: @longquanzheng I don't think this is a valid syntax for defining the default value
I guess there is no construct of providing defaults in go-yaml

longquanzheng: Yes you are right. There is some third party library can do that but I don't think we want to use. 
I think we just need to check if the string is empty, and if it is then set a default value to it.

swayamraina: Yes, this is what seems right.

longquanzheng:
Hello @swayamraina , are you still working on it?

swayamraina:
Hi, I did some changes to the implementation but didn't get enough time to continue on this

longquanzheng:
already implemented in https://github.com/uber/cadence/pull/3825
@swayamraina lmk if I missed anything

PR #3584, opened 2020-10-04
Fix history scanner panic

longquanzheng:
**What changed?**
Add the missing scanner context to worker option

**Why?**
It was removed somehow when adding the execution scanner to workers. 

**How did you test it?**
Run locally to make sure the issue https://github.com/uber/cadence/issues/3582 is gone

**Potential risks**
No.

### DIFF
--- a/service/worker/scanner/executions/workflows.go
+++ b/service/worker/scanner/executions/workflows.go
@@ -35,13 +35,13 @@ import (
 
 const (
 	// ConcreteScannerContextKey is the key used to access ScannerContext in activities for concrete executions
-	ConcreteScannerContextKey = ContextKey(0)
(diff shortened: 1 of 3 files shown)


service/worker/scanner/executions/workflows.go
```diff
@@ -35,13 +35,13 @@ import (
 
 const (
 	// ConcreteScannerContextKey is the key used to access ScannerContext in activities for concrete executions
-	ConcreteScannerContextKey = ContextKey(0)
```
longquanzheng, line 38: Using string instead of number will make it much unlikely to conflict. The performance penalty should be minimal enough to ignore.

yycptt (approved):

PR #3585, opened 2020-10-05
Add metric plugin model and implement statsd as an example

longquanzheng:
**What changed?**
Add a plugin model and statsd implementation, with README.

**Why?**
For https://github.com/uber/cadence/issues/3530 
And people have asked in Slack channel that they need to report metrics to Datadog.

**How did you test it?**
Will run locally later with Statsd.

**Potential risks**
No.


longquanzheng:
close in favor of https://github.com/uber/cadence/pull/3611

PR #3586, opened 2020-10-05
Added initial internal types and their mappers to/from thrift

vytautas-karpavicius:
**What changed?**
This change introduces a sample of common internal types to be used throughout the Cadence codebase in place of generated Thrift counterparts.

**Why?**
This is a prerequisite for migration to gRPC as we want to stop using Thrift generated types throughout Cadence codebase. Instead start using internal types that are independent of serialization protocol used at network and persistence boundaries.

**How did you test it?**
100% coverage with unit tests

**Potential risks**
None yet. Types are not used yet.

### DIFF
--- a/common/types/common.go
+++ b/common/types/common.go
@@ -1,0 +1,68 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+//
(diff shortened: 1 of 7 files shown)


vytautas-karpavicius:
Closing in favor of https://github.com/uber/cadence/pull/3610

PR #3595, opened 2020-10-06
Add admin command to force update shard rangeID

yycptt:
**What changed?**
Add admin command to force update shard rangeID

**Why?**
https://github.com/uber/cadence/issues/3566

**How did you test it?**
Tested locally

**Potential risks**
N/A. Only CLI is changed. However the command is dangerous and should be executed only when necessary and with great care.


tools/cli/adminCommands.go
```diff
@@ -414,6 +414,37 @@ func AdminDescribeShard(c *cli.Context) {
...
+	updatedShardInfo.Owner = ""
+	updatedShardInfo.UpdatedAt = time.Now()
+
+	if err := shardManager.UpdateShard(ctx, &persistence.UpdateShardRequest{
+		PreviousRangeID: previousRangeID,
+		ShardInfo:       updatedShardInfo,
+	}); err != nil {
+		ErrorAndExit("Failed to reset shard rangeID.", err)
+	}
+}
```
vancexu, line 446: consider adding a success output including PreviousRangeID and new rid

vancexu (approved):

tools/cli/adminCommands.go
yycptt: Good point. Added.

PR #3598, opened 2020-10-06
Separate some types between taskManager and taskStore

anish531213:
Separate some types between taskManager and taskStore that have timer type as int64

### DIFF
--- a/common/persistence/persistenceInterface.go
+++ b/common/persistence/persistenceInterface.go
@@ -46,8 +46,30 @@ type (
 		GetShard(ctx context.Context, request *InternalGetShardRequest) (*InternalGetShardResponse, error)
 		UpdateShard(ctx context.Context, request *InternalUpdateShardRequest) error
 	}
+
 	// TaskStore is a lower level of TaskManager
-	TaskStore = TaskManager
+	TaskStore interface {
+		Closeable
+		GetName() string
+		LeaseTaskList(ctx context.Context, request *LeaseTaskListRequest) (*LeaseTaskListResponse, error)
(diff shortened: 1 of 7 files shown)


common/persistence/persistenceInterface.go
```diff
@@ -46,8 +46,30 @@ type (
 		GetShard(ctx context.Context, request *InternalGetShardRequest) (*InternalGetShardResponse, error)
 		UpdateShard(ctx context.Context, request *InternalUpdateShardRequest) error
 	}
+
 	// TaskStore is a lower level of TaskManager
-	TaskStore = TaskManager
+	TaskStore interface {
+		Closeable
+		GetName() string
+		LeaseTaskList(ctx context.Context, request *LeaseTaskListRequest) (*LeaseTaskListResponse, error)
```
andrewjdawson2016, line 54: `LeaseTaskListRequest` takes a TaskListKind which is an int. That in represents a value of a thrift enum. Which is a serialization detail. Therefore `LeaseTaskListRequest` should be separated.

andrewjdawson2016: The same is true of `LeaseTaskListResponse`

andrewjdawson2016 (changes requested):

common/persistence/persistenceInterface.go
andrewjdawson2016: There are other cases in which this is true as well please take a another pass over those request and response structs. I think many of them have taskListKind. We will want to represent task list kind as an application level enum eventually.

anish531213: The taskListKind is an enum in dataInterface. So, I didn't understand how it is related to serialization here.

andrewjdawson2016 (approved):

PR #3606, opened 2020-10-07
Separate visibility store from visibility manager

andrewjdawson2016:
https://github.com/uber/cadence/issues/3572


anish531213 (approved):
LGTM

PR #3609, opened 2020-10-08
Extend ScheduleActivityTask decision attributes with a requestLocalDispatch field

mkolodezny:
This is basically the prerequisite for the activity dispatch optimization feature to allow worker to dispatch activity tasks through local tunnel after decisions are made. This is a performance optimization to skip activity scheduling efforts.

The follow up is to populate ActivityLocalDispatchInfo response struct for each activity such as: ScheduledTimestamp, StartedTimestamp and probably StartedEventId if needed.     

The corresponding IDL pr https://github.com/uber/cadence-idl/pull/45

**How did you test it?**
unit tests
e2e testing

**Potential risks**
No risk. It is off by default


common/service/dynamicconfig/constants.go
```diff
@@ -280,6 +280,7 @@ var keys = map[Key]string{
 	NotifyFailoverMarkerInterval:                          "history.NotifyFailoverMarkerInterval",
 	NotifyFailoverMarkerTimerJitterCoefficient:            "history.NotifyFailoverMarkerTimerJitterCoefficient",
 	EnableDropStuckTaskByDomainID:                         "history.DropStuckTaskByDomain",
+	EnableRequestActivityLocalDispatchByDomain:            "history.enableRequestActivityLocalDispatchByDomain",
```
meiliang86, line 283: EnableRequestActivityLocalDispatchByDomain -> EnableActivityLocalDispatchByDomain

common/service/dynamicconfig/constants.go
```diff
@@ -904,6 +905,9 @@ const (
 
 	// lastKeyForTest must be the last one in this const group for testing purpose
 	lastKeyForTest
+
+	// EnableRequestActivityLocalDispatchByDomain indicates if client can request to start activity locally
+	EnableRequestActivityLocalDispatchByDomain
```
meiliang86, line 910: Same name change here

common/service/dynamicconfig/constants.go
```diff
@@ -904,6 +905,9 @@ const (
 
 	// lastKeyForTest must be the last one in this const group for testing purpose
 	lastKeyForTest
+
+	// EnableRequestActivityLocalDispatchByDomain indicates if client can request to start activity locally
```
meiliang86, line 909: // EnableActivityLocalDispatchByDomain allows worker to dispatch activity tasks through local tunnel after decisions are made. This is an performance optimization to skip activity scheduling efforts.

service/history/config/config.go
```diff
@@ -254,6 +254,9 @@ type Config struct {
 	NotifyFailoverMarkerInterval               dynamicconfig.DurationPropertyFn
 	NotifyFailoverMarkerTimerJitterCoefficient dynamicconfig.FloatPropertyFn
 	EnableGracefulFailover                     dynamicconfig.BoolPropertyFn
+
+	// Activity dispatch optimization
```
meiliang86, line 258: Apply same comment and name change here.

meiliang86: Same below.

service/history/decisionTaskHandler.go
```diff
@@ -219,9 +220,17 @@ func (handler *decisionTaskHandlerImpl) handleDecisionScheduleActivity(
 		return err
 	}
 
-	_, _, err = handler.mutableState.AddActivityTaskScheduledEvent(handler.decisionTaskCompletedID, attr)
+	dispatchLocally := handler.config.EnableRequestActivityLocalDispatchByDomain(handler.domainEntry.GetInfo().Name) &&
```
meiliang86, line 223: This logic should be put inside AddActivityTaskScheduledEvent

meiliang86: You have TODO below to return additional information anyways. So you can use that to decide if you need to ad the started event.

meiliang86:
Any additional unit test you can add?

service/history/decisionTaskHandler.go
mkolodezny: agree, lets to update decision handler func's to return interface{} to collect all that info

.gen/proto/gogo.pb.go
```diff
@@ -1,1321 +0,0 @@
-// The MIT License (MIT)
```
yux0, line 1: why we remove this one?

.gen/proto/message.pb.go
```diff
@@ -1,15581 +0,0 @@
-// The MIT License (MIT)
```
yux0: same?

service/history/execution/history_builder_test.go
```diff
@@ -841,7 +841,8 @@ func (s *historyBuilderSuite) addActivityTaskScheduledEvent(decisionCompletedID
 			HeartbeatTimeoutSeconds:       common.Int32Ptr(hearbeatTimeout),
 			StartToCloseTimeoutSeconds:    common.Int32Ptr(1),
 			RetryPolicy:                   retryPolicy,
-		})
+		},
+		false)
```
yux0, line 845: Add some test with true case?

yycptt (commented):
Looks good to me! I think it may worth to add an integration test for the e2e behavior so that it won't be break accidentally. I am not sure how much change is needed for the integration test framework though.

mkolodezny:
> Looks good to me! I think it may worth to add an integration test for the e2e behavior so that it won't be break accidentally. I am not sure how much change is needed for the integration test framework though.

sure, thats scheduled as a follow up on client side.

mkolodezny:
> Any additional unit test you can add?

added

meiliang86 (approved):

PR #3611, opened 2020-10-08
Add metric plugin model and implement statsd as an example

longquanzheng:
**What changed?**
Add a plugin model and statsd implementation, with README.

**Why?**
For https://github.com/uber/cadence/issues/3530 
And people have asked in Slack channel that they need to report metrics to Datadog.

**How did you test it?**
Will run locally later with Statsd.

**Potential risks**
No.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -24,6 +24,7 @@ import (
 	"os"
 
 	"github.com/uber/cadence/cmd/server/cadence"
+	_ "github.com/uber/cadence/common/metrics/plugins/statsd"             // needed to load statsd plugin
 	_ "github.com/uber/cadence/common/persistence/sql/sqlplugin/mysql"    // needed to load mysql plugin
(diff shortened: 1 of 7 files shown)


andrewjdawson2016:
Internally at Uber as you know we inject dependencies by having our own main.go wrapper and it injects bootstrap params into the services. 

It looks to me like you are going to try to bootstrap cadence fully with plugins and depend on the main.go in cadence open source repo. Is this right?

longquanzheng:
> Internally at Uber as you know we inject dependencies by having our own main.go wrapper and it injects bootstrap params into the services.
> 
> It looks to me like you are going to try to bootstrap cadence fully with plugins and depend on the main.go in cadence open source repo. Is this right?

Now I think it makes sense to stick with the existing interface model. This plugin model is easy to use(only import is needed) but also has a drawback, which is the config is not strongly typed. And you have a good point that it could be confusing to have two different ways to extend Cadence.

PR #3613, opened 2020-10-08
Convert task store/manager to use internal types

andrewjdawson2016:
https://github.com/uber/cadence/issues/3612

### DIFF
--- a/common/persistence/cassandra/cassandraPersistence.go
+++ b/common/persistence/cassandra/cassandraPersistence.go
@@ -2718,8 +2718,8 @@ func (d *cassandraPersistence) CreateTasks(
 				rowTypeTask,
 				task.TaskID,
 				domainID,
-				task.Execution.GetWorkflowId(),
-				task.Execution.GetRunId(),
+				*task.Execution.WorkflowId,
+				*task.Execution.RunId,
 				scheduleID,
 				cqlNowTimestamp)
 		} else {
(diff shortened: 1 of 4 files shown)


common/persistence/taskManager.go
```diff
@@ -108,7 +110,7 @@ func (t *taskManager) toInternalCreateTaskInfo(createTaskInfo *CreateTaskInfo) *
 		return nil
 	}
 	return &InternalCreateTasksInfo{
-		Execution: createTaskInfo.Execution,
+		Execution: *thrift.ToWorkflowExecution(&createTaskInfo.Execution),
```
anish531213, line 113: how do we handle the case where createTaskInfo.Execution is nil.

anish531213: Also if it returns nil, it would panic right.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -2718,8 +2718,8 @@ func (d *cassandraPersistence) CreateTasks(
 				rowTypeTask,
 				task.TaskID,
 				domainID,
-				task.Execution.GetWorkflowId(),
-				task.Execution.GetRunId(),
+				*task.Execution.WorkflowId,
```
anish531213, line 2721: What will happen if WorkflowId or RunId is nil.

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -2733,8 +2733,8 @@ func (d *cassandraPersistence) CreateTasks(
...
+				*task.Execution.RunId,
```
anish531213, line 2737: Same question here

common/persistence/cassandra/cassandraPersistence.go
yux0, line 2721: I agree with your concern about the type. One way is to have a unit test for it. I think another way is to have cast the type first before passing it in.

var workflowID string
workflowID = *task.Execution.WorkflowId

anish531213 (approved):

PR #3618, opened 2020-10-09
Improve docs search

just-at-uber:
**What changed:** 
- switched to `vuepress-plugin-fulltext-search` plugin (was `vuepress-plugin-flexsearch`) to increase reliability of search function.
  - See **before screenshot** showing 10 entries, 8 of which are false positives (pages don't contain "batch" at all)
  - See **after screenshot** showing 3 correct entries, of which show "batch" at least once on the page.
- Added style overrides for new search box.

**Screenshots**

Before:
Screen Shot 2020-10-09 at 2 44 10 PM

After:
Screen Shot 2020-10-09 at 2 43 46 PM

### DIFF
--- a/src/.vuepress/config.js
+++ b/src/.vuepress/config.js
@@ -13,7 +13,7 @@ module.exports = {
   plugins: [
     '@vuepress/back-to-top',
     'code-switcher',
-    'flexsearch',
+    'fulltext-search',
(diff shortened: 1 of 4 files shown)


meiliang86 (approved):
Thanks @just-at-uber Looks great!

PR #3619, opened 2020-10-10
Move batcher to new domain

vancexu:
**What changed?**
Move batcher to use new domain

**Why?**
We want to enable auth on cadence-system so that only cadence maintainer/team can use CLI or Web to perform admin related tasks, e.g managed failover, scanner, archiver 
To enable it, we will add auth policy on cadence-system domain.
If not moving batcher out from system domain, only cadence team will be able to do batch in CLI, which is wrong.  

I considered several other approaches, this is probably the best long-term solution. but I am open to other suggestions if concerns. 

**How did you test it?**
local test

**Potential risks**
No risk. also batcher is off by default.

### DIFF
--- a/common/constants.go
+++ b/common/constants.go
@@ -105,12 +105,13 @@ const (
 	SystemGlobalDomainName = "cadence-system-global"
 	// SystemLocalDomainName is domain name for cadence system workflows running in local cluster
 	SystemLocalDomainName = "cadence-system"
-	// SystemDomainID is domain id for all cadence system workflows
-	SystemDomainID = "32049b68-7872-4094-8e63-d0dd59896a83"
 	// SystemDomainRetentionDays is retention config for all cadence system workflows
 	SystemDomainRetentionDays = 7
 	// DefaultAdminOperationToken is the default dynamic config value for AdminOperationToken
 	DefaultAdminOperationToken = "CadenceTeamONLY"
+	// BatcherLocalDomainName is domain name for batcher workflows running in local cluster
+	// Batcher cannot use SystemLocalDomain because auth
+	BatcherLocalDomainName = "cadence-batcher"
 )
 
 const (
--- a/common/constants.go
+++ b/common/constants.go
@@ -103,12 +103,16 @@ const (
 const (
 	// SystemGlobalDomainName is global domain name for cadence system workflows running globally
 	SystemGlobalDomainName = "cadence-system-global"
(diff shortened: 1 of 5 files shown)


common/constants.go
```diff
@@ -105,12 +105,13 @@ const (
 	SystemGlobalDomainName = "cadence-system-global"
 	// SystemLocalDomainName is domain name for cadence system workflows running in local cluster
 	SystemLocalDomainName = "cadence-system"
-	// SystemDomainID is domain id for all cadence system workflows
-	SystemDomainID = "32049b68-7872-4094-8e63-d0dd59896a83"
```
mkolodezny, line 109: lets keep it the same, to be 100% backwards compatible :)

service/worker/service.go
```diff
@@ -340,26 +343,26 @@ func (s *Service) startFailoverManager() {
...
 
-func (s *Service) registerSystemDomain() {
+func (s *Service) registerSystemDomain(domain string) {
 
 	currentClusterName := s.GetClusterMetadata().GetCurrentClusterName()
 	_, err := s.GetMetadataManager().CreateDomain(context.Background(), &persistence.CreateDomainRequest{
 		Info: &persistence.DomainInfo{
-			ID:          common.SystemDomainID,
-			Name:        common.SystemLocalDomainName,
+			ID:          uuid.New().String(),
```
mkolodezny, line 364: probably better to stick to the previous approach and use constants for safety sake?

common/constants.go
```diff
@@ -105,12 +105,13 @@ const (
...
 	// SystemDomainRetentionDays is retention config for all cadence system workflows
 	SystemDomainRetentionDays = 7
 	// DefaultAdminOperationToken is the default dynamic config value for AdminOperationToken
 	DefaultAdminOperationToken = "CadenceTeamONLY"
+	// BatcherLocalDomainName is domain name for batcher workflows running in local cluster
+	// Batcher cannot use SystemLocalDomain because auth
+	BatcherLocalDomainName = "cadence-batcher"
```
mkolodezny, line 114: idem

service/worker/service.go
vancexu: agree it might be safer for sql. changed back.

mkolodezny (approved):
Looking good now. The approach seems simple and neat to me. cc @meiliang86

PR #3626, opened 2020-10-12
Satisfy error interface for errors and add FromError mapper

vytautas-karpavicius:
**What changed?**
- Satisfy `error` interface for error related internal types.
- Add common `thrift.FromError` mapper.

Keeping them in separate files, to not touch autogenerated code for now.

**Why?**
Once we start changing types to internal, error related change becomes problematic as errors propagate up everywhere. Hence created this thrift.FromError mapper that checks if error is in internal form and if so - converts it to thrift equivalent. Next we could add Thrift handler wrapper, that would use this mapper as the last step. All internal propagates errors will be converted to thrift there. Once this is done changes switch errors to internal types can be started.

**How did you test it?**

**Potential risks**


andrewjdawson2016 (approved):
Yes looks very reasonable to me.

PR #3627, opened 2020-10-12
Update ttl max value

yux0:
**What changed?**
Update max TTL value to 5 years

**Why?**
There is a max timestamp Cassandra supports. Use the max timestamp, some weird behavior could happen.

**How did you test it?**
DB interaction test

**Potential risks**


vancexu (approved):
Just curious what is the weird error you are seeing caused by this?

yux0:
> Just curious what is the weird error you are seeing caused by this?

From https://docs.datastax.com/en/dse-trblshoot/doc/troubleshooting/recoveringTtlYear2038Problem.html, the data could immediately expire. This explains what we see in the matching task loss

yycptt (approved):

PR #3628, opened 2020-10-12
Include getters for generated internal types

andrewjdawson2016:
This adds getters to internal types. This is needed because we are attempting to move from generated thrift types to internal types and in some places in the code we use thrift accessors. The purpose of these generated getters is to do the same thing that the thrift access methods


common/types/generator/main.go
```diff
@@ -287,9 +310,10 @@ func newType(fullName string) Type {
 		name := fullName[pos+1:]
 		pos = strings.LastIndexByte(pkg, '/')
 		short := pkg[pos+1:]
-		return Type{pkg, short, name, false, isArray, isMap, isPointer}
+		_, ok := enumTypes[name]
+		return Type{pkg, short, name, false, isArray, isMap, isPointer, ok}
```
anish531213, line 314: Maybe create something like isEnum and set the default value as false.

anish531213 (approved):
LGTM

vytautas-karpavicius (approved):

PR #3632, opened 2020-10-12
Unify shard rangeID column and field in Cass persistence

yycptt:
**What changed?**
Unify shard rangeID column and field in Cass persistence

**Why?**
When data stored in cass in corrupted, it's possible that value stored in shard rangeID column is different from the one stored in shardInfo struct, which will result in CAS operation failure when we try to acquire the shard.

This pr will always return the rangeID  stored in range_id column to prevent CAS failure. It will also perform a sanity check to detect in consistency between two rangeIDs and try to fixed the inconsistency.

**How did you test it?**
Tested locally.

**Potential risks**
In the worst case, we may failed to acquire a shard.


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1078,11 +1089,77 @@ func (d *cassandraPersistence) GetShard(
...
+	// shardInfoRangeID and return rangeID instead. Later when updating the shard, CAS can still succeed
+	// as the value from rangeID columns is returned, shardInfoRangeID will also be updated to the correct value.
+	rangeID := result["range_id"].(int64)
+	shard := result["shard"].(map[string]interface{})
+	shardInfoRangeID := shard["range_id"].(int64)
+	if shardInfoRangeID > rangeID {
+		// In this case we need to fix the rangeID column before returning the result as:
+		// 1. if we return shardInfoRangeID, then later shard CAS operation will fail
+		// 2. if we still return rangeID, CAS will work but rangeID will move backward which
+		// result in lost tasks, corrupted workflow history, etc.
```
yux0, line 1105: Maybe add a fatal logging there

common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1078,11 +1089,77 @@ func (d *cassandraPersistence) GetShard(
...
+	// check if rangeID column and rangeID field in shard column matches, if not we need to pick the larger
+	// rangeID.
+	//
+	// If shardInfoRangeID < rangeID, we don't need to do anything here as createShardInfo will ignore
+	// shardInfoRangeID and return rangeID instead. Later when updating the shard, CAS can still succeed
+	// as the value from rangeID columns is returned, shardInfoRangeID will also be updated to the correct value.
+	rangeID := result["range_id"].(int64)
+	shard := result["shard"].(map[string]interface{})
+	shardInfoRangeID := shard["range_id"].(int64)
+	if shardInfoRangeID > rangeID {
```
yux0, line 1101: If the shard info range ID is deprecated, do we still need this check? 
Should we handle if rangeID > shard info range ID?

common/persistence/cassandra/cassandraPersistence.go
yycptt: We don't need to panic here, the logic below can fix the corrupted rangeID. I can add a Warning log though.

common/persistence/cassandra/cassandraPersistence.go
yycptt: We will update rangeID field in shardInfo. So we have more chance for catching data corruption. 

I explained the case for rangeID > shard info range ID in the comments on L1095. Basically we don't need to handle that case.

yux0 (approved):

PR #3633, opened 2020-10-12
Fix race condition in task redispatcher test

yycptt:
**What changed?**
Fix race condition in task redispatcher test

**Why?**
Fix race condition

**How did you test it?**
`go test -run TestRedispatcherSuite  -race -v -count=100` passes

**Potential risks**
N/A


service/history/task/redispatcher_test.go
```diff
@@ -72,7 +72,9 @@ func (s *redispatcherSuite) SetupTest() {
 
 func (s *redispatcherSuite) TearDownTest() {
 	s.redispatcher.Stop()
+	s.redispatcher.Lock()
 	s.Nil(s.redispatcher.redispatchTimer)
+	s.redispatcher.Unlock()
```
mkolodezny, line 77: Nice, thanks for the quick fix!

mkolodezny (approved):
Looking good!

PR #3638, opened 2020-10-13
[scanner] add generic shard scanner

mantas-sidlauskas:
**What changed?**
Shard scanner is a generic shard scanner. 

To use shard scanner you need to provide: 
* invariant manager - to set-up check and fix actions defined by scanner user.
* iterator - shard scanner uses iterator interface to iterate over entities.
* config resolver (optional) - scanner users can have their own custom dynamic configs. Config values are recorded for better determinism.

**Why?**
This shard scanner helps to reduce boilerplate when introducing new types of scanners, no need to reimplement/duplicate common logic.

**How did you test it?**
Unit tests.

**Potential risks**
None, this is noop change


service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+			Enabled:                 dc.Enabled(),
+			Concurrency:             dc.Concurrency(),
+			PageSize:                dc.PageSize(),
+			BlobstoreFlushThreshold: dc.BlobstoreFlushThreshold(),
+			ActivityBatchSize:       dc.ActivityBatchSize(),
+		},
+	}
+
+	if ctx.Hooks != nil && ctx.Hooks.Config != nil {
+		result.CustomScannerConfig = ctx.Hooks.Config(ctx)
```
andrewjdawson2016, line 74: Is config the right name for this? These actually contain callback functions right?

service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func scanShard(
+	activityCtx context.Context,
+	params ScanShardActivityParams,
+	shardID int,
+	heartbeatDetails ScanShardHeartbeatDetails,
+) (*ScanReport, error) {
+	ctx := activityCtx.Value(params.ContextKey).(Context)
+	scope := ctx.Scope.Tagged(metrics.ActivityTypeTag(params.ContextKey.String()))
```
andrewjdawson2016, line 137: Does params.ContextKey.String() identify the scan type or this activity name?

andrewjdawson2016: It seems like we want a tag called scan - or maybe a subscope created at the root each scanner to the separation between different scanner's metrics is clear.

service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+	resources := ctx.Resource
+	execManager, err := resources.GetExecutionManager(shardID)
+	if err != nil {
+		scope.IncCounter(metrics.CadenceFailures)
+		return nil, err
+	}
+
+	pr := persistence.NewPersistenceRetryer(execManager, resources.GetHistoryManager(), c.CreatePersistenceRetryPolicy())
+
+	scanner := NewScanner(
```
andrewjdawson2016, line 154: Nice 👍

service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+		shardID,
+		ctx.Hooks.Iterator(pr, params, *ctx.Config),
+		resources.GetBlobstoreClient(),
+		params.BlobstoreFlushThreshold,
+		ctx.Hooks.Manager(pr, params, *ctx.Config),
+		func() { activity.RecordHeartbeat(activityCtx, heartbeatDetails) },
+	)
+	report := scanner.Scan(activityCtx)
```
andrewjdawson2016, line 162: It seems to me that report should be a struct composed of two types: GenericReportInfo and ScanSpecificyInfo

Just like you have for config.

andrewjdawson2016: There are parts of the report that deal only at the level of if shard was successful or not but there are also going to be parts of the report that are specific to the type of scan being run. We can do this in a follow up diff because this one is already very large.

service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+	heartbeatDetails FixShardHeartbeatDetails,
+) (*FixReport, error) {
+	ctx := activityCtx.Value(params.ContextKey).(FixerContext)
+	resources := ctx.Resource
+	scope := ctx.Scope.Tagged(metrics.ActivityTypeTag(ActivityFixShard))
+	sw := scope.StartTimer(metrics.CadenceLatency)
+	defer sw.Stop()
+
+	if ctx.Hooks == nil {
+		return nil, errors.New("fixer hooks are not provided")
```
andrewjdawson2016, line 283: I think the fixShardActivity and scanShardActivity has a retry policy on them. If that is the case then I think we want to make sure we are treating this case as a non-retryable error.

service/worker/scanner/shardscanner/activities.go
```diff
@@ -0,0 +1,336 @@
...
+		scope.Tagged(metrics.InvariantTypeTag(string(k))).UpdateGauge(metrics.ScannerCorruptionByTypeGauge, float64(v))
+	}
+	shardStats := params.ShardDistributionStats
+	scope.UpdateGauge(metrics.ScannerShardSizeMaxGauge, float64(shardStats.Max))
+	scope.UpdateGauge(metrics.ScannerShardSizeMedianGauge, float64(shardStats.Median))
+	scope.UpdateGauge(metrics.ScannerShardSizeMinGauge, float64(shardStats.Min))
+	scope.UpdateGauge(metrics.ScannerShardSizeNinetyGauge, float64(shardStats.P90))
+	scope.UpdateGauge(metrics.ScannerShardSizeSeventyFiveGauge, float64(shardStats.P75))
+	scope.UpdateGauge(metrics.ScannerShardSizeTwentyFiveGauge, float64(shardStats.P25))
+	scope.UpdateGauge(metrics.ScannerShardSizeTenGauge, float64(shardStats.P10))
```
andrewjdawson2016, line 334: As per my earlier comment about the scan type specific report I think right here we should have something that is basically like 

if hooks != nil {
    hooks.EmitMetrics(report)
}

service/worker/scanner/shardscanner/fixer_workflow.go
```diff
@@ -0,0 +1,226 @@
...
+
+	corruptKeys, err := GetCorruptedKeys(ctx, wf.Params)
+	if err != nil {
+		return nil, err
+	}
+	wf.Keys = corruptKeys
+	wf.Aggregator = NewShardFixResultAggregator(corruptKeys.CorruptedKeys, *corruptKeys.MinShard, *corruptKeys.MaxShard)
+
+	for name, fn := range setHandlers(wf.Aggregator) {
+		if err := workflow.SetQueryHandler(ctx, name, fn); err != nil {
```
andrewjdawson2016, line 108: So are different scan types able to set different types of queries? For example what if I want a query on timers that says how many timers are going to fire between time X and Y?

service/worker/scanner/shardscanner/activities.go
mantas-sidlauskas: Yes, this is an actual bug. I need to change ActivityType to actual activity name, will add workflow type to distinguish between scanners. Will add check to tests as well.

Regarding subscope: our `metrics` scope does not implement SubScope interface method.

service/worker/scanner/shardscanner/fixer_workflow.go
mantas-sidlauskas: Yes, we are on the same page here. (the same applies to metrics part) I will make those parts extendable in a separate PR.

service/worker/scanner/shardscanner/activities.go
mantas-sidlauskas: renamed to `GetScannerConfig`

service/worker/scanner/shardscanner/activities.go
mantas-sidlauskas: good catch. Added non-retryable error.

andrewjdawson2016 (approved):

PR #3645, opened 2020-10-14
Fix if the replication state is nil

yux0:
**What changed?**
Fix if the replication state is nil

**Why?**
NPE

**How did you test it?**

**Potential risks**


common/persistence/cassandra/cassandraPersistence.go
```diff
@@ -1684,12 +1684,16 @@ func (d *cassandraPersistence) GetCurrentExecution(
 	currentRunID := result["current_run_id"].(gocql.UUID).String()
 	executionInfo := createWorkflowExecutionInfo(result["execution"].(map[string]interface{}))
 	replicationState := createReplicationState(result["replication_state"].(map[string]interface{}))
+	lastWriteVersion := common.EmptyVersion
+	if replicationState != nil {
```
mkolodezny, line 1688: was it not supposed to be nil at all times?

mkolodezny (approved):
Thanks for the quick fix!

common/persistence/cassandra/cassandraPersistence.go
yux0: With the current logic, we still write into replication state

PR #3646, opened 2020-10-14
Separate thrift handler for history service

vytautas-karpavicius:
**What changed?**
- Created Handler interface for history service
- Previous Handler struct renamed to HandlerImpl
- Created ThriftHandler wrapper on top of Handler interface. At this point it only forward requests as is to underlying handler.
- Created Mock for Handler interface.
- Covered ThriftHandler with units tests. Test that requests are forwarded to underlying handler and response is returned as is.
- Replaced handler registration. Instead of registering HandlerImpl directly, wrap it with ThriftHandler and register it instead.

**Why?**
This is purely a mechanical refactoring that introduces additional ThriftHandler wrapper. For now it only forwards request to the original handler. Later on this will become a point for conversion between internal types and thrift types. Underlying handler will be independent of rpc protocol types.

**How did you test it?**
- Covered ThriftHandler with unit tests
- Started server locally

**Potential risks**
None, purely mechanical change.


service/history/handler.go
```diff
@@ -53,9 +52,55 @@ import (
...
+		SignalWithStartWorkflowExecution(context.Context, *hist.SignalWithStartWorkflowExecutionRequest) (*gen.StartWorkflowExecutionResponse, error)
+		SignalWorkflowExecution(context.Context, *hist.SignalWorkflowExecutionRequest) error
+		StartWorkflowExecution(context.Context, *hist.StartWorkflowExecutionRequest) (*gen.StartWorkflowExecutionResponse, error)
+		SyncActivity(context.Context, *hist.SyncActivityRequest) error
+		SyncShardStatus(context.Context, *hist.SyncShardStatusRequest) error
+		TerminateWorkflowExecution(context.Context, *hist.TerminateWorkflowExecutionRequest) error
+	}
+
+	// HandlerImpl is an implementation for history service independent of wire protocol
+	HandlerImpl struct {
```
yux0, line 103: handlerImpl?

service/history/handler.go
```diff
@@ -92,8 +137,8 @@ var (
 func NewHandler(
 	resource resource.Resource,
 	config *config.Config,
-) *Handler {
-	handler := &Handler{
+) *HandlerImpl {
```
yux0, line 140: could we return interface here?

service/history/service.go
```diff
@@ -42,7 +42,7 @@ type Service struct {
 	resource.Resource
 
 	status  int32
-	handler *Handler
+	handler *HandlerImpl
```
yux0, line 45: Just use the interface rather than the implementation?

service/history/MockHandler.go
```diff
@@ -0,0 +1,643 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/uber/cadence/service/history (interfaces: Handler)
+
+// Package history is a generated GoMock package.
+package history
```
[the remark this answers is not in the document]
vytautas-karpavicius: Agree, updated to handler_mock.go

service/history/handler.go
vytautas-karpavicius: Agree, no need to expose this. Updated.

service/history/handler.go
vytautas-karpavicius: As per Go guidelines we should return concrete implementations instead of interfaces.

service/history/service.go
vytautas-karpavicius: At this point `handler` also use other methods (`Start`, `Stop`,...) defined on `HandlerImpl` but not on `Handler` interface. I wanted to keep this interface lean and only contain actual handler methods, not the utilities like `Start` and `Stop`.

Alternatively we could separate them into another interface and use that here instead. But for the time being and do not see immediate benefit for that and I wanted to keep this diff simple and mechanical.

service/history/handler.go
yux0: Thanks. Good to know about this

yux0 (approved):

PR #3650, opened 2020-10-14
Add getters to internal generated types

andrewjdawson2016:
This adds getters to internal types. This is needed because we are attempting to move from generated thrift types to internal types and in some places in the code we use thrift accessors. The purpose of these generated getters is to do the same thing that the thrift access methods

### DIFF
--- a/common/types/shared.go
+++ b/common/types/shared.go
@@ -25,10 +25,42 @@ type AccessDeniedError struct {
 	Message string
 }
 
+func (v *AccessDeniedError) GetMessage() (o string) {
+	if v != nil {
+		return v.Message
+	}
+	return
+}
+
+// ActivityLocalDispatchInfo is an internal type (TBD...)
+type ActivityLocalDispatchInfo struct {
+	ActivityID *string
+}
+
+func (v *ActivityLocalDispatchInfo) GetActivityID() (o string) {
+	if v != nil && v.ActivityID != nil {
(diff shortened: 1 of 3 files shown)


anish531213 (approved):

common/types/shared.go
```diff
@@ -25,12 +25,44 @@ type AccessDeniedError struct {
...
+func (v *AccessDeniedError) GetMessage() (o string) {
+	if v != nil {
+		return v.Message
+	}
+	return
+}
+
+// ActivityLocalDispatchInfo is an internal type (TBD...)
+type ActivityLocalDispatchInfo struct {
+	ActivityID *string
```
mkolodezny, line 37: please add the all fields to match the original struct https://github.com/uber/cadence/blob/6e5f9ece2d632d070a3ad5640dcdc7d90abb4f7d/.gen/go/shared/shared.go#L179:

```
type ActivityLocalDispatchInfo struct {
	ActivityId                      *string `json:"activityId,omitempty"`
	ScheduledTimestamp              *int64  `json:"scheduledTimestamp,omitempty"`
	StartedTimestamp                *int64  `json:"startedTimestamp,omitempty"`
	ScheduledTimestampOfThisAttempt *int64  `json:"scheduledTimestampOfThisAttempt,omitempty"`
	TaskToken                       []byte  `json:"taskToken,omitempty"`
}
```

Issue #3612, opened 2020-10-08
Persistence stores move to internal types


andrewjdawson2016:
We have now converted persistence stores to use types which are decoupled from persistence manager types. We have also landed a PR which introduces internal types and coverts to/from thrift. 

The next step to continue to introduce these internal types into the codebase is start to use them at the persistence store level and below. Currently persistence managers do serialization/deserialization of thrift fields in the request/responses of persistence stores. This serialization/deserialization will be replaced with logic that converts to/from application types. So at the persistence store level and below we will work with these application types. All code at the persistence manager layer and above will be unchanged. 

So the types will look as follows 

- Application code will still deal with thrift types and pass those types to persistence managers 
- Persistence managers will accept/return request/response structs which have thrift fields. 
- Persistence managers will convert those thrift fields to internal types (instead of serializing them like we do today) and hand that new type to persistence stores (this is why persistence store types had to be decoupled from persistence manager types).  
- Persistence stores will now be operating on application level types (decoupled from the serialization format). The stores will convert back to thrift, serialize and write to their backing database. 
- When persistence stores send a response to persistence managers that response should contain only application level types - no thrift or data blobs. The persistence managers will convert those application level types back to thrift for the rest of the application to work with. 

This change increases the number of type translations we have to do. Once we have plumbed these internal types all the way up to the network layer there will not be this need to translate to and from application types as much. Instead we will simply do one translation at the network layer from/to a thrift/proto type to an internal type. 

There are two risks associated with this change 
1. If we do any of the translations incorrectly it can result in writing corrupted data to the database. 
2. This change will increase the number of small structs we create which can put pressure on golang GC. 

After this change is done we have the following next steps 
- Converting to use application layer types instead of thrift types is not good enough. We also need to represent time as time.Time and time.Duration instead of int64 and int32. The persistence store request/response structs should convert to using time.Time and time.Duration. This will be done as a separate pass to minimize the scope of these changes.
- We need to plumb the application layer types higher and higher in the stack until they reach the network layer. 
- Once we plumb the application layer types higher than the persistence managers we should clean up the types between persistence managers and persistence stores. As of today persistence managers just do serialization/deserialization, in a world where both managers and stores are using application level types I see no reason for persistence managers to exist at all.

PR #3656, opened 2020-10-14
Convert visibilityStore/manager to use internal types

anish531213:
#3612


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -116,6 +119,7 @@ func (v *esVisibilityStore) RecordWorkflowExecutionStarted(
 	request *p.InternalRecordWorkflowExecutionStartedRequest,
 ) error {
 	v.checkProducer()
+	memo := v.serializeMemo(thrift.FromMemo(request.Memo), request.DomainUUID, request.WorkflowID, request.RunID)
```
andrewjdawson2016, line 122: Looks to me that everywhere you call serializeMemo you call thrift.FromMemo - probably it makes sense to just pass request.Memo and call thrift.FromMemo in serializeMemo

andrewjdawson2016 (approved):

PR #3658, opened 2020-10-15
Leaner frontend Handler interface

vytautas-karpavicius:
**What changed?**
Leaner frontend `Handler` interface by dropping all methods that are not actual handler methods.

**Why?**
This simplifies decorating implementations, as they do not need to forward these utility calls. Instead explicitly pass required dependencies (Config, Resources) via constructors. Also keep base handler reference within Service, to Stop it and set Health status. 

I will follow up this, with another Thrift specific handler, that will implement this leaner interface as well. While this refactoring is not strictly required for that, I believe only actual handler methods should be decorated - not the utility methods that are specific to base implementation only.

**How did you test it?**
Existing unit tests.

**Potential risks**


vancexu (approved):
Looks great

PR #3664, opened 2020-10-16
Support TTL where needed for SQL plugin interface

andrewjdawson2016:
MySQL does not support TTLs. Cadence's Cassandra implementation depends on TTLs. The way we handle this for our SQL implementation is we have a workflow that scans over tasks and task lists and removes old ones. Internally we are implementing a plugin on top of a sql store which does support TTLs. We prefer to use this TTL support instead of depending on a workflow to do the cleanup. In order to do this we need to extend the sql plugin interface to support TTL. 

This diff adds support for  creating matching tasks and task lists with a ttl.


common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -524,6 +545,7 @@ type (
 		WriteLockShards(ctx context.Context, filter *ShardsFilter) (int, error)
 
 		InsertIntoTasks(ctx context.Context, rows []TasksRow) (sql.Result, error)
+		InsertIntoTasksWithTTL(ctx context.Context, rows []TasksRowWithTTL) (sql.Result, error)
```
longquanzheng, line 548: Can you add some comments why there are two similar APIs and the relation ship with SupportsTTL()? I think it will make it easier for anyone want to implement this interface can is not confused.

longquanzheng (approved):
lgtm

PR #3666, opened 2020-10-18
Refactor ElasticSearch visibility code path for multi version support

longquanzheng:
**What changed?**
Refactor ElasticSearch visibility code path for multi version support
The current way of implementing ElasticSearch is very intrusive in our application layer code. The v6 library is being used everywhere in the service layer code. 
However, the v7/v8 versions of ElasticSearch are major version difference. They need to be run with a completely different library.  See https://github.com/olivere/elastic
Therefore, to support multiple major versions of ElasticSearch, we have to do a big refactoring around the code path, to extract and logic of v6 specific, and making abstraction as interfaces for all future versions of ElasticSearch clients. 

**Why?**
For https://github.com/uber/cadence/issues/2866

**How did you test it?**
Local tests and existing tests.

**Potential risks**
Even though the intention is no behavior change, this refactoring is more risker than other persistence refactoring. 
We need to run stress test to make sure edge handling has no changes.


cmd/server/cadence/server.go
```diff
@@ -175,7 +175,7 @@ func (s *server) startService() common.Daemon {
 		}
 
 		params.ESConfig = advancedVisStore.ElasticSearch
-		esClient, err := elasticsearch.NewClient(params.ESConfig)
+		esClient, err := elasticsearch.NewGenericElasticSearchClient(params.ESConfig, s.cfg.Persistence.VisibilityConfig, params.Logger)
```
vancexu, line 178: Why we need a new `NewGenericElasticSearchClient`? My feeling is es7/8 just need to implement existing Client interface (https://github.com/uber/cadence/blob/6eb918ea60ab63c2f9896ef2d712abfe29a44619/common/elasticsearch/client.go#L34)

longquanzheng: yeah agreed. I plan to have that config in the next PR when introducing V7 as client.

longquanzheng: The client interface is not implementable by v7/v8. The v7/v8 versions of ElasticSearch are major version difference. They need to be run with a completely different library. The current Client interface is already binded with v6. For example, `*elastic.SearchResult` is the result just for v6.

longquanzheng: So the very  basic idea is that the interface needs to be defined in a way that it doesn't touch/refer to any v6 library struct/interfaces.

longquanzheng: Agreed about the config. I will add that in the next PR when introducing V7 as the implementation. This PR is just refactoring to make sure no behavior changes.

common/elasticsearch/interfaces.go
```diff
@@ -0,0 +1,225 @@
...
+	visibilityConfig *config.VisibilityConfig,
+	logger log.Logger,
+) (GenericElasticSearch, error) {
+	// TODO hardcoded to V6 for now
+	return newV6Client(connectConfig, visibilityConfig, logger)
+}
+
+type (
+	// GenericElasticSearch is a generic interface for all versions of ElasticSearch clients
+	GenericElasticSearch interface {
```
meiliang86, line 50: Call it GenericClient? It's under elasticsearch folder so we can omit that part?

common/elasticsearch/interfaces.go
```diff
@@ -0,0 +1,225 @@
...
+	// ScanByQueryRequest is request for SearchByQuery
+	ScanByQueryRequest struct {
+		Index         string
+		Query         string
+		NextPageToken []byte
+		PageSize      int
+	}
+
+	// GenericBulkProcessor is a bulk processor
+	GenericBulkProcessor interface {
```
yycptt, line 95: Maybe just `BulkProcessor`? I think if a definition lives in this file, it implies it's "Generic" to all elastic version, so probably we can drop all "Generic" in the interface or struct definition?

common/elasticsearch/interfaces.go
longquanzheng: Could we use GenericClient? This makes it clear the purpose of this interface is to generalize all elasticSearch libraries. 
 
Btw not really strong opioned at that. You could make a decision and let me know.

common/elasticsearch/interfaces.go
```diff
@@ -0,0 +1,225 @@
...
+type (
+	// GenericElasticSearch is a generic interface for all versions of ElasticSearch clients
+	GenericElasticSearch interface {
+		Search(ctx context.Context, request *SearchRequest) (*p.InternalListWorkflowExecutionsResponse, error)
+		SearchByQuery(ctx context.Context, request *SearchByQueryRequest) (*p.InternalListWorkflowExecutionsResponse, error)
+		ScanByQuery(ctx context.Context, request *ScanByQueryRequest) (*p.InternalListWorkflowExecutionsResponse, error)
+		CountByQuery(ctx context.Context, index, query string) (int64, error)
+		PutMapping(ctx context.Context, index, root, key, valueType string) error
+		CreateIndex(ctx context.Context, index string) error
+		GetClosedWorkflowExecution(ctx context.Context, index string, request *p.InternalGetClosedWorkflowExecutionRequest) (*p.InternalGetClosedWorkflowExecutionResponse, error)
```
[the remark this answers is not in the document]
longquanzheng: Good point. I initially wanted to define different struct but at the end realize that it will be just the same as the persistence one, and to avoid having lots of code copying struct I simply use the same struct. 
I think a better idea is to use type alias -- this is the way we were using between persitenceInterfaces and dataInterfaces when some structs are exactly the same. This will make things more clear. And when it comes the time that one struct need to be different, we could separate them .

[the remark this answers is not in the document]
longquanzheng: Because the implementation of this method doesn't share the request struct with others. If you look at the code today https://github.com/uber/cadence/blob/51355edb4e3769666e5e8a4c0ccd9a0f224cdb79/common/persistence/elasticsearch/esVisibilityStore.go#L392 it cannot reuse any logic like ListCloseWorkflows etc. 

This method seems odd indeed, and not generic enough. I am going to use type alias and hope it look better :)

common/elasticsearch/interfaces.go
```diff
@@ -0,0 +1,225 @@
...
+		Memo          []byte
+		Encoding      string
+		TaskList      string
+		Attr          map[string]interface{}
+	}
+)
+
+// ExponentialBackoff implements the simple exponential backoff described by
+// Douglas Thain at http://dthain.blogspot.de/2009/02/exponential-backoff-in-distributed.html.
+type ExponentialBackoff struct {
```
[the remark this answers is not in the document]
longquanzheng: Technically yes, but not that simple. I simply copied it from the elasticsearch library, and make an interface 
```
	GenericBackoff interface {
		// Next implements a BackoffFunc.
		Next(retry int) (time.Duration, bool)
	}
```
The common/backoff one is more complicated than that. And to lower the risk of changing behavior, since there are too many changes, I recommend to keep this for now. I could leave a TODO and an issue for this and improve it later. How does that sound?

common/elasticsearch/interfaces.go
longquanzheng: I am fine with naming, because it's better to let more people be happy with names to maintain the code. 
I am actually coming from another side that I feel like making this explicitly "Generic" is better. But again this is very minor things.

common/elasticsearch/interfaces.go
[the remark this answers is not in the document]
longquanzheng: I added a TODO comment and an issue for that.

common/elasticsearch/interfaces.go
meiliang86: I am fine with GenericClient.

yycptt: I see. Yeah, I also don't have strong opinion on naming. `GenericClient` looks fine to me.

common/elasticsearch/interfaces.go
yycptt: Let's keep the name to be consistent with `GenericClient`.

common/elasticsearch/interfaces.go
longquanzheng: Thanks both :D

common/elasticsearch/backoff.go
```diff
@@ -0,0 +1,58 @@
...
+	"math"
+	"math/rand"
+	"time"
+)
+
+// ExponentialBackoff implements the simple exponential backoff described by 
+// Douglas Thain at http://dthain.blogspot.de/2009/02/exponential-backoff-in-distributed.html.
+// TODO https://github.com/uber/cadence/issues/3676
+type ExponentialBackoff struct {
+	t float64 // initial timeout (in msec)
```
meiliang86, line 33: Can we use proper names? :)

common/elasticsearch/client_v6.go
```diff
@@ -0,0 +1,664 @@
...
+var (
+	oneMilliSecondInNano = int64(1000)
+)
+
+// NewWrapperClient returns a new implementation of Client
+func newV6Client(
+	connectConfig *config.ElasticSearchConfig,
+	visibilityConfig *config.VisibilityConfig,
+	logger log.Logger,
+) (GenericClient, error) {
```
meiliang86, line 88: Return struct (pointer) instead. And add type assertion.

common/elasticsearch/client_v6.go
```diff
@@ -0,0 +1,664 @@
...
+		BulkSize      int
+		FlushInterval time.Duration
+		Backoff       elastic.Backoff
+		BeforeFunc    elastic.BulkBeforeFunc
+		AfterFunc     elastic.BulkAfterFunc
+	}
+)
+
+var (
+	oneMilliSecondInNano = int64(1000)
```
meiliang86, line 80: millis/nanos = 10^6
use time.Millisecond / time.Nanosecond instead?

service/worker/indexer/indexer.go
```diff
@@ -60,7 +62,7 @@ const (
 )
 
 // NewIndexer create a new Indexer
-func NewIndexer(config *Config, client messaging.Client, esClient es.Client, esConfig *es.Config,
+func NewIndexer(config *Config, client messaging.Client, esClient es.GenericClient, esConfig *config.ElasticSearchConfig,
```
meiliang86, line 65: One param per line?

service/worker/indexer/indexer.go
```diff
@@ -23,6 +23,8 @@ package indexer
 import (
 	"fmt"
 
+	"github.com/uber/cadence/common/service/config"
+
```
meiliang86, line 27: remove extra line

common/elasticsearch/backoff.go
longquanzheng: yeah sure. (i copied from the v6 library but didn't look into the details).

common/elasticsearch/client_v6.go
longquanzheng: No I think we should return interface. Same as the `ExponentialBackoff`. If returning struct, the struct needs to be public, otherwise you will get compiler warning. I don't see it's necessary to return the v6 struct here.

service/worker/indexer/indexer.go
longquanzheng: sure.

service/worker/indexer/indexer.go
longquanzheng: okay

common/elasticsearch/client_v6.go
longquanzheng: Interesting. I copied it from https://github.com/uber/cadence/blob/85a2c9c1c50ae1274e6f5c6d9f87008d1d79dbf5/common/persistence/elasticsearch/esVisibilityStore.go#L97
but didn't pay attention. 
looks like it should be renamed to oneMicroSecondInNano? @vancexu

longquanzheng: Maybe rename it to oneMicroSecondInNano for now and leave an issue/TODO comment to tackle later. What do you think?

common/elasticsearch/backoff.go
longquanzheng: done, also make the struct private.

common/elasticsearch/client_v6.go
meiliang86: sure

yycptt (approved):
LGTM, just some nit comments.

common/elasticsearch/client_v6.go
```diff
@@ -0,0 +1,663 @@
...
+		searchResult, scrollService, err = c.scrollFirstPage(ctx, request.Index, request.Query)
+	} else {
+		searchResult, scrollService, err = c.scroll(ctx, token.ScrollID)
+	}
+
+	isLastPage := false
+	if err == io.EOF { // no more result
+		isLastPage = true
+		if scrollService != nil {
+			err := scrollService.Clear(context.Background())
```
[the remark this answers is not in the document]
longquanzheng: sure, done

common/elasticsearch/interfaces.go
```diff
@@ -0,0 +1,213 @@
...
+
+package elasticsearch
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/uber/cadence/common/metrics"
+
```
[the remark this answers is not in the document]
longquanzheng: updated :D

PR #3667, opened 2020-10-19
Improve reset CLI: add DecisionCompletedTime as resetType and also provide …

longquanzheng:
…lastest runID as default

**What changed?**
1. The primitive reset command will use latest runID by default. 
2. Provide one more predefined reset type "DecisionStartedTime" for reset and reset-batch commands. 

**Why?**
1. It's annoying to require runID. Most of the time people just want to reset the current run or latest run. 
2. It will be useful to allow reset by time as resetType for https://github.com/uber/cadence/issues/2237

**How did you test it?**
```
./cadence  --do sample wf reset -w e8988544-5852-42c7-a74c-e4b74ffb1f25 --reason test --reset_type DecisionCompletedTime --earliest_time 2020-11-01T12:00:31-08:00
resetType: DecisionCompletedTime
{
  "runId": "20a4845d-9c1d-419b-9469-e78a1f61eac7"
}

```

**Potential risks**
No risk

### DIFF
--- a/tools/cli/workflow.go
+++ b/tools/cli/workflow.go
@@ -260,11 +260,6 @@ func newWorkflowCommands() []cli.Command {
 					Name:  FlagResetBadBinaryChecksum,
 					Usage: "Binary checksum for resetType of BadBinary",
 				},
-				cli.BoolFlag{
-					Name:  FlagSkipSignalReapply,
-					Usage: "whether or not skipping signals reapply after the reset point",
-				},
-
 				cli.StringFlag{
 					Name: FlagEarliestTimeWithAlias,
 					Usage: "EarliestTime of decision start time, required for resetType of DecisionCompletedTime." +
@@ -272,6 +267,10 @@ func newWorkflowCommands() []cli.Command {
 						"time range (N<duration>), where 0 < N < 1000000 and duration (full-notation/short-notation) can be second/s, " +
 						"minute/m, hour/h, day/d, week/w, month/M or year/y. For example, '15minute' or '15m' implies last 15 minutes, " +
 						"meaning that workflow will be reset to the first decision that completed in last 15 minutes.",
+				},
+				cli.BoolFlag{
+					Name:  FlagSkipSignalReapply,
+					Usage: "whether or not skipping signals reapply after the reset point",
 				},
 			},
 			Action: func(c *cli.Context) {
--- a/tools/cli/workflow.go
+++ b/tools/cli/workflow.go
@@ -244,8 +244,9 @@ func newWorkflowCommands() []cli.Command {
 					Usage: "RunID, optional, default to the current/latest RunID",
 				},
 				cli.StringFlag{
-					Name:  FlagEventID,
-					Usage: "The eventID of any event after DecisionTaskStarted you want to reset to (exclusive). It can be DecisionTaskCompleted, DecisionTaskFailed or others",
+					Name: FlagEventID,
+					Usage: "The eventID of any event after DecisionTaskStarted you want to reset to (this event is exclusive in a new run. The new run " +
+						"history will fork and continue from the previous eventID of this). It can be DecisionTaskCompleted, DecisionTaskFailed or others",
 				},
 				cli.StringFlag{
 					Name:  FlagReason,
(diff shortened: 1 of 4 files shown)


tools/cli/workflowCommands.go
```diff
@@ -1792,12 +1798,59 @@ func getResetEventIDByType(ctx context.Context, c *cli.Context, resetType, domai
...
+		resp, err := frontendClient.GetWorkflowExecutionHistory(ctx, req)
+		if err != nil {
+			return "", 0, printErrorAndReturn("GetWorkflowExecutionHistory failed", err)
+		}
+		found := false
+		for _, e := range resp.GetHistory().GetEvents() {
+			if e.GetEventType() == shared.EventTypeDecisionTaskCompleted {
+				if e.GetTimestamp() >= earliestTime {
+					found = true
+					decisionFinishID = e.GetEventId()
```
yycptt, line 1835: nit: set decision eventID and break the outer loop directly instead of using `found`?

tools/cli/workflowCommands.go
```diff
@@ -1792,12 +1798,59 @@ func getResetEventIDByType(ctx context.Context, c *cli.Context, resetType, domai
...
+			return
+		}
 	default:
 		panic("not supported resetType")
 	}
 	return
 }
 
+func getEarliestDecisionID(ctx context.Context, domain string, wid string, rid string, earliestTime int64, frontendClient workflowserviceclient.Interface) (resetBaseRunID string, decisionFinishID int64, err error) {
+	resetBaseRunID = rid
```
yycptt, line 1814: nit: seems like `resetBaseRunID` is just `rid` in this case, why we still need to return it?

tools/cli/workflow.go
```diff
@@ -259,6 +259,14 @@ func newWorkflowCommands() []cli.Command {
 					Name:  FlagResetBadBinaryChecksum,
 					Usage: "Binary checksum for resetType of BadBinary",
 				},
+				cli.StringFlag{
+					Name: FlagEarliestTimeWithAlias,
+					Usage: "EarliestTime of decision start time, for resetType of DecisionStartedTime." +
+						"Supported formats are '2006-01-02T15:04:05+07:00', raw UnixNano and " +
+						"time range (N<duration>), where 0 < N < 1000000 and duration (full-notation/short-notation) can be second/s, " +
+						"minute/m, hour/h, day/d, week/w, month/M or year/y. For example, '15minute' or '15m' implies last 15 minutes, " +
+						"meaning that workflow will be reset to the first decision that started in last 15 minutes.",
```
yycptt, line 268: Is this a required option for DecisionStartedTime? It seems like if this flag is not set, this type is the same as FristDecisionCompleted？Shall we make it more clear in the comments?

longquanzheng: Yes this is a required option. 
https://github.com/uber/cadence/blob/a9337097fb6410818ccf02be99faf3f2b27a3f14/tools/cli/workflowCommands.go#L1449

https://github.com/uber/cadence/blob/a9337097fb6410818ccf02be99faf3f2b27a3f14/tools/cli/workflowCommands.go#L1540

We will check the extra option is existing for the particular resetType

tools/cli/workflowCommands.go
longquanzheng: Good catch, will remove it.

tools/cli/workflowCommands.go
```diff
@@ -1798,36 +1804,92 @@ func isLastEventDecisionTaskFailedWithNonDeterminism(ctx context.Context, domain
...
 	switch resetType {
-	case "LastDecisionCompleted":
+	case resetTypeLastDecisionCompleted:
 		resetBaseRunID, decisionFinishID, err = getLastDecisionCompletedID(ctx, domain, wid, rid, frontendClient)
 		if err != nil {
 			return
 		}
-	case "LastContinuedAsNew":
+	case resetTypeLastContinuedAsNew:
 		resetBaseRunID, decisionFinishID, err = getLastContinueAsNewID(ctx, domain, wid, rid, frontendClient)
```
yycptt, line 1824: nit: Seems like only this reset type will change baseRunID, shall we also update other util functions like `getLastDecisionCompletedID`, `getFirstDecisionCompletedID`, etc. not to return `resetBaseRunID`?

yycptt (approved):

tools/cli/workflowCommands.go
longquanzheng: yeah that sounds right thing to do.

PR #3673, opened 2020-10-20
Convert shardStore/manager to use internal types

anish531213:
#3612


common/persistence/persistenceInterface.go
```diff
@@ -798,24 +798,24 @@ type (
...
+		TimerAckLevel                 time.Time                         `json:"timer_ack_level"`
+		ClusterTransferAckLevel       map[string]int64                  `json:"cluster_transfer_ack_level"`
+		ClusterTimerAckLevel          map[string]time.Time              `json:"cluster_timer_ack_level"`
+		TransferProcessingQueueStates *types.ProcessingQueueStates      `json:"transfer_processing_queue_states"`
+		TimerProcessingQueueStates    *types.ProcessingQueueStates      `json:"timer_processing_queue_states"`
+		TransferFailoverLevels        map[string]TransferFailoverLevel  // uuid -> TransferFailoverLevel
+		TimerFailoverLevels           map[string]TimerFailoverLevel     // uuid -> TimerFailoverLevel
+		ClusterReplicationLevel       map[string]int64                  `json:"cluster_replication_level"`
+		DomainNotificationVersion     int64                             `json:"domain_notification_version"`
+		PendingFailoverMarkers        []*types.FailoverMarkerAttributes `json:"pending_failover_markers"`
```
yycptt, line 818: Why are we using the generated types here instead of `DataBlob` here? My understanding is that previously persistence interface layer and below is ignorant of any data serialization. 

If we do want to let the persistence layer handled the serialization, we should only modify `ShardInfo` in data interface layer instead of `InternalShardInfo` I think. Otherwise, all persistence layer will need the same logic for serializing and deserializing objects.

anish531213: My understanding is that we want to do serialization (thrift/proto) in the persistence layer. And above that, everything should use internal types. Right now, we are making sure that the persistence layer is receiving the internal types and serialization and deserialization only happens in the persistence layer and we have only changed manager till now. The manager is converting to internal types right now but as we will be pushing up the stack, the data interface layer will also change to use internal types.

anish531213: maybe @andrewjdawson2016 can explain this better.

yycptt: I think using internal types and serialization are two unrelated things. I am all for using internal types in data/persistence interface and I think the plan is to start from persistence interface and gradually push the boundary to application layer. I also understand that the `DataBlob` type we currently use is from thrift and we need to replace it with internal types. We do have an internal version of `DataBlob` we can use for this migration and all we need to do is copying over all the fields. For thrift types like `FailoverMarkerAttributes` or `Timer/TransferProcessingQueueStates`, they are only used in application layer and is target of the next step in migration. 

Now in terms of serialization, my biggest concern of doing serialization at the lowest persistence interface level (persistence store) is that each persistence implementation (cassandra, mysql, postgres, etc.) will have the exact same logic for serialization and deserialization, which is something we should try to avoid. We can either have all the code for serialization in data interface layer (persistence manager) or we can keep the logic in application as is, which will make later changes to those types easier since changes to those types will affects code in application layer, not in persistence layer. Either way, I don't see a strong reason to change the persistence store interface.

andrewjdawson2016: I see your point on this. If I understand your concern correctly you are worried about duplication of the serialization logic in each store implementation on the basis that these serializations will look identical. My claim is there is no reason they would have to look identical. How MySQL decides to serialize and store blobs has nothing to do with how Cassandra decides to do it. As long as each store implementation can implement the interface the format of storage/serialization does not matter to higher levels. 

My goal was to push the point of serialization to the thinnest possible portion of the stack - so basically we only serialize when we have to (e.g. right before enter/exit of the wire and right before enter/exit of the db). 

In order to address your duplication concern I think we can have common functions for the serialization logic that happens to be shared between these stores.

andrewjdawson2016: I do not see the internal types as decoupled form serialization. In order to get a datablob at the manager we have to serialize. So even if we convert to internal.DataBlob at the store level the serialization still would live in the manager.

PR #3674, opened 2020-10-21
Add correct license header to attribute to temporal and update license generation tool

andrewjdawson2016:


LICENSE
```diff
@@ -1,7 +1,3 @@
-The MIT License (MIT)
-
-Copyright (c) 2017-2020 Uber Technologies Inc.
-
```
yycptt, line 4: The LICENSE file is also used by the go generate command to generate mock files, will this change affect that?

andrewjdawson2016: good call I will check - thanks

yycptt (approved):

PR #3678, opened 2020-10-21
Notify new tasks when persistence operation timeout

yycptt:
**What changed?**
Notify new tasks to transfer/timer processor when persistence operation in shard context timeout

**Why?**
When operations in shard contexts timedout, we should still notify task processor, as we don't know if the operation succeeded or not.

Consider the following case: 
- User worker call respond decision task completed and want to schedule an activity. 
- History engine will update workflow's state and create an activity task. 
- When shard context tries to persist the update to DB, the operation timedout. Assume despite the timeout, the update actually got applied in DB. 
- Shard context will increase the rangeID to make sure the update will be seen on next read, and then return the timeout error to workflow execution context. 
- Workflow execution context will return the error directly without notifying the task processing logic. 
- Frontend or user got this timeout error and retry respond decision task completed, but the retry will also fail because the workflow state has been updated in the first call and the decision task no longer exists. 
- Now this activity task is persisted in DB but task processor not never notified, and the activity will get Schedule to Start timeout error. 

The same logic also applies to startworkflow as later retry will get workflow execution already started error.

**How did you test it?**
Integration test.

**Potential risks**
The only risk is that  notify task will trigger a DB load and result in more DB requests.


service/history/execution/context.go
```diff
@@ -415,14 +415,14 @@ func (c *contextImpl) CreateWorkflowExecution(
 
 	_, err := c.createWorkflowExecutionWithRetry(ctx, createRequest)
 	if err != nil {
+		if c.isPersistenceTimeoutError(err) {
```
yux0, line 418: Will it be a problem to notify task regardless error type?

yycptt: No, you can notify task processor as many times as you want. It will only trigger a DB load.

yux0 (approved):

PR #3687, opened 2020-10-23
reset workflow with no decision task complete

yux0:
**What changed?**
reset workflow with no decision task complete

**Why?**
https://github.com/uber/cadence/issues/2420

If there is no decision task completed, add the decision task time out in this case.

**How did you test it?**
local test

**Potential risks**
TODO


service/history/historyEngine.go
```diff
@@ -2458,6 +2458,11 @@ func (e *historyEngineImpl) ResetWorkflowExecution(
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := baseMutableState.GetInFlightDecision(); !ok && baseMutableState.IsWorkflowExecutionRunning() {
+		return nil, &workflow.BadRequestError{
+			Message: "Cannot reset open workflow without completed decision task.",
```
longquanzheng, line 2463: Is it accurate? -- A workflow can be open, but waiting for a timer to fire, so there is no in-flight decision, but it already have a completed decision. 

Also why we want to restrict this feature for only closed workflows?

service/history/reset/resetter.go
```diff
@@ -202,32 +202,32 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
 		return nil, err
 	}
 
-	// TODO add checking of reset until event ID == decision task started ID + 1
 	decision, ok := resetMutableState.GetInFlightDecision()
-	if !ok || decision.StartedID+1 != resetMutableState.GetNextEventID() {
-		return nil, &shared.BadRequestError{
-			Message: fmt.Sprintf("Can only reset workflow to DecisionTaskStarted + 1: %v", baseRebuildLastEventID+1),
+	if ok {
```
longquanzheng, line 206: Not sure this will work. InFlightDecision == HasDecisionStarted. 
So there is also case that decision scheduled but not started yet. So if you simply schedule another decision, I am not sure what will happen

longquanzheng: double check the code that MutableState allows calling scheduleDecisoin multiple times. So probably this is not an issue. But make sure you test this case, and strongly recommend you to add an integration test.

service/history/historyEngine.go
yycptt: I think using `GetPendingDecision()` is more appropriate here. 

@longquanzheng  Basically we want to block the case when the decision task is not even scheduled when a reset is triggered as if workflow history only contains started event, the reset is kind of like a no-op.

longquanzheng: yeah I understand. But doesn't seem necessary to have that. There could be case that we need it to mitigate some server issues. So it would be nice to not strict that IMO.

longquanzheng: GetPendingDecision() also have the same issue. If a workflow is waiting for a timer to fire, there is no decision scheduled until the timer fires. The workflow has no pending decision nor started decision, but it doesn't mean that the workflow hasn't completed a decision before.

yux0: The functions might have bad naming. From my understanding, getPending or getInFlight will return a decision task if there is completed one

service/history/reset/resetter.go
yux0: Yes. I am testing with a decision task timed out

service/history/historyEngine.go
yux0: Changed to HasProcessedOrPendingDecision()

service/history/execution/mutable_state_decision_task_manager.go
```diff
@@ -272,7 +278,47 @@ func (m *mutableStateDecisionTaskManagerImpl) AddDecisionTaskScheduleToStartTime
...
+		return nil, m.msb.createInternalServerError(opTag)
+	}
+
+	// Clear stickiness whenever decision fails
+	m.msb.ClearStickyness()
+
+	event := m.msb.hBuilder.AddDecisionTaskTimedOutEvent(
+		scheduleEventID,
+		0,
+		workflow.TimeoutTypeScheduleToStart,
```
longquanzheng, line 317: Should it be TimeoutTypeReset?

service/history/reset/resetter.go
```diff
@@ -202,32 +202,58 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
...
+		// reset workflow has decision task start
+		if decision.StartedID+1 != resetMutableState.GetNextEventID() {
+			return nil, &shared.BadRequestError{
+				Message: fmt.Sprintf("Can only reset workflow to DecisionTaskStarted + 1: %v", baseRebuildLastEventID+1),
```
longquanzheng, line 210: This error is no longer a bad request. ResetMutableState.GetNextEventID() must be equal to decision.StartedID+1. This is an internal mutableState corruption if not true.

service/history/reset/resetter.go
```diff
@@ -202,32 +202,58 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
...
-		execution.IdentityHistoryService,
-		resetReason,
-		"",
-		baseRunID,
-		resetRunID,
-		baseLastEventVersion,
-	)
-	if err != nil {
-		return nil, err
+	decision, ok = resetMutableState.GetPendingDecision()
```
longquanzheng, line 235: This whole block need to be in the `else` of the previous `resetMutableState.GetInFlightDecision()` check. Otherwise you will add decisionTaskFailed and then addDecisionTaskTimeout.

service/history/reset/resetter.go
```diff
@@ -202,32 +202,58 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
...
+				Message: fmt.Sprintf("Can only reset workflow to DecisionTaskStarted + 1: %v", baseRebuildLastEventID+1),
+			}
 		}
-	}
-	if len(resetMutableState.GetPendingChildExecutionInfos()) > 0 {
-		return nil, &shared.BadRequestError{
-			Message: fmt.Sprintf("Can only reset workflow with pending child workflows"),
+		if len(resetMutableState.GetPendingChildExecutionInfos()) > 0 {
+			return nil, &shared.BadRequestError{
+				Message: fmt.Sprintf("Can only reset workflow with pending child workflows"),
```
longquanzheng, line 215: typo: "Can not only reset workflow with pending child workflows"

service/history/reset/resetter.go
```diff
@@ -202,32 +202,58 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
...
+	if ok {
+		//reset workflow has decision task schedule
+		if decision.ScheduleID+1 != resetMutableState.GetNextEventID() {
+			return nil, &shared.BadRequestError{
+				Message: "Can only reset workflow to DecisionTaskScheduled.",
+			}
+		}
+		if len(resetMutableState.GetPendingChildExecutionInfos()) > 0 {
+			return nil, &shared.BadRequestError{
+				Message: fmt.Sprintf("Can only reset workflow with pending child workflows"),
```
longquanzheng, line 245: typo: "Can not only reset workflow with pending child workflows"

service/history/reset/resetter.go
```diff
@@ -202,32 +202,58 @@ func (r *workflowResetterImpl) prepareResetWorkflow(
...
-		baseLastEventVersion,
-	)
-	if err != nil {
-		return nil, err
+	decision, ok = resetMutableState.GetPendingDecision()
+	if ok {
+		//reset workflow has decision task schedule
+		if decision.ScheduleID+1 != resetMutableState.GetNextEventID() {
+			return nil, &shared.BadRequestError{
+				Message: "Can only reset workflow to DecisionTaskScheduled.",
```
longquanzheng, line 240: This is not necessary to be true, consider this:
```
1: WF start
2: Decision Scheduled
3: WF signaled
4 WF signaled
5: WF timeouted
```
People may want to reset to 4/5, as DecisionFinishEventID, == resetMutableState.GetNextEventID()

I think we can remove this check.

service/history/reset/resetter.go
```diff
@@ -289,16 +315,22 @@ func (r *workflowResetterImpl) persistToDB(
 	if err != nil {
 		return err
 	}
-	if len(resetWorkflowEventsSeq) != 1 {
+
+	resetHistorySize := int64(0)
+	switch len(resetWorkflowEventsSeq) {
+	case 0:
+		// reset workflow without decision task completes
```
longquanzheng, line 322: I don't get this, we still need to write DecisionTaskTImeout, why this is zero?

tools/cli/workflowCommands.go
```diff
@@ -1816,9 +1834,11 @@ func getLastDecisionCompletedID(ctx context.Context, domain, wid, rid string, fr
 			return "", 0, printErrorAndReturn("GetWorkflowExecutionHistory failed", err)
 		}
 		for _, e := range resp.GetHistory().GetEvents() {
-			if e.GetEventType() == shared.EventTypeDecisionTaskCompleted {
+			if e.GetEventType().Equals(decisionType) {
 				decisionFinishID = e.GetEventId()
+				return resetBaseRunID, decisionFinishID, nil
 			}
+			decisionFinishID = e.GetEventId()
```
longquanzheng, line 1841: Should it be `e.GetEventId()+1`?  this decisionFinishID is exclusive in reset.

service/history/execution/mutable_state_decision_task_manager.go
yux0: Added the reason field for reset

service/history/reset/resetter.go
yux0: removed

service/history/reset/resetter.go
yux0: good catch. Added it

service/history/reset/resetter.go
yux0: Updated

service/history/reset/resetter.go
yux0: removed

tools/cli/workflowCommands.go
yux0: updated

service/history/reset/resetter.go
yux0: removed

service/history/execution/mutable_state_decision_task_manager.go
longquanzheng: I actually meant cause == RESET. it's strongly typed and can be safely used for programming.

service/history/reset/resetter.go
```diff
@@ -632,3 +619,56 @@ func (r *workflowResetterImpl) getPaginationFn(
...
+	resetRunID string,
+	baseLastEventVersion int64,
+	resetReason string,
+) (execution.MutableState, error) {
+
+	if decision, ok := resetMutableState.GetInFlightDecision(); ok {
+		// reset workflow has decision task start
+		if len(resetMutableState.GetPendingChildExecutionInfos()) > 0 {
+			return nil, &shared.BadRequestError{
+				Message: fmt.Sprintf("Can not only reset workflow with pending child workflows"),
```
longquanzheng, line 635: typo: remove `only`

service/history/reset/resetter.go
```diff
@@ -632,3 +619,56 @@ func (r *workflowResetterImpl) getPaginationFn(
...
+			resetRunID,
+			baseLastEventVersion,
+		)
+		if err != nil {
+			return nil, err
+		}
+	} else if decision, ok := resetMutableState.GetPendingDecision(); ok {
+		if ok {
+			//reset workflow has decision task schedule
+			if len(resetMutableState.GetPendingChildExecutionInfos()) > 0 {
```
longquanzheng, line 656: Nit: maybe move it to the beginning of the method.

longquanzheng:
Almost lgtm. Can you also add a test case in https://github.com/uber/cadence/blob/master/host/resetworkflow_test.go ?

service/history/reset/resetter.go
yux0: updated

service/history/reset/resetter.go
yux0: updated

service/history/reset/resetter.go
yux0: updated

service/history/execution/mutable_state_decision_task_manager.go
yux0: updated

longquanzheng (approved):

common/types/shared.go
```diff
@@ -8521,30 +8617,30 @@ func (e TimeoutType) String() string {
...
+		return "StartToClose"
 	case 1:
-		return "SCHEDULE_TO_START"
+		return "ScheduleToStart"
 	case 2:
-		return "SCHEDULE_TO_CLOSE"
+		return "ScheduleToClose"
 	case 3:
-		return "HEARTBEAT"
+		return "Heartbeat"
```
yycptt, line 8626: We can't simply change this representation, as it will be used as the timeout error reason. User needs to type the exact same string when specifying non-retryable errors. See https://github.com/uber/cadence/blob/master/service/history/execution/timer_sequence.go#L440

yycptt (changes requested):

tools/cli/workflowCommands.go
```diff
@@ -1904,60 +1904,41 @@ func getResetEventIDByType(
...
-	if decisionFinishID == 0 {
-		return 0, printErrorAndReturn("Get DecisionFinishID failed", fmt.Errorf("no DecisionFinishID"))
-	}
-	return
-}
+	decisionType types.EventType,
+) (resetBaseRunID string, decisionFinishID int64, err error) {
 
-func getLastDecisionCompletedID(ctx context.Context, domain, wid, rid string, frontendClient frontend.Client) (decisionFinishID int64, err error) {
+	resetBaseRunID = runID
```
yycptt, line 1936: I think this return value was remove in a previous diff as it's value is always equal to the input runID. Any reason for adding it back?

yycptt (approved):

tools/cli/workflowCommands.go
yux0: You are correct. It reverted during conflict resolution. I updated it

PR #3695, opened 2020-10-27
Thrift clients

vytautas-karpavicius:
**What changed?**
- Explicitly listed `Client` interfaces instead of embedding thrift client interfaces.
- Created `thriftClient` wrappers. Currently they only forward the request as is.

**Why?**
This is prep-work for introducing internal application types. For now this change simply adds no-op thrift wrappers. Later on they will become a place for internal types to thrift conversion. Explicitly listed `Client` interfaces will operate on on internal types and only at the thriftClient they will be converted to thrift counterparts.

**How did you test it?**
- Existing tests unit tests.
- Run canary locally.

**Potential risks**

### DIFF
--- a/client/admin/interface.go
+++ b/client/admin/interface.go
@@ -20,9 +20,34 @@
 
 package admin
 
-import "github.com/uber/cadence/.gen/go/admin/adminserviceclient"
+import (
+	"context"
+
+	"go.uber.org/yarpc"
+
+	"github.com/uber/cadence/.gen/go/admin"
+	"github.com/uber/cadence/.gen/go/replicator"
+	"github.com/uber/cadence/.gen/go/shared"
+)
 
 // Client is the interface exposed by admin service client
 type Client interface {
-	adminserviceclient.Interface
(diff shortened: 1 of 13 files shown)


client/admin/interface.go
```diff
@@ -20,9 +20,34 @@
...
+
+	"github.com/uber/cadence/.gen/go/admin"
+	"github.com/uber/cadence/.gen/go/replicator"
+	"github.com/uber/cadence/.gen/go/shared"
+)
 
 // Client is the interface exposed by admin service client
 type Client interface {
-	adminserviceclient.Interface
+	AddSearchAttribute(context.Context, *admin.AddSearchAttributeRequest, ...yarpc.CallOption) error
```
yux0, line 35: qq: why we will not use the embedded thrift interface?

vytautas-karpavicius: This was changed on purpose so that we can change method signatures to accept and return internal types instead of thrift counterparts. Thrift types will be only be used in thrift clients and handlers. These changes will come with upcoming PRs.

PR #3697, opened 2020-10-27
Service busy error will not put replication task into DLQ

yux0:
**What changed?**
Service busy error will not put replication task into DLQ

**Why?**
when encountering service busy error, do retry and don't put it into DLQ

**How did you test it?**
cluster testing 

**Potential risks**


service/history/replication/task_processor.go
```diff
@@ -393,21 +393,22 @@ func (p *taskProcessorImpl) processSingleTask(replicationTask *r.ReplicationTask
...
-			p.logger.Warn("Skip adding new messages to DLQ.", tag.Error(err))
-		default:
-			p.logger.Error(
-				"Failed to apply replication task after retry. Putting task into DLQ.",
-				tag.TaskID(replicationTask.GetSourceTaskId()),
-				tag.Error(err),
-			)
-			return p.putReplicationTaskToDLQ(replicationTask)
-		}
+	if _, ok := err.(*shared.ServiceBusyError); err == nil || ok {
```
yycptt, line 396: nit: use `common.IsServiceBusyError`?

yycptt (approved):
Shall we update the PR title to be more specific?

service/history/replication/task_processor.go
yux0: updated

PR #3699, opened 2020-10-27
Improve Error Handling For Cassandra Persistence Implementation

yycptt:
**What changed?**
- Check if `err == context.DeadlineExceeded` when checking for timeout error
- Replace duplicated error handling logic with a common error conversion function, which can return more detailed error types instead of just InternalServiceError.

**Why?**
Improve error handling

**How did you test it?**
Tested locally

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/db.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/db.go
@@ -81,16 +81,9 @@ func (db *cdb) IsTimeoutError(err error) bool {
 }
 
 func (db *cdb) IsThrottlingError(err error) bool {
-	if req, ok := err.(gocql.RequestError); ok {
-		// gocql does not expose the constant errOverloaded = 0x1001
-		return req.Code() == 0x1001
-	}
-	return false
+	return IsThrottlingError(err)
 }
 
 func (db *cdb) IsConditionFailedError(err error) bool {
-	if err == errConditionFailed {
-		return true
-	}
-	return false
+	return IsConditionFailedError(err)
 }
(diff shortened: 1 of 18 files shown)


common/persistence/nosql/nosqlplugin/cassandra/queue.go
```diff
@@ -183,7 +183,7 @@ func (db *cdb) InsertQueueMetadata(
 	query := db.session.Query(templateInsertQueueMetadataQuery, queueType, clusterAckLevels, version).WithContext(ctx)
 	_, err := query.ScanCAS()
 	if err != nil {
-		return fmt.Errorf("failed to insert initial queue metadata record: %v, Type: %v", err, queueType)
+		return err
```
longquanzheng, line 186: Nit: I thought we want to keep some information about the error like queueType, why should we remove the wrapping?

common/persistence/nosql/nosqlplugin/cassandra/db.go
```diff
@@ -79,14 +77,7 @@ func (db *cdb) IsNotFoundError(err error) bool {
...
 func (db *cdb) IsTimeoutError(err error) bool {
-	if err == gocql.ErrTimeoutNoResponse {
-		return true
-	}
-	if err == gocql.ErrConnectionClosed {
-		return true
-	}
-	_, ok := err.(*gocql.RequestErrWriteTimeout)
-	return ok
+	return IsTimeoutError(err)
```
longquanzheng, line 80: Nit: it's good to centralize the logic. Could you do the same thing for the other two?

longquanzheng (approved):
lgtm, only some nit comments

common/persistence/nosql/nosqlplugin/cassandra/queue.go
yycptt: The error information is still preserved in the caller of this function where we do error conversion. Return raw error here so that the caller in cassandraQueue.go can do the conversion, otherwise it won't be able to tell the difference between errors.

common/persistence/nosql/nosqlplugin/cassandra/db.go
yycptt: Sure. Forgot to do the same thing for the other two. Will update.

yycptt: Done.

PR #3705, opened 2020-10-28
Make id length dynamic config per domain

andrewjdawson2016:
Make dynamic config to control id length limit filtered by domain name. 

So that we can provide domains exception as we reduce the limits.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -447,14 +447,14 @@ func (wh *WorkflowHandler) PollForActivityTask(
 		return nil, wh.error(errDomainNotSet, scope)
 	}
 
-	if !wh.validIDLength(pollRequest.GetDomain(), scope) {
+	if !wh.validIDLength(pollRequest.GetDomain(), scope, pollRequest.GetDomain()) {
 		return nil, wh.error(errDomainTooLong, scope)
 	}
 
-	if err := wh.validateTaskList(pollRequest.TaskList, scope); err != nil {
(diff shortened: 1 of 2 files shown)


service/frontend/workflowHandler.go
```diff
@@ -447,14 +447,14 @@ func (wh *WorkflowHandler) PollForActivityTask(
 		return nil, wh.error(errDomainNotSet, scope)
 	}
 
-	if !wh.validIDLength(pollRequest.GetDomain(), scope) {
+	if !wh.validIDLength(pollRequest.GetDomain(), scope, pollRequest.GetDomain()) {
```
meiliang86, line 450: I would expect the check is only enforced upon creation (i.e. start workflow/create domain, etc)? Otherwise there's nothing user can do to fix this?

andrewjdawson2016: We do check on workflow start. The check in PollActivityTask should never fail. We can delete all nop checks if you want. But I don't see a very compelling reason to change it from how its already been.

andrewjdawson2016: I think I have a slight preference for having validation logic in all APIs instead of having APIs have their validation logic depend on which other APIs we know have already run. I find this simpler.

meiliang86: ok. For now let's add a comment in all the non-start method and leave the changes there.

meiliang86 (approved):

service/frontend/workflowHandler.go
andrewjdawson2016: sounds good - will do.

PR #3711, opened 2020-10-30
Frontend client internal types

vytautas-karpavicius:
**What changed?**
Switch frontend client to internal types. Changed immediate client usage by passing already constructed types or using thrift mappers at the call boundary. Upcoming PRs will propagate those conversions further up, until internal types reaches handlers eventually.

**Why?**

**How did you test it?**
Existing tests.

**Potential risks**


service/frontend/dcRedirectionHandler_test.go
```diff
@@ -24,12 +24,15 @@ import (
 	"context"
 	"testing"
 
+	"github.com/uber/cadence/client/frontend"
+	"github.com/uber/cadence/common/types"
+	"github.com/uber/cadence/common/types/mapper/thrift"
```
yycptt, line 29: nit: please group with other cadence imports

yycptt (approved):

service/worker/batcher/workflow.go
```diff
@@ -367,27 +367,27 @@ func startTaskProcessor(
 				err = processTask(ctx, limiter, task, batchParams, client,
 					batchParams.CancelParams.CancelChildren,
 					func(workflowID, runID string) error {
-						return client.RequestCancelWorkflowExecution(ctx, &shared.RequestCancelWorkflowExecutionRequest{
+						return client.RequestCancelWorkflowExecution(ctx, &types.RequestCancelWorkflowExecutionRequest{
```
yycptt, line 370: I am not entirely sure if we need to convert the error. Seems like EntityNotExist is the only error we care about and it's already updated? Same for terminate and signal?

service/frontend/dcRedirectionHandler_test.go
vytautas-karpavicius: Done.

service/worker/batcher/workflow.go
vytautas-karpavicius: We don't need that here. Both batcher and failovermanager workers were simple enough to update in one shot. Thus no thrift related types (including errors) left there.
This particular error returned here is checked against `*types.EntityNotExistsError` just as you mentioned, which is already converted as well.

PR #3713, opened 2020-10-30
Matching client internal types

vytautas-karpavicius:
**What changed?**
Switch matching client to internal types. Changed immediate client usage by passing already constructed types or using thrift mappers at the call boundary. Upcoming PRs will propagate those conversions further up, until internal types reaches handlers eventually.

**Why?**

**How did you test it?**
Existing tests.

**Potential risks**


andrewjdawson2016 (approved):
This looks good to me. Please locally try to introduce some random bugs in the translation and make sure unit tests catch these bugs. If unit tests not catch then please extend unit tests.

PR #3721, opened 2020-11-01
Don't retry when reset to a middle of a batch to save perf

longquanzheng:
**What changed?**
Change the error return from StateBuilder from InternalServiceError to BadRequestError

**Why?**
For https://github.com/uber/cadence/issues/3720
InternalServiceError in harmful because FrontendService will keep retrying until context timeouts. 
Reset relys on replaying, which is very expensive and potentially making down the history service.  

**How did you test it?**
Local tested.

**Potential risks**
StateBuilder is also being used by NDC replication. I believe returning BadRequestError should also work(better) fro NDC. 
@yux0

### DIFF
--- a/service/history/execution/state_rebuilder.go
+++ b/service/history/execution/state_rebuilder.go
@@ -162,8 +162,10 @@ func (r *stateRebuilderImpl) Rebuild(
 		baseLastEventID,
 		baseLastEventVersion,
 	)) {
-		return nil, 0, &shared.InternalServiceError{Message: fmt.Sprintf(
-			"nDCStateRebuilder unable to rebuild mutable state to event ID: %v, version: %v",
+		return nil, 0, &shared.BadRequestError{Message: fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


service/history/execution/state_rebuilder.go
```diff
@@ -162,8 +162,10 @@ func (r *stateRebuilderImpl) Rebuild(
 		baseLastEventID,
 		baseLastEventVersion,
 	)) {
-		return nil, 0, &shared.InternalServiceError{Message: fmt.Sprintf(
-			"nDCStateRebuilder unable to rebuild mutable state to event ID: %v, version: %v",
+		return nil, 0, &shared.BadRequestError{Message: fmt.Sprintf(
```
yux0, line 165: It makes sense to be a non retryable error. I understand the use of a bad request here but the bad request is more for user request. Could you add a TODO to mention have retryable error and non-retryable error?

yux0 (approved):

yycptt (approved):

PR #3722, opened 2020-11-02
Update sqlx dependency to get rid of hacking fix

longquanzheng:
**What changed?**

**Why?**
for https://github.com/uber/cadence/issues/2863
This will make people's life easier when depending on cadence server as library. They don't have to repeat the override(replace) command.

**How did you test it?**
Existing tests

**Potential risks**
It should be minimal. sqlx is a wrapper library, usually a bug is already shown up if any during the tests.


yycptt (commented):
Do we need the same changes as this diff?: https://github.com/temporalio/temporal/pull/924/files

longquanzheng:
> Do we need the same changes as this diff?: https://github.com/temporalio/temporal/pull/924/files

No we don't have that legacy code about BindNamed. It's their own code change no in Cadence. 
see https://github.com/uber/cadence/blob/master/common/persistence/sql/sqlplugin/mysql/execution_maps.go#L128

yycptt (approved):

PR #3723, opened 2020-11-03
Domains manager and store serialization refactor

andrewjdawson2016:


common/util.go
```diff
@@ -792,3 +792,13 @@ func IsStickyTaskConditionError(err error) bool {
 	}
 	return false
 }
+
+// DurationToInt32 converts duration to int32 at the specified unit
+func DurationToInt32(d time.Duration, unit time.Duration) int32 {
```
vytautas-karpavicius, line 797: Maybe we could have several helpers like `DurationToSeconds`, `DurationToDays`, etc instead?

andrewjdawson2016: Addressed

emrahs (approved):

PR #3727, opened 2020-11-04
Various fixes and improvements for reset and retry error, add integration test for reset

longquanzheng:
**What changed?**
1. Fix reset failure when current run is open. It's the same issue that I should have fixed in https://github.com/uber/cadence/pull/3500 (sorry for missing this case). 
2. Add reset integration test for 3 cases: reset open, reset closed and reset of reset. For https://github.com/uber/cadence/issues/1688 
3. Dumping history is making the buildkite output really large. Disable dumpHistory flag in all integ test. I don't think anyone is looking at that. It's only useful when reproducing locally for debugging. But LMK if I am wrong. 
4. Improve error handling for retry policy. After introducing context, we need to return the previous error if exists instead of the last error. This will significantly improve our life when troubleshooting. 

**Why?**
1. reset is broken
2. we have been missing reset integ for a while(even though we have canary)
3. see above explanation. 

**How did you test it?**
local test, unit test , integ test.

before any fixes:
```
[qlong] ~/cadence [qlong-reset-integ-test] M?? % ./cadence --ct 30 --do qlong wf reset -w cancel_9db64e27-3882-4854-bb27-45ed898971e2 --event_id 4 --reason test
Error: reset failed
Error Details: code:deadline-exceeded message:timeout
```

after the retry fix:
```
[qlong] ~/cadence [qlong-reset-integ-test] M?? % ./cadence --ct 30 --do qlong wf reset -w cancel_494d32a1-122d-4619-9254-2925143cab48  --event_id 4 --reason test

Error: reset failed
Error Details: code:internal message:cadence internal error, msg: AppendHistoryNodes: pq: duplicate key value violates unique constraint "history_tree_pkey"
('export CADENCE_CLI_SHOW_STACKS=1' to see stack traces)
```

after the reset fix:
```
[qlong] ~/cadence [qlong-reset-integ-test] M?? % ./cadence --do qlong wf list --open
        WORKFLOW TYPE       |                 WORKFLOW ID                 |                RUN ID                | TASK LIST | START TIME | EXECUTION TIME
  main.sampleCancelWorkflow | cancel_d32a3a20-11c0-4459-822a-b43a7b3b8739 | 40f1238f-184b-4e4f-9520-7ae5f68177c8 |           | 18:11:59   | 18:11:59
[qlong] ~/cadence [qlong-reset-integ-test] M?? % ./cadence --ct 30 --do qlong wf reset -w cancel_d32a3a20-11c0-4459-822a-b43a7b3b8739 --event_id 4 --reason test
{
  "runId": "e5fcdca2-4be5-4cdd-8156-a47c3d557139"
}
```

**Potential risks**
no.

### DIFF
--- a/common/backoff/retry.go
+++ b/common/backoff/retry.go
@@ -89,23 +89,35 @@ func NewConcurrentRetrier(retryPolicy RetryPolicy) *ConcurrentRetrier {
 }
 
 // Retry function can be used to wrap any call with retry logic using the passed in policy
+// The returned error will be preferred to a previous one if one exists. That's because the
+// very last error is very likely a timeout error, and it's not useful for logging/troubleshooting
 func Retry(operation Operation, policy RetryPolicy, isRetryable IsRetryable) error {
+	var prevErr error
 	var err error
 	var next time.Duration
 
 	r := NewRetrier(policy, SystemClock)
 	for {
+		// record the previous error before an operation
+		prevErr = err
+
 		// operation completed successfully.  No need to retry.
 		if err = operation(); err == nil {
 			return nil
 		}
 
 		if next = r.NextBackOff(); next == done {
+			if prevErr != nil {
+				return prevErr
+			}
 			return err
 		}
 
 		// Check if the error is retryable
 		if isRetryable != nil && !isRetryable(err) {
+			if prevErr != nil {
+				return prevErr
+			}
 			return err
 		}
 
--- a/common/backoff/retry_test.go
+++ b/common/backoff/retry_test.go
@@ -105,6 +105,8 @@ func (s *RetrySuite) TestRetryFailed_ReturnPreviousError() {
 
 	policy := NewExponentialRetryPolicy(1 * time.Millisecond)
 	policy.SetMaximumInterval(5 * time.Millisecond)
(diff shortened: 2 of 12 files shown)


common/backoff/retry.go
```diff
@@ -89,23 +89,35 @@ func NewConcurrentRetrier(retryPolicy RetryPolicy) *ConcurrentRetrier {
...
+			if prevErr != nil {
+				return prevErr
+			}
 			return err
 		}
 
 		// Check if the error is retryable
 		if isRetryable != nil && !isRetryable(err) {
+			if prevErr != nil {
+				return prevErr
```
yux0, line 119: This could be confused to the caller if the err is non retryable but the returned prev error is retryable

longquanzheng: Retryable error could also be returned, if that's the last attempt. So it shouldn't be confusing.
We did the samething on the client side:https://github.com/uber-go/cadence-client/pull/878 
Because the last error usually is just timeout error because of maxing out the timeout value. It's useless.

longquanzheng: Essentially, operations with retries got a lot of errors returned, and we need to decide which one. Technically anyone should make sense. Returning last one is the simplest approach, but not useful for handling and logging.

yux0 (approved):

PR #3730, opened 2020-11-05
History persistence store manager types

andrewjdawson2016:

### DIFF
--- a/common/persistence/historyStore.go
+++ b/common/persistence/historyStore.go
@@ -32,6 +32,7 @@ import (
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/service/dynamicconfig"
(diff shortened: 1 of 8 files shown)


common/persistence/historyStore.go
```diff
@@ -169,8 +171,12 @@ func (m *historyV2ManagerImpl) GetHistoryTree(
 	if err != nil {
 		return nil, err
 	}
+	var branches []*workflow.HistoryBranch
```
vytautas-karpavicius, line 174: nit: initialize with known capacity to avoid resizing - `make([]*workflow.HistoryBranch, 0, len(resp.Branches))`

vytautas-karpavicius (approved):

common/persistence/historyStore.go
andrewjdawson2016: This will not handle nil correctly without adding another check.

PR #3734, opened 2020-11-05
Persistence Error Injection Client

yycptt:
**What changed?**
Add an error injection client for persistence layer

**Why?**
Injection random persistence error during bench test to expose potential error handling bugs.

**How did you test it?**
Tested locally using canary with 0.001 error rate

**Potential risks**
No risk. Error injection client won't be created if error rate is 0 on service start up.


common/persistence/client/factory.go
```diff
@@ -169,6 +171,9 @@ func (f *factoryImpl) NewShardManager() (p.ShardManager, error) {
 		return nil, err
 	}
 	result := p.NewShardManager(store)
+	if errorRate := f.config.ErrorInjectionRate(); errorRate != 0 {
+		result = p.NewShardPersistenceErrorInjectionClient(result, errorRate, f.logger)
```
longquanzheng, line 175: How about passing in `f.config.ErrorInjectionRate`? since it's dynamic config we should take the advantage. Otherwise may just use static config then.

yycptt: I was worry about memory issue by querying dynamic config value too often. I retied once on staging by reading a dynamicconfig value once per second for each shard and it resulted some memory issues (GC can't keep up). For persistence client I think the rate will be even higher so would like to keep it safe.

Also I think we will never use this error injection client in production so the value won't be changed, and we can have one less wrapper around persistence manger. The only place it will be used is in our testing environment, where I think restarting the cluster is acceptable.

longquanzheng: Makes sense to me. I am fine with that.

common/persistence/persistenceErrorInjectionClients.go
```diff
@@ -0,0 +1,1686 @@
...
+}
+
+func (p *queueErrorInjectionPersistenceClient) Close() {
+	p.persistence.Close()
+}
+
+func shouldForwardCallToPersistence(
+	err error,
+) bool {
+	return err == nil || err == ErrFakeTimeout || err == ErrFakeUnhandled
```
yux0, line 1675: For those errors, there should be a rate that some calls will forward to DB and some won't?

common/persistence/persistenceErrorInjectionClients.go
```diff
@@ -0,0 +1,1686 @@
...
+}
+
+func generateFakeError(
```
yux0, line 1678: Different APIs could return some different errors. Should we also consider this? Currently, it returns some generic errors. Will we consider to expand the error type to workflowalreadystarted/conditionupdatefailure etc.?

common/persistence/persistenceErrorInjectionClients.go
```diff
@@ -0,0 +1,1686 @@
...
+) error {
+	fakeErr := generateFakeError(p.errorRate)
+
+	var persistenceErr error
+	if shouldForwardCallToPersistence(fakeErr) {
+		persistenceErr = p.persistence.CreateShard(ctx, request)
+	}
+
+	if fakeErr != nil {
+		p.logger.Error(msgInjectedFakeErr, tag.Error(fakeErr))
```
yux0, line 216: also include the API name and persistenceErr if not nil?

common/persistence/persistenceErrorInjectionClients.go
yycptt: I went through all the error types we defined and I don't think we can randomly return those errors (like entity not exist, workflow already started, domain not active, conditional update failed, etc) without checking the actual state in DB. Those errors are typically non-retryable and will cause the load we are running to fail.

common/persistence/persistenceErrorInjectionClients.go
yycptt: Sure, will update.

common/persistence/persistenceErrorInjectionClients.go
yycptt: Makes sense. I will update the function to return `false` with 50% chance if err is timeout or unhandled.

common/persistence/persistenceErrorInjectionClients.go
yycptt: Updated the log message to include the name of the operation, whether the call is forwarded and the persistenceErr (even if the error is nil, the information is still useful).

All store operations now have pre-defined tags in `log/tag/values.go`

common/persistence/persistenceErrorInjectionClients.go
yycptt: Updated.

yux0 (approved):

PR #3750, opened 2020-11-10
Improve logging when loading a dynamic config

longquanzheng:
**What changed?**
Use info log level for first time loading a dynamic config

**Why?**
It will be useful for debugging dynamic config. Because default logging level is Info in Cadence.
This would cause too many logging becuase it only emit the logs for the firs time loading.

**How did you test it?**
N/A

**Potential risks**
No

### DIFF
--- a/common/service/dynamicconfig/config.go
+++ b/common/service/dynamicconfig/config.go
@@ -44,7 +44,7 @@ func NewCollection(
 	return &Collection{
 		client:        client,
 		logger:        logger,
-		keys:          &sync.Map{},
+		logKeys:       &sync.Map{},
 		errCount:      -1,
 		filterOptions: filterOptions,
 	}
@@ -56,7 +56,7 @@ func NewCollection(
 type Collection struct {
 	client        Client
 	logger        log.Logger
-	keys          *sync.Map // map of config Key to strongly typed value
+	logKeys       *sync.Map // map of config Keys for logging to capture changes
 	errCount      int64
 	filterOptions []FilterOption
 }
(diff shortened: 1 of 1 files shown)


common/service/dynamicconfig/config.go
```diff
@@ -75,9 +75,10 @@ func (c *Collection) logValue(
 	cmpValueEquals func(interface{}, interface{}) bool,
 ) {
 	loadedValue, loaded := c.keys.LoadOrStore(key, value)
-	if !loaded || !cmpValueEquals(loadedValue, value) {
-		c.logger.Debug("Get dynamic config",
-			tag.Name(key.String()), tag.Value(value), tag.DefaultValue(defaultValue))
+	// If not loaded before, then print the log for the first loaded 
+	if !loaded {
```
yycptt, line 79: I am a little bit confused by the logic here. It seems like we want to emit a log when the value is first loaded and when the value changes. But the values in `keys` was never updated after the first load, so we will keep printing logs if the values changes. If we fix that, we should be able to emit log in both cases without generating too many logs?

longquanzheng: I misunderstood the code here because of the names. Just updated the PR, does it make more sense to you now?

yycptt: 👍  Yup, looks good to me! 

nit: Log messages should begin with capital letter?

yycptt (approved):

common/service/dynamicconfig/config.go
longquanzheng: yeah for sure. Thank you so much for reviewing it.

PR #3757, opened 2020-11-14
Improve logging for dynamic config

longquanzheng:
**What changed?**
Change the Warn from Info for key not found when reading dynamic config

**Why?**
It's noisy to report the warnings. And it's not accurate to use Warn for the case that config is not set. Afterall using default values is expected to do. 

**How did you test it?**
Existing tests.

**Potential risks**
No.

### DIFF
--- a/common/service/dynamicconfig/clientInterface.go
+++ b/common/service/dynamicconfig/clientInterface.go
@@ -24,6 +24,8 @@ package dynamicconfig
 
 import (
 	"time"
+
+	"github.com/uber/cadence/.gen/go/shared"
 )
 
 // Client allows fetching values from a dynamic configuration system NOTE: This does not have async
@@ -44,4 +46,8 @@ type Client interface {
 	) (time.Duration, error)
 	// UpdateValue takes value as map and updates by overriding. It doesn't support update with filters.
 	UpdateValue(name Key, value interface{}) error
+}
+
+var notFoundError = &shared.EntityNotExistsError{
+	Message: "unable to find key",
 }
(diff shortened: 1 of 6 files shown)


common/service/dynamicconfig/config.go
```diff
@@ -26,6 +26,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/uber/cadence/.gen/go/shared"
+
```
yycptt, line 30: nit: remove empty line and use internal types

common/service/dynamicconfig/clientInterface.go
```diff
@@ -45,3 +47,7 @@ type Client interface {
 	// UpdateValue takes value as map and updates by overriding. It doesn't support update with filters.
 	UpdateValue(name Key, value interface{}) error
 }
+
+var notFoundError = &shared.EntityNotExistsError{
```
yycptt, line 51: As we are migrating to internal types. Can we use types defined in `common/types` instead?

longquanzheng: Yeah good point. almost forgot that.

common/service/dynamicconfig/clientInterface.go
```diff
@@ -23,6 +23,7 @@
 package dynamicconfig
 
 import (
+	"github.com/uber/cadence/common/types"
 	"time"
```
yycptt, line 27: nit: `make fmt`

yycptt (approved):

PR #3759, opened 2020-11-16
Switch errors to internal types

vytautas-karpavicius:
**What changed?**
- Switched all thrift errors in cadence server to internal types.
- All intermediate `thrift.FromError` mappers were removed. They are only kept within handlers where errors are converted back to thrift.

**Why?**

I have decided to change all error types at once, after several other attempts to partition them. Otherwise it become very difficult to reason about such changes, as errors of particular type are often expected somewhere upwards. Unlike switching regular types compiler does not help here as all errors share the same interface. One have to be very careful to not forget adding `thrift.FromError` mappers if going by some partitioned approach. 

**How did you test it?**
Existing unit tests and integration tests.

With IDE support, ensured that thrift types are not used anywhere, except:
- Handler unit tests
- Parts that interact via thrift generated client:
  - CLI
  - Integration tests

**Potential risks**


common/types/errors.go
```diff
@@ -61,11 +64,15 @@ func (err DomainNotActiveError) Error() string {
 }
 
 func (err EntityNotExistsError) Error() string {
-	return fmt.Sprintf("EntityNotExistsError{Message: %v, CurrentCluster: %v, ActiveCluster: %v}",
```
andrewjdawson2016, line 64: So in the thrift error type were these not nullable?

common/types/errors.go
```diff
@@ -89,32 +96,59 @@ func (err RemoteSyncMatchedError) Error() string {
...
 	return fmt.Sprintf("ServiceBusyError{Message: %v}", err.Message)
 }
 
 func (err WorkflowExecutionAlreadyStartedError) Error() string {
-	return fmt.Sprintf("WorkflowExecutionAlreadyStartedError{Message: %v, StartRequestId: %v, RunId: %v}",
-		err.Message,
-		err.StartRequestID,
-		err.RunID,
-	)
+	fields := []string{}
```
andrewjdawson2016, line 130: What about using a string buffer instead of an array - I think it will be more efficient.

andrewjdawson2016 (approved):

common/types/errors.go
vytautas-karpavicius: Updated.

common/types/errors.go
vytautas-karpavicius: Those were nullable int thrift. However `Error()` methods are not generated and initial commited version was simplified and not included optional printing of nil fields.

PR #3771, opened 2020-11-19
Create domain filter for fixer workflow

andrewjdawson2016:
This enables controlling which domains get fixed by fixer workflow.


mantas-sidlauskas (approved):
looks good

PR #3775, opened 2020-11-19
Convert execution persistence types to use time.Time, time.Duration and internal types

andrewjdawson2016:
This removes anything that has to do with serialization from the persistenceInterface.go level in executions.


vytautas-karpavicius (approved):
LGTM

PR #3776, opened 2020-11-20
Set default MaximumSignalsPerExecution to 10K

longquanzheng:
**What changed?**
Set default MaximumSignalsPerExecution to 10K

**Why?**
If keep receiving signals in very high rate, existing behavior will keep failing decision and workflow can't do continueAsNew, because continueAsNew is not allowed when there is any bufferedEvents. So this can make an infinite loop:
1. workflow worker has started a decision because there is some signals
2. incoming signals go to buffered events
3. workflow made a decision to do continueAsNew because the total signals. has reached a big number(like 1K)
4. RespondDecisionTaskComplete failed and it becomes decisionTaskFailed, because there is buffered events. 
6. A new decision task is scheduled...go to 1 again. 

Eventually the workflow will get terminated when exceeding 200K event limit. This was design to protect this scenario. 
10K should be high even given our history length limit of 200K.

Related to discussion in https://uber-cadence.slack.com/archives/CL22WDF70/p1605881855052000?thread_ts=1605767738.022500&cid=CL22WDF70

**How did you test it?**
existing tests. 

**Potential risks**
Internally there should be set to some value.
This is to protect the external customers because mostly people rely on the default values.

### DIFF
--- a/service/history/config/config.go
+++ b/service/history/config/config.go
@@ -406,7 +406,7 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
 		ExecutionMgrNumConns:            dc.GetIntProperty(dynamicconfig.ExecutionMgrNumConns, 50),
 		HistoryMgrNumConns:              dc.GetIntProperty(dynamicconfig.HistoryMgrNumConns, 50),
 		MaximumBufferedEventsBatch:      dc.GetIntProperty(dynamicconfig.MaximumBufferedEventsBatch, 100),
-		// 10K signals should big enough given workflow has 200K history lengh limit. It needs to be non-zero to protect continueAsNew from infinit loop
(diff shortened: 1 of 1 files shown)


yycptt (approved):
The change looks good to me. But I think our some of our production clusters are relying on this default value defined in the code so not sure if this will break any workflows in production. We'll need to be careful when releasing this change.

longquanzheng:
> The change looks good to me. But I think our some of our production clusters are relying on this default value defined in the code so not sure if this will break any workflows in production. We'll need to be careful when releasing this change.

Yeah please take a look at the Uber's internal cluster when doing release. I think I updated the internal one to a smaller number last year when fixing the same issue for an internal customer. But I can't remember exactly. 
If not you can override the config to zero in dynamic config to keep the same behavior.

PR #3780, opened 2020-11-20
Fix workflow retry policy for childWorkflow

longquanzheng:
**What changed?**
Fix workflow retry policy for childWorkflow

**Why?**
For https://github.com/uber/cadence/issues/3351

**How did you test it?**
Local test.
We need https://github.com/uber/cadence/issues/3779 

**Potential risks**
No, it's already broken. The fix is straightfoward that it shouldn't cause any issue.

### DIFF
--- a/service/history/task/transfer_active_task_executor.go
+++ b/service/history/task/transfer_active_task_executor.go
@@ -1267,50 +1267,49 @@ func (t *transferActiveTaskExecutor) startWorkflowWithRetry(
 	attributes *workflow.StartChildWorkflowExecutionInitiatedEventAttributes,
 ) (string, error) {
 
-	now := t.shard.GetTimeSource().Now()
-	request := &h.StartWorkflowExecutionRequest{
-		DomainUUID: common.StringPtr(task.TargetDomainID),
-		StartRequest: &workflow.StartWorkflowExecutionRequest{
-			Domain:                              common.StringPtr(targetDomain),
-			WorkflowId:                          attributes.WorkflowId,
-			WorkflowType:                        attributes.WorkflowType,
-			TaskList:                            attributes.TaskList,
-			Input:                               attributes.Input,
-			Header:                              attributes.Header,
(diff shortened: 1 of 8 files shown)


service/history/task/transfer_active_task_executor.go
```diff
@@ -1272,50 +1272,49 @@ func (t *transferActiveTaskExecutor) startWorkflowWithRetry(
...
-			backoff.GetBackoffForNextScheduleInSeconds(
-				attributes.GetCronSchedule(),
-				now,
-				now,
-			),
-		),
+		InitiatedId: common.Int64Ptr(task.ScheduleID),
 	}
+	now := t.shard.GetTimeSource().Now()
+	historyStartReq.FirstDecisionTaskBackoffSeconds = common.Int32Ptr(
```
yycptt, line 1304: This field is already set by `common.CreateHistoryStartWorkflowRequest` I believe, but we may want to modify that util function to also take `now` as a parameter.

yycptt (approved):
Thanks for fixing this issue! The comments below is not related to your changes.

- I think we can also do better in our mutable state code to catch this case, i.e: if expirationTimestamp is 0 in the request, but the expirationSeconds in retry policy is not 0. 

- It's probably also not a good idea to let the caller figure what the expiration time should be. This logic should be internal to history service or maybe just mutable state builder.

yycptt (commented):
Is it possible to update the unit test in `transfer_active_task_executor_test.go` to cover this case?

longquanzheng:
> Is it possible to update the unit test in `transfer_active_task_executor_test.go` to cover this case given that we don't have an integration test for now?

Yes, will do it.

service/history/task/transfer_active_task_executor.go
longquanzheng: good point. I have refactored it to passing time source as parameter.

longquanzheng: Added an issue for the refactoring https://github.com/uber/cadence/issues/3800

PR #3781, opened 2020-11-21
Improve building tools and docs

longquanzheng:
**What changed?**
Improve building tools and docs

**Why?**
1. The links to using persistence tools is not connected from main page
2. Building tools is too slow because it is running with other unrelated stuff about server/thrif/canary/etc

**How did you test it?**
Local run

**Potential risks**
No

### DIFF
--- a/tools/cassandra/README.md
+++ b/tools/cassandra/README.md
@@ -9,7 +9,7 @@ make install-schema
 ## For production
 
 ### Create the binaries
-- Run `make tools`
+- Run `make cadence-cassandra-tool`
 - You should see an executable `cadence-cassandra-tool`
 
 ### Do one time database creation and schema setup for a new cluster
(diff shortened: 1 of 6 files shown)


tools/cassandra/README.md
```diff
@@ -9,7 +9,7 @@ make install-schema
 ## For production
 
 ### Create the binaries
-- Run `make bins`
+- Run `make tools`
```
yycptt, line 12: Shall we only build the specific tool here instead of building all three tools? Same question for cli and sql README.

longquanzheng: Yeah that works too. I thought we might want to run a simpler command like `make bins` here.

longquanzheng: updated :D and add a tip for use `make tools` to build all tools.

yycptt (approved):

PR #3785, opened 2020-11-22
Add cadence_service tag for all metrics in rootScope

longquanzheng:
**What changed?**
1. Add cadence_service tag for all metrics in rootScope
2. Remove unused metric tag values

**Why?**
https://github.com/uber/cadence/issues/3783
https://github.com/uber/cadence/issues/3766
Internally Uber service metric client emit tag value for each service, while the open source version doesn't have equivalent metrics tag. The service tag is important because there are metrics that have the same name across services, for example cadence_request and cadence_errors. Without differentiating the service, there is no way to set up correct dashboard and monitors. 

**How did you test it?**
```
# HELP cadence_authorization_latency cadence_authorization_latency histogram
# TYPE cadence_authorization_latency histogram
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.001"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.002"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.005"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.01"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.02"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.05"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.1"} 10
cadence_authorization_latency_bucket{cadence_service="cadence_frontend",domain="all",operation="DescribeDomain",le="0.2"} 10
```

**Potential risks**
No


common/metrics/defs.go
```diff
@@ -1344,13 +1349,6 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		HistoryArchiverScope:    {operation: "HistoryArchiver"},
 		VisibilityArchiverScope: {operation: "VisibilityArchiver"},
 
-		BlobstoreClientUploadScope:          {operation: "BlobstoreClientUpload", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientDownloadScope:        {operation: "BlobstoreClientDownload", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientGetMetadataScope:     {operation: "BlobstoreClientGetMetadata", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientExistsScope:          {operation: "BlobstoreClientExists", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientDeleteScope:          {operation: "BlobstoreClientDelete", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
-		BlobstoreClientDirectoryExistsScope: {operation: "BlobstoreClientDirectoryExists", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
```
yycptt, line 1352: Those definitions are used by our internal archiver implementations I believe. 
https://github.com/uber/cadence/blob/master/common/metrics/defs.go#L647

longquanzheng: okay I will add it back. I thought it is from some deprecated code.

yycptt (commented):
Actually I think the same thing can be achieved by modifying the config file? There's a field for default tags and we can set it to different values for different services: https://github.com/uber/cadence/blob/master/common/service/config/config.go#L303

longquanzheng:
> Actually I think the same thing can be achieved by modifying the config file? There's a field for default tags and we can set it to different values for different services: https://github.com/uber/cadence/blob/master/common/service/config/config.go#L303

Yes, but that's not required. I think it's better to make it default to everyone so that they don't need to learn about using that configuration. 

I think that tags config can be used for some other purpose for example if they have different clusters to manager.

common/service/config/metrics.go
```diff
@@ -23,6 +23,8 @@ package config
 import (
 	"time"
 
+	"github.com/uber/cadence/common/metrics"
+
```
yycptt, line 27: nit: fix format

yycptt (approved):

PR #3787, opened 2020-11-22
Support different SSL modes

longquanzheng:
**What changed?**
Support different SSL modes

**Why?**
For https://github.com/uber/cadence/pull/3488#discussion_r528247325

**How did you test it?**
TODO: Test with local Postgres

**Potential risks**
No, it's completely backward compatible

### DIFF
--- a/common/auth/tls.go
+++ b/common/auth/tls.go
@@ -27,7 +27,7 @@ type (
 
 		// For Postgres(https://www.postgresql.org/docs/9.1/libpq-ssl.html) and MySQL
 		// default to require if Enable is true.
-		// For MySQL: https://github.com/go-sql-driver/mysql , it also can be set in ConnectAttributes, default is tls-custom
+		// For MySQL: https://github.com/go-sql-driver/mysql , it also can be set by ConnectAttributes, default is tls-custom
 		SSLMode string `yaml:"sslmode" `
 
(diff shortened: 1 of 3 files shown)


common/auth/tls.go
```diff
@@ -25,6 +25,11 @@ type (
 	TLS struct {
 		Enabled bool `yaml:"enabled"`
 
+		// for Postgres: see more https://www.postgresql.org/docs/9.1/libpq-ssl.html
+		// default to require if Enable is true
+		// TODO: support MySQL/Cassandra as well
```
longquanzheng, line 30: @JosefWN thanks for providing the idea. Do you happen to know how to support different modes for MySQL/Cassandra as well?

JosefWN: Wow, that was quick, thanks! I've been using Postgres for the past few years, but MySQL seems quite similar: https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode

Postgres, for reference: https://www.postgresql.org/docs/13/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS

Not sure about Cassandra, perhaps more info here: https://cassandra.apache.org/doc/latest/operating/security.html

yux0 (approved):

PR #3788, opened 2020-11-23
Switch common/archiver to internal types

vytautas-karpavicius:
**What changed?**
- Switched `common/archive` package to use internal types.
- Since json marshaler is used on internal types, also added `MarshalText`/`UnmarshalText` on internal type enums. Otherwise they get serialized as numbers.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/worker/archiver/activities.go
+++ b/service/worker/archiver/activities.go
@@ -24,6 +24,8 @@ import (
 	"context"
 	"errors"
 
+	"github.com/uber/cadence/common/types/mapper/thrift"
+
 	"go.uber.org/cadence"
 	"go.uber.org/cadence/activity"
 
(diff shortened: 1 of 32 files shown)


service/worker/archiver/activities.go
```diff
@@ -24,6 +24,8 @@ import (
 	"context"
 	"errors"
 
+	"github.com/uber/cadence/common/types/mapper/thrift"
```
yycptt, line 27: nit: group cadence imports

service/worker/archiver/client.go
```diff
@@ -27,6 +27,8 @@ import (
 	"math/rand"
 	"time"
 
+	"github.com/uber/cadence/common/types/mapper/thrift"
```
yycptt, line 30: nit: same here.

yycptt (approved):
LGTM. I think we also have some internal implementations for archiver. We may also want to change that as well.

PR #3790, opened 2020-11-23
Allow custom ES client options

yycptt:
**What changed?**
Allow custom ES client options

**Why?**
For internal use.

**How did you test it?**

**Potential risks**


longquanzheng (approved):
lgtm

PR #3794, opened 2020-11-24
Fix ES client config NPE

yycptt:
**What changed?**
Fix ES client config NPE by removing visibility config from es client.

**Why?**
Visibility config used by es client is not initialized in `cmd/server.go`, instead it's initialized by each service separately in `service.go` when setting up the visibility store and can potentially have different values in different services.

**How did you test it?**

**Potential risks**


longquanzheng (approved):
Lgtm. But better create a ticket for later refactoring to clear confusion.

yycptt:
> Lgtm. But better create a ticket for later refactoring to clear confusion.

Yup, created an issue for moving this field to `ElasticSearchConfig`: https://github.com/uber/cadence/issues/3795

PR #3796, opened 2020-11-24
History Error Injection Client

yycptt:
**What changed?**
Add an error injection client for history client

**Why?**
Injection random client error during in bench test to expose potential error handling bugs.

**How did you test it?**
Local test.

**Potential risks**
No risk. Error injection client won't be created if error rate is 0 on service start up.


service/worker/parentclosepolicy/workflow.go
```diff
@@ -24,6 +24,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/uber/cadence/common/types/mapper/thrift"
```
vytautas-karpavicius, line 27: nit: this seems like it is not used anywhere

vytautas-karpavicius (approved):
I know this is a common trend in this repo to add such decorators. But did anyone considered to use code generation for such occasions? This seems like it could be easily generated from a template.

service/worker/parentclosepolicy/workflow.go
yycptt: Good catch. Removed.

yycptt:
> I know this is a common trend in this repo to add such decorators. But did anyone considered to use code generation for such occasions? This seems like it could be easily generated from a template.

I think that's a good idea, let me create an issue to track it.

PR #3801, opened 2020-11-25
Remove protobuf dependency from go.mod

anish531213:
Removing protobuf dependency.

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -30,6 +30,8 @@ import (
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


service/frontend/workflowHandler.go
```diff
@@ -257,6 +259,10 @@ func (wh *WorkflowHandler) RegisterDomain(ctx context.Context, registerRequest *
 	scope, sw := wh.startRequestProfile(metrics.FrontendRegisterDomainScope)
 	defer sw.Stop()
 
+	var x proto.TimeoutType
+	x = 5
+	_ = x.String()
```
yycptt, line 264: Revert the change?

PR #3808, opened 2020-11-30
Run fixer workflow automatically

andrewjdawson2016:
Run fixer automatically be detecting the most recent scan to finish.


just-at-uber (approved):
LGTM. Might want another person to review also.

PR #3810, opened 2020-12-01
Introduce internal types for sql blobs

andrewjdawson2016:
This is step 1 of the following 

**1. Introduce these types**
2. Add mappers which will convert to and from thrift types 
3. Add mappers which will convert to and from proto types 
4. Update the encoder interface methods to accept these internal types - the encoder will then use the mapper written in step 2 to convert to thrift blob before encoding. 
5. Update the decoder interface methods to return these internal types - the decoder will then use the mapper written in step 2 to convert from thrift blob to internal before returning. 
6. Implement encoder and decoder for proto just like we did for thrift

### DIFF
--- a/common/persistence/serialization/interfaces.go
+++ b/common/persistence/serialization/interfaces.go
@@ -60,7 +60,7 @@ type (
 		Description                 *string
 		Owner                       *string
 		Status                      *int32
-		RetentionDays               *time.Duration
+		Retention                   *time.Duration
 		EmitMetric                  *bool
 		ArchivalBucket              *string
 		ArchivalStatus              *int16
@@ -77,9 +77,9 @@ type (
 		HistoryArchivalURI          *string
 		VisibilityArchivalStatus    *int16
 		VisibilityArchivalURI       *string
-		FailoverEndTime             *time.Time
+		FailoverEndTimestamp        *time.Time
 		PreviousFailoverVersion     *int64
-		LastUpdatedTime             *time.Time
+		LastUpdatedTimestamp        *time.Time
 	}
 
 	// HistoryBranchRange blob in a serialization agnostic format
@@ -91,9 +91,9 @@ type (
 
 	// HistoryTreeInfo blob in a serialization agnostic format
 	HistoryTreeInfo struct {
-		CreatedTime *time.Time
-		Ancestors   []*HistoryBranchRange
-		Info        *string
+		CreatedTimestamp *time.Time
+		Ancestors        []*HistoryBranchRange
+		Info             *string
 	}
 
 	// WorkflowExecutionInfo blob in a serialization agnostic format
@@ -107,8 +107,8 @@ type (
 		CompletionEventEncoding            *string
 		TaskList                           *string
 		WorkflowTypeName                   *string
-		WorkflowTimeoutSeconds             *time.Duration
-		DecisionTaskTimeoutSeconds         *time.Duration
+		WorkflowTimeout                    *time.Duration
+		DecisionTaskTimeout                *time.Duration
 		ExecutionContext                   []byte
 		State                              *int32
 		CloseStatus                        *int32
@@ -117,8 +117,8 @@ type (
 		LastEventTaskID                    *int64
 		LastFirstEventID                   *int64
 		LastProcessedEvent                 *int64
-		StartTime                          *time.Time
-		LastUpdatedTime                    *time.Time
+		StartTimestamp                     *time.Time
+		LastUpdatedTimestamp               *time.Time
 		DecisionVersion                    *int64
 		DecisionScheduleID                 *int64
 		DecisionStartedID                  *int64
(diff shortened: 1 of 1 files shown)


common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		ParentDomainID                     *string
+		ParentWorkflowID                   *string
+		ParentRunID                        *string
+		InitiatedID                        *int64
+		CompletionEventBatchID             *int64
+		CompletionEvent                    []byte
+		CompletionEventEncoding            *string
+		TaskList                           *string
+		WorkflowTypeName                   *string
+		WorkflowTimeoutSeconds             *time.Duration
```
vytautas-karpavicius, line 110: As this is `time.Duration`, rename it to `WorkflowTimeout`

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		DecisionTaskTimeoutSeconds         *time.Duration
```
vytautas-karpavicius, line 111: Same here, rename it to `DecisionTaskTimeout`

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		DecisionScheduledTimestamp         *time.Time
+		CancelRequested                    *bool
+		DecisionOriginalScheduledTimestamp *time.Time
+		CreateRequestID                    *string
+		DecisionRequestID                  *string
+		CancelRequestID                    *string
+		StickyTaskList                     *string
+		StickyScheduleToStartTimeout       *time.Duration
+		RetryAttempt                       *int64
+		RetryInitialIntervalSeconds        *time.Duration
```
vytautas-karpavicius, line 137: Same here, drop `Seconds`

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		RetryMaximumIntervalSeconds        *time.Duration
```
vytautas-karpavicius, line 138: Same here, drop `Seconds`

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		RetryMaximumAttempts               *int32
+		RetryExpirationSeconds             *time.Duration
```
vytautas-karpavicius, line 140: Same here, drop `Seconds`

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		RetryBackoffCoefficient            *float64
+		RetryExpirationTime                *time.Time
```
vytautas-karpavicius, line 142: Just a suggestion, some `time.Time` fields have `Time` suffix while others uses `Timestamp` suffix. Maybe we should unify this?

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		ScheduledEvent                []byte
+		ScheduledEventEncoding        *string
+		ScheduledTime                 *time.Time
+		StartedID                     *int64
+		StartedEvent                  []byte
+		StartedEventEncoding          *string
+		StartedTime                   *time.Time
+		ActivityID                    *string
+		RequestID                     *string
+		ScheduleToStartTimeoutSeconds *time.Duration
```
vytautas-karpavicius, line 174: For this and following lines below, drop `Seconds` suffix.

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		StartToCloseTimeoutSeconds    *time.Duration
+		HeartbeatTimeoutSeconds       *time.Duration
+		CancelRequested               *bool
+		CancelRequestID               *int64
+		TimerTaskStatus               *int32
+		Attempt                       *int32
+		TaskList                      *string
+		StartedIdentity               *string
+		HasRetryPolicy                *bool
+		RetryInitialIntervalSeconds   *time.Duration
```
vytautas-karpavicius, line 185: Same here

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
 )
 
+type (
+	// ShardInfo blob in a serialization agnostic format
+	ShardInfo struct {
+		StolenSinceRenew                      *int32
+		UpdatedAt                             *time.Time
+		ReplicationAckLevel                   *int64
+		TransferAckLevel                      *int64
+		TimerAckLevel                         *time.Time
```
vytautas-karpavicius, line 42: I think this should be `int64`

vytautas-karpavicius (commented):
As all those are newly defines types which are serialization agnostic, maybe we should start dropping pointer for fields which are always present?

common/persistence/serialization/interfaces.go
andrewjdawson2016: In persistenceInterface.go this has been converted to time.Time.

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		DomainNotificationVersion             *int64
+		ClusterTransferAckLevel               map[string]int64
+		ClusterTimerAckLevel                  map[string]time.Time
```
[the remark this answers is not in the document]
andrewjdawson2016: Same here.

common/persistence/serialization/interfaces.go
```diff
@@ -23,13 +23,281 @@
...
+		TimerProcessingQueueStatesEncoding    *string
+	}
+
+	// DomainInfo blob in a serialization agnostic format
+	DomainInfo struct {
+		Name                        *string
+		Description                 *string
+		Owner                       *string
+		Status                      *int32
+		RetentionDays               *time.Duration
```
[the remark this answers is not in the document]
andrewjdawson2016: Agreed

vytautas-karpavicius (approved):

PR #3812, opened 2020-12-01
Regenerated internal types to include ActivitiesToDispatchLocally

vytautas-karpavicius:
**What changed?**
Regenerated internal types to include `ActivitiesToDispatchLocally`

**Why?**
Previous version was generated using older version of thrift package, therefore was missing this field.

**How did you test it?**

**Potential risks**


mkolodezny (approved):
Looks good, thanks for the quick fix!

PR #3814, opened 2020-12-02
Add thrift mapper from internal sql types to and from thrift 

andrewjdawson2016:

### DIFF
--- a/common/persistence/serialization/thrift_mapper.go
+++ b/common/persistence/serialization/thrift_mapper.go
@@ -1,3 +1,25 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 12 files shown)


common/persistence/serialization/thrift_mapper.go
```diff
@@ -0,0 +1,746 @@
...
+func historyTreeInfoToThrift(info *HistoryTreeInfo) *sqlblobs.HistoryTreeInfo {
+	if info == nil {
+		return nil
+	}
+	result := &sqlblobs.HistoryTreeInfo{
+		CreatedTimeNanos: unixNanoPtr(info.CreatedTimestamp),
+		Info:             info.Info,
+	}
+	if info.Ancestors != nil {
+		result.Ancestors = make([]*shared.HistoryBranchRange, len(info.Ancestors), len(info.Ancestors))
```
vytautas-karpavicius, line 169: nit: there is no need specify array capacity when it is equal to length

common/persistence/serialization/thrift_mapper_test.go
```diff
@@ -0,0 +1,482 @@
...
+	assert.Equal(t, expected.ParentDomainID, actual.ParentDomainID)
+	assert.Equal(t, expected.ParentWorkflowID, actual.ParentWorkflowID)
+	assert.Equal(t, expected.ParentRunID, actual.ParentRunID)
+	assert.Equal(t, expected.InitiatedID, actual.InitiatedID)
+	assert.Equal(t, expected.CompletionEventBatchID, actual.CompletionEventBatchID)
+	assert.Equal(t, expected.CompletionEvent, actual.CompletionEvent)
+	assert.Equal(t, expected.CompletionEventEncoding, actual.CompletionEventEncoding)
+	assert.Equal(t, expected.TaskList, actual.TaskList)
+	assert.Equal(t, expected.WorkflowTypeName, actual.WorkflowTypeName)
+	assert.True(t, (*expected.WorkflowTimeout-*actual.WorkflowTimeout) < time.Second)
```
vytautas-karpavicius, line 227: Can you explain why this is needed? Do we loose precision with rounding on conversion?

vytautas-karpavicius (approved):

common/persistence/serialization/thrift_mapper_test.go
andrewjdawson2016: Yes we only store at seconds. But in practice we won't actually lose anything because we will also truncate to seconds at the RPC layer.

PR #3818, opened 2020-12-02
Log short context timeout for long poll GetWorkflowExecutionHistory

yycptt:
**What changed?**
Log short context timeout for long poll GetWorkflowExecutionHistory.

**Why?**
By default long poll GetWorkflowExecutionHistory will return after 20s if workflow's no progress or doesn't reach the specified eventID. If customer specified a shorter timeout for the call and their workflow takes much longer to run due to some downstream dependencies issues, they will see timeouts and the timeout will impact our worker SLA. 

Add this logging so that we have evidence that customer has specified an invalid timeout for this API. We can also use this information to adjust the default wait time for polling mutable state. 

**How did you test it?**
Integration test

**Potential risks**
N/A, this diff only adds logging for invalid timeout.


vytautas-karpavicius (approved):
LGTM

PR #3819, opened 2020-12-02
Add CLI command for signalWithStart workflow execution

yycptt:
**What changed?**
Add CLI command for signalWithStart workflow execution
https://github.com/uber/cadence/issues/3817

**Why?**
- Customer may want to use this API through CLI
- Currently SignalWithStart can be used to trigger cron workflow, which is very useful for development and testing purpose. 

**How did you test it?**
Tested locally.

**Potential risks**
N/A, server side code is not changed, only modified CLI code

### DIFF
--- a/tools/cli/util.go
+++ b/tools/cli/util.go
@@ -801,6 +801,9 @@ func processJSONInputHelper(c *cli.Context, jType jsonType) string {
 	case jsonTypeHeader:
 		flagNameOfRawInput = FlagHeaderValue
 		flagNameOfInputFileName = FlagHeaderFile
+	case jsonTypeSignal:
+		flagNameOfRawInput = FlagSignalInput
+		flagNameOfInputFileName = FlagSignalInputFile
 	default:
 		return ""
 	}
--- a/tools/cli/workflow.go
+++ b/tools/cli/workflow.go
@@ -82,50 +82,24 @@ func newWorkflowCommands() []cli.Command {
 			Name:    "signal",
 			Aliases: []string{"s"},
 			Usage:   "signal a workflow execution",
-			Flags: []cli.Flag{
-				cli.StringFlag{
-					Name:  FlagWorkflowIDWithAlias,
-					Usage: "WorkflowID",
-				},
-				cli.StringFlag{
-					Name:  FlagRunIDWithAlias,
-					Usage: "RunID",
-				},
-				cli.StringFlag{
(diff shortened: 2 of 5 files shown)


tools/cli/workflow.go
```diff
@@ -82,50 +82,24 @@ func newWorkflowCommands() []cli.Command {
...
-				},
-			},
+			Flags:   getFlagsForSignal(),
 			Action: func(c *cli.Context) {
 				SignalWorkflow(c)
 			},
 		},
+		{
+			Name:  "signalwithstart",
+			Usage: "signal with start a workflow execution",
```
longquanzheng, line 92: how about : Signal the workflow if an existing run is running, or start a workflow if a new run can be started based on IDResuePolicy, or ,otherwise return an error when attempting to start a new run.

tools/cli/util.go
```diff
@@ -801,6 +801,9 @@ func processJSONInputHelper(c *cli.Context, jType jsonType) string {
 	case jsonTypeHeader:
 		flagNameOfRawInput = FlagHeaderValue
 		flagNameOfInputFileName = FlagHeaderFile
+	case jsonTypeSignal:
+		flagNameOfRawInput = FlagSignalInput
+		flagNameOfInputFileName = FlagSignalInputFile
 	default:
 		return ""
```
longquanzheng, line 808: what is the meaning of this return?

tools/cli/workflow.go
```diff
@@ -82,50 +82,24 @@ func newWorkflowCommands() []cli.Command {
...
-				},
-				cli.StringFlag{
-					Name:  FlagRunIDWithAlias,
-					Usage: "RunID",
-				},
-				cli.StringFlag{
-					Name:  FlagNameWithAlias,
-					Usage: "SignalName",
-				},
-				cli.StringFlag{
```
longquanzheng, line 98: Do you intend to make this breaking changes from --input to --signal_input? 
If so you need to put it into the release notes :P 

Or maybe just keep it. I think it makes sense for a signal command to use --input instead of --signal_input because it's already clear that the input is for signal.

yycptt: Good catch, I change the name in the wrong place. Will revert the change here.

yycptt: Reverted to use `--input`

longquanzheng (approved):

PR #3821, opened 2020-12-03
Use internal types in persistence store parser

andrewjdawson2016:
This converts parser at the persistence store level to use internal types. After this is landed the remaining steps are 
1. Move the UUID lower in the stack. UUID is a serialization detail and it should be pushed down as low as possible. 
2. Implement proto encoder and decoder


common/persistence/serialization/thrift_mapper_test.go
```diff
@@ -27,6 +27,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/uber/cadence/common/types"
```
vytautas-karpavicius, line 30: nit: move this together with other uber import

vytautas-karpavicius (approved):

PR #3824, opened 2020-12-03
Add admin CLI scan command for unsupported workflows

yux0:
**What changed?**
Add admin CLI scan command for unsupported workflows

**Why?**
0.16 release cleans up the 2dc workflow code path. This 2dc workflow will not be supported after release 0.16.

**How did you test it?**
Cassandra local test
MySQL local test
TODO: Postgres test

**Potential risks**
The command may take a long time to finish with large data set.


common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -121,6 +121,7 @@ type (
 		DomainID   serialization.UUID
 		WorkflowID string
 		RunID      serialization.UUID
+		Size       int
```
yycptt, line 124: Add some comments regarding two semantics of this Filter?

tools/cli/adminDBScanCommand.go
```diff
@@ -157,3 +163,137 @@ func checkExecution(
 
 	return execution, invariant.NewInvariantManager(ivs).RunChecks(ctx, execution)
 }
+
+func AdminDBScanUnsupportedWorkflow(c *cli.Context) {
+	numberOfShards := c.Int(FlagNumberOfShards)
```
yycptt, line 168: I think this is a required parameter?

tools/cli/adminDBScanCommand.go
```diff
@@ -157,3 +163,137 @@ func checkExecution(
...
+	shardID int,
+	rps int,
+) persistence.ExecutionManager {
+
+	var execStore persistence.ExecutionStore
+	dbType := c.String(FlagDBType)
+	logger := loggerimpl.NewNopLogger()
+	switch dbType {
+	case "cassandra":
+		execStore = initialCassandraExecutionClient(c, shardID, logger)
```
yycptt, line 241: nit: rename `initial` to `initialize`?

tools/cli/adminDBScanCommand.go
```diff
@@ -31,9 +32,14 @@ import (
...
 
 	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/collection"
+	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/loggerimpl"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/persistence/cassandra"
+	"github.com/uber/cadence/common/persistence/serialization"
+	"github.com/uber/cadence/common/persistence/sql"
+	quotas2 "github.com/uber/cadence/common/quotas"
```
yycptt, line 42: Is this alias required?

tools/cli/adminDBScanCommand.go
```diff
@@ -157,3 +163,137 @@ func checkExecution(
...
+	rps int,
+	outputFile *os.File,
+) {
+
+	client := initializeExecutionStore(c, shardID, rps)
+	defer client.Close()
+
+	paginationFunc := func(paginationToken []byte) ([]interface{}, []byte, error) {
+		resp, err := client.ListConcreteExecutions(
+			context.Background(),
```
yycptt, line 189: not sure if this will cause the call to hang

tools/cli/adminDBScanCommand.go
```diff
@@ -157,3 +163,137 @@ func checkExecution(
...
 	return execution, invariant.NewInvariantManager(ivs).RunChecks(ctx, execution)
 }
+
+func AdminDBScanUnsupportedWorkflow(c *cli.Context) {
+	numberOfShards := c.Int(FlagNumberOfShards)
+	rps := c.Int(FlagRPS)
+	outputFile := getOutputFile(c.String(FlagOutputFilename))
+	defer outputFile.Close()
+	for i := 0; i < numberOfShards; i++ {
+		listExecutionsByShardID(c, i, rps, outputFile)
```
yycptt, line 173: Since the script may be long running, I will we should provide user with some progress information, like shard xxx has completed.

We can also consider give user the option to start from middle in case the scan file due to some transient error. Probably add flags for specifying shardID range instead of just the max shardID?

yux0: Changed to use lower shard boundary and upper shard boundary.

tools/cli/adminDBScanCommand.go
yux0: Updated with a timeout value

tools/cli/adminDBScanCommand.go
yux0: removed

tools/cli/adminDBScanCommand.go
yux0: updated

tools/cli/adminDBScanCommand.go
yux0: removed this argument

common/persistence/sql/sqlplugin/interfaces.go
yux0: Updated

RELEASES.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Note:** This CLI is a long-running process to scan the database for unsupported workflows.
+If you have TLS configurations or use customized encoding/decoding type. Please use
+
+`cadence adm db unsupported-workflow --help`
+
+to configurate the correct connection.
+
+
+After the CLI completes, a list of unsupported workflows will be display in the file `cadence_scan`.
```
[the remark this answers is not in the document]
yux0: Updated

tools/cli/adminDBScanCommand.go
```diff
@@ -157,3 +168,144 @@ func checkExecution(
...
+// AdminDBScanUnsupportedWorkflow is to scan DB for unsupported workflow for a new release
+func AdminDBScanUnsupportedWorkflow(c *cli.Context) {
+	rps := c.Int(FlagRPS)
+	outputFile := getOutputFile(c.String(FlagOutputFilename))
+	startShardID := c.Int(FlagLowerShardBound)
+	endShardID := c.Int(FlagUpperShardBound)
+
+	defer outputFile.Close()
+	for i := startShardID; i <= endShardID; i++ {
+		listExecutionsByShardID(c, i, rps, outputFile)
```
yycptt, line 181: I kind of feel like if the scan is done in only goroutine, it will take a long time to finish and the rps limit is kind of useless and will never be reached.

yycptt (approved):

PR #3825, opened 2020-12-03
Add various default config to startup

longquanzheng:
**What changed?**
1. add default values for SQL encodingType and decodingTypes
2. add default value for RPCName in clusterInformation

**Why?**
Because https://github.com/uber/cadence/pull/3534 our new release is going to break all the existing customers unless you upgrade their config. And also reuiqre us to update the helm chart https://github.com/banzaicloud/banzai-charts/blob/master/cadence/templates/server-configmap.yaml#L40

For https://github.com/uber/cadence/pull/3577

**How did you test it?**
Unit test and local test

**Potential risks**

### DIFF
--- a/config/development_mysql.yaml
+++ b/config/development_mysql.yaml
@@ -14,8 +14,6 @@ persistence:
         maxConns: 20
         maxIdleConns: 20
         maxConnLifetime: "1h"
-        encodingType: "thriftrw"
-        decodingTypes: [ "thriftrw" ]
     mysql-visibility:
       sql:
         pluginName: "mysql"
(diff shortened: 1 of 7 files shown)


andrewjdawson2016 (approved):

config/development_mysql.yaml
```diff
@@ -27,8 +25,6 @@ persistence:
         maxConns: 2
         maxIdleConns: 2
         maxConnLifetime: "1h"
-        encodingType: "thriftrw"
```
yycptt, line 30: Also remove those fields in postgres yaml and docker config template?

yycptt (approved):

config/development_mysql.yaml
longquanzheng: yeah, removed.

PR #3830, opened 2020-12-04
Improve context timeout handling for PollMutableState

yycptt:
**What changed?**
Improve context timeout handling for PollMutableState. 

**Why?**
Current implementation of PollMutableState by default will wait 20s for workflow to make progress. If the context timeout is less than 20s and there's no progress within that time period, a context deadline exceed error will be returned, and user will see timeout error returned from the rpc framework. 

Even if user initially specified a long timeout for the long poll get history, say 60s, as long as their workflow takes longer time to run, the (java) client side, will keep decreasing the timeout and call server with shorter and shorter timeout, which eventually leads to timeout.

This means if user workflow takes longer time to run due to downstream dependencies issues, our SLO will also be impacted by those timeout errors.  

**How did you test it?**
Integration test

**Potential risks**
N/A

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -78,6 +78,10 @@ const (
 	replicationTimeout                        = 30 * time.Second
 	contextLockTimeout                        = 500 * time.Millisecond
 
+	// pollMutableStateBufferDuration should be smaller than minPollMutableStateTimeout
+	pollMutableStateBufferDuration = 50 * time.Millisecond
+	minPollMutableStateTimeout     = 100 * time.Millisecond
+
 	// TerminateIfRunningReason reason for terminateIfRunning
 	TerminateIfRunningReason = "TerminateIfRunning Policy"
 	// TerminateIfRunningDetailsTemplate details template for terminateIfRunning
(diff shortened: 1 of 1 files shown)


service/history/historyEngine.go
```diff
@@ -1031,7 +1035,25 @@ func (e *historyEngineImpl) getMutableStateOrPolling(
...
+		expirationInterval := e.shard.GetConfig().LongPollExpirationInterval(domainCache.GetInfo().Name)
+		if deadline, ok := ctx.Deadline(); ok {
+			timeout := deadline.Sub(e.shard.GetTimeSource().Now())
+			if timeout < minPollMutableStateTimeout {
+				// timeout is too short, return directly
+				// note that minPollMutableStateTimeout is larger than pollMutableStateBufferDuration,
+				// this ensure that:
+				//     1. timeout-pollMutableStateBufferDuration > 0
+				//     2. leaves a reasonable amount of time to poll mutable state
+				return response, nil
```
yycptt, line 1048: In this case, I think we have 3 options:
1. return current response and nil error. when client side see the empty response, it will keep calling the API with a shorter timeout, and eventually leads to a timeout. This will impact SLA and result in more API calls.
2. return timeout error, will impact our SLA if user workflow can't complete within the timeout they initially specified.
3. return an application error, like BadRequest. My understanding is that user will see this error, and will be confused.

Without change on the client side (to stop calling server when the timeout is very short), none of them seems to be a good option. Any thoughts on it? @meiliang86 Thanks.

Groxx: 2: this means the SLA is bad and should be fixed, not that we should consider adding complexity to the code to work around it.

yycptt: I agree. That's why in the comment I suggest making changes to the client side instead of here so that it won't call this API with an unreasonable small context timeout. But I guess I still need the same logic for 1 (making sure the timeout is > buffer duration so that we can return an empty response just before the context timeout).

The issue right now is that the client side logic will keep calling this API will smaller and smaller timeout which almost guarantees a timeout error and affect our SLA if the initial context timeout they specified on the client side is less than their workflow execution time.

service/history/historyEngine.go
```diff
@@ -78,6 +78,10 @@ const (
 	replicationTimeout                        = 30 * time.Second
 	contextLockTimeout                        = 500 * time.Millisecond
 
+	// pollMutableStateBufferDuration should be smaller than minPollMutableStateTimeout
```
mantas-sidlauskas, line 81: nit: this can be enforced if `minPollMutableStateTimeout` is declared like `minPollMutableStateTimeout = pollMutableStateBufferDuration * 2`

PR #3842, opened 2020-12-09
Switch Health status endpoints to internal types

vytautas-karpavicius:
**What changed?**
- Generated health related internal types and their mappers
- Updated health status handler endpoints to use internal types

**Why?**

**How did you test it?**

**Potential risks**


yux0:
LGTM. I also retried to run the main.go to generate the internal types. Could we add a README about how to add new thrift types and how to run the generator?

yux0 (approved):

vytautas-karpavicius:
> LGTM. I also tried to run the main.go to generate the internal types. Could we add a README about how to add new thrift types and how to run the generator?

Generator is not intended to stay. We should remove it once the migration is finished and proto IDLs are added. After that new types should be added manually as they will need to be in sync with all thrift, proto and internal structures anyway.

PR #3844, opened 2020-12-10
Set config for shardscanner fixer

mantas-sidlauskas:
**What changed?**
Fixing missing `config` to fixer config

**Why?**
to not have nil pointer exception

**How did you test it?**
Added a test to check if `config` is set to scanner and fixer configs

**Potential risks**


mkolodezny (approved):
LGTM

PR #3862, opened 2020-12-18
Allow using Kafka TLS without cert ca and key 

longquanzheng:
**What changed?**
Allow using Kafka TLS without cert ca and key 

**Why?**
This is to align with the Kafka client samples: https://github.com/Shopify/sarama/blob/0616f68815691527d3b6ab4d95a2881e09400fe3/examples/sasl_scram_client/main.go#L41
We have community users that need to connect to Kafka in this way.

**How did you test it?**
Tested by community users.

**Potential risks**
Minimal: see https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#

### DIFF
--- a/common/messaging/kafka/clientImpl.go
+++ b/common/messaging/kafka/clientImpl.go
@@ -175,19 +175,26 @@ func convertTLSConfig(tlsConfig auth.TLS) (*tls.Config, error) {
 		return nil, nil
 	}
 
-	cert, err := tls.LoadX509KeyPair(tlsConfig.CertFile, tlsConfig.KeyFile)
-	if err != nil {
-		return nil, err
-	}
-	caCertPool := x509.NewCertPool()
-	pemData, err := ioutil.ReadFile(tlsConfig.CaFile)
-	if err != nil {
-		return nil, err
-	}
-	caCertPool.AppendCertsFromPEM(pemData)
+	if tlsConfig.CertFile != "" && tlsConfig.CaFile != "" && tlsConfig.KeyFile != "" {
+		cert, err := tls.LoadX509KeyPair(tlsConfig.CertFile, tlsConfig.KeyFile)
+		if err != nil {
+			return nil, err
+		}
+		caCertPool := x509.NewCertPool()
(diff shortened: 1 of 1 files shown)


yux0 (approved):

common/messaging/kafka/clientImpl.go
```diff
@@ -175,19 +175,26 @@ func convertTLSConfig(tlsConfig auth.TLS) (*tls.Config, error) {
...
-		return nil, err
-	}
-	caCertPool := x509.NewCertPool()
-	pemData, err := ioutil.ReadFile(tlsConfig.CaFile)
-	if err != nil {
-		return nil, err
+	if tlsConfig.CertFile != "" && tlsConfig.CaFile != "" && tlsConfig.KeyFile != "" {
+		cert, err := tls.LoadX509KeyPair(tlsConfig.CertFile, tlsConfig.KeyFile)
+		if err != nil {
+			return nil, err
```
AngerM, line 181: This isn't technically correct. You can use just a caFile and it works to enable TLS. @longquanzheng 

our version looked something like this:
```
	config := tls.Config{
	if err != nil {			InsecureSkipVerify: !tlsConfig.EnableHostVerification,
		return nil, err	
	}		}
	caCertPool := x509.NewCertPool()	
	pemData, err := ioutil.ReadFile(tlsConfig.CaFile)		if tlsConfig.CaFile != "" {
	if err != nil {			caCertPool := x509.NewCertPool()
		return nil, err			pemData, err := ioutil.ReadFile(tlsConfig.CaFile)
		if err != nil {
			return nil, err
		}
		caCertPool.AppendCertsFromPEM(pemData)

		config.RootCAs = caCertPool
	}

	if tlsConfig.CertFile != "" && tlsConfig.KeyFile != "" {
		cert, err := tls.LoadX509KeyPair(tlsConfig.CertFile, tlsConfig.KeyFile)
		if err != nil {
			return nil, err
		}

		config.Certificates = []tls.Certificate{cert}
	}		}
	caCertPool.AppendCertsFromPEM(pemData)	


	return &tls.Config{		return &config, nil
```

PR #3863, opened 2020-12-21
Initial protobuf types and API

vytautas-karpavicius:
**What changed?**
This is an initial protobuf types and gRPC service definitions for Cadence. Current Thrift types were used as a reference to define proto types. However multiple changes were applied to make proto types idiomatic, more consistent, pass Uber style guide and provide better extensibility options in the future.
- Whole API surface is split into public and internal. Public exposes only what is needed to communicate outside of Cadence server (clients, cli, web). Everything else resides within internal. For now, public part is committed here, later once refined and confirmed as working it will be moved out to [cadence-idl](https://github.com/uber/cadence-idl) repo.
- Public WorkflowService is split into `WorkflowAPI`, `DomainAPI`, `VisibilityAPI` and `WorkerAPI`. This will allow to create thinner clients in other languages that could expose only WorfklowAPI for example, for just starting new workflows but not hosting entire worker.
- IDLs are defined to be idiomatic and pass Uber style. [Buf](https://github.com/bufbuild/buf) linter is used to ensure it with [Uber style rules](https://docs.buf.build/migration-prototool/#uber1-uber2). This includes:
  - Proto package naming, versioning and matching file layout
  - `API` suffix for all services
  - `_INVALID` suffix for zero value enums
  - Casing for types, fields
  - [Requirement](https://docs.buf.build/lint-checkers#rpc_request_standard_name-rpc_response_standard_name-rpc_request_response_unique) for request/response types to be only used within one service. Currently this is not the case in Thrift, as request/response types are reused in more that one service.
- Split API into several proto files for easier search and navigation.
- Introduced wrapper types for future needs:
  - `Payload` to wrap data bytes used as input, details or results. Later on this is expected to carry additional metadata such as encoding format.
  - `Failure` to wrap and group failure related fields `reason` and `details`. Later on this may include additional failure related fields.
- Error types no longer include `message` field as it is default field in grpc [status.Status](https://godoc.org/google.golang.org/grpc/status). Remaining fields will be used as error details.
- Use standard time and [duration](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Duration) and [timestamp](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp) types. Suffix the with `_timeout`, `_interval` or `_time`. No longer use time units in field name such as seconds or days.
- Additional unifications for consistency:
  - Use `domain_id` everywhere (no `domain_uuid`)
  - Use `page_size` everywhere (no `maximum_page_size`)
  - Use `int32` type for all `attempt` fields
  - Use `int32` type for all `shard_id` fields
  - Use `workflow_execution` everywhere (no `execution`)

This does not yet include persistence types.

**Why?**
To add gRPC support for Cadence.

**How did you test it?**
`make proto`

**Potential risks**
None yet. This introduces proto IDLs and generates them. None are used yet.


proto/public/uber/cadence/api/v1/error.proto
```diff
@@ -0,0 +1,65 @@
...
+
+package uber.cadence.api.v1;
+
+option go_package = "apiv1";
+option java_multiple_files = true;
+option java_outer_classname = "ApiProto";
+option java_package = "com.uber.cadence.api.v1";
+
+message WorkflowExecutionAlreadyStartedError {
+  string start_request_id = 1;
```
Groxx, line 31: `workflow_id`?  or does this mean something else?

proto/public/uber/cadence/api/v1/error.proto
```diff
@@ -0,0 +1,65 @@
...
+}
+
+message DomainNotActiveError {
+  string domain = 1;
+  string current_cluster = 2;
+  string active_cluster = 3;
+}
+
+message RetryTaskV2Error {
+  string domain_id = 1;
```
Groxx, line 47: though this is in multiple places, I just noticed it here:
there's some inconsistency between `domain` and `domain_id`.  do they ever mean different things?  or can they be standardized?

proto/public/uber/cadence/api/v1/error.proto
```diff
@@ -0,0 +1,65 @@
...
+  string workflow_id = 2;
+  string run_id = 3;
```
Groxx, line 49: similar to above: there's both `workflow_id + run_id` field pairs and `api.v1.WorkflowExecution` used in what seem like inconsistent ways, with a handful of exceptions where only `workflow_id` is relevant or where the whole request is just these two fields (so nesting seems unnecessary).  standardize?  or is there a reason they're separate in some places?

proto/internal/uber/cadence/indexer/v1/messages.proto
```diff
@@ -0,0 +1,58 @@
...
+  map<string, Field> fields = 6;
+}
+
+message Field {
+  FieldType type = 1;
+  oneof data {
+    string string_data = 2;
+    int64 int_data = 3;
+    bool bool_data = 4;
+    bytes binary_data = 5;
```
Groxx, line 42: irrelevant for this set of changes, but:
a canonical timestamp field would probably be a good idea, so users don't have to come up with a string format or decide on what precision to use for integers.

proto/public/uber/cadence/api/v1/error.proto
yux0: Tes. domain and domain_id are different. Domain is the actual user domain name. Domain ID is a UUID.

proto/public/uber/cadence/api/v1/error.proto
vytautas-karpavicius: I like this proposal, it would be more consistent. Unless team have any semantic reasons why were declared separately in the first place.

proto/public/uber/cadence/api/v1/error.proto
vytautas-karpavicius: The way I understand it, this is correct.
User receiving this error already knows `workflow_id` because he was trying to start it with this id.
`start_request_id` may give him more info which request was used to start it and which specific `run_id` was started.

Groxx: ah, right, start requests do have a `request_id`.  this makes sense then, my mistake :)

proto/public/uber/cadence/api/v1/error.proto
Groxx: after a closer look: it kinda looks like `domain_id` is used only for internal / worker purposes?  that's probably fine then / makes sense - uuids are easier for identity.

there is one exception though: `WorkflowExecutionInfo` is user-visible and only contains `domain_id`, not `domain` (unless it's nested in one of the other fields that I haven't noticed?).  worth adding `domain` there maybe?

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,286 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+syntax = "proto3";
```
andrewjdawson2016, line 21: Are we using the same proto compiler that is used for internal? Otherwise we will get errors when we pull to internal.

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,286 @@
...
+  rpc ResendReplicationTasks(ResendReplicationTasksRequest) returns (ResendReplicationTasksResponse);
+}
+
+message DescribeWorkflowExecutionRequest {
+  string domain = 1;
+  api.v1.WorkflowExecution workflow_execution = 2;
+}
+
+message DescribeWorkflowExecutionResponse {
+  string shard_id = 1;
```
andrewjdawson2016, line 102: Why is shard_id a string instead of an int32?

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,286 @@
...
+  string history_addr = 2;
+  string mutable_state_in_cache = 3;
+  string mutable_state_in_database = 4;
```
andrewjdawson2016, line 105: Why are these strings instead of booleans?

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,286 @@
...
+}
+
+message CloseShardResponse {
+}
+
+message RemoveTaskRequest {
+  int32 shard_id = 1;
+  shared.v1.TaskType task_type = 2;
+  int64 task_id = 3;
+  google.protobuf.Timestamp visibility_time = 4 [(gogoproto.stdtime) = true];
```
andrewjdawson2016, line 133: What is this = true?

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,286 @@
...
+  api.v1.WorkflowExecution workflow_execution = 2;
+}
+
+message RefreshWorkflowTasksResponse {
+}
+
+message ResendReplicationTasksRequest {
+  string domain_id = 1;
+  string workflow_id = 2;
+  string run_id = 3;
```
andrewjdawson2016, line 279: Is there a uuid type in proto we should consider using? Or is string the standard way to represent uuids in proto?

proto/internal/uber/cadence/admin/v1/service.proto
Groxx: `[(gogoproto.stdtime) = true]` -> an annotation.
in this case, it's to tell the gogoproto code generator to use the stdlib time type rather than something that directly fits the google.protobuf.Timestamp structure (which is basically a `struct{seconds int64, nanos int32}`)

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: This was missed, as it was ported from Thrift where this is weirdly a string: https://github.com/uber/cadence-idl/blob/3a2b613eaf9cebcd419e9926e09d02de9c7f15da/thrift/admin.thrift#L207

Thanks for spotting it, I have updated it to be int32, like everywhere else.

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: Those are internal mutable state serialized as json: https://github.com/uber/cadence/blob/master/service/history/historyEngine.go#L1422

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: To my knowledge there is no default UUID type for proto, and [this](https://stackoverflow.com/questions/36344826/how-do-i-represent-a-uuid-in-a-protobuf-message) confirms that it best represented with either string or []byte. Given we already use strings everywhere, I would say we keep them here as well.

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: I'm not sure what you mean by that. This is only one set of proto files and they all use same compiler. Later, once we verify that everything works as expected, public part of API will be moved out to cadence-idls repo.

proto/public/uber/cadence/api/v1/error.proto
vytautas-karpavicius: Updated to wrap all `workflow_id + run_id` pairs with `api.v1.WorkflowExecution`.

andrewjdawson2016 (approved):

proto/public/uber/cadence/api/v1/decision.proto
```diff
@@ -0,0 +1,171 @@
...
+import "google/protobuf/duration.proto";
+import "dependencies/gogoproto.proto";
+import "uber/cadence/api/v1/common.proto";
+import "uber/cadence/api/v1/tasklist.proto";
+import "uber/cadence/api/v1/workflow.proto";
+
+message Decision {
+  DecisionType decision_type = 1;
+  oneof attributes {
+    ScheduleActivityTaskDecisionAttributes schedule_activity_task_decision_attributes = 2;
```
Groxx, line 39: this applies across a few such messages, but: why both "type" and "oneof"?  seems like oneof already provides whatever type would offer

proto/public/uber/cadence/api/v1/decision.proto
```diff
@@ -0,0 +1,171 @@
...
+  Memo memo = 14;
+  SearchAttributes search_attributes = 15;
+}
+
+message SignalExternalWorkflowExecutionDecisionAttributes {
+  string domain = 1;
+  WorkflowExecution workflow_execution = 2;
+  string signal_name = 3;
+  Payload input = 4;
+  bytes control = 5;
```
Groxx, line 165: what's "control" for?

proto/public/uber/cadence/api/v1/workflow.proto
```diff
@@ -0,0 +1,217 @@
...
+import "uber/cadence/api/v1/tasklist.proto";
+
+message WorkflowExecutionInfo {
+  WorkflowExecution workflow_execution = 1;
+  WorkflowType type = 2;
+  google.protobuf.Timestamp start_time = 3 [(gogoproto.stdtime) = true];
+  google.protobuf.Timestamp close_time = 4 [(gogoproto.stdtime) = true];
+  WorkflowExecutionCloseStatus close_status = 5;
+  int64 history_length = 6;
+  string parent_domain_id = 7;
```
Groxx, line 43: this and error.proto are the only public mentions of domain_id.  maybe change them to domain?  or perhaps include both.

proto/public/uber/cadence/api/v1/common.proto
```diff
@@ -0,0 +1,111 @@
...
+  string name = 1;
+}
+
+message ActivityType {
+  string name = 1;
+}
+
+message Payload {
+  bytes data = 1;
+}
```
Groxx, line 48: I kinda wonder if it'd be worth replacing all use of `Payload` with `DataBlob`, with an `ENCODING_TYPE_UNKNOWN / _OPAQUE`... might make it possible to "upgrade" things in-place later.

e.g. `ScheduleActivityTaskDecisionAttributes` -> if `_PROTO3`, parse the `data` as an `Any`, and you might be able to understand and display it in a UI.  Then the unknown/opaque behavior is just like it is now: try to display, maybe it's json, maybe not.

proto/public/uber/cadence/api/v1/domain.proto
```diff
@@ -0,0 +1,83 @@
...
+import "google/protobuf/timestamp.proto";
+import "dependencies/gogoproto.proto";
+
+message DomainInfo {
+  string name = 1;
+  DomainStatus status = 2;
+  string description = 3;
+  string owner_email = 4;
+  map<string, string> data = 5;
+  string uuid = 6;
```
Groxx, line 40: should this be "domain_id" (or "id") instead?

proto/public/uber/cadence/api/v1/workflow.proto
Groxx: as another example: `WorkflowExecutionStartedEventAttributes` has `string parent_workflow_domain` (and no domain_id)

proto/public/uber/cadence/api/v1/history.proto
```diff
@@ -0,0 +1,499 @@
...
+  TaskList task_list = 5;
+  Payload input = 6;
+  google.protobuf.Duration execution_start_to_close_timeout = 7 [(gogoproto.stdduration) = true];
+  google.protobuf.Duration task_start_to_close_timeout = 8 [(gogoproto.stdduration) = true];
+  string continued_execution_run_id = 9;
+  ContinueAsNewInitiator initiator = 10;
+  Failure continued_failure = 11;
+  Payload last_completion_result = 12;
+  // This is the runID when the WorkflowExecutionStarted event is written
+  string original_execution_run_id = 13;
```
Groxx, line 107: iirc this != current run id when you do a reset?  or is it something else?  may be worth mentioning in a comment.

proto/public/uber/cadence/api/v1/history.proto
```diff
@@ -0,0 +1,499 @@
...
+  string identity = 2;
+  string request_id = 3;
+}
+
+message DecisionTaskCompletedEventAttributes {
+  int64 scheduled_event_id = 1;
+  int64 started_event_id = 2;
+  string identity = 3;
+  string binary_checksum = 4;
+  bytes execution_context = 5;
```
Groxx, line 153: same "maybe DataBlob" note for execution_context (also in service_worker.proto), though I'm fuzzier on this one's purpose.

proto/public/uber/cadence/api/v1/service_domain.proto
```diff
@@ -0,0 +1,129 @@
...
+  string visibility_archival_uri = 14;
+}
+
+message RegisterDomainResponse {
+}
+
+message DescribeDomainRequest {
+  string name = 1;
+  string id = 2;
+}
```
Groxx, line 79: not necessarily worthwhile, but: I'm guessing these are mutually exclusive?  may be worth mentioning optional / use `oneof`.

proto/public/uber/cadence/api/v1/service_domain.proto
```diff
@@ -0,0 +1,129 @@
...
+message ListDomainsResponse {
+  repeated DescribeDomainResponse domains = 1;
+  bytes next_page_token = 2;
+}
+
+message UpdateDomainInfo {
+  string description = 1;
+  string owner_email = 2;
+  map<string, string> data = 3;
+}
```
Groxx, line 103: only used in one place (below), and the others have these just flat in the parent message.  maybe worth flattening into the update request?  though that will have the semantic change of no longer being nullable.

proto/public/uber/cadence/api/v1/service_visibility.proto
```diff
@@ -0,0 +1,140 @@
...
+message ListClosedWorkflowExecutionsRequest {
+  string domain = 1;
+  int32 page_size = 2;
+  bytes next_page_token = 3;
+  StartTimeFilter start_time_filter = 4;
+  oneof filters {
+    WorkflowExecutionFilter execution_filter = 5;
+    WorkflowTypeFilter type_filter = 6;
+    StatusFilter status_filter = 7;
+  }
```
Groxx, line 94: IMO sorta weird that you can't search for both type and status.  (execution is pretty specific, maybe not necessary to combine with others... except maybe status?)
is there a technical issue preventing it on the backend?

proto/internal/uber/cadence/matching/v1/service.proto
```diff
@@ -0,0 +1,214 @@
...
+  int64 backlog_count_hint = 8;
+  bool sticky_execution_enabled = 9;
+  api.v1.WorkflowQuery query = 10;
+  shared.v1.TransientDecisionInfo decision_info = 11;
+  api.v1.TaskList workflow_execution_task_list = 12;
+  int32 event_store_version = 13;
+  bytes branch_token = 14;
+  google.protobuf.Timestamp scheduled_time = 15 [(gogoproto.stdtime) = true];
+  google.protobuf.Timestamp started_time = 16 [(gogoproto.stdtime) = true];
+  map<string, api.v1.WorkflowQuery> queries = 17;
```
Groxx, line 108: both "query" and "queries"?  I assume this mimics the thrift API, but that's a bit odd

proto/public/uber/cadence/api/v1/service_worker.proto
```diff
@@ -0,0 +1,323 @@
...
+  int64 started_event_id = 5;
+  int64 attempt = 6;
+  int64 backlog_count_hint = 7;
+  History history = 8;
+  bytes next_page_token = 9;
+  WorkflowQuery query = 10;
+  TaskList workflow_execution_task_list = 11;
+  google.protobuf.Timestamp scheduled_time = 12 [(gogoproto.stdtime) = true];
+  google.protobuf.Timestamp started_time = 13 [(gogoproto.stdtime) = true];
+  map<string, WorkflowQuery> queries = 14;
```
Groxx, line 167: both "query" and "queries" is fairly odd, especially since the RespondDecisionTaskCompletedRequest has only a query_results map.  is "query" perhaps just un-cleaned-up thrift legacy?

proto/public/uber/cadence/api/v1/service_worker.proto
```diff
@@ -0,0 +1,323 @@
...
+  StickyExecutionAttributes sticky_attributes = 5;
+  bool return_new_decision_task = 6;
+  bool force_create_new_decision_task = 7;
+  string binary_checksum = 8;
+  map<string, WorkflowQueryResult> query_results = 9;
+}
+
+message RespondDecisionTaskCompletedResponse {
+  PollForDecisionTaskResponse decision_task = 1;
+  map<string, ActivityLocalDispatchInfo> activities_to_dispatch_locally = 2;
```
Groxx, line 184: any reason this wouldn't be part of `PollForDecisionTaskResponse`?  granted it only saves 1 poll/respond cycle per process, but still kinda odd.

proto/public/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -0,0 +1,255 @@
...
+  Payload input = 4;
+  string identity = 5;
+  string request_id = 6;
+  bytes control = 7;
+}
+
+message SignalWorkflowExecutionResponse {
+}
+
+message SignalWithStartWorkflowExecutionRequest {
```
Groxx, line 129: may be worth embedding `StartWorkflowExecutionRequest`, and adding `SignalWorkflowExecutionRequest` fields separately?  Or splitting out the unique fields of both to a separate message type.  Since they really are "the same thing" - if we add something to Start, we probably want it available on SignalWithStart too.

proto/public/uber/cadence/api/v1/common.proto
```diff
@@ -0,0 +1,111 @@
...
+
+message WorkerVersionInfo {
+  string impl = 1;
+  string feature_version = 2;
+}
+
+message SupportedClientVersions {
+  string go_sdk = 1;
+  string java_sdk = 2;
+}
```
Groxx, line 87: probably worth documenting the format / how comparison is defined.  I'm _guessing_ it's a standard semver string... but in that case string comparison won't get the right result, as `1.2.3` is less than `1.2.3-rc1`, so hopefully it's something else.

proto/public/uber/cadence/api/v1/common.proto
Groxx: hmms.  on more thinking: this could also provide a way to point to external resources (encoding_type_uri -> schema://path/to/resource) for large blobs.  which could be valid in _many_ uses of Payload, though not all (e.g. it'd be kinda crazy for `SearchAttributes`, but it'd be perfect for `RespondActivityTaskCompletedRequest`)

proto/public/dependencies/gogoproto.proto
```diff
@@ -0,0 +1,144 @@
...
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto2";
+package gogoproto;
```
Groxx, line 30: after a bunch more reading about gogoproto: is this only used for the `buf` linter?
it seems like as long as we're using the gogo-extended `protoc`, these extensions might be already understood, and we could import them as e.g. `github.com/gogo/protobuf/gogoproto/gogo.proto` instead.  (the file is effectively identical to this one though, so this being proto2 is fine/expected/ideal)

that said, it seems effectively harmless, as it's not actually _doing_ anything beyond modifying the parsing environment.  so I guess YOLO, though I'd be broadly interested in un-weirding a lot of the build process on this repo...

Groxx:
I've read the whole public api pretty carefully, not yet private though.  I'll try to do that tomorrow, but broadly I'd say don't wait on me :)  until this is actually rolled out, changes are still "free".

proto/internal/uber/cadence/admin/v1/service.proto
Groxx: yea, personally I think string is best for uuids.  it's dramatically easier to deal with when things fall apart than a bunch of bytes.  if representing UUIDs as strings vs bytes is proving to be a performance bottleneck, we have a truly awesomely optimized system - otherwise, optimizing for humans/failure modes seems best.

proto/public/uber/cadence/api/v1/decision.proto
vytautas-karpavicius: TBH, I'm not entirely sure about their purpose. They seem to go all way along the commit history and allows some arbitrary data passed along when dealing with external or child workflows. Couple of places that actually uses them: [here](https://github.com/uber-go/cadence-client/blob/4653b9fd2a2ee0b7b9b6cfd5245b3f5f0414e325/internal/internal_decision_state_machine.go#L878) and [here](https://github.com/uber-go/cadence-client/blob/4653b9fd2a2ee0b7b9b6cfd5245b3f5f0414e325/internal/internal_decision_state_machine.go#L938). Both of them are strings converted to bytes.

I'm not aware about the origin intention of these fields, therefore it is hard to reason about them. Currently they are the same as in Thrift.

proto/public/uber/cadence/api/v1/common.proto
vytautas-karpavicius: Introduction of `Payload` wrapper was exactly for this purpose, to add encoding type (or maybe other metadata) next to the actual data later, if needed. Now, your suggestion to reuse the same DataBlob type, actually makes a lot of sense. I'm considering doing it.

proto/public/uber/cadence/api/v1/service_domain.proto
vytautas-karpavicius: Not sure about this. It kind of holds intent to update these fields. If we flatten them, then these fields come together with `name`. This might imply that we want to update name as well, rather than use it a domain selection field.

proto/public/uber/cadence/api/v1/domain.proto
vytautas-karpavicius: Agree, updated.

proto/public/uber/cadence/api/v1/history.proto
vytautas-karpavicius: This is for previous runId when doing continue as new. Updated the comment.

proto/public/uber/cadence/api/v1/service_domain.proto
vytautas-karpavicius: These are indeed mutually exclusive fields. Updated to use oneof.

proto/public/uber/cadence/api/v1/service_worker.proto
vytautas-karpavicius: This is not yet legacy, but it might be a good timing to drop this.

proto/public/dependencies/gogoproto.proto
vytautas-karpavicius: I did not manage to get it working when referencing like `github.com/gogo/protobuf/gogoproto/gogo.proto`. Both `protoc` and `buf` are complaining that they can not find such file.

proto/public/uber/cadence/api/v1/common.proto
vytautas-karpavicius: Yes, this is standard semver. We use https://github.com/hashicorp/go-version for comparison.

proto/public/uber/cadence/api/v1/service_worker.proto
vytautas-karpavicius: Maybe @mkolodezny can answer this.

proto/public/uber/cadence/api/v1/decision.proto
Groxx: heh.  good enough for me at least :)  maybe I'll understand them after reading more of the source.

proto/public/uber/cadence/api/v1/common.proto
Groxx: yea.  if there's a clear separation between payload-purposes and datablob-purposes, maybe it makes sense to separate them.  otherwise "a binary blob" seems pretty generic to me, and runtime limits on what's allowed in what API seem fine (similar to size limits.  not necessarily worth ensuring it via the type system)

proto/public/uber/cadence/api/v1/history.proto
Groxx: perfect, thanks!

proto/public/uber/cadence/api/v1/service_domain.proto
Groxx: I'm guessing pretty much just "name" isn't modifiable in the update request?  or are there others?

also, something I'm not familiar with: are _all_ of the fields (that are modifiable, including in nested messages) required to be non-empty?  otherwise we may not be able to tell "remove X" from "do not modify X"

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,285 @@
...
+}
+
+message DescribeWorkflowExecutionRequest {
+  string domain = 1;
+  api.v1.WorkflowExecution workflow_execution = 2;
+}
+
+message DescribeWorkflowExecutionResponse {
+  int32 shard_id = 1;
+  string history_addr = 2;
```
Groxx, line 103: name appears to be unique - does this mean something unique, or is it maybe `history_archival_uri`?

proto/public/uber/cadence/api/v1/service_worker.proto
meiliang86: This is for consistent query. @andrewjdawson2016 has more context.

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -0,0 +1,285 @@
...
+message DescribeClusterResponse {
+  api.v1.SupportedClientVersions supported_client_versions = 1;
+  shared.v1.MembershipInfo membership_info = 2;
+}
+
+message ReadDLQMessagesRequest {
+  shared.v1.DLQType type = 1;
+  int32 shard_id = 2;
+  string source_cluster = 3;
+  int64 inclusive_end_message_id = 4;
```
Groxx, line 234: excellent name 👍

proto/public/uber/cadence/api/v1/service_worker.proto
meiliang86: Talked to @Groxx offline. This should be fine.

Groxx: after some reading + asking Qs of @meiliang86 : this makes sense!  tl;dr is:
- worker polls for a decision task, gets one
- decides it can/wants to execute some activities locally / eagerly
- adds those to `RespondDecisionTaskCompletedRequest`'s decisions, with a marker that it can run them locally
- server decides if / which ones should be executed locally, which is in this response

so it's not a field that makes sense on a poll, only on a response to a completed decision task.

proto/internal/uber/cadence/indexer/v1/messages.proto
```diff
@@ -0,0 +1,59 @@
...
+  MessageType message_type = 1;
+  string domain_id = 2;
+  api.v1.WorkflowExecution workflow_execution = 3;
+  int64 version = 4;
+  map<string, Field> fields = 5;
+}
+
+message Field {
+  FieldType type = 1;
+  oneof data {
```
Groxx, line 39: same "both type-identifier and oneof-typed-fields" note as elsewhere

proto/internal/uber/cadence/matching/v1/service.proto
```diff
@@ -0,0 +1,214 @@
...
+  string forwarded_from = 7;
+}
+
+message AddDecisionTaskResponse {
+}
+
+message AddActivityTaskRequest {
+  string domain_id = 1;
+  api.v1.WorkflowExecution workflow_execution = 2;
+  string source_domain_id = 3;
```
Groxx, line 153: huh.  does this mean domain X can start an activity in domain Y?

proto/public/uber/cadence/api/v1/service_domain.proto
Groxx: - `DomainConfiguration.emit_metric` is a bool, so if the configuration is provided it'll have to be set in the request, or it'll be disabled.
- `*_archival_uri` are just strings, so `""` is ambiguous between "leave alone" and "remove", but maybe that's fine in practice (e.g. if status is invalid assume "no change", require explicit disable to delete)

otherwise they _appear_ pretty zero-value friendly to me.

Groxx:
Broadly:
- LGTM, ship it whenever you're happy
- `field-type-enum field_type` + `oneof field-type { ... }` pairs seem unnecessary / only make things more error-prone than just using the oneof list, so I'll vote to get rid of the enums/field-type-fields
- unifying most `bytes` uses/`Payload`/`DataBlob` sounds good in my head, but I'm open to putting it off or keeping them separate.
  - `Payload` and `DataBlob` in particular are probably fine to delay on as they are already wrappers and can be extended, but a couple of the `bytes` cases seem like they might benefit from a wrapper too (e.g. `control`, `execution_context`, arguably `details` in `Failure`).  But if they're going to remain opaque blobs, meh - `bytes` is accurate (task/next-page tokens are a good example there).
- given there are only two `domain_id` references in the public API: I'll vote to add domain name, and maybe remove the id field.  Only-ID is unnecessary friction for users who see the values.
- the rest of the comments are pretty minor

With caveats:
- I haven't been super careful about default-value behavior, but I agree from earlier discussion: the _vast_ majority seem safe, most fields seem required.  and since messages are nullable, at least some of the possible exceptions still work.

Groxx (approved):

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: This is ip:port of history service that returned the response.

proto/public/uber/cadence/api/v1/service_domain.proto
vytautas-karpavicius: At least in CLI (I think currently the only place), before issuing update request the current state is read and only changed fields are are overwritten. Thus all of them are provided in the update request.

However you made a good point and leaving a distinction for "do not modify" makes sense to me. In that case it makes sense to flatten all inner nested fields wrap each of them with a pointer.

proto/public/uber/cadence/api/v1/service_visibility.proto
vytautas-karpavicius: It goes all the way down to persistence https://github.com/uber/cadence/blob/master/common/persistence/visibilityInterfaces.go#L181. TBH I'm not sure why this was design that way. But if want to change this, it should be a separate PR.

proto/internal/uber/cadence/matching/v1/service.proto
vytautas-karpavicius: Not sure about this one's purpose. @meiliang86 do you know what is the use of this field?

proto/public/uber/cadence/api/v1/service_workflow.proto
vytautas-karpavicius: Updated.

proto/public/uber/cadence/api/v1/history.proto
vytautas-karpavicius: I have check [Go](https://github.com/uber-go/cadence-client/blob/4653b9fd2a2ee0b7b9b6cfd5245b3f5f0414e325/internal/internal_task_handlers.go#L1548) and [Java](https://github.com/uber/cadence-java-client/blob/5e9b33fdb0b8002c92ae50e2141e6b4d02fc74f8/src/main/java/com/uber/cadence/internal/replay/ReplayDecisionTaskHandler.java#L242) clients, and they seems not to set this field.

@meiliang86 do we still need this field? What is the purpose of it?

proto/internal/uber/cadence/matching/v1/service.proto
meiliang86: Not sure either. @yycptt or @yux0 do you know?

proto/public/uber/cadence/api/v1/service_domain.proto
meiliang86: minor: imo we should drop DomainConfiguration.emit_metric. This option adds more confusion than usefulness. Practically I haven't see any real use case to require turning metrics off (and I am not sure how to operate the cluster if that's the case).
We can do it in a different PR though if that makes things easier.

proto/public/uber/cadence/api/v1/history.proto
meiliang86: @yycptt @yux0 @andrewjdawson2016 any idea?

proto/public/uber/cadence/api/v1/service_domain.proto
vytautas-karpavicius: Dropped `emit_metric`

vytautas-karpavicius:
> Broadly:
> 
> * LGTM, ship it whenever you're happy
> * `field-type-enum field_type` + `oneof field-type { ... }` pairs seem unnecessary / only make things more error-prone than just using the oneof list, so I'll vote to get rid of the enums/field-type-fields
> * unifying most `bytes` uses/`Payload`/`DataBlob` sounds good in my head, but I'm open to putting it off or keeping them separate.
>   
>   * `Payload` and `DataBlob` in particular are probably fine to delay on as they are already wrappers and can be extended, but a couple of the `bytes` cases seem like they might benefit from a wrapper too (e.g. `control`, `execution_context`, arguably `details` in `Failure`).  But if they're going to remain opaque blobs, meh - `bytes` is accurate (task/next-page tokens are a good example there).
> * given there are only two `domain_id` references in the public API: I'll vote to add domain name, and maybe remove the id field.  Only-ID is unnecessary friction for users who see the values.
> * the rest of the comments are pretty minor
> 
> With caveats:
> 
> * I haven't been super careful about default-value behavior, but I agree from earlier discussion: the _vast_ majority seem safe, most fields seem required.  and since messages are nullable, at least some of the possible exceptions still work.

- I have addressed `oneof` + type proposal.
- For `Payload` and `DataBlob`, I will leave them separate for now.
- As for `domain_id` in public API, I agree with you, but will change those in a separate PR, as it will require deeper changes to push missing domain name through.

PR #3865, opened 2020-12-23
 Support visibility query with close status represented in string

yycptt:
**What changed?**
Accept string type as valid arguments for workflow close status in visibility query

**Why?**
https://github.com/uber/cadence/issues/3852
Previously user can only use number to represent close status and the query looks like `CloseStatus = 0`, now they can write `CloseStatus = "closed"`. This improves the user experience as they don't need to remember the mapping between actual status and the corresponding value. Also hided the value from user. 

**How did you test it?**
Tested locally.

**Potential risks**
In the worst case visibility query will stop to work. Parse of each individual query may a take a bit longer.


longquanzheng (commented):
That's really cool! I think that will greatly improve the experience of using query. Can you put a note in https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#?
1.  add release notes about breaking changes in list command.
2.  update cadence-docs after this is released

yycptt:
> That's really cool! I think that will greatly improve the experience of using query. Can you put a note in https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#?
> 
> 1. add release notes about breaking changes in list command.
> 2. update cadence-docs after this is released

Sure, I will update the doc. The change in the list command is not breaking BTW, we still support old syntax and treat them as aliases for the close status.

longquanzheng:
> > That's really cool! I think that will greatly improve the experience of using query. Can you put a note in https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#?
> > 
> > 1. add release notes about breaking changes in list command.
> > 2. update cadence-docs after this is released
> 
> Sure, I will update the doc. The change in the list command is not breaking BTW, we still support old syntax and treat them as aliases for the close status.

Including the number?
eg
```
cadence workflow list --query "WorkflowStatus=1"
```

yycptt:
> > > That's really cool! I think that will greatly improve the experience of using query. Can you put a note in https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#?
> > > 
> > > 1. add release notes about breaking changes in list command.
> > > 2. update cadence-docs after this is released
> > 
> > 
> > Sure, I will update the doc. The change in the list command is not breaking BTW, we still support old syntax and treat them as aliases for the close status.
> 
> Including the number?
> eg
> 
> ```
> cadence workflow list --query "WorkflowStatus=1"
> ```

Yes, user can continue to use number as CloseStatus, so `CloseStatus = 0` is the same thing as `CloseStatus = "completed"`.

longquanzheng:
> > > > That's really cool! I think that will greatly improve the experience of using query. Can you put a note in https://docs.google.com/document/d/1wbDyZyYkZ17LtxrevZn9uNzy6FkePgUk1ifPEQGS5ts/edit#?
> > > > 
> > > > 1. add release notes about breaking changes in list command.
> > > > 2. update cadence-docs after this is released
> > > 
> > > 
> > > Sure, I will update the doc. The change in the list command is not breaking BTW, we still support old syntax and treat them as aliases for the close status.
> > 
> > 
> > Including the number?
> > eg
> > ```
> > cadence workflow list --query "WorkflowStatus=1"
> > ```
> 
> Yes, user can continue to use number as CloseStatus, so `CloseStatus = 0` is the same thing as `CloseStatus = "completed"`.

got it. Thanks

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -817,6 +817,22 @@ func processAllValuesForKey(dsl *fastjson.Value, keyFilter func(k string) bool,
 	return nil
 }
 
+func combinedKeyFilter(key string) bool {
```
longquanzheng, line 820: Nit: isCombinedKeyFilter / isCompoundKeyFilter

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -817,6 +817,22 @@ func processAllValuesForKey(dsl *fastjson.Value, keyFilter func(k string) bool,
...
+	}
+
+	if closeStatusKeyFilter(key) {
+		return closeStatusProcessFunc(obj, key, value)
+	}
+
+	return fmt.Errorf("unknown es dsl key %v for processing value", key)
+}
+
 func timeKeyFilter(key string) bool {
```
longquanzheng, line 836: isTimeKeyFilter

common/persistence/elasticsearch/esVisibilityStore.go
yycptt: Makes sense. Updated to `isCombinedKey`, `isTimeKey` and `isCloseStatusKey`.

longquanzheng (approved):

PR #3868, opened 2020-12-28
Add instructions to setup local MySQL and Postgres

yux0:
**What changed?**
Add README for local SQL DB setup

**Why?**
Better contribution support

**How did you test it?**

**Potential risks**


readme/contribution/CONTRIBUTING.md
```diff
@@ -57,15 +57,24 @@ Connected to Test Cluster at 127.0.0.1:9042.
...
+# If you use SQL DB, then run:
+make install-schema-mysql OR make install-schema-postgres
 ```
 
 Then you will be able to run a basic local Cadence server for development:
 ```bash
 ./cadence-server start
+
+# If you use SQL DB, then run:
+./cadence-server --zone <mysql/postgres> start
```
longquanzheng, line 75: Isn't it development_mysql/development_postgres?

readme/contribution/MYSQL_SETUP.md
```diff
@@ -0,0 +1,35 @@
+# Setup local MySQL with Docker
```
longquanzheng, line 1: There are too many files in root and these two setup don't seem important enough to be in the root. How above move them to docs/?

readme/contribution/CONTRIBUTING.md
yycptt: I think the loading logic will automatically append the value for `zone` to `environment` when resolving the config file path.

readme/contribution/MYSQL_SETUP.md
yycptt: +1 for this. I think we already have a `docs` folder at root and we can reuse it instead of creating an extra `readme` folder?

andrewjdawson2016 (approved):

readme/contribution/MYSQL_SETUP.md
yux0: Sounds good. Moved to the docs folder

docs/readme/MYSQL_SETUP.md
```diff
@@ -0,0 +1,35 @@
+# Setup local MySQL with Docker
```
longquanzheng, line 1: why `readme`? this name doesn't give any useful information. Maybe `setup`?

yux0: updated the dir name

PR #3869, opened 2020-12-30
Update read DLQ messages API to return raw task info

yux0:
**What changed?**
Update read DLQ messages API to return raw task info

**Why?**
If the workflow deleted from the source cluster, the current read dlq message returns zero task

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/history/replication/dlq_handler.go
+++ b/service/history/replication/dlq_handler.go
@@ -49,7 +49,7 @@ type (
 			lastMessageID int64,
 			pageSize int,
 			pageToken []byte,
-		) ([]*types.ReplicationTask, []byte, error)
+		) ([]*types.ReplicationTask, []*types.ReplicationTaskInfo, []byte, error)
 		PurgeMessages(
 			ctx context.Context,
 			sourceCluster string,
@@ -96,7 +96,7 @@ func (r *dlqHandlerImpl) ReadMessages(
 	lastMessageID int64,
 	pageSize int,
 	pageToken []byte,
-) ([]*types.ReplicationTask, []byte, error) {
+) ([]*types.ReplicationTask, []*types.ReplicationTaskInfo, []byte, error) {
 
 	return r.readMessagesWithAckLevel(
 		ctx,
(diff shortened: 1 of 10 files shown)


tools/cli/flags.go
```diff
@@ -231,6 +231,7 @@ const (
 	FlagTLSEnableHostVerification         = "tls_enable_host_verification"
 	FlagDLQType                           = "dlq_type"
 	FlagDLQTypeWithAlias                  = FlagDLQType + ", dt"
+	FlagDLQRawTask                        = "show_dlq_raw_task"
```
yycptt, line 234: nit: maybe just `dlq_raw_task`?

service/history/replication/dlq_handler.go
```diff
@@ -159,11 +159,11 @@ func (r *dlqHandlerImpl) readMessagesWithAckLevel(
 			},
 		)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
 	}
 
-	return response.ReplicationTasks, resp.NextPageToken, nil
+	return response.ReplicationTasks, taskInfo, resp.NextPageToken, nil
```
yycptt, line 166: just want to clarify the behavior here: `taskInfo` will always contain info for all raw tasks, while `response.ReplicationTasks` only contains tasks whose history can be retrieved from the remote cluster?

If that's the case probably in CLI output we should make it more clear that some tasks are missing and urge the user to check raw task info.

yycptt (commented):
Is there a corresponding thrift IDL change?

tools/cli/adminDLQCommands.go
```diff
@@ -100,6 +105,24 @@ func AdminGetDLQMessages(c *cli.Context) {
 			ErrorAndExit("fail to print dlq messages.", err)
 		}
 	}
+
+	if showRawTask {
+		_, err := outputFile.WriteString(fmt.Sprintf("#### REPLICARION DLQ RAW TASKS INFO ####\n"))
```
vytautas-karpavicius, line 110: typo in: `REPLICARION`

andrewjdawson2016 (approved):

tools/cli/flags.go
yux0: updated

service/history/replication/dlq_handler.go
yux0: Added a warning message if the tasks are empty but the metadata task is not empty

tools/cli/adminDLQCommands.go
yux0: Good catch. Updated

PR #3871, opened 2021-01-05
replace string based logging with tagged logs

mantas-sidlauskas:
**What changed?**
Tag debug logs instead of using `Sprintf()`

**Why?**
Tags makes filtering logs possible.

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -52,6 +52,11 @@ type (
 	pollerIDCtxKey string
 	identityCtxKey string
 
+	queryResult struct {
+		workerResponse *types.MatchingRespondQueryTaskCompletedRequest
+		internalError  error
+	}
+
 	// lockableQueryTaskMap maps query TaskID (which is a UUID generated in QueryWorkflow() call) to a channel
 	// that QueryWorkflow() will block on. The channel is unblocked either by worker sending response through
 	// RespondQueryTaskCompleted() or through an internal service error causing cadence to be unable to dispatch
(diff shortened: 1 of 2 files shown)


service/matching/matchingEngine.go
```diff
@@ -215,13 +228,12 @@ func (e *matchingEngineImpl) AddDecisionTask(
...
 	taskListKind := request.TaskList.Kind
-
-	e.logger.Debug(
-		fmt.Sprintf("Received AddDecisionTask for taskList=%v, WorkflowID=%v, RunID=%v, ScheduleToStartTimeout=%v",
-			request.TaskList.GetName(),
-			request.Execution.GetWorkflowID(),
-			request.Execution.GetRunID(),
-			request.GetScheduleToStartTimeoutSeconds()))
+	e.logger.Debug("Received AddDecisionTask",
+		tag.WorkflowTaskListName(request.TaskList.GetName()),
```
andrewjdawson2016, line 232: I have seen issues with sticky task list before - may consider logging task list type and/or sticky task list name vs base task list name.

andrewjdawson2016 (approved):

service/matching/matchingEngine.go
mantas-sidlauskas: good idea, will add this. Thanks

yycptt: I think it will be helpful to also log the task scheduledID, which can help us identify a certain decision task in the workflow. Otherwise if the workflow is long and have many decision task, it's hard to find the log for one particular decision task. Same suggestion for the AddActivityTask log. 

I am actually making the same change for those log messages when improving the visibility into transfer task processing logic as I need some messaged logged at Info level. :) I will resolve the conflicts once you land the diff.

service/matching/matchingEngine.go
```diff
@@ -215,15 +228,19 @@ func (e *matchingEngineImpl) AddDecisionTask(
...
 	taskListName := request.TaskList.GetName()
 	taskListKind := request.TaskList.Kind
+	taskListType := persistence.TaskListTypeDecision
 
-	e.logger.Debug(
-		fmt.Sprintf("Received AddDecisionTask for taskList=%v, WorkflowID=%v, RunID=%v, ScheduleToStartTimeout=%v",
-			request.TaskList.GetName(),
-			request.Execution.GetWorkflowID(),
-			request.Execution.GetRunID(),
-			request.GetScheduleToStartTimeoutSeconds()))
```
vytautas-karpavicius, line 224: `GetScheduleToStartTimeoutSeconds` was dropped, is this intended?

mantas-sidlauskas: yes, this is intentional

PR #3873, opened 2021-01-06
Add more options to config_template for docker image and disable archival by default

longquanzheng:
**What changed?**
for https://github.com/uber/cadence/issues/3870
1. Disable archival by default in `docker run`
2. Add more options to config_template for docker image

**Why?**
It's impossible now to change or disable config like archival when using `docker run`. That's not a problem in helm chart because Helm chart already override it. https://github.com/banzaicloud/banzai-charts/blob/master/cadence/templates/server-configmap.yaml#L159  However, there are some users not in K8s yet, they are still using `docker run` to start cadence. For those users, most of them need to disable archival to save the disk space.
 

**How did you test it?**
Local test

**Potential risks**
External user that rely on disk based archival from raw docker image will be broken, and they have to explicitly passing in parameter in `docker run` to enable it back.
It's very unlikely that people will rely on that, but we need to clarify it in our release note that from this change, the local file based archival is disabled by default. If anyone wants to enable it, they should pass in the parameters in `docker run` command. Helm charts has disabled it by default.

### DIFF
--- a/common/service/config/archival.go
+++ b/common/service/config/archival.go
@@ -22,7 +22,6 @@ package config
 
 import (
 	"errors"
-
 	"github.com/uber/cadence/common"
 )
 
@@ -50,6 +49,6 @@ func isArchivalConfigValid(
 	URISet := len(domianDefaultURI) != 0
 
 	validEnable := archivalEnabled && URISet && specifiedProvider
-	validDisabled := !archivalEnabled && !enableRead && domainDefaultStatus != common.ArchivalEnabled && !URISet && !specifiedProvider
+	validDisabled := !archivalEnabled && !enableRead && domainDefaultStatus != common.ArchivalEnabled && !URISet
 	return validEnable || validDisabled
 }
--- a/common/service/config/archival.go
+++ b/common/service/config/archival.go
@@ -22,7 +22,6 @@ package config
 
 import (
 	"errors"
-
 	"github.com/uber/cadence/common"
 )
 
--- a/common/service/config/archival.go
+++ b/common/service/config/archival.go
@@ -22,6 +22,7 @@ package config
 
 import (
 	"errors"
+	
 	"github.com/uber/cadence/common"
 )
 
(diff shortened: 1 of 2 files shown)


docker/config_template.yaml
```diff
@@ -194,31 +194,31 @@ services:
...
 
 dcRedirectionPolicy:
-    policy: "noop"
+    policy: {{ default .Env.DC_REDIRECT_POLICY "selected-apis-forwarding" }}
     toDC: ""
 
 archival:
   history:
-    status: "enabled"
+    status: {{ default .Env.HISTORY_ARCHIVAL "disabled" }}
```
longquanzheng, line 214: @andrewjdawson2016 this doesn't work because of https://github.com/uber/cadence/blob/master/common/service/config/archival.go#L53

Background: a customer need to disable archival, they are running pure docker container using our images. 

Do you have any suggestion? One way is to relax the "validDisable" check, not sure if you like that.
Another way is completely remove the archival section for everyone --- I don't think file base archival is useful here as the docker container will be destroyed. but it's a little riscky.

common/service/config/archival.go
```diff
@@ -50,6 +49,6 @@ func isArchivalConfigValid(
 	URISet := len(domianDefaultURI) != 0
 
 	validEnable := archivalEnabled && URISet && specifiedProvider
-	validDisabled := !archivalEnabled && !enableRead && domainDefaultStatus != common.ArchivalEnabled && !URISet && !specifiedProvider
+	validDisabled := !archivalEnabled && !enableRead && domainDefaultStatus != common.ArchivalEnabled && !URISet
```
longquanzheng, line 52: I have to remove this check, otherwise there is no way to workaround in the config_template. 
And I don't think it's necessary to enforce this. LMK if I am wrong.

yycptt (approved):

common/service/config/archival.go
longquanzheng: @andrewjdawson2016 LMK if you have concern about that. We could emit a warn log if you think this is useful to prevent users making mistake in configuration.

longquanzheng: I believe it should be fine :-)  So I am going to merge it in.

PR #3877, opened 2021-01-06
Task processing debug logs

yycptt:
**What changed?**
- Add event logger for logging for transfer/time task lifecycle events.
- Add logs for recordActivity/DecisionTaskStarted calls in history service.
- Add logs for AddActivity/DecisionTask calls in matching.
- Add logs when matching handler returns an internal service error or an uncategorized error.

**Why?**
Help debugging lost task issue.

**How did you test it?**
Tested in staging2

**Potential risks**
No risk when all feature flags are disabled. If enabled, may increase GC pressure due to frequent access of dynamicconfig client for determining if log need to be emitted.

### DIFF
--- a/service/history/task/event_logger.go
+++ b/service/history/task/event_logger.go
@@ -77,7 +77,7 @@ func (e *eventLoggerImpl) AddEvent(
 		details:   details,
 	}
 
-	e.nextEventIdx = (e.nextEventIdx + 1) % len(e.events)
+	e.nextEventIdx = (e.nextEventIdx + 1) % e.maxSize
 }
 
 func (e *eventLoggerImpl) FlushEvents(
--- a/service/history/task/event_logger.go
+++ b/service/history/task/event_logger.go
@@ -1,0 +1,111 @@
+// Copyright (c) 2017-2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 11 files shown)


service/history/task/task.go
```diff
@@ -385,9 +405,16 @@ func (t *taskBase) Ack() {
 		t.scope.RecordTimer(metrics.TaskLatencyPerDomain, time.Since(t.submitTime))
 		t.scope.RecordTimer(metrics.TaskQueueLatencyPerDomain, time.Since(t.GetVisibilityTimestamp()))
 	}
+
+	if t.eventLogger != nil && t.shouldProcessTask && t.attempt != 0 {
+		// only dump events when the task should be processed and has been retried
+		t.eventLogger.FlushEvents("Task processing events")
```
yux0, line 411: Could it happen if the task always Nack?

service/history/task/event_logger.go
```diff
@@ -0,0 +1,111 @@
...
+	eventMsg string,
+	details ...interface{},
+) {
+	e.events[e.nextEventIdx] = event{
+		timestamp: e.timeSource.Now(),
+		message:   eventMsg,
+		details:   details,
+	}
+
+	e.nextEventIdx = (e.nextEventIdx + 1) % len(e.events)
```
yux0, line 80: (e.nextEventIdx + 1) % e.maxSize?

service/history/task/task.go
yycptt: Assumption here is all submitted tasks will eventually be acked, otherwise two possible situations
1. task may be lost during the processing and redispatching logic -> ack level won't move -> alert for task lag will fire
2. task got stuck and keep retrying, alert for task attempts will fire and the reason why it's keep retrying will be logged.

The main purpose for this events is for the case that neither of the above two situations occurs, but a task still appears to be lost or not processed. In that case we can be sure that a task is acked, and we can check its life cycle events.

yux0 (approved):

service/history/task/event_logger.go
yycptt: Updated.

PR #3880, opened 2021-01-09
Fix dynamic config collection logValue function

yycptt:
**What changed?**
Add filter to dynamicconfig key when storing, comparing and emitting logs for retrieved values.

**Why?**
Current logValue function emit a log whenever the value for a dynamicconfig key changes. However, each key may have a set of filters associated with it when retrieving the values (for example domain filter or shardID filter), and the value is expected to be different when the filter values are different and we should not emit the log in this case.

**How did you test it?**
Tested locally.

**Potential risks**
Increase memory cost as we need to store more values. Also each access to dynamicconfig collection will be more expensive as we need to construct the filtered key. But we will also emit much less logs, which will improve the performance.

### DIFF
--- a/common/service/dynamicconfig/config.go
+++ b/common/service/dynamicconfig/config.go
@@ -21,7 +21,9 @@
 package dynamicconfig
 
 import (
+	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


common/service/dynamicconfig/config.go
```diff
@@ -157,346 +166,354 @@ func (c *Collection) GetProperty(key Key, defaultValue interface{}) PropertyFn {
...
+	for _, opt := range opts {
+		opt(m)
+	}
+	for _, opt := range c.filterOptions {
+		opt(m)
+	}
+	return m
+}
+
+func getFilteredKey(
```
longquanzheng, line 507: Nit: how about getFilteredKeyAsString. It will be less confusing when reading the code.

common/service/dynamicconfig/config.go
```diff
@@ -157,346 +166,354 @@ func (c *Collection) GetProperty(key Key, defaultValue interface{}) PropertyFn {
...
-			c.logError(key, err)
+			c.logError(key, filters, err)
 		}
-		c.logValue(key, val, defaultValue, boolCompareEquals)
+		c.logValue(key, filters, val, defaultValue, boolCompareEquals)
 		return val
 	}
 }
+
+func (c *Collection) getFilterMap(opts ...FilterOption) map[Filter]interface{} {
```
longquanzheng, line 495: Nit: toFilterMap(...) GetXxx is generally used when it is got from c *Collection, while this is converting the input paramters into another object

longquanzheng (approved):
lgtm except for some nit comments. Address them as you agree :D

PR #3882, opened 2021-01-09
Add domain name tag in failover metrics

yux0:
**What changed?**
Add domain name tag in failover metrics

**Why?**
To better search domain related metrics

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/history/failover/coordinator.go
+++ b/service/history/failover/coordinator.go
@@ -32,6 +32,7 @@ import (
 	"github.com/uber/cadence/client/history"
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/backoff"
+	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/clock"
 	"github.com/uber/cadence/common/domain"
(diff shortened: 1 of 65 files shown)


service/history/failover/coordinator.go
```diff
@@ -255,6 +259,14 @@ func (c *coordinatorImpl) handleFailoverMarkers(
 		record.shards[shardID] = struct{}{}
 	}
 
+	domainEntry, err := c.domainCache.GetDomainByID(domainID)
```
yycptt, line 262: nit: if all you need is the domainName, consider using the `GetDomainName()` method which can avoid a deep copy of domainEntry.

yycptt (approved):

service/history/failover/coordinator.go
yux0: Good call. will update

yycptt: Let we create an issue for this actually, to use `GetDomainName` whenever possible. The deep copy of domain entry allocated many objects and created quite some GC pressure when I was testing in staging2.

PR #3894, opened 2021-01-13
Add larger test timeout for large test

anish531213:
Some test might take longer than the context deadline so introduced this new test timeout value.


andrewjdawson2016 (approved):

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -56,6 +56,7 @@ type (
 
 var (
 	testContextTimeout = 5 * time.Second
+	largeTestContextTimeout = 10 * time.Second
```
mkolodezny, line 59: nit: is this the case for the Cassandra only(not sql)?

PR #3895, opened 2021-01-14
Handle data corruption error in replication

yux0:
**What changed?**
Update some internal service error to internal data inconsistency error and handle data corruption error in replication.

**Why?**
Those data inconsistency error could block the replication stack.

**How did you test it?**
Added unit tests.

**Potential risks**

### DIFF
--- a/common/persistence/cassandra/cassandraHistoryPersistence.go
+++ b/common/persistence/cassandra/cassandraHistoryPersistence.go
@@ -176,11 +176,11 @@ func (h *nosqlHistoryManager) ReadHistoryBranch(
 
 		switch {
 		case nodeID < lastNodeID:
-			return nil, &types.InternalServiceError{
+			return nil, &types.InternalDataInconsistencyError{
 				Message: fmt.Sprintf("corrupted data, nodeID cannot decrease"),
 			}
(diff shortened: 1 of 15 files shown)


common/persistence/cassandra/cassandraHistoryPersistence.go
```diff
@@ -176,11 +176,11 @@ func (h *nosqlHistoryManager) ReadHistoryBranch(
 
 		switch {
 		case nodeID < lastNodeID:
-			return nil, &types.InternalServiceError{
+			return nil, &types.InternalDataInconsistencyError{
```
yycptt, line 179: There's some more checks in `historyStore.go` regarding corrupted history. Can we also return this inconsistency error if those checks fail?

yux0: Updated those errors in the history store. I think there might be more cases to be converted. I don't plan to convert all of them in this change. We can continue with other changes.

yycptt: Sure thing. I method those in historyStore because I saw them a lot in production logs. We can gradually convert the rest.

yycptt (approved):

PR #3898, opened 2021-01-15
Add a "help" target to the makefile

Groxx:
Since this isn't a universal Makefile trick, I've also switched the default target to `help` so newcomers / random `make`-ers can discover the high-level things the makefile can do.
I'm happy to add more in this first PR, I just grabbed a few obvious ones to serve for starters.

As a concrete example, this is now what running `make` will do:
Screen Shot 2021-01-14 at 7 13 59 PM
The one-liner isn't all that hard to understand and modify, e.g. to change the column width or colors.


yux0:
Thank you for making this change.

yux0 (approved):

PR #3899, opened 2021-01-15
Use GetDomainName instead of GetDomainByID for retrieving domain names

yycptt:
**What changed?**
https://github.com/uber/cadence/issues/3883

**Why?**
https://github.com/uber/cadence/issues/3883

**How did you test it?**
Unit test.

**Potential risks**
No logic change in this diff so the risk is quite low


service/history/task/transfer_active_task_executor.go
```diff
@@ -606,28 +604,24 @@ func (t *transferActiveTaskExecutor) processStartChildExecution(
 		return nil
 	}
 
-	// Get parent domain name
-	var domain string
-	if domainEntry, err := t.shard.GetDomainCache().GetDomainByID(task.DomainID); err != nil {
+	// Get parent domainName name
```
vytautas-karpavicius, line 607: nit: no need to update comment

vytautas-karpavicius (approved):

service/history/task/transfer_active_task_executor.go
yycptt: Thanks for catching. Reverted the change.

PR #3902, opened 2021-01-19
Proto mappers

vytautas-karpavicius:
This PR introduces `internal types` <-> `proto` mappers.

This is constructed in the same fashion as `internal types` <-> `thrift` mappers. For each type we will have `From<Type>` and `To<Type>` functions performing the conversion.

These mappers will be used in the grpc handlers and clients that will come later.

## Context

Since protobuf v3 does not differentiate between primitive zero values and value not being sent on the wire, multiple preparation were done before this PR. It included either:
  - Dropping pointers on primitive values within internal types. Many types are always required and make meaningful use of zero value, therefore pointer was not nessesary on those.
  - Where not possible to do so (missing field on wire has some meaning), proto IDLs were adjusted. Either by wrapping field with wrapper message or grouping multiple optional fields into an optional message.

This means that generated proto types can be directly mapped to internal types without involving pointer dereference and assigning zero default values in nil cases. This in turns mean that we are not changing required/optional semantics within this layer. This can be validated that none of the mappers call `t.Get<Field>` helpers of internal types.

## Testing

There is **99.7%** unit test coverage for all conversion functions. All unit tests are constructed to test the following invariant:
```
    assert.Equal(t, initialValue, ToType(FromType(initialValue)))
```

To construct initial values another `testdata` package is introduces within `common/types`. It contains all types initialized with all fields to some arbitrary non-zero value. It ensures, that fields are not forgotten/dropped during mapper conversion. Each test value is named the same as its type.

These test values were put into a different package, as they could be reused in other unit tests that just need some non-zero value of that type, but does not really case about specific values set.


andrewjdawson2016 (approved):

Groxx:
fwiw I'm waiting to review until #3905 is done, since codegen changes for v2 probably mean this changes as well

common/types/shared.go
```diff
@@ -2929,6 +2929,8 @@ const (
 	EncodingTypeThriftRW EncodingType = iota
 	// EncodingTypeJSON is an option for EncodingType
 	EncodingTypeJSON
+	// EncodingTypeProto3 is an option for EncodingType
+	EncodingTypeProto3
```
Groxx, line 2933: doesn't seem to be reflected in the `String()` / `UnmarshalText(...)` implementations above.
generally speaking, it may also be worth adding an 'undefined' zero value, for the same kinds of reason as proto3.  better to detect invalid init immediately rather than e.g. after a database round-trip.

PR #3903, opened 2021-01-19
Makefile cleanup / thrift revamp / gobin removed

Groxx:
A few "safe" / simple changes to the Makefile, in prep for future changes.

Specifically, this includes:
- made `// ` (with trailing space) license comments into `//` so they do not conflict with `goimports`
- moving "build" to ".build" so it's hidden and automatically ignored by most things.
  - in particular this _slightly_ improves the odds that the first build (no build-folder at parse time) and the Nth build (build folder already exists) behave the same, as e.g. `ALL_SRC` will implicitly ignore it now.
- `ifndef` -> `?=` because that's basically what they're doing.
  - this does mean that `go env GOOS` may run multiple times, as `?=` acts like `=` and not `:=`, but it's fast enough to not be noticeable (single-digit ms).
  - I did not change `TEST_TAG`, as the `override` is needed to modify the var when it's set via `make TEST_TAG=adsf targets...`
- `ALL_SRC` now uses a find query that does not even descend into vendor / etc, rather than filtering them out afterward.
  - this is visibly faster when the files are not in caches, and will essentially remain fast regardless of how large the project grows.
  - future changes can just modify the find filter (it's pretty simple) or add a `grep -v` to the end like before.
- got rid of `TOOLS_SRC` because it was completely incorrect: the tools bins depend on cadence source outside of the tools folder.
  - the copyright bin is the only exception, and its target dependency doesn't use `TOOLS_SRC` anyway.
- revamped how thrift sources are generated:
  - `*.thrift` files are found rather than listed.  guaranteed to be complete + smaller = perhaps better than a handwritten list.
  - switched to using a "normal" multi-target rule rather than define+foreach+eval+call.  hopefully this is simpler to understand and modify.
  - `--no-recurse` + separate target files = you can safely generate thrift in parallel!  try `touch idls/thrift/cadence.thrift` + `make .gen/go/cadence/cadence.go -j8` to see this in action.
    - `make -j8` in general does not yet work though, as other dependencies are not declared correctly, e.g. copyright is not run when appropriate.  that's a goal for later commits...
- got rid of `gobin`, now using go.mod
  - this approach keeps runtime-library-versions in sync with codegen-library-versions, which is important for some libraries, like yarpc and mockgen.
    - other tools _could_ be moved to a separately-versioned tools.mod for greater flexibility, but that is not done here.
  - I have not yet adapted the protobuf changes, but will do that soon
  - to update tools, `go get -u` them, like normal.  to add tools, add them to `cmd/server/tools.go` (this file will probably be moved soon)
- some changes to protoc to align it with everything else

After pulling, you'll need to manually delete your `build` folder.  Generated code / copyright headers / etc should all be identical after this change, though the actual golint / yarpc / etc binaries are not necessarily _identical_ due to incomplete versioning before.


Makefile
```diff
@@ -47,43 +27,80 @@ GO_BUILD_LDFLAGS          := $(shell $(GO_BUILD_LDFLAGS_CMD) LDFLAG)
...
+THRIFT_SRCS := $(shell find idls -name '*.thrift')
+# concrete targets to build / the "sentinel" go files that need to be produced per thrift file.
+# idls/thrift/thing.thrift -> thing.thrift -> thing -> .gen/go/thing/thing.go
+THRIFT_GEN_SRC := $(foreach tsrc,$(basename $(subst idls/thrift/,,$(THRIFT_SRCS))),$(THRIFT_GENDIR)/$(tsrc)/$(tsrc).go)
+
+# how to generate each thrift file.
+# note that each generated file depends on ALL thrift files - this is necessary because they can import each other.
+$(THRIFT_GEN_SRC): $(THRIFT_SRCS) $(THRIFTRW_BIN)
+	@# .gen/go/thing/thing.go -> thing.go -> "thing " -> thing -> idls/thrift/thing.thrift
+	$(THRIFTRW_BIN) --plugin=yarpc --pkg-prefix=$(PROJECT_ROOT)/$(THRIFT_GENDIR) --out=$(THRIFT_GENDIR) --no-recurse idls/thrift/$(strip $(basename $(notdir $@))).thrift
```
vytautas-karpavicius, line 81: nit: maybe split one options per line, for easier readability?

Makefile
```diff
@@ -47,43 +27,80 @@ GO_BUILD_LDFLAGS          := $(shell $(GO_BUILD_LDFLAGS_CMD) LDFLAG)
...
+		-o -path './.*' \
+		-o -path '*/mocks*' \
+	\) \
+	-prune \
+	-o -name '*.go' -print \
+)
+
+# make sure a sentinel thrift-generated file is in ALL_SRC, so thrift is (re)generated if necessary
+ALL_SRC += $(THRIFT_GEN_SRC)
+FMT_SRC := $(filter-out .gen/%, $(ALL_SRC))
```
vytautas-karpavicius, line 100: Does this mean we will not run goimport for generated files?

Makefile
```diff
@@ -185,34 +194,23 @@ cadence-canary: $(ALL_SRC)
 
 go-generate-format: go-generate fmt
 
-go-generate:
-	GO111MODULE=off go get -u github.com/myitcv/gobin
-	GOOS= GOARCH= gobin -mod=readonly github.com/golang/mock/mockgen
-	GOOS= GOARCH= gobin -mod=readonly github.com/dmarkham/enumer
+go-generate: $(MOCKGEN_BIN) $(ENUMER_BIN)
```
vytautas-karpavicius, line 197: This is probably not related to this refactoring, but I have tried running `make go-generate` which regenerates mock files without licence headers and then runs `copyright` with `--verifyOnly` which then complains. Making this command kind of useless.

This is not true for generated thrift files though.  Looks like licengen handles `.gen` files differently: https://github.com/uber/cadence/blob/makethrift/cmd/tools/copyright/licensegen.go#L238

vytautas-karpavicius (approved):
Overall looks great to me!

Makefile
```diff
@@ -1,38 +1,18 @@
...
-  idls/thrift/shared.thrift \
-  idls/thrift/admin.thrift \
-  idls/thrift/sqlblobs.thrift \
-  idls/thrift/checksum.thrift \
-
-PROGS = cadence
 TEST_TIMEOUT = 20m
 TEST_ARG ?= -race -v -timeout $(TEST_TIMEOUT)
-BUILD := ./build
+BUILD := .build
```
vytautas-karpavicius, line 15: You will also need to update https://github.com/uber/cadence/blob/master/.buildkite/pipeline.yml
to pick code coverage report from the new location.

Makefile
Groxx: can do - I'll suppress and add an echo then, it's sorta space-consuming in the default output

Makefile
Groxx: when I was running all this, that seemed to be the case currently.  if I ran goimports on those files, it moved imports around a bit.
possibly that's just a different version of goimports though, I'm not quite sure.  I'll look into it a bit more and find out exactly what's happening.

generally speaking though I don't feel strongly either way.  not-running it is potentially simpler (no need to consider it for task sequencing), but running it does avoid accidental IDE tweaks.  would you / we prefer to goimports it all?

Makefile
Groxx: yeah, this command will go away eventually I think.  it's fast enough to make thrift/proto re-generate _correctly_ when needed (which is not part of this PR), which makes an explicit command rather pointless.

as far as the `copyright` tool: what is the _desired_ behavior?  is there any reason to _not_ run it all the time without `--verifyOnly` so it prepends it anywhere it's missing?

vytautas-karpavicius: Yeah, I usually regenerated mock files as needed (within the IDE). Plus this runs very long anyway.

Makefile
vytautas-karpavicius: Since we already do modifications to generated files by adding license header, I think it also makes sense to modify it further with goimports. So yeah, I would vote to run goimport on generated files as well.

Makefile
Groxx: ah, right, this is mockgen not protoc-gen.
yeah, that makes sense (running as-needed and where-needed).  if the mocks are out of date they'll fail tests anyway, so there's no real harm in leaving it manual.

PR #3907, opened 2021-01-20
Proto domain api improvements

vytautas-karpavicius:
**What changed?**
This is a follow-up to #3863 focused on few improvements on Domain API. None of them are strictly necessary, we can discuss whether to take none, some or all of them.
- Archival status and URI wrapped into `Archival` message https://github.com/uber/cadence/commit/f9be08345c0bc4c9e0694b94fe640992d81e7c2f
- Flattened `UpdateDomainInfo` into `UpdateDomainRequest` as it is not used anywhere else. https://github.com/uber/cadence/commit/7c506386f25eead9720090778639f0f553863508
- Add common `Domain` message which can be reused where complete domain is returned, in: `UpdateDomainResponse`, `DescribeDomainResponse` and `ListDomainsResponse` https://github.com/uber/cadence/commit/e32a6f2044c6d25f8b42582d6c5b9ff9b412cba6
- Move `security_token` to first position, to better indicate that these requests require authentication. https://github.com/uber/cadence/commit/77c024e08f1f297c5c1787b6904679126586105c
- Use `google.protobuf.FieldMask` in  `UpdateDomainRequest` to specify which fields should be updated. This is used to distinguish between "do not update" and "reset to zero value" semantics. https://github.com/uber/cadence/commit/6544b3398448ac70ec937a552160fbb644b67fac
- Finally flatten `DomainInfo`, `DomainConfiguration` and `DomainReplicationConfiguration` structures. Those are related how we store various domain attributes internally. Exposing that as part of API is not necessary and could be simplified as basic domain properties. https://github.com/uber/cadence/commit/0714e0967422ef2d668068d39b84603b8cf1e232

**Why?**
To simplify DomainAPI and make it more consistent and idiomatic.

**How did you test it?**

**Potential risks**


Groxx (approved):
Looks great overall.  The same-typing of `Archival` is the only thing that gives me a _bit_ of pause, as it seems possible that those could diverge.... but it does feel like options for one would be added to the other eventually even if not immediately.

vytautas-karpavicius:
> Looks great overall. The same-typing of `Archival` is the only thing that gives me a _bit_ of pause, as it seems possible that those could diverge.... but it does feel like options for one would be added to the other eventually even if not immediately.

Reverted `Archival` wrapper for now. While it is nicer, it is probably safer not to do it.

yux0 (approved):

proto/public/uber/cadence/api/v1/domain.proto
```diff
@@ -30,33 +30,29 @@ option java_package = "com.uber.cadence.api.v1";
...
-  string visibility_archival_uri = 6;
+  google.protobuf.Duration workflow_execution_retention_period = 7;
+  BadBinaries bad_binaries = 8;
+  ArchivalStatus history_archival_status = 9;
+  string history_archival_uri = 10;
+  ArchivalStatus visibility_archival_status = 11;
+  string visibility_archival_uri = 12;
+  string active_cluster_name = 13;
+  repeated ClusterReplicationConfiguration clusters = 14;
+  int64 failover_version = 15;
```
yycptt, line 48: This field looks like an internal implementation detail. Shall we expose this info to users?

mkolodezny (approved):

andrewjdawson2016 (approved):

proto/public/uber/cadence/api/v1/domain.proto
vytautas-karpavicius: This is currently used within `checkOngoingFailover`. But in general I agree, this looks like an implementation detail.

PR #3909, opened 2021-01-21
Fix scan unsupported workflow in SQl DB

yux0:
**What changed?**
1. Fix the result ordering of SQL
2. Fix version histories deserialization

**Why?**

**How did you test it?**
Local testing

**Potential risks**

### DIFF
--- a/common/persistence/sql/sqlTaskManager.go
+++ b/common/persistence/sql/sqlTaskManager.go
@@ -267,6 +267,7 @@ type taskListPageToken struct {
 // ListTaskList lists tasklist from DB
 // DomainID translates into byte array in SQL. The minUUID is not the minimum byte array.
 // This API could return incomplete result set.
+// https://github.com/uber/cadence/issues/3911
 func (m *sqlTaskManager) ListTaskList(
 	ctx context.Context,
 	request *persistence.ListTaskListRequest,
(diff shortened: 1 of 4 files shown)


yycptt (approved):

yycptt (commented):
Can we add some comment for the new contract of list executions?

yycptt (commented):
It looks like `minUUID` is also used by ListTaskList, will it cause the same issue there?

yux0:
> Can we add some comment for the new contract of list executions?
Added comments about the behavior.

> It looks like `minUUID` is also used by ListTaskList, will it cause the same issue there?

I think so. Added a comment on this API. This API is only used in Scanner.

common/persistence/sql/sqlTaskManager.go
```diff
@@ -264,6 +264,9 @@ type taskListPageToken struct {
 	TaskType int64
 }
 
+// ListTaskList lists tasklist from DB
+// DomainID translates into byte array in SQL. The minUUID is not the minimum byte array.
+// This API could return incomplete result set.
```
yycptt, line 269: Shall we create an issue for this? Looks like this is something we need to fix.

yux0: https://github.com/uber/cadence/issues/3911

PR #3913, opened 2021-01-22
Version goveralls, remove unused go bins from docker setup

Groxx:
A few things seem unused / unnecessary to bake into scripts / docker, as now these are versioned by the makefile.

I also noticed that the "host" folder has code-being-tested and tests, but wasn't included in coverage.
So that's now included.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,11 @@
 MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
+# a literal space value, for makefile purposes
+SPACE :=
+SPACE +=
+COMMA := ,
+
 .PHONY: git-submodules test bins clean cover cover_ci help
 default: help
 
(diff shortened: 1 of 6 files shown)


Groxx:
huh.  seems like buildkite / github gets confused if you delete a commit from a branch.
oh well.

mantas-sidlauskas (approved):
lgtm

mkolodezny (approved):

Makefile
```diff
@@ -97,13 +105,13 @@ $(BIN)/protoc: | $(BIN)
 	rm $(BIN)/protoc.zip
 
 # any generated file - they all depend on each other / are generated at once, so any will work
-PROTO_GEN_SRC = .gen/proto/admin/v1/service.pb.go
+PROTO_GEN_SRC = ./.gen/proto/admin/v1/service.pb.go
```
Groxx, line 108: makes this consistent with the `find` results, so `sort` will put them where they belong / deduplicate.

PR #3916, opened 2021-01-22
Drop pointers for domain name/id fields

vytautas-karpavicius:
**What changed?**
Drop pointers for domain name/id fields within internal types. This is mostly mechanical change touching only those fields which are always required. This removes `common.StringPtr` helpers everywhere and modifies checks for empty domain instead of nil in handler validation. This is equivalent to previous logic, because incoming nil from request was masked everywhere with `GetDomain` helper, which returned empty string anyway. Now this is done at the mapping layer.

There were a few cases were domain id could be optional. In particular:
- `ParentExecutionInfo.DomainUUID` / `ParentExecutionInfo.Domain`
- `WorkflowExecutionInfo.ParentDomainID`
- `WorkflowExecutionStartedEventAttributes.ParentWorkflowDomain`
- `ActivityTaskScheduledEventAttributes.Domain`
In those cases there were additional logic attached to `!= nil` case. I will check if it makes sense to use zero value for those and address that in a separate PR later if needed.

**Why?**

**How did you test it?**

**Potential risks**


andrewjdawson2016 (approved):

host/integration_test.go
```diff
@@ -3157,7 +3157,7 @@ func (s *integrationSuite) TestStickyTimeout_NonTransientDecision() {
 			// send a signal on third failure to be buffered, forcing a non-transient decision when buffer is flushed
 			/*if failureCount == 3 {
 				err := s.engine.SignalWorkflowExecution(createContext(), &types.SignalWorkflowExecutionRequest{
-					Domain:            common.StringPtr(s.domainName),
+					Domain:            s.domainName,
```
Groxx, line 3160: commented out?

not relevant for this diff, but is this something we should restore?

service/history/replication/task_ack_manager.go
```diff
@@ -470,7 +470,7 @@ func (t *taskAckManagerImpl) generateFailoverMarkerTask(
 		TaskType:     types.ReplicationTaskType.Ptr(types.ReplicationTaskTypeFailoverMarker),
 		SourceTaskID: common.Int64Ptr(taskInfo.GetTaskID()),
 		FailoverMarkerAttributes: &types.FailoverMarkerAttributes{
-			DomainID:        common.StringPtr(taskInfo.GetDomainID()),
```
Groxx, line 473: some of these original lines are a bit facepalm-ridden :|
e.g. this one takes a nullable `*string`, converts it to a non-null `string` defaulting to `""`, then gets a pointer back to `*string`.  effectively just forgetting if the original was null or not.

well.  this is why I like this PR, I suppose.

common/types/mapper/thrift/admin.go
```diff
@@ -900,7 +900,7 @@ func FromAdminDescribeWorkflowExecutionRequest(t *types.AdminDescribeWorkflowExe
 		return nil
 	}
 	return &admin.DescribeWorkflowExecutionRequest{
-		Domain:    t.Domain,
+		Domain:    &t.Domain,
```
Groxx, line 903: same kind of nullable-pointer to non-null-pointer note as in the wid/rid diff.

Groxx (approved):
(same as on wid/rid PR)

Well.  The nullable pointer -> non-null pointer changes in the mappers are quite hard to trace and verify... so I'm not sure what can be done beyond "yolo, test it in prod".  or spend an incredible amount of time trying to verify it all.

wid/rid/domain all seem pretty likely to always be present where required, so odds of this working are pretty good I think.  works for me 👍

PR #3917, opened 2021-01-22
Drop pointer for worflow / run id fields

vytautas-karpavicius:
**What changed?**
Drop pointers for workflow id and run id fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


common/persistence/sql/sqlVisibilityStore.go
```diff
@@ -250,7 +250,7 @@ func (s *sqlVisibilityStore) GetClosedWorkflowExecution(
 	rows, err := s.db.SelectFromVisibility(ctx, &sqlplugin.VisibilityFilter{
 		DomainID: request.DomainUUID,
 		Closed:   true,
-		RunID:    execution.RunID,
+		RunID:    &execution.RunID,
```
Groxx, line 253: I'll resolve this if it turns out safe as I read more: this has a semantic change from allowing nulls to pass through -> no more nulls.  does that matter?

common/types/errors.go
```diff
@@ -101,12 +101,8 @@ func (err RetryTaskV2Error) Error() string {
 	if err.DomainID != nil {
 		printField(sb, "DomainID", *err.DomainID)
 	}
-	if err.WorkflowID != nil {
-		printField(sb, "WorkflowID", *err.WorkflowID)
-	}
-	if err.RunID != nil {
-		printField(sb, "RunID", *err.RunID)
```
Groxx, line 108: obviously minor / a passing thought: may be worth putting nil checks into `printField` so you can pass all (top-level) nils straight to it without `if != nil` checks.

common/persistence/sql/sqlVisibilityStore.go
Groxx: ah.  this one's fine, it'd be useless to not pass at least a runid.  great.

on further reading though: maybe this func's args are just too imprecise? it accepts but ignores domain and workflow-id, and... _overwrites_ the workflowid on the returned data with whatever was passed in?  that's weird looking.

service/frontend/thriftHandler_test.go
```diff
@@ -182,7 +183,7 @@ func TestThriftHandler(t *testing.T) {
 	t.Run("ResetWorkflowExecution", func(t *testing.T) {
 		h.EXPECT().ResetWorkflowExecution(ctx, &types.ResetWorkflowExecutionRequest{}).Return(&types.ResetWorkflowExecutionResponse{}, internalErr).Times(1)
 		resp, err := th.ResetWorkflowExecution(ctx, &shared.ResetWorkflowExecutionRequest{})
-		assert.Equal(t, shared.ResetWorkflowExecutionResponse{}, *resp)
+		assert.Equal(t, shared.ResetWorkflowExecutionResponse{RunId: common.StringPtr("")}, *resp)
 		assert.Equal(t, expectedErr, err)
 	})
```
Groxx, line 188: pretty much all of these tests don't reflect what happens on error - the implementations I can see return `nil, err`, not pointers to zero-valued responses.  that seems to be the source of the now-required runid -> `&""` in the response.

are these tests doing anything useful, aside from double-checking that types compile?  or should we just delete them?

service/frontend/workflowHandler.go
```diff
@@ -1941,11 +1941,11 @@ func (wh *WorkflowHandler) GetWorkflowExecutionHistory(
 		if err != nil {
 			return nil, wh.error(errInvalidNextPageToken, scope, getWfIDRunIDTags(wfExecution)...)
 		}
-		if execution.RunID != nil && execution.GetRunID() != token.RunID {
+		if execution.RunID != "" && execution.GetRunID() != token.RunID {
```
Groxx, line 1944: hmm.  these handlers are... rather painful to read.  there's a lot of:
- random flipping between nil and zero-value-getters for the same fields
- checks of .GetSomething() values before checking if the request itself is nil, which is also an error
- checks like this, which probably belong a lot earlier in the method to ensure the request is sane before doing things.  e.g. this appears to be relevant _immediately_ after the `if !getRequest.GetSkipArchival() {`, but it's nearly 60 lines later.

and there are _four thousand lines_ like this in this file alone.

wat.

service/history/historyEngine.go
```diff
@@ -3046,7 +3046,7 @@ func getWorkflowAlreadyStartedError(errMsg string, createRequestID string, workf
 	return &types.WorkflowExecutionAlreadyStartedError{
 		Message:        common.StringPtr(fmt.Sprintf(errMsg, workflowID, runID)),
 		StartRequestID: common.StringPtr(fmt.Sprintf("%v", createRequestID)),
-		RunID:          common.StringPtr(fmt.Sprintf("%v", runID)),
+		RunID:          fmt.Sprintf("%v", runID),
```
Groxx, line 3049: weird.  why not just runID?  (not relevant to this change, so leave it if you want, just odd)

service/history/historyEngine_test.go
```diff
@@ -187,8 +187,8 @@ func (s *engineSuite) TestGetMutableStateSync() {
 	ctx := context.Background()
 
 	workflowExecution := types.WorkflowExecution{
-		WorkflowID: common.StringPtr("test-get-workflow-execution-event-id"),
-		RunID:      common.StringPtr(constants.TestRunID),
+		WorkflowID: "test-get-workflow-execution-event-id",
+		RunID:      constants.TestRunID,
 	}
```
Groxx, line 192: > 528  service/history/historyEngine_test.go 

bleh.
I wonder if it'd be worth cleaning up the extreme duplication in so many of these test files... maybe after all the type changing is done?

andrewjdawson2016 (approved):

service/history/thriftHandler_test.go
```diff
@@ -198,7 +199,7 @@ func TestThriftHandler(t *testing.T) {
 	t.Run("ResetWorkflowExecution", func(t *testing.T) {
 		h.EXPECT().ResetWorkflowExecution(ctx, &types.HistoryResetWorkflowExecutionRequest{}).Return(&types.ResetWorkflowExecutionResponse{}, internalErr).Times(1)
 		resp, err := th.ResetWorkflowExecution(ctx, &hist.ResetWorkflowExecutionRequest{})
-		assert.Equal(t, shared.ResetWorkflowExecutionResponse{}, *resp)
+		assert.Equal(t, shared.ResetWorkflowExecutionResponse{RunId: common.StringPtr("")}, *resp)
 		assert.Equal(t, expectedErr, err)
 	})
```
Groxx, line 204: same kind of note as `service/frontend/thriftHandler_test.go`

common/types/mapper/thrift/admin.go
```diff
@@ -1051,8 +1051,8 @@ func FromResendReplicationTasksRequest(t *types.ResendReplicationTasksRequest) *
 	}
 	return &admin.ResendReplicationTasksRequest{
 		DomainID:      t.DomainID,
-		WorkflowID:    t.WorkflowID,
-		RunID:         t.RunID,
+		WorkflowID:    &t.WorkflowID,
+		RunID:         &t.RunID,
```
Groxx, line 1055: there are only a few of these "nullable pointer -> non-null pointer"-containing files, but I think I need to read them more carefully.  it seems like there's a decent chance they're correct/safe, but it's definitely a semantic change.

my brain needs a bit of a break after reading everything else though, so I'll come back to this later :)

service/frontend/adminHandler.go
```diff
@@ -247,7 +247,7 @@ func (adh *adminHandlerImpl) DescribeWorkflowExecution(
 		return nil, adh.error(err, scope)
 	}
 
-	shardID := common.WorkflowIDToHistoryShard(*request.Execution.WorkflowID, adh.numberOfHistoryShards)
+	shardID := common.WorkflowIDToHistoryShard(request.Execution.WorkflowID, adh.numberOfHistoryShards)
 	shardIDstr := string(shardID)
```
Groxx, line 251: I noticed on the domain-id diff: https://github.com/uber/cadence/blob/master/service/frontend/adminHandler.go#L940-L948
lower in this file, you can drop the `Get...` calls too

common/types/errors.go
vytautas-karpavicius: It would be nicer. 

However `printField` accepts `interface{}` thus we can not simply do `if != nil` inside. It would require some some reflection to check if value within interface is nil. Not sure if this is  worth it.

service/history/historyEngine.go
vytautas-karpavicius: Yep, fixed it.

service/frontend/adminHandler.go
vytautas-karpavicius: I agree with all you comments, but as you can see, this PR is already pretty big. I wanted to avoid additional changes like these to keep it mostly mechanical, even though such simplifications are tempting. 

Once this is merged we can do another pass to drop `Get...` helpers as well, further simplifying our code.

service/frontend/thriftHandler_test.go
vytautas-karpavicius: Yes, these needs improvement. I was planning to do something about them. For now they just check that thrift handler forwards the call to the underlying handler.

common/types/errors.go
Groxx: ah, right.  typed nils.
yea, probably not worth it.

service/frontend/adminHandler.go
Groxx: yep, totally agreed 👍

common/types/mapper/thrift/admin.go
Groxx: admin's good - everything that uses this clearly requires the fields

common/types/mapper/thrift/replicator.go
```diff
@@ -302,8 +302,8 @@ func FromHistoryTaskV2Attributes(t *types.HistoryTaskV2Attributes) *replicator.H
 	return &replicator.HistoryTaskV2Attributes{
 		TaskId:              t.TaskID,
 		DomainId:            t.DomainID,
-		WorkflowId:          t.WorkflowID,
-		RunId:               t.RunID,
+		WorkflowId:          &t.WorkflowID,
+		RunId:               &t.RunID,
```
Groxx, line 306: woo.  the call chains involving this go through quite a lot of code :|

Groxx (approved):
Well.  The nullable pointer -> non-null pointer changes in the mappers are quite hard to trace and verify... so I'm not sure what can be done beyond "yolo, test it in prod".  or spend an incredible amount of time trying to verify it all.

wid/rid/domain all seem pretty likely to always be present where required, so odds of this working are pretty good I think.  works for me 👍

PR #3918, opened 2021-01-22
add server side metrics for activity & decision poller counts for task lists

mkolodezny:
**What changed?**
subj

**Why?**
currently we can see the pollers as of now, but it’s just serving data from memory. So we need that as a metric so that we can look back

**How did you test it?**
tested locally


service/matching/taskListManager.go
```diff
@@ -166,7 +165,15 @@ func newTaskListManager(
...
 	}
-
+	var taskListTypeTag metrics.Tag
+	if taskList.taskType == persistence.TaskListTypeActivity {
+		taskListTypeTag = metrics.TaskListTypeTag("activity")
+	} else {
+		taskListTypeTag = metrics.TaskListTypeTag("decision")
+	}
+	tlMgr.pollerHistory = newPollerHistory(func(newSize int) {
+		tlMgr.metricScope().Tagged(taskListTypeTag).UpdateGauge(metrics.PollerPerTaskListCounter, float64(newSize))
```
yycptt, line 175: probably create the tagged scope outside the inline func to avoid creating a new scope every time the updateFunc is called.

Looks like we can update the `metricScope()` func to return a scope with taskListType tag and let other metrics benefit from the change? We don't do this to avoid breaking existing dashboards/alerts?

service/matching/taskListManager.go
```diff
@@ -166,7 +165,15 @@ func newTaskListManager(
 			metrics.MatchingTaskListMgrScope,
 		))
 	}
-
+	var taskListTypeTag metrics.Tag
+	if taskList.taskType == persistence.TaskListTypeActivity {
+		taskListTypeTag = metrics.TaskListTypeTag("activity")
+	} else {
+		taskListTypeTag = metrics.TaskListTypeTag("decision")
```
yycptt, line 172: nit: define global variables for those two tags somewhere in the matching package?

common/cache/cache.go
```diff
@@ -102,6 +105,9 @@ type SimpleOptions struct {
 // deletion, Cache calls go f(i)
 type RemovedFunc func(interface{})
 
+// UpdatedFunc is a type for notifying applications when the cache item was put/deleted
+type UpdatedFunc func(newSize int)
```
yycptt, line 109: I guess three things confuse me a little bit when reading the interface.
- The name. I was expecting the function will be called if an item's value was updated, but seems like it's not the case from the implementation. Currently, it's more like a `SizeUpdatedFunc`.
- I think the update function will be more useful if the callback call also gets the new value and probably the key.  `newSize` might be useful but for this particular use case of emitting metrics, ignoring the race condition and get the size using `Size()` method seems also acceptable.
- On item deletion, both `RemovedFunc` and `UpdatedFunc` will be called, we probably should make it clear in the comment in which order are those functions invoked, update first, remove first or maybe in parallel.

yycptt: - The name is bit tricky. Since we are adding this new function callback, I still think we should make the behavior more general/clear in the beginning and not only call the function when size is changed because: 1. it's in the common package and other places can also benefit from it. 2. I don't think that require many code change.
- Regarding parameter, one idea from @yux0 is that we can make it a struct so that it easier to make modifications later.
- Undefined order is probably fine, I just think we need to make it clear in the comments.

common/cache/cache.go
```diff
@@ -102,6 +105,16 @@ type SimpleOptions struct {
 // deletion, Cache calls go f(i)
 type RemovedFunc func(interface{})
 
+// OnCacheUpdatedFunc is a type for notifying applications when the cache item was put/deleted
```
yycptt, line 108: nit: probably also update the comment as now the func will also be called when value is updated.

yycptt (approved):

PR #3927, opened 2021-01-26
Drop pointer for cluster name fields

vytautas-karpavicius:
**What changed?**
Drop pointer for cluster name fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


service/frontend/adminHandler.go
```diff
@@ -579,7 +579,7 @@ func (adh *adminHandlerImpl) GetReplicationMessages(
 	if request == nil {
 		return nil, adh.error(errRequestNotSet, scope)
 	}
-	if request.ClusterName == nil {
+	if request.ClusterName == "" {
```
vytautas-karpavicius, line 582: Currently this is then only place that has explicit check for nil. This may not be valid change if we allow empty cluster name. 

This eventually goes down to [ClusterReplicationLevel](https://github.com/uber/cadence/blob/master/service/history/shard/context.go#L374). It is not validated anywhere and could be empty string.

From what I can tell, the request for it is only constructed in [taskFetcher](https://github.com/uber/cadence/blob/master/service/history/replication/task_fetcher.go#L287). It in turn get current cluster name from server config, which also does not seem to be explicitly validated against empty string. However there is an additional lookup when constructing [metadata](https://github.com/uber/cadence/blob/master/common/cluster/metadata.go#L114). It should not be empty here as it would result in invalid config file.

Given this reasoning I thing it should be safe to change this check.

Groxx: yeah, I was broadly wondering (for all of these, but for this more-so than previous ones) "when might this be nil in a valid request".

seems like cluster-name (as it's fully internally controlled?  I assume?) should be pretty safe.  Or if it errors, it'll be quick to identify and fix, and not really impact users.

Groxx (approved):

PR #3928, opened 2021-01-26
Drop pointer on request id fields

vytautas-karpavicius:
**What changed?**
Drop pointer for request id fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


service/history/historyEngine.go
```diff
@@ -2084,11 +2084,8 @@ func (e *historyEngineImpl) RequestCancelWorkflowExecution(
 			isCancelRequested, cancelRequestID := mutableState.IsCancelRequested()
 			if isCancelRequested {
 				cancelRequest := req.CancelRequest
-				if cancelRequest.RequestID != nil {
-					requestID := *cancelRequest.RequestID
-					if requestID != "" && cancelRequestID == requestID {
-						return updateWorkflowWithNewDecision, nil
-					}
+				if cancelRequest.RequestID != "" && cancelRequest.RequestID == cancelRequestID {
```
vytautas-karpavicius, line 2087: This is the only place that has `!= nil` logic attached. However this can be simplified since `nil` will be mapped to empty string in mapping layer.

Groxx: I'm guessing `""` isn't a "valid" requestID anyway?

Groxx (approved):

service/history/historyEngine.go
vytautas-karpavicius: Generally RequestID can be empty. At least it is not validated against, nor it is set it in all client requests.
However in all cases it used via GetRequestID helper, which turns nil to empty string anyway. Only this place checks agains nil directly, but this can be changed to equivalent check I did.

PR #3929, opened 2021-01-26
Emit number of task list managers started

mantas-sidlauskas:
**What changed?**
During creation and deletion of Task List Manager gauge metric is emitted with a count of Task List Managers active

**Why?**
This will expose a number of task list managers created per matching instance

**How did you test it?**

**Potential risks**

### DIFF
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -203,6 +203,10 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 		logger.Info("Task list manager state changed", tag.LifeCycleStartFailed, tag.Error(err))
 		return nil, err
 	}
+	e.metricsClient.Scope(metrics.MatchingTaskListMgrScope).UpdateGauge(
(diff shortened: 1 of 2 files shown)


service/matching/matchingEngine.go
```diff
@@ -203,6 +203,10 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 		logger.Info("Task list manager state changed", tag.LifeCycleStartFailed, tag.Error(err))
 		return nil, err
 	}
+	e.metricsClient.Scope(metrics.MatchingTaskListMgrScope).UpdateGauge(
+		metrics.TaskListManagersGauge,
+		float64(len(e.taskLists)),
```
Groxx, line 208: this is a racy read, since this is outside the locked scope

mantas-sidlauskas: yep, missed that. thanks!

yycptt (approved):

PR #3931, opened 2021-01-26
Drop pointer for timerid fields

vytautas-karpavicius:
**What changed?**
Drop pointer for timerid fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


host/integration_test.go
```diff
@@ -3722,7 +3722,7 @@ func (s *integrationSuite) TestCancelTimer() {
 			return nil, []*types.Decision{{
 				DecisionType: types.DecisionTypeStartTimer.Ptr(),
 				StartTimerDecisionAttributes: &types.StartTimerDecisionAttributes{
-					TimerID:                   common.StringPtr(fmt.Sprintf("%v", TimerID)),
+					TimerID:                   fmt.Sprintf("%v", TimerID),
```
Groxx, line 3725: could probably change timerid on 3711 to a string.  but also fine to leave it 👍

Groxx (approved):

host/integration_test.go
vytautas-karpavicius: Fixed it.

PR #3934, opened 2021-01-27
Sanitize common metrics tag values, allow dashes

Groxx:
The existence of a sanitizer implies that there are *un*safe values, so we should probably always sanitize.  Many of these values are user-controllable anyway, leaving them untouched may not be a good idea.
So now everything's sanitized.  I've excluded the kafka one simply because the current builder + sanitizer are guaranteed to not have any effect.

Of note:
- `sanitizer` was only used in this file, so I've moved it.
- `-` may be reasonable to allow for values, but it's a slightly larger semantic change and is not necessarily desirable.
- `.` I'm *intentionally excluding*, as it's used as a conceptual-group separator, e.g. as in canary/metrics.go.
  Since these tag APIs seem more focused on single-"thing" values, separators don't seem like they should be allowed.
  I've left a comment in code for this, for future potential-modifiers.

The main risk here is that some stats could change, e.g. domains / instances / etc with `.`s or other blocked characters in their names will now have underscores there.  Double-sanitizing is harmless / a no-op as there's no escaping sequence, just outright replacement.

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -20,7 +20,11 @@
 
 package metrics
 
-import "strconv"
+import (
+	"strconv"
+
+	"github.com/uber-go/tally"
+)
 
 const (
 	revisionTag     = "revision"
@@ -58,6 +62,16 @@ type (
 	}
 )
 
+// sanitizer is used to sanitize tag values, to prevent collisions and invalid data.
+// values should not allow dots for "namespacing" / "grouping" metrics, as they seem to refer to "one thing" in all cases.
+// `-` is not allowed in Prometheus keys or names, but it may be reasonable to allow it in values.
+var sanitizer = tally.NewSanitizer(tally.SanitizeOptions{
+	NameCharacters:       tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
+	KeyCharacters:        tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
+	ValueCharacters:      tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
+	ReplacementCharacter: '_',
+})
+
 func (s simpleMetric) Key() string   { return s.key }
 func (s simpleMetric) Value() string { return s.value }
 
@@ -65,7 +79,7 @@ func metricWithUnknown(key, value string) Tag {
 	if len(value) == 0 {
 		value = unknownValue
 	}
-	return simpleMetric{key: key, value: value}
+	return simpleMetric{key: key, value: sanitizer.Value(value)}
 }
 
 // DomainTag returns a new domain tag. For timers, this also ensures that we
@@ -82,7 +96,7 @@ func DomainUnknownTag() Tag {
 
 // InstanceTag returns a new instance tag
 func InstanceTag(value string) Tag {
-	return simpleMetric{key: instance, value: value}
+	return metricWithUnknown(instance, value)
 }
 
 // TargetClusterTag returns a new target cluster tag.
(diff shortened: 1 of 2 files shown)


common/metrics/tags.go
```diff
@@ -58,14 +62,23 @@ type (
 	}
 )
 
+// sanitizer is used to sanitize tag values, to prevent collisions and invalid data.
+// values should not allow dots for "namespacing" / "grouping" metrics, as they seem to refer to "one thing" in all cases.
+var sanitizer = tally.NewSanitizer(tally.SanitizeOptions{
+	NameCharacters:       tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreDashCharacters},
```
mantas-sidlauskas, line 68: Just curious.
AFAIK, Prometheus is not supporting dashes: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels (`It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.`)

If user selects Prometheus  as a reporter, will dashes be replaced to underscore?

Groxx: good question... hopefully?  that'd be a good reason to not add dashes though, if not

Groxx: let me see if I can find any other stat-sanitizing code that might be happening in this repo.  it may reveal more details.

common/metrics/tags.go
```diff
@@ -58,14 +62,23 @@ type (
...
+
 func (s simpleMetric) Key() string   { return s.key }
 func (s simpleMetric) Value() string { return s.value }
 
 func metricWithUnknown(key, value string) Tag {
 	if len(value) == 0 {
 		value = unknownValue
 	}
-	return simpleMetric{key: key, value: value}
+	return simpleMetric{key: key, value: sanitizer.Value(value)}
```
yycptt, line 81: I was looking into the same issue as some metrics with value constructed by our java client was dropped by the metrics system due to invalid characters. I also sent out an announcement in our open source slack channel to see if this change will impact open source customer or not.

Will doing sanitization for all metrics tag too expensive and cause performance issues? My original thought is that we can only apply sanitizer to the values we don't have control over, for example, tasklist, workflow type, activity type. For other types the value is set by the server code so we know their value won't contain invalid characters.

Groxx: sanitization cost: should be practically non-existent, it's doing a very simple transformation on a fairly short string.  other code in metrics alone likely costs *significantly* more, to say nothing of all the other code being executed.

common/metrics/tags.go
meiliang86: Can we leave the dash out and merge this PR?

Groxx: yeah, I'd be game to do that.  I haven't yet looked at other stats, but "the value is one conceptual thing" seems desirable from all calls to these at least.

Groxx: updated, dropped `-` (and added to the comment).  so this is now:
```go
// sanitizer is used to sanitize tag values, to prevent collisions and invalid data.
// values should not allow dots for "namespacing" / "grouping" metrics, as they seem to refer to "one thing" in all cases.
// `-` is not allowed in Prometheus keys or names, but it may be reasonable to allow it in values.
var sanitizer = tally.NewSanitizer(tally.SanitizeOptions{
	NameCharacters:       tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
	KeyCharacters:        tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
	ValueCharacters:      tally.ValidCharacters{tally.AlphanumericRange, tally.UnderscoreCharacters},
	ReplacementCharacter: '_',
})
```

meiliang86 (approved):

taylanisikdemir:

PR #3937, opened 2021-01-27
Drop pointer for attempt fields

vytautas-karpavicius:
**What changed?**
Drop pointer for attempt fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


common/util.go
```diff
@@ -313,7 +313,7 @@ func CreateMatchingPollForDecisionTaskResponse(historyResponse *types.RecordDeci
 	matchingResp := &types.MatchingPollForDecisionTaskResponse{
 		WorkflowExecution:         workflowExecution,
 		TaskToken:                 token,
-		Attempt:                   Int64Ptr(historyResponse.GetAttempt()),
+		Attempt:                   historyResponse.GetAttempt(),
```
Groxx, line 316: huh. what type is this actually?  can it just be `historyResponse.Attempt`?
but generally: `Get` -> field can wait if you want, we can tackle those later.

service/history/execution/history_builder.go
```diff
@@ -561,7 +561,7 @@ func (b *HistoryBuilder) newWorkflowExecutionStartedEvent(
 	attributes.PrevAutoResetPoints = resetPoints
 	attributes.Identity = request.Identity
 	attributes.RetryPolicy = request.RetryPolicy
-	attributes.Attempt = common.Int32Ptr(startRequest.GetAttempt())
+	attributes.Attempt = startRequest.GetAttempt()
```
Groxx, line 564: same "can it be `.Attempt`" Q

service/matching/matchingEngine.go
```diff
@@ -806,7 +806,7 @@ func (e *matchingEngineImpl) createPollForActivityTaskResponse(
 	}
 
 	response.TaskToken, _ = e.tokenSerializer.Serialize(token)
-	response.Attempt = common.Int32Ptr(int32(token.ScheduleAttempt))
+	response.Attempt = int32(token.ScheduleAttempt)
```
Groxx, line 809: I'm guessing it's an int64?
I don't believe this came up in the string/int cast linting I saw, but mind triple checking that this cast is sane? (I wish go casts weren't so yolo)

service/matching/thriftHandler_test.go
```diff
@@ -80,13 +80,13 @@ func TestThriftHandler(t *testing.T) {
...
 		resp, err := th.PollForActivityTask(ctx, &m.PollForActivityTaskRequest{})
-		assert.Equal(t, s.PollForActivityTaskResponse{WorkflowDomain: common.StringPtr("")}, *resp)
+		assert.Equal(t, s.PollForActivityTaskResponse{WorkflowDomain: common.StringPtr(""), Attempt: common.Int32Ptr(0)}, *resp)
 		assert.Equal(t, expectedErr, err)
 	})
 	t.Run("PollForDecisionTask", func(t *testing.T) {
 		h.EXPECT().PollForDecisionTask(ctx, &types.MatchingPollForDecisionTaskRequest{}).Return(&types.MatchingPollForDecisionTaskResponse{}, internalErr).Times(1)
 		resp, err := th.PollForDecisionTask(ctx, &m.PollForDecisionTaskRequest{})
-		assert.Equal(t, m.PollForDecisionTaskResponse{}, *resp)
+		assert.Equal(t, m.PollForDecisionTaskResponse{Attempt: common.Int64Ptr(0)}, *resp)
```
Groxx, line 89: ...huh.  one's int32 and one's int64?  that's sorta funky.
if it is, it is, but it's probably worth checking to make sure this makes sense.

common/types/mapper/thrift/history.go
```diff
@@ -2514,7 +2514,7 @@ func FromRecordActivityTaskStartedResponse(t *types.RecordActivityTaskStartedRes
 	return &history.RecordActivityTaskStartedResponse{
 		ScheduledEvent:                  FromHistoryEvent(t.ScheduledEvent),
 		StartedTimestamp:                t.StartedTimestamp,
-		Attempt:                         t.Attempt,
+		Attempt:                         &t.Attempt,
```
Groxx, line 2517: same "somewhat concerned about stored data" note, but a 0 here makes sense too even if it changes from not-stored to stored.  so probably not a problem either way.

Groxx (approved):

service/matching/matchingEngine.go
vytautas-karpavicius: Yeah, those are not consistent across codebase and there multiple conversions back and forth like this.
I have converted all attempt fields to `int32` in newly introduced proto IDLs. 
I think we should make them consistent internally as well. But this is another task.

common/util.go
vytautas-karpavicius: It can and it should be `historyResponse.Attempt`. All those helpers will become useless now, and should be simplified. However I'm planning to do it separately.

PR #3941, opened 2021-01-28
Drop pointer for poller id fields

vytautas-karpavicius:
**What changed?**
Drop pointer for poller id fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/common/types/mapper/thrift/matching.go
+++ b/common/types/mapper/thrift/matching.go
@@ -101,7 +101,7 @@ func FromCancelOutstandingPollRequest(t *types.CancelOutstandingPollRequest) *ma
 		DomainUUID:   &t.DomainUUID,
 		TaskListType: t.TaskListType,
 		TaskList:     FromTaskList(t.TaskList),
-		PollerID:     t.PollerID,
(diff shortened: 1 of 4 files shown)


common/types/mapper/thrift/matching.go
```diff
@@ -101,7 +101,7 @@ func FromCancelOutstandingPollRequest(t *types.CancelOutstandingPollRequest) *ma
 		DomainUUID:   &t.DomainUUID,
 		TaskListType: t.TaskListType,
 		TaskList:     FromTaskList(t.TaskList),
-		PollerID:     t.PollerID,
+		PollerID:     &t.PollerID,
```
Groxx, line 104: Same Q/assumption as the "forwarded from" field PR

Groxx (approved):

common/types/mapper/thrift/matching.go
vytautas-karpavicius: Yes, the `pollerID` is internal. It is set in the frontend with `uuid.New()`.

PR #3942, opened 2021-01-28
Drop pointer for forwarded from fields

vytautas-karpavicius:
**What changed?**
Drop pointer for forwarded from fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/common/types/mapper/thrift/matching.go
+++ b/common/types/mapper/thrift/matching.go
@@ -39,7 +39,7 @@ func FromAddActivityTaskRequest(t *types.AddActivityTaskRequest) *matching.AddAc
 		ScheduleId:                    t.ScheduleID,
 		ScheduleToStartTimeoutSeconds: t.ScheduleToStartTimeoutSeconds,
 		Source:                        FromTaskSource(t.Source),
-		ForwardedFrom:                 t.ForwardedFrom,
+		ForwardedFrom:                 &t.ForwardedFrom,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


common/types/mapper/thrift/matching.go
```diff
@@ -39,7 +39,7 @@ func FromAddActivityTaskRequest(t *types.AddActivityTaskRequest) *matching.AddAc
 		ScheduleId:                    t.ScheduleID,
 		ScheduleToStartTimeoutSeconds: t.ScheduleToStartTimeoutSeconds,
 		Source:                        FromTaskSource(t.Source),
-		ForwardedFrom:                 t.ForwardedFrom,
+		ForwardedFrom:                 &t.ForwardedFrom,
```
Groxx, line 42: I assume this is all internal, so no concerns about old things that won't send this?

Groxx (approved):

common/types/mapper/thrift/matching.go
vytautas-karpavicius: Yes this is internal only.

But important thing for this and related PRs is the actual usage of these fields. Most of them are used through `Get...` wrappers anyway. So the logic does not change as the code using them always receive empty string (even if it was nil originally). Therefore with these changes we only move nil to empty conversion to a different place.

Only in occasions where field is explicitly checked for nil value, we should pay attention, whether some optionality semantics were intended there.

PR #3945, opened 2021-01-29
Switch to revive, drop golint

Groxx:
See the individual commits for a detailed breakdown of lints / changes and what they mean / why they are made.
I intend to merge without squashing (as a fast-forward), so they are all retained.

As a high-level tl;dr, these commits:
- Drop golint, as it has somewhat significant annoyances
- Add [revive](https://github.com/mgechev/revive), as it's faster, more capable, and configurable
- Remove a pet peeve lint of mine
- Add many "free" lints we're already good on
- Add a couple significant lints, with fixes
- Add another significant lint, sadly with no fixes yet
- Last but not least, mention _every lint currently available_, and briefly documents why it's disabled / if we should consider enabling it.  The full list can be seen in the revive repo's readme, and there are additional docs as well: https://github.com/mgechev/revive


vytautas-karpavicius (approved):
This is very nice! It will allow us to improve code quality in the long run.

One question on the initial rule set. As I understand this is currently roughly equivalent to golint + few more rules that you see as high value, but not do not have too much violations currently, correct?

Would it make sense to add all available rules in the config file, but comment some them out with either:
- TODO: too much failures currently, once fixed enable it
- We will not use this rule because of <reason>
That way it would be more visible what we are using, what we are planning to use and what is not suitable for us and why.

Groxx:
I've kinda debated doing that, yeah.  The full list is rather large though, and some of them are just kinda clearly "meh" or "no" due to kinda innately having high false-positives or applying pressures that IMO lead to gaming metrics: https://github.com/mgechev/revive
Like `max-public-structs`: I'm pretty sure everyone agrees that too many things in one file is undesirable, but... what value is there in an arbitrary limit?  Sorta similar for the cyclomatic complexity check - simple is of course better, but artificially forcing a split of a conceptual unit can make it harder to understand.

Also you can eyeball cyclomatic complexity and some other lints easily enough during code review.  tbh I'm more interested in the things that are _hard_ to review - those are the ones that are particularly valuable to automate.

Anyway though.  Yep, I could do that - rejects really only take a few words, and it'd at least document the full list that we've seen.

Groxx:
ah, to answer the first part of your comment: yep, exactly.  I'd include the `unhandled-errors` immediately, but having too many warnings is almost as bad as having none - new ones get lost in the noise.  collectively I think this only adds ~20, and fixing the string-int-cast ones will cut that in half.

Groxx:
All lints now documented and organized!

meiliang86 (approved):

service/frontend/adminHandler.go
```diff
@@ -248,7 +248,7 @@ func (adh *adminHandlerImpl) DescribeWorkflowExecution(
 	}
 
 	shardID := common.WorkflowIDToHistoryShard(request.Execution.WorkflowID, adh.numberOfHistoryShards)
-	shardIDstr := string(shardID)
+	shardIDstr := strconv.Itoa(shardID)
```
mantas-sidlauskas, line 251: nit: we can drop one of shardID strings here

Groxx: I looked into this one a bit more (and failed to make a smaller repro), and have decided to undo it / change it to a form that `go vet` allows (and is recommended in the github issue thread that lead to the vet check being added) but produces the same output as before.

Since this appears to change how shards in a hash ring are selected, is it safe to change at all?  Or could it cause Problems™ on deploy, as different instances will compute different shards for the same keys?

service/frontend/adminHandler.go
```diff
@@ -248,7 +248,7 @@ func (adh *adminHandlerImpl) DescribeWorkflowExecution(
 	}
 
 	shardID := common.WorkflowIDToHistoryShard(request.Execution.WorkflowID, adh.numberOfHistoryShards)
-	shardIDstr := string(shardID)
+	shardIDstr := string(rune(shardID)) // originally `string(int_shard_id)`, but changing it will change the ring hashing
 	shardIDForOutput := strconv.Itoa(shardID)
```
Groxx, line 252: @mantas-sidlauskas ^ this is the new version

Makefile
```diff
@@ -270,13 +270,8 @@ go-generate: $(BIN)/mockgen $(BIN)/enumer
...
 
-lint: $(BIN)/golint fmt
-	@echo "running linter"
-	@lintFail=0; for file in $(sort $(LINT_SRC)); do \
-		$(BIN)/golint "$$file"; \
-		if [ $$? -eq 1 ]; then lintFail=1; fi; \
-	done; \
-	if [ $$lintFail -eq 1 ]; then exit 1; fi;
+lint: $(BIN)/revive
+	$(BIN)/revive -config revive.toml -exclude './canary/...' -exclude './vendor/...' -formatter unix ./... | sort
```
Groxx, line 274: strangely it also doesn't exclude vendor by default.  maybe if I set my go flags to include -mod=vendor...?

meh.  at least you can _tell it to ignore it_.

go.mod
```diff
@@ -18,8 +18,7 @@ require (
 	github.com/dmetzgar/goveralls v0.0.3
 	github.com/eapache/go-resiliency v1.2.0 // indirect
 	github.com/emirpasic/gods v0.0.0-20190624094223-e689965507ab
-	github.com/fatih/color v0.0.0-20181010231311-3f9d52f7176a
-	github.com/fatih/structtag v1.1.0 // indirect
+	github.com/fatih/color v1.10.0
```
Groxx, line 21: (to pick a random version change)
these changes are the result of a `go get` for revive, not a `go get -u`, so it likely has these versions as minimums.
either way though, none of these seem particularly concerning.

mantas-sidlauskas (approved):

client/history/client.go
```diff
@@ -1048,7 +1049,7 @@ func (c *clientImpl) getClientForDomainID(domainID string) (Client, error) {
 }
 
 func (c *clientImpl) getClientForShardID(shardID int) (Client, error) {
-	client, err := c.clients.GetClientForKey(string(shardID))
+	client, err := c.clients.GetClientForKey(strconv.Itoa(shardID))
```
Groxx, line 1052: this doesn't go into any load-distribution system, just an in-memory client cache.  safe to change.

PR #3946, opened 2021-01-30
Support AWS signing for ElasticSearch client

longquanzheng:
**What changed?**
Add optional support for AWS signing 
following https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service
https://github.com/olivere/elastic/blob/release-branch.v7/recipes/aws-connect-v4/main.go

**Why?**
From a Cadence user:
```
If turn on protections for ES
We use IAM to ensure only cadence has read/write access to our ES
in order for that to work properly you have to enable the signing of the requests
the client itself has the support baked in, but you have to turn it on
```

**How did you test it?**
User will test it. 

**Potential risks**
Very low/minimum risk as it's optional.

### DIFF
--- a/common/service/config/elasticsearch.go
+++ b/common/service/config/elasticsearch.go
@@ -42,6 +42,21 @@ type (
 		DisableSniff bool `yaml:"disableSniff"`
 		// optional to disable health check
 		DisableHealthCheck bool `yaml:"disableHealthCheck"`
+		// optional to use AWS signing client
+		// See more info https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service
+		AWSSigning AWSSigning `yaml:"awsSigning"`
+	}
+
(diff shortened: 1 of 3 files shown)


AngerM-DD:
This LGTM. I think it should work

longquanzheng:
> This LGTM. I think it should work

Thanks for reviewing the code.

common/service/config/elasticsearch.go
```diff
@@ -58,3 +90,13 @@ func (cfg *ElasticSearchConfig) SetUsernamePassword() {
 		cfg.URL.User = url.UserPassword(cfg.Username, cfg.Password)
 	}
 }
+
+func CheckAWSSigningConfig(config AWSSigning) error {
```
yycptt, line 94: nit: lint error for no comment on exported function?

yycptt (approved):

PR #3947, opened 2021-02-01
Lint preview (may be merged into #3945)

Groxx:
Integration tests don't run locally (still trying to figure that one out), so I'm using this to test changes.

I'm happy to switch to something else!  this was just the only thing I could find quickly, and I wanted to do a small check.


Groxx:
intended to be temporary, dropping it

PR #3948, opened 2021-02-01
Drop pointer for event id/version fields

vytautas-karpavicius:
**What changed?**
Drop pointers for event id/version fields within internal types. Some of the fields did have optional semantics. Those were not included in this PR. Everything else are never checked for nil and used via `Get<Field>()` helpers. 

**Why?**

**How did you test it?**

**Potential risks**


common/types/mapper/thrift/history.go
```diff
@@ -217,9 +217,9 @@ func ToGetMutableStateResponse(t *history.GetMutableStateResponse) *types.GetMut
 	return &types.GetMutableStateResponse{
 		Execution:                            ToWorkflowExecution(t.Execution),
 		WorkflowType:                         ToWorkflowType(t.WorkflowType),
-		NextEventID:                          t.NextEventId,
+		NextEventID:                          t.GetNextEventId(),
 		PreviousStartedEventID:               t.PreviousStartedEventId,
-		LastFirstEventID:                     t.LastFirstEventId,
+		LastFirstEventID:                     t.GetLastFirstEventId(),
```
Groxx, line 222: not specific to this one, and I know it's sorta late in these reviews to add this, but: tbh I'm kinda getting increasingly concerned about not detecting "things we think are always sent, but occasionally are not".

is it worth making a `notnull(thing)` helper to panic if that expectation turns out wrong?  presumably we have a per-request `defer func() { recover() }()` in some form (e.g. rpc layer) so it just becomes an error, so it'd be "safe" to do so.  it'd probably only be needed / most-valuable in these mapper files.

Groxx (approved):
LGTM, though check the comments (both pretty minor / we can definitely do validation later, I just think it may be good to have before we do a new deploy)

common/types/mapper/thrift/history.go
vytautas-karpavicius: We actually are not changing behavior when the field is not sent occasionally. Currently pretty much everywhere pointer fields are retrieved with `.GetThing()` helper. Thus it may be the case that those fields are not sent, but code consuming it is OK with zero value that `.GetThing()` helper returns. So I'm essentially moving where this helper is called. Instead of being scattered around the codebase it will immediately be called in RPC mapping layer. Having `notnull(thing)`  helper that panics would actually change this behavior.

Of course, there are several places that do not use `GetThing` helpers. They check for nil values and do something about that. I'm leaving such fields out of these PRs, and will address them separately.

common/types/history.go
```diff
@@ -187,9 +187,9 @@ func (v *GetMutableStateRequest) GetCurrentBranchToken() (o []byte) {
 type GetMutableStateResponse struct {
 	Execution                            *WorkflowExecution `json:"execution,omitempty"`
 	WorkflowType                         *WorkflowType      `json:"workflowType,omitempty"`
-	NextEventID                          *int64             `json:"NextEventId,omitempty"`
+	NextEventID                          int64              `json:"NextEventId,omitempty"`
 	PreviousStartedEventID               *int64             `json:"PreviousStartedEventId,omitempty"`
-	LastFirstEventID                     *int64             `json:"LastFirstEventId,omitempty"`
+	LastFirstEventID                     int64              `json:"LastFirstEventId,omitempty"`
```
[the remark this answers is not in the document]
vytautas-karpavicius: So, have been digging through code to understand this myself. As events are persisted in batches, this is the first event of the latest batch stored. This seems to be used when finding close event during long poll.

common/types/mapper/thrift/history.go
Groxx: so that'd be the rare exception like... domain was it?  which had a bigger branch on null than just the internal "get" helpers.

yea, that makes sense - they'd be rather visible during review I guess.  and no doubt there are a lot of these "nobody checks for nil" fields.

PR #3950, opened 2021-02-01
Cleanup internal types and remove generator

vytautas-karpavicius:
**What changed?**
- Removed internal types generator. It was used to generate internal types from Thrift types. It is no longer needed as additional changes to IDLs will be small can be done manually.
- Cleanup internal types. There were generated `...Args` and `...Result` types that are not used anywhere. Removed these types and their mappers.

**Why?**
To cleanup codebase.

**How did you test it?**
Passing build.

**Potential risks**
None. Removed types are not used.


Groxx (approved):
tbh I'm not sure what all the things are that were deleted... but if it compiles I don't see how they can be incorrect, so 👍

PR #3954, opened 2021-02-02
Flatten UpdateDomainRequest

vytautas-karpavicius:
**What changed?**
Flattened `UpdateDomainRequest` to explicitly list all updatable fields.

**Why?**
Protobuf API contains several changes for usability improvements. One such [change](https://github.com/uber/cadence/pull/3907/files#diff-80aa5b2a2fc654662fda4e36738c7ee08083c7b74db722605be8468805c5026eR75) flattens update request so that a subset of fields could be updated using a field mask. This is the only message that can contain nilable fields in our internal types as they can be set or not. By listing them separately in the request we can retain pointers on them to have such (set / not set) semantics. Fields for other messages will then drop pointers, as they are always required.

**How did you test it?**

**Potential risks**


common/domain/handler_GlobalDomainEnabled_MasterCluster_test.go
```diff
@@ -401,25 +401,19 @@ func (s *domainHandlerGlobalDomainEnabledMasterClusterSuite) TestUpdateGetDomain
...
+		Data:                                   data,
+		WorkflowExecutionRetentionPeriodInDays: common.Int32Ptr(retention),
+		EmitMetric:                             common.BoolPtr(emitMetric),
+		HistoryArchivalStatus:                  types.ArchivalStatusDisabled.Ptr(),
+		HistoryArchivalURI:                     common.StringPtr(""),
+		VisibilityArchivalStatus:               types.ArchivalStatusDisabled.Ptr(),
+		VisibilityArchivalURI:                  common.StringPtr(""),
+		BadBinaries:                            &types.BadBinaries{Binaries: map[string]*types.BadBinaryInfo{}},
+		ActiveClusterName:                      common.StringPtr(s.ClusterMetadata.GetCurrentClusterName()),
+		Clusters:                               clusters,
```
Groxx, line 416: seems identical except for `IsGlobalDomain: common.BoolPtr(isGlobalDomain),` at the top.

and more below...
since these are all overwhelmingly the same, maybe a cleanup (later?) would be to move all update tests to a separate file, since "update" is the thing that's happening regardless of cluster/global/etc?  the duplication would be pretty obvious then, may be easier to find a large chunk to dedup to a helper func or something.

common/types/mapper/thrift/shared.go
```diff
@@ -5474,60 +5474,80 @@ func ToTransientDecisionInfo(t *shared.TransientDecisionInfo) *types.TransientDe
...
-		Configuration:            FromDomainConfiguration(t.Configuration),
-		ReplicationConfiguration: FromDomainReplicationConfiguration(t.ReplicationConfiguration),
+	request := shared.UpdateDomainRequest{
+		Name:                     &t.Name,
 		SecurityToken:            t.SecurityToken,
 		DeleteBadBinary:          t.DeleteBadBinary,
 		FailoverTimeoutInSeconds: t.FailoverTimeoutInSeconds,
 	}
+	if t.Description != nil || t.OwnerEmail != nil || t.Data != nil {
+		request.UpdatedInfo = &shared.UpdateDomainInfo{
```
Groxx, line 5489: I'm not entirely sure if this is a good idea, but an alternate tactic could be:
```go
request := shared.UpdateDomainRequest{
  // ...
  UpdatedInfo: &shared.UpdateDomainInfo{
    // ... all values
  }
  // same for others, just do it all in one big struct init
}

// erase subsections that are zero-valued
if *request.UpdatedInfo == shared.UpdateDomainInfo{} {
  request.UpdatedInfo = nil
}
```
it might be a little more resistant to forgetting a `!= nil` check, and seems like it might match the semantics a bit better ("zero value means nothing changed means nil" vs "these fields are nil means nil")

common/domain/handler.go
```diff
@@ -765,10 +765,10 @@ func (d *handlerImpl) getHistoryArchivalState(
...
 
-	if requestedConfig != nil && clusterHistoryArchivalConfig.ClusterConfiguredForArchival() {
+	if clusterHistoryArchivalConfig.ClusterConfiguredForArchival() {
 		archivalEvent, err := d.toArchivalUpdateEvent(
-			requestedConfig.HistoryArchivalStatus,
-			requestedConfig.GetHistoryArchivalURI(),
+			updateRequest.HistoryArchivalStatus,
+			updateRequest.GetHistoryArchivalURI(),
 			clusterHistoryArchivalConfig.GetDomainDefaultURI(),
 		)
```
Groxx, line 773: took a while to be sure that this would lead to no-ops on nils / empty values... but yep.  looks good.

I feel like there _has_ to be a simpler way to write `archivalConfigStateMachine.go`, but I guess it is at least explicit about covering all possibilities...

Groxx (approved):

common/types/mapper/thrift/shared.go
vytautas-karpavicius: This would be nicer, I agree. Unfortunately we can not compare these struct like that since they contain slices or maps.

Groxx: ah, right, un-comparable things.  fun.
well, there is `reflect.IsZero()`

PR #3955, opened 2021-02-03
Improve docker-compose local unit-test, minor cleanup

Groxx:
Unfortunately the "unit-test" isn't much of a unit test, clearly connecting to the database containers...
... but after these changes, it at least starts up on my machine, and doesn't collide with other instances / port use / etc.
It also finally connects to postgres, though that should still be fixed eventually.

To summarize the changes:
- `ports` -> `expose` to keep networking internal.  ports are no longer publicly mapped, so you can run other things without colliding.
- `command` array -> a string.  because wow was that verbose.
- adding postgres connection info to unit-test, as it was using the wrong host + username previously.

---

Unfortunately tests still don't pass locally, I have not yet traced down the reason.
They stop here for me, on a few errors like this:
```
--- FAIL: TestSQLHistoryV2PersistenceSuite (3.70s)
    --- PASS: TestSQLHistoryV2PersistenceSuite/TestConcurrentlyCreateAndAppendBranches (1.23s)
    --- FAIL: TestSQLHistoryV2PersistenceSuite/TestConcurrentlyForkAndAppendBranches (0.61s)
        historyV2PersistenceTest.go:752:
            	Error Trace:	historyV2PersistenceTest.go:752
				historyV2PersistenceTest.go:570
				asm_amd64.s:1357
            	Error:      	Expected nil, but got: &types.InternalDataInconsistencyError{Message:"corrupted history event batch, eventID is not continouous"}
            	Test:       	TestSQLHistoryV2PersistenceSuite/TestConcurrentlyForkAndAppendBranches
```

Still, it's progress.

### DIFF
--- a/docker/buildkite/docker-compose-local.yml
+++ b/docker/buildkite/docker-compose-local.yml
@@ -3,8 +3,8 @@ version: "3.5"
 services:
   cassandra:
     image: cassandra:3.11
-    ports:
-      - "9042:9042"
+    expose:
+      - "9042"
     networks:
       services-network:
         aliases:
@@ -17,8 +17,8 @@ services:
       MYSQL_ROOT_PASSWORD: uber
     volumes:
       - ./mysql-init:/docker-entrypoint-initdb.d
-    ports:
-      - "3306:3306"
+    expose:
+      - "3306"
     networks:
       services-network:
         aliases:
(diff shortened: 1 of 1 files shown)


just-at-uber:
Looks like you have some integration tests failing with the new changes.

Groxx:
should be totally unrelated - buildkite doesn't use the -local files. (restarting tests)

just-at-uber (approved):

PR #3956, opened 2021-02-03
Drop pointers for bool fields

vytautas-karpavicius:
**What changed?**
Drop pointers for bool fields within internal types.

**Why?**

**How did you test it?**

**Potential risks**


Groxx (approved):
Yea, all these seem like safe ones to change.  `*bool` in general really, it's kinda rare that bools are used for ternary logic.

PR #3961, opened 2021-02-05
Attempt to make some racy tests more reliable

Groxx:
Unfortunately these tests do not have any guaranteed way to sequence these things,
so I wrote down the best-effort strategies and used them to hopefully make the code a bit more clear on intent.

At the core, what this code needs is a way to simultaneously block on one action (push into queue) and unblock another (read from queue), and then do the reverse later / in other tests.
You can do that by e.g. pushing a message on a non-buffered channel...
... but that would require significant complicating changes to this code just for testing.
And you can't guarantee you can push to *two* channels at before unblocking, so the push/message must be *the actual job to perform synchronously*, moving in lock-step between the two sides, which is wildly different than how it works now.

So it's inherently racy.  Throw stuff at the wall until it sticks, but try to be clear about the goals.

### DIFF
--- a/service/matching/matcher_test.go
+++ b/service/matching/matcher_test.go
@@ -86,20 +86,13 @@ func (t *MatcherTestSuite) TestLocalSyncMatch() {
 	<-t.fwdr.AddReqTokenC()
 	<-t.fwdr.PollReqTokenC()
 
-	pollStarted := make(chan struct{})
-
-	go func() {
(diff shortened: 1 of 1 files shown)


Groxx:
With this change, all my (docker-compose) unit-test tests finally pass locally!  So that's good to see.

mantas-sidlauskas (approved):

PR #3965, opened 2021-02-08
Provide background context to all scanners

### LINKED ISSUE #3963: TaskListScavengerActivity panic due to scanner context is missing
**Describe the bug**
```
lastFailureReason
cadenceInternal:Panic
lastWorkerIdentity
1@prod-cadence-worker-c7c77b94b-sv8gz@cadence-sys-tl-scanner-tasklist-0
lastFailureDetails.0
interface conversion: interface {} is nil, not scanner.scannerContext
lastFailureDetails.1
activity for cadence-sys-tl-scanner-tasklist-0 [panic]: github.com/uber/cadence/service/worker/scanner.TaskListScavengerActivity(0x2bf1100, 0xc000098ea0, 0x0, 0x0) /cadence/service/worker/scanner/workflow.go:155 +0x6b3 reflect.Value.call(0x211f5a0, 0x2694fa8, 0x13, 0x25ae4c7, 0x4, 0xc0010f60e0, 0x1, 0x1, 0x1, 0x0, ...) /usr/local/go/src/reflect/value.go:460 +0x5f6 reflect.Value.Call(0x211f5a0, 0x2694fa8, 0x13, 0xc0010f60e0, 0x1, 0x1, 0x0, 0x0, 0x0) /usr/local/go/src/reflect/value.go:321 +0xb4 go.uber.org/cadence/internal.(*activityExecutor).Execute(0xc0003f6a20, 0x2bf1100, 0xc000098ea0, 0x0, 0x0, 0x0, 0x3eb9a80, 0xc000dcabd0, 0x24, 0xc0009600c0, ...) /go/pkg/mod/go.uber.org/cadence@v0.14.1/internal/internal_worker.go:710 +0x3b3 go.uber.org/cadence/internal.(*activityTaskHandlerImpl).Execute(0xc0000e32c0, 0x25eb5ed, 0x21, 0xc000620370, 0x0, 0x0, 0x0, 0x0) /go/pkg/mod/go.uber.org/cadence@v0.14.1/internal/internal_task_handlers.go:1822 +0x969 go.uber.org/cadence/internal.(*activityTaskPoller).ProcessTask(0xc00035c930, 0x2050ec0, 0xc000a35200, 0xc0006845b0, 0x0) /go/pkg/mod/go.uber.org/cadence@v0.14.1/internal/internal_task_pollers.go:888 +0x1b8 go.uber.org/cadence/internal.(*baseWorker).processTask(0xc0006c6380, 0x2050ec0, 0xc000a35200) /go/pkg/mod/go.uber.org/cadence@v0.14.1/internal/internal_worker_base.go:323 +0x127 created by go.uber.org/cadence/internal.(*baseWorker).runTaskDispatcher /go/pkg/mod/go.uber.org/cadence@v0.14.1/internal/internal_worker_base.go:252 +0x105
```

It's broken in https://github.com/uber/cadence/pull/3772 
the context KV is not being set anymore after this refactoring
**To Reproduce**
Is the issue reproducible? 
- Yes

**Steps to reproduce the behavior:**
Start Cadence worker with MySQL should reproduce it. 

**Expected behavior**

**Screenshots**

**Additional context**
This is the second time we broke system workflow like this...
Last time was fixed by this PR: https://github.com/uber/cadence/pull/3584

I would highly recommend the next person fixing it also add some integration test to protect this being broken again.

mantas-sidlauskas:
**What changed?**
Context keys are now typed and private. Every scanner has context key named after workflow name. Activity context is used to get workflow name. Shardscanner now don't pass workflow name in activity param
Fixes #3963

**Why?**
This will reduce possibility to misuse context keys

**How did you test it?**
Unit tests, local test

**Potential risks**

### DIFF
--- a/service/worker/scanner/scanner.go
+++ b/service/worker/scanner/scanner.go
@@ -117,7 +117,7 @@ func New(
 
 // Start starts the scanner
 func (s *Scanner) Start() error {
-	backgroundActivityContext := context.Background()
+	backgroundActivityContext := newBackgroundContext(s.context)
 	var workerTaskListNames []string
 	var wtl []string
 
@@ -161,23 +161,33 @@ func (s *Scanner) startShardScanner(
 	ctx context.Context,
 	config *shardscanner.ScannerConfig,
 ) (context.Context, []string) {
-	scannerContextKey := shardscanner.ScannerContextKey(config.ScannerWFTypeName)
-	fixerContextKey := shardscanner.ScannerContextKey(config.FixerWFTypeName)
 	workerTaskListNames := []string{}
-	backgroundActivityContext := context.WithValue(ctx, scannerContextKey, s.context)
-
 	if config.DynamicParams.ScannerEnabled() {
-		backgroundActivityContext = context.WithValue(
-			backgroundActivityContext,
-			scannerContextKey,
+		ctx = context.WithValue(
+			ctx,
+			shardscanner.ScannerContextKey(config.ScannerWFTypeName),
 			getShardScannerContext(s.context, config),
 		)
+
+		go s.startWorkflowWithRetry(
+			config.StartWorkflowOptions,
+			config.ScannerWFTypeName,
+			shardscanner.ScannerWorkflowParams{
+				Shards: shardscanner.Shards{
+					Range: &shardscanner.ShardRange{
+						Min: 0,
+						Max: s.context.cfg.Persistence.NumHistoryShards,
+					},
+				},
+			})
+
+		workerTaskListNames = append(workerTaskListNames, config.StartWorkflowOptions.TaskList)
 	}
 
 	if config.DynamicParams.FixerEnabled() {
-		backgroundActivityContext = context.WithValue(
-			backgroundActivityContext,
-			fixerContextKey,
+		ctx = context.WithValue(
+			ctx,
+			shardscanner.ScannerContextKey(config.FixerWFTypeName),
 			getShardFixerContext(s.context, config),
 		)
 
@@ -192,23 +202,7 @@ func (s *Scanner) startShardScanner(
 		workerTaskListNames = append(workerTaskListNames, config.StartFixerOptions.TaskList)
 	}
 
-	if config.DynamicParams.ScannerEnabled() {
-		workerTaskListNames = append(workerTaskListNames, config.StartWorkflowOptions.TaskList)
-
-		go s.startWorkflowWithRetry(
-			config.StartWorkflowOptions,
-			config.ScannerWFTypeName,
-			shardscanner.ScannerWorkflowParams{
-				Shards: shardscanner.Shards{
-					Range: &shardscanner.ShardRange{
-						Min: 0,
-						Max: s.context.cfg.Persistence.NumHistoryShards,
-					},
-				},
-			})
-	}
-
-	return backgroundActivityContext, workerTaskListNames
+	return ctx, workerTaskListNames
 }
 
 func (s *Scanner) startWorkflowWithRetry(
@@ -259,6 +253,12 @@ func (s *Scanner) startWorkflow(
 	}
 	s.context.GetLogger().Info("workflow successfully started", tag.WorkflowType(workflowType))
 	return nil
+}
+
+func newBackgroundContext(scannerContext scannerContext) context.Context {
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, scannerContextKey, scannerContext)
+	return ctx
 }
 
 func getShardScannerContext(ctx scannerContext, config *shardscanner.ScannerConfig) shardscanner.Context {
--- a/service/worker/scanner/scanner.go
+++ b/service/worker/scanner/scanner.go
@@ -25,13 +25,13 @@ import (
 	"fmt"
 	"time"
 
-	"go.uber.org/cadence/activity"
-
 	"github.com/uber-go/tally"
-	"go.uber.org/cadence/.gen/go/shared"
-	cclient "go.uber.org/cadence/client"
-	"go.uber.org/cadence/worker"
 	"go.uber.org/zap"
+
+	"go.uber.org/cadence/.gen/go/shared"
+	"go.uber.org/cadence/activity"
+	"go.uber.org/cadence/client"
+	"go.uber.org/cadence/worker"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/backoff"
@@ -77,10 +77,10 @@ type (
 		TallyScope tally.Scope
 	}
 
-	// ScannerContext is the context object that get's
+	// scannerContext is the context object that get's
 	// passed around within the scanner workflows / activities
-	ScannerContext struct {
-		resource.Resource
+	scannerContext struct {
+		resource   resource.Resource
 		cfg        Config
 		tallyScope tally.Scope
 		zapLogger  *zap.Logger
@@ -90,7 +90,7 @@ type (
 	// of database tables to cleanup resources, monitor anomalies
 	// and emit stats for analytics
 	Scanner struct {
-		context ScannerContext
+		context scannerContext
 	}
 )
 
@@ -109,8 +109,8 @@ func New(
 		resource.GetLogger().Fatal("failed to initialize zap logger", tag.Error(err))
 	}
 	return &Scanner{
-		context: ScannerContext{
-			Resource:   resource,
+		context: scannerContext{
+			resource:   resource,
 			cfg:        params.Config,
 			tallyScope: params.TallyScope,
 			zapLogger:  zapLogger.Named("scanner"),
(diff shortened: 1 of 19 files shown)


service/worker/scanner/scanner.go
```diff
@@ -161,23 +161,33 @@ func (s *Scanner) startShardScanner(
...
 	workerTaskListNames := []string{}
-	backgroundActivityContext := context.WithValue(ctx, scannerContextKey, s.context)
-
 	if config.DynamicParams.ScannerEnabled() {
-		backgroundActivityContext = context.WithValue(
-			backgroundActivityContext,
-			scannerContextKey,
+		ctx = context.WithValue(
+			ctx,
+			shardscanner.ScannerContextKey(config.ScannerWFTypeName),
```
longquanzheng, line 168: I am a bit worried about the way of managing contextKeys. Two potential issues here --
1. They could conflict each other because the next person may not be able to find all the existing keys already used
2. The bug will come back again on another refactoring. Before we can implement any test(it's quite tricky and I don't even have a good idea to test yet), may be just write a code in a more organized way:

So How about having a separate file/module to define all the context keys in a `map[string]interface{}`, and then return an error when starting scanner if any key is not set in the backgroundContext?

Groxx: yea, personally I dislike exposing keys _ever_.  e.g. rather than `context.WithValue(ctx, getKeyFunc(..), value)`, do `setValue(ctx, key_args, value)` + a typed get-value func.  makes it effectively impossible to misuse, trivial to find all calls, you can add validation -> return err, etc.

longquanzheng: yeah dislike those global keys too. That's the main reasons that we ran into the bug twice, so we should improve this piece of code. 
 
Not 100% following what you are suggesting but seems interesting. Are you saying having a method `setValue` of Scanner to manage that?

Groxx: basically:
- all keys should be private + uniquely typed (+ personally: strings, so they're readable in logs/debuggers).  that way the value does not matter.
- in that same file: make as-strongly-typed-as-possible setter and getter funcs, which read/write/cast the value with the keys
- literally nothing except those funcs ever uses the keys, ideally guaranteed by the compiler because they're private
- since everything's trivial and in one file, you can verify it at a glance and move on with your life

that way it's all typed, you can add validation in insertion and/or retrieval, and there's zero chance of collisions or wrong type casts.

mantas-sidlauskas:
Will make this a bit bigger refactoring

service/worker/scanner/scanner.go
mantas-sidlauskas: Just to add more context here, we are dealing here with two separate context types:

- shardscanner context which is using typed string for context key. While adding new shardscanner, you don't even have to think about context as it uses workflow name as context key,  which also helps to avoid collisions as you are not allowed to have different workflows under the same name.

- TaskListScanner/HistoryScanner context which shares the same context key provided by a constant which usage was mistakenly moved to another scope]

service/worker/scanner/scanner.go
```diff
@@ -117,36 +120,40 @@ func New(
 
 // Start starts the scanner
 func (s *Scanner) Start() error {
-	backgroundActivityContext := context.Background()
+	bac := context.Background()
```
longquanzheng, line 123: Nit: lol, I haven't seen this name as a convention in the code base. I think ctx should be fine, the rest of the code may not care if it's from background or not.

service/worker/scanner/shardscanner/types.go
```diff
@@ -419,6 +414,69 @@ func (s Shards) Flatten() ([]int, int, int) {
...
+	val, ok := ctx.Value(contextKey(info.WorkflowType.Name)).(*Context)
+	if !ok {
+		return nil, fmt.Errorf("context type is not %T for a key %q", val, info.WorkflowType.Name)
+	}
+	return val, nil
+}
+
+// GetFixerContext extracts fixer context from activity context
+// it uses typed, private key to reduce access scope
+func GetFixerContext(
```
longquanzheng, line 470: GetScannerContext and GetFixerContext are almost the same except for casting. Maybe just have a pair of symmetric methods:
```
SetActivityContext
```
and
```
GetActivityContext
```

service/worker/scanner/shardscanner/types.go
```diff
@@ -419,6 +414,69 @@ func (s Shards) Flatten() ([]int, int, int) {
...
+	ctx context.Context,
+	workflowName string,
+	scannerContext interface{},
+) context.Context {
+	return context.WithValue(ctx, contextKey(workflowName), scannerContext)
+}
+
+// GetScannerContext extracts scanner context from activity context
+// it uses typed, private key to reduce access scope
+func GetScannerContext(
```
longquanzheng, line 454: Also I think this context management should be outside of shardscanner package.

service/worker/scanner/shardscanner/types.go
```diff
@@ -419,6 +414,69 @@ func (s Shards) Flatten() ([]int, int, int) {
 	return shardList, min, max
 }
 
-func (c ScannerContextKey) String() string {
-	return string(c)
+// NewShardScannerContext sets scanner context up
+func NewShardScannerContext(
+	res resource.Resource,
+	config *ScannerConfig,
+) Context {
```
longquanzheng, line 421: I am a bit sick of this name `Context`. Could we make a more explicit? maybe ShardScannerContext. It will help people with less knowledge or you become less familar after 1 year ago, read it easier I think

Groxx (approved):
To summarize:
- Changes look correct and/or with minor enough requests that I'm game to ✅ to unblock.  Seems like it'll work.
- I sorta agree that the `scanner.Context` + `shardscanner.Context` + `context.Context` + `workflow.Context` is a bit confusing.  But tbh I'm not sure what name would work better; "context" describes it pretty well, and it's just an unfortunate collision that Go tried to claim "Context" for one specific kind of context.  Also I don't have much trouble with following when reading, when I get my head out of "`context.Context` only" mode.
  - So I think I'm fine with the names.  If you do come up with better ones, great.  If not: 👍
- I have a few broader thoughts on this chunk of code as a whole (below).  But there's no need to block this PR on that for another day or more, unless you want to do so.

---

Broader Qs / notes / maybe for a refactor later:
1. What _exactly_ are the various `Context`s for / what's going into them?
    1. It appears that every instance (per type) is identical and never mutated, but it's added to the `context.Context` quite a few times, and each activity retrieves its unique copy.  Especially visible in the loop at the beginning of `scanner.Scanner.Start()`.
        - If that's accurate: this could probably be inserted once (per type), no need for workflow-name-specific versions at all.
        - If not: what values are unique?  I may have just missed something obvious / hidden in collapsed code.
    2. It looks like this can all be computed at process-startup (after reading config) regardless of the workflows / activities that are going to be used.
        - If that's accurate: especially when combined with the previous, it seems like this could become just a single small set of additions to `context.Context` that's always done, no conditional logic whatsoever.  Probably worth doing so it's trivial to follow?
    3. Actually, is it just inserting the same-keyed items N times?  So there are N copies in BAC, but only the latest one can be retrieved?
2. ... How does this all work, anyway?  It seems like the scanner starts up, begins some workflows, sets some BACs, and then _the same process_ works on activities?
    - If that's accurate: why not local activities?  Why a workflow at all, really?  Or is this all needed for durability?  (I suppose they're kinda slow too, right?)
    - If it's for durability: it seems like that'd also imply 1ii^ so these values are not at all based on "how the workflows are configured to start", just "what activities are running" (or even "all activities possible"), as otherwise resuming from a server shutdown wouldn't really be possible.  Seems safer and simpler to make the activity BACs reflect that.
3. We would really benefit from a nicer dependency-injection thing for workflows and activities :|  And so would our users.  I'm not quite sure what that should look like though.
    - Registering activity-instances here might achieve it, and get rid of the need for BAC / type-erasure entirely.  That may be nicer, and it's available now.
4. Potentially this all demonstrates a conceptual issue with local activities / a gap in our semantics?
    - Currently they're "run at-least-once in same process, very low latency, but do not complete decision tasks until [some time later]".
        - Useful, but a bit risky / surprising with long-running things... unless you know far too many internal details and know how and when to yield so it's recorded (so IMO: they're just risky/surprising).
        - Maybe we should make it clear(er) that they're primarily intended for "trivial" operations?
        - Or just do away with them and say "use SideEffect" as it's practically the same thing (I think?).
    - Maybe there needs to be another category?  "a normal activity, but run by this same process".  A "local session" perhaps.  "Same memory space as made the decision to run the activity" is a separate desire, durability / other details are orthogonal.
        - It'd probably also be a closer fit for this kind of "run a durable set of actions, but only on this one machine" purpose.  Assuming I'm understanding what this scanner does, anyway.
        - It could still be faster than a normal activity, as the decision-task-commit response is sufficient to begin processing immediately, which may save a round-trip.  (or do we have an optimization like that already?)

mantas-sidlauskas:
Addressing only part of comments just to keep scope narrow. 

Thanks @longquanzheng and @Groxx for your suggestions and questions

PR #3966, opened 2021-02-09
Drop pointer for signal name fields

vytautas-karpavicius:
Drop pointer for signal name fields within internal types.


Groxx (approved):
simple 👍

PR #3532, opened 2020-09-21
Add e2e integration tests for Postgres and remove 2dc runs

longquanzheng:
**What changed?**
Adding postgres to run with integration test suits

**Why?**
Follow up to a conversation in https://github.com/uber/cadence/pull/3488#discussion_r483220280 

**How did you test it?**
It's adding tests 

**Potential risks**
No

Local run succeeded:
```
docker-compose -f docker/buildkite/docker-compose-local.yml run integration-test-postgres
...
...
"service": "cadence-matching", "logging-call-at": "service.go:115"}
--- PASS: TestSizeLimitIntegrationSuite (14.23s)
    --- PASS: TestSizeLimitIntegrationSuite/TestTerminateWorkflowCausedBySizeLimit (2.44s)
PASS
coverage: 29.0% of statements in github.com/uber/cadence/common/..., github.com/uber/cadence/service/..., github.com/uber/cadence/client/..., github.com/uber/cadence/tools/...
ok  	github.com/uber/cadence/host	546.843s	coverage: 29.0% of statements in github.com/uber/cadence/common/..., github.com/uber/cadence/service/..., github.com/uber/cadence/client/..., github.com/uber/cadence/tools/...
1539.65user 518.08system 10:40.22elapsed 321%CPU (0avgtext+0avgdata 1528492maxresident)
136720inputs+1913456outputs (50major+10853577minor)pagefaults 0swaps
```

PR #3967, opened 2021-02-09
Fix xdc failover integration test

vytautas-karpavicius:
I have just noticed that after #3954 xdc integration test was not updated. My IDE started showing errors for this.

I fixed what was missing, but the bigger question is why is this not included in the build?

### DIFF
--- a/host/xdc/integration_failover_test.go
+++ b/host/xdc/integration_failover_test.go
@@ -158,10 +158,8 @@ func (s *integrationClustersTestSuite) TestDomainFailover() {
 
 	// update domain to fail over
 	updateReq := &types.UpdateDomainRequest{
-		Name: common.StringPtr(domainName),
-		ReplicationConfiguration: &types.DomainReplicationConfiguration{
-			ActiveClusterName: common.StringPtr(clusterName[1]),
-		},
+		Name: domainName,
+		ActiveClusterName: common.StringPtr(clusterName[1]),
 	}
 	updateResp, err := client1.UpdateDomain(createContext(), updateReq)
 	s.NoError(err)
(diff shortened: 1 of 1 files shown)


yycptt:
My understanding is that xdc tests will be run on buildkite. 

But looks like they are removed by https://github.com/uber/cadence/pull/3532.

Maybe @yux0 @longquanzheng can provide more context on that change?

yycptt (approved):

vytautas-karpavicius:
> My understanding is that xdc tests will be run on buildkite.
> 
> But looks like they are removed by #3532.
> 
> Maybe @yux0 @longquanzheng can provide more context on that change?

Maybe just remove those tests if we are not running anymore? I see this was also considered in the linked #3532

yycptt:
> > My understanding is that xdc tests will be run on buildkite.
> > But looks like they are removed by #3532.
> > Maybe @yux0 @longquanzheng can provide more context on that change?
> 
> Maybe just remove those tests if we are not running anymore? I see this was also considered in the linked #3532

Yeah, I also think we can remove the failover test in xdc. But it looks like there's another elasticsearch test in xdc that's not covered by the ndc integration test?

vytautas-karpavicius:
> > > My understanding is that xdc tests will be run on buildkite.
> > > But looks like they are removed by #3532.
> > > Maybe @yux0 @longquanzheng can provide more context on that change?
> > 
> > 
> > Maybe just remove those tests if we are not running anymore? I see this was also considered in the linked #3532
> 
> Yeah, I also think we can remove the failover test in xdc. But it looks like there's another elasticsearch test in xdc that's not covered by the ndc integration test?

Created an issue to address that #3979

PR #3969, opened 2021-02-09
Drop pointer for task id fields

vytautas-karpavicius:
Drop pointer for task id fields within internal types.


service/matching/matchingEngine_test.go
```diff
@@ -1123,13 +1123,13 @@ func (s *matchingEngineSuite) TestMultipleEnginesActivitiesRangeStealing() {
 	// History service is using mock
 	s.mockHistoryClient.EXPECT().RecordActivityTaskStarted(gomock.Any(), gomock.Any()).DoAndReturn(
 		func(ctx context.Context, taskRequest *types.RecordActivityTaskStartedRequest) (*types.RecordActivityTaskStartedResponse, error) {
-			if _, ok := startedTasks[*taskRequest.TaskID]; ok {
+			if _, ok := startedTasks[taskRequest.TaskID]; ok {
```
Groxx, line 1126: I wonder if we can lint on "do not pass pointers to maps" :|

Groxx (approved):

service/matching/matchingEngine_test.go
vytautas-karpavicius: Well, `revive` does not seem to have such rule.
On the other hand, we are not passing pointer here (nor did we before the change).

PR #3972, opened 2021-02-09
Drop pointer on remaining domain fields

vytautas-karpavicius:
Dropped remaining pointer on domain related fields within internal types.

Apart from the usual changes, this PR also includes 2 behavior changes:
- When registering a domain, `IsGlobalDomain` flag will default to false if not specified. Currently it is explicitly checked for nil and an error returned. This can be debated though. I image this explicit error could be useful if we want to avoid silently creating local domains when user forgets the flag. 
- When registering a domain, `ActiveClusterName` will be checked against empty string instead of nil. CLI updated accordingly. This should not be a problem as both can be used to represent missing value which then would default to `clusterMetadata.GetCurrentClusterName()`. As empty cluster name is not required this should be ok.


host/xdc/integration_failover_test.go
```diff
@@ -158,10 +158,8 @@ func (s *integrationClustersTestSuite) TestDomainFailover() {
 
 	// update domain to fail over
 	updateReq := &types.UpdateDomainRequest{
-		Name: common.StringPtr(domainName),
-		ReplicationConfiguration: &types.DomainReplicationConfiguration{
-			ActiveClusterName: common.StringPtr(clusterName[1]),
-		},
+		Name:              domainName,
+		ActiveClusterName: common.StringPtr(clusterName[1]),
```
Groxx, line 162: Since there's just one, this makes sense IMO, but I am kinda curious why it changed.

Groxx (approved):
Yeah, I think defaulting to false is fine.  I could see making it a nullable field so callers must specify it.... but supporting global at all is optional (e.g. local ones won't have it), so I don't really think it's *necessary*.

Can people delete their (local) domains and create a new global one with the same name, e.g. if there have been no workflows yet?  Or is the name forever claimed?

vytautas-karpavicius:
> Yeah, I think defaulting to false is fine. I could see making it a nullable field so callers must specify it.... but supporting global at all is optional (e.g. local ones won't have it), so I don't really think it's _necessary_.
> 
> Can people delete their (local) domains and create a new global one with the same name, e.g. if there have been no workflows yet? Or is the name forever claimed?

It is possible to delete it directly from DB if there is no workflows yet, but there is no cli command to support that.

host/xdc/integration_failover_test.go
vytautas-karpavicius: This is was actually fixed with #3967 and no longer there after rebase.

PR #3973, opened 2021-02-09
Decouple decision handler and history engine impl logic

yycptt:
**What changed?**
- Decouple decision handler and history engine impl logic. 
- Create separate files for loading and updating function.
- Next step will be moving all decision related logic to a separate package.

**Why?**
https://github.com/uber/cadence/issues/3181

**How did you test it?**
Tested locally

**Potential risks**
n/a, only refactor


service/history/workflow_util.go
```diff
@@ -0,0 +1,318 @@
...
+		CreateDecision: true,
+	}
+	UpdateWorkflowWithoutDecision = &UpdateWorkflowAction{
+		CreateDecision: false,
+	}
+)
+
+///////////////////  Util function for loading workflow ///////////////////
+
+func LoadWorkflowOnce(
```
yux0, line 56: are those functions cover in the handler tests?

yux0 (approved):

service/history/workflow_util.go
yycptt: Good point. I think right now those util functions are covered by `history_engine_test.go` as when calling history engine APIs, those functions will also be called. But there's no specific unit test for those util functions. 

When moving those functions to a separate `workflow` package, I think it make sense to add those specific tests. I will do that in the diff for separating packages.

PR #3976, opened 2021-02-10
Use a test-logger in tests rather than stdout

Groxx:
Currently, on failure, e.g. `matchingEngine_test.go` tests spew out *ridiculous* amounts of text.
This happens because Go cannot tell which logs are related to the failed test(s), so it does what it can: show all stdout.

Unfortunately, stdout blends logs from parallel tests (I don't believe it's even line-synchronized), and it's hard or impossible to find the ones relevant to the failure simply by reading.
So: stop doing that.  Zap has a `zaptest` package for this purpose, we just have to use it.
For tests using the workflow suite, there's an annoying call-order dependency that means we have to `suite.SetLogger` before `suite.NewTestWorkflowEnvironment` or it's ignored... but that's not actually hard to deal with.  I just wish we (and our users) didn't have to.

I'm not sure how many other places may need to be changed, but this helps the ones I'm looking at at the moment.

### DIFF
--- a/canary/workflow_test.go
+++ b/canary/workflow_test.go
@@ -31,6 +31,7 @@ import (
 	"go.uber.org/cadence/mocks"
 	"go.uber.org/cadence/testsuite"
 	"go.uber.org/cadence/worker"
+	"go.uber.org/zap/zaptest"
 )
 
 type (
@@ -51,7 +52,9 @@ func TestWorkflowTestSuite(t *testing.T) {
 }
 
 func (s *workflowTestSuite) SetupTest() {
+	s.SetLogger(zaptest.NewLogger(s.T()))
 	s.env = s.NewTestWorkflowEnvironment()
+	s.env.Test(s.T())
 }
 
 func (s *workflowTestSuite) TearDownTest() {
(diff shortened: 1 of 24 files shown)


canary/workflow_test.go
```diff
@@ -51,7 +52,9 @@ func TestWorkflowTestSuite(t *testing.T) {
 }
 
 func (s *workflowTestSuite) SetupTest() {
+	s.SetLogger(zaptest.NewLogger(s.T()))
```
meiliang86, line 55: Any reason we are not changing the default [here](https://github.com/uber-go/cadence-client/blob/master/internal/internal_workflow_testsuite.go#L272)?

Groxx: I think that's the right choice, yeah.  Among possibly other locations - any calls to `zap.New...` in pretty much any go code anywhere are immediately suspect IMO.

Groxx: hm.  unfortunately adding it there is somewhere between "not trivial" and "requires breaking changes".  there doesn't appear to be any `*testing.T` available at that point, and if I start changing code to require it, I eventually end up at public APIs.

still very much worthwhile, but tbh I think it's worth tackling separately.

meiliang86 (approved):

PR #3977, opened 2021-02-10
Fix/0.19.x build

just-at-uber:
Running `go mod tidy` in attempt to fix broken build.


Groxx:
this one can probably be dropped, right?

PR #3978, opened 2021-02-10
Repair 0.19.x branch

Groxx:
Fixing a bad merge of dependencies.  Effectively:
- restore imports in cmd/server/tools.go
- run `go mod tidy`

the golang.org/x/tools upgrade was causing issues internally.
that's now fixed, so the updated versions of libs (including revive, which depends on golang.org/x/tools 0.1.0) can be used.


just-at-uber (approved):

go.mod
```diff
@@ -65,13 +67,13 @@ require (
 	go.uber.org/zap v1.12.0
 	golang.org/x/net v0.0.0-20201031054903-ff519b6c9102
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
-	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9 // indirect
-	golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4 // indirect
 	golang.org/x/text v0.3.4 // indirect
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	golang.org/x/tools v0.1.0
```
mkolodezny, line 72: Unfortunately the tools needs to be reverted to golang.org/x/tools v0.0.0-20200127195909-ed30b9180dd3 as we did for 0.18 otherwise the diff runs into issues upon land to the monorepo :(

Groxx: yea.  I think I've got that figured out and an upgrade is in the works - unfortunately it's mutually exclusive with `revive` :|

we could (... should?) move to a separate `tools.mod` file, and that would probably resolve this for a while at least.  monorepo's gonna have to update eventually though, static analysis code is going to start depending on newer x/tools at some point.

yux0 (approved):

PR #3983, opened 2021-02-13
Fix IP address in docker template

longquanzheng:
**What changed?**
Fix IP address in docker template. It's hardcoded to 127.0.0.1, this won't work with docker container :D 

**Why?**
See discussion in https://uber-cadence.slack.com/archives/CL22WDF70/p1613176534165800?thread_ts=1612951969.125100&cid=CL22WDF70

**How did you test it?**
Locally running batcher workflow

**Potential risks**
No

### DIFF
--- a/docker/config_template.yaml
+++ b/docker/config_template.yaml
@@ -203,7 +203,7 @@ clusterMetadata:
             enabled: true
             initialFailoverVersion: 0
             rpcName: "cadence-frontend"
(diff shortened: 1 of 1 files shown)


Groxx (approved):
Given the rest of the stuff in this file: looks right / reasonable to me 👍

longquanzheng:
@Groxx  @meiliang86  thanks for reviewing and merging it.

PR #3988, opened 2021-02-17
Add oneof for DescribeHistoryHostRequest in proto IDLs

vytautas-karpavicius:
**What changed?**
Added `oneof` wrapper in `DescribeHistoryHostRequest` message within proto IDLs for **admin** service.

Also removed all fields from `DescribeHistoryHostRequest` message for **history** service. Fields in admin service are used to [select](https://github.com/uber/cadence/blob/master/client/history/client.go#L151) appropriate history client. In history service itself [none of those fields are used](https://github.com/uber/cadence/blob/master/service/history/handler.go#L728). This is also conceptually correct as concrete history host is already resolved and none arguments are required to describe it.

**Why?**
To better reflect a contract that history host can be describe by one of these options. Similar construct is already used for [DescribeDomainRequest](https://github.com/uber/cadence/blob/master/proto/public/uber/cadence/api/v1/service_domain.proto#L109)

**How did you test it?**

**Potential risks**


Groxx (approved):
Makes sense 👍

PR #3991, opened 2021-02-17
Revert changes to use string casting for shard ID

yux0:
**What changed?**
Change the shard ID integer casting to use string(rune())

**Why?**
Make backward compatible with ringpop mapping.

**How did you test it?**

**Potential risks**

### DIFF
--- a/client/history/client.go
+++ b/client/history/client.go
@@ -22,7 +22,6 @@ package history
 
 import (
 	"context"
-	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


Groxx (approved):
Yep, looks right, all those lookups are hash-related.
That said: is it possible to change any of these funcs to take an int instead of a string?

yux0:
> Yep, looks right, all those lookups are hash-related.
> That said: is it possible to change any of these funcs to take an int instead of a string?

We can abstract it and pass in int. But we already use this int -> char. So even in the func, we have to do int -> char -> string.

Groxx:
Yeah, somewhere it has to happen.  I'm just hoping to take that need out of calling locations, as that kind of abnormal semantic requirement not really visible during code review.

yycptt (approved):

yycptt:
> > Yep, looks right, all those lookups are hash-related.
> > That said: is it possible to change any of these funcs to take an int instead of a string?
> 
> We can abstract it and pass in int. But we already use this int -> char. So even in the func, we have to do int -> char -> string.

Yeah having a function may be a good idea. Although we still do int -> rune -> string conversion in the function, at least there's a central place for the conversion and situation where history client using a different logic than the shard controller won't happen.

PR #3993, opened 2021-02-17
Rename metric name SignalInfo to SignalInfoSize

AngerM:
Describe what has changed in this PR:
Fix the `SignalInfoSize` metric name from `signal_info` to `signal_info_size`

Tell your future self why have you made these changes:
`signal_info` is a timer which means it implicitly also creates a `signal_info_count` metric in prometheus/Otel style metrics, this conflicts with the `SignalInfoCount` metric defined on line 2345

How have you verified this change? Tested locally? Added a unit test? Checked in staging env?:
Deployed and tested in a production environment

Assuming the worst case, what can be broken when deploying this change to production?:
Metric names is changed and Cadence users should update dashboard if they use this metric.


AngerM:
cc @longquanzheng

longquanzheng (approved):

yux0 (approved):

PR #4003, opened 2021-02-19
Drop pointer for DomainTaskAttributes fields

vytautas-karpavicius:
Drop pointers for DomainTaskAttributes fields within internal types.


yycptt (approved):
LGTM. Not sure if @yux0 wants to take another look.

PR #4007, opened 2021-02-20
Add more default Prometheus histogram buckets for non-latency timer metrics like history_size

longquanzheng:
**What changed?**
Add more default Prometheus histogram buckets for non-latency timer metrics like history_size

**Why?**
The default buckets from tally won't work for Cadence's non-latency metrics
See more: https://github.com/uber/cadence/issues/4006

**How did you test it?**
Local test

**Potential risks**
No.

### DIFF
--- a/common/metrics/tally/prometheus/buckets.go
+++ b/common/metrics/tally/prometheus/buckets.go
@@ -1,0 +1,72 @@
+// Copyright (c) 2021 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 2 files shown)


Groxx:
> history_count_bucket{..., le="5.0000000000000004e-08"} 24

bleh.  is there a non-terrible way to improve that output?

longquanzheng:
> > history_count_bucket{..., le="5.0000000000000004e-08"} 24
> 
> bleh. is there a non-terrible way to improve that output?
> it might just be due to division not producing an ideal value, i.e. there's a value that's closer to `5.0e-08`.

My best guess is same. It doesn't seem to be an issue when I used it in Prometheus/Grafana. 

There is a way to improve. The current output is not only terrible/silly, but also in-efficient. NanoSecond granularity is useless for latency metrics in Cadence. The minimum latency to Cassandra could be a few microseconds. Because of those additional buckets, all the latency metrics will be recorded and output in NanoSecond as well. 

To improve this, we can change all the non-latency timer metrics to emit values with 1000 times of the current. That way we can remove the 4 buckets  under microSecond:
```
 		{Upper: 5 * OneOrNanoSecond},
		{Upper: 10 * OneOrNanoSecond},
		{Upper: 50 * OneOrNanoSecond},
		{Upper: 100 * OneOrNanoSecond},
```
However, this is a backward-incompatible change. All dashboards of non-latency timer metrics will need to divided by 1000 to get the correct values as before. 

Although it's a breaking change, this should only impact Uber's Cadence clusters(because non-latency timer metrics are already broken for the community if they don't change the Prometheus timer type to Summary). 

So if I would suggest Uber Cadence to make this change. M3QL supports this easily but just need to change all the dashboards and monitor for non-latency timer metrics. 

cc @meiliang86 LMK what you think :D

common/metrics/tally/prometheus/buckets.go
```diff
@@ -0,0 +1,72 @@
...
+// creating a new Histogram in the prometheus registry.
+// Cadence uses Histogram to report both latency and non-latency metrics(like history_count/history_size).
+// The buckets need to set correct boundary for both.
+// See more in https://github.com/uber/cadence/issues/4006
+func DefaultHistogramBuckets() []prometheus.HistogramObjective {
+
+	// For latency metrics, this is a nanoSecond.
+	// For non-latency metrics, this is ONE.
+	// divided by Second because of https://github.com/uber-go/tally/blob/04828d51c63b1d09b46824d7c0d7904b5eb1b3b6/prometheus/reporter.go#L183
+	OneOrNanoSecond := float64(time.Nanosecond) / float64(time.Second)
```
meiliang86, line 39: nit: any reason you need to capitalize this var?

meiliang86 (approved):

common/metrics/tally/prometheus/buckets.go
longquanzheng: No reason at all. That's a mistake. I didn't notice that LoL. Will change it. Thanks!

PR #4008, opened 2021-02-22
Drop pointer on remaining fields

vytautas-karpavicius:
Dropped pointers on remaining fields that are relevant for proto migration.

After this PR only primitives that still have pointers are: enums, timestamps, durations and failure reason fields. Enums map to special `_INVALID` value in proto, so leaving them for now. Timestamps and durations fields are nullable in proto (as they are wrapped in special types), thus are left as well. Later those could be converted appropriate `time.Time` and `time.Duration`. But for the time being they do not block further work on proto migration. Also failure reason fields still contain pointers. However in proto they (together with failure details) are wrapped into separate `Failure` message. Therefore dropping pointer here is not blocking currently, and may be refactored later.

I realise this is rather big, but otherwise there would be a series of many very small PRs, since all remaining fields are rather random.

Special attention is required to `ContinuedExecutionRunID` and `ExpectedNextEventID` fields. I was back and forth between dropping their pointers and adding _optionality_ into proto messages. Both variants are possible. However since they both contain usable zero values, I have chosen to drop pointers here as well. However, this is debatable and I'm open for other option as well.


common/types/shared.go
```diff
@@ -8938,14 +8938,14 @@ func (v *UpsertWorkflowSearchAttributesEventAttributes) GetSearchAttributes() (o
 
 // VersionHistories is an internal type (TBD...)
 type VersionHistories struct {
-	CurrentVersionHistoryIndex *int32            `json:"currentVersionHistoryIndex,omitempty"`
+	CurrentVersionHistoryIndex int32             `json:"currentVersionHistoryIndex,omitempty"`
```
Groxx, line 8941: 0 and nil seem fairly different in this case.  or is there always a 'current' index?

service/frontend/adminHandler.go
```diff
@@ -284,7 +284,7 @@ func (adh *adminHandlerImpl) RemoveTask(
 	scope, sw := adh.startRequestProfile(metrics.AdminRemoveTaskScope)
 	defer sw.Stop()
 
-	if request == nil || request.Type == nil || request.TaskID == nil {
+	if request == nil {
```
Groxx, line 287: hmms.  is 0-type a valid type?  (actually, what type is it?  thrift files just leave it as an int....)

service/history/historyEngine.go
```diff
@@ -974,7 +974,7 @@ func (e *historyEngineImpl) getMutableStateOrPolling(
 
 	// expectedNextEventID is 0 when caller want to get the current next event ID without blocking
 	expectedNextEventID := common.FirstEventID
-	if request.ExpectedNextEventID != nil {
+	if request.ExpectedNextEventID != 0 {
```
Groxx, line 977: I suppose a 0-id would be fairly meaningless, as you can't wait for the first (i.e. "start workflow") event?  so all practical ones are 1+... if so, this seems fine (and matches the comment)

Groxx (approved):
A few comments, but if they're fine, ship it.

common/types/shared.go
vytautas-karpavicius: This index starts from zero and is always set, check here: https://github.com/uber/cadence/blob/5ffd942d386e80e9e26820149914353e181303a0/common/persistence/versionHistory.go#L383

service/frontend/adminHandler.go
vytautas-karpavicius: Yeah, lack of enums makes it harder to understand. You can track it down to this: https://github.com/uber/cadence/blob/master/common/constants.go#L164
We can see here that valid values starts from `2`. Maybe update this check to reject everything under `2`?

Also for proto, I have introduced enum for those: https://github.com/uber/cadence/blob/master/proto/internal/uber/cadence/shared/v1/queue.proto#L27

Groxx: yea, yay improved RPC APIs :)

I'd prefer a check of some kind, yeah.  `>=2` is strictly accurate and might lead to discovering the right type too, but `!=0` might be enough - it at least identifies "did not provide -> default 0" as an invalid value, which is probably good enough until we switch to better types.

vytautas-karpavicius: Actually, I have reverted pointer drop on those fields. As it is mapped to enum value in proto, it also contains `_INVALID` value for it. So we can map nil to `_INVALID` just like for all other enums, making more consistent.

vytautas-karpavicius: Regarding all those checks. I have a rough idea how it should be done. Since we now have pure internal types, it opens quite a few doors for improvement. Once such idea - internal type validation functions. Those types could have:
```
func (t SomeInternalType) Validate() error {
  if t.Field1 <= 0 {
    return Error1
  }
  if t.Field2 == nil {
    return Error2
  }
  etc...
}
```

Then bigger types such as all `Request`s could call `Validate()` of all composed types, like:
```
func (t *SomeFrontendRequest) Validate() error {
  var errors error
  if t == nil {
    errors = multierr.Append(errors, NilRequestError)
  }
  if err := t.Field1.Validate(); err != nil {
    errors = multierr.Append(errors, err)
  }
  if err := t.Field2.Validate(); err != nil {
    errors = multierr.Append(errors, err)
  }
  return errors
}
```

All validation logic would be in once place and easily testable. Then in each handler, we could simple call this `Validate` at the very beginning of a call:
```
func (h *Handler) SomeRPC(request *SomeRequest) (*SomeResponse, error) {
  if err := request.Validate(); err != nil {
    return nil, err
  }
}
```
With this setup we would not need to worry about invalid values possible somewhere down the line. It would be consistent, testable and isolated in once place.

Of course this is for another (not small) task, but that is the direction that I would like to eventually proceed.

PR #4014, opened 2021-02-25
feature/changelog

just-at-uber:
Adding a CHANGELOG.md file that follows [keep a changelog](https://keepachangelog.com/en/1.0.0/) format which documents what is in upcoming or past releases. It helps when preparing for a release as the notes are written as the work is completed. These notes can then be copied into the github release page when publishing a new release. The only change would be to move all notes accumulated in `## [Unreleased]` section to a versioned section.

I have not updated previous releases as we need to understand what is noteworthy and what is not (or what multiple commits could be summarized into one point).

When completing work in the future, please update this file as it will help the team with the next release and communication to our end users.

## Screenshots
Screen Shot 2021-02-24 at 4 06 13 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


Groxx (approved):
SGTM - I think it's fine for old releases to be documented incrementally / later.

Maybe add a reference (at the top?) to look at github releases for older ones, in the meantime?

just-at-uber:
Added a note to reference github releases page. Thanks!

PR #4015, opened 2021-02-25
Move workflow utility components to separate package

yycptt:
**What changed?**
- Move workflow utility components to separate package
- Added unit test for util functions to update workflow execution and get domain entry
- Renamed some util functions

**Why?**
Decomposing history package

**How did you test it?**
Existing tests.

**Potential risks**
Risk is minimal as it only moves code around.


service/history/workflow/util.go
```diff
@@ -272,12 +288,11 @@ func GetActiveDomainEntry(
 	domainUUID string,
```
yux0, line 288: Can this move to shard context?

vytautas-karpavicius: +1
This would makes more sense as it does not have anything workflow related.

yycptt: Yeah, I was thinking probably this should be moved to `domain` package instead of `shard` as it has not much to do with shard, which is only for getting the domain cache.

yycptt: Synced with @yux0 offline. So `GetActivityDomainEntry` now is a method of the `DomainCache` interface and `ValidateDomainUUID` is moved to `common` package.

yux0 (approved):

PR #4017, opened 2021-02-25
Add missing fields/messages to proto IDLs

vytautas-karpavicius:
Over the time new fields were introduced to Thrift IDLs, while proto IDLs (while still in development) got behind.
This PR syncs missing fields and messages, so that proto IDL would be up to date.


mantas-sidlauskas (approved):
sgtm

PR #4020, opened 2021-02-25
fixed the poller count metric

mkolodezny:
**What changed?**
clean the expired pollers before reporting the metric

**Why?**
follow up on https://github.com/uber/cadence/pull/3918

**How did you test it?**
tested locally


service/matching/pollerHistory.go
```diff
@@ -68,6 +74,7 @@ func (pollers *pollerHistory) updatePollerInfo(id pollerIdentity, ratePerSecond
 		rps = *ratePerSecond
 	}
 	pollers.history.Put(id, &pollerInfo{ratePerSecond: rps})
+	pollers.onHistoryUpdatedFunc()
```
mantas-sidlauskas, line 77: nit: do we need to check for a nil here?

service/matching/taskListManager.go
```diff
@@ -174,8 +174,9 @@ func newTaskListManager(
 	} else {
 		taskListTypeMetricScope = tlMgr.metricScope().Tagged(taskListDecisionTypeTag)
 	}
-	tlMgr.pollerHistory = newPollerHistory(func(state *cache.UpdatedState) {
-		taskListTypeMetricScope.UpdateGauge(metrics.PollerPerTaskListCounter, float64(state.NewSize))
+	tlMgr.pollerHistory = newPollerHistory(func() {
+		taskListTypeMetricScope.UpdateGauge(metrics.PollerPerTaskListCounter,
+			float64(len(tlMgr.pollerHistory.getAllPollerInfo())))
```
mantas-sidlauskas, line 179: sadly we can't (?) use history.Size() here

mantas-sidlauskas (approved):
smgt

service/matching/taskListManager.go
mkolodezny: yes, as discussed offline, for current cache implementation we need to evict the expired pollers everytime now.

PR #4022, opened 2021-02-28
Add support for ScyllaDB

adambabik:
**What changed?**

This PR adds support for ScyllaDB. The codebase is 99.9% compatible, I needed to change one line.

**Why?**

ScyllaDB is an alternative implementation of Cassandra. It was requested in https://github.com/uber/cadence/issues/2607 and a related bug was reported in https://github.com/uber/cadence/issues/3314.

**How did you test it?**

I tested it locally using docker-compose and running a few tests from https://github.com/uber-common/cadence-samples. using Cassandra and ScyllaDB as persistent stores.

**Potential risks**

I don't see any huge risks. Cassandra is still a default database and there was almost no code changes.

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/queue.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/queue.go
@@ -188,7 +188,7 @@ func (db *cdb) InsertQueueMetadata(
 ) error {
 	clusterAckLevels := map[string]int64{}
 	query := db.session.Query(templateInsertQueueMetadataQuery, queueType, clusterAckLevels, version).WithContext(ctx)
-	_, err := query.ScanCAS()
+	_, err := query.ScanCAS(nil, nil, nil)
 	if err != nil {
 		return err
 	}
@@ -209,7 +209,7 @@ func (db *cdb) UpdateQueueMetadataCas(
 		row.QueueType,
 		row.Version-1,
 	).WithContext(ctx)
-	applied, err := query.ScanCAS()
+	applied, err := query.ScanCAS(nil, nil, nil, nil)
 	if err != nil {
 		return err
 	}
--- a/common/persistence/nosql/nosqlplugin/cassandra/queue.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/queue.go
@@ -188,6 +188,10 @@ func (db *cdb) InsertQueueMetadata(
 ) error {
 	clusterAckLevels := map[string]int64{}
 	query := db.session.Query(templateInsertQueueMetadataQuery, queueType, clusterAckLevels, version).WithContext(ctx)
+	
+	// NOTE: Must pass nils to be compatible with ScyllaDB's LWT behavior
+	// "Scylla always returns the old version of the row, regardless of whether the condition is true or not."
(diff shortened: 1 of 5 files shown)


common/persistence/nosql/nosqlplugin/cassandra/queue.go
```diff
@@ -188,7 +188,7 @@ func (db *cdb) InsertQueueMetadata(
 ) error {
 	clusterAckLevels := map[string]int64{}
 	query := db.session.Query(templateInsertQueueMetadataQuery, queueType, clusterAckLevels, version).WithContext(ctx)
-	_, err := query.ScanCAS()
+	_, err := query.ScanCAS(nil, nil, nil)
```
adambabik, line 191: I haven't researched this yet but it looks like Scylla returns all values for LWT while Cassandra returns only one. This code works for both, though.

adambabik: This might be answered by in [ScyllaDB docs](https://docs.scylladb.com/kb/lwt-differences/):
> Conditional statements return a result set, and unlike Cassandra, Scylla result set metadata doesn’t change from execution to execution: Scylla always returns the old version of the row, regardless of whether the condition is true or not. This ensures conditional statements work well with prepared statements.

longquanzheng (approved):

common/persistence/nosql/nosqlplugin/cassandra/queue.go
longquanzheng: Thanks for contributing. 
This looks good to me, however, have you checked is there any other places need to change like this?

docker/config_template.yaml
```diff
@@ -13,6 +13,19 @@ persistence:
...
+                keyspace: {{ default .Env.KEYSPACE "cadence" }}
+                user: {{ default .Env.CASSANDRA_USER "" }}
+                password: {{ default .Env.CASSANDRA_PASSWORD "" }}
+        visibility:
+            cassandra:
+                hosts: {{ default .Env.CASSANDRA_SEEDS "" }}
+                keyspace: {{ default .Env.VISIBILITY_KEYSPACE "cadence_visibility" }}
+                user: {{ default .Env.CASSANDRA_USER "" }}
+                password: {{ default .Env.CASSANDRA_PASSWORD "" }}
+        {{- else if eq $db "scylla" }}
```
longquanzheng, line 28: Nit: This looks redundant. 
Can you help us  figure out how to improve this by using OR operator? I think it's using this templating language: https://github.com/jwilder/dockerize

adambabik: I simplified this.

common/persistence/nosql/nosqlplugin/cassandra/queue.go
adambabik: I checked the source code and found one more place, however, it seems that that code path is not executed when running samples from https://github.com/uber-common/cadence-samples. It would panic otherwise. All other calls use `MapScanCAS()` which should be fine.

longquanzheng: Thanks. That's also the only place that I can see. 
The code path will be executed in XDC mode which is not used in sample by default.

PR #4030, opened 2021-03-03
Implement domain deprecate CLI command

Shaddoll:
**What changed?**
Implement domain deprecate CLI command

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/tools/cli/domainCommands.go
+++ b/tools/cli/domainCommands.go
@@ -31,7 +31,6 @@ import (
 
 	"github.com/olekukonko/tablewriter"
 	"github.com/urfave/cli"
-	s "go.uber.org/cadence/.gen/go/shared"
 
 	"github.com/uber/cadence/client/frontend"
 	"github.com/uber/cadence/common"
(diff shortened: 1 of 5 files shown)


tools/cli/domainCommands.go
```diff
@@ -282,6 +282,28 @@ func (d *domainCLIImpl) UpdateDomain(c *cli.Context) {
...
+
+	ctx, cancel := newContext(c)
+	defer cancel()
+
+	err := d.deprecateDomain(ctx, &types.DeprecateDomainRequest{
+		Name: domainName,
+		SecurityToken: securityToken,
+	})
+	if err != nil {
+		if _, ok := err.(*s.EntityNotExistsError); !ok {
```
vytautas-karpavicius, line 297: As call uses types from `types` package, you should also check agains an error (with same name) from this package.
This was probably copy-paste, in that case could you check if there are any more cases like this?

yycptt (approved):

PR #4032, opened 2021-03-03
Add compose file for multi-cluster setup

Shaddoll:
**What changed?**
Add compose file for multi-cluster setup

**Why?**
It makes it easier for us to run cadence with 2 clusters on laptop.

**How did you test it?**

**Potential risks**

### DIFF
--- a/docker/config_template.yaml
+++ b/docker/config_template.yaml
@@ -117,7 +117,7 @@ services:
         metrics:
             statsd:
                 hostPort: {{ .Env.STATSD_ENDPOINT }}
-                prefix: "cadence-frontend"
+                prefix: {{ default .Env.STATSD_FRONTEND_PREFIX "cadence-frontend" }}
         {{- else if .Env.PROMETHEUS_ENDPOINT }}
         metrics:
             prometheus:
@@ -138,7 +138,7 @@ services:
         metrics:
             statsd:
                 hostPort: {{ .Env.STATSD_ENDPOINT }}
-                prefix: "cadence-matching"
+                prefix: {{ default .Env.STATSD_MATCHING_PREFIX "cadence-matching" }}
         {{- else if .Env.PROMETHEUS_ENDPOINT }}
         metrics:
             prometheus:
@@ -159,7 +159,7 @@ services:
         metrics:
             statsd:
                 hostPort: {{ .Env.STATSD_ENDPOINT }}
-                prefix: "cadence-history"
+                prefix: {{ default .Env.STATSD_HISTORY_PREFIX "cadence-history" }}
         {{- else if .Env.PROMETHEUS_ENDPOINT }}
         metrics:
             prometheus:
@@ -180,7 +180,7 @@ services:
         metrics:
             statsd:
                 hostPort: {{ .Env.STATSD_ENDPOINT }}
-                prefix: "cadence-worker"
+                prefix: {{ default .Env.STATSD_WORKER_PREFIX "cadence-worker" }}
         {{- else if .Env.PROMETHEUS_ENDPOINT }}
         metrics:
             prometheus:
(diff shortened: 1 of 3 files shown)


docker/docker-compose-multiclusters.yml
```diff
@@ -0,0 +1,62 @@
...
+    image: ubercadence/server:master-auto-setup
+    ports:
+     - "7933:7933"
+     - "7934:7934"
+     - "7935:7935"
+     - "7939:7939"
+    environment:
+      - "CASSANDRA_SEEDS=cassandra"
+      - "STATSD_ENDPOINT=statsd:8125"
+      - "DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development.yaml"
```
yux0, line 24: The replication has to be config from the yaml file. Take a look at the development_active.yaml and development_standby.yaml

docker/config_template.yaml
```diff
@@ -197,13 +197,25 @@ clusterMetadata:
     enableGlobalDomain: {{ default .Env.ENABLE_GLOBAL_DOMAIN "false" }}
     failoverVersionIncrement: 10
     masterClusterName: "master"
+    {{- if .Env.IS_NOT_MASTER }}
```
yux0, line 200: For some reason, let's not use the word master. What about is_not_active? And if you can replace the master with other word in this file. It would be great.

docker/docker-compose-multiclusters.yml
Shaddoll: Yes, it uses config_template.yaml to generate config file. This environment variable is set to get dynamic config.

yux0: Oh, I see. This is a dynamic config yaml. Then it looks good to me.

yux0 (approved):

docker/config_template.yaml
```diff
@@ -196,14 +196,26 @@ services:
...
-    currentClusterName: "master"
+    masterClusterName: "active"
+    {{- if .Env.IS_NOT_ACTIVE }}
+    currentClusterName: "standby"
+    {{- else }}
+    currentClusterName: "active"
+    {{- end }}
     clusterInformation:
-        master:
+        active:
```
longquanzheng, line 206: Can we not use active/standby? I purposely changed this because this is confusing in the XDC concepts. Active/Standby is a concept for domain, not cluster. If master is not the right word, how about `clusterDCA`, `clusterDCB`? @yux0

longquanzheng (changes requested):
Please do not use active/standby to name clusters. This will confuse users.

docker/config_template.yaml
Shaddoll: how about `primary` and `secondary`

longquanzheng: yeah that works too. I like the idea :D

docker/docker-compose-multiclusters.yml
```diff
@@ -0,0 +1,72 @@
...
+    depends_on:
+      - cassandra
+      - statsd
+  cadence-standby:
+    image: ubercadence/server:master-auto-setup
+    ports:
+     - "7943:7933"
+     - "7944:7934"
+     - "7945:7935"
+     - "7949:7939"
```
yycptt, line 41: nit: Can we make the port number consistent with those in `development_standby.yaml`? Makes it a bit easier when switching back and forth between different setup.
```
  bootstrapHosts: [ "127.0.0.1:8933", "127.0.0.1:8934", "127.0.0.1:8935", "127.0.0.1:8940" ]
```

yycptt (approved):

longquanzheng (approved):
Nit: could you also make change to config/ that:
```
development_active.yaml -> development_primary.yaml
development_standby.yaml -> development_secondary.yaml
```
And change the related README.md
Or create a ticket and do it separately :D

PR #4036, opened 2021-03-04
Fix warnings when compiling the source code

git-hulk:
**What changed?**

Fix some warnings when `make bins`

**Why?**

The suggestion was reasonable, so we should fix it as we can

**How did you test it?**

* `make bins` to check whether those warnings were disappeared or not 
* make test

**Potential risks**

no

### DIFF
--- a/common/domain/replication_queue.go
+++ b/common/domain/replication_queue.go
@@ -229,16 +229,11 @@ func (q *replicationQueueImpl) UpdateDLQAckLevel(
 	ctx context.Context,
 	lastProcessedMessageID int64,
 ) error {
-
-	if err := q.queue.UpdateDLQAckLevel(
+	return q.queue.UpdateDLQAckLevel(
 		ctx,
 		lastProcessedMessageID,
 		localDomainReplicationCluster,
-	); err != nil {
-		return err
-	}
-
-	return nil
+	)
 }
 
 func (q *replicationQueueImpl) GetDLQAckLevel(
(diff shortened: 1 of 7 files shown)


service/worker/scanner/scanner.go
```diff
@@ -49,7 +49,6 @@ const (
 )
 
 type (
-	contextKey string
```
Groxx, line 52: these are intentionally a different type - e.g. if we ever had a workflow named `processorContext`, they could collide.

common/constants.go
```diff
@@ -176,3 +176,8 @@ const MemoKeyForOperator = "operator"
 
 // ReservedTaskListPrefix is the required naming prefix for any task list partition other than partition 0
 const ReservedTaskListPrefix = "/__cadence_sys/"
+
+type (
+	// ContextKey is the alias type for the context key
+	ContextKey string
```
Groxx, line 182: not-sharing a type is the main point of that "don't use a string" lint.  it's better to make all the types unique.

e.g. don't use common.ContextKey elsewhere, make a unique `type contextKey string` like what `scanner.go` does.  then collisions are literally impossible, and you don't have to think about it or look for other uses of the type anywhere in the repo.

Groxx (changes requested):
context-key-types should change, but the err-returns look good.  and thank you!

service/worker/scanner/scanner.go
git-hulk: okay, thanks for your review, I have removed the Common.Context

Groxx:
Well that's not a very clear error :\
```
INFO  Event name: pull_request
INFO  Updating PR #4036 Fix warnings when compiling the source code
INFO  fix-warning-when-building HEAD: 4b7d7f27df2e7c49324ad2a14fccc2805241ef0f
INFO  Rebasing onto master f2f10d4dbb4cad55b3444ea9dff1bae81008e4bb
ERROR command failed with code 128: git push --quiet --force-with-lease origin fix-warning-when-building
```
mind merging master / rebasing?  I wonder if something is conflicting.

git-hulk:
@Groxx I have merged and rebased the master, it seems no conflict was found. You can help to squash the PR if need, thanks.

Groxx (approved):
Thanks for the contribution, and all the changes!

PR #4038, opened 2021-03-04
Disable starting new workflows in deprecated domain

Shaddoll:
**What changed?**
Disable starting new workflows in deprecated domain

**Why?**

**How did you test it?**
unit test

**Potential risks**

### DIFF
--- a/service/history/historyEngine3_eventsv2_test.go
+++ b/service/history/historyEngine3_eventsv2_test.go
@@ -263,6 +263,34 @@ func (s *engine3Suite) TestStartWorkflowExecution_BrandNew() {
 	s.NotNil(resp.RunID)
 }
 
+func (s *engine3Suite) TestStartWorkflowExecution_DeprecatedDomain() {
+	domainID := constants.TestDomainID
+	workflowID := "workflowID"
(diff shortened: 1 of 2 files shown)


service/history/historyEngine3_eventsv2_test.go
```diff
@@ -263,6 +263,34 @@ func (s *engine3Suite) TestStartWorkflowExecution_BrandNew() {
...
+			WorkflowID:                          workflowID,
+			WorkflowType:                        &types.WorkflowType{Name: workflowType},
+			TaskList:                            &types.TaskList{Name: taskList},
+			ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(1),
+			TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(2),
+			Identity:                            identity,
+			RequestID:                           requestID,
+		},
+	})
+	s.Error(err)
```
yycptt, line 292: nit: probably check if the error has type `BadRequestError`, so that it's will be treated as a non-retryable error by the caller. Same for signal with start.

yycptt (approved):

PR #4041, opened 2021-03-08
Disable signal workflow in deprecated domains

Shaddoll:
**What changed?**
Disable signal workflow in deprecated domains

**Why?**

**How did you test it?**
new unit tests

**Potential risks**

### DIFF
--- a/service/history/historyEngine.go
+++ b/service/history/historyEngine.go
@@ -2110,6 +2110,9 @@ func (e *historyEngineImpl) SignalWorkflowExecution(
 	if err != nil {
 		return err
 	}
+	if domainEntry.GetInfo().Status != persistence.DomainStatusRegistered {
+		return errDomainDeprecated
+	}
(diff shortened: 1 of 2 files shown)


service/history/historyEngine.go
```diff
@@ -2110,6 +2110,9 @@ func (e *historyEngineImpl) SignalWorkflowExecution(
 	if err != nil {
 		return err
 	}
+	if domainEntry.GetInfo().Status != persistence.DomainStatusRegistered {
```
yycptt, line 2113: I thought we discussed something about creating a dynamic config for this behavior? I am not sure what's the final decision we made for that?

Shaddoll: I don't think it's worth the effort.

yycptt (approved):
LGTM. @meiliang86 I am not sure yours opinion regarding the config.

PR #4042, opened 2021-03-09
Shadower workflow implementation

yycptt:
**What changed?**
Shadower workflow implementation

**Why?**
Workflow shadowing

**How did you test it?**
unit test

**Potential risks**
N/A, not enabled yet


common/constants.go
```diff
@@ -116,6 +118,10 @@ const (
 	// BatcherLocalDomainName is domain name for batcher workflows running in local cluster
 	// Batcher cannot use SystemLocalDomain because auth
 	BatcherLocalDomainName = "cadence-batcher"
+	// ShadowerDomainID is domain id for workflow shadower local domain
+	ShadowerDomainID = "59c51119-1b41-4a28-986d-d6e377716f82"
```
yux0, line 122: Why do we need the domain ID?

yycptt: Worker will talk to metadata store directly for registering system domain instead of calling frontend. so it needs both the domain name and ID.

yux0: So we also need to add this domain register in the DB setup script?

yycptt: I don't think so. Worker will ensure the domain exists on start up and register the domain if not. I also don't see anything related to `cadence-batcher` in DB setup scripts.

mkolodezny: Will we have the local shadower domain per region?

service/worker/shadower/workflow.go
```diff
@@ -0,0 +1,298 @@
...
+func shadowWorkflow(
+	ctx workflow.Context,
+	params shadower.WorkflowParams,
+) (shadower.WorkflowResult, error) {
+	if err := validateAndFillWorkflowParams(&params); err != nil {
+		return shadower.WorkflowResult{}, err
+	}
+
+	lao := workflow.LocalActivityOptions{
+		ScheduleToCloseTimeout: time.Second * 5,
```
mkolodezny, line 65: what's the average expected execution time for the activity? I thought for a local one i should be up to few seconds.

common/constants.go
yycptt: Yes. It's ok for local domains to have the same domainID in different regions, all system domains follows this pattern.

service/worker/shadower/workflow.go
yycptt: Yes this local activity only accesses domain cache and check if domain is active or not so it will be very fast. 

I don't have strong opinion on the timeout. 5s, 3s or 1s, I don't see a big difference here. Used 5 sec here just to be conservative, in case processing got slowed down when worker hosts are busy with other workflows, like archival or batch.

yux0 (approved):

PR #4044, opened 2021-03-09
Add the service name to the log entry

git-hulk:
It's difficult to identify logs were generated by which service,
so we add the service name to help us filter faster.

**What changed?**

Add the service name to the log entry

**Why?**

It's difficult to identify logs were generated by which service,
so we add the service name to help us filter faster.

**How did you test it?**

Check the log entry whether has the right service name or not

**Potential risks**

No


cmd/server/cadence/server.go
```diff
@@ -108,9 +108,10 @@ func (s *server) startService() common.Daemon {
 	params := service.BootstrapParams{}
 	params.Name = "cadence-" + s.name
 	params.Logger = loggerimpl.NewLogger(s.cfg.Log.NewZapLogger())
+	params.UpdateLoggerWithServiceName(s.name)
```
vytautas-karpavicius, line 111: In general this makes a lot of sense.

However this is not the only place where `service.BootstrapParams` parameters are constructed. There are:
- `host/onebox.go` used for integration tests. It needs updating as well.
- We also construct these params on the internal "Uber" version of Cadence. And we need to make sure to not forget adding this tag on the next release.

git-hulk: Thanks for your review, I have also added the `UpdateLoggerWithServiceName` to other places in onebox.

git-hulk:
hi @vytautas-karpavicius How do you think about PR?

vytautas-karpavicius (approved):
Looks good now!

PR #4056, opened 2021-03-15
Fix missing root dir prefix when loading dynamic config

git-hulk:
The cadence server would fail to load the dynamic config file
when the working dir wasn't the parent dir of dynamic config
file path.

**What changed?**

Fix missing root dir prefix when loading dynamic config

**Why?**

    The cadence server would fail to load the dynamic config file
    when the working dir wasn't the parent dir of dynamic config
    file path.

**How did you test it?**

local test

**Potential risks**

no


git-hulk:
@longquanzheng Can you have a look at this PR?

cmd/server/cadence/cadence.go
```diff
@@ -54,6 +55,7 @@ func startHandler(c *cli.Context) {
 	if cfg.Log.Level == "debug" {
 		log.Printf("config=\n%v\n", cfg.String())
 	}
+	cfg.DynamicConfigClient.Filepath = constructPath(rootDir, cfg.DynamicConfigClient.Filepath)
```
longquanzheng, line 58: Just a very minor comment: 

Could you move this line to server.go right before creating the client? 
```
params.DynamicConfig, err = dynamicconfig.NewFileBasedClient....
```
This will be slightly better in organizing the logic. When finding issues with dynamicConfigClient it's together. 

Here is loading the whole configuration, so better to keep it not involving specific config transformation.

longquanzheng:
> @longquanzheng Can you have a look at this PR?

Thanks for finding and fixing it.

cmd/server/cadence/cadence.go
git-hulk: Thanks for your review,  the `root` dir was from the argument in cli.Context and it's wired if passed the `cli.Context` into newServer. Is it ok that add a private variable like `rootDir` in Config?

longquanzheng: You are right. 

No I feel that's too Intrusive to add that. I don't see a pattern in the code base. 

I'm okay to do it as you have here . 

@yycptt do you have better suggestion?

longquanzheng (approved):

cmd/server/cadence/cadence.go
longquanzheng: Maybe we could have a method called  replaceWithRootDir(cfg) to do that thing. For now only dynamicConfigClient needs it.

git-hulk: another way to work around maybe refactor `config.Load`  to `config.Load(ctx *cli.Context)`, also can avoid too many parameters.

longquanzheng (changes requested):
Actually, this will break the case that using absolute path. 
Maybe we should check if a path is absolute before transformation

git-hulk:
> Actually, this will break the case that using absolute path.
> Maybe we should check if a path is absolute before transformation

I didn't check  the path  since the comment says that should be relative to the root directory

```
// Filepath should be relative to the root directory
```

longquanzheng:
> > Actually, this will break the case that using absolute path.
> > Maybe we should check if a path is absolute before transformation
> 
> I didn't check the path since the comment says that should be relative to the root directory
> 
> ```
> // Filepath should be relative to the root directory
> DynamicConfigClient dynamicconfig.FileBasedClientConfig `yaml:"dynamicConfigClient"`
> ```

I see. That is misleading, sorry for that. We should remove the comment.

maybe that’s the initial intent but the code doesn’t restrict it. As a result, people are using it with absolute paths now. See https://uber-cadence.slack.com/archives/CNC31V3K7/p1616084908003400?thread_ts=1616084757.003300&channel=CNC31V3K7&message_ts=1616084908.003400

so I think we have to support it anyway. Otherwise I don’t think those people have a workaround

git-hulk:
> so I think we have to support it anyway. Otherwise I don’t think those people have a workaround

Okay, I'll fix it and remove the comment later 👍

longquanzheng:
> > so I think we have to support it anyway. Otherwise I don’t think those people have a workaround
> 
> Okay, I'll fix it and remove the comment later 👍

Cool. Can you do the same for loading config? I think it’s okay if config is outside root directory. This will probably make it easier/simpler  to use homebrew. And also add description to the root CLI parameter: it will not apply to absolute path.

git-hulk:
> > > so I think we have to support it anyway. Otherwise I don’t think those people have a workaround
> > 
> > 
> > Okay, I'll fix it and remove the comment later 👍
> 
> Cool. Can you do the same for loading config? I think it’s okay if config is outside root directory. This will probably make it easier/simpler to use homebrew. And also add description to the root CLI parameter: it will not apply to absolute path.

Done, you can have a look again

cmd/server/cadence/cadence.go
```diff
@@ -145,8 +148,13 @@ func getRootDir(c *cli.Context) string {
...
+// constructPathIfNeed would append the dir as the root dir
+// when the file wasn't absolute path.
+func constructPathIfNeed(dir string, file string) string {
+	if !filepath.IsAbs(file) {
+		return dir + "/" + file
+	}
+	return file
 }
 
 // BuildCLI is the main entry point for the cadence server
```
longquanzheng, line 160: Can you also fix line 178? 
maybe :
`config dir path relative to root, or an absolute path`.

longquanzheng (approved):
Just a comment about the CLI parameter. The rest LGTM

cmd/server/cadence/cadence.go
git-hulk: Fixed it, sorry for missing that.

PR #4057, opened 2021-03-16
Created GRPC handlers and internal clients

vytautas-karpavicius:
**What changed?**
This change introduces GRPC handlers and internal clients. It brings simultaneous support for TChannel/Thrift and GRPC/protobuf within cadence server.

For this to work, it opens up new port GRPC port. For consistency all GRPC ports are `-100` from Thrift ports. For example:
``` 
frontend: 7933 -> 7833
history: 7934 -> 7834
matching: 7935 -> 7835
```
On this new port new YARPC inbound is configured to accept GRPC traffic. New GRPC handlers were created to accepts proto encoded requests. They are converted to internal types and passed along to the main handler to process.

The outbound GRPC communication between server components is off by default, and can be enabled with dynamic config flag: `system.enableGRPCOutbound`. If enabled all requests will be encoded to proto messages within newly added GRPC clients. Additionally those clients are created with a different GRPC dispatcher.

When making an outbound call we need to know GRPC port. This is a little bit tricky, as all addresses are resolved using ringpop, which automatically carries TChannel port. Thus we need to drop it and append grpc port by using a lookup from additionally passed service configuration.

**Why?**
To migration Cadence to GRPC.

**How did you test it?**
Integration tests were updated to use GPRC configuration by default.
Locally passing cadence canary with GRPC communication enabled.

**Potential risks**
There may be some mapping issues between proto and internal types that were not noticed yet. Some of them were caught by with integration and canary tests. To reduce risk, proto communication is off by default and will be enabled with dynamic config. Additionally all frontend metrics are tagged with grpc/thrift flag, which will allow observing error increase when switching to grpc dynamically. Mitigation is easy - switching back to Thrift.


common/config/rpc.go
```diff
@@ -71,17 +77,33 @@ func (d *RPCFactory) GetDispatcher() *yarpc.Dispatcher {
 func (d *RPCFactory) createDispatcher() *yarpc.Dispatcher {
```
Shaddoll, line 77: nit: createInboundDispatcher

common/config/rpc.go
```diff
@@ -90,20 +112,44 @@ func (d *RPCFactory) CreateDispatcherForOutbound(
...
+	return d.createOutboundDispatcher(callerName, serviceName, hostName, d.ch.NewSingleOutbound(hostName))
+}
+
+// CreateGRPCDispatcherForOutbound creates a dispatcher for GRPC outbound connection
+func (d *RPCFactory) CreateGRPCDispatcherForOutbound(
+	callerName string,
+	serviceName string,
+	hostName string,
+) (*yarpc.Dispatcher, error) {
+	grpcAddress, err := d.grpcPorts.GetGRPCAddress(serviceName, hostName)
```
Shaddoll, line 125: Are we assuming that all instances use the same grpc port on different hosts?

vytautas-karpavicius: Yes, at least for open source version here.

common/config/rpc.go
vytautas-karpavicius: done

client/frontend/grpcClient.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (g grpcClient) CountWorkflowExecutions(ctx context.Context, request *types.CountWorkflowExecutionsRequest, opts ...yarpc.CallOption) (*types.CountWorkflowExecutionsResponse, error) {
+	response, err := g.visibility.CountWorkflowExecutions(ctx, proto.FromCountWorkflowExecutionsRequest(request), opts...)
+	return proto.ToCountWorkflowExecutionsResponse(response), proto.ToError(err)
+}
+
+func (g grpcClient) DeprecateDomain(ctx context.Context, request *types.DeprecateDomainRequest, opts ...yarpc.CallOption) error {
+	_, err := g.domain.DeprecateDomain(ctx, proto.FromDeprecateDomainRequest(request), opts...)
+	return proto.ToError(err)
```
just-at-uber, line 56: Should this error here?

just-at-uber (approved):

PR #4059, opened 2021-03-17
Improve tasklist and implement history scavenger for SQL

kraney:
There are deficiencies in the TaskList scanner (which only runs for SQL storage) that allow tasks to build up in the `tasks` table.
* scavenger stops processing when it reaches a task that has not expired, regardless of whether that task is completed. But tasks may have infinite timeouts. Not only is such a task itself not garbage collected, no task that comes sequentially after it is garbage collected from that point, either.
* Entries may be deleted from the `task_lists` table while related entries remain in the `tasks` table (no foreign key constraints.) In the case, the scavenger will never attempt to clean up those tasks.

This changeset attempts to address these deficiencies in the following ways
* The "stop" logic in the tasklist scanner has been altered to match the logic used by TaskGC in history; it deletes all tasks less than the persisted ACK level, regardless of expiration time. Since history never reads tasks with an ID less than the ackLevel, this is deemed safe.
* An additional task was added to the tasklist scanner that queries for orphaned tasks, and deletes them when found.
* In support of the above, a new GetOrphanTasks() query was added to the persistence interface. It is left unimplemented for Cassandra, since the tasklist scanner is unused there.

In a similar vein, for SQL storage the history scanner does not run. This allows garbage to accumulate indefinitely into the `history_tree` and `history_node` tables. Seemingly, the reason for this omission is that the GetAllHistoryBranches() query is unimplemented for SQL. This changeset implements that query, and then takes advantage of this to go ahead and enable the history scanner for SQL storage.

Without this change, cruft will accumulate in the database indefinitely, increasing storage used and also increasing DB load and query latency.

These change are patched in from an operational fork of this project. The changes are working in operation, however the patch process may have introduced regressions, the post-patch code is much less well tested.

I ran the automated tests. Some tests that are likely important fail
* I don't know how to make the postgres and mysql tests run correctly; they seem to require a local running instance of the db as a prerequisite.
* I declined to rewrite the test suite for the tasklist scavenger, which is fully predicated on assumptions around stopping once an unexpired task is reached. Those assumptions no longer hold. Someone upstream needs to reason through whether the change is valid & acceptable. (However, one bit of evidence in its favor is it simply mimics the logic of TaskGC.)

If you're using Cassandra, nothing.

If using SQL, the scavenger could consume excessive resources, especially on the first run where there's an existing buildup of unmanaged garbage. It could delete tasks from the task table that should not be deleted. The history scanner similarly could consume excess resources, although the history scanner code is not new; it's identically the same code used for Cassandra. It's only newly enabled for SQL.


longquanzheng:
Wow, thanks for fixing this!!! These are lots of improvements. 

Btw, is this to replace https://github.com/uber/cadence/pull/4019 ?

kraney:
This change picks up on that branch and includes that change, so this PR could be considered a replacement, yes.

common/persistence/persistenceInterface.go
```diff
@@ -70,6 +70,7 @@ type (
 		//  - number of rows actually deleted, if limit is honored
 		//  - UnknownNumRowsDeleted, when all rows below value are deleted
 		CompleteTasksLessThan(ctx context.Context, request *CompleteTasksLessThanRequest) (int, error)
+		GetOrphanTasks(ctx context.Context, request *GetOrphanTasksRequest) (*GetOrphanTasksResponse, error)
```
longquanzheng, line 73: Can you add a comment here. It's better to add a definition of "Orphan", from both user and technical point of view. 
I am still reading the code(thanks for your detailed PR description!), seems like for users, "Orphan task" means the task is completed or expired/timeout? Maybe nice to add a conclusion about technical definition of it.

kraney: An orphan task would be a task record in the task table in the database for which no task list record exists. It no longer represents a valid item that can be acted upon. I can add a comment to that effect.

common/persistence/sql/sqlHistoryManager.go
```diff
@@ -440,10 +446,48 @@ func (m *sqlHistoryV2Manager) GetAllHistoryTreeBranches(
 	ctx context.Context,
 	request *p.GetAllHistoryTreeBranchesRequest,
 ) (*p.GetAllHistoryTreeBranchesResponse, error) {
-
-	// TODO https://github.com/uber/cadence/issues/2458
-	// Implement it when we need
-	panic("not implemented yet")
+	page := historyTreePageToken{
+		ShardID:  0,
+		TreeID:   serialization.MustParseUUID(minUUID),
```
longquanzheng, line 451: Can we replace this with empty binary?
I think It's lower than any other non-empty. 
```
mysql> select * from test_binary order by bid;
+------------------+
| bid              |
+------------------+
|                  |
| 0                |
| a                |
| b                |
+------------------+
4 rows in set (0.01 sec)

mysql> show create table test_binary;
+-------------+----------------------------------------------------------------------------------------------------------------------+
| Table       | Create Table                                                                                                         |
+-------------+----------------------------------------------------------------------------------------------------------------------+
| test_binary | CREATE TABLE `test_binary` (
  `bid` binary(16) NOT NULL,
  PRIMARY KEY (`bid`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 |
+-------------+----------------------------------------------------------------------------------------------------------------------+
```

common/persistence/sql/sqlHistoryManager.go
```diff
@@ -440,10 +446,48 @@ func (m *sqlHistoryV2Manager) GetAllHistoryTreeBranches(
 	ctx context.Context,
 	request *p.GetAllHistoryTreeBranchesRequest,
 ) (*p.GetAllHistoryTreeBranchesResponse, error) {
```
longquanzheng, line 448: Can you help remove this line to help us verify the code? https://github.com/uber/cadence/blob/ed82bb73b4aa7887529773a8fb7b1b1c2926775b/common/persistence/persistence-tests/historyV2PersistenceTest.go#L122
The persistence is already written. 

We like to have high test coverage for persistence layer as they could be building block for any critical features. 

It's fine that we don't have tests for scanner/scavenger, because leaking is very slow, usually people already identify and fix it before they could be an issue.

longquanzheng:
>I don't know how to make the postgres and mysql tests run correctly; they seem to require a local running instance of the db as a prerequisite.

Yes, we have a very rough instructions about testing. Need more improvement, hehe

>I declined to rewrite the test suite for the tasklist scavenger
Totally acceptable. leaking is very slow, usually people already identify and fix it before they could be an issue. In fact it's quite hard to write the tests :(

common/persistence/persistenceInterface.go
longquanzheng: sounds good to me.

common/persistence/sql/sqlplugin/mysql/task.go
```diff
@@ -78,6 +78,12 @@ task_type = :task_type
 	rangeDeleteTaskQry = `DELETE FROM tasks ` +
 		`WHERE domain_id = ? AND task_list_name = ? AND task_type = ? AND task_id <= ? ` +
 		`ORDER BY domain_id,task_list_name,task_type,task_id LIMIT ?`
+
+	getOrphanTaskQry = `SELECT task_id, domain_id, task_list_name, task_type FROM tasks AS t ` +
+		`WHERE NOT EXISTS ( ` +
+		`	SELECT domain_id, name, task_type FROM task_lists AS tl ` +
+		`	WHERE t.domain_id=tl.domain_id and t.task_list_name=tl.name and t.task_type=tl.task_type ` +
```
longquanzheng, line 85: `shard_id` is the first column of the PK, it would be more efficient to pass in that value. 

The shardID is decided by domainId, tasklistName and nShards. So it should be easy to get: 
https://github.com/uber/cadence/blob/bfcb03fe53883ff4d574d9cc41e20dab202d4d52/common/persistence/sql/sqlTaskManager.go#L532


You may argue that will increase the complexity of the code.

common/persistence/sql/sqlplugin/postgres/task.go
```diff
@@ -48,7 +48,8 @@ task_type = :task_type
 
 	listTaskListQry = `SELECT domain_id, range_id, name, task_type, data, data_encoding ` +
 		`FROM task_lists ` +
-		`WHERE shard_id = $1 AND domain_id > $2 AND name > $3 AND task_type > $4 ORDER BY domain_id,name,task_type LIMIT $5`
+		`WHERE shard_id = $1 AND ((domain_id = $2 AND name = $3 AND task_type > $4) OR (domain_id=$2 AND name > $3) OR (domain_id > $2)) ` +
```
longquanzheng, line 51: any reason this is only changed for Postgres, not MySQL?

longquanzheng: Can you add more context/thoughts about this change. Even though the existing one looks broken to me. But I still have a hard time to understand why this new one will work.

kraney: Just an oversight; we actively use the postgres path and not the mysql one.

longquanzheng: I see. thanks.

common/persistence/sql/sqlplugin/mysql/task.go
kraney: It's not clear to me how I could use this knowledge from code in order to update the query. It needs to scan over all tasks. I would argue the efficiency gained by using the index on task_lists is vastly overwhelmed by creating a need to make a separate client request per task so that the shard ID can be calculated in Go for each task row.

The cardinality of the task_lists table is relatively low; much less than the tasks table.

common/persistence/sql/sqlplugin/postgres/task.go
```diff
@@ -48,7 +48,8 @@ task_type = :task_type
...
+		`ORDER BY domain_id,task_type,name LIMIT $5`
```
longquanzheng, line 53: Trying to understand what's the problem and how are you solving it. Here is my learning, please correct me if I am wrong. 
The goal of `ListTasklists` persistence API is to go through every tasklists. Implicitly, there are the requirements:
* Cannot miss any tasklist, cannot return same tasklist more than once
* Must aways move forward, given whatever pagesize. 

Basically we are using PK to implement cursor based pagination(because it's more efficient/scalable than SQL's LIMIT+Offset).


The code here is working tightly with sqlTaskManager's `ListTaskList`. :
sqlTaskManager is relying on PaginationToken to record the status. sqlTaskManager will start with `shardID=0` and loop until `maxShards-1`. 
The contract between the two sqlTaskManger and here(task.go) is that sqlTaskManager will pass in the last record as the cursor. 
For each shard, sqlTaskManager will start with this fake record as cursor:
`taskType=math.MinInt, domainID=minUUID`. And then moving on, it will use last returned record in the result set.
```
		lastRow := &rows[request.PageSize-1]
		nextPageToken, err = gobSerialize(&taskListPageToken{
			ShardID:  pageToken.ShardID,
			DomainID: lastRow.DomainID.String(),
			Name:     lastRow.Name,
			TaskType: lastRow.TaskType,
		})
```

Hence the contract for task.go is to move forward(exclusively) from this last record. 

The existing one is apparently broken, because for example -- 
`domain_id > $2 AND name > $3 AND task_type > $4`, domain_id > $2 would have to skip some records, if the last record's domain still have some other records(tasklists). 

Therefor you are doing:
`((domain_id = $2 AND name = $3 AND task_type > $4) OR (domain_id=$2 AND name > $3) OR (domain_id > $2))`
This will ensure:
* If last record's domain still have other tasklists, then will try to get from them, but they must be greater than last tasklist
* If Not, then get next domain_id( domain_id > $2 )

longquanzheng: Thanks for doing it. Took me a little while to understand. 

btw I was thinking to point you out to another way that I implemented cursor based pagination for our event history store( see https://github.com/uber/cadence/blob/bfcb03fe53883ff4d574d9cc41e20dab202d4d52/common/persistence/sql/sqlHistoryManager.go#L146), but now I believe your change is smaller and safer to do. 

Would you mind help us add some comments to the code? If my above understanding is correct, you may just copy /modify most of them, hehe. Maybe it's better put in sqlTaskManager(like we added comment for sqlHistoryManager), as it has the most logic. 
Comment will help us maintain the code much easier. As people can understand it now, but they will forget it a few months later.

longquanzheng (commented):
Thanks for making such a big contribution. Really, really appreciate it.

common/persistence/sql/sqlplugin/mysql/events.go
```diff
@@ -46,6 +46,8 @@ const (
 	getHistoryTreeQuery = `SELECT branch_id, data, data_encoding FROM history_tree WHERE shard_id = ? AND tree_id = ? `
 
 	deleteHistoryTreeQuery = `DELETE FROM history_tree WHERE shard_id = ? AND tree_id = ? AND branch_id = ? `
+
+	getAllHistoryTreeQuery = `SELECT branch_id, data, data_encoding FROM history_tree WHERE (shard_id = ? AND tree_id = ? AND branch_id > ?) OR (shard_id = ? AND tree_id > ?) OR (shard_id > ?) ORDER BY shard_id, tree_id, branch_id LIMIT ?`
```
longquanzheng, line 50: I believe you have noticed that listTaskList query doesn't use `shard_id>?`, but instead it uses a forloop to iterate over all shards. The reasons is that we were trying to implement in a way that can work for some sharded/partitioned MySQL. In partitioned MySQL, all shards are independent hence we cannot query across shards directly.

However, there is no real usecase of partitioned MySQL...so I won't force you to implement it. And it won't work for your getOrphanTask query which will have to do corss-partition query. 

Would you mind add a TODO here linking to issue that I just opened? https://github.com/uber/cadence/issues/4064

longquanzheng: The comment may be better put in sqlTaskManager

common/persistence/sql/sqlplugin/mysql/task.go
longquanzheng: You are right, I forgot that it's using `WHERE NOT EXISTS` so we can't map to a particular shardID.

common/persistence/sql/sqlTaskManager.go
```diff
@@ -528,6 +528,29 @@ func (m *sqlTaskManager) CompleteTasksLessThan(
 	return int(nRows), nil
 }
 
+// GetOrphanTasks gets tasks from the tasks table that belong to a task_list no longer present
+// in the task_lists table.
+func (m *sqlTaskManager) GetOrphanTasks(ctx context.Context, request *persistence.GetOrphanTasksRequest) (*persistence.GetOrphanTasksResponse, error) {
```
longquanzheng, line 533: Same here: Would you mind add a TODO comment here linking to issue that I just opened? #4064

common/service/dynamicconfig/constants.go
```diff
@@ -827,6 +830,12 @@ const (
 	WorkerThrottledLogRPS
 	// ScannerPersistenceMaxQPS is the maximum rate of persistence calls from worker.Scanner
 	ScannerPersistenceMaxQPS
+	// ScannerMaxOrphanTasks is the maximum number of orphans to delete in one batch
```
longquanzheng, line 833: Can you open another PR in https://github.com/uber/cadence-docs to add the new configuration in https://cadenceworkflow.io/docs/operation-guide/setup/#sysworker-dynamic-configuration ? 

This is a new process we started recently to help our document updated. Otherwise it's super difficult to maintain our documentation.

common/service/dynamicconfig/constants.go
```diff
@@ -827,6 +830,12 @@ const (
...
+	ScannerMaxOrphanTasks
+	// ScannerTaskBatchSize is the number of tasks to query per call in the scavenger
+	ScannerTaskBatchSize
+	// ScannerMaxTasksPerJob is the number of tasks to process before letting another task list be processed
+	ScannerMaxTasksPerJob
```
longquanzheng, line 838: I am still going to understand what exactly the configs are. But so far, there is a gap between the name and the comment. Do you think you have better names for them? The names are important for configuration because many people don't really read description carefully.

kraney: Here I just exposed variables that already existed having hardcoded values. I'm open to suggestions on better naming.

common/service/dynamicconfig/constants.go
kraney: https://github.com/uber/cadence-docs/pull/29

common/service/dynamicconfig/constants.go
longquanzheng: Looks like some names are from existing code. It's okay if the names are not perfect in code. But when we expose as contract for users, we want to improve it. 
Just some ideas to rename:

`ScannerMaxOrphanTasks`=>ScannerGetOrphanTasksPageSize`

longquanzheng: e.g. `ScannerMaxTasksPerJob`=>`ScannerMaxTasksProcessedPerTasklistJob`

service/worker/scanner/workflow.go
```diff
@@ -79,7 +79,8 @@ var (
 		TaskList:                     historyScannerTaskListName,
 		ExecutionStartToCloseTimeout: infiniteDuration,
 		WorkflowIDReusePolicy:        cclient.WorkflowIDReusePolicyAllowDuplicate,
-		CronSchedule:                 "0 */12 * * *",
+		CronSchedule:                 "*/15 * * * *",
```
longquanzheng, line 82: How about making it a dynamic config?

common/service/dynamicconfig/constants.go
longquanzheng: e.g. `ScannerTaskBatchSize` => `ScannerBatchSizeForCompletingExpiredTasks`

common/service/dynamicconfig/constants.go
longquanzheng: thanks!

But also sorry, you may have to update the PR after addressing comments about renames

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,45 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
...
 //
 // Each loop of the handler proceeds as follows
-//    - Retrieve the next batch of tasks sorted by task_id for this task-list from persistence
+//    - Attempt to delete tasks up to the persisted ACK level.
 //    - If there are 0 tasks for this task-list, try deleting the task-list if its idle
-//    - If any of the tasks in the batch isn't expired, we are done. Since tasks are retrieved
-//      in sorted order, if one of the tasks isn't expired, chances are, none of the tasks above
-//      it are expired as well - so, we give up and wait for the next run
-//    - Delete the entire batch of tasks
 //    - If the number of tasks retrieved is less than batchSize, there are no more tasks in the task-list
```
longquanzheng, line 53: this comment is quite hard to understand, maybe too misleading. Maybe say :
`If the number of tasks from completeTasks is less than the batch size, ...`

common/service/dynamicconfig/constants.go
longquanzheng: Also can you update comment/descripiton for`ScannerBatchSizeForCompleteTasksLessThanAckLevel`: 
```
The scavenger rely on it to decide if a tasklist can be deleted. So it's better to keep it a relatively high number to let be more efficient.

**Review by longquanzheng — COMMENTED** [review]
above are mostly all my comments. This PR is very good work. Probably the best one from community so far. 
cc @meiliang86

File: service/worker/scanner/workflow.go (hunk 11 above)
Comment (kraney) [reply service/worker/scanner/workflow.go:82]: Whoops, I wasn't planning on including that change in the patch. OTOH I don't see much advantage in letting these accumulate for 12 hours.

It wouldn't do much good as a dynamic config, because the workflow schedule isn't actually changed unless you explicitly terminate that workflow and let it be recreated.

Let me just put this back to original value for this for this bug fix. Making it changeable should probably be separate work.

File: common/persistence/persistence-tests/visibilityPersistenceTest.go [hunk 13]
```diff
@@ -43,7 +43,7 @@ type (
 		// override suite.Suite.Assertions with require.Assertions; this means that s.NotNil(nil) will stop the test,
 		// not merely log an error
 		*require.Assertions
-		VisibilityMgr             p.VisibilityManager
```
longquanzheng, line 46: make sure run `make lint` and `make fmt` to reformat our code. Otherwise buildkite will fail.

common/persistence/persistenceInterface.go
```diff
@@ -70,6 +70,10 @@ type (
 		//  - number of rows actually deleted, if limit is honored
 		//  - UnknownNumRowsDeleted, when all rows below value are deleted
 		CompleteTasksLessThan(ctx context.Context, request *CompleteTasksLessThanRequest) (int, error)
+		// GetOrphanTasks returns tasks that exist as records in the database but are part of task lists which
+		// _do not_ exist in the database. They are therefore unreachable and no longer represent valid items
+		// that can be legitimately acted upon.
+		GetOrphanTasks(ctx context.Context, request *GetOrphanTasksRequest) (*GetOrphanTasksResponse, error)
```
longquanzheng, line 76: Would you mind adding a persistence test for this API? It should be pretty straightforward to write. We have a convention of high test coverage for persistence as it's very critical layer for Cadence. The test will protect us from breaking it especially when people do refactoring.

https://github.com/uber/cadence/blob/master/common/persistence/persistence-tests/matchingPersistenceTest.go
You can see examples for other APIs.

common/service/dynamicconfig/constants.go
```diff
@@ -827,6 +830,12 @@ const (
 	WorkerThrottledLogRPS
 	// ScannerPersistenceMaxQPS is the maximum rate of persistence calls from worker.Scanner
 	ScannerPersistenceMaxQPS
+	// ScannerGetOrphanTasksPageSize is the maximum number of orphans to delete in one batch
+	ScannerGetOrphanTasksPageSize
+	// ScannerBatchSizeForCompleteTasksLessThanAckLevel is the number of tasks to query per call in the scavenger
+	ScannerBatchSizeForCompleteTasksLessThanAckLevel
+	// ScannerMaxTasksProcessedPerTasklistJob is the number of tasks to process before letting another task list be processed
```
longquanzheng, line 837: Also need to fix the comment/description for ScannerMaxTasksProcessedPerTasklistJob and ScannerBatchSizeForCompleteTasksLessThanAckLevel

maybe:
```
ScannerMaxTasksProcessedPerTasklistJob is the number of tasks to process for a tasklist in each workflow run
```
and
```
ScannerBatchSizeForCompleteTasksLessThanAckLevel is the batch size passed into CompleteTasksLessThanAckLevel to delete completed tasks. When the number of deleted tasks is less than the batch size, the tasklist will also be deleted. 
```

service/worker/scanner/workflow.go
longquanzheng: yeah that sounds good too.

service/worker/scanner/scanner.go
```diff
@@ -289,3 +290,22 @@ func GetScannerContext(ctx context.Context) (scannerContext, error) {
...
+func (cfg *Config) MaxTasksPerJob() int {
+	return cfg.MaxTasksPerJobFn()
+}
+
+// GetOrphanTasksPageSize the number of orphans to query in one call
+func (cfg *Config) GetOrphanTasksPageSize() int {
+	if cfg.GetOrphanTasksPageSizeFn != nil {
+		return cfg.GetOrphanTasksPageSizeFn()
+	}
+	return 16
```
longquanzheng, line 305: We don't set default this way. Can you check other configuration(like in other scanners). I think these tree methods can be removed if following the current convention.

longquanzheng: actually you have set default values in service.go. So you just need to remove the three methods.

service/worker/scanner/tasklist/scavenger_test.go
```diff
@@ -55,6 +55,19 @@ const (
 
 var errTest = errors.New("transient error")
 
+type MockConfig struct {
```
longquanzheng, line 58: like above comment, if it's for testing, you can do similar as https://github.com/uber/cadence/blob/bfcb03fe53883ff4d574d9cc41e20dab202d4d52/service/worker/scanner/history/scavenger_test.go#L73

common/persistence/sql/sqlTaskManager.go
```diff
@@ -309,7 +309,7 @@ func (m *sqlTaskManager) ListTaskList(
 			TaskType: lastRow.TaskType,
 		})
 	case pageToken.ShardID+1 < m.nShards:
-		nextPageToken, err = gobSerialize(&taskListPageToken{ShardID: pageToken.ShardID + 1, TaskType: math.MinInt16})
+		nextPageToken, err = gobSerialize(&taskListPageToken{ShardID: pageToken.ShardID + 1, TaskType: math.MinInt16, DomainID: minUUID})
```
longquanzheng, line 312: also should use empty here instead of minUUID.

kraney: The problem with that change is here it's storing a string, not a struct. I either need a min value that's a string, or I'd need to refactor the struct to store []byte which could have ripple effects.

kraney: Looking at the code that parses UUIDs, it seems to me like the "000..." string would parse to an array of zeroes, which should actually be a minimum.

longquanzheng: That's a valid concern. 

To be backward-compatible, a solution is to add a wrapper function in this file for
line 283 `domainID := serialization.MustParseUUID(pageToken.DomainID)`
to return empty striing, if `pageToken.DomainID` is equal to minUUID.

longquanzheng: Alternatively, I think it's fine if you change the field to `[]byte`. Serialization will fail the activity and then workflow, and after restarting, it will be back. 

I am fine with either way.

longquanzheng: alternatively, we could use another field instead of change the field to []byte. 
For example `DomainUUID`. The existing pagination token will become as empty.

longquanzheng: >"000..." string would parse to an array of zeroes, which should actually be a minimum.

I was under the same impression too. but this issue seems saying that's not true.
https://github.com/uber/cadence/issues/3911
The only explanation is that MySQL did something else when converting a string into binary. 

@yux0 do you have more context about this?

longquanzheng: @kraney how about we add a unit test for `uuid.MustParseUUID`? by comparing that minUUID to 1000 random UUID. That's should be the simplies way to verify your assumption.

longquanzheng: @kraney I am not sure that assumption is valid as MySQL is still a black box there. I tried to find something to validate it but didn't find enough information to confirm. 

If that becomes too much for you(since you have done a lot), you can open another ticket like "Potential bug: Tasklist Scavenger may not scan all domains"  and we can tackle it later.

common/persistence/sql/sqlplugin/mysql/task.go
```diff
@@ -46,9 +46,23 @@ name = :name AND
...
+	//	113
+	//	122
+	//	211
+	// where the hundreds digit corresponds to domain_id, the tens digit
+	// corresponds to name, and the ones digit corresponds to task_type,
+	// Imagine recurring queries with a limit of 1.
+	// For the second query to skip the first result and return 112, it must allow equal values in hundreds & tens, but it's OK because the ones digit is higher.
+	// For the third query, the ones digit is now lower but that's irrelevant because the tens digit is greater.
+	// For the fourth query, both the tens digit and ones digit are now lower but that's again irrelevant because now the hundreds digit is higher.
+	// This technique is useful since the size of the table can easily change between calls, making SKIP an unreliable method, while other db-specific things like rowids are not portable
```
mkolodezny, line 61: nit: just to clarify is it potential bug fix and performance optimization here?

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -154,6 +154,9 @@ func (s *Scavenger) run() {
 		s.stopWG.Done()
 	}()
 
+	// Start a task to delete orphaned tasks from the tasks table
+	s.executor.Submit(&orphanExecutorTask{scvg: s})
```
yycptt, line 158: Can you help me understand what circumstance will result in an orphan task? Thanks.

My understanding is that only this scanner will attempt to delete a task list and it only deletes a task list when:
1. The task list has been idle for a long time. This makes sure no matching hosts still owns the task list.
2. All the tasks in the tasklist have passed the expiration interval and have been successfully deleted.
3. The deletion is a conditional delete based on rangeID. If new tasks are added to the task list, rangeID will be changed and this conditional delete will fail.

common/persistence/sql/sqlplugin/mysql/task.go
yycptt: I think it's a bug fix. The original query will skip records during pagination.

yycptt (commented):
Thanks so much for the contribution! Definitely one of the highest quality PRs from the open-source community and it requires a very deep understanding of Cadence internals. Thank you!

service/worker/scanner/scanner.go
kraney: I put this there in deference to a test failure. For some reason during tests the default was not set causing a nilpointerexception. I couldn't work out where it was supposed to be set in that test - actually I couldn't even tell exactly which test was failing from the test output - so I gave up and did this.

The three methods must be there so that the config object implements the Config interface. This was the least intrusive way to introduce these dynamic config vars to scanner that I could find.

service/worker/scanner/tasklist/scavenger_test.go
kraney: The problem with this approach is it means scavenger needs to store a pointer to scanner.Config. That creates an import cycle, which Go does not allow. Using an interface here breaks the loop.

common/persistence/sql/sqlTaskManager.go
kraney: An all-zero UUID is not strictly the minimum possible value - as you showed an empty UUID is smaller. Maybe that's the issue? But it does seem to be the practical minimum given the actual usage.

common/persistence/sql/sqlplugin/mysql/task.go
kraney: Yes, I would only describe this as a bug fix. I'm not sure it affects performance.

In the prior implementation, to draw from the analogy in the comment, only the first row in the example would be returned. None of the remaining three rows have all of their elements greater than the first, which is what it previously required. This mean the query skipped rather broadly through the dataset.

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,45 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
 // with the assumption that the executor will schedule this task later
 //
 // Each loop of the handler proceeds as follows
-//    - Retrieve the next batch of tasks sorted by task_id for this task-list from persistence
+//    - Attempt to delete tasks up to the persisted ACK level.
```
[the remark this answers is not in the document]
kraney: I can't say with certainty that infinite would be stored, but I saw in direct observation that we had tasks stored which had expiration dates more than 300 years in the future. That effectively neutered scavenger.

service/worker/scanner/tasklist/scavenger.go
kraney: There's an important distinction here between a task being logically expired or invalid and having its database record removed. The DB record always outlasts the validity since you wouldn't want to delete a valid entry.

The TaskGC only runs as a side effect of a PollFor*Tasks call. If the poller for a task list stops polling, then TaskGC won't run and also the ACK level won't be updated. Under that circumstance (and perhaps others) eventually the task list will age out and be eligible for scavenger deletion - despite still having lingering task records which have not been handled.

It's true that all of the tasks would have passed the expiration interval by that time. But that does not imply that they have been handled. The records are still there. This is true both for scavenger when it's time to delete or for history if it updates the lease. In either case, if there are any lingering task records for the old lease, they'll still be there. History should eventually delete them via TaskGC.

The scavenger deletes the task_list entry without regard for whether (expired) task entries exist or not. So if they do, they become orphans. Previously, scavenger only iterated over the task list, so orphans were ignored and never removed.

Between this, some 300-year expiration dates holding off the scavenger, and the query bug that skips large numbers of tasks, our tasks table was over 80% orphans.

common/persistence/sql/sqlTaskManager.go
longquanzheng: I am not sure either. But I did another look at the code, it shouldn't cause any issue if you want to change the field from string to []byte. This long running activity doesn't store any progress in heartbeat details. So when a worker restart, the activity will have to start from very beginning.

service/worker/scanner/tasklist/scavenger_test.go
longquanzheng: >it means scavenger needs to store a pointer to scanner.Config

this is not what I meant. You can put these three fields here, similar to other fields: https://github.com/uber/cadence/blob/master/service/worker/scanner/scanner.go#L69

service/worker/scanner/scanner.go
longquanzheng: Yeah I know what you are looking for. 
So you will create default like this: https://github.com/uber/cadence/blob/master/service/worker/scanner/history/scavenger_test.go#L73

service/worker/scanner/tasklist/scavenger_test.go
kraney: They are already there. The problem is scavenger cannot reference scanner.Config in order to access them because that requires importing scanner - but scanner already imports scavenger. That makes a loop and is not allowed.

kraney: nm I think I see what you mean

common/persistence/sql/sqlplugin/mysql/task.go
mkolodezny: I see, thanks much for the great contribution. We really, really appreciate it!

service/worker/scanner/tasklist/scavenger.go
longquanzheng: From your above info, I think these are the two reasons that you have orphan tasks:
1.  `Previously, scavenger only iterated over the task list, so orphans were ignored and never removed`
2. ` The TaskGC only runs as a side effect of a PollFor*Tasks call. If the poller for a task list stops polling, then TaskGC won't run` 

I think 1 may be the main reason. 2 is a very edge case with much lower possibility. 

I don't understand `History should eventually delete them via TaskGC`

Also, `query bug that skips large numbers of tasks`, the bug of skipping stuff is for tasklist, not task, right? 

Correct me if I am wrong. 

What we are trying to get from here is to understand how an orphan task could occur, as our design try to guarantee this from no happening. 

However, bugs do happen sometimes, so your orphan task scavenger is important in any way.

longquanzheng: ahh I think I know what you meant by `History should eventually delete them via TaskGC`

What you really want to say is :
`Cadence service should eventually delete them via TaskGC`.  Current TaskGC is in Matching service, not History.

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,45 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
...
-		taskID := resp.Tasks[nTasks-1].TaskID
-		if _, err = s.completeTasks(key, taskID, nTasks); err != nil {
+		if err != nil {
+			s.logger.Error(fmt.Sprintf("Scavenger error completing tasks: %v", err))
 			return handlerStatusErr
 		}
-
+		nProcessed += nTasks
 		nDeleted += nTasks
 		if nTasks < taskBatchSize {
```
longquanzheng, line 76: After double thinking, I think this is wrong. Feel free to correct me:

`nTasks < taskBatchSize` is not a safe signal to delete a tasklist. `nTasks` is only the number of tasks that are already acked(polled). It doesn't mean the tasklist is expired, or there are no anymore valid tasks for the tasklist. 

Imagine this extreme case -- when `nTasks = ZERO`, but there are still tasks in the tasklist.  And if you delete the tasklist, then the remaining tasks become orphan(maybe that's the reason that you see so many orphan tasks). The problem is that, the task will be lost...You will see wrong behavior in Cadence workflow. 

However, like you said, the existing logic is also problematic that if see an non-expired task and then stops. So IMO, the deal way is this:
1. use `completeTasks` to complete acked tasks as much as we can. 
2. if `nTasks < taskBatchSize`, then do what the existing code is doing
   2.1 `getTasks(key, taskBatchSize)` to get all the tasks
   2.2 check if expired, and return if one of them is not(because we won't be able to delete the tasklist if so)
   2.3  complete(delete) the tasks up to the last taskID
   2.4. if 2.3 delete less than taskBatchSize, then the tasklist is safe to delete

kraney: This call isn't truly a delete it's a "try delete" and only succeeds if the tasklist LastUpdated value is more than 48 hours old. If it isn't - meaning a history service is still handling the task list - then nothing happens. (And this was preexisting behavior for tryDeleteTaskList.)

I don't think scavenger should do anything at all above the acklevel, whether the task is expired or not.

longquanzheng: >  if the tasklist LastUpdated value is more than 48 hours old

This is not safe, because a task associated to the tasklist could have timeout greater than 48 hours. Simply deleting the tasklist will make the task become orphan. The method is preexisting, but the current way is safely checking all the tasks are expired. 

> I don't think scavenger should do anything at all above the acklevel, whether the task is expired or not.

This is another misunderstanding. The major purpose of this scavenger is to clean up the task and tasklist after expiration. This is a major design for SQL because SQL DBs are missing TTL feature. 

There are some context about the original design here. Sorry for not having enough documentation in Cadence yet. I am happy to provide more information if you have question. 

The main use case for TTL and deleting tasklist is that there is a feature in Cadence called `sticky tasklist`. The tasklist is auto-matcially created per instance per deployment(randomUUID). So we have to clean otherwise they will keep growing as people restart workers. Even though non-sticky tasklist could have the same issue, but less worrying than the sticky ones.

kraney: What you say may be true, but is not related to a behavior introduced by this change. tryDeleteTaskList was always present and always had this behavior. What happens to tasks has changed; what happens to tasklists has not.

kraney: Let me add some context behind this change in hopes it'll help reach mutual understanding.
* Our excess of orphans existed before I changed any code; this change is in response to that, among other things.
* We also had lingering tasks and tasklists for workflows that themselves were long expired and gone.

If I'm understanding your point, I think we're both agreed that completing tasks that are below the ack level is fine, regardless of expiration. The question remains what's correct to do above the ack level.

So the assertion you are making is it is safe to complete tasks that have expired even if the history service might be processing the task list? 

And secondly, that it is unsafe to delete a task list if it has unexpired tasks? If that's so, then I think at the very least the overall workflow state / timeout should also be taken as another factor. It should be safe to delete tasks and tasklists for a nonexistent workflow, regardless of the state of those items.

kraney: I'm still trying to make sure I understand your point, so sorry for more questions...

In the Cassandra case, what happens when TTL is reached? Is there anything that preserves a tasklist entry due to unexpired tasks if nothing has happened to extend the TTL? How long is the TTL?

longquanzheng: > In the Cassandra case, what happens when TTL is reached ...

TTL is one day. It only applies to sticky tasklist. So normal tasklists will never get deleted...
I just realized what we did for SQL is quite different from Cassandra when reading the code --
Every time updating a sticky takslist(when updating ackLevel and rangeID), it will update the TTL again. 
Sticky tasklist is only for sticky decision tasks, they have very short timeout(usually seconds, default to 10s). So when TTL is reached, it means the sticky tasklist hasn't been used for last 24 hours, hence all tasks should have been expired. 

So from reading the code, we will never delete a normal tasklist in Cassandra. 

See more details in 
https://github.com/uber/cadence/blob/b78809c691b08c33bc2e11e6222b0e8fde75d7ee/common/persistence/cassandra/cassandraTaskPersistence.go#L291

longquanzheng: Just a little more background about tasklist table. It stores the metadata mainly for 
* rangeID: ownership management, and taskID management
  * everytime a matching host claim ownership of a tasklist, it must increase the value succesfully 
  * also used as base section of the taskID. E.g, if rangeID is 1, then allowed taskID ranged will be [100K, 2*100K-1], see more details in https://github.com/uber/cadence/blob/b78809c691b08c33bc2e11e6222b0e8fde75d7ee/service/matching/taskWriter.go#L129
* ackLevel: max taskID that can be safely deleted.

If there are non-expired tasks, but we delete the associated tasklist, the rangeID and ackLevel will be lost.  As a result, if   historySerive is calling matching service to add a new task, these two number will be starting from ZERO. This will cause some tasks will be overwritten as some taskIDs will be reused.

longquanzheng: Back to the current case of SQL tasklist scanner: It grabs a batch of tasks, and check if all of them expired, then delete the tasks and also the tasklist. 
This logic matches what we did in Cassandra for sticky tasklist. As sticky tasks only have short timeouts. 

However, this doesn't match what we did in Cassandra for normal tasklist. In Cassandra, most of the tasks will use TTL to clean up the tasks automatically. See https://github.com/uber/cadence/blob/b78809c691b08c33bc2e11e6222b0e8fde75d7ee/common/persistence/cassandra/cassandraTaskPersistence.go#L421 However, the  current SQL scanner will stop once it sees a task hasn't expired yet, without looking at other tasks. This could be a problem because in normal tasklist, decision task has a 1 year timeout by default. As a result, all the tasks will be only get deleted after one year. 

The existing code doesn't seem to have real issue of orphan tasks. I believe they are introduced by your change. By design tasklist will only get deleted only tasks are all deleted. 
However, I think your orphan tasks scanner is till nice to have. There could be bugs that we don't know. And should add some error logs into this scanner, because this is not expected to see.

service/worker/scanner/tasklist/handler.go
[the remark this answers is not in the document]
yycptt: Sorry, just want to make myself clear here, in case there's any misunderstanding. 
- I am not against the idea of deleting acked tasks in scanner, as it can cover some rare cases where acked tasks are not deleted in time. 
- **Regardless of a task is acked or not, as long as it has passed the expiration, it needs to be deleted, even if it has a taskID higher than the ack level.** This is the main purpose of task list scanner since there're cases where a task may never be acked.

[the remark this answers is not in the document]
yycptt: I am not sure where the 300 year expiration comes from. From design perspective it should not happen and there should be a 1 year limit for the expiration in history service when pushing tasks to matching. If you do see an unreasonable expiration interval, it must be a bug somewhere in the system and we need to fix it.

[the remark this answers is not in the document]
kraney: Thanks for the confirmation about deleting expired tasks.

I'm not sure on the long expiration either. I am going to add some logging to see if I may get to the bottom of it. (It's difficult to directly search for such tasks since the expiration date is "hidden" in the binary data from a db query point of view.) One thought I had on the topic; if a retry policy is set its duration will override the start-to-close timeout. Maybe it's possible to set a retry policy that implies a very long timeout and thus bypass the limits?

service/worker/scanner/tasklist/handler.go
kraney: After reflecting on this overnight I was able to (I believe) understand your explanation. I'll bring back code checking for expiration. This will also bring back the "stuck scavenger" issue we were having for long-expiration tasks, so I'm also going to think about how to address that safely.

service/worker/scanner/tasklist/handler.go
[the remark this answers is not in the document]
kraney: I still see unreasonable expiration times on a handful of tasks. Now that I'm logging them, I can see they have the following properties:
* The date stored in the DB is the value 11651379494838206464 which convert to Tuesday, March 21, 2339 10:18:14.838 PM
* This actually is interpreted as 1754-08-30 22:43:41.128654848 +0000 UTC presumably due to overflow when converting from 64 bit epoch nanoseconds to 32 bit epoch seconds.
* However isTaskExpired also checks for t.Expiry.After(epochStartTime) so it still treats this negative time as unexpired. The presence of the check makes me think these overflow values are not a new surprise.

All of these tasks have identically the same expiration value stored in the database, which makes me think it must have been statically initialized from some value.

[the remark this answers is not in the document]
longquanzheng: Can you file another bug/issue with more information so that we can fix it separately?:
* server version
* tasklist type(decision/activity)
* tasklist kind(sticky/non-stikcy)
* the deserialized taskInfo object so that we can see `workflowID,runID, scheduledID, expiryTime and createdTime`

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,84 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
...
+		if nTasks < taskBatchSize {
+			break
+		}
+	}
+	for nProcessed < max {
+		// if we finished completing tasks below the ack level, but still have budget, then focus on expired tasks
+		resp, err1 := s.getTasks(info, taskBatchSize)
 		if err1 != nil {
 			err = err1
+			s.logger.Error(fmt.Sprintf("Scavenger error getting tasks: %v", err))
```
longquanzheng, line 86: We want to use tag.Error(err) instead of `Sprintf`. 
This is because having a stable/static message string will make it possible for us to use it to build some dashboard

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,84 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
...
-			return handlerStatusDone
-		}
-
+		nTasks := 0
+		unexpiredTaskFound := false
 		for _, task := range resp.Tasks {
 			nProcessed++
 			if !s.isTaskExpired(task) {
-				return handlerStatusDone
+				s.logger.Info(fmt.Sprintf("Scavenger stopping at an unexpired task. (Expires %v)", task.Expiry), tag.WorkflowTaskListName(info.Name), tag.TaskType(info.TaskType), tag.WorkflowDomainID(task.DomainID), tag.WorkflowID(task.WorkflowID), tag.WorkflowRunID(task.RunID), tag.TaskID(task.TaskID))
```
longquanzheng, line 95: Same here, use `tag.Timestamp(taskExpiry)`

service/worker/scanner/tasklist/handler.go
```diff
@@ -47,60 +48,84 @@ const scannerTaskListPrefix = "cadence-sys-tl-scanner"
...
+				unexpiredTaskFound = true
```
longquanzheng, line 96: Why not returning directly like the exisitng code does?

common/service/dynamicconfig/constants.go
```diff
@@ -828,6 +831,12 @@ const (
 	WorkerThrottledLogRPS
 	// ScannerPersistenceMaxQPS is the maximum rate of persistence calls from worker.Scanner
 	ScannerPersistenceMaxQPS
+	// ScannerGetOrphanTasksPageSize is the maximum number of orphans to delete in one batch
+	ScannerGetOrphanTasksPageSize
+	// ScannerBatchSizeForCompleteTasksLessThanAckLevel is the number of tasks to query per call in the scavenger. The scavenger then uses the return of CompleteTasksLessThanAckLevel, to decide if a tasklist can be deleted. So it's better to keep it a relatively high number to let it be more efficient.
+	ScannerBatchSizeForCompleteTasksLessThanAckLevel
```
longquanzheng, line 837: Now it's being used for both completeTasksLessThanAckLevel and `getTaskForCheckExpiry`...
Do you think we should separate them? or rename it to something else that make more sense(and revise the comment too). 

I am open to either one.

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -154,6 +153,9 @@ func (s *Scavenger) run() {
...
```
longquanzheng, line 157: Can you help to use a new config `worker.enableCleaningOrphanTaskInTasklistScavenger` and disable the behavior by default. 

Mostly because it could have some perf issue for some users and we don't want to introduce this by default.

longquanzheng (changes requested):

longquanzheng (approved):

PR #4063, opened 2021-03-19
Fix time convertion to nanos

vytautas-karpavicius:
**What changed?**
Trying to handle zero value of `time.Time{}` conversion.

**Why?**
Converting zero value of `time.Time{}` to nanos and back does not result in the same value. It is documented that calling `UnixNano()` on zero value is undefined. It returns weird number which converted back to time results in `1754-08-30T22:43:41Z`. Repro: https://play.golang.org/p/qvhsG0_S6Vs

**How did you test it?**
Updated unit tests with such case.

**Potential risks**

### DIFF
--- a/common/persistence/serialization/thrift_mapper.go
+++ b/common/persistence/serialization/thrift_mapper.go
@@ -690,6 +690,8 @@ func replicationTaskInfoFromThrift(info *sqlblobs.ReplicationTaskInfo) *Replicat
 	}
 }
 
+var zeroTimeNanos = time.Time{}.UnixNano()
+
 func unixNanoPtr(t *time.Time) *int64 {
 	if t == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


andrewjdawson2016 (approved):

common/persistence/serialization/thrift_mapper.go
```diff
@@ -701,6 +703,11 @@ func timePtr(t *int64) *time.Time {
 	if t == nil {
 		return nil
 	}
+	// Calling UnixNano() on zero time is undefined an results in a number that is converted back to 1754-08-30T22:43:41Z
+	// Handle such case explicitly
+	if *t == zeroTimeNanos {
```
mkolodezny, line 708: idem

common/persistence/serialization/thrift_mapper.go
```diff
@@ -690,6 +690,8 @@ func replicationTaskInfoFromThrift(info *sqlblobs.ReplicationTaskInfo) *Replicat
 	}
 }
 
+var zeroTimeNanos = time.Time{}.UnixNano()
```
mkolodezny, line 693: nit: can the Time.IsZero() function be used instead: https://golang.org/src/time/time.go?s=15971:15998#L416

vytautas-karpavicius: Not in this case. IsZero can tell whether time.Time isself is zero. But we need to check corresponding unix nanos value, that got persisted. And it gets this weird value, due to some internal overflow.

PR #4069, opened 2021-03-23
Add a domain cluster filter during replication

yux0:
**What changed?**
Add a domain cluster filter during replication

**Why?**
Save one DB call to fetch workflow history in replication

**How did you test it?**
Unit tests

**Potential risks**
N/A

### DIFF
--- a/service/history/replication/task_ack_manager.go
+++ b/service/history/replication/task_ack_manager.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/backoff"
+	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/collection"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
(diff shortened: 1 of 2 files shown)


service/history/replication/task_ack_manager.go
```diff
@@ -155,14 +156,23 @@ func (t *taskAckManagerImpl) GetTasks(
 	readLevel := lastReadTaskID
 TaskInfoLoop:
 	for _, taskInfo := range taskInfoList {
+		// filter task info by domain clusters.
+		domainEntity, err := t.shard.GetDomainCache().GetDomainByID(taskInfo.GetDomainID())
```
mkolodezny, line 160: do we assume here the domain from the db is always in the cache?

service/history/replication/task_ack_manager.go
```diff
@@ -614,6 +624,15 @@ func (t *taskAckManagerImpl) generateHistoryReplicationTask(
 	)
 }
 
+func skipTask(pollingCluster string, domainEntity *cache.DomainCacheEntry) bool {
+	for _, cluster := range domainEntity.GetReplicationConfig().Clusters {
```
mkolodezny, line 628: idem

service/history/replication/task_ack_manager.go
yux0: Yes. Domains are always in cache

yycptt (approved):

PR #4070, opened 2021-03-24
YARPC error mapping

vytautas-karpavicius:
**What changed?**
Use YARPC error model instead of vanilla grpc errors. As we use YARPC framework, errors needs to use it as well, to get properly serialized and deserialized. Otherwise all errors gets mapped to Unknown category and breaks logic for the caller which expect specific errors to be returned.

**Why?**
This is a prerequisite for https://github.com/uber/cadence/pull/4057. Splitting it off here for easier review.

**How did you test it?**
Unit tests.

**Potential risks**

### DIFF
--- a/common/types/mapper/proto/errors.go
+++ b/common/types/mapper/proto/errors.go
@@ -23,8 +23,8 @@ package proto
 import (
 	"errors"
 
-	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


mantas-sidlauskas (approved):
lgtm

PR #4071, opened 2021-03-24
Use gogofast to register types using gogo registry

vytautas-karpavicius:
**What changed?**
Changed proto compiler from `gofast` to `gogofast`. Those are identical except for the part that `gogofast` imports gogo library within generated files intsead golang/protobuf.

**Why?**
This is a necessary prerequisite for https://github.com/uber/cadence/pull/4057. Serializing errors with details involves marshaling them as `google.protobuf.Any` messages. This need to know [message name](https://github.com/gogo/protobuf/blob/master/proto/properties.go#L578). It can either be set using gogo extensions  (which we want to avoid using for easier migration off gogo in the future) or type must be registered within gogo registry. Currently this is [not the case](https://github.com/uber/cadence/blob/master/.gen/proto/api/v1/error.pb.go#L480).

**How did you test it?**

**Potential risks**


mantas-sidlauskas (approved):
sgtm

PR #4072, opened 2021-03-24
Add version to Kafka config

mistikel:
**What changed?**
Expose Kafka version for configuration

**Why?**
This need to be allowed to configured, so that Cadence can work with different versions of Kafka.

**How did you test it?**

**Potential risks**
Breaking changes if the config is not updated


longquanzheng (approved):
Thanks for the contribution.

common/service/config/kafkaConfig.go
```diff
@@ -62,6 +64,12 @@ func (k *KafkaConfig) Validate(checkApp bool) {
 	if len(k.Topics) == 0 {
 		panic("Empty Topics Config")
 	}
+	if len(k.Version) < 5 {
+		panic("Invalid Kafka Version")
```
longquanzheng, line 68: Let's avoid breaking change. We can set a default version here if this is empty

longquanzheng (changes requested):

common/messaging/kafka/clientImpl.go
```diff
@@ -86,16 +86,20 @@ func NewKafkaClient(
...
 func (c *clientImpl) NewConsumer(app, consumerName string) (messaging.Consumer, error) {
 	topics := c.config.GetTopicsForApplication(app)
+	version, err := sarama.ParseKafkaVersion(c.config.GetKafkaVersionForApplication(app))
+	if err != nil {
+		return nil, err
+	}
 	saramaConfig := sarama.NewConfig()
 	// bellow config is copied from uber/kafka-clientImpl bo keep the same behavior
-	saramaConfig.Version = sarama.V0_10_2_0
+	saramaConfig.Version = version
```
mkolodezny, line 95: do we presume to support the all kafka versions here?

common/service/config/kafkaConfig.go
```diff
@@ -100,3 +108,8 @@ func (k *KafkaConfig) GetBrokersForKafkaCluster(kafkaCluster string) []string {
 func (k *KafkaConfig) GetTopicsForApplication(app string) TopicList {
 	return k.Applications[app]
 }
+
+// GetKafkaVersionForApplication gets kafka version from application
+func (k *KafkaConfig) GetKafkaVersionForApplication(app string) string {
+	return k.Version
```
mkolodezny, line 114: idem

common/messaging/kafka/clientImpl.go
longquanzheng: I think so, unless there is a reason not to do so.

mistikel:
@longquanzheng @mkolodezny any update?

common/messaging/kafka/clientImpl.go
```diff
@@ -86,16 +86,20 @@ func NewKafkaClient(
 // NewConsumer is used to create a Kafka consumer
 func (c *clientImpl) NewConsumer(app, consumerName string) (messaging.Consumer, error) {
 	topics := c.config.GetTopicsForApplication(app)
+	version, err := sarama.ParseKafkaVersion(c.config.GetKafkaVersionForApplication())
+	if err != nil {
+		version = sarama.V0_10_2_0
```
longquanzheng, line 91: Swallowing errors is not a good idea. There could be some errors that we want to let users be aware of. 
You could do:
```go
kafkaConfig := c.config.GetKafkaVersionForApplication()
if kafkaConfig.version == ""{
     kafkaConfig.version = "0.10.2.0"
}
version, err := sarama.ParseKafkaVersion(kafkaConfig)
...
```

longquanzheng (changes requested):

common/service/config/kafkaConfig.go
```diff
@@ -100,3 +101,8 @@ func (k *KafkaConfig) GetBrokersForKafkaCluster(kafkaCluster string) []string {
 func (k *KafkaConfig) GetTopicsForApplication(app string) TopicList {
 	return k.Applications[app]
 }
+
+// GetKafkaVersionForApplication gets kafka version from application
+func (k *KafkaConfig) GetKafkaVersionForApplication() string {
```
longquanzheng, line 106: I think we should remove this method. 
The pattern of "ForApplication" is to exact config from `k.Application` 
For this version, just get it directly

common/messaging/kafka/clientImpl.go
```diff
@@ -86,16 +86,26 @@ func NewKafkaClient(
 // NewConsumer is used to create a Kafka consumer
 func (c *clientImpl) NewConsumer(app, consumerName string) (messaging.Consumer, error) {
 	topics := c.config.GetTopicsForApplication(app)
+	kafkaVersion := c.config.GetKafkaVersionForApplication()
```
longquanzheng, line 89: kafKaVerison := c.Version

longquanzheng (changes requested):
sorry, one more.

longquanzheng (approved):
thanks for your contribution. 
it LGTM now. 

@mkolodezny let us know if you have other comments.

yux0 (approved):

PR #4073, opened 2021-03-24
Fix emit 0 count if encounters error

yux0:
**What changed?**
Fix emit 0 count if encounters error

**Why?**
Fix emit 0 count if encounters error

**How did you test it?**

**Potential risks**

### DIFF
--- a/common/domain/replication_queue.go
+++ b/common/domain/replication_queue.go
@@ -336,13 +336,14 @@ func (q *replicationQueueImpl) emitDLQSize() {
 				).IncCounter(
 					metrics.DomainReplicationQueueSizeErrorCount,
 				)
+			} else {
+				q.metricsClient.Scope(
+					metrics.DomainReplicationQueueScope,
+				).UpdateGauge(
+					metrics.DomainReplicationQueueSizeGauge,
+					float64(size),
+				)
 			}
-			q.metricsClient.Scope(
-				metrics.DomainReplicationQueueScope,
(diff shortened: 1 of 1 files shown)


yycptt (approved):

common/domain/replication_queue.go
```diff
@@ -336,13 +336,14 @@ func (q *replicationQueueImpl) emitDLQSize() {
 				).IncCounter(
 					metrics.DomainReplicationQueueSizeErrorCount,
 				)
+			} else {
+				q.metricsClient.Scope(
```
mkolodezny, line 340: is there the corresponding metric on the dash?

common/domain/replication_queue.go
```diff
@@ -336,13 +336,14 @@ func (q *replicationQueueImpl) emitDLQSize() {
...
+					metrics.DomainReplicationQueueScope,
+				).UpdateGauge(
+					metrics.DomainReplicationQueueSizeGauge,
+					float64(size),
+				)
 			}
-			q.metricsClient.Scope(
```
mkolodezny, line 340: nit: lets make sure it wont make the existing alerts going off.

common/domain/replication_queue.go
yux0: Yes

common/domain/replication_queue.go
yux0: I updated the alert to use this one.

PR #4076, opened 2021-03-25
Added missing enum value for WorkflowExecutionCloseStatus

vytautas-karpavicius:
**What changed?**
Include missing value `WORKFLOW_EXECUTION_CLOSE_STATUS_NONE ` value for `WorkflowExecutionCloseStatus`. Those values are based on persistence constants defined [here](https://github.com/uber/cadence/blob/master/common/persistence/dataInterfaces.go#L120).

**Why?**
This enum was introduced for proto only. Thrift continues to pass integers, thus it is not relevant there.
`None` value was not included initially. Including it now, otherwise other values are not mapped correctly.

**How did you test it?**
Noticed this while testing grpc with integration tests.

**Potential risks**


vytautas-karpavicius:
This is actually causes other problems. We will need another mapping from persistence constants to internal types.

PR #4079, opened 2021-03-26
Add domain queries to scanner and fixer, and skip terminated scans

emrahs:
- Added `domain_report` query handlers to both Scanner and Fixer workflows
- Updated the Fixer workflow to ensure it only picks up the `continued_as_new` scanner executions by default

- Domain queries are useful for troubleshooting and fixer rollouts. 
- Picking the latest `continued_as_new` scanner execution is important to avoid picking up a terminated or failed execution which could provide an incomplete/incorrect report. This only affects the default behavior, and operators can still run the fixer with a specific Scanner execution as they wish.

- Unit testing
- Manual e2e testing with intentionally created corruptions

In the worst case, the availability of Scanner or Fixer could be impacted such that they may just stop working or fail to serve data through the query handlers.

### DIFF
--- a/service/worker/scanner/shardscanner/aggregators.go
+++ b/service/worker/scanner/shardscanner/aggregators.go
@@ -24,6 +24,7 @@ package shardscanner
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/uber/cadence/common/reconciliation/invariant"
 	"github.com/uber/cadence/common/reconciliation/store"
@@ -169,6 +170,7 @@ type (
 		maxShard int
 
 		reports       map[int]FixReport
+		domainStats   map[string]*FixStats
 		status        ShardStatusResult
 		statusSummary ShardStatusSummaryResult
 		aggregation   AggregateFixReportResult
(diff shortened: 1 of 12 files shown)


service/worker/scanner/shardscanner/aggregators.go
```diff
@@ -234,6 +255,38 @@ func (a *ShardFixResultAggregator) GetStatusResult(req PaginatedShardQueryReques
...
+// GetStatusResult returns stats broken by domain IDs
+func (a *ShardFixResultAggregator) GetDomainStatus(req DomainReportQueryRequest) (*DomainFixReportQueryResult, error) {
+	responseStats := []DomainFixStats{}
+
+	if req.DomainID == nil {
+		domainIDs := make([]string, 0)
+		for domainID := range a.domainStats {
+			domainIDs = append(domainIDs, domainID)
+		}
+		sort.Strings(domainIDs)
```
mantas-sidlauskas, line 267: just curious, why sort domain ids?

emrahs: Since the domainStats is a map, ordering isn't consistent. I just find it easier to view reports when sorted, compare between versions, write the unit test, etc. Performance shouldn't be an issue since the # of domains per cluster is around 100s only.

mantas-sidlauskas (approved):

PR #4082, opened 2021-03-26
Revert "Drop pointer on ActivityTaskScheduledEventAttributes.Domain (#4009)"

yycptt:
**What changed?**
This reverts commit 4cd869c49a3ea3d1abd405d53be9cd9f8ac8db22. as it's not compatible with 0.19 and previous releases.

Since logic in our previous releases assumes if this field is not nil then it's also not an empty string, the solution is:
- For internal types, added the pointer back for domain field, otherwise the history persisted will be viewed as a pointer to empty string by previous releases and cause issues.
- For thrift types, nothing changes, it always has the pointer. When converting to internal types, domain field will never be a pointer to empty string (technically it can, but there's no code path lead to it).
- For proto, nothing changes, it **DOES NOT** have the pointer. When converting to internal types, domain field **will be** a pointer to empty string. The logic for checking that field has been updated to check if the string is empty or not instead of if just a nil pointer. Since proto is will not be production ready soon, it's unlikely user are still using 0.19 server when it's ready.

**Why?**
Domain field in ActivityScheduledEvent is an optional field for scheduling activity in a different domain.

**How did you test it?**
Local canary.

**Potential risks**
N/A

### DIFF
--- a/common/types/shared.go
+++ b/common/types/shared.go
@@ -252,7 +252,7 @@ func (v *ActivityTaskFailedEventAttributes) GetIdentity() (o string) {
 type ActivityTaskScheduledEventAttributes struct {
 	ActivityID                    string        `json:"activityId,omitempty"`
 	ActivityType                  *ActivityType `json:"activityType,omitempty"`
-	Domain                        string        `json:"domain,omitempty"`
+	Domain                        *string       `json:"domain,omitempty"`
 	TaskList                      *TaskList     `json:"taskList,omitempty"`
 	Input                         []byte        `json:"input,omitempty"`
 	ScheduleToCloseTimeoutSeconds *int32        `json:"scheduleToCloseTimeoutSeconds,omitempty"`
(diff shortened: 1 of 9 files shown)


common/types/shared.go
```diff
@@ -252,7 +252,7 @@ func (v *ActivityTaskFailedEventAttributes) GetIdentity() (o string) {
 type ActivityTaskScheduledEventAttributes struct {
 	ActivityID                    string        `json:"activityId,omitempty"`
 	ActivityType                  *ActivityType `json:"activityType,omitempty"`
-	Domain                        string        `json:"domain,omitempty"`
+	Domain                        *string       `json:"domain,omitempty"`
```
mkolodezny, line 255: does the idl needs to be updated as well?

mkolodezny (approved):

common/types/shared.go
yycptt: thrift generated files always have pointer. We only dropped pointer for internal types I think.

vytautas-karpavicius (approved):

PR #4083, opened 2021-03-26
Fix Dockerfile

yycptt:
**What changed?**
- Remove `.bin` and `.build` folder when building docker images
- Fix typo in `.fake-codegen` and `.fake-protoc`
- Build copyright binary if not exists

**Why?**
Current Dockerfile is broken

**How did you test it?**
Tested locally

**Potential risks**
N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,9 +26,10 @@ COPY go.* ./
 RUN go mod download
 
 COPY . .
+RUN rm -fr .bin .build
 
 # bypass codegen, use committed files.  must be run separately, before building things.
-RUN make .faux-codegen
+RUN make .fake-codegen
 RUN CGO_ENABLED=0 make copyright cadence-cassandra-tool cadence-sql-tool cadence cadence-server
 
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -320,7 +320,7 @@ lint: ## (re)run the linter
 fmt: $(BUILD)/fmt ## run goimports
 
 # not identical to the intermediate target, but does provide the same codegen (or more).
-copyright: ## update copyright headers
+copyright: $(BIN)/copyright ## update copyright headers
```
Groxx, line 323: oops.  I must've forgotten this one.  sorry!

Groxx (approved):

Dockerfile
```diff
@@ -26,9 +26,10 @@ COPY go.* ./
 RUN go mod download
 
 COPY . .
+RUN rm -fr .bin .build
 
 # bypass codegen, use committed files.  must be run separately, before building things.
-RUN make .faux-codegen
```
Groxx, line 31: bleh.  how did I miss this one too

mkolodezny (approved):

Makefile
mkolodezny: did it cause the "syntax error: unexpected word (expecting ")")"?

Groxx: I don't think so. The error message seemed to be from something make was running, not make itself

PR #4088, opened 2021-03-29
Treat persistence timeout as transient error

yycptt:
**What changed?**
- Treat persistence timeout as transient error
- Move `IsPersistenceTransientError` to persistence package to avoid cycle dependencies between `common` and `persistence` package.

**Why?**
Persistence timeout is a transient error and operation should be retried when encountering that error

**How did you test it?**
Unit test

**Potential risks**
If timeout is due to high number of requests, retry will result in more requests and make the situation worse.


yux0 (approved):

yycptt:
Timeout error is not also retryable, need to double check all references to this function.

PR #4089, opened 2021-03-29
Add delayed start support in handler

demirkayaender:
**What changed?**
Changed workflowHandler to use delayStartSeconds param from
StartWorkflowExecutionRequest and delay starting the workflow accordingly. 

**Why?**
Workflows will be able to start with a delay

**How did you test it?**
```
go test -v github.com/uber/cadence/service/frontend -run TestWorkflowHandlerSuite -testify.m TestStartWorkflowExecution_Failed_BadDelayStartSeconds
go test -v github.com/uber/cadence/common -testify.m TestCreateHistoryStartWorkflowRequest_DelayStart
make test
```

**Potential risks**
If there is a bug in how delays work, workflow starts might get delayed infinitely.


common/util.go
```diff
@@ -431,12 +432,22 @@ func CreateHistoryStartWorkflowRequest(
 	domainID string,
 	startRequest *types.StartWorkflowExecutionRequest,
 	now time.Time,
-) *types.HistoryStartWorkflowExecutionRequest {
+) (resp *types.HistoryStartWorkflowExecutionRequest, retError error) {
```
yycptt, line 435: do we need named return parameter in this case, looks like `resp` is never used and a new `histRequest` is created and returned

common/util.go
```diff
@@ -103,6 +103,7 @@ var (
 	ErrContextTimeoutTooShort = &types.BadRequestError{Message: "Context timeout is too short."}
 	// ErrContextTimeoutNotSet is error for not setting a context timeout when calling a long poll API
 	ErrContextTimeoutNotSet = &types.BadRequestError{Message: "Context timeout is not set."}
+	ErrDelayStartSeconds    = &types.BadRequestError{Message: "A valid DelayStartSeconds is not set on request."}
```
yycptt, line 106: nit: can we make the message a bit more specific?

service/history/historyEngine.go
```diff
@@ -2758,7 +2765,7 @@ func getScheduleID(
 func getStartRequest(
 	domainID string,
 	request *types.SignalWithStartWorkflowExecutionRequest,
-) *types.HistoryStartWorkflowExecutionRequest {
+) (historyReq *types.HistoryStartWorkflowExecutionRequest, retError error) {
```
yycptt, line 2768: same here?

service/frontend/workflowHandler.go
```diff
@@ -1829,8 +1834,13 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 	}
 
 	wh.GetLogger().Debug("Start workflow execution request domainID", tag.WorkflowDomainID(domainID))
-	resp, err = wh.GetHistoryClient().
-		StartWorkflowExecution(ctx, common.CreateHistoryStartWorkflowRequest(domainID, startRequest, time.Now()))
+	historyRequest, err := common.CreateHistoryStartWorkflowRequest(
+		domainID, startRequest, time.Now())
+	if historyRequest == nil {
```
yycptt, line 1839: let's check for error instead of returned requests

service/frontend/workflowHandler.go
```diff
@@ -1787,6 +1788,10 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 		return nil, wh.error(errInvalidTaskStartToCloseTimeoutSeconds, scope)
 	}
 
+	if startRequest.GetDelayStartSeconds() <= 0 {
```
yux0, line 1791: Does delay start seconds allow 0?

common/util.go
```diff
@@ -103,6 +103,7 @@ var (
 	ErrContextTimeoutTooShort = &types.BadRequestError{Message: "Context timeout is too short."}
 	// ErrContextTimeoutNotSet is error for not setting a context timeout when calling a long poll API
 	ErrContextTimeoutNotSet = &types.BadRequestError{Message: "Context timeout is not set."}
+	ErrDelayStartSeconds    = &types.BadRequestError{Message: "Conflicting inputs: both DelayStartSeconds and Cron schedule is set"}
```
yux0, line 106: Still think we should support this. We can sync offline

common/types/shared.go
```diff
@@ -7955,6 +7964,7 @@ type StartChildWorkflowExecutionInitiatedEventAttributes struct {
 	Header                              *Header                `json:"header,omitempty"`
 	Memo                                *Memo                  `json:"memo,omitempty"`
 	SearchAttributes                    *SearchAttributes      `json:"searchAttributes,omitempty"`
+	DelayStartSeconds                   *int32                 `json:"delayStartSeconds,omitempty"`
```
yux0, line 7967: what about DelayStartInSeconds

common/util.go
```diff
@@ -431,12 +432,22 @@ func CreateHistoryStartWorkflowRequest(
...
 	now time.Time,
-) *types.HistoryStartWorkflowExecutionRequest {
+) (*types.HistoryStartWorkflowExecutionRequest, error) {
 	histRequest := &types.HistoryStartWorkflowExecutionRequest{
 		DomainUUID:   domainID,
 		StartRequest: startRequest,
 	}
-	firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(startRequest.GetCronSchedule(), now, now)
+
+	firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(
```
yux0, line 441: I am thinking:
startTime := now.Add(time.Second * delayStartSeconds)
firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(startRequest.GetCronSchedule(), startTime, startTime)

So. you don't need to return an error.

service/frontend/workflowHandler.go
demirkayaender: yes, I replaced this with "<", will update the pr

common/util.go
demirkayaender: I think, in the meeting we decided to make this mutually exclusive for now, then support with another change.

common/types/shared.go
demirkayaender: I chose this naming since it was consistent with other variables like taskStartToCloseTimeoutSeconds, executionStartToCloseTimeoutSeconds in the same struct

common/util.go
demirkayaender: I think there's more than to this than how to implement it. We need to clearly explain this behavior. Let's talk about this behavior in an ERD after this change then update here.

PR #4090, opened 2021-03-29
Fix CLI unit tests

emrahs:
### Describe what has changed in this PR
Fixed broken unit tests

### DIFF
--- a/tools/cli/adminClusterCommands.go
+++ b/tools/cli/adminClusterCommands.go
@@ -29,6 +29,9 @@ import (
 	"github.com/uber/cadence/common/types"
 )
 
+// An indirection for the prompt function so that it can be mocked in the unit tests
+var promptFn = prompt
+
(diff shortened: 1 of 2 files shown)


tools/cli/adminClusterCommands.go
```diff
@@ -29,6 +29,9 @@ import (
 	"github.com/uber/cadence/common/types"
 )
 
+// An indirection for the prompt function so that it can be mocked in the unit tests
+var promptFn = prompt
+
```
Groxx, line 34: :| 

I mean, it works, but I/O is a dependency like anything else, we should just inject it somehow.  Though I'm not familiar with how much of a change that'd be to this code.

Groxx (approved):
LGTM though I am kinda curious how these weren't flagged earlier.  do we run these in CI?

demirkayaender (approved):
Thanks for fixing these tests!

PR #4092, opened 2021-03-30
Fix flaky integration test

emrahs:
**What changed?**
Attempt to fix flaky integration test that seems to be failing due to time precisions affecting timestamp math. Also added some extra logs to help with the investigation if the test continues being flaky.

**Why?**

**How did you test it?**

**Potential risks**

### DIFF
--- a/host/integration_test.go
+++ b/host/integration_test.go
@@ -1208,15 +1208,26 @@ func (s *integrationSuite) TestCronWorkflow() {
 	}
 	for i := 1; i != 4; i++ {
 		executionInfo := closedExecutions[i]
-		// Roundup to compare on the precision of seconds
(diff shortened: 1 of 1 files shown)


demirkayaender (approved):
This is actually the test I have been looking today for the delay start integration test, thanks for improving the output for better debugging. 

Looks like Zijian is refactoring the integration tests. You might want to wait until that change goes in for better consistency.

Shaddoll:
> This is actually the test I have been looking today for the delay start integration test, thanks for improving the output for better debugging.
> 
> Looks like Zijian is refactoring the integration tests. You might want to wait until that change goes in for better consistency.

I haven't done the refactoring yet. This is not a big change. I can rebase this change.

yux0 (approved):
Thank you for fixing this

PR #4098, opened 2021-03-31
Seperate integration test suite definition from actual integration tests

Shaddoll:
**What changed?**
Seperate integration test suite definition from actual integration tests

**Why?**
Allow these integration test suites to be reused.

**How did you test it?**
integration tests

**Potential risks**

**Release notes**

**Documentation Changes**


host/sizelimitTest.go
```diff
@@ -0,0 +1,191 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
yux0, line 1: Looks good to me. Could we have the updated header for all touched files?

yux0 (approved):

host/sizelimitTest.go
Shaddoll: Of course, I think the linter forces us to do it.

PR #4102, opened 2021-04-01
Fix don't swallow the error when building zap logger

git-hulk:
**What changed?**
Fix don't swallow the error when building zap logger

**Why?**

Currently, start the cadence server would be panic when
failed to build the zap logger. It will cause confusion since
the error was swallowed.

**How did you test it?**

test local with no permission log dir

**Potential risks**
no


tools/cli/domainUtils.go
```diff
@@ -299,7 +299,8 @@ func initializeDomainHandler(
 func initializeLogger(
 	serviceConfig *config.Config,
 ) log.Logger {
-	return loggerimpl.NewLogger(serviceConfig.Log.NewZapLogger())
+	zapLogger, _ := serviceConfig.Log.NewZapLogger()
```
meiliang86, line 302: log fatal or panic here in case of error?

meiliang86 (approved):

tools/cli/domainUtils.go
meiliang86: I guess you don't have a logger, so the only thing you can do is to panic

git-hulk: Yeah, I have used the `ErrorAndExited` to log the error and exit

longquanzheng (approved):

bench/lib/context.go
```diff
@@ -54,7 +54,7 @@ type RuntimeContext struct {
 
 // NewRuntimeContext builds a runtime context from the config
 func NewRuntimeContext(cfg *Config) (*RuntimeContext, error) {
-	logger := cfg.Log.NewZapLogger()
+	logger, _ := cfg.Log.NewZapLogger()
```
longquanzheng, line 57: We should just panic here. having error in the result means the logger could be invalid.

git-hulk: fixed

Groxx:
Huh.  Why don't we have this passed in?  Constructing a logger seems doomed to be over-complicated (as we expose more and more logger-configuration fields) and less flexible (zap loggers can do effectively anything, there's no reason for us to limit them).

Accepting a logger would also simplify the `log, _ := config.NewZapLogger()` case in tests: pass in a `zaptest.NewLogger(t)` instead of constructing one and trying to work around the test environment.

longquanzheng:
> Huh. Why don't we have this passed in? Constructing a logger seems doomed to be over-complicated (as we expose more and more logger-configuration fields) and less flexible (zap loggers can do effectively anything, there's no reason for us to limit them).
> 
> Accepting a logger would also simplify the `log, _ := config.NewZapLogger()` case in tests: pass in a `zaptest.NewLogger(t)` instead of constructing one and trying to work around the test environment.

Which one are you referring to? I think we have to do this way because they are in from different `main()` : server, bench, canary and CLI. So we can't pass a logger from server to cli code, hehe. 
Or in other words, we already try the best to pass around/initiate only once.

longquanzheng:
@Groxx I merged this PR but if you see a way that we can improve by passing around logger, we make do a refactoring

PR #4106, opened 2021-04-03
Expose server version tag to be set during build time

longquanzheng:
**What changed?**
Expose server version tag to be set during build time

**Why?**
for https://github.com/uber/cadence/issues/4052
following https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

**How did you test it?**
locally tested:
```
(qlong_version_tag)$./cadence-server
NAME:
   cadence - Cadence server

USAGE:
   cadence-server [global options] command [command options] [arguments...]

VERSION:
   test-6c9b260b
```

**Potential risks**
No

**Release notes**
No

**Documentation Changes**
No


meiliang86 (approved):

meiliang86:
Is it possible to include the first part of git sha as part of the version as well?
So it will look like "development-XXXXXX" and "0.19.1-XXXXXX"

longquanzheng:
> Is it possible to include the first part of git sha as part of the version as well?
> So it will look like "development-XXXXXX" and "0.19.1-XXXXXX"

It's not usual to do that way. How about having another field `commit`?

scripts/go-build-ldflags.sh
```diff
@@ -13,13 +13,15 @@ export GIT_VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo unknown)
...
+SERVER_VERSION="${SERVER_VERSION:-'development'}"
 
 if [ "$MODE" = "LDFLAG" ]; then
   LD_FLAGS="-X ${BASE_PACKAGE}.Revision=${GIT_REVISION} \
   -X ${BASE_PACKAGE}.Branch=${GIT_BRANCH}               \
   -X ${BASE_PACKAGE}.Version=${GIT_VERSION}             \
   -X ${BASE_PACKAGE}.BuildDate=${BUILD_DATE}            \
-  -X ${BASE_PACKAGE}.BuildTimeUnix=${BUILD_TS_UNIX}"
+  -X ${BASE_PACKAGE}.BuildTimeUnix=${BUILD_TS_UNIX}    \
+  -X main.Version=${SERVER_VERSION}"
```
Groxx, line 24: - @meiliang86 we have revision already in vars here, so no need to add it to the version.  though adding it to the output may be nice, yeah.
- @longquanzheng is SERVER_VERSION noticeably different than GIT_VERSION?  and why not put this in BASE_PACKAGE as well?

longquanzheng: good point. I didn't pay much attention. doing it.

longquanzheng: @meiliang86 okay I ended up append the revision  in the version output like you said, because there is no easy way to add such a field. 

@Groxx again, thanks for your comment.

longquanzheng:
@Groxx LMK if you have any other comments

scripts/go-build-ldflags.sh
```diff
@@ -9,10 +9,17 @@ fi
...
 export BUILD_TS_UNIX=$(date '+%s') # second since epoch
 export BASE_PACKAGE=github.com/uber/cadence/common/metrics
+if [ -z ${SERVER_VERSION} ]; then
+  # If not set SERVER_VERSION, then use the most recent tag.
+  export GIT_VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo unknown)
+else
+  # If passing a version explicitly, then use it
+  export GIT_VERSION=${SERVER_VERSION}
+fi
+
```
Groxx, line 22: Two Qs / options:
1. Maybe include `--dirty` so you see if the git status is clean?  I've found a fair number of flawed build systems this way, since it'll tell you if you're un-knowingly modifying things in your releases.
2. If ya want to be fancy:
    ```bash
    export GIT_VERSION="${SERVER_VERSION:-$(git describe --tags --abbrev=0 2>/dev/null || echo unknown)}"
    ```
    it'll only run the `$(...)` operation if SERVER_VERSION is not set.  Arguably this could be done for all the vars, though I mostly doubt we'll ever want to customize the others.
    
    (I generally avoid fancy bash-isms, but some are simple / clear / common enough to be useful)

But this is good as-is 👍 Same as before, and unambiguous.

Groxx (approved):
Works for me 👍  Check comment if you agree, but either way I think it's good to go.

scripts/go-build-ldflags.sh
longquanzheng: adopted 1 but reserve for 2 -- probably I am not very good at bash and feel like less fancyness will be more readable for people like me :P 
(really wish some we could write all python for all script).

Groxx: yep, more-explicit bash is rarely the wrong choice I think :)  easier to read and modify later.

PR #4108, opened 2021-04-03
Fix docker build with empty idls

longquanzheng:
**What changed?**
Fix docker build with empty idls

**Why?**
for https://github.com/uber/cadence/issues/4105

**How did you test it?**
local test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 	touch $(BUILD)/protoc
 
 .fake-thrift: | $(BIN) $(BUILD)
-	git submodule update --init --recursive
+	$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)
 	touch $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc
 	touch $(THRIFT_GEN)
 
--- a/Makefile
+++ b/Makefile
@@ -251,6 +251,7 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 	touch $(BUILD)/protoc
 
 .fake-thrift: | $(BIN) $(BUILD)
+	git submodule update --init --recursive
 	touch $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc
 	touch $(THRIFT_GEN)
 
--- a/Makefile
+++ b/Makefile
@@ -251,9 +251,8 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 	touch $(BUILD)/protoc
 
 .fake-thrift: | $(BIN) $(BUILD)
-	$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)
 	touch $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc
-	touch $(THRIFT_GEN)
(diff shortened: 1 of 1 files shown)


Groxx (changes requested):
no, this belongs somewhere else.  it's a totally separate 'thing' to do, it needs its own target

Groxx:
Is this just for CI builds?  we can / probably should just change it to `git clone --recursive` instead

Groxx:
if it's just for empty `THRIFT_GEN` (which is probably fine to leave), `$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)` would likely work as well, and avoids the clone entirely

longquanzheng:
> Is this just for CI builds? we can / probably should just change it to `git clone --recursive` instead

I ran into this when I tried to manually build docker image for testing.
But actually, this is also breaking out CI build -- out master images hasn't been updated for a while.

longquanzheng:
> if it's just for empty `THRIFT_GEN` (which is probably fine to leave), `$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)` would likely work as well, and avoids the clone entirely

That works! thanks! Updated my PR.

longquanzheng:
> if it's just for empty `THRIFT_GEN` (which is probably fine to leave), `$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)` would likely work as well, and avoids the clone entirely

Tested that works. Thanks. Updated my PR. 
btw what is the purpose of that then? if we could do noop of it, why we need it?

longquanzheng:
> at the end, similar reasons as .fake-protoc: order matters. and you'll need to _replace_ the `touch $(THRIFT_GEN)` or it'll fail in the same way, because it's still executing `touch [nothing]`

No, it doesn't fail. Looks like they are  different from the Makefile

Groxx:
> > at the end, similar reasons as .fake-protoc: order matters. and you'll need to _replace_ the `touch $(THRIFT_GEN)` or it'll fail in the same way, because it's still executing `touch [nothing]`
> 
> No, it doesn't fail. Looks like they are different from the Makefile

see if you have an idls folder locally - you probably created it with your earlier changes.

anyway.  what it is is a small number of lines above: it's a marker for "have we generated thrift file X", for each file.
- when we're generating code, those files have a last-modified date.  if it's newer than the generated source files based on them, thrift-gen is re-done.  so touching them updates their last-modified to 'now', and later things are updated or ignored based on that.
- when we're _not_ generating code, they may not exist, so it's an empty variable.  so `$(if [empty],touch [nothing],)` takes the "else" branch and does nothing, because it'd be an error anyway.
- last but not least: since the var is empty, it's not anything that can be depended on, so even if it's in a dependency list somewhere, it does nothing.

longquanzheng:
@Groxx yes there is an empty `idls/`. But it's added since the checkout. 
I also verified by cloning a new repo:
```
(master)$ls
CHANGELOG.md  LICENSE	PROPOSALS.md  RELEASES.md  canary  cmd	   config  docs		go.mod	host  proto	   schema   service
Dockerfile    Makefile	README.md     bench	   client  common  docker  environment	go.sum	idls  revive.toml  scripts  tools
qlong@~/tmp/tmp2/cadence:
(master)$
qlong@~/tmp/tmp2/cadence:
(master)$ls idls/
qlong@~/tmp/tmp2/cadence:
(master)$make .fake-protoc
touch .bin/protoc-3.14.0 .bin/protoc-gen-gogofast .bin/protoc-gen-yarpc-go
touch .build/protoc
```

Groxx:
I just tried it locally too.  I can reproduce + this fixes it, so 👍 

if you'd like, a `# comment` after that `$(..)` will work, e.g. `# maybe ignoring empty idl folder`.  it'll be printed since the line isn't prefixed with `@`, but that's probably a good thing (makes it clear if/when something happens in CI)

Groxx (approved):

PR #4116, opened 2021-04-06
Fix flaky test

demirkayaender:
**What changed?**
There's a subtle race condition for this test causing it to be flaky, added some buffer to expiration timestamp comparison to make the test more stable.

**Why?**
TestCreateHistoryStartWorkflowRequest_DelayStart tests if setting DelayStart param will add to the expiration time. We set the start time before we create the workflow; then create the workflow, then take the expiration time, then compare against the start time and expect the difference to be at at least the value of DelayStartSeconds. However, since we take the expirationTime after the we create the workflow, that difference is not guaranteed due to test thread sleeping or workflow creation taking a bit more time than anticipated.

With this change, I'm adding some buffer to avoid this test to be flaky and block people's changes.

**How did you test it?**
`go test -v github.com/uber/cadence/common`

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -120,8 +120,25 @@ func TestCreateHistoryStartWorkflowRequest_DelayStart(t *testing.T) {
 
 	expirationTime := startRequest.GetExpirationTimestamp()
 	require.NotNil(t, expirationTime)
-	require.True(t, time.Unix(0, expirationTime).Sub(now) > (100+60)*time.Second)
-	require.True(t, time.Unix(0, expirationTime).Sub(now) < (100+65)*time.Second)
+
+	// Since we assign the expiration time after we create the workflow request,
+	// There's a chance that the test thread might sleep or get deprioritized and
+	// expirationTime - now may not be equal to DelayStartSeconds. Adding 2 seconds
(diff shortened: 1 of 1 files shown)


common/util_test.go
```diff
@@ -120,8 +120,20 @@ func TestCreateHistoryStartWorkflowRequest_DelayStart(t *testing.T) {
 
 	expirationTime := startRequest.GetExpirationTimestamp()
 	require.NotNil(t, expirationTime)
-	require.True(t, time.Unix(0, expirationTime).Sub(now) > (100+60)*time.Second)
-	require.True(t, time.Unix(0, expirationTime).Sub(now) < (100+65)*time.Second)
+	require.True(
```
meiliang86, line 123: Add the same explanation as comments so other people understand what's going on here?

meiliang86 (approved):

Groxx:
For future reference: there's a `require.InDelta` that effectively does these two tests.

Though in this case: seems like `time.Unix(0, expirationTime).Sub(now)` should never be _below_ a precise minimum of `(100+60)*time.Second`, regardless of thread delays, since expirationTime can only grow larger.

PR #4118, opened 2021-04-07
Refactor config: move config and dynamiconfig one level up

longquanzheng:
**What changed?**
Refactor config: move config and dynamiconfig one level up

**Why?**
I have been sick with this layout for a very long time. This config and dynamicconfig is one of the most frequent lookup but they are hidden very deep. Moreover, I don't know why they have to be under `service/`. 

So of course services use the configs...but so are all other packages. 

**How did you test it?**
existing test.
Mostly are done by IDE, I only need to fix something that IDE missed. 

**Potential risks**
No. 

**Release notes**

**Documentation Changes**


yycptt (approved):

yycptt (commented):
Is the path for dynamic config updated in the config file?

yycptt:
> Is the path for dynamic config updated in the config file?

nvm, I confused the import path with the actual config file path. Change LGTM!

PR #4120, opened 2021-04-07
Allow setting levelKey for logging level

IgnacioGonzalezSomoza:
**What changed?**
levelkey is hardcode now. This PR will allow to set it from yaml config file, giving de change to set it to "Level" (actual value), "severity" (in order to ingest it from cloud logging) or other value.

**Why?**

**How did you test it?** It's a small test, tested locally

**Potential risks**

**Release notes**

**Documentation Changes**


cmd/server/server.go
```diff
@@ -106,6 +106,9 @@ func (s *server) startService() common.Daemon {
 
 	params := service.BootstrapParams{}
 	params.Name = "cadence-" + s.name
+	if len(s.cfg.Log.LevelKey) == 0 {
+		s.cfg.Log.LevelKey = "level"
+	}
```
longquanzheng, line 111: Can you move this code to the `NewZapLogger`? 

also use `if s.cfg.Log.LevelKey == "" {` as our convention

common/service/config/config.go
```diff
@@ -232,6 +232,8 @@ type (
 		Level string `yaml:"level"`
 		// OutputFile is the path to the log output file
 		OutputFile string `yaml:"outputFile"`
+		// levelKey is the desired log level
```
longquanzheng, line 235: how about: levelKey is the desired log level, defaults to "level"

common/service/config/log.go
```diff
@@ -38,7 +38,7 @@ const fileMode = os.FileMode(0644)
 func (cfg *Logger) NewZapLogger() *zap.Logger {
 	encodeConfig := zapcore.EncoderConfig{
```
longquanzheng, line 39: like I said above, set the default value here:
```
levelKey := s.cfg.Log.LevelKey
if levelKey == ""{
    levelKey = "level"
}
```

longquanzheng (changes requested):
Thanks for your contribution

longquanzheng:
there is some conflict with cmd/server/server.go. But if you address the comment the conflict should go away

common/service/config/config.go
```diff
@@ -232,6 +232,8 @@ type (
...
+		LevelKey string `yaml:"levelKey"`
```
mkolodezny, line 236: maybe:
` LevelKey string `default:"level" yaml:"levelKey"``

longquanzheng: I have tried that before, it doesn't work with the yaml library that we are using. (otherwise we have done this for lots of other fields)

LMK if I am wrong.

cmd/server/server.go
IgnacioGonzalezSomoza: done

common/service/config/config.go
IgnacioGonzalezSomoza: done

common/service/config/log.go
IgnacioGonzalezSomoza: done

common/service/config/config.go
IgnacioGonzalezSomoza: I've tried with --> LevelKey string `default:"mylevel" yaml:"levelKey"`
but it doesn't work for me either

longquanzheng:
@IgnacioGonzalezSomoza can you also rebase and resolve the conflict?

longquanzheng (approved):

PR #4123, opened 2021-04-09
Support workflow already completed error

demirkayaender:
**What changed?**
When a workflow is completed, signals were returning EntityNotExistsError
With this diff we will start returning WorkflowExecutionAlreadyCompletedError
which is more descriptive to the user

See https://t3.uberinternal.com/browse/CDNC-615 for more details

**NOTE**: when releasing, these changes need to go with the client changes here (https://github.com/uber-go/cadence-client/pull/1081) If these changes are deployed before the client changes; instead of returning error, CLI shows signal being succeeded.

**Why?**
This change helps differentiate non existing workflows vs completed workflows.

**How did you test it?**
*Unit tests:*
`go test -v github.com/uber/cadence/service/history -run TestEngineSuite -testify.m TestSignalWorkflowExecution_Failed`
`go test -v github.com/uber/cadence/service/history -run engine2Suite -testify.m TestRequestCancelWorkflowExecutionFail`

*Using CLI*: Used helloWorld sample, finished a workflow instance then sent a signal to the workflow. 
```
enderd[04/08 21:22]:~/cadence$ ./cadence --do samples-domain wf signal --wid helloworld_7b7e9bfe-47d9-4242-9f30-96051827bb4f --name signal_test
Error: Signal workflow failed.
Error Details: WorkflowExecutionAlreadyCompletedError{Message: workflow execution already completed}
('export CADENCE_CLI_SHOW_STACKS=1' to see stack traces)
```

**Potential risks**

**Release notes**

**Documentation Changes**


demirkayaender:
Failures are due to the missing client repo changes. This PR fixes it https://github.com/uber-go/cadence-client/pull/1081

common/archiver/s3store/historyArchiver.go
```diff
@@ -303,7 +303,7 @@ func (h *historyArchiver) Get(
 				return nil, &types.InternalServiceError{Message: err.Error()}
 			}
 			switch err.(type) {
-			case *types.BadRequestError, *types.InternalServiceError, *types.EntityNotExistsError:
+			case *types.BadRequestError, *types.InternalServiceError, *types.EntityNotExistsError, *types.WorkflowExecutionAlreadyCompletedError:
```
yycptt, line 306: I don't think already completed error will be returned here by the download func?

common/metrics/defs.go
```diff
@@ -1712,6 +1712,8 @@ const (
 	ParentClosePolicyProcessorSuccess
 	ParentClosePolicyProcessorFailures
 
+	CadenceErrWorkflowExecutionAlreadyCompletedCounter
```
yycptt, line 1715: nit: group with other `CadenceErrXXXCounter` metrics?

service/history/task/transfer_active_task_executor.go
```diff
@@ -925,7 +925,7 @@ func (t *transferActiveTaskExecutor) recordChildExecutionStarted(
 	return t.updateWorkflowExecution(ctx, wfContext, true,
 		func(ctx context.Context, mutableState execution.MutableState) error {
 			if !mutableState.IsWorkflowExecutionRunning() {
-				return &types.EntityNotExistsError{Message: "Workflow execution already completed."}
+				return &types.WorkflowExecutionAlreadyCompletedError{Message: "Workflow execution already completed."}
```
yycptt, line 928: Also need to update the error handling logic in `task.go` if the returned error is changed I think.

service/history/task/transfer_active_task_executor.go
```diff
@@ -1054,7 +1054,7 @@ func (t *transferActiveTaskExecutor) signalExternalExecutionCompleted(
 	err := t.updateWorkflowExecution(ctx, wfContext, true,
 		func(ctx context.Context, mutableState execution.MutableState) error {
 			if !mutableState.IsWorkflowExecutionRunning() {
-				return &types.EntityNotExistsError{Message: "Workflow execution already completed."}
+				return &types.WorkflowExecutionAlreadyCompletedError{Message: "Workflow execution already completed."}
```
yycptt, line 1057: Do we need to update the code on L1076? Same for other changes.

service/history/task/transfer_active_task_executor.go
```diff
@@ -1015,7 +1015,7 @@ func (t *transferActiveTaskExecutor) requestCancelExternalExecutionCompleted(
...
```
yycptt, line 1018: Not related to change on this line.  Just want to make sure that: is the error handling logic able to handle the new error type returned from `historyClient` calls in this file?

common/metrics/defs.go
demirkayaender: I thought about that but wasn't sure how the enums were stored in the metrics; putting this up there might change the enum values of the other counters and metrics might show up wrong in the dashboards. Do you know if it's okay to put new enums in between the other ones?

service/history/task/transfer_active_task_executor.go
demirkayaender: > is the error handling logic able to handle the new error type 

Could you explain this a bit more? How can I check this?

common/metrics/defs.go
yycptt: yes, it's ok. Those enums will be mapped to strings.

service/history/task/transfer_active_task_executor.go
yycptt: Probably find all calls to history client in this file and see if the call might return already completed error or not. If so, maybe check if it is correctly treated as non-retryable and the same way as the entity not exists error returned previously.

demirkayaender: As far as this file goes, there was only one place to be handled in the caller to `applyParentClosePolicy`. I fixed this one.

We return this error also from `requestCancelExternalExecutionWithRetry` and `signalExternalExecutionWithRetry` which forward the error to their external callers from `Execute`. And those seem to be handled in task.go:HandleErr which I also changed (you'll see in the updated PR).

I couldn't find other places but let me know if there are other places if you can think of other places.

service/history/task/transfer_active_task_executor.go
demirkayaender: Good catch! I didn't realize we have special treatments for this error in this file.

demirkayaender (commented):
Will update the pr shortly.

service/history/task/transfer_active_task_executor.go
yycptt: Probably also update the logic in `createFirstDecisionTask`? It's ok if it's not updated I think as the logic in `task.go` will handle it, but to avoid confusion, probably also update that?

service/history/workflow/errors.go
```diff
@@ -34,7 +34,7 @@ var (
 	// ErrActivityTaskNotFound is the error to indicate activity task could be duplicate and activity already completed
 	ErrActivityTaskNotFound = &types.EntityNotExistsError{Message: "activity task not found"}
 	// ErrAlreadyCompleted is the error to indicate workflow execution already completed
-	ErrAlreadyCompleted = &types.EntityNotExistsError{Message: "workflow execution already completed"}
+	ErrAlreadyCompleted = &types.WorkflowExecutionAlreadyCompletedError{Message: "workflow execution already completed"}
```
yycptt, line 37: This error is used by many other APIs (for example RecordChildExecutionCompleted) besides the three that have their IDL definition updated. We will need to add already completed error to the IDL definition for all those APIs I think.

demirkayaender: Yeah I realized this while looking at the integration test failures. Right now the differentiation is based on if the request was signal/terminate/cancel type or not. I will probably update this diff with the idl repo later. Let me know if you have any other ideas. I updated the pr to fix the integration tests though.

go.mod
```diff
@@ -58,7 +58,7 @@ require (
 	github.com/xwb1989/sqlparser v0.0.0-20180606152119-120387863bf2
 	go.opencensus.io v0.22.5 // indirect
 	go.uber.org/atomic v1.5.1
-	go.uber.org/cadence v0.16.1-0.20210330222409-fcc703e94e2f
+	go.uber.org/cadence v0.16.1-0.20210409233821-65903f75c15d
```
demirkayaender, line 61: I have these lines even after `go mod tidy`. Should I do manually revert them?

common/client/versionChecker.go
```diff
@@ -53,14 +53,14 @@ const (
...
 	// JavaWorkerRawHistoryQueryVersion indicates the minimum client version of the java worker which supports RawHistoryQuery
-	JavaWorkerRawHistoryQueryVersion = "1.3.0"
+	JavaWorkerRawHistoryQueryVersion = "1.4.0"
 	// GoWorkerRawHistoryQueryVersion indicates the minimum client version of the go worker which supports RawHistoryQuery
-	GoWorkerRawHistoryQueryVersion = "1.6.0"
+	GoWorkerRawHistoryQueryVersion = "1.7.0"
 	// CLIRawHistoryQueryVersion indicates the minimum CLI version of the go worker which supports RawHistoryQuery
 	// Note: cli uses go client feature version
-	CLIRawHistoryQueryVersion = "1.6.0"
+	CLIRawHistoryQueryVersion = "1.7.0"
```
yycptt, line 63: Those versions should not be updated. They specify versions those features become available.

common/client/versionChecker.go
```diff
@@ -40,11 +40,11 @@ const (
 	CLI = "cli"
 
 	// SupportedGoSDKVersion indicates the highest go sdk version server will accept requests from
-	SupportedGoSDKVersion = "1.6.0"
+	SupportedGoSDKVersion = "1.7.0"
 	// SupportedJavaSDKVersion indicates the highest java sdk version server will accept requests from
-	SupportedJavaSDKVersion = "1.6.0"
+	SupportedJavaSDKVersion = "1.7.0"
```
yycptt: This should actually be 1.4.0 as specified in the java client. The previously version was not correct. I have already updated the version and landed to master, you can simply merge my changes.

common/util.go
```diff
@@ -35,6 +35,8 @@ import (
 	"github.com/pborman/uuid"
 	"go.uber.org/yarpc/yarpcerrors"
```
yycptt, line 37: nit: remove empty line

service/history/workflow/util.go
```diff
@@ -37,7 +37,7 @@ type (
 		CreateDecision bool
 	}
 
-	UpdateActionFunc func(execution.Context, execution.MutableState) (*UpdateAction, error)
+	UpdateActionFunc func(context.Context, execution.Context, execution.MutableState) (*UpdateAction, error)
```
yycptt, line 40: what's the purpose of this new parameter?

service/history/task/transfer_active_task_executor.go
```diff
@@ -1134,7 +1153,12 @@ func (t *transferActiveTaskExecutor) signalExternalExecutionFailed(
 	err := t.updateWorkflowExecution(ctx, wfContext, true,
 		func(ctx context.Context, mutableState execution.MutableState) error {
 			if !mutableState.IsWorkflowExecutionRunning() {
-				return &types.EntityNotExistsError{Message: "Workflow is not running."}
+				return execution.VersionBasedError(
+					&types.WorkflowExecutionAlreadyCompletedError{Message: "Workflow execution already completed."},
+					"1.7.0",
+					yarpc.CallFromContext(ctx).Header(common.FeatureVersionHeaderName),
+					&types.EntityNotExistsError{Message: "Workflow execution already completed."},
+				)
```
yycptt, line 1161: My understanding it that all calls from history service will not go through client, so there's no version here. Same for other checks in the file.

service/history/historyEngine.go
```diff
@@ -2357,9 +2383,14 @@ func (e *historyEngineImpl) TerminateWorkflowExecution(
 		domainID,
 		workflowExecution,
 		e.timeSource.Now(),
-		func(wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
+		func(ctx context.Context, wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
```
yycptt, line 2386: we can use `ctx`  directly in the func without adding it as a parameter.

service/history/historyEngine.go
```diff
@@ -2069,9 +2080,14 @@ func (e *historyEngineImpl) RequestCancelWorkflowExecution(
...
 	return workflow.UpdateCurrentWithActionFunc(ctx, e.executionCache, e.executionManager, domainID, workflowExecution, e.timeSource.Now(),
-		func(wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
+		func(ctx context.Context, wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
 			if !mutableState.IsWorkflowExecutionRunning() {
-				return nil, workflow.ErrAlreadyCompleted
+				return nil, execution.VersionBasedError(
+					workflow.ErrAlreadyCompleted,
+					"1.7.0",
+					yarpc.CallFromContext(ctx).Header(common.FeatureVersionHeaderName),
+					workflow.ErrNotExists,
```
yycptt, line 2089: I am not entirely sure about this, does the call from frontend to history contains client version in the header?

service/history/historyEngine.go
```diff
@@ -2357,9 +2383,14 @@ func (e *historyEngineImpl) TerminateWorkflowExecution(
 		domainID,
 		workflowExecution,
 		e.timeSource.Now(),
-		func(wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
+		func(ctx context.Context, wfContext execution.Context, mutableState execution.MutableState) (*workflow.UpdateAction, error) {
 			if !mutableState.IsWorkflowExecutionRunning() {
-				return nil, workflow.ErrAlreadyCompleted
+				return nil, execution.VersionBasedError(
+					workflow.ErrAlreadyCompleted,
+					"1.7.0",
```
yycptt, line 2390: let's define this value in `versionChecker.go`.

yycptt: I am still thinking maybe we should do this check in frontend instead of history so that only frontend interacts with the concept of client version. Then we can add a new method to versionChecker for checking the support of already complete error at client side. 

In terms of implementation, one approach might be, history always return the newly defined type (just in 3 APIs for now), and frontend can convert it back to entity not exists error if necessary.

demirkayaender: That sounds like a good idea. Do we know which client is calling the service? How do I know if the client is java or go client?

common/client/versionChecker_test.go
```diff
@@ -339,6 +344,89 @@ func (s *VersionCheckerSuite) TestSupportsRawHistoryQuery() {
...
+		{
+			clientImpl:           JavaSDK,
+			clientFeatureVersion: "1.4.0",
+			expectErr:            false,
+		},
+		{
+			clientImpl:           JavaSDK,
+			clientFeatureVersion: "1.4.0",
+			expectErr:            false,
+		},
```
yycptt, line 380: nit: remove the duplicated case?

go.mod
yycptt: nit: maybe update to the latest client commit?

service/frontend/workflowHandler.go
```diff
@@ -4027,3 +4033,23 @@ func checkRequiredDomainDataKVs(requiredDomainDataKeys map[string]interface{}, d
 	}
 	return nil
 }
+
+// Some error types are introduced later that some clients might not support
+// To make them backward compatible, we continue returning the legacy error types
+// for older clients
+func (wh *WorkflowHandler) NormalizeVersionedErrors(ctx context.Context, err error) error {
```
yycptt, line 4040: nit: it seems like we don't need to export this method?

service/history/workflow/util.go
```diff
@@ -186,11 +186,12 @@ func UpdateWithAction(
 		domainID,
 		execution,
 		now,
-		getUpdateActionFunc(createDecisionTask, action),
+		getUpdateActionFunc(ctx, createDecisionTask, action),
 	)
 }
 
 func getUpdateActionFunc(
+	ctx context.Context,
```
yycptt, line 194: looks like we can revert this change.

service/history/historyEngine.go
```diff
@@ -691,6 +691,11 @@ func (e *historyEngineImpl) startWorkflowHelper(
 				startRequest,
 				nil,
 			)
+			// By the time we try to terminate the workflow, it was already terminated
+			// So continue as if we didn't need to terminate it in the first place
+			if _, ok := err.(*types.WorkflowExecutionAlreadyCompletedError); !ok {
+				return resp, err
```
yycptt, line 697: @yux0 Can you double check the logic here? Looks good to me, but if possible please double check.

service/frontend/workflowHandler.go
```diff
@@ -2490,6 +2492,7 @@ func (wh *WorkflowHandler) RequestCancelWorkflowExecution(
 		CancelRequest: cancelRequest,
 	})
 	if err != nil {
+		err = wh.NormalizeVersionedErrors(ctx, err)
```
yux0, line 2495: Could we do normalize after the wh.error so that we have the correct error metrics?

service/history/historyEngine.go
yux0: Technically, this is correct. However, we acquire locking before shouldTerminateAndStart. So the workflow state should not be changed. Maybe we can add a log meesage if the AlreadyCompletedError encounters.

demirkayaender: Added a warning log, let me know if this looks good to you

PR #4126, opened 2021-04-09
Update domain deprecation command to fail if workflow exists

Shaddoll:
**What changed?**
Update the domain deprecation command to check if there exists any workflow history in that domain before deprecating the domain by default. And also add a new flag '--force' to allow user to skip the workflow history check.

**Why?**
We want to be cautious when deprecating a domain, so the default behavior is not to do anything if workflow history exists. We also want to keep the original behavior to deprecate a domain when we don't care workflow history in that domain.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/domainCommands.go
+++ b/tools/cli/domainCommands.go
@@ -285,10 +285,24 @@ func (d *domainCLIImpl) UpdateDomain(c *cli.Context) {
 func (d *domainCLIImpl) DeprecateDomain(c *cli.Context) {
 	domainName := getRequiredGlobalOption(c, FlagDomain)
 	securityToken := c.String(FlagSecurityToken)
+	force := c.Bool(FlagForce)
 
 	ctx, cancel := newContext(c)
 	defer cancel()
 
+	if !force {
+	// check if there is any workflow in this domain, if exists, do not deprecate
+		wfs, _ := listClosedWorkflow(getWorkflowClient(c), 1, 0, time.Now().UnixNano(), "", "", workflowStatusNotSet, nil, c)
+		if len(wfs) > 0 {
+			ErrorAndExit("Operation DeprecateDomain failed.", errors.New("Workflow history not cleared in this domain."))
+			return
+		}
+		wfs, _ = listOpenWorkflow(getWorkflowClient(c), 1, 0, time.Now().UnixNano(), "", "", nil, c)
+		if len(wfs) > 0 {
+			ErrorAndExit("Operation DeprecateDomain failed.", errors.New("Workflow still running in this domain."))
+			return
+		}
+	}
 	err := d.deprecateDomain(ctx, &types.DeprecateDomainRequest{
 		Name:          domainName,
 		SecurityToken: securityToken,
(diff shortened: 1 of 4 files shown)


tools/cli/domainCommands.go
```diff
@@ -288,6 +288,16 @@ func (d *domainCLIImpl) DeprecateDomain(c *cli.Context) {
 	ctx, cancel := newContext(c)
 	defer cancel()
 
+	// check if there is any workflow in this domain, if exists, do not deprecate
+	wfs, _ := listClosedWorkflow(getWorkflowClient(c), 1, 0, time.Now().UnixNano(), "", "", workflowStatusNotSet, nil, c)
+	if len(wfs) > 0 {
+		ErrorAndExit("Operation DeprecateDomain failed.", errors.New("Workflow history not cleared in this domain."))
```
longquanzheng, line 294: why closed workflows still matter? Is that a requirements or some technical workaround?

Shaddoll: it's a requirement. We want to be conservative.

yycptt: May I ask what's purpose of this change? I thought only start or signal workflow is blocked for deprecated domains. Trying to understand why do closed workflow and running workflow matter here?

If we want to enforce this requirement that domain deprecation only happens for domains in which all workflows have passed retention, I think this check should be done inside the handler of domain deprecation call instead of just in the CLI?

Shaddoll: The reason I made the change in CLI but not in server is to let the caller  do the check.
And within CLI, we can add another flag (e.g. --force) to make this check optional, but the default behavior is to be conservative and does the check.

Shaddoll: @meiliang86 What's your opinion on this? I prefer to adding the check in CLI and later we can add a flag to skip the check if we'd like to.

meiliang86: Yeah adding the check to the CLI is fine. Technically we can deprecate domains even there's open/close workflows, so it's ok to not check it on the server side.
This extra check is just to make sure we don't shoot ourselves in the foot (i.e. accidentally deprecating domains that are still in use), as right now we don't support resurrecting domains once deprecated.

tools/cli/domainCommands.go
```diff
@@ -284,10 +284,24 @@ func (d *domainCLIImpl) UpdateDomain(c *cli.Context) {
...
 	domainName := getRequiredGlobalOption(c, FlagDomain)
 	securityToken := c.String(FlagSecurityToken)
+	force := c.Bool(FlagForce)
 
 	ctx, cancel := newContext(c)
 	defer cancel()
 
+	if !force {
+	// check if there is any workflow in this domain, if exists, do not deprecate
+		wfs, _ := listClosedWorkflow(getWorkflowClient(c), 1, 0, time.Now().UnixNano(), "", "", workflowStatusNotSet, nil, c)
```
yycptt, line 294: do we need to check the error and block deprecation when failed to verify workflows?

Shaddoll: Yes, it's done within the helper functions.

yycptt: Oh I see. I thought the second return value is error. My bad.

yycptt (approved):

PR #4128, opened 2021-04-10
Config layout refactor: centralize config file for documentation

longquanzheng:
**What changed?**

**Why?**
1. After this configuration, we will have all config defined in the same package, so that our documentation can simply link a place: https://pkg.go.dev/github.com/uber/cadence@v0.19.1/common/service/config#Service

2. packages of auth + authorization are confusing. 

3. Remove the gocql.Client from configuration to make it more client, and avoid cyclular dependency

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**
Will update cadence-docs after this change is released: https://github.com/uber/cadence-docs/pull/32


common/persistence/nosql/nosqlplugin/cassandra/gocql/native/client.go
```diff
@@ -18,15 +18,56 @@
...
+	return err == gogocql.ErrNotFound
+}
+
+func (c client) IsThrottlingError(err error) bool {
+	if req, ok := err.(gogocql.RequestError); ok {
+		// gocql does not expose the constant errOverloaded = 0x1001
+		return req.Code() == 0x1001
+	}
+	return false
+}
```
yycptt, line 73: nit: new line

common/persistence/nosql/nosqlplugin/cassandra/gocql/session.go
```diff
@@ -48,7 +48,7 @@ type (
 	}
 )
 
-func newSession(
+func NewSession(
```
yycptt, line 51: why do we need to export this function? I think Client interface has a CreateSession method

tools/cassandra/handler.go
```diff
@@ -193,9 +191,6 @@ func validateCQLClientConfig(config *CQLClientConfig) error {
 	if config.numReplicas == 0 {
 		config.numReplicas = defaultNumReplicas
 	}
-	if config.CQLClient == nil {
-		config.CQLClient = defaultGoCQLClient
```
yycptt, line 197: looks like we can remove the file which defines `defaultGoCQLClient`?

common/persistence/persistence-tests/persistenceTestBase.go
```diff
@@ -45,6 +45,7 @@ import (
 	p "github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/persistence/cassandra"
 	"github.com/uber/cadence/common/persistence/client"
+	_ "github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra/gocql/native" // needed to load the default gocql client
```
yycptt, line 48: so it means we can't reuse the tests for testing our internal gocql client impl?

maybe we can put it in `cassandra_test.go` so that gocql client only get registered in the open source test?

tools/cli/adminCommands.go
```diff
@@ -298,7 +297,7 @@ func connectToCassandra(c *cli.Context) (gocql.Client, gocql.Session) {
 		Timeout:           10 * time.Second,
 	}
 	if c.Bool(FlagEnableTLS) {
-		clusterConfig.TLS = &auth.TLS{
+		clusterConfig.TLS = &config.TLS{
```
yycptt, line 300: Not related to this line. 

CLI gets gocql client through a client factory on L309. Can you also update that to use the registered gocql client?

common/persistence/nosql/nosqlplugin/cassandra/gocql/native/client.go
```diff
@@ -18,15 +18,56 @@
...
+var _ gocql.Client = client{}
+
+type (
+	client struct {
+	}
+)
+
+func init() {
+	gocql.RegisterClient(client{})
+}
```
yycptt, line 41: Personally I would prefer to call the RegisterClient directly so it's more clear we are doing and won't surprise people when they import this package or import a package that import this package. But no strong opinion on it.

longquanzheng: It's certainly a good concern. 
The reason behind this is that module variable is better to be immutable once being set. Allowing override (using Replace ) is dangerous can lead to undefined behavior. 

Because of the immutability, this client can only be set once. If not using import, we have to make sure not calling RegisterClient more than noce. 

I think we have to do it anyway. Looks like your internal repo needs to call this API for testing.  

I will change to explict calling pattern.

tools/cli/adminCommands.go
longquanzheng: It's already updated -- I updated the client factory to use the reigstered gocql client. LMK if I am wrong.

tools/cassandra/handler.go
longquanzheng: Yes, I deleted ` tools/cassandra/gocqlClient.go` but Github showns as moving file for some reasons.

common/persistence/nosql/nosqlplugin/cassandra/gocql/session.go
longquanzheng: because the native client needs to call it to create a session. LMK if I missed anything.

common/persistence/nosql/nosqlplugin/cassandra/gocql/native/client.go
longquanzheng: yeah for sure.

common/persistence/nosql/nosqlplugin/cassandra/gocql/native/client.go
longquanzheng: okay I still keep using `init`. Because it turns out that it's not possible to call only once in the test(which started by `go test ...` command). So if we want to call RegisterClient directly, we must allowing the registeredClient to be immutable.

However, I don't like the idea of adding "_ import ..." in every "..._test.go" files. Because it's error prone and people may forget about doing that. 

So I ended up making an special package "external" as a wrapper for NewTestBaseWithCassandra. This is the best idea I have right now for all the concerns.

tools/cli/adminCommands.go
yycptt: I see. But looks like we can remove this methods from client factory interface? internal cli will also do a client registration, so the internal implementation of this method will be the same as the open source version.

common/persistence/nosql/nosqlplugin/cassandra/gocql/public/testBase.go
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2017-2020 Uber Technologies, Inc.
+// Copyright (c) 2017-2021 Uber Technologies, Inc.
+// Portions of the Software are attributed to Copyright (c) 2020 Temporal Technologies Inc.
```
yycptt, line 2: oh, temporal has the same structure for those tests?

yycptt (approved):
LGTM overall, just some nit comments.

common/persistence/nosql/nosqlplugin/cassandra/gocql/public/testBase.go
longquanzheng: No I didn’t look at any temporal code for this refactoring. Moving the files is because of circle dependency.

yycptt: then we can remove the second line?

longquanzheng: ah I see. Yeah I copied it from somewhere else by mistake.

tools/cli/adminCommands.go
longquanzheng: yeah good catch. Removed.

PR #4129, opened 2021-04-11
Refactor dynamicconfig comments for documentation

longquanzheng:
**What changed?**
Refactor dynamicconfig comments

**Why?**
To introduce a better way to manage/maintain the [documentation of dynamic configuration](https://cadenceworkflow.io/docs/operation-guide/setup/#dynamic-configuration-overview) for the community

Currently it requires developer to go to cadence-docs to update the configuration. 
This is tedious and inconvenient for our developers. Moreover, there is no easy way to know what configurations is available in which release versions. 

A better way is just update in the comments, and go-docs will automatically update the document with the right release version. See example: https://pkg.go.dev/github.com/uber/cadence@v0.19.1/common/service/dynamicconfig#Key

**How did you test it?**
Existing

**Potential risks**
No. only comment change.

**Release notes**

**Documentation Changes**
Will update cadence-docs to use go-docs after this is released. 
https://github.com/uber/cadence-docs/issues/33

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -699,6 +699,10 @@ const (
 	// KeyName: history.transferProcessorMaxRedispatchQueueSize
 	// Default value: 10000
 	TransferProcessorMaxRedispatchQueueSize
+	// TransferProcessorEnableValidator is whether validator should be enabled for transferQueueProcessor
+	// KeyName: history.transferProcessorEnableValidator
+	// Default value: FALSE
+	TransferProcessorEnableValidator
 	// TransferProcessorVisibilityArchivalTimeLimit is the upper time limit for archiving visibility records
 	// KeyName: history.transferProcessorVisibilityArchivalTimeLimit
 	// Default value: 200*time.Millisecond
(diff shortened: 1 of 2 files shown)


common/dynamicconfig/constants.go
```diff
@@ -31,937 +31,1437 @@ func (k Key) String() string {
...
 	// TransferProcessorCompleteTransferInterval is complete timer interval for transferQueueProcessor
+	// KeyName: history.transferProcessorCompleteTransferInterval
+	// Default value: 60*time.Second
 	TransferProcessorCompleteTransferInterval
 	// TransferProcessorMaxRedispatchQueueSize is the threshold of the number of tasks in the redispatch queue for transferQueueProcessor
+	// KeyName: history.transferProcessorMaxRedispatchQueueSize
+	// Default value: 10000
 	TransferProcessorMaxRedispatchQueueSize
-	// TransferProcessorEnableValidator indicates whether validator should be enabled for transferQueueProcessor
-	TransferProcessorEnableValidator
```
yycptt, line 721: nit: maybe we can preserve the original order if that doesn't cost too much effort?

yycptt (approved):
👍 👍 👍

common/dynamicconfig/constants.go
longquanzheng: Good catch. I accidently moved this one. will  move it back

PR #4132, opened 2021-04-13
Update domain registration to check required data field

Shaddoll:
**What changed?**
RegisterDomain handler is updated to check if required data field exists in request.

**Why?**
To allow user to configure required data fields for domains. For example, we'd like to enforce `Tier` info for all domains.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -84,6 +84,7 @@ var keys = map[Key]string{
 	EnableWorkflowShadower:              "system.enableWorkflowShadower",
 	EnableStickyQuery:                   "system.enableStickyQuery",
 	EnableDebugMode:                     "system.enableDebugMode",
+	RequiredDomainDataKeys:              "system.requiredDomainDataKeys",
 
 	// size limit
 	BlobSizeLimitError:      "limit.blobSize.error",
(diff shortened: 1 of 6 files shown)


common/dynamicconfig/constants.go
```diff
@@ -426,6 +427,8 @@ const (
 	DisallowQuery
 	// EnableDebugMode is the key for enabling debugging components, logs and metrics
 	EnableDebugMode
+	// RequiredDomainDataKeys is the key for the list of data keys required for domain registeration
```
vytautas-karpavicius, line 430: typo in: registeration

service/frontend/workflowHandler_test.go
```diff
@@ -479,6 +479,23 @@ func (s *workflowHandlerSuite) TestStartWorkflowExecution_Failed_InvalidTaskStar
 	s.Equal(errInvalidTaskStartToCloseTimeoutSeconds, err)
 }
 
+func (s *workflowHandlerSuite) TestRegisterDomain_Failure_MissingDomainDataKey() {
```
vytautas-karpavicius, line 482: Could you a test case for checking behavior when dynamic config returns nil? I imagine it will be default until we create a value for it.

Shaddoll: Other test cases have already cover the case.

vytautas-karpavicius (approved):

PR #4134, opened 2021-04-14
Add print json flag to list domains command

Shaddoll:
**What changed?**
List domains CLI command is updated to support JSON output.

**Why?**
JSON output can be piped with jq. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/domainCommands.go
+++ b/tools/cli/domainCommands.go
@@ -22,6 +22,7 @@ package cli
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


tools/cli/domainCommands.go
```diff
@@ -447,13 +448,36 @@ func (d *domainCLIImpl) ListDomains(c *cli.Context) {
...
 	}
 
 	domains := d.getAllDomains(c)
 
+	if printJSON {
+		var filteredDomains []*types.DescribeDomainResponse
+		if printAll {
+			filteredDomains = domains
+		} else {
+			filteredDomains = make([]*types.DescribeDomainResponse, 0, len(domains))
```
vytautas-karpavicius, line 464: It seems that we are duplicating filtering logic, as it there is similar code for filtering when printing a table.

Could we make this filtering happen before, to be available for both rendering modes (table and json) ?

vytautas-karpavicius (approved):

yycptt (approved):

PR #4136, opened 2021-04-14
Add persistence info to DescribeCluster API call

mantas-sidlauskas:
**What changed?**
`DescribeCluster` API call will add `PersistenceInfo` to response.
To fully implement this feature, will update IDL and handler accordingly.

**Why?**
If we expose available persistence capabilities, it helps to decide if advanced UI features can be enabled.

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/types/admin.go
+++ b/common/types/admin.go
@@ -46,6 +46,7 @@ func (v *AddSearchAttributeRequest) GetSecurityToken() (o string) {
 type DescribeClusterResponse struct {
 	SupportedClientVersions *SupportedClientVersions `json:"supportedClientVersions,omitempty"`
 	MembershipInfo          *MembershipInfo          `json:"membershipInfo,omitempty"`
+	PersistenceInfo         *PersistenceInfo         `json:"persistenceInfo,omitempty"`
 }
 
 // GetSupportedClientVersions is an internal getter (TBD...)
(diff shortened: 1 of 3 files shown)


Shaddoll (approved):

common/types/admin.go
```diff
@@ -277,6 +278,12 @@ func (v *MembershipInfo) GetRings() (o []*RingInfo) {
 	return
 }
 
+// PersistenceInfo is an internal type which exposes persistence configuration
+type PersistenceInfo struct {
+	Name                      *string `json:"name,omitempty"`
```
yycptt, line 283: Looks like the `Name` field is only for visibility persistence? I assume we will also add info for the persistence for execution and history to the describe cluster result, so may be we can rename this to something like VisibilityPersistenceName?

mantas-sidlauskas: err, yes. Good catch!

just-at-uber (approved):

yycptt (approved):

common/types/admin.go
```diff
@@ -44,8 +44,9 @@ func (v *AddSearchAttributeRequest) GetSecurityToken() (o string) {
 
 // DescribeClusterResponse is an internal type (TBD...)
 type DescribeClusterResponse struct {
```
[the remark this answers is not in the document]
mantas-sidlauskas: yes, that is next up

PR #4138, opened 2021-04-16
Add debug info when List API is hititng ratelimiting/sampling

longquanzheng:
**What changed?**
Add debug info when List API is hititng ratelimiting/sampling

**Why?**
This has been an ongoing forever complaining from Cadence users: https://github.com/uber/cadence/issues/3900#issuecomment-765246025

Even though we have increased the default to 10.

I need a better way to help users to debug it. 

**How did you test it?**
local test(reduce the default maxQPS to 1 and enable debug logs:
```
{"level":"debug","ts":"2021-04-15T23:30:50.086-0700","msg":"List API request consumed QPS token","service":"cadence-frontend","wf-domain-name":"samples-domain","name":"github.com/uber/cadence/common/persistence.(*visibilitySamplingClient).ListClosedWorkflowExecutions","logging-call-at":"visibilitySamplingClient.go:328"}
```
and
```
{"level":"debug","ts":"2021-04-15T19:00:21.956-0700","msg":"List API request is being sampled","service":"cadence-frontend","wf-domain-name":"samples-domain","name":"github.com/uber/cadence/common/persistence.(*visibilitySamplingClient).ListClosedWorkflowExecutions","logging-call-at":"visibilitySamplingClient.go:326"}
```

**Potential risks**
No

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/visibilitySamplingClient.go
+++ b/common/persistence/visibilitySamplingClient.go
@@ -22,8 +22,6 @@ package persistence
 
 import (
 	"context"
-	"fmt"
-	"path/filepath"
 	"runtime"
 	"sync"
 
@@ -325,14 +323,15 @@ func getRequestPriority(request *RecordWorkflowExecutionClosedRequest) int {
 }
 
 func (p *visibilitySamplingClient) returnServiceBusyErrorForList(domainName string) error {
-	p.logger.Debug("List API request is being sampled", tag.WorkflowDomainName(domainName), tag.Name(caller(1)))
+	p.logger.Debug("List API request is being sampled", tag.WorkflowDomainName(domainName), tag.Name(callerFuncName(2)))
 	return errPersistenceLimitExceededForList
 }
 
-func caller(skip int) string {
-	_, path, lineno, ok := runtime.Caller(skip)
-	if !ok {
-		return ""
+func callerFuncName(skip int) string {
+	pc, _, _, ok := runtime.Caller(skip)
+	details := runtime.FuncForPC(pc)
+	if ok && details != nil {
+		return details.Name()
 	}
-	return fmt.Sprintf("%v:%v", filepath.Base(path), lineno)
+	return ""
 }
--- a/common/persistence/visibilitySamplingClient.go
+++ b/common/persistence/visibilitySamplingClient.go
@@ -22,6 +22,9 @@ package persistence
 
 import (
 	"context"
+	"fmt"
+	"path/filepath"
+	"runtime"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


common/persistence/visibilitySamplingClient.go
```diff
@@ -176,8 +180,8 @@ func (p *visibilitySamplingClient) ListOpenWorkflowExecutions(
 	domain := request.Domain
 
 	rateLimiter := p.rateLimitersForList.getRateLimiter(domain, numOfPriorityForList, p.config.VisibilityListMaxQPS(domain))
-	if ok, _ := rateLimiter.GetToken(0, 1); !ok {
-		return nil, ErrPersistenceLimitExceededForList
+	if !p.tryConsumeListToken(request.Domain, rateLimiter) {
```
mantas-sidlauskas, line 183: Nit/for a consideration:
we can reduce code duplication a bit and keep logic in a single place with a helper method:
```go
func (p *visibilitySamplingClient) isRateLimited(domain string) bool {
	rateLimiter := p.rateLimitersForList.getRateLimiter(domain, numOfPriorityForList, p.config.VisibilityListMaxQPS(domain))
	ok, _ := rateLimiter.GetToken(0, 1)
	if ok {
		p.logger.Debug("List API request consumed QPS token", tag.WorkflowDomainName(domain), tag.Name(callerFuncName(2)))
	}
	return !ok
}
```
Then methods looks simply like this:
```go
	if p.isRateLimited(request.Domain) {
		return nil, p.returnServiceBusyErrorForList(request.Domain)
	}

	return <...>
```

longquanzheng: that's a good idea. Thanks Mantas. Will do it.

meiliang86:
Do we print the actual config somehow? I have a feeling that we are not using the correct value somehow.
Hard to imagine that the rate limiter logic itself has some bug.

longquanzheng:
> Do we print the actual config somehow? I have a feeling that we are not using the correct value somehow.
> Hard to imagine that the rate limiter logic itself has some bug.

Yes we do, but only when `log level` is `debug`. Yeah I agreed, my guess is there is some misuse somewhere on their side. So this PR is to provide the information for their debugging.

common/persistence/visibilitySamplingClient.go
longquanzheng: Adopted your idea with a lightly difference. :D thanks so much.

longquanzheng:
@mantas-sidlauskas @meiliang86 LMK if you have any other comments about it

meiliang86 (approved):

PR #4142, opened 2021-04-18
Fixing typos in README.md File

noiarek:
**Describe what has changed in this PR** 
`README.md` was modified, 2 words typos were fixed. The first one, the `docker-compose.yaml` file doesn't match with the actual filename and the second one, `PosstgreSQL`, was fixed (an extra s in the name)

Also, the linter modified to files so I commited them too.

**Tell your future self why have you made these changes**
To point to the right file and to have the right database name.

**How have you verified this change? Tested locally? Added a unit test? Checked in staging env?** 
No need to verified this change, it's a readme file

**Assuming the worst case, what can be broken when deploying this change to production?**
Nothing

**Is it notable for release? e.g. schema updates, configuration or data migration required? If so, please mention it, and also update CHANGELOG.md**
No

**Is there any documentation updates should be made for config, https://cadenceworkflow.io/docs/operation-guide/setup/ ? If so, please open an PR in https://github.com/uber/cadence-docs**
No

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -26,7 +26,7 @@ docker pull ubercadence/server:master-auto-setup
 
 Using different docker-compose files
 -----------------------
-By default `docker-compose up` will run with `docker-compose.yaml` in this folder.
+By default `docker-compose up` will run with `docker-compose.yml` in this folder.
 This compose file is running with Cassandra, with basic visibility, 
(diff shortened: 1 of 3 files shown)


longquanzheng (approved):
Thanks for the good catches!

PR #4154, opened 2021-04-23
Elasticsearch TLS support

spmistry:
**What changed?**
Additional support for connection to Elastic search over TLS for v6 and v7 clients

**Why?**
Enhancements to support Elasticsearch providers enforcing TLS connections using signed certificates.

**How did you test it?**
Connecting to Elasticsearch deployment requiring TLS and pem files

**Potential risks**
None. If TLS config isn't set default behavior will be enforced.

**Release notes**
ES configuration will have an addition nested field for TLS config [here](https://github.com/uber/cadence/blob/12579bd153360679f8442a3589f35f279f19ec6b/common/config/tls.go#L25)

**Documentation Changes**
Added documentation in visibility-on-elasticsearch.md


CONTRIBUTING.md
```diff
@@ -118,7 +118,8 @@ This will run all the tests excluding end-to-end integration test in host/ packa
 ```bash
 make test
 ```
-> Note that you will see some test failures because of errors connecting to MySQL/Postgres if only Cassandra is up. This is okay if you don't write any code related to persistence layer. 
+:warning: Note:
```
longquanzheng, line 121: TIL! I didn't know that we can have a eye catching sign like this. Thank you!

common/config/elasticsearch.go
```diff
@@ -48,6 +48,9 @@ type (
 		// optional to use AWS signing client
 		// See more info https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service
 		AWSSigning AWSSigning `yaml:"awsSigning"`
+		// optional to use Signed Certificates over https
+		// See more info https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service
```
longquanzheng, line 52: This comment is not relevant.

docs/visibility-on-elasticsearch.md
```diff
@@ -75,6 +75,25 @@ This part is used to config advanced visibility store to ElasticSearch.
  - `url` is for Cadence to discover ES 
  - `indices/visibility` is ElasticSearch index name for the deployment.  
 
+Optional TLS Support can be enabled by setting the TLS config as follows:
```
longquanzheng, line 78: Thanks for adding documentation. Can you also open a PR in cadence-docs to add the same docs in https://cadenceworkflow.io/docs/concepts/search-workflows/ ?

longquanzheng (approved):
LGTM, just some minor comments. 

Also since we don't have integ tests for this, can you try run it locally to make sure it works?

common/config/elasticsearch.go
spmistry: removed

docs/visibility-on-elasticsearch.md
spmistry: Created: https://github.com/uber/cadence-docs/pull/40

longquanzheng (approved):

PR #4161, opened 2021-04-25
Add Signal Name to Signal Metrics

demirkayaender:
**What changed?**
Adding signalName as a tag to the signal metrics. This will help filtering signal metrics by signal name and create alerts. 

See https://t3.uberinternal.com/browse/CDNC-1232 for more details. 

Original ask was to include workflowType in the metric name but that doesn't seem to be accessible within the SignalWorkflowExecution requests. Instead of the workflowType signalName seems to be a good candidate. This approach needs to be discussed before landing, sending the changes first as the intro of the discussion.

**Why?**
This PR is from [this task](https://t3.uberinternal.com/browse/CDNC-1232) as a followup to the incident [here](https://incidents.uberinternal.com/incident/27d9dec4-a849-4242-b948-106f6dabf738)

The customer wants to trigger an alert whenwhen the signal metric shows abnormally. (e.g too much error, no signal, and signal spike abnormally)

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -43,6 +43,7 @@ const (
 	invariantType  = "invariantType"
 	kafkaPartition = "kafkaPartition"
 	transport      = "transport"
+	signalName     = "signalName"
 
 	domainAllValue = "all"
 	unknownValue   = "_unknown_"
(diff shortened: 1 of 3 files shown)


mkolodezny (approved):
LGTM, how did you test it?

service/history/handler.go
```diff
@@ -1030,6 +1034,7 @@ func (h *handlerImpl) SignalWorkflowExecution(
 	defer log.CapturePanic(h.GetLogger(), &retError)
 	h.startWG.Wait()
 
+	ctx = withSignalName(ctx, wrappedRequest.SignalRequest.GetSignalName())
```
yux0, line 1037: As this is for customer use, we just need to add it in front end service. No need to add the tag in history.

meiliang86 (approved):

PR #4165, opened 2021-04-29
Deep merge config files

vytautas-karpavicius:
**Why?**
While current config loading support hierarchy:
```
// The hierarchy is as follows from lowest to highest
//
//   base.yaml
//       env.yaml   -- environment is one of the input params ex-development
//         env_az.yaml -- zone is another input param
```
It will **replace** entire config block found in the most specific configuration, but not **merge** it. As our config blocks are rather large, this means duplicating them in many config files for different environments. This make config files harder to maintain.

**What changed?**
Instead use `go.uber.org/config` library that natively supports deep merging multiple configuration sources.

Also updated our `development_*.yaml` files for different persistence stores to only include minimal set changes compared to `development.yaml`.

**How did you test it?**
Updated unit tests, to test deep merging.

I have tried starting Cadence locally with different zones (`CADENCE_AVAILABILTY_ZONE`), to pickup different config files, and output their value to console for inspection.

**Potential risks**
Cadence could fail to start on some particular environment if it is configured incorrectly.

**Release notes**

**Documentation Changes**


longquanzheng:
Note: We should add release notes to let users update their config for this change

vytautas-karpavicius:
> Note: We should add release notes to let users update their config for this change

Thanks. I have updated the change log.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ You can find a list of previous releases on the [github releases](https://github
 
 ### Changed
 - Bump CLI version to v0.18.3 (#3959)
+- Server will deep merge configuration files when loading. No need to copy the whole config section, specify only those fields that needs overriding. (#4165)
```
yux0, line 18: Maybe update this in latest release section

vytautas-karpavicius: Rebased and updated

yux0 (approved):

PR #4169, opened 2021-04-30
Support filtering Cron workflows

demirkayaender:
**What changed?**
This change adds is_cron search attribute to workflow search attributes. That attribute will help us filter cron workflows so we can see scheduled workflows in the list workflow UI

Note: this change requires new attribute to be whitelisted. Here is the whitelist task: https://code.uberinternal.com/T8781809

**Why?**
Customer wanted to see scheduled workflows so they can cancel some of them.
See https://t3.uberinternal.com/browse/CDNC-1249 for more details.

**How did you test it?**
new workflowHandler unit tests added with this change.

**Potential risks**

**Release notes**

**Documentation Changes**


just-at-uber (approved):

longquanzheng:
Should we also add release instructions about how to add search attributes in ES and dynamicconfig? I guess we have done it internally, so maybe just for open source users

demirkayaender:
@longquanzheng, that's a good idea! I will check with the team and create a doc for this

PR #4172, opened 2021-04-30
Detect when IDL submodule is not initialized and warn user, improve .fake-codegen

### LINKED ISSUE #4162: cadence-workflow 0.20.0 build issue on apple arm64
👋 trying to build the latest release, but run into some build issue. The error log is as below:

<details>
<summary>build failure</summary>

```
==> make cadence cadence-server cadence-canary cadence-sql-tool cadence-cassandra-tool
building revive from github.com/mgechev/revive...
building goimports from golang.org/x/tools/cmd/goimports...
downloading protoc 3.14.0
building protoc-gen-gofast from github.com/gogo/protobuf/protoc-gen-gofast...
building protoc-gen-yarpc-go from go.uber.org/yarpc/encoding/protobuf/protoc-gen-yarpc-go...
[.bin/protoc.zip]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of .bin/protoc.zip or
        .bin/protoc.zip.zip, and cannot find .bin/protoc.zip.ZIP, period.
make: *** [.bin/protoc-3.14.0] Error 9
make: *** Waiting for unfinished jobs....
```

</details>

Full build log is in here, https://github.com/Homebrew/homebrew-core/runs/2372116200
relates to https://github.com/Homebrew/homebrew-core/pull/75387

Groxx:
This is largely a continuation of #4155 , and is related to / helps resolve #4162 .

When the IDL submodule doesn't have files in it, builds can fail in a variety of obtuse ways.
Currently it fails when running protoc, as there's a symlink from proto/ into idls/, e.g. causing error from #4155 :
```
uber/cadence/matching/v1/service.proto:213:12: "api.v1.TaskListPartitionMetadata" is not defined.
proto/public: warning: directory does not exist.
uber/cadence/api/v1/history.proto: File not found.
...
```

Since this needs to be fixed *before* running a `make ...`, as its files are used as target prerequisites,
it can't be auto-initialized by the makefile.  Instead, it now fails the build with a descriptive error,
and the user needs to resolve it before trying again.

---

While building and testing this I also noticed that the fake protoc/thrift binaries were sometimes surprising me,
so I added a warning about them to `make clean`.  They *might* be reasonable to automatically clean up as well,
but I'm leaving them in place for now, so e.g. a build process doesn't accidentally fake -> clean -> do a full build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ PROTOC_UNZIP_DIR = $(BIN)/protoc-$(PROTOC_VERSION)-zip
 # otherwise this must be a .PHONY rule, or the buf bin / symlink could become out of date.
 PROTOC_VERSION_BIN = protoc-$(PROTOC_VERSION)
 $(BIN)/$(PROTOC_VERSION_BIN): | $(BIN)
-	@echo "downloading protoc $(PROTOC_VERSION)"
+	@echo "downloading protoc $(PROTOC_VERSION): $(PROTOC_URL)"
 	@# recover from partial success
 	@rm -rf $(BIN)/protoc.zip $(PROTOC_UNZIP_DIR)
 	@# download, unzip, copy to a normal location
@@ -177,6 +177,15 @@ $(BIN)/$(PROTOC_VERSION_BIN): | $(BIN)
 # ====================================
 # Codegen targets
 # ====================================
+
+# IDL submodule must be populated, or files will not exist -> prerequisites will be wrong -> build will fail.
+# Because it must exist before the makefile is parsed, this cannot be done automatically as part of a build.
+# Instead: call this func in targets that require the submodule to exist, so that target will not be built.
+#
+# THRIFT_FILES is just an easy identifier for "the submodule has files", others would work fine as well.
+define ensure_idl_submodule
+$(if $(THRIFT_FILES),,$(error idls/ submodule must exist, or build will fail.  Run `git submodule update --init` and try again))
+endef
 
 # codegen is done when thrift and protoc are done
 $(BUILD)/codegen: $(BUILD)/thrift $(BUILD)/protoc | $(BUILD)
@@ -190,6 +199,7 @@ THRIFT_GEN := $(subst idls/thrift/,.build/,$(THRIFT_FILES))
 
 # thrift is done when all sub-thrifts are done
 $(BUILD)/thrift: $(THRIFT_GEN) | $(BUILD)
+	$(call ensure_idl_submodule)
 	@touch $@
 
 # how to generate each thrift book-keeping file.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -252,7 +266,10 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 
 .fake-thrift: | $(BIN) $(BUILD)
 	touch $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc
-	$(if $(THRIFT_GEN),touch $(THRIFT_GEN),) # maybe ignoring empty idl folder
+	@# if the submodule exists, touch thrift_gen markers to fake their generation.
+	@# if it does not, do nothing - there are none.
+	$(if $(THRIFT_GEN),touch $(THRIFT_GEN),)
+	touch $(BUILD)/thrift
```
Groxx, line 272: now that $(BUILD)/thrift does more, we need to skip it as well.

Makefile
```diff
@@ -166,7 +166,7 @@ PROTOC_UNZIP_DIR = $(BIN)/protoc-$(PROTOC_VERSION)-zip
 # otherwise this must be a .PHONY rule, or the buf bin / symlink could become out of date.
 PROTOC_VERSION_BIN = protoc-$(PROTOC_VERSION)
 $(BIN)/$(PROTOC_VERSION_BIN): | $(BIN)
-	@echo "downloading protoc $(PROTOC_VERSION)"
+	@echo "downloading protoc $(PROTOC_VERSION): $(PROTOC_URL)"
```
Groxx, line 169: Hopefully helps #4162 - I'm not sure what's happening there, but it may be a bad URL.

Makefile
```diff
@@ -178,6 +178,15 @@ $(BIN)/$(PROTOC_VERSION_BIN): | $(BIN)
 # Codegen targets
 # ====================================
 
+# IDL submodule must be populated, or files will not exist -> prerequisites will be wrong -> build will fail.
+# Because it must exist before the makefile is parsed, this cannot be done automatically as part of a build.
+# Instead: call this func in targets that require the submodule to exist, so that target will not be built.
```
longquanzheng, line 183: > "this cannot be done automatically as part of a build"

Just curious: why this can be done for thrift but not proto?

longquanzheng (approved):
Thank you so much. Not very familiar with Makefile, but lgtm.

Makefile
Groxx: neither works / it's used in both.  but they are a bit different:
- thrift runs a "read file -> produce generated code" _per thrift file_, and they're passed in as arguments
- protoc runs a _single_ generator call, and it discovers all relevant files in the folders you tell it about

So $(BUILD)/thrift depends on [each thrift file's generation], which means those have to be defined at parse-time, or the makefile won't set up the commands / infer those dependencies (since there are none).  This is difficult to deal with in make, since that "what will be executed" tree is created once at parse-time, and you can't update it later.

Protoc doesn't actually depend on the files at a makefile-level, so as long as the submodule exists before it runs, it's fine.  Or it can easily be made to be fine, anyway, since the dependency tree doesn't need to change whether those files exist or not.

Groxx: ah, and just to be clear: the only reason I'm using `THRIFT_FILES` is because it's easy and sufficient - it's empty if the idls submodule is empty.  `PROTO_FILES` is not, because some of the files are in the `proto` folder instead.

Groxx:
well apparently that label isn't the (only?) reason automerge is failing.
oh well.

PR #4173, opened 2021-05-03
Add dynamic configs and metrics to control/monitor per id lengths

andrewjdawson2016:


service/frontend/workflowHandler.go
```diff
@@ -3579,19 +3702,18 @@ func (wh *WorkflowHandler) validateTaskList(t *types.TaskList, scope metrics.Sco
...
+
+	if !common.ValidIDLength(
+		t.GetName(),
+		scope,
+		wh.config.MaxIDLengthWarnLimit(),
+		wh.config.TaskListNameMaxLength(domain),
+		metrics.CadenceErrTaskListNameExceededWarnLimit) {
 		return wh.error(errTaskListTooLong, scope)
 	}
-	if wh.isRawListList(t) && len(t.GetName()) > wh.config.MaxRawTaskListNameLimit(domain) {
```
yycptt, line 3585: For scalable tasklist I think we will generate task list names based on the value user provided, and the generated name may exceed the limit. Will that be a concern and stop it from working? I guess it's probably fine as those checks is only enforced frontend and decision where only raw task list name is used. Not sure if there's a case I miss.

andrewjdawson2016: Yes I think its fine. Basically these checks will operate on the raw task list and we just need to make sure we reduce this limit to a point where we have enough buffer to tolerate the extension of scalable task list.

yycptt (approved):

service/frontend/service.go
```diff
@@ -112,9 +119,14 @@ func NewConfig(dc *dynamicconfig.Collection, numHistoryShards int, enableReadFro
...
 		MaxIDLengthWarnLimit:                        dc.GetIntProperty(dynamicconfig.MaxIDLengthWarnLimit, 128),
-		MaxRawTaskListNameLimit:                     dc.GetIntPropertyFilteredByDomain(dynamicconfig.MaxRawTaskListNameLimit, 1000),
+		DomainNameMaxLength:                         dc.GetIntPropertyFilteredByDomain(dynamicconfig.DomainNameMaxLength, 1000),
+		IdentityMaxLength:                           dc.GetIntPropertyFilteredByDomain(dynamicconfig.IdentityMaxLength, 1000),
+		WorkflowIDMaxLength:                         dc.GetIntPropertyFilteredByDomain(dynamicconfig.WorkflowIDMaxLength, 1000),
+		SignalNameMaxLength:                         dc.GetIntPropertyFilteredByDomain(dynamicconfig.SignalNameMaxLength, 1000),
+		WorkflowTypeMaxLength:                       dc.GetIntPropertyFilteredByDomain(dynamicconfig.WorkflowTypeMaxLength, 1000),
+		RequestIDMaxLength:                          dc.GetIntPropertyFilteredByDomain(dynamicconfig.RequestIDMaxLength, 1000),
+		TaskListNameMaxLength:                       dc.GetIntPropertyFilteredByDomain(dynamicconfig.TaskListNameMaxLength, 1000),
```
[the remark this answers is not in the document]
yycptt: Created constants in common/constant.go for warn and error limit.

yux0 (approved):

PR #4175, opened 2021-05-04
Implement new matching and frontend API to get all tasklists for a domain

anish531213:
**What changed?**
Introduced new API called GetTaskListForDomain to get all tasklists for a domain. For this introduced a cache in matching engine that keeps track of tasklist and domain relation.

**Why?**
Introduced this API so this can be used to get granular tasklist information for a domain. This is planned to be used in failover manager but can have other uses in the future.

**How did you test it?**
Just introduced this API. Will add the unit test and verify in staging eng.

**Potential risks**
I don't see any potential risks.

**Release notes**

**Documentation Changes**


client/frontend/client.go
```diff
@@ -690,3 +690,20 @@ func (c *clientImpl) ListTaskListPartitions(
 
 	return client.ListTaskListPartitions(ctx, request, opts...)
 }
+
+func (c *clientImpl) GetTaskListsForDomain(
```
yux0, line 694: what about GetTaskListsByDomain

service/matching/handler.go
```diff
@@ -391,6 +392,32 @@ func (h *handlerImpl) ListTaskListPartitions(
 	return response, hCtx.handleErr(err)
 }
 
+// ListTaskListPartitions returns information about partitions for a taskList
```
yux0, line 395: API doc needs to be updated

service/matching/matchingEngine.go
```diff
@@ -211,6 +216,24 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 	return mgr, nil
 }
 
+func (e *matchingEngineImpl) cacheTaskListByDomain(taskList *taskListID) error {
+	domainName, err := e.domainCache.GetDomainName(taskList.domainID)
+	if err != nil {
+		return err
+	}
+	cachedTL := e.taskListCache.Get(domainName)
```
yux0, line 224: can we store domain ID to tasklist mapping?

common/clientCache.go
```diff
@@ -98,3 +100,11 @@ func (c *clientCacheImpl) GetClientForClientKey(clientKey string) (interface{},
 	c.clients[clientKey] = client
 	return client, nil
 }
+
+func (c *clientCacheImpl) GetAllTaskLists() []string {
+	taskLists := make([]string, 0, len(c.clients))
+	for tl, _ := range c.clients {
```
yux0, line 106: These are clients. Why the API name is related to task list?

client/matching/client.go
```diff
@@ -232,6 +232,33 @@ func (c *clientImpl) ListTaskListPartitions(
 	return client.ListTaskListPartitions(ctx, request, opts...)
 }
 
+func (c *clientImpl) GetTaskListsForDomain(
+	ctx context.Context,
+	request *types.MatchingGetTaskListsForDomainRequest,
+	opts ...yarpc.CallOption,
+) (*types.GetTaskListsForDomainResponse, error) {
+	opts = common.AggregateYarpcOptions(ctx, opts...)
+	clients := c.clients.GetAllTaskLists()
```
yux0, line 241: The result could be large?

client/matching/client.go
```diff
@@ -232,6 +232,33 @@ func (c *clientImpl) ListTaskListPartitions(
...
+
+	var response *types.GetTaskListsForDomainResponse
+	for _, tl := range clients {
```
yux0, line 244: call getTaskListForDomain for each client?

service/matching/matchingEngine.go
```diff
@@ -211,6 +216,24 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
...
+	var taskLists map[string]bool
+	if cachedTL != nil {
+		taskLists = cachedTL.(map[string]bool)
+	} else {
+		taskLists = map[string]bool{}
+	}
+
+	taskLists[taskList.name] = true
```
yycptt, line 232: When will cached tasklist be removed? I think matching engine has an idle check for each tasklist and we can remove tasklist from the cache if it's idle for too long and be unloaded.

service/matching/matchingEngine.go
```diff
@@ -167,6 +169,8 @@ func (e *matchingEngineImpl) String() string {
 // if successful creates one.
 func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 	taskListKind *types.TaskListKind) (taskListManager, error) {
+	// Cache the task list in taskListCache
+	e.cacheTaskListByDomain(taskList)
```
yycptt, line 173: Only cache user defined task lists? Other tasklists like sticky tasklist should not be cached I think?

service/matching/matchingEngine.go
```diff
@@ -211,6 +216,24 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
...
 
+func (e *matchingEngineImpl) cacheTaskListByDomain(taskList *taskListID) error {
+	domainName, err := e.domainCache.GetDomainName(taskList.domainID)
+	if err != nil {
+		return err
+	}
+	cachedTL := e.taskListCache.Get(domainName)
+	var taskLists map[string]bool
+	if cachedTL != nil {
+		taskLists = cachedTL.(map[string]bool)
```
yycptt, line 227: In this case, PutIfNotExist below won't do anything as the domainName already exists in the cache?

service/matching/matchingEngine.go
```diff
@@ -211,6 +216,24 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
...
+	} else {
+		taskLists = map[string]bool{}
+	}
+
+	taskLists[taskList.name] = true
+	e.taskListCache.PutIfNotExist(domainName, taskLists)
```
yycptt, line 233: Looks like there might be some race condition here like two goroutines tries to cache two different tasklists at the same time, one of them will be lost. Not sure if this is an issue or not as this is just for caching. So, if not an issue, maybe add some comments explaining why the race condition is ignored.

service/matching/matchingEngine.go
yux0: Good point. I think we can support TTL in cache?

client/frontend/client.go
anish531213: Updated

service/matching/handler.go
anish531213: Updated

service/matching/matchingEngine.go
anish531213: How can i differentiate user defined and stiky tasklist?

service/matching/matchingEngine.go
anish531213: I think this is not an issue as we have all the tasklistCache gets populated eventually.

client/matching/client.go
```diff
@@ -232,6 +232,33 @@ func (c *clientImpl) ListTaskListPartitions(
...
+		if err != nil {
+			return nil, err
+		}
+
+		resp, err := c.getTaskListForDomain(client, ctx, request, opts...)
+		if err != nil {
+			return nil, err
+		}
+		for _, tl := range resp.TaskListNames {
+			response.TaskListNames = append(response.TaskListNames, tl)
```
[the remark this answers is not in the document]
anish531213: Same, not an issue here because as we can handle this when we get the response.

service/matching/matchingEngine.go
anish531213: Updated with a comment.

client/matching/client.go
anish531213: Yes. I am calling getTaskListForDomain for each client.

client/matching/client.go
anish531213: Should I use pagination here? What do you think?

service/matching/matchingEngine.go
yycptt: Check the `isRawListList()` method in frontend.

common/clientCache.go
```diff
@@ -98,3 +100,11 @@ func (c *clientCacheImpl) GetClientForClientKey(clientKey string) (interface{},
 	c.clients[clientKey] = client
 	return client, nil
 }
+
+func (c *clientCacheImpl) GetAllClients() []string {
```
yux0, line 104: This returns all keys. You can just return all clients here.

client/matching/client.go
yux0: Leave a comment in GetAllClients()

client/matching/client.go
yux0: After getting all clients from GetAllClients, you don't need to call getClientForTaskList

anish531213: Updated.

common/clientCache.go
anish531213: Updated.

service/matching/matchingEngine.go
```diff
@@ -167,6 +169,10 @@ func (e *matchingEngineImpl) String() string {
 // if successful creates one.
 func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 	taskListKind *types.TaskListKind) (taskListManager, error) {
+	// Cache user defined task list in taskListCache
+	if *taskListKind == types.TaskListKindNormal {
```
yux0, line 173: can we also add a feature flag here? a little concern about the perf here

service/matching/matchingEngine.go
yux0: let's update the cache in removeTaskListManager as well

service/matching/matchingEngine.go
yux0: it might be ok to move this under a read lock

anish531213: Updated with a feature flag.

common/clientCache.go
```diff
@@ -98,3 +100,11 @@ func (c *clientCacheImpl) GetClientForClientKey(clientKey string) (interface{},
 	c.clients[clientKey] = client
 	return client, nil
 }
+
+func (c *clientCacheImpl) GetAllClients() []interface{} {
```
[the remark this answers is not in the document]
anish531213: Updated.

service/matching/matchingEngine.go
```diff
@@ -211,6 +221,33 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
...
+func (e *matchingEngineImpl) removeFromTaskListCache(taskList *taskListID) {
+	cachedTL := e.taskListCache.Get(taskList.domainID)
+	var taskLists map[string]bool
+	if cachedTL != nil {
+		taskLists := cachedTL.(map[string]bool)
+		// there might be a race condition here but delete will be a no-op if the tasklist is already deleted
+		delete(taskLists, taskList.name)
+	}
+
+	e.taskListCache.Put(taskList.domainID, taskLists)
```
yux0, line 248: nit: this could be in the if statement.

yux0 (approved):

client/matching/client.go
```diff
@@ -232,6 +232,30 @@ func (c *clientImpl) ListTaskListPartitions(
...
 
+func (c *clientImpl) GetTaskListsByDomain(
+	ctx context.Context,
+	request *types.MatchingGetTaskListsByDomainRequest,
+	opts ...yarpc.CallOption,
+) (*types.GetTaskListsByDomainResponse, error) {
+	opts = common.AggregateYarpcOptions(ctx, opts...)
+
+	// This result could be large
+	clients := c.clients.GetAllClients()
```
yycptt, line 243: This only returns all cached matching clients on the hosts right? If a matching host is never cached on this host then it won't be returned and we may miss tasks list there.

client/matching/client.go
```diff
@@ -250,10 +274,26 @@ func (c *clientImpl) createLongPollContext(
...
+	opts ...yarpc.CallOption,
+) (*types.GetTaskListsByDomainResponse, error) {
+	ctx, cancel := c.createContext(ctx)
+	defer cancel()
+
+	resp, err := cl.GetTaskListsByDomain(ctx, request, opts...)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
```
yycptt, line 298: nit: can be simplified to `return cl.GetTaskListsByDomain(ctx, request, opts...)` ?

common/clientCache.go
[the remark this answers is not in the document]
yycptt: Do we still need to add the note?

client/matching/client.go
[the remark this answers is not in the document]
yycptt: I don't see any dedup logic in frontend handler where the matching client is used?

service/matching/config.go
```diff
@@ -117,6 +118,7 @@ func NewConfig(dc *dynamicconfig.Collection) *Config {
 		ForwarderMaxOutstandingTasks:    dc.GetIntPropertyFilteredByTaskListInfo(dynamicconfig.MatchingForwarderMaxOutstandingTasks, 1),
 		ForwarderMaxRatePerSecond:       dc.GetIntPropertyFilteredByTaskListInfo(dynamicconfig.MatchingForwarderMaxRatePerSecond, 10),
 		ForwarderMaxChildrenPerNode:     dc.GetIntPropertyFilteredByTaskListInfo(dynamicconfig.MatchingForwarderMaxChildrenPerNode, 20),
+		EnableTaskListCache:             dc.GetBoolProperty(dynamicconfig.MatchingEnableTaskListCacheByDomainID, false),
```
yycptt, line 121: dc.GetBoolPropertyByDomainID?

service/matching/matchingEngine.go
```diff
@@ -165,8 +167,17 @@ func (e *matchingEngineImpl) String() string {
...
 // Returns taskListManager for a task list. If not already cached gets new range from DB and
 // if successful creates one.
-func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
-	taskListKind *types.TaskListKind) (taskListManager, error) {
+func (e *matchingEngineImpl) getTaskListManager(
+	taskList *taskListID,
+	taskListKind *types.TaskListKind,
+) (taskListManager, error) {
+	// Cache user defined task list in taskListCache
+	if e.config.EnableTaskListCache() {
```
yycptt, line 175: I think this is a per domain config?

service/matching/matchingEngine.go
```diff
@@ -165,8 +167,17 @@ func (e *matchingEngineImpl) String() string {
...
+		if *taskListKind == types.TaskListKindNormal {
```
yycptt, line 176: Is generated scalable tasklist Kind Normal?

service/matching/matchingEngine.go
```diff
@@ -165,8 +167,17 @@ func (e *matchingEngineImpl) String() string {
...
+			e.cacheTaskListByDomain(taskList)
```
yycptt, line 177: Is the taskListsLock held when calling this func? Otherwise two goroutines may operated on the same cached map I think?

client/matching/client.go
yux0: Change to use get memebers from ringpop

client/matching/client.go
yux0: updated

service/matching/config.go
yux0: updated

service/matching/matchingEngine.go
yux0: updated

service/matching/matchingEngine.go
yux0: Yes. I think so

service/matching/matchingEngine.go
yux0: Yes. Updated.

service/matching/matchingEngine.go
yux0: Yes. They are normal with a partition number

service/matching/matchingEngine.go
```diff
@@ -211,6 +223,38 @@ func (e *matchingEngineImpl) getTaskListManager(taskList *taskListID,
 	return mgr, nil
 }
 
+func (e *matchingEngineImpl) cacheTaskListByDomainLocked(
```
yycptt, line 226: Is it possible to reuse the existing `e.taskLists` field which is basically a cache of all tasklists on the host, instead of creating a new cache? Is the new cache for avoiding lock contention?

service/frontend/thriftHandler.go
```diff
@@ -134,6 +134,12 @@ func (t ThriftHandler) ListTaskListPartitions(ctx context.Context, request *shar
 	return thrift.FromListTaskListPartitionsResponse(response), thrift.FromError(err)
 }
 
+// GetTaskListsByDomain forwards request to the underlying handler
+func (t ThriftHandler) GetTaskListsByDomain(ctx context.Context, request *shared.GetTaskListsByDomainRequest) (*shared.GetTaskListsByDomainResponse, error) {
+	response, err := t.h.GetTaskListsByDomain(ctx, thrift.ToGetTaskListsByDomainRequest(request))
```
yycptt, line 139: `withThriftTag(ctx)`

service/matching/thriftHandler.go
```diff
@@ -85,6 +85,12 @@ func (t ThriftHandler) ListTaskListPartitions(ctx context.Context, request *m.Li
 	return thrift.FromListTaskListPartitionsResponse(response), thrift.FromError(err)
 }
 
+// GetTaskListsByDomain forwards request to the underlying handler
+func (t ThriftHandler) GetTaskListsByDomain(ctx context.Context, request *s.GetTaskListsByDomainRequest) (*s.GetTaskListsByDomainResponse, error) {
+	response, err := t.h.GetTaskListsByDomain(ctx, thrift.ToGetTaskListsByDomainRequest(request))
```
yycptt, line 90: `withThriftTag(ctx)`

client/matching/client.go
```diff
@@ -232,6 +241,49 @@ func (c *clientImpl) ListTaskListPartitions(
...
+		var resp *types.GetTaskListsByDomainResponse
+		if err = future.Get(ctx, &resp); err != nil {
+			return nil, err
+		} else {
+			for _, tl := range resp.TaskListNames {
+				if _, ok := taskListMap[tl]; !ok {
+					taskListMap[tl] = struct{}{}
+				}
+			}
+		}
```
yycptt, line 276: nit: remove else and outdent the for loop?

client/matching/client.go
```diff
@@ -232,6 +241,49 @@ func (c *clientImpl) ListTaskListPartitions(
...
+	clients, err := c.clientIterator()
+	if err != nil {
+		return nil, err
+	}
+
+	var futures []future.Future
+	for _, client := range clients {
+		future, settable := future.NewFuture()
+		go func() {
+			settable.Set(client.(Client).GetTaskListsByDomain(ctx, request, opts...))
```
yycptt, line 259: Do we need to make sure the ctx here has a finite/reasonable timeout here? Otherwise, the created goroutine may be blocked, especially when one future returns an error and we just return from the function without waiting for other futures?

client/matching/client.go
```diff
@@ -26,6 +26,8 @@ import (
 
 	"go.uber.org/yarpc"
 
+	"github.com/uber/cadence/common/future"
+
```
yycptt, line 30: nit: remove empty line

yycptt (approved):
Looks good overall. Feel free to land after addressing the rest of the comments (withThriftTag, ctx timeout, and mapper tests).

client/matching/client.go
yycptt: Looks like the loop variable (`client`) is being captured here. Will that be an issue? shall we pass in the client as a parameter to this inline func?

client/matching/client.go
```diff
@@ -40,25 +41,32 @@ const (
...
 )
 
-type clientImpl struct {
-	timeout         time.Duration
-	longPollTimeout time.Duration
-	clients         common.ClientCache
-	loadBalancer    LoadBalancer
-}
+type (
+	clientIterator func() ([]interface{}, error)
```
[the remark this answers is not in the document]
yux0: Yes. Updated

common/types/mapper/proto/matching.go
```diff
@@ -200,6 +200,42 @@ func ToMatchingListTaskListPartitionsResponse(t *matchingv1.ListTaskListPartitio
 	}
 }
 
+func FromMatchingGetTaskListsByDomainRequest(t *types.GetTaskListsByDomainRequest) *matchingv1.GetTaskListsByDomainRequest {
```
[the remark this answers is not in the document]
yux0: updated

common/types/mapper/thrift/shared.go
```diff
@@ -3052,6 +3052,46 @@ func ToListTaskListPartitionsResponse(t *shared.ListTaskListPartitionsResponse)
 	}
 }
 
+// FromGetTaskListsByDomainRequest converts internal GetTaskListsByDomainRequest type to thrift
+func FromGetTaskListsByDomainRequest(t *types.GetTaskListsByDomainRequest) *shared.GetTaskListsByDomainRequest {
```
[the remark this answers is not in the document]
yux0: updated

service/frontend/thriftHandler.go
yux0: updated

service/matching/thriftHandler.go
yux0: updated

client/matching/client.go
yux0: updated

client/matching/client.go
yux0: updated

client/matching/client.go
yux0: updated

PR #4176, opened 2021-05-04
Support Listing Scheduled Cron Workflows

demirkayaender:
**What changed?**
This change adds IsCron search attribute to workflow search attributes. That attribute will help us filter cron workflows so we can see scheduled workflows in the list workflow UI

Note that changes here depend on following changes; ignore the build/test failures until those changes are in
IDL: https://github.com/uber/cadence-idl/pull/67
Client: https://github.com/uber-go/cadence-client/pull/1095

**Why?**
Customer wanted to see scheduled workflows so they can cancel some of them.

**How did you test it?**
```
go test -v ./host -run TestElasticsearchIntegrationSuite -tags esintegration -testify.m TestListCronWorkflows
```
and the E2E test here: https://github.com/uber-go/cadence-client/pull/1095

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -700,16 +702,27 @@ func (v *esVisibilityStore) checkProducer() {
...
+func getVisibilityMessage(
+	domainID string,
+	wid,
+	rid string,
+	workflowTypeName string,
+	taskList string,
+	startTimeUnixNano,
+	executionTimeUnixNano int64,
+	taskID int64,
+	memo []byte, encoding common.EncodingType,
```
just-at-uber, line 714: Should encoding be on a new line?

just-at-uber (approved):
LGTM. one minor comment

environment/env.go
```diff
@@ -60,7 +60,7 @@ const (
 	// ESVersion is the ElasticSearch version
 	ESVersion = "ES_VERSION"
 	// ESDefaultVersion is the default version
-	ESDefaultVersion = "v6"
+	ESDefaultVersion = "v7"
```
yycptt, line 63: Why do we need to update this version? This may break some developers' dev environment?

yycptt (commented):
I think we also need to update the index template for es?

environment/env.go
demirkayaender: I thought this was an unintentional default as Cadence server uses v7 since that's the running ES version but integration tests use v6. I can revert this to v6 if it's useful

demirkayaender:
> I think we also need to update the index template for es?

How do we update the index template for ES?

environment/env.go
yycptt: Our internal ES cluster is using v6, so I'd say let's revert it to v6?

common/persistence/elasticsearch/esVisibilityStore_test.go
```diff
@@ -127,6 +127,7 @@ func (s *ESVisibilitySuite) TestRecordWorkflowExecutionStarted() {
 	request.StartTimestamp = time.Unix(0, int64(123))
 	request.ExecutionTimestamp = time.Unix(0, int64(321))
 	request.TaskID = int64(111)
+	request.IsCron = false
```
yux0, line 130: could we also validate this field?

yux0 (approved):

common/types/shared.go
```diff
@@ -9520,6 +9520,7 @@ type WorkflowExecutionInfo struct {
 	SearchAttributes  *SearchAttributes             `json:"searchAttributes,omitempty"`
 	AutoResetPoints   *ResetPoints                  `json:"autoResetPoints,omitempty"`
 	TaskList          string                        `json:"taskList,omitempty"`
+	IsCron            bool                          `json:"isCron,omitempty"`
```
yycptt, line 9523: I think this type is also used by the response for DescribeWorkflowExecution API, so shall we also update that API handler to set this field?

demirkayaender: Thanks for that case. Let me take a look.

host/xdc/integration_failover_test.go
```diff
@@ -109,13 +111,18 @@ func (s *integrationClustersTestSuite) SetupSuite() {
...
 	s.Require().NoError(yaml.Unmarshal(confContent, &clusterConfigs))
 
-	c, err := host.NewCluster(clusterConfigs[0], s.logger.WithTags(tag.ClusterName(clusterName[0])))
-	s.Require().NoError(err)
-	s.cluster1 = c
-
-	c, err = host.NewCluster(clusterConfigs[1], s.logger.WithTags(tag.ClusterName(clusterName[1])))
-	s.Require().NoError(err)
-	s.cluster2 = c
+	/*
```
demirkayaender, line 114: @yux0 let me know if this looks good to you until we decide what to do with this test.

PR #4177, opened 2021-05-05
Support managed failover workflow with drill mode

yux0:
**What changed?**
1. workflow registration update
2. some linting fix
3. add a step to reset the domain active cluster after a failover drill

**Why?**
Support managed failover workflow with drill mode

**How did you test it?**
unit tests and local tests

**Potential risks**
N/A

**Release notes**
Internal usage only

**Documentation Changes**
Internal usage only


service/worker/failovermanager/workflow_test.go
```diff
@@ -194,28 +210,46 @@ func (s *failoverWorkflowTestSuite) TestWorkflow_Pause() {
...
+	domains := []string{"d1"}
+	mockFailoverActivityResult := &FailoverActivityResult{
+		SuccessDomains: []string{"d1"},
+	}
+	s.workflowEnv.OnActivity(getDomainsActivityName, mock.Anything, mock.Anything).Return(domains, nil)
+	s.workflowEnv.OnActivity(failoverActivityName, mock.Anything, mock.Anything).Return(mockFailoverActivityResult, nil)
+	params := &FailoverParams{
+		TargetCluster: "t",
+		SourceCluster: "s",
+		DrillWaitTime: 1,
```
yycptt, line 230: nit: probably use `SetOnTimerFiredListener` to validate that the wait timer is scheduled and fired.

service/worker/failovermanager/workflow_test.go
```diff
@@ -39,15 +41,35 @@ import (
...
 
+func (s *failoverWorkflowTestSuite) SetupTest() {
+	s.activityEnv = s.NewTestActivityEnvironment()
+	s.workflowEnv = s.NewTestWorkflowEnvironment()
+	s.workflowEnv.RegisterWorkflowWithOptions(FailoverWorkflow, workflow.RegisterOptions{Name: WorkflowTypeName})
+	s.workflowEnv.RegisterActivityWithOptions(FailoverActivity, activity.RegisterOptions{Name: failoverActivityName})
+	s.workflowEnv.RegisterActivityWithOptions(GetDomainsActivity, activity.RegisterOptions{Name: getDomainsActivityName})
+	s.activityEnv.RegisterActivityWithOptions(FailoverActivity, activity.RegisterOptions{Name: failoverActivityName})
+	s.activityEnv.RegisterActivityWithOptions(GetDomainsActivity, activity.RegisterOptions{Name: getDomainsActivityName})
+}
```
yycptt, line 72: Add a `TearDownTest()` and assert all expectations for mock inside it?

tools/cli/admin.go
```diff
@@ -1029,6 +1029,15 @@ func newAdminFailoverCommands() []cli.Command {
 					Usage: "Optional domains to failover, eg d1,d2..,dn. " +
 						"Only provided domains in source cluster will be failover.",
 				},
+				cli.IntFlag{
+					Name: FlagFailoverDrillWaitTimeWithAlias,
```
yycptt, line 1033: Is this a required flag when cron schedule is specified?

service/worker/failovermanager/workflow.go
```diff
@@ -48,6 +48,7 @@ const (
 	WorkflowTypeName = "cadence-sys-failoverManager-workflow"
 	// WorkflowID will be reused to ensure only one workflow running
 	WorkflowID             = "cadence-failover-manager"
+	DrillWorkflowID        = WorkflowID + "-drill"
```
yycptt, line 51: I thought if we have only one workflow ID we can ensure there's only one failover workflow running at a time?

tools/cli/admin.go
```diff
@@ -1043,7 +1052,13 @@ func newAdminFailoverCommands() []cli.Command {
 					Name:  FlagRunIDWithAlias,
 					Usage: "Optional Failover workflow runID, default is latest runID",
 				},
+				cli.BoolFlag{
+					Name: FlagFailoverDrillWithAlias,
+					Usage: "Optional to query failover workflow or failover drill workflow." +
```
yycptt, line 1057: nit: `query` -> `pause`
Similar for `resume` and `abort` command.

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,69 @@ func getRunID(c *cli.Context) string {
...
+	wf, err := client.StartWorkflow(tcCtx, options, failovermanager.WorkflowTypeName, foParams)
+	if err != nil {
+		ErrorAndExit("Failed to start failover workflow", err)
+	}
+	fmt.Println("Failover workflow started")
+	fmt.Println("wid: " + wf.ID)
+	fmt.Println("rid: " + wf.RunID)
+}
+
+func getWorkflowID(c *cli.Context) string {
```
yycptt, line 259: nit: shall we make the fn name a bit more specific like `getFailoverWorkflowID`? I think `cli` is a quite large package with many util function, so it may be confusing what this function is doing just by looking at the function name. Same suggestion for the `query` function above.

tools/cli/adminFailoverCommands.go
```diff
@@ -281,3 +271,53 @@ func getOperator() string {
...
+	}
+	if params.batchFailoverWaitTimeInSeconds <= 0 {
+		params.batchFailoverWaitTimeInSeconds = defaultBatchFailoverWaitTimeInSeconds
+	}
+	if params.failoverTimeout <= 0 {
+		params.failoverTimeout = defaultFailoverTimeoutInSeconds
+	}
+}
+
+func validateWorkflowID(workflowID string) string {
```
yycptt, line 318: Maybe I missed something, but I don't see this function used anywhere?

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,69 @@ func getRunID(c *cli.Context) string {
...
+	options := cclient.StartWorkflowOptions{
+		ID:                           failovermanager.WorkflowID,
+		WorkflowIDReusePolicy:        cclient.WorkflowIDReusePolicyAllowDuplicate,
+		TaskList:                     failovermanager.TaskListName,
+		ExecutionStartToCloseTimeout: workflowTimeout,
+		Memo: map[string]interface{}{
+			common.MemoKeyForOperator: getOperator(),
+		},
+	}
+	if len(params.cron) > 0 {
```
yycptt, line 227: I think in the workflow implementation whether it's a drill or not is determined based on waitTime?

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,69 @@ func getRunID(c *cli.Context) string {
...
+		options.ID = failovermanager.DrillWorkflowID
+		options.CronSchedule = params.cron
+	} else {
+		client := getCadenceClient(c)
+		tcCtx, cancel := newContext(c)
+		defer cancel()
+
+		// block managed failover if an on-going failover drill
```
yycptt, line 235: Do we need to check for the reverse situation? 

Also if we have only one workflowID then we can get rid of this check?

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,69 @@ func getRunID(c *cli.Context) string {
...
+
+	targetCluster := params.targetCluster
+	sourceCluster := params.sourceCluster
+	batchFailoverSize := params.batchFailoverSize
+	batchFailoverWaitTimeInSeconds := params.batchFailoverWaitTimeInSeconds
+	workflowTimeout := time.Duration(params.failoverTimeout) * time.Second
+	domains := params.domains
+	drillWaitTime := time.Duration(params.drillWaitTime) * time.Second
+
+	client := getCadenceClient(c)
```
yycptt, line 214: I may not have the full context, looks like the failover workflow should run in a global system domain, if we have that?

service/worker/failovermanager/workflow.go
yux0: Yes. In a managed failover, it checks if this drill workflow is running. With this, we don't need to terminate the scheduled failover wf. But I am open to use one wf ID

tools/cli/admin.go
yux0: Yes. Will update it

tools/cli/admin.go
yux0: updated

tools/cli/adminFailoverCommands.go
yux0: I think we decide to run in local domain. We can sync offline.

tools/cli/adminFailoverCommands.go
yux0: yes

tools/cli/adminFailoverCommands.go
yux0: Yes. This should be removed

service/worker/failovermanager/workflow_test.go
yux0: All mocks are in cadence testing framework

service/worker/failovermanager/workflow_test.go
yux0: Added

tools/cli/adminFailoverCommands.go
yux0: Sync offline. Changed to send pause signal

tools/cli/adminFailoverCommands.go
yux0: updated

service/worker/failovermanager/workflow_test.go
yycptt: Yeah, I understand that. but we still need to call `env.AssertExpectations()` to check if all mocks are indeed used. like this: https://github.com/uber-common/cadence-samples/blob/master/cmd/samples/recipes/branch/workflow_test.go#L29

yux0: Sounds good. Added it.

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,73 @@ func getRunID(c *cli.Context) string {
...
+		Memo: map[string]interface{}{
+			common.MemoKeyForOperator: getOperator(),
+		},
+	}
+	if params.drillWaitTime > 0 {
+		options.ID = failovermanager.DrillWorkflowID
+		options.CronSchedule = params.cron
+	} else {
+		// block if there is an on-going failover drill
+		if err := executePauseOrResume(c, failovermanager.DrillWorkflowID, true); err != nil {
```
yycptt, line 232: Looks like it's possible that one drill workflow may receive multiple pause signals? In that case how many resume signal is needed to unblock it?

tools/cli/adminFailoverCommands.go
```diff
@@ -273,6 +200,73 @@ func getRunID(c *cli.Context) string {
...
+		WorkflowIDReusePolicy:        cclient.WorkflowIDReusePolicyAllowDuplicate,
+		TaskList:                     failovermanager.TaskListName,
+		ExecutionStartToCloseTimeout: workflowTimeout,
+		Memo: map[string]interface{}{
+			common.MemoKeyForOperator: getOperator(),
+		},
+	}
+	if params.drillWaitTime > 0 {
+		options.ID = failovermanager.DrillWorkflowID
+		options.CronSchedule = params.cron
```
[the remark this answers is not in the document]
yux0: added a check and return error

tools/cli/adminFailoverCommands.go
yux0: Added a step to clean up the pause channel after receiving a resume signal. This should dedup the pause signals.

yycptt (approved):

PR #4188, opened 2021-05-10
Adding Cassandra client proto version as config option

noiarek:
**What changed?**
ProtoVersion can now be defined in development.yaml. If not present, the default value will be 4.

**Why?**
Labeled as `Up for grabs` as `good first issue` in https://github.com/uber/cadence/issues/3536

**How did you test it?**
Playing with some values in the development.yaml file. Happy to add some unit test if you indicate me where

**Potential risks**
There was a default value set in `newCassandraCluster`, however the value is sent by value and not by reference so it won't be set somewhere else. Potentially the right place would be `tools/cassandra/handler.go` but please let me know where you think I should set this value in case this is not the right place

**Release notes**
Maybe? Let me know if you want me to do that

**Documentation Changes**
Likely to be the case. Let me know if I need to do this.


noiarek:
Hi @vytautas-karpavicius, friendly ping to review this :)

vytautas-karpavicius (approved):
Looks good.

tools/cassandra/cqlclient.go
```diff
@@ -105,7 +105,7 @@ func NewCQLClient(cfg *CQLClientConfig) (*CqlClient, error) {
 		Keyspace:     cfg.Keyspace,
 		TLS:          cfg.TLS,
 		Timeout:      time.Duration(cfg.Timeout) * time.Second,
-		ProtoVersion: cqlProtoVersion,
+		ProtoVersion: cfg.ProtoVersion,
```
yycptt, line 108: - Is this field initialized in `newCQLClientConfig`?
- Shall we provide a flag for specifying the value in the cassandra tool CLI, which is different from the cadence CLI?

config/development.yaml
```diff
@@ -7,10 +7,12 @@ persistence:
       cassandra:
         hosts: "127.0.0.1"
         keyspace: "cadence"
+        protoVersion: 4
```
yycptt, line 10: Please also update `config_template.yaml` in the `docker` folder.

yycptt (changes requested):

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
```diff
@@ -51,7 +51,13 @@ func RegisterClient(c Client) {
...
 	hosts := parseHosts(cfg.Hosts)
 	cluster := gocql.NewCluster(hosts...)
-	cluster.ProtoVersion = 4
+
+	// Keeping compatibility
+	if cfg.ProtoVersion == 0 {
+		cfg.ProtoVersion = 4
+	}
+
+	cluster.ProtoVersion = cfg.ProtoVersion
```
yycptt, line 60: I think the if statement should be moved after the assignment.

common/persistence/nosql/nosqlplugin/cassandra/session.go
```diff
@@ -45,7 +44,7 @@ func CreateSession(cfg config.Cassandra) (gocql.Session, error) {
 		Datacenter:        cfg.Datacenter,
 		MaxConns:          cfg.MaxConns,
 		TLS:               cfg.TLS,
-		ProtoVersion:      cassandraProtoVersion,
+		ProtoVersion:      cfg.ProtoVersion,
```
yycptt, line 47: Can you moved the backward compatibility logic to here? We have an internal implementation for gocql and the implementation in `gocql/client.go` is not used by our internal impl. This function however, is shared by both the internal and open source impl, so having the backward compatibility logic here can ensure our internal impl won't be break.

longquanzheng: +1. Adding config like this should be backward compatible. Even when breaking changes is necessary, we will have a middle version for transition.

tools/cassandra/cqlclient.go
noiarek: 1) It is now :)
2) Is that in `tools/common/schema/types.go` ? If that's the case, that's done

config/development.yaml
noiarek: Done

common/persistence/nosql/nosqlplugin/cassandra/session.go
noiarek: @yycptt Done, thanks for pointing it out. Definitely trying to keep it backward compatible and let me know if you can think in another place where I should add it, I might have missed one or two places.

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
noiarek: If I add it later `cluster` wouldn't have the right value afaik

noiarek (commented):
@yycptt I've addressed your comments and also modified other files where `Protoversion` might be needed.

Two things to add:

1. I don't know if this is related to my change but I'm getting an error when I run `make bins`. This is not allowing me to run the test in local

```
common/types/mapper/thrift/shared.go:3415:3: unknown field 'NextEventId' in struct literal of type shared.PollForDecisionTaskResponse
common/types/mapper/thrift/shared.go:3439:31: t.GetNextEventId undefined (type *shared.PollForDecisionTaskResponse has no field or method GetNextEventId)
```
2. The change is quite big already (not only to review but to merge). I'm happy to split this in smaller changes if you want me to (i.e add compatibility first and then modified config files)

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
yycptt: Oh, Sorry. You are right, I misread the code.

yycptt (commented):
I think `make bins` failed because the IDL submodule is changed to an old version so the generated thrift files are also changed. The build and test should pass once you updated the IDL to the latest commit 68a720ef059629a0cba5db58f71053100acb7f19 by using `git submodule update --remote`

PR #4192, opened 2021-05-10
Add cluster rebalance CLI

yux0:
**What changed?**
Add cluster rebalance CLI

**Why?**
To rebalance domain active cluster based on the preference

**How did you test it?**
local. tests and unit tests

**Potential risks**
N/A

**Release notes**
Internal usage

**Documentation Changes**
Internal usage


tools/cli/admin.go
```diff
@@ -688,6 +688,15 @@ func newAdminClusterCommands() []cli.Command {
 				newDomainCLI(c, false).FailoverDomains(c)
 			},
 		},
+		{
+			Name:    "rebalance",
+			Aliases: []string{"rb"},
+			Usage: "Rebalance cluster will failover managed failover domains to preferred cluster. " +
+				"You can set the preference in domain data with PreferredCluster.",
```
yycptt, line 695: When setting PreferredCluster in domain data, is there any check performed on the value for making sure the value is a valid cluster name?

tools/cli/domainCommands.go
```diff
@@ -353,6 +353,38 @@ func (d *domainCLIImpl) failoverDomains(c *cli.Context) ([]string, []string) {
...
+func (d *domainCLIImpl) RebalanceDomains(c *cli.Context) {
+	// ask user for confirmation
+	prompt("You are trying to rebalance all managed domains, continue? Y/N")
+	d.rebalanceDomains(c)
+}
+
+func (d *domainCLIImpl) rebalanceDomains(c *cli.Context) ([]string, []string) {
+	domains := d.getAllDomains(c)
+	shouldRebalance := func(domain *types.DescribeDomainResponse) bool {
+		return len(getPreferredClusterName(domain)) != 0 && isDomainFailoverManagedByCadence(domain)
```
yycptt, line 366: Also check if domain is deprecated?

tools/cli/domainCommands.go
```diff
@@ -353,6 +353,38 @@ func (d *domainCLIImpl) failoverDomains(c *cli.Context) ([]string, []string) {
...
+	}
+	var succeedDomains []string
+	var failedDomains []string
+	for _, domain := range domains {
+		if shouldRebalance(domain) {
+			domainName := domain.GetDomainInfo().GetName()
+			err := d.failover(c, domainName, getPreferredClusterName(domain))
```
yycptt, line 373: Just want to check my understanding: domain cache is refreshed every 10s so all failover requests happens within the 10 seconds will take affect together and only one failover queue per shard will be created?

If too many domain need to rebalanced, do we need to do them in batches?

tools/cli/domainCommands.go
```diff
@@ -353,6 +353,38 @@ func (d *domainCLIImpl) failoverDomains(c *cli.Context) ([]string, []string) {
 	return succeedDomains, failedDomains
 }
 
+// RebalanceDomains is used for managed failover to rebalance with domain data PreferredCluster
+func (d *domainCLIImpl) RebalanceDomains(c *cli.Context) {
+	// ask user for confirmation
+	prompt("You are trying to rebalance all managed domains, continue? Y/N")
```
yycptt, line 359: A dry run option may be useful here? so that operator can have an understand of the result of the rebalance without actually doing it.

tools/cli/domainCommands_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (s *domainCommandsSuite) TestRebalanceDomains_NoPreferredCluster_Skipped() {
+	listDomainsResp := getListDomainsResponse()
+	listDomainsResp.Domains[0].DomainInfo.Data = map[string]string{
+		common.DomainDataKeyForManagedFailover: "true",
+	}
+	s.serverFrontendClient.EXPECT().ListDomains(gomock.Any(), gomock.Any()).Return(listDomainsResp, nil)
+	s.serverFrontendClient.EXPECT().
+		UpdateDomain(gomock.Any(), gomock.Any()).
+		Return(&types.UpdateDomainResponse{}, nil).Times(0)
```
yycptt, line 114: nit: Looks like failure case is not tested? Probably adding a test to return random value (success/fail) for update domain and assert # success + # fail returned by `rebalanceDomains` equals to # of domains should be failovered.

tools/cli/admin.go
yux0: currently no. but the update domain will failed

tools/cli/domainCommands.go
yux0: Most likely it can finish in 10s so at most two failover queue will be created

tools/cli/domainCommands.go
yycptt: nit: maybe don't print this prompt in dry run mode, so that operation know he's in dry run mode and there'll be no actual impact.

tools/cli/admin.go
yux0: This step is manual. I added a step to verify the data

PR #4196, opened 2021-05-14
Support delaying cron workflows

demirkayaender:
**What changed?**
About a month ago, we added delaying workflow feature which wasn't supporting delaying cron workflows. There was a customer ask to see if we could delay starting cron workflows too; this PR does that. 

**Why?**

**How did you test it?**
- TestCreateHistoryStartWorkflowRequest_DelayStart
- TestCreateHistoryStartWorkflowRequest_DelayStartWithCron

```
./cadence --do samples-domain workflow start --tl cronGroup --wt main.sampleCronWorkflow --et 60 -i '"cadence"' --delay_start_seconds 5 --cron "@every 10s"
```
Verified that the first run starts after 15s and every subsequent one fires after 10 seconds

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -103,7 +103,6 @@ var (
 	ErrContextTimeoutTooShort = &types.BadRequestError{Message: "Context timeout is too short."}
 	// ErrContextTimeoutNotSet is error for not setting a context timeout when calling a long poll API
 	ErrContextTimeoutNotSet = &types.BadRequestError{Message: "Context timeout is not set."}
-	ErrDelayStartSeconds    = &types.BadRequestError{Message: "Conflicting inputs: both DelayStartSeconds and Cron schedule is set"}
 )
 
 // AwaitWaitGroup calls Wait on the given wait
@@ -428,20 +427,22 @@ func CreateHistoryStartWorkflowRequest(
 	domainID string,
 	startRequest *types.StartWorkflowExecutionRequest,
 	now time.Time,
-) (*types.HistoryStartWorkflowExecutionRequest, error) {
+) *types.HistoryStartWorkflowExecutionRequest {
 	histRequest := &types.HistoryStartWorkflowExecutionRequest{
 		DomainUUID:   domainID,
 		StartRequest: startRequest,
 	}
 
-	firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(
-		startRequest.GetCronSchedule(), now, now)
 	delayStartSeconds := startRequest.GetDelayStartSeconds()
-	if delayStartSeconds > 0 && firstDecisionTaskBackoffSeconds > 0 {
-		return nil, ErrDelayStartSeconds
-	}
-	if delayStartSeconds > 0 {
-		firstDecisionTaskBackoffSeconds = delayStartSeconds
+	firstDecisionTaskBackoffSeconds := delayStartSeconds
+	if len(startRequest.GetCronSchedule()) > 0 {
+		delayedStartTime := now.Add(time.Second * time.Duration(delayStartSeconds))
+		firstDecisionTaskBackoffSeconds = backoff.GetBackoffForNextScheduleInSeconds(
+			startRequest.GetCronSchedule(), delayedStartTime, delayedStartTime)
+
+		// backoff seconds was calculated based on delayed start time, so we need to
+		// add the delayStartSeconds to that backoff.
+		firstDecisionTaskBackoffSeconds += delayStartSeconds
 	}
 
 	histRequest.FirstDecisionTaskBackoffSeconds = Int32Ptr(firstDecisionTaskBackoffSeconds)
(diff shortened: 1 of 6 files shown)


common/util.go
```diff
@@ -437,11 +436,8 @@ func CreateHistoryStartWorkflowRequest(
 	firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(
 		startRequest.GetCronSchedule(), now, now)
 	delayStartSeconds := startRequest.GetDelayStartSeconds()
-	if delayStartSeconds > 0 && firstDecisionTaskBackoffSeconds > 0 {
-		return nil, ErrDelayStartSeconds
-	}
 	if delayStartSeconds > 0 {
-		firstDecisionTaskBackoffSeconds = delayStartSeconds
+		firstDecisionTaskBackoffSeconds += delayStartSeconds
```
demirkayaender, line 440: Do we add this delay for the very first cron run or for every scheduled cron runs? 

Let's say if the Cron is running every hour and they wanted to start that workflow in 10 minutes expected delays for each cron runs should be
1st run: delayStartSeconds + BackoffForNextScheduleInSeconds
2nd run: 1hr
3rd run: 1hr
...

NOT
1st run: delayStartSeconds + BackoffForNextScheduleInSeconds
2nd run: delayStartSeconds + 1hr
3rd run: delayStartSeconds + 1hr
...

Could someone verify that change here isn't going to cause that?

just-at-uber: Have you tried running a test locally with the above changes and observed when the workflows kick off? You should have the workflow started time for each run so you can manually check if it's working as expected.

common/util.go
```diff
@@ -434,14 +433,16 @@ func CreateHistoryStartWorkflowRequest(
 		StartRequest: startRequest,
 	}
 
-	firstDecisionTaskBackoffSeconds := backoff.GetBackoffForNextScheduleInSeconds(
-		startRequest.GetCronSchedule(), now, now)
 	delayStartSeconds := startRequest.GetDelayStartSeconds()
-	if delayStartSeconds > 0 && firstDecisionTaskBackoffSeconds > 0 {
-		return nil, ErrDelayStartSeconds
```
yux0, line 441: We may update the function signature to only return the request. Error is not needed?

common/util.go
```diff
@@ -434,14 +433,16 @@ func CreateHistoryStartWorkflowRequest(
...
-		firstDecisionTaskBackoffSeconds = delayStartSeconds
+	firstDecisionTaskBackoffSeconds := delayStartSeconds
+	if len(startRequest.GetCronSchedule()) > 0 {
+		delayedStartTime := now.Add(time.Second * time.Duration(delayStartSeconds))
+		firstDecisionTaskBackoffSeconds = backoff.GetBackoffForNextScheduleInSeconds(
+			startRequest.GetCronSchedule(), delayedStartTime, delayedStartTime)
+
+		// backoff seconds was calculated based on delayed start time, so we need to
+		// add the delayStartSeconds to that backoff.
+		firstDecisionTaskBackoffSeconds += delayStartSeconds
```
yux0, line 445: My understanding is the cron is calculated based on the now+delay time. Then we don't need to add this back?

yux0: I think we need to define the delay + cron behavior is:
1. Delay then calculate cron 
2. Calculate the cron based on delay start time

common/util.go
demirkayaender: Thanks, tested with this and verified that first one is the behavior. Updated the test plan in the PR too

common/util.go
demirkayaender: > My understanding is the cron is calculated based on the now+delay time. Then we don't need to add this back?

I tried to put a comment to explain this but let me know how I should update this. Imagine we want to fire the cron scheduler every 10 seconds, want to delay 3 seconds and we are at 0th second.

- backoffSeconds are calculated based on 0+3 seconds so it returns 7 seconds
- we still need to add 3 seconds so we can start at the right time. 

Let me know if this makes sense.

common/util.go
demirkayaender: With this PR, user can set both delayStart and CronSchedule. Previously they were treated as conflicting params. When do you think we should return an error?

just-at-uber (approved):

common/util.go
yux0: I see. It makes sense to me now.

common/util.go
yux0: Yeah. I think after it supports both delayStart and CronSchedule. It doesn't need to return an error. Feel free to do it in this PR or in a different PR.

yux0 (approved):

common/util.go
demirkayaender: Oh I see, what you mean now, I will update the signature

PR #4198, opened 2021-05-14
Update admin DB commands to support SQL databases

Shaddoll:
**What changed?**
Update admin DB commands to support SQL databases

**Why?**
To address https://github.com/uber/cadence/issues/2777

**How did you test it?**
local verification

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/adminDBScanCommand.go
+++ b/tools/cli/adminDBScanCommand.go
@@ -125,29 +125,14 @@ func checkExecution(
 	invariants []executions.InvariantFactory,
 	fetcher executions.ExecutionFetcher,
 ) (interface{}, invariant.ManagerCheckResult) {
-	client, session := connectToCassandra(c)
-	defer session.Close()
-	logger := loggerimpl.NewNopLogger()
+	execManager := initializeExecutionStore(c, common.WorkflowIDToHistoryShard(req.WorkflowID, numberOfShards), 0)
+	defer execManager.Close()
 
-	execStore, err := cassandra.NewWorkflowExecutionPersistence(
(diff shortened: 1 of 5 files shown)


tools/cli/adminDBScanCommand.go
```diff
@@ -125,29 +125,14 @@ func checkExecution(
 	invariants []executions.InvariantFactory,
 	fetcher executions.ExecutionFetcher,
 ) (interface{}, invariant.ManagerCheckResult) {
-	client, session := connectToCassandra(c)
-	defer session.Close()
-	logger := loggerimpl.NewNopLogger()
+	execManager := initializeExecutionStore(c, common.WorkflowIDToHistoryShard(req.WorkflowID, numberOfShards), 0)
```
yycptt, line 128: not related to your change but looks like we can do the initialization only once for the db scan/clean command (also also provide an rps limit) instead of creating a new session for each workflow record. Would you mind creating an issue for it?

yycptt (commented):
Is there any plan for supporting sql database for other admin commands like `adm shard describe`, `adm workflow show`, etc? Currently those commands have the option to specify sql as the db but I think the actual impl will only attempt to connect to cassandra.

yycptt (approved):

Shaddoll:
> Is there any plan for supporting sql database for other admin commands like `adm shard describe`, `adm workflow show`, etc? Currently those commands have the option to specify sql as the db but I think the actual impl will only attempt to connect to cassandra.

Thanks for mentioning those commands. I don't know those commands are also communicating with databases. We definitely need them for docstore. I'll create tickets for that.

PR #4202, opened 2021-05-17
Use frontend client in parentClosePolicy workflow to allow auto-forwarding

yycptt:
**What changed?**
Use frontend client in parentClosePolicy workflow

**Why?**
If workflow domain failovered to another cluster after parent close policy workflow receives the request, the request will keeping retrying as the domain is no-long active in the current cluster and there's no auto-forwarding for history client, which is currently used.

Switch to frontend client so that the request can be auto-forwarded to the new active cluster.

**How did you test it?**
Tested locally with samples

**Potential risks**
In the worst case, parent close policy workflow may stop working.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/parentclosepolicy/workflow.go
+++ b/service/worker/parentclosepolicy/workflow.go
@@ -103,7 +103,6 @@ func ProcessorWorkflow(ctx workflow.Context) error {
 
 // ProcessorActivity is activity for processing batch operation
 func ProcessorActivity(ctx context.Context, request Request) error {
-	time.Sleep(2 * time.Minute)
 	processor := ctx.Value(processorContextKey).(*Processor)
 	for _, execution := range request.Executions {
 		var err error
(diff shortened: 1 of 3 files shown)


service/worker/parentclosepolicy/workflow.go
```diff
@@ -103,38 +103,32 @@ func ProcessorWorkflow(ctx workflow.Context) error {
 
 // ProcessorActivity is activity for processing batch operation
 func ProcessorActivity(ctx context.Context, request Request) error {
+	time.Sleep(2 * time.Minute)
```
yux0, line 106: why sleeps 2 minutes?

yux0 (approved):

service/worker/parentclosepolicy/workflow.go
yycptt: ah, sorry I forgot to delete the test code. Removed.

longquanzheng (approved):

PR #4206, opened 2021-05-19
Shift pointer conversions down to thrift encoder

Shaddoll:
**What changed?**
Refactored persistence serialization to shift pointer conversions down to thrift encoder.

**Why?**
Some encoders don't need to shift pointers, so original pointer conversion in common path for all encoders will be shifted back if the encoder don't need to shift pointers.

**How did you test it?**
unit test. 

**Potential risks**
For cassandra, there is no risk. For SQL, we need to make sure the read path is compatible with old data.

**Release notes**
For cassandra, there is nothing to be worried.

**Documentation Changes**

### DIFF
--- a/common/persistence/serialization/interfaces.go
+++ b/common/persistence/serialization/interfaces.go
@@ -35,267 +35,267 @@ import (
 type (
 	// ShardInfo blob in a serialization agnostic format
 	ShardInfo struct {
-		StolenSinceRenew                      *int32
-		UpdatedAt                             *time.Time
-		ReplicationAckLevel                   *int64
-		TransferAckLevel                      *int64
-		TimerAckLevel                         *time.Time
-		DomainNotificationVersion             *int64
+		StolenSinceRenew                      int32
+		UpdatedAt                             time.Time
+		ReplicationAckLevel                   int64
+		TransferAckLevel                      int64
+		TimerAckLevel                         time.Time
+		DomainNotificationVersion             int64
 		ClusterTransferAckLevel               map[string]int64
 		ClusterTimerAckLevel                  map[string]time.Time
-		Owner                                 *string
+		Owner                                 string
 		ClusterReplicationLevel               map[string]int64
 		PendingFailoverMarkers                []byte
-		PendingFailoverMarkersEncoding        *string
+		PendingFailoverMarkersEncoding        string
 		ReplicationDlqAckLevel                map[string]int64
 		TransferProcessingQueueStates         []byte
-		TransferProcessingQueueStatesEncoding *string
+		TransferProcessingQueueStatesEncoding string
 		TimerProcessingQueueStates            []byte
-		TimerProcessingQueueStatesEncoding    *string
+		TimerProcessingQueueStatesEncoding    string
 	}
 
 	// DomainInfo blob in a serialization agnostic format
 	DomainInfo struct {
-		Name                        *string
-		Description                 *string
-		Owner                       *string
-		Status                      *int32
-		Retention                   *time.Duration
-		EmitMetric                  *bool
-		ArchivalBucket              *string
-		ArchivalStatus              *int16
-		ConfigVersion               *int64
-		NotificationVersion         *int64
-		FailoverNotificationVersion *int64
-		FailoverVersion             *int64
-		ActiveClusterName           *string
+		Name                        string // TODO: This field seems not to be required. We already store domain name in another column.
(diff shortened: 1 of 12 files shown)


common/persistence/serialization/interfaces.go
```diff
@@ -35,267 +35,267 @@ import (
...
-		HistoryArchivalStatus       *int16
-		HistoryArchivalURI          *string
-		VisibilityArchivalStatus    *int16
-		VisibilityArchivalURI       *string
+		BadBinariesEncoding         string
+		HistoryArchivalStatus       int16
+		HistoryArchivalURI          string
+		VisibilityArchivalStatus    int16
+		VisibilityArchivalURI       string
 		FailoverEndTimestamp        *time.Time
```
Shaddoll, line 80: It's not updated because we have logic checking if the pointer is `nil`.

common/persistence/serialization/interfaces.go
```diff
@@ -35,267 +35,267 @@ import (
...
 
 	// WorkflowExecutionInfo blob in a serialization agnostic format
 	WorkflowExecutionInfo struct {
 		ParentDomainID                     UUID
-		ParentWorkflowID                   *string
+		ParentWorkflowID                   string
 		ParentRunID                        UUID
-		InitiatedID                        *int64
+		InitiatedID                        int64
 		CompletionEventBatchID             *int64
```
Shaddoll, line 105: This is not updated because of backward compatibility issue. Currently if this field is set to nil in thrift, it will be converted to nil in this interface, and later it will be converted to `common.EmptyEventID = -23`, which is not a zero value.
We can move the `common.EmptyEventID` logic to thrift mapper, but I don't think it's a good practice. We'll revisit it later.

common/persistence/serialization/interfaces.go
```diff
@@ -35,267 +35,267 @@ import (
...
+		WorkflowTimeout                    time.Duration
+		DecisionTaskTimeout                time.Duration
 		ExecutionContext                   []byte
-		State                              *int32
-		CloseStatus                        *int32
-		StartVersion                       *int64
+		State                              int32
+		CloseStatus                        int32
+		StartVersion                       int64
 		LastWriteEventID                   *int64
```
Shaddoll, line 117: We have logic checking if `LastWriteEventID != nil`

common/persistence/serialization/interfaces.go
```diff
@@ -35,267 +35,267 @@ import (
...
 
 	// TimerTaskInfo blob in a serialization agnostic format
 	TimerTaskInfo struct {
 		DomainID        UUID
-		WorkflowID      *string
+		WorkflowID      string
 		RunID           UUID
-		TaskType        *int16
+		TaskType        int16
 		TimeoutType     *int16
```
Shaddoll, line 278: The default value for `TimeoutType` doesn't make sense. For example, user timers previously have this field set to `nil`, but if we change it back to int16, the default value doesn't make sense to user timers. 
Ideally, for all enum types, the default values should be an invalid value.

common/persistence/serialization/thrift_mapper.go
```diff
@@ -36,22 +36,22 @@ func shardInfoToThrift(info *ShardInfo) *sqlblobs.ShardInfo {
...
 		ReplicationDlqAckLevel:                info.ReplicationDlqAckLevel,
 		TransferProcessingQueueStates:         info.TransferProcessingQueueStates,
-		TransferProcessingQueueStatesEncoding: info.TransferProcessingQueueStatesEncoding,
+		TransferProcessingQueueStatesEncoding: &info.TransferProcessingQueueStatesEncoding,
 		TimerProcessingQueueStates:            info.TimerProcessingQueueStates,
-		TimerProcessingQueueStatesEncoding:    info.TimerProcessingQueueStatesEncoding,
-		UpdatedAtNanos:                        unixNanoPtr(info.UpdatedAt),
-		TimerAckLevelNanos:                    unixNanoPtr(info.TimerAckLevel),
+		TimerProcessingQueueStatesEncoding:    &info.TimerProcessingQueueStatesEncoding,
+		UpdatedAtNanos:                        common.Int64Ptr(info.UpdatedAt.UnixNano()),
```
vytautas-karpavicius, line 53: Any reason we don't use conversion helpers for time related fields like in many places below?

vytautas-karpavicius (approved):

PR #4208, opened 2021-05-20
do not use init() when registering scavenger workflow

mantas-sidlauskas:
**What changed?**
Scavenger workflows are not registered using `init()` anymore

Other changes:
* Moved activity and workflow code to appropriate packages
* Dynamic collection is now passed to config as param, reused in history scavenger
* Context code used for background activities is removed

**Why?**
 Global workflow registration methods are deprecated

**How did you test it?**
Unit tests


service/worker/scanner/scanner.go
```diff
@@ -129,26 +123,28 @@ func (s *Scanner) Start() error {
...
-			ctx = s.startScanner(
-				ctx,
-				tlScannerWFStartOptions,
-				tlScannerWFTypeName)
-			workerTaskListNames = append(workerTaskListNames, tlScannerTaskListName)
+	if s.cfg.Persistence.DefaultStoreType() == config.StoreTypeSQL {
+		if s.cfg.TaskListScannerEnabled() {
+			w := s.newWorker(tasklist.WFStartOptions.TaskList)
+			tasklist.RegisterWorkflow(w, s.resource, &s.cfg.TaskListScannerOptions)
+			if err := w.Start(); err != nil {
```
Shaddoll, line 135: The worker isn't stopped.

service/worker/scanner/history/workflow.go
```diff
@@ -0,0 +1,127 @@
...
+// Activities is used to pass resources to activities
+type Activities struct {
+	resource resource.Resource
+	// scannerPersistenceMaxQPS the max rate of calls to persistence
+	scannerPersistenceMaxQPS   dynamicconfig.IntPropertyFn
+	maxWorkflowRetentionInDays dynamicconfig.IntPropertyFn
+}
+
+// RegisterWorkflow registers workflow and activity with worker
+func RegisterWorkflow(w worker.Worker, resource resource.Resource, dc *dynamicconfig.Collection) {
```
Shaddoll, line 80: nit: Instead of passing `dc` to this function, you can define `HistoryScannerOptions` like `TaskListScannerOptions`, so that we don't need to change the signature of `worker.NewConfig`

PR #4212, opened 2021-05-21
Fix update domain replication ack level

yux0:
**What changed?**
1. Pass remote cluster name into get domain replication request
2. Check ack level before updating the ack level to make sure it will not decrease
3. Simplify domain replication queue cleanup

**Why?**
1. Pass remote cluster name into get domain replication request
2. Check ack level before updating the ack level to make sure it will not decrease

**How did you test it?**
Unit tests and integration tests

**Potential risks**
It may impact the data in queue metadata table. A one time cleanup might be needed

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/queueManager.go
+++ b/common/persistence/queueManager.go
@@ -68,20 +68,7 @@ func (q *queueManager) DeleteMessagesBefore(ctx context.Context, messageID int64
 }
 
 func (q *queueManager) UpdateAckLevel(ctx context.Context, messageID int64, clusterName string) error {
-	ackLevels, err := q.persistence.GetAckLevels(ctx)
-	if err != nil {
-		return err
-	}
-	if ackLevel, ok := ackLevels[clusterName]; ok && messageID > ackLevel {
-		// Update if the ack level moves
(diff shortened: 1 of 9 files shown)


common/persistence/queueManager.go
```diff
@@ -68,7 +68,20 @@ func (q *queueManager) DeleteMessagesBefore(ctx context.Context, messageID int64
 }
 
 func (q *queueManager) UpdateAckLevel(ctx context.Context, messageID int64, clusterName string) error {
-	return q.persistence.UpdateAckLevel(ctx, messageID, clusterName)
+	ackLevels, err := q.persistence.GetAckLevels(ctx)
```
Shaddoll, line 71: Do we need to do it in a transaction?

yux0: Yes. Let me update this one

yux0: The persistence layer already does transaction updates. I move this verification logic one layer down.

yycptt (approved):

PR #4213, opened 2021-05-21
Wire up rebalance CLI

yux0:
**What changed?**
Wire up rebalance CLI

**Why?**
Use CLI to start and list rebalance workflow

**How did you test it?**
local test

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**

### DIFF
--- a/tools/cli/adminClusterCommands.go
+++ b/tools/cli/adminClusterCommands.go
@@ -110,7 +110,9 @@ func AdminRebalanceStart(c *cli.Context) {
 }
 
 func AdminRebalanceList(c *cli.Context) {
-
+	c.Set(FlagWorkflowID, failovermanager.RebalanceWorkflowID)
+	c.GlobalSet(FlagDomain, common.SystemLocalDomainName)
+	ListWorkflow(c)
 }
 
 func intValTypeToString(valType int) string {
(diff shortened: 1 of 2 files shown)


tools/cli/adminClusterCommands.go
```diff
@@ -110,7 +110,9 @@ func AdminRebalanceStart(c *cli.Context) {
 }
 
 func AdminRebalanceList(c *cli.Context) {
-
+	c.Set(FlagWorkflowID, failovermanager.RebalanceWorkflowID)
+	c.GlobalSet(FlagDomain, common.SystemLocalDomainName)
```
yycptt, line 114: nit: not sure if we want to increase the retention period of the system so that the history for managed failover or rebalancing workflow can be stored for a longer period of time.

yycptt (approved):

tools/cli/adminClusterCommands.go
yux0: Good point. I will create a task to track this.

yux0: it is already 30 days.

yycptt: Interesting. Maybe we manually updated it internally. The retention specified in the schema is 3 days I think. 
See: https://github.com/uber/cadence/blob/master/schema/cassandra/cadence/schema.cql#L447

PR #4224, opened 2021-05-26
Add Admin Shard List Command

demirkayaender:
**What changed?**
With these changes we can see the shard distribution for history or for other shard roles

See IDL repo changes: https://github.com/uber/cadence-idl/pull/70

**Why?**

**How did you test it?**
```
enderd[05/25 23:35]:~/cadence$ ./cadence --do samples-domain admin shard list --role all
{
  "numberOfShards": 4,
  "shardIDs": {
    "0": "cadence-history,127.0.0.1:7934",
    "1": "cadence-history,127.0.0.1:7934",
    "2": "cadence-history,127.0.0.1:7934",
    "3": "cadence-history,127.0.0.1:7934"
  }
}
enderd[05/25 23:37]:~/cadence$ ./cadence --do samples-domain admin shard list --role cadence-history2
{}
enderd[05/25 23:37]:~/cadence$ ./cadence --do samples-domain admin shard list --role cadence-history
{
  "numberOfShards": 4,
  "shardIDs": {
    "0": "cadence-history,127.0.0.1:7934",
    "1": "cadence-history,127.0.0.1:7934",
    "2": "cadence-history,127.0.0.1:7934",
    "3": "cadence-history,127.0.0.1:7934"
  }
}

// testing pagination
enderd[05/26 10:13]:~/cadence$ ./cadence --do samples-domain admin shard list --pagesize 2 --page_id 2
{
  "numberOfShards": 4
}
enderd[05/26 10:13]:~/cadence$ ./cadence --do samples-domain admin shard list --pagesize 2 --page_id 1
{
  "numberOfShards": 4,
  "shardIDs": {
    "2": "cadence-history,127.0.0.1:7934",
    "3": "cadence-history,127.0.0.1:7934"
  }
}
enderd[05/26 10:13]:~/cadence$ ./cadence --do samples-domain admin shard list --pagesize 2 --page_id 0
{
  "numberOfShards": 4,
  "shardIDs": {
    "0": "cadence-history,127.0.0.1:7934",
    "1": "cadence-history,127.0.0.1:7934"
  }
}
enderd[05/26 10:13]:~/cadence$ ./cadence --do samples-domain admin shard list --pagesize 3 --page_id 1
{
  "numberOfShards": 4,
  "shardIDs": {
    "3": "cadence-history,127.0.0.1:7934"
  }
}

// help
enderd[05/26 10:16]:~/cadence$ ./cadence --do samples-domain admin shard list --help
NAME:
   cadence admin shard list - List shard distribution

USAGE:
   cadence admin shard list [command options] [arguments...]

OPTIONS:
   --role value      Role of the shard (default: "cadence-history")
   --pagesize value  Max number of results to return (default: 100)
   --page_id value   Option to show results offset from pagesize * page_id (default: 0)
```

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/handler.go
```diff
@@ -703,6 +704,42 @@ func (h *handlerImpl) StartWorkflowExecution(
...
+	// count all the shards but put only the ones falling into the requested page
+	addShardToResponse := func(shardID int, identity string) {
+		if resp.NumberOfShards >= offset && len(resp.Shards) < int(request.PageSize) {
+			resp.Shards[int32(shardID)] = identity
+		}
+		resp.NumberOfShards++
+	}
+
+	for shardID := 0; shardID < numShards; shardID++ {
+		info, err := h.GetHistoryServiceResolver().Lookup(string(rune(shardID)))
```
yycptt, line 730: I think both `HistoryServiceResolver` and `numberOfHistoryShards` (number of shards is in the static yaml config) are available in the frontend. So it looks like we can handle the request directly in frontend instead of forwarding it to history service.

tools/cli/admin.go
```diff
@@ -139,6 +139,31 @@ func newAdminShardManagementCommands() []cli.Command {
...
 		},
+		{
+			Name:    "list",
+			Aliases: []string{"l"},
+			Usage:   "List shard distribution",
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:  FlagRole,
+					Value: "cadence-history",
+					Usage: "Role of the shard",
```
yycptt, line 150: I am little bit confused by the meaning of role here. Is `cadence-history` the only accepted value or we plan to add more like the tasklist distribution for matching?

service/history/handler.go
```diff
@@ -703,6 +704,42 @@ func (h *handlerImpl) StartWorkflowExecution(
...
+	offset := request.PageID * request.PageSize
+	// count all the shards but put only the ones falling into the requested page
+	addShardToResponse := func(shardID int, identity string) {
+		if resp.NumberOfShards >= offset && len(resp.Shards) < int(request.PageSize) {
+			resp.Shards[int32(shardID)] = identity
+		}
+		resp.NumberOfShards++
+	}
+
+	for shardID := 0; shardID < numShards; shardID++ {
```
yycptt, line 729: Looks like we can just look up shardIDs falling into the request page since the total number of shards will never change?

tools/cli/admin.go
demirkayaender: Value here is the default value, you can pass any value. I made it a parameter so we can gather other type of shards in the future if we wanted to.

service/history/handler.go
demirkayaender: Let me try and see if this works

service/history/handler.go
demirkayaender: Is this condition true: number_of_shards == any_type_of_shard ? 

If there's a filter we need to count how many shards have this role right?

service/history/handler.go
demirkayaender: The config on admin side has `NumHistoryShards`; is this just history shards. If so, do we have access to total number of shards of all types ?

demirkayaender: Talked offline this is now addressed

common/types/mapper/thrift/shared.go
```diff
@@ -1556,6 +1567,17 @@ func ToDescribeHistoryHostRequest(t *shared.DescribeHistoryHostRequest) *types.D
 	}
 }
 
+// ToDescribeShardDistributionRequest converts thrift DescribeHistoryHostRequest type to internal
+func ToDescribeShardDistributionRequest(t *shared.DescribeShardDistributionRequest) *types.DescribeShardDistributionRequest {
+	if t == nil {
+		return nil
+	}
+	return &types.DescribeShardDistributionRequest{
+		PageSize: *t.PageSize,
```
Shaddoll, line 1576: Nit: use t.GetPageSize()/t.GetPageID() instead to avoid nil pointer dereference.

Shaddoll (approved):

service/frontend/adminHandler.go
```diff
@@ -355,6 +356,35 @@ func (adh *adminHandlerImpl) DescribeQueue(
 	return adh.GetHistoryClient().DescribeQueue(ctx, request)
 }
 
+// DescribeHistoryHost returns information about the internal states of a history host
+func (adh *adminHandlerImpl) DescribeShardDistribution(
```
yycptt, line 360: nit: update the comment.

tools/cli/adminCommands.go
```diff
@@ -503,6 +503,26 @@ func AdminCloseShard(c *cli.Context) {
...
+
+	ctx, cancel := newContext(c)
+	defer cancel()
+
+	req := &types.DescribeShardDistributionRequest{
+		PageSize: int32(c.Int(FlagPageSize)),
+		PageID:   int32(c.Int(FlagPageID)),
+	}
+
+	resp, err := adminClient.DescribeShardDistribution(ctx, req)
```
yycptt, line 518: nit: not sure if it's useful to have some logic there to automatically fetch next page upon user request. 

and  if `len(resp.Shards) < req.PageSize` it means there's no more pages?

yycptt (approved):

tools/cli/adminCommands.go
demirkayaender: For the second one, maybe I can split NumberOfShards into 2 as TotalNumberOfShards and NumberOfShardsReturned; although this will require another idl change; just landed the previous one. Let me know if that sounds good or if you have any other ideas.

For the first one, we need to persist what user already saw to achieve this right? It might be better to give this control to the user. Or if there's any example, let me know.

yycptt: For the first one, what I am thinking is somewhat like our list workflow commands. The default pagesize is 10, and at the end of the each page, it will ask user if the next page should be displayed. If user hit enter then the next page will be displayed. It's kind of useful when user just want to get a sense of the shard distribution. I don't have strong opinion on it though. 

For the second one, I think the current impl is fine. I just want to confirm my understanding and see if this is something caller can rely on when trying to fetch all pages. 
Originally I was thinking we can follow the nextPageToken pattern we use for most APIs that needs pagination. But this impl also makes sense I think.

PR #4225, opened 2021-05-26
Add cross-cluster task related types and methods to data/persistence interface

yycptt:
**What changed?**
Add cross-cluster task related types and methods to data/persistence interface

**Why?**
For supporting cross-domain/cluster operation

**How did you test it?**
- Existing persistence test
- Since this pr only add methods and fields to existing struct & interface def, the fact that the code compiles also shows the correctness to some degree

**Potential risks**
N/A, the newly added fields is not used or persisted.

**Release notes**

**Documentation Changes**


common/persistence/dataInterfaces.go
```diff
@@ -152,6 +152,13 @@ const (
 	TransferTaskTypeUpsertWorkflowSearchAttributes
 )
 
+// Types of cross-cluster tasks
+const (
+	CrossClusterTaskTypeStartChildExecution = iota
```
Shaddoll, line 157: nit: it's good to have something like `CrossClusterTaskTypeInvalid` as default value, so that when we forget to set the value at some place, we'll get an explicit error.

Shaddoll (approved):

common/mocks/ExecutionManager.go
```diff
@@ -18,6 +20,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+// Code generated by mockery v2.2.1. DO NOT EDIT.
```
yux0, line 23: Is this auto generated?

common/persistence/dataInterfaces.go
yycptt: Good point. I changed the enum value to start from 1 so that if the value is unset the value will be an invalid value 0.

common/log/tag/values.go
```diff
@@ -177,21 +177,25 @@ var (
...
+	StoreOperationIsWorkflowExecutionExists        = storeOperation("is-wf-execution-exists")
+	StoreOperationListConcreteExecution            = storeOperation("list-concrete-execution")
+	StoreOperationGetTransferTasks                 = storeOperation("get-transfer-tasks")
+	StoreOperationGetCrossClusterTasks             = storeOperation("get-cross-cluster-tasks")
+	StoreOperationGetReplicationTasks              = storeOperation("get-replication-tasks")
+	StoreOperationCompleteTransferTask             = storeOperation("complete-transfer-task")
+	StoreOperationRangeCompleteTransferTask        = storeOperation("range-complete-transfer-task")
+	StoreOperationCompleteCrossClusterTask         = storeOperation("complete-cross-cluster-task")
+	StoreOperationRangeCompleteCrossClusterTask    = storeOperation("range-complete-cross-cluster-task")
+
```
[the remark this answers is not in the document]
yycptt: Removed. Thanks for catching.

common/mocks/ExecutionManager.go
yycptt: Yes, it's auto-generated by mockery.

yux0 (approved):

Issue #2777, opened 2019-11-06
Support different database for Admin CLI


longquanzheng:
right now we only support Cassandra, we need another parameter to let customer to chose different persistence type.

PR #4226, opened 2021-05-26
Add conn_attrs flag to admin db commands

Shaddoll:
**What changed?**

- Introduce StringMap type to parse `key=val` CLI input into string map
- Add `conn_attrs` flag to admin db commands and use it for SQL database connection

**Why?**
To address #2777. Docstore will use `conn_attrs` for database connection.

**How did you test it?**
local test, new unit test

```
zijian@zijian-C02XW43AJGH6 cli % ./cli adm shard d --conn_attrs "a=b" --shard_id 0 --db_port 9042 --conn_attrs cd --db_type mysql 
Incorrect Usage: invalid value "cd" for flag -conn_attrs: should be in 'key=value' format

NAME:
   cadence admin shard describe - Describe shard by Id

USAGE:
   cadence admin shard describe [command options] [arguments...]

OPTIONS:
   --db_type value                 persistence type. Current supported options are [mysql postgres cassandra] (default: "cassandra")
   --db_address value              persistence address (right now only cassandra is fully supported) (default: "127.0.0.1")
   --db_port value                 persistence port (default: 9042)
   --db_region value               persistence region
   --username value                persistence username
   --password value                persistence password
   --keyspace value                cassandra keyspace (default: "cadence")
   --db_name value                 sql database name (default: "cadence")
   --encoding_type value           sql database encoding type (default: "thriftrw")
   --decoding_types value          sql database decoding types (default: "thriftrw")
   --tls                           enable TLS over cassandra connection
   --tls_cert_path value           cassandra tls client cert path (tls must be enabled)
   --tls_key_path value            cassandra tls client key path (tls must be enabled)
   --tls_ca_path value             cassandra tls client ca path (tls must be enabled)
   --tls_enable_host_verification  cassandra tls verify hostname and server cert (tls must be enabled)
   --conn_attrs value              a key-value set of sql database connection attributes (must be in key=value format) (default: map[a:b])
   --shard_id value                The Id of the shard to describe (default: 0)
   
invalid value "cd" for flag -conn_attrs: should be in 'key=value' format
```

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/admin.go
+++ b/tools/cli/admin.go
@@ -31,6 +31,7 @@ import (
 	"github.com/uber/cadence/common/persistence/sql"
 	"github.com/uber/cadence/common/reconciliation/invariant"
 	"github.com/uber/cadence/service/worker/scanner/executions"
+	"github.com/uber/cadence/tools/common/flag"
 )
 
 func newAdminWorkflowCommands() []cli.Command {
@@ -1001,6 +1002,11 @@ func getDBFlags() []cli.Flag {
 		cli.BoolFlag{
 			Name:  FlagTLSEnableHostVerification,
 			Usage: "cassandra tls verify hostname and server cert (tls must be enabled)",
+		},
+		cli.GenericFlag{
+			Name:  FlagConnectionAttributes,
+			Usage: "a key-value set of sql database connection attributes (must be in key1=value1,key2=value2,...,keyN=valueN format, e.g. cluster=dca or cluster=dca,instance=cadence)",
+			Value: &flag.StringMap{},
 		},
 	}
 }
--- a/tools/cli/adminCommands.go
+++ b/tools/cli/adminCommands.go
@@ -43,6 +43,7 @@ import (
 	"github.com/uber/cadence/common/persistence/sql/sqlplugin"
 	"github.com/uber/cadence/common/types"
 	"github.com/uber/cadence/common/types/mapper/thrift"
+	"github.com/uber/cadence/tools/common/flag"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


tools/cli/admin.go
```diff
@@ -1002,6 +1003,11 @@ func getDBFlags() []cli.Flag {
 			Name:  FlagTLSEnableHostVerification,
 			Usage: "cassandra tls verify hostname and server cert (tls must be enabled)",
 		},
+		cli.GenericFlag{
+			Name:  FlagConnectionAttributes,
+			Usage: "a key-value set of sql database connection attributes (must be in key=value format)",
```
longquanzheng, line 1008: what is the format for multiple KVs? Can you add an example of one KV and two KVs?

tools/cli/admin.go
```diff
@@ -1002,6 +1003,11 @@ func getDBFlags() []cli.Flag {
...
+			Value: &flag.StringMap{},
```
longquanzheng, line 1009: Nit: what do you think about keeping consistent with the sql-tool?
https://github.com/uber/cadence/blob/d5fbc5e3b9c978f7ba9b1c93cb162e261df6b650/tools/sql/handler.go#L171
The format is URL encoded: `k1=v1&k2=v2`

longquanzheng (approved):
Just minor comments

tools/cli/admin.go
Shaddoll: Yes, it should be `a=b,c=d,...,x=y` or just `a=b`. I've updated the Usage.

tools/cli/admin.go
Shaddoll: I didn't realize the flag is already defined in `sql-tool`. I'll update this PR to keep them consistent with each other.

longquanzheng: Thanks! Can you also help update the usage in the sql-tool? that's also missing the details

Shaddoll: Sure, updated!

longquanzheng:
LGTM

PR #4229, opened 2021-05-27
Refactoring Cassandra task persistence manager for NoSQL support

longquanzheng:
**What changed?**
1. Create NoSQL interface for task operation
2. Extract the logic of Cassandra to implement the interface
3. Replace matchingStore with NoSQL plugin db style

**Why?**
For https://github.com/uber/cadence/issues/3514

**How did you test it?**
existing tests

**Potential risks**
Low risks

**Release notes**
No

**Documentation Changes**
No

### DIFF
--- a/common/persistence/cassandra/cassandraTaskPersistence.go
+++ b/common/persistence/cassandra/cassandraTaskPersistence.go
@@ -24,166 +24,58 @@ package cassandra
 import (
 	"context"
 	"fmt"
-	"strings"
+	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin"
 	"time"
 
 	"github.com/uber/cadence/common/config"
 	"github.com/uber/cadence/common/log"
 	p "github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra"
-	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra/gocql"
 	"github.com/uber/cadence/common/types"
 )
 
 type (
 	// Implements TaskManager
-	cassandraTaskPersistence struct {
-		cassandraStore
-		shardID            int
-		currentClusterName string
+	nosqlTaskManager struct {
+		db     nosqlplugin.DB
+		logger log.Logger
 	}
 )
 
 const (
-	// Row types for table tasks
-	rowTypeTask = iota
-	rowTypeTaskList
+	initialRangeID  = 1 // Id of the first range of a new task list
+	initialAckLevel = 0
 )
 
-const (
-	taskListTaskID = -12345
-	initialRangeID = 1 // Id of the first range of a new task list
-)
-
-const (
-	templateTaskListType = `{` +
-		`domain_id: ?, ` +
-		`name: ?, ` +
-		`type: ?, ` +
-		`ack_level: ?, ` +
-		`kind: ?, ` +
-		`last_updated: ? ` +
-		`}`
-
-	templateTaskType = `{` +
-		`domain_id: ?, ` +
-		`workflow_id: ?, ` +
-		`run_id: ?, ` +
-		`schedule_id: ?,` +
-		`created_time: ? ` +
-		`}`
-
-	templateCreateTaskQuery = `INSERT INTO tasks (` +
-		`domain_id, task_list_name, task_list_type, type, task_id, task) ` +
-		`VALUES(?, ?, ?, ?, ?, ` + templateTaskType + `)`
-
-	templateCreateTaskWithTTLQuery = `INSERT INTO tasks (` +
-		`domain_id, task_list_name, task_list_type, type, task_id, task) ` +
-		`VALUES(?, ?, ?, ?, ?, ` + templateTaskType + `) USING TTL ?`
-
-	templateGetTasksQuery = `SELECT task_id, task ` +
-		`FROM tasks ` +
-		`WHERE domain_id = ? ` +
-		`and task_list_name = ? ` +
-		`and task_list_type = ? ` +
-		`and type = ? ` +
-		`and task_id > ? ` +
-		`and task_id <= ?`
-
-	templateCompleteTaskQuery = `DELETE FROM tasks ` +
-		`WHERE domain_id = ? ` +
-		`and task_list_name = ? ` +
-		`and task_list_type = ? ` +
-		`and type = ? ` +
-		`and task_id = ?`
-
-	templateCompleteTasksLessThanQuery = `DELETE FROM tasks ` +
-		`WHERE domain_id = ? ` +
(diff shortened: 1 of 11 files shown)


common/persistence/nosql/nosqlplugin/cassandra/tasks.go
```diff
@@ -0,0 +1,492 @@
...
+		`VALUES(?, ?, ?, ?, ?, ` + templateTaskType + `) USING TTL ?`
+
+	templateGetTasksQuery = `SELECT task_id, task ` +
+		`FROM tasks ` +
+		`WHERE domain_id = ? ` +
+		`and task_list_name = ? ` +
+		`and task_list_type = ? ` +
+		`and type = ? ` +
+		`and task_id >= ? ` +
+		`and task_id < ?`
```
yycptt, line 79: Is this intentional? It used to be (Exclusive, Inclusive].

longquanzheng: Yes, that’s intentional for consistency

common/persistence/cassandra/cassandraTaskPersistence.go
```diff
@@ -24,174 +24,68 @@ package cassandra
 import (
```
yux0, line 24: Should update this file name?

common/persistence/nosql/nosqlplugin/cassandra/constants.go
```diff
@@ -78,5 +78,5 @@ const (
 	rowTypeShardTaskID     = int64(-11)
 	emptyInitiatedID       = int64(-7)
 
-	stickyTaskListTTL = int32(24 * time.Hour / time.Second) // if sticky task_list stopped being updated, remove it in one day
+	maxCassandraTTL = int64(157680000)
```
yux0, line 81: How does this value get calculated

common/persistence/nosql/nosqlplugin/interfaces.go
```diff
@@ -203,6 +204,119 @@ type (
...
+		TaskListType int
+		TaskID       int64
+
+		WorkflowID  string
+		RunID       string
+		ScheduledID int64
+		CreatedTime time.Time
+	}
+
+	TaskRowPK struct {
```
yycptt, line 291: Can we reuse `TasksFilter`? Also not sure if `BatchSize` belongs to `TasksFilter` logically.

common/persistence/nosql/nosqlplugin/cassandra/tasks.go
longquanzheng: Yeah, maybe just let it be inconsistent. I just found that the queueCRUD also has an API using exclusive,inclusive manner. 

I will change it back.

common/persistence/cassandra/cassandraTaskPersistence.go
longquanzheng: I will update them later when we move the files to nosql package.

common/persistence/nosql/nosqlplugin/interfaces.go
longquanzheng: Yeah that's a bit tricky here. TasksFilter has `MinTaskID` which is not supported in this API. Other than this, we have several options:
1. Reuse it any way, and tell the developer to ignore `MinTaskID` in this API. 
2. Reuse it, and implement it, but we don't have tests to cover this, as the persistenceIntefaces+dataInterfaces doesn't have `MinTaskID` in rangeDelete. Unless we also change the persistenceIntefaces and dataInterfaces. We probably don't want to do this in this refactoring. 
3. Create a different taskFilter, e.g. `TasksFilterNoMin`. 

I chose this TaskRowPK because 1. it can be resued in other methods 2. It help indicating what is PrimaryKey in this table. 
Do you have any preference?

longquanzheng: I think the most elegant way is to use TasksFilter like you said. So I changed to option 2, and create an issue https://github.com/uber/cadence/issues/4243

common/persistence/cassandra/cassandraTaskPersistence.go
```diff
@@ -201,350 +95,264 @@ func (d *cassandraTaskPersistence) LeaseTaskList(
...
 		} else {
-			return nil, convertCommonErrors(d.client, "LeaseTaskList", err)
+			return nil, convertCommonErrors(t.db, "LeaseTaskList", err)
 		}
 	} else {
 		// if request.RangeID is > 0, we are trying to renew an already existing
 		// lease on the task list. If request.RangeID=0, we are trying to steal
-		// the tasklist from its current owner
-		if request.RangeID > 0 && request.RangeID != rangeID {
+		// the currTL from its current owner
```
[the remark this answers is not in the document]
longquanzheng: yeah good catch :D

yycptt (approved):

common/persistence/cassandra/cassandraTaskPersistence.go
```diff
@@ -201,350 +95,258 @@ func (d *cassandraTaskPersistence) LeaseTaskList(
...
+
+	if err != nil {
+		return nil, convertCommonErrors(t.db, "GetTasks", err)
 	}
 
-	if err := iter.Close(); err != nil {
-		return nil, convertCommonErrors(d.client, "GetTasks", err)
+	response := &p.InternalGetTasksResponse{}
+	for _, t := range resp {
+		response.Tasks = append(response.Tasks, toTaskInfo(t))
```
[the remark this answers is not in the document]
longquanzheng: >Some random thoughts: looks like we can just have one rangeDelete API in the plugin interface so that development don't need to implement both delete and rangeDelete?

Ha, that's a genius idea! Will do it. 

>  there's no nextPageToken in the response even when batchSize in the request is less than the number of tasks in the requested range
yeah, the way it works it that in tasklist scanner, the activity will run again and delete next page.

common/persistence/nosql/nosqlplugin/interfaces.go
yycptt: I am not sure if there's any concern from performance side when using rangeDelete to delete a single task. But we always handle the special case of deleting only one task with a point delete query in the plugin implementation.

yycptt (commented):
Please also add no-op impl for dynamoDB.

common/persistence/nosql/nosqlplugin/interfaces.go
longquanzheng: It depends on the DB implementation, but in theory this should not be a concern. 
Also, this CompleteTask data API is only being used by orphan task scanner, which should be very rarely used.

longquanzheng:
> Please also add no-op impl for dynamoDB.

Yeah, also added test for ddb plugin.

PR #4232, opened 2021-05-27
Remove transfer/timer failover queue level from persistence layer

yycptt:
**What changed?**
- Remove transfer/timer failover queue level from persistence layer
- Don't update/persist shard info when failover level is updated since it's only used in-memory

**Why?**
- transfer/timer failover queue level is not persisted.

**How did you test it?**
- Integration test

**Potential risks**
N/A, the field is never persisted to DB.

**Release notes**

**Documentation Changes**


Shaddoll (approved):

yux0:
LGTM

PR #4241, opened 2021-05-29
Refactor: not require db visibility when ES visibility is provided

longquanzheng:
**What changed?**
1. Remove the requirement of db visibility when ES visibility is provided
2. Move dynamic config to dynamicconfig package 

**Why?**
1. This will make implementing nosql plugin much easier. It's unecessary to implement basic visibility with nosql as ES is much better
2. For consistency

**How did you test it?**
existing tests

**Potential risks**
Low

**Release notes**
Announce this change in release to the community

**Documentation Changes**
https://github.com/uber/cadence-docs/issues/51


common/config/persistence.go
```diff
@@ -39,8 +39,20 @@ func (c *Persistence) DefaultStoreType() string {
 
 // Validate validates the persistence config
 func (c *Persistence) Validate() error {
-	stores := []string{c.DefaultStore, c.VisibilityStore}
-	for _, st := range stores {
+	if _, ok := c.DataStores[c.DefaultStore]; !ok {
```
Shaddoll, line 42: nit: line 42 - 44 is already covered in the for loop

longquanzheng: Removed. Good catch, thanks!

common/persistence/client/factory.go
```diff
@@ -246,13 +249,81 @@ func (f *factoryImpl) NewExecutionManager(shardID int) (p.ExecutionManager, erro
...
+		visibilityFromDB, err = f.newDBVisibilityManager(resourceConfig)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if params.PersistenceConfig.AdvancedVisibilityStore != "" {
+		visibilityIndexName := params.ESConfig.Indices[common.VisibilityAppName]
+		visibilityProducer, err := params.MessagingClient.NewProducer(common.VisibilityAppName)
+		if err != nil {
+			f.logger.Fatal("Creating visibility producer failed", tag.Error(err))
```
Shaddoll, line 272: why not return error but log fatal?

common/dynamicconfig/resource.go
```diff
@@ -0,0 +1,47 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package dynamicconfig
+
+type (
+	// ResourceConfig is a subset of the service dynamic config
+	ResourceConfig struct {
```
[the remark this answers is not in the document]
longquanzheng: I did that initially, but that would cause a circular dependency issue. We can also have a separate package for this: common/resource/config package.

common/persistence/client/factory.go
longquanzheng: It was copied it from https://github.com/uber/cadence/blob/723ecf572b2391ad3acc93c455445d9968170eef/service/history/service.go#L83 I am fine with either way. Returning error should result in the same behavior.

longquanzheng: Lots of our code is using fatal.
Some thoughts: this fatal would result in a panic and even the caller ignores the error, the startup would still fail. Also, this can report a "Fatal" log type instead of "error". which is more obvious for us to catch(since we may have many other errors). And this one single line is good for :
```
logger.Error(...)
return err
```

Do you feel okay to keep Fatal?

common/persistence/visibilityDualManager.go
```diff
@@ -0,0 +1,282 @@
...
+			return v.esVisibilityManager.RecordWorkflowExecutionClosed(ctx, request)
+		},
+	)
+}
+
+func (v *visibilityDualManager) DeleteWorkflowExecution(
+	ctx context.Context,
+	request *VisibilityDeleteWorkflowExecutionRequest,
+) error {
+	return v.chooseVisibilityManagerForWrite(
```
Shaddoll, line 115: Original behavior deletes workflows from all available storages, which is better for data cleaning. What do you think of it?

common/persistence/client/factory.go
```diff
@@ -267,8 +338,12 @@ func (f *factoryImpl) NewVisibilityManager() (p.VisibilityManager, error) {
 	if ds.ratelimit != nil {
 		result = p.NewVisibilityPersistenceRateLimitedClient(result, ds.ratelimit, f.logger)
 	}
-	if visConfig != nil && visConfig.EnableSampling() {
-		result = p.NewVisibilitySamplingClient(result, visConfig, f.metricsClient, f.logger)
+	if visibilityConfig.EnableDBVisibilitySampling() {
+		result = p.NewVisibilitySamplingClient(result, &p.SamplingConfig{
```
Shaddoll, line 342: Fields in `ResourceConfig` can be `nil`, but we are not doing any checks on the caller side or callee side.

common/persistence/client/factory.go
Shaddoll: I'm okay with it.

service/worker/service.go
```diff
@@ -94,15 +92,12 @@ func NewService(
...
-		common.WorkerServiceName,
-		serviceConfig.PersistenceMaxQPS,
-		serviceConfig.PersistenceGlobalMaxQPS,
-		serviceConfig.ThrottledLogRPS,
-		func(
-			persistenceBean persistenceClient.Bean,
-			logger log.Logger,
-		) (persistence.VisibilityManager, error) {
-			return persistenceBean.GetVisibilityManager(), nil
+		common.MatchingServiceName,
```
Shaddoll, line 95: common.WorkerServiceName

common/persistence/visibilityDualManager.go
longquanzheng: Yes I changed that. I don't see it necessary. People uses dual write mode only for migration. After the migration, one of them will be removed anyway. Changing it here will make the behavior consistent with the configuration, or under the expectation of the configuration. It would be very hard to explain/documenting to have such a exceptional behavior in deleting.

common/persistence/client/factory.go
longquanzheng: What kind of checking do you think we should have? I am open to add more. There is one checking in the factoryImpl which would return nil when both read and write mode are nil.

common/dynamicconfig/resource.go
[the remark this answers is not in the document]
longquanzheng: Moved to resource/config/ pkg

common/persistence/client/factory.go
Shaddoll: Something similar to line 324 before invoking the functions

longquanzheng: Ah I see. Added the nil check here.

Shaddoll (approved):

yycptt (approved):

PR #4242, opened 2021-06-01
Add the ability to query cron workflows

demirkayaender:
**What changed?**
Added cron filters to the list workflow calls. Changed Cassandra, MySQL and Postgres plugins so we can filter workflows based on whether they are cron or non-cron workflows. 

IDL changes: https://github.com/uber/cadence-idl/pull/72

**Why?**
So the customers can filter cron workflows in Cadence UI.

**How did you test it?**
```
go test -v github.com/uber/cadence/common/persistence/sql/sqlplugin/mysql
go test -v github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra/tests
go test -v github.com/uber/cadence/common/persistence/sql/sqlplugin/postgres
```

**Potential risks**

**Release notes**

**Documentation Changes**


common/types/mapper/thrift/shared.go
```diff
@@ -3055,6 +3062,7 @@ func ToListWorkflowExecutionsRequest(t *shared.ListWorkflowExecutionsRequest) *t
 		PageSize:      t.GetPageSize(),
 		NextPageToken: t.NextPageToken,
 		Query:         t.GetQuery(),
+		IsCron:        t.IsCron,
```
just-at-uber, line 3065: similar to IDL change, will impact how frontend is built.

common/persistence/cassandra/cassandraVisibilityPersistence.go
```diff
@@ -187,6 +187,19 @@ type (
 	}
 )
 
+func (v *cassandraVisibilityPersistence) Query(
+	request *p.InternalListWorkflowExecutionsRequest,
+	template string,
+	params ...interface{},
+) gocql.Query {
+	if request.IsCron != nil {
+		template += `AND is_cron = ? ALLOW FILTERING `
```
yycptt, line 196: I don't think we want to use `allow filtering`?

common/types/mapper/thrift/shared.go
demirkayaender: Frontend needs to set this field only when user sets it, frontend should work as is without making any changes for other cases.

yycptt (commented):
What's scope of this task? I originally thought we only need to support it for elasticsearch. 

For Cassandra I don't think we want to use `allow filtering` or adding another secondary index to support this. Not supporting it is fine as right now only a limited type of visibility queries are supported by Cassandra. 

For SQL, I guess the story is similar to Cassandra.

common/persistence/cassandra/cassandraVisibilityPersistence.go
demirkayaender: If I don't put ALLOW FILTERING, cassandra complains since this is not an indexed column. Is there another way to do this?

yycptt: Yeah, that's why I mentioned a secondary index is needed if we want to support it. We can sync during standup but my current understanding is we don't have to support it for cassandra and sql, just like we don't support querying based on tasklist when using those two DB for visibility.

longquanzheng: Non-indexed query will not scale well in Cassandra(same as MySQL/Postgres, but they don't complains).

What is the background of this? I wonder why not just implement this in ESVisibility only? We have decided to move to ES for new requirements long time ago as database doesn't have good support for searching like ElasticSearch.

longquanzheng:
This is going to conflict with my refactoring https://github.com/uber/cadence/pull/4223

Would you mind review my PR and let it go in first(because it's quite hard to resolve conflict in my refactoring PR >< 

I would really appreciate it! Thanks!

common/persistence/cassandra/cassandraVisibilityPersistence.go
longquanzheng: A little more some background, even secondary index will not perform/scale well in production. @vancexu did a comprehensive scalability test 1-2 yrs ago, then we decided to add Sampling to C* visibility https://github.com/uber/cadence/blob/master/common/persistence/visibilitySamplingClient.go 
And later on we moved to ES based visibility. 

ES visibility should be default for Cadence. You may change the internal wiki that es based docker compose is the default so that internal developer can use it by default. I am refactor visibility in https://github.com/uber/cadence/pull/4241 so that other databases will not have to implement visibility.

yycptt:
> This is going to conflict with my refactoring #4223 and #4241
> 
> Would you mind review my PR and let it go in first(because it's quite hard to resolve conflict in my refactoring PR ><
> 
> I would really appreciate it! Thanks!

I guess if only elasticsearch implementation is touched (for List/Scan/CountWorkflowExecution API), there should be no conflicts as only the query parser for ES needs to be changed?

longquanzheng:
> > This is going to conflict with my refactoring #4223 and #4241
> > Would you mind review my PR and let it go in first(because it's quite hard to resolve conflict in my refactoring PR ><
> > I would really appreciate it! Thanks!
> 
> I guess if only elasticsearch implementation is touched (for List/Scan/CountWorkflowExecution API), there should be no conflicts as only the query parser for ES needs to be changed?

Yes, agreed. that will be great

demirkayaender:
Abandoning this PR since we are going with ES only solution

PR #4247, opened 2021-06-02
Fix Cassandra schema version check

yycptt:
**What changed?**
- Fix Cassandra schema version check

**Why?**
`Cassandra` field in `Datastore` config is deprecated as part of the nosql plugin refactoring. As a result, the schema version check for cassandra is skipped on service start up.

**How did you test it?**
Tested locally.

**Potential risks**

**Release notes**

**Documentation Changes**


tools/cassandra/handler.go
```diff
@@ -47,24 +48,48 @@ type SetupSchemaConfig struct {
...
+) error {
+	var expectedVersion string
+	var cfg config.NoSQL
+	if ds.NoSQL != nil {
+		var ok bool
+		if expectedVersion, ok = expectedVersions[ds.NoSQL.PluginName]; !ok {
+			return fmt.Errorf("unknown NoSQL plugin name: %v", ds.NoSQL.PluginName)
+		}
+		cfg = *ds.NoSQL
+	} else if ds.Cassandra != nil {
```
longquanzheng, line 82: You can remove this check I believe. We are always refilling NoSQL from Cassandra.

tools/cassandra/handler.go
```diff
@@ -47,24 +48,48 @@ type SetupSchemaConfig struct {
...
+		expectedVersion = cassandra.VisibilityVersion
+		cfg = *ds.Cassandra
+	} else {
+		// not using nosql or cassandra
```
longquanzheng, line 86: Nit: returning error here should be fine.

tools/cassandra/handler.go
```diff
@@ -47,24 +48,48 @@ type SetupSchemaConfig struct {
...
 		}
 	}
-	ds, ok = cfg.DataStores[cfg.VisibilityStore]
-	if ok && ds.Cassandra != nil {
-		err := CheckCompatibleVersion(*ds.Cassandra, cassandra.VisibilityVersion)
-		if err != nil {
+
+	if ds, ok := cfg.DataStores[cfg.VisibilityStore]; ok {
+		if err := verifyCompatibleVersion(ds, map[string]string{
+			cassandra_db.PluginName: cassandra.VisibilityVersion,
```
longquanzheng, line 61: It's generally not a good practice to pass a map like this. If you have to, maybe define a new struct. But here I think we can keep passing the version only..

It's a bigger work for supporting different plugins. We could do that separately(feel free to do it if you are interested!)

tools/cassandra/handler.go
yycptt: I can't return error here, otherwise server using sql config can't start as we always execute the verification function for both cassandra and sql. See: https://github.com/uber/cadence/blob/master/cmd/server/cadence/cadence.go#L65

tools/cassandra/handler.go
yycptt: I think we need to support old config files that may still be used by some open source customers and in our internal repo?

tools/cassandra/handler.go
yycptt: Yeah probably I can only pass in cassandra version here for now and return an error if plugin name is not equal to `cassandra`?

tools/cassandra/handler.go
longquanzheng: Yes, but it’s back filled here so NoSQL struct should not be nil https://github.com/uber/cadence/blob/723ecf572b2391ad3acc93c455445d9968170eef/common/config/persistence.go#L53

tools/cassandra/handler.go
longquanzheng: Yes, that’s fine for now.

tools/cassandra/handler.go
yycptt: Ah, don't know we have the backfill logic there. Thanks for pointing that out. Removed the check on `ds.Cassandra` field.

tools/cassandra/handler.go
yycptt: Updated to only pass in cassandra version to the func.

longquanzheng (approved):

PR #4248, opened 2021-06-03
Remove timerGate usage in transfer queue processor

yycptt:
**What changed?**
- Remove timerGate usage in transfer queue processor
- Only schedule timer when a queue is throttled

**Why?**
- Improve performance

**How did you test it?**
Integration test and canary

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/queue/transfer_queue_processor_base.go
```diff
@@ -173,8 +179,12 @@ func (t *transferQueueProcessorBase) Stop() {
 	t.logger.Info("Transfer queue processor state changed", tag.LifeCycleStopping)
 	defer t.logger.Info("Transfer queue processor state changed", tag.LifeCycleStopped)
 
-	t.nextPollTimer.Close()
 	close(t.shutdownCh)
+	t.processingLock.Lock()
+	for _, timer := range t.backoffTimer {
+		timer.Stop()
+	}
```
yux0, line 186: Do you need to change shouldProcess to false?

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -259,22 +316,14 @@ processorPumpLoop:
...
-				if !now.Before(pollTime.time) {
-					levels[level] = struct{}{}
-					delete(t.nextPollTime, level)
-				} else {
-					t.nextPollTimer.Update(pollTime.time)
+				select {
+				case t.processCh <- struct{}{}:
+				default:
 				}
+				continue processorPumpLoop
```
yux0, line 323: Any reason to send signal and continue? Could it just do t.processQueueCollections()?

yycptt: The purpose here is to block all task processing when there are too many pending tasks in the redispatcher (which means we are buffering too many tasks and may run oom), so we can't just do `t.processQueueCollections()`. Continue the processorLoop here so that the check will be executed and backoff again if needed.

service/history/queue/transfer_queue_processor_base.go
yycptt: We don't really care about the value in `shouldProcess` actually. As long as the `shutdownCh` is closed the processing loop will be stopped. But I can set it to false, so that the for loop in `processQueueCollections` can stop early. Will update.

yycptt: Updated.

PR #4250, opened 2021-06-04
Generate cross cluster tasks in mutable state task generator

yycptt:
**What changed?**
- Generate cross cluster tasks in mutable state task generator
- Allocate cross cluster taskID and timestamp in shardContext

**Why?**
- For supporting cross-domain operations

**How did you test it?**
- Unit test

**Potential risks**
N/A, currently cross-domain/cluster operation is not allowed in decision checker.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_task_generator.go
+++ b/service/history/execution/mutable_state_task_generator.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/cluster"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/types"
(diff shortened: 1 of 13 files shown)


service/history/execution/mutable_state_task_generator.go
```diff
@@ -86,8 +87,9 @@ type (
 	}
 
 	mutableStateTaskGeneratorImpl struct {
-		domainCache cache.DomainCache
-		logger      log.Logger
+		clusterMetadata cluster.Metadata
```
yux0, line 90: Only store current cluster name?

yux0 (approved):

service/history/execution/mutable_state_task_generator.go
yycptt: I can change to currentClusterName if you have strong opinion on it. But I kinda think it's useful to keep clusterMetadata here which can be used if we want to do additional checks in the future for example if targetCluster is a valid cluster name defined in the metadata.

PR #4252, opened 2021-06-05
Adding protoversion as a config value

noiarek:
**What changed?**
ProtoVersion can now be defined in development.yaml. If not present, the default value will be 4. This will be done in a 2 steps process:

1. Adding the variable in the structs without adding them in the config files (to make sure compatibility is working)
2. Adding the config files in a different PR

**Why?**
Labeled as `Up for grabs` as `good first issue` in https://github.com/uber/cadence/issues/3536

**How did you test it?**
Playing with some values in the development.yaml file. Happy to add some unit test if you indicate me where

**Potential risks**
There was a default value set in `newCassandraCluster`, however the value is sent by value and not by reference so it won't be set somewhere else. Potentially the right place would be `tools/cassandra/handler.go` but please let me know where you think I should set this value in case this is not the right place

**Release notes**
Maybe? Let me know if you want me to do that in the second step

**Documentation Changes**
Likely to be the case. Let me know if I need to do this in the second step

### DIFF
--- a/common/persistence/cassandra/cassandraPersistenceTest.go
+++ b/common/persistence/cassandra/cassandraPersistenceTest.go
@@ -48,7 +48,7 @@ type TestCluster struct {
 }
 
 // NewTestCluster returns a new cassandra test cluster
-func NewTestCluster(keyspace, username, password, host string, port int, schemaDir string) *TestCluster {
+func NewTestCluster(keyspace, username, password, host string, port int, schemaDir string, protoVersion int) *TestCluster {
 	var result TestCluster
 	result.keyspace = keyspace
 	if port == 0 {
@@ -60,14 +60,18 @@ func NewTestCluster(keyspace, username, password, host string, port int, schemaD
 	if host == "" {
 		host = environment.GetCassandraAddress()
 	}
+	if protoVersion == 0 {
+		protoVersion = environment.GetCassandraProtoVersion()
+	}
 	result.schemaDir = schemaDir
 	result.cfg = config.Cassandra{
-		User:     username,
-		Password: password,
-		Hosts:    host,
-		Port:     port,
-		MaxConns: 2,
-		Keyspace: keyspace,
+		User:         username,
+		Password:     password,
+		Hosts:        host,
+		Port:         port,
+		MaxConns:     2,
+		Keyspace:     keyspace,
+		ProtoVersion: protoVersion,
 	}
 	return &result
 }
@@ -118,13 +122,14 @@ func (s *TestCluster) TearDownTestDatabase() {
 // CreateSession from PersistenceTestCluster interface
 func (s *TestCluster) CreateSession() {
 	s.cluster = &gocql.ClusterConfig{
-		Hosts:       s.cfg.Hosts,
-		Port:        s.cfg.Port,
-		User:        s.cfg.User,
-		Password:    s.cfg.Password,
-		Keyspace:    "system",
-		Consistency: gocql.One,
-		Timeout:     40 * time.Second,
+		Hosts:        s.cfg.Hosts,
+		Port:         s.cfg.Port,
+		User:         s.cfg.User,
+		Password:     s.cfg.Password,
+		ProtoVersion: s.cfg.ProtoVersion,
+		Keyspace:     "system",
+		Consistency:  gocql.One,
+		Timeout:      40 * time.Second,
 	}
 
 	var err error
@@ -155,7 +160,7 @@ func (s *TestCluster) DropDatabase() {
 // LoadSchema from PersistenceTestCluster interface
 func (s *TestCluster) LoadSchema(fileNames []string, schemaDir string) {
 	workflowSchemaDir := schemaDir + "/cadence"
-	err := loadCassandraSchema(workflowSchemaDir, fileNames, s.cluster.Hosts, s.cluster.Port, s.DatabaseName(), true, nil)
+	err := loadCassandraSchema(workflowSchemaDir, fileNames, s.cluster.Hosts, s.cluster.Port, s.DatabaseName(), true, nil, s.cluster.ProtoVersion)
 	if err != nil && !strings.Contains(err.Error(), "AlreadyExists") {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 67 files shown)


noiarek:
@yycptt @longquanzheng  I was having some weird issues when I was running `make bins` in https://github.com/uber/cadence/pull/4188 so I closed that PR and created a new branch/PR. 

Since this is a really big change I thought it might be a good idea if I first make all the changes without the config files (to see if it's working properly and passing all the test, my test fail in local because I dont have mysql). If  everything looks correct, I would create a 2nd PR where I would add all the changes on the yaml, config and env files (if that's ok to you)

longquanzheng:
> @yycptt @longquanzheng I was having some weird issues when I was running `make bins` in #4188 so I closed that PR and created a new branch/PR.
> 
> Since this is a really big change I thought it might be a good idea if I first make all the changes without the config files (to see if it's working properly and passing all the test, my test fail in local because I dont have mysql). If everything looks correct, I would create a 2nd PR where I would add all the changes on the yaml, config and env files (if that's ok to you)

I am fine with it. Thanks for the contribution.

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
```diff
@@ -51,7 +51,10 @@ func RegisterClient(c Client) {
 func newCassandraCluster(cfg ClusterConfig) *gocql.ClusterConfig {
 	hosts := parseHosts(cfg.Hosts)
 	cluster := gocql.NewCluster(hosts...)
-	cluster.ProtoVersion = 4
+	if cfg.ProtoVersion == 0 {
+		cfg.ProtoVersion = 4
```
longquanzheng, line 55: Since you have declare the global constant, this cone should become
```
cfg.ProtoVersion = env.CassandraDefaultProtoVersion
```

tools/cassandra/cqlclient.go
```diff
@@ -93,6 +94,9 @@ var _ schema.DB = (*CqlClient)(nil)
 func NewCQLClient(cfg *CQLClientConfig) (*CqlClient, error) {
 	var err error
 
+	if cfg.ProtoVersion == 0 {
+		cfg.ProtoVersion = cqlProtoVersion
+	}
```
longquanzheng, line 99: I believe the default will be set later. So these three lines can be removed, along with the local constant `cqlProtoVersion`?

longquanzheng (approved):
Mostly LGTM. Only a few minor comments.

tools/cassandra/cqlclient.go
noiarek: Good catch. About removing the const, I'm using it in `tools/cassandra/main.go:L108` in this diff, I'm wondering thou if it's a good idea to use `environment.GetCassandraProtoVersion()` (I created that function) in `tools/cassandra/main.go:L108` instead?

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
noiarek: What about using the function `environment.GetCassandraProtoVersion()` that I created instead? (to avoid the int parsing). And another question would be, if the answer is yes, should I use it everywhere instead of a const? (i.e `common/persistence/nosql/nosqlplugin/cassandra/session.go:39` in this diff)

tools/cassandra/cqlclient.go
longquanzheng: You already did it in `CASSANDRA_PROTOCOL_VERSION` in tools/cassandra/main.go.
So here just use the constant value. Getting from another env will be overlapping(and conusing)

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
longquanzheng: Instead of parsing, how about create another `environment.CassandraDefaultProtoVersionInteger` beside `environment.CassandraDefaultProtoVersion`? and always use environment.CassandraDefaultProtoVersionInteger to avoid parsing string.

common/persistence/nosql/nosqlplugin/cassandra/session.go
```diff
@@ -35,6 +35,9 @@ const (
 // CreateSession creates a new session
 // TODO this will be converted to private later, after all cassandra code moved to plugin pkg
 func CreateSession(cfg config.Cassandra) (gocql.Session, error) {
+	if cfg.ProtoVersion == 0 {
+		cfg.ProtoVersion = cassandraProtoVersion
```
longquanzheng, line 39: Also here

common/persistence/cassandra/cassandraPersistenceTest.go
```diff
@@ -60,14 +60,18 @@ func NewTestCluster(keyspace, username, password, host string, port int, schemaD
 	if host == "" {
 		host = environment.GetCassandraAddress()
 	}
+	if protoVersion == 0 {
+		protoVersion = environment.GetCassandraProtoVersion()
```
noiarek, line 64: Would you like me to replace these  for the const `CassandraDefaultProtoVersionInteger` @longquanzheng ?

here, in `common/persistence/nosql/nosqlplugin/cassandra/tests/cassandra_tool_version_test.go` and `common/persistence/nosql/nosqlplugin/cassandra/tests/utils.go`, all in this diff.

longquanzheng: The tests have been using `environment.GetCassandraXYZ` so it's better to keep consistent. For others yes, if they were using the hardcoded value, then replace with the `CassandraDefaultProtoVersionInteger`
Thank you!

common/persistence/nosql/nosqlplugin/cassandra/session.go
```diff
@@ -23,18 +23,22 @@ package cassandra
...
-	cassandraProtoVersion = 4
 	defaultSessionTimeout = 10 * time.Second
 )
 
 // CreateSession creates a new session
 // TODO this will be converted to private later, after all cassandra code moved to plugin pkg
 func CreateSession(cfg config.Cassandra) (gocql.Session, error) {
+	if cfg.ProtoVersion == 0 {
+		cfg.ProtoVersion = environment.CassandraDefaultProtoVersionInteger
+	}
```
longquanzheng, line 41: I believe this is not needed as the default will be set within `CreateSession`, can you double check?

common/persistence/nosql/nosqlplugin/cassandra/session.go
```diff
@@ -23,18 +23,22 @@ package cassandra
 import (
 	"time"
 
+	"github.com/uber/cadence/environment"
+
```
longquanzheng, line 27: can you remove this empty line?

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
```diff
@@ -24,6 +24,8 @@ import (
 	"crypto/tls"
 	"strings"
 
+	"github.com/uber/cadence/environment"
```
longquanzheng, line 27: can you move this under `"github.com/gocql/gocql"`? 
If you look at any other files, the convention is:
```
 imports of golang standard libary
<newline>
imports of other libraries
<newline>
imports of current repo
```

common/persistence/nosql/nosqlplugin/cassandra/session.go
noiarek: You're right, removed

common/persistence/nosql/nosqlplugin/cassandra/gocql/client.go
noiarek: Fixed, funny thing is that `make fmt` wrote it there

PR #4256, opened 2021-06-06
Rename persistence files for consistency 

longquanzheng:
**What changed?**
Manager is the first level, Store is the 2nd level.

**Why?**
For consistency and clearness 

**How did you test it?**
Existing test

**Potential risks**
No, only file name changes and struct name changes. 

**Release notes**

**Documentation Changes**


service/history/reset/resetter.go
```diff
@@ -23,7 +23,7 @@
 package reset
 
 import (
-	context "context"
+	"context"
 	ctx "context"
```
yycptt, line 27: use only one `context` import?

common/persistence/utils/clusterMetadata.go
```diff
@@ -18,7 +18,9 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package persistence
+package utils
```
yycptt, line 21: nit: I kind of think having persistence in the package and import name is more informative than just `utils`, which is not a recommended package name (https://blog.golang.org/package-names#TOC_5.)

yycptt (commented):
- Shall we also rename contents in sqlQueue.go also to `sqlQueueStore` to be consistence and match the filename?
- Is it possible to merge `dataVisibilityManagerInterfaces.go` into `dataManagerInterfaces.go`?

common/persistence/utils/clusterMetadata.go
longquanzheng: yeah good point. I dislike that too. I moved the ClusterMetadata to `cluster` package, and renamed this to `persistence-util`

service/history/reset/resetter.go
longquanzheng: Yeah good catch. Fixed

common/cluster/metadata.go
```diff
@@ -33,6 +33,7 @@ import (
 
 type (
 	// Metadata provides information about clusters
+	// TODO: use a more informative name, e.g. FailoverMonitor
```
yycptt, line 36: Does the TODO apply here? cluster.Metadata name seems reasonable to me for those methods.

longquanzheng: Yeah naming is controversial.  I just dislike the overloaded name of `metadata` everywhere. 
Maybe `ClusterMeta` will make more sense for me. 
I will remove this TODO and see if more people agree with me in the future.

common/persistence/persistenceutils/historyManagerUtil.go
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
longquanzheng: Removed. Thanks

yycptt (approved):

PR #4259, opened 2021-06-08
Add GetTasks in processing queue

yux0:
**What changed?**
Add GetTasks in processing queue

**Why?**
This is for cross cluster task polling

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/queue/interface.go
```diff
@@ -56,6 +56,7 @@ type (
 		Split(ProcessingQueueSplitPolicy) []ProcessingQueue
 		Merge(ProcessingQueue) []ProcessingQueue
 		AddTasks(map[task.Key]task.Task, task.Key)
+		GetTasks() map[task.Key]task.Task
```
yycptt, line 59: Is it possible to use a different name, like `GetPendingTasks()` or `GetOutstandingTasks()`? I think in the future we will moving task loading logic into processing queue and the `GetTask` name should be reserved for loading new tasks from persistence or the in memory cache.

PR #4261, opened 2021-06-08
Update generating close event to use event version

yux0:
**What changed?**
Update generating close event to use event version

**Why?**
The task refresher may generate a task with newer version and gets ignored later

**How did you test it?**
Unit tests.

**Potential risks**
If there is bugs with this change, there could be race condition in handling closed event

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -1152,6 +1152,36 @@ func (e *mutableStateBuilder) GetStartEvent(
 	return startEvent, nil
 }
 
+// GetCloseEvent returns the last event in history
+func (e *mutableStateBuilder) GetCloseEvent(
+	ctx context.Context,
+) (*types.HistoryEvent, error) {
+
+	if e.GetExecutionInfo().CloseStatus == persistence.WorkflowCloseStatusNone {
+		return nil, nil
+	}
+
+	currentBranchToken, err := e.GetCurrentBranchToken()
+	if err != nil {
+		return nil, err
+	}
+
+	closeEvent, err := e.eventsCache.GetEvent(
+		ctx,
+		e.shard.GetShardID(),
+		e.executionInfo.DomainID,
+		e.executionInfo.WorkflowID,
+		e.executionInfo.RunID,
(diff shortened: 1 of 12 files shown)


service/history/execution/mutable_state_task_refresher.go
```diff
@@ -218,10 +218,14 @@ func (r *mutableStateTaskRefresherImpl) refreshTasksForWorkflowClose(
...
 	executionInfo := mutableState.GetExecutionInfo()
-
 	if executionInfo.CloseStatus != persistence.WorkflowCloseStatusNone {
+		closeEvent, err := mutableState.GetCloseEvent(ctx)
+		if err != nil {
+			return err
+		}
 		return taskGenerator.GenerateWorkflowCloseTasks(
 			now,
+			closeEvent,
```
yycptt, line 228: it's possible that the closeEvent returned is nil even when error is nil right? then it will cause a NPE in taskGenerator?

service/history/execution/mutable_state_builder.go
```diff
@@ -1152,6 +1152,36 @@ func (e *mutableStateBuilder) GetStartEvent(
 	return startEvent, nil
 }
 
+// GetCloseEvent returns the last event in history
+func (e *mutableStateBuilder) GetCloseEvent(
+	ctx context.Context,
+) (*types.HistoryEvent, error) {
+
+	if e.GetExecutionInfo().CloseStatus == persistence.WorkflowCloseStatusNone {
+		return nil, nil
```
yycptt, line 1161: not sure if it's better to return an error here. Usually caller will assume historyEvent is not nil when error is nil?

yycptt (approved):

service/history/execution/mutable_state_task_refresher.go
yux0: Will return error when close event is nil

service/history/execution/mutable_state_builder.go
yux0: Will return error when close event is nil

PR #4265, opened 2021-06-09
preserve some time to return partial of the result

yux0:
**What changed?**
preserve some time to return part of the result

**Why?**
if the context is timing out, we should return part of the result instead of time out the request

**How did you test it?**
Local test

**Potential risks**

**Release notes**

**Documentation Changes**


client/history/client.go
```diff
@@ -821,16 +822,24 @@ func (c *clientImpl) GetReplicationMessages(
...
 	}
 
+	// preserve some time to return partial of the result if context is timing out
+	now := time.Now()
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = now.Add(c.timeout)
+	}
+	requestTimeout := time.Duration(math.Ceil(float64(deadline.Sub(now)) * 0.95))
+	requestContext, cancel := context.WithTimeout(context.Background(), requestTimeout)
```
yycptt, line 832: nit: not sure if using `ctx` instead of `context.Backgroud()` can preserve more information (like for tracing) here?

yycptt (approved):

client/history/client.go
yux0: Good point. Will change to the original context

PR #4268, opened 2021-06-10
Convert transfer to cross cluster task if target domain is active in remote cluster

yycptt:
**What changed?**
- Convert transfer to cross cluster task if target domain is active in remote cluster
- Refactor unit test for transfer active task executor

**Why?**
Cross domain task life cycle

**How did you test it?**
Unit test

**Potential risks**
N/A, no cross cluster task will be generated right now

**Release notes**

**Documentation Changes**


service/history/task/transfer_active_task_executor.go
```diff
@@ -630,6 +635,25 @@ func (t *transferActiveTaskExecutor) processStartChildExecution(
...
+	var targetDomainName string
+	var targetDomainEntry *cache.DomainCacheEntry
+	if targetDomainEntry, err = t.shard.GetDomainCache().GetDomainByID(task.TargetDomainID); err != nil {
+		if _, ok := err.(*types.EntityNotExistsError); !ok {
+			return err
+		}
+		// TODO: handle the case where target domain does not exist
+
+		// it is possible that the domain got deleted. Use domainID instead as this is only needed for the history event
+		targetDomainName = task.TargetDomainID
```
yux0, line 648: any reason to assgin domain ID to domain name?

service/history/task/transfer_active_task_executor.go
```diff
@@ -630,6 +635,25 @@ func (t *transferActiveTaskExecutor) processStartChildExecution(
...
+	} else {
+		if targetCluster := targetDomainEntry.GetReplicationConfig().ActiveClusterName; targetCluster != t.shard.GetClusterMetadata().GetCurrentClusterName() {
```
yux0, line 650: move to a function isCrossClusterTask(activeClusterName string) bool ?

yycptt: Sure, will do.

service/history/task/transfer_active_task_executor.go
yycptt: This is the old logic for handling domain not exist case. I actually don't think it's correct as the start child request will fail anyway due to domain not exists. Since for other task types (signal and cancel) we also need to handle the domain not exists case, I left a TODO in those places and keep the old logic untouched.

yux0 (approved):

service/history/task/transfer_active_task_executor.go
yycptt: Updated

PR #4269, opened 2021-06-11
Add cross cluster queue processor

yux0:
**What changed?**
Add cross cluster queue processor

**Why?**
Support cross cluster queue task fetching and processing

**How did you test it?**
TODO

**Potential risks**
New feature

**Release notes**

**Documentation Changes**


service/history/queue/action.go
```diff
@@ -50,14 +56,35 @@ type (
...
 	// ActionTypeReset is the ActionType for reseting processing queue states
 	ActionTypeReset ActionType = iota + 1
 	// ActionTypeGetState is the ActionType for reading processing queue states
 	ActionTypeGetState
+	// ActionTypeGetTasks is the ActionType for get cross cluster tasks
+	ActionTypeGetTasks
+	// ActionTypeUpdateTask is the ActionType to update outstanding task
+	ActionTypeUpdateTask
 	// add more ActionType here
+
```
yux0, line 87: remove

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	if c.isGlobalDomainEnabled {
+		for _, queueProcessor := range c.queueProcessorByCluster {
+			queueProcessor.queue.Start()
+		}
+	}
+
+	c.shutdownWG.Add(1)
+	go c.completeTaskLoop()
```
yux0, line 173: no need

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,317 @@
...
+	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/service/history/config"
+	"github.com/uber/cadence/service/history/engine"
+	"github.com/uber/cadence/service/history/shard"
+	"github.com/uber/cadence/service/history/task"
+)
+
+type (
+	queueProcessorWithAckLevel struct {
```
yux0, line 42: ack level no need

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+
+processorPumpLoop:
+	for {
+		select {
+		case <-c.shutdownCh:
+			break processorPumpLoop
+		case <-c.notifyCh:
+			c.notifyAllQueueCollections()
+		case <-updateAckTimer.C:
+			processFinished, err := c.updateAckLevel()
```
yux0, line 177: 1. queue base to update queue state
2. delete tasks prior to the ack level

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+		case <-c.processCh:
+			maxRedispatchQueueSize := c.options.MaxRedispatchQueueSize()
+			if c.redispatcher.Size() > maxRedispatchQueueSize {
+				// has too many pending tasks in re-dispatch queue, block loading tasks from persistence
+				c.redispatcher.Redispatch(maxRedispatchQueueSize)
+				if c.redispatcher.Size() > maxRedispatchQueueSize {
+					// if redispatcher still has a large number of tasks
+					// this only happens when system is under very high load
+					// we should backoff here instead of keeping submitting tasks to task processor
+					time.Sleep(backoff.JitDuration(
```
yux0, line 201: setup backoffTimer instead of sleep

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+
+func (c *crossClusterQueueProcessorBase) pollTasks() []task.Task {
+
+	var result []task.Task
+	for _, queueTask := range c.getTasks() {
+		crossClusterTask, ok := queueTask.(task.CrossClusterTask)
+		if !ok {
+			panic("received non cross cluster task")
+		}
+		if crossClusterTask.IsReadyForPoll() {
```
yux0, line 255: TODO: check domain active cluster

yux0: workflow version check state check

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+		tasks := make(map[task.Key]task.Task)
+		newTaskID := readLevel.(transferTaskKey).taskID
+
+		c.Lock()
+		for _, taskInfo := range taskInfos {
+			if !domainFilter.Filter(taskInfo.GetDomainID()) {
+				continue
+			}
+
+			if c.outstandingTaskCount > c.options.MaxPendingTaskSize() {
```
yux0, line 468: check by batch

service/history/queue/interface.go
```diff
@@ -73,6 +75,8 @@ type (
 		Queues() []ProcessingQueue
 		ActiveQueue() ProcessingQueue
 		AddTasks(map[task.Key]task.Task, task.Key)
+		GetTask(task.Key) (task.Task, error)
```
yux0, line 78: GetOutstandingTask
GetOutstandingTasks

service/history/queue/processing_queue_test.go
```diff
@@ -110,6 +110,140 @@ func (s *processingQueueSuite) TestAddTasks() {
...
+	queue := s.newTestProcessingQueue(
+		0,
+		ackLevel,
+		ackLevel,
+		maxLevel,
+		NewDomainFilter(nil, true),
+		make(map[task.Key]task.Task),
+	)
+
+	newReadLevel := testKey{ID: 10}
```
demirkayaender, line 180: What does level mean in this context? I have seen it used with ackLevel, readLevel and readMaxLevel.

Are those offsets where sender and receiver send to each other. Is readLevel the offset receiver uses to read tasks starting from the given id and ackLevel the offset used by the sender to understand until which task id the tasks were transferred successfully?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+	c.processingLock.Unlock()
+
+	if success := common.AwaitWaitGroup(&c.shutdownWG, time.Minute); !success {
+		c.logger.Warn("", tag.LifeCycleStopTimedout)
+	}
+
+	c.redispatcher.Stop()
+}
+
+func (c *crossClusterQueueProcessorBase) processorPump() {
```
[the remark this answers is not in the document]
yux0: updated

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,317 @@
...
+) (*ActionResult, error) {
+
+	queueProcessor, ok := c.queueProcessorByCluster[clusterName]
+	if !ok {
+		return nil, fmt.Errorf("failed to find the cross cluster queue with cluster name: %v", clusterName)
+	}
+
+	resultNotificationCh, added := queueProcessor.queue.addAction(action)
+	if !added {
+		return nil, errProcessorShutdown
```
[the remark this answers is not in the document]
yux0: Only when the base queue is shutitng down. The caller of this function can log the message.

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	select {
+	case resultNotification := <-resultNotificationCh:
+		return resultNotification.result, resultNotification.err
+	case <-c.shutdownChan:
+		return nil, errProcessorShutdown
```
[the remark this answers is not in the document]
yux0: Yes. It need to wait for the result to return. If the execute is interrupted, when the shutdown channel should close

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,317 @@
...
+			ackLevel = minTaskKey(ackLevel, state.AckLevel())
+		}
+
+		if ackLevel == maximumTransferTaskKey {
+			panic(fmt.Sprintf("no ack level is defined in cross cluster queeu with cluster: %v", clusterName))
+		}
+
+		queueProcessorByCluster[clusterName] = &queueProcessorWithAckLevel{
+			queue:    queueProcessor,
+			ackLevel: ackLevel.(transferTaskKey).taskID,
```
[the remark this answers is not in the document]
yux0: updated

service/history/queue/cross_cluster_queue_processor.go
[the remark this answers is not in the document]
yux0: This is expected when the processor base is closed.

service/history/queue/action.go
```diff
@@ -29,6 +31,8 @@ type (
 		ActionType               ActionType
 		ResetActionAttributes    *ResetActionAttributes
 		GetStateActionAttributes *GetStateActionAttributes
+		GetTasksAttributes       *GetTasksAttributes
+		UpdateTaskAttributes     *UpdateTaskAttributes
```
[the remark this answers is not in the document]
yux0: Yes. Those will be integrated with the task fetcher.

service/history/queue/processing_queue_test.go
yux0: Ack level is the max task id that has been processed. 
Read level is the max task id loaded into the queue(in memory)
Max read level is the max task id that can be read from DB. 

Those offsets are stored in x cross queue only.

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,538 @@
...
+			&persistence.GetCrossClusterTasksRequest{
+				TargetCluster: c.targetCluster,
+				ReadLevel:     readLevel.(transferTaskKey).taskID,
+				MaxReadLevel:  maxReadLevel.(transferTaskKey).taskID,
+				BatchSize:     c.options.BatchSize(),
+			})
+		return err
+	}
+
+	err := backoff.Retry(op, persistenceOperationRetryPolicy, persistence.IsTransientError)
```
Shaddoll, line 307: Probably, we should retry on ServiceBusyError. If the error is from database, retry can create more requests which will lead to more errors.

yux0: Yes. Remove this error from retry policy. The queue processor is going to retry with a lower rate

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+	logger log.Logger,
+	metricsClient metrics.Client,
+) *crossClusterQueueProcessorBase {
+	processorBase := newProcessorBase(
+		shard,
+		processingQueueStates,
+		taskProcessor,
+		options,
+		updateMaxReadLevel,
+		nil,
```
Shaddoll, line 82: It will panic if `EnablePersistQueueStates ` is false.

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+				c.options.UpdateAckInterval(),
+				c.options.UpdateAckIntervalJitterCoefficient(),
+			))
+		case <-maxPollTimer.C:
+			c.notifyAllQueueCollections()
+			if time.Now().After(lastPollTime.Add(c.options.ValidationInterval())) {
+				// No polling is happening and loop all outstanding task and validate task status
+				c.validateOutstandingTasks()
+				lastPollTime = time.Now()
+			}
```
yycptt, line 221: I am thinking if we can do this validation on task load and when `FailoverDomain()` method is called on the Processor. Then I guess we don't need to perform this check periodically and can react more quickly upon domain failover?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+			break processorPumpLoop
+		case <-c.notifyCh:
+			c.notifyAllQueueCollections()
+		case <-updateAckTimer.C:
+			processFinished, err := c.updateAckLevel()
+			if err == shard.ErrShardClosed || (err == nil && processFinished) {
+				go c.Stop()
+				break processorPumpLoop
+			}
+			ackLevel, err = c.completeAckTasks(ackLevel)
```
yycptt, line 206: - Looks like we can change `updateAckLevel()` to also return the new ackLevel and pass it into `completeAckTasks()` so that we don't have to compute it again? 
- The current ackLevel can maybe be a field of `crossClusterQueueProcessorBase`? Currently, it's kind of confusing to me that the ackLevel pass into the function is actually the old ack level that is not used for deleting tasks and a new one is calculated inside the func.

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+	}
+	newAckLevelTaskID := newAckLevel.(transferTaskKey).taskID
+	c.logger.Debug(fmt.Sprintf("Start completing cross cluster task from: %v, to %v.", ackLevel, newAckLevelTaskID))
+	c.metricsScope.IncCounter(metrics.TaskBatchCompleteCounter)
+
+	if ackLevel >= newAckLevelTaskID {
+		return ackLevel, nil
+	}
+
+	if err := c.shard.GetExecutionManager().RangeCompleteCrossClusterTask(context.Background(), &persistence.RangeCompleteCrossClusterTaskRequest{
```
yycptt, line 277: nit: probably use a finite and small timeout for this persistence op so that it won't hang or block the entire processing loop for too long?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+		TargetCluster:        c.targetCluster,
+		ExclusiveBeginTaskID: ackLevel,
+		InclusiveEndTaskID:   newAckLevelTaskID,
+	}); err != nil {
+		return ackLevel, nil
```
yycptt, line 282: return the error?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+}
+
+func (c *crossClusterQueueProcessorBase) addOutstandingTaskCountLocked(count int) {
+	c.outstandingTaskCount += count
+}
+
+// TODO: this will be called in cross cluster task Acked function.
+func (c *crossClusterQueueProcessorBase) removeOutstandingTaskCountLocked(count int) {
+	c.outstandingTaskCount -= count
+}
```
yycptt, line 383: `Ack()` method will be called concurrently on many tasks, so do we need a lock on the `outstandingTaskCount` field?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+	c.shouldProcess[level] = true
+
+	// trigger the actual processing
+	select {
+	case c.processCh <- struct{}{}:
+	default:
+	}
+}
+
+func (c *crossClusterQueueProcessorBase) resetBackoffTimer() {
```
yycptt, line 409: - I guess it also works if we just call the `setupBackoffTimer` for all queue levels? Whether the existing backoff timer is overwritten or not probably doesn't matter too much?
- With the current impl, I think you'll need to `Stop()` the exiting backoffTimer in the map if there's one?

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+	for _, task := range tasks {
+		if c.isDomainChanged(task) || c.isSourceWorkflowClosed(task) {
+			_, err := c.submitTask(task)
+			if err != nil {
+				// the queue is shutting down
+				return
+			}
+		}
+	}
+}
```
yycptt, line 613: - I am thinking this should probably be a method on the CrossClusterTask itself since different task types may have different conditions. For example, signal/child workflow tasks should be executed only when the workflow is open while recordChild or applyParentClosePolicy tasks should be executed only when the workflow is closed.
- Current `isDomainChanged` only checks for the source domain, I believe we also need to check for the target domain.
- Task version check is also needed I think? Can be done as part of the `IsReadyForPoll` method I think.

service/history/queue/action.go
```diff
@@ -50,13 +56,33 @@ type (
...
+
+	// GetTasksAttributes contains the parameter to get tasks
+	GetTasksAttributes struct{}
+	// GetTasksResult is the result for performing GetTasks Action
+	GetTasksResult struct {
+		tasks []task.Task
+	}
+	// UpdateTaskAttributes contains the parameter to update task
+	// TODO: replace result interface with defined data struct
+	UpdateTaskAttributes struct {
```
yycptt, line 68: - Since tasks will also be responded in batch, probably make this a batch operation as well?
- Update the type of result to `[]*types.CrossClusterTaskResponse` now that it has been defined? TaskID is probably not necessary as it's included in the TaskResponse type.

service/history/queue/action.go
```diff
@@ -50,13 +56,33 @@ type (
 	GetStateActionResult struct {
 		States []ProcessingQueueState
 	}
+
+	// GetTasksAttributes contains the parameter to get tasks
+	GetTasksAttributes struct{}
+	// GetTasksResult is the result for performing GetTasks Action
+	GetTasksResult struct {
+		tasks []task.Task
```
yycptt, line 64: Maybe update the type to be `[]*types.CrossClusterTaskRequest`? There's not much the caller can do with the Task interface and construct a proper TaskRequest I guess?

service/history/queue/cross_cluster_queue_processor_base.go
yux0: Yes. Set it with a static value true

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+func newCrossClusterQueueProcessor(
+	shard shard.Context,
+	clusterName string,
+	taskProcessor task.Processor,
+	taskExecutor task.Executor,
+	logger log.Logger,
+) *crossClusterQueueProcessorBase {
```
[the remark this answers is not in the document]
yux0: updated

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,215 @@
...
+		)
+		queueProcessors[clusterName] = queueProcessor
+	}
+
+	return &crossClusterQueueProcessor{
+		shard:                 shard,
+		historyEngine:         historyEngine,
+		taskProcessor:         taskProcessor,
+		config:                config,
+		isGlobalDomainEnabled: shard.GetClusterMetadata().IsGlobalDomainEnabled(),
```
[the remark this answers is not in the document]
yux0: remove it

service/history/queue/processor_base.go
```diff
@@ -72,6 +72,7 @@ type (
 		EnableLoadQueueStates                dynamicconfig.BoolPropertyFn
 		EnableValidator                      dynamicconfig.BoolPropertyFn
 		ValidationInterval                   dynamicconfig.DurationPropertyFn
+		MaxPendingTaskSize                   dynamicconfig.IntPropertyFn
```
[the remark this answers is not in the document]
yux0: updated

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,688 @@
...
+
+	options.EnableSplit = config.QueueProcessorEnableSplit
+	options.SplitMaxLevel = config.QueueProcessorSplitMaxLevel
+	options.EnableRandomSplitByDomainID = config.QueueProcessorEnableRandomSplitByDomainID
+	options.RandomSplitProbability = config.QueueProcessorRandomSplitProbability
+	options.EnablePendingTaskSplitByDomainID = config.QueueProcessorEnablePendingTaskSplitByDomainID
+	options.PendingTaskSplitThreshold = config.QueueProcessorPendingTaskSplitThreshold
+	options.EnableStuckTaskSplitByDomainID = config.QueueProcessorEnableStuckTaskSplitByDomainID
+	options.StuckTaskSplitThreshold = config.QueueProcessorStuckTaskSplitThreshold
+	options.SplitLookAheadDurationByDomainID = config.QueueProcessorSplitLookAheadDurationByDomainID
```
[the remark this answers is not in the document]
yux0: updated

service/history/queue/cross_cluster_queue_processor_base.go
yux0: updated

service/history/queue/cross_cluster_queue_processor_base.go
yux0: I can move the locking logic into this method

service/history/queue/transfer_queue_processor_base.go
```diff
@@ -514,7 +509,7 @@ func (t *transferQueueProcessorBase) readTasks(
 		return err
 	}
 
-	err := backoff.Retry(op, persistenceOperationRetryPolicy, persistence.IsTransientError)
+	err := backoff.Retry(op, persistenceOperationRetryPolicy, persistence.IsBackgroundTransientError)
```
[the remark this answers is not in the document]
yux0: Yes. The IsBackgroundTransientError will not retry on service busy error

service/history/queue/cross_cluster_queue_processor_base.go
yux0: updated

service/history/queue/cross_cluster_queue_processor_base.go
yux0: This is going to be a change that affects all three queue types. I added a TODO and will do it in the following PR

service/history/queue/cross_cluster_queue_processor_base.go
yux0: Updated. The stop() is already been called in processor stop()

service/history/queue/cross_cluster_queue_processor_base.go
yux0: Moved to task interface.

service/history/queue/action.go
yux0: updated

service/history/queue/action.go
yux0: updated

service/history/queue/cross_cluster_queue_processor_base.go
yux0: I included the change in this PR

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,696 @@
...
+		case <-maxPollTimer.C:
+			c.notifyAllQueueCollections()
+
+			maxPollTimer.Reset(backoff.JitDuration(
+				c.options.MaxPollInterval(),
+				c.options.MaxPollIntervalJitterCoefficient(),
+			))
+		case <-c.processCh:
+			maxRedispatchQueueSize := c.options.MaxRedispatchQueueSize()
+			if c.redispatcher.Size() > maxRedispatchQueueSize {
```
yycptt, line 272: probably also check p.outstandingTaskCount?

yycptt (approved):

service/history/queue/cross_cluster_queue_processor.go
```diff
@@ -0,0 +1,173 @@
...
+	select {
+	case resultNotification := <-resultNotificationCh:
+		return resultNotification.result, resultNotification.err
+	case <-c.shutdownChan:
+		return nil, errProcessorShutdown
+	}
+}
+
+func (c *crossClusterQueueProcessor) FailoverDomain(map[string]struct{}) {
+	for _, queueBase := range c.queueProcessors {
```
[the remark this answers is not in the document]
yux0: Yes. It is using signal now.

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -0,0 +1,696 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), taskCleanupTimeout)
+	defer cancel()
+
+	if err := c.shard.GetExecutionManager().RangeCompleteCrossClusterTask(ctx, &persistence.RangeCompleteCrossClusterTaskRequest{
+		TargetCluster:        c.targetCluster,
+		ExclusiveBeginTaskID: c.ackLevel,
+		InclusiveEndTaskID:   nextAckLevel,
+	}); err != nil {
+		return c.ackLevel, nil
```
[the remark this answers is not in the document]
yux0: updated

PR #4271, opened 2021-06-14
Support Cross Cluster Operations in SQL type DBs

demirkayaender:
**What changed?**
Previously cross cluster support was added to Cassandra (https://github.com/uber/cadence/pull/4237), this change adds it for SQL

**Why?**
So that we can support x-cluster operations in SQL DBs

**How did you test it?**
```
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestCrossClusterTasks
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestGetShard
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestUpdateShard
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestReplicationTasks
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestCreateGetShardBackfill
go test -v ./common/persistence/sql/sqlplugin/mysql -testify.m TestCreateGetUpdateGetShard

go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestCrossClusterTasks
go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestGetShard
go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestUpdateShard
go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestReplicationTasks
go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestCreateGetShardBackfill
go test -v ./common/persistence/sql/sqlplugin/postgres -testify.m TestCreateGetUpdateGetShard
```

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/serialization/thrift_mapper.go
```diff
@@ -616,6 +620,46 @@ func transferTaskInfoFromThrift(info *sqlblobs.TransferTaskInfo) *TransferTaskIn
 	}
 }
 
+func crossClusterTaskInfoToThrift(info *CrossClusterTaskInfo) *sqlblobs.CrossClusterTaskInfo {
```
yycptt, line 623: We can reuse the sqlblobs.TransferTaskInfo type definition for this and the following function I think?

demirkayaender: Wouldn't this cause any issues in the future if TransferTaskInfo and CrossClusterTaskInfo diverge in the future? When we serialize we keep the struct name in the blob right? 

We might normally derive both structs from a common ancestor but not sure if Thrift supports inheritance. We can have it as composition but that also looks like a hack?

yycptt: My understanding is that by design/definition transfer task and cross-cluster task are the same task type. The difference is that when the target domain for a transfer task is in a different cluster we put it in the cross cluster queue, call it cross cluster task and process it using a polling method. If target domain did a failover and later become active in the same cluster as the source domain, that cross cluster task will be convert back to be transfer task. Since the processing logic for transfer and cross-cluster task is identical (and we need convert back and forth between them), and taskInfo only contains information needed for processing the task, I think we can reuse the same struct. 

I think we also mentioned this in Database section of the ERD. @yux0 Can you confirm?

I agree that having two definition works and will give us more flexibility. I guess the downside however, is that in the future for any changes we want to make to transfers tasks info, we will need to change both places.

demirkayaender: Looks like thrift isn't putting the struct name to the serialized data but even then, reusing the same struct name might confuse the developers working on this codebase later. The gain by reusing the same name is less number of lines of code they don't really share the same meaning semantically

common/persistence/sql/sqlplugin/postgres/execution.go
```diff
@@ -83,46 +83,55 @@ domain_id = :domain_id AND
...
 
-	createTransferTasksQuery = `INSERT INTO transfer_tasks(shard_id, task_id, data, data_encoding) 
+	createTransferTasksQuery = `INSERT INTO transfer_tasks(shard_id, task_id, data, data_encoding)
  VALUES(:shard_id, :task_id, :data, :data_encoding)`
 
 	deleteTransferTaskQuery      = `DELETE FROM transfer_tasks WHERE shard_id = $1 AND task_id = $2`
 	rangeDeleteTransferTaskQuery = `DELETE FROM transfer_tasks WHERE shard_id = $1 AND task_id > $2 AND task_id <= $3`
 
+	getCrossClusterTasksQuery = `SELECT task_id, data, data_encoding
+ FROM cross_cluster_tasks WHERE target_cluster = $1 AND shard_id = $2 AND task_id > $3 AND task_id <= $4 ORDER BY target_cluster, shard_id, task_id`
```
yycptt, line 96: looks like we only need to order by task_id? Same question for get transfer task query and sql?

yycptt (approved):
Please update the idl submodule before landing.

PR #4278, opened 2021-06-19
Remove IsConditionalFailed from NoSQL

longquanzheng:
**What changed?**
1. Remove IsConditionalFailed from NoSQL
2. Also add some minor improvement for shardStore to be consistent for the error message

**Why?**
As a follow-up of a comment in previous PR, we want to remove the IsConditionFailed error interface to be consistent with Golang standard
https://github.com/uber/cadence/issues/4273

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlplugin/errors.go
+++ b/common/persistence/nosql/nosqlplugin/errors.go
@@ -20,6 +20,8 @@
 
 package nosqlplugin
 
+import "fmt"
+
 // Condition Errors for NoSQL interfaces
 type (
 	// Only one of the fields must be non-nil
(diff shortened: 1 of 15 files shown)


common/persistence/nosql/nosqlplugin/errors.go
```diff
@@ -0,0 +1,77 @@
...
+		Details string // detail info for logging
+	}
+
+	ConditionFailure struct{}
+)
+
+var _ error = (*WorkflowOperationConditionFailure)(nil)
+
+func (e *WorkflowOperationConditionFailure) Error() string {
+	return "condition failure"
```
yycptt, line 58: nit: shall we indicate the operation in the message? like shard/workflow/task, etc?

yycptt (approved):

common/persistence/nosql/nosqlplugin/errors.go
longquanzheng: yeah, good idea. though this error is not being used yet.

PR #4279, opened 2021-06-20
Use global domain by default

longquanzheng:
**What changed?**
Use global domain by default 

**Why?**
Local domain has no more value compared to global domain, even in a single cluster setup. 
It's only limiting users from migrating to new clusters(e.g. for scaling up)

**How did you test it?**
Existing tests

**Potential risks**
No. It's been used for a very long time.

**Release notes**
Need to add release notes for migrating to update config

**Documentation Changes**
N/A

### DIFF
--- a/common/domain/handler.go
+++ b/common/domain/handler.go
@@ -128,7 +128,7 @@ func (d *handlerImpl) RegisterDomain(
 	if !d.clusterMetadata.IsGlobalDomainEnabled() {
 		if registerRequest.IsGlobalDomain {
 			return &types.BadRequestError{Message: "Cannot register global domain when not enabled. Please update config to enable global domain(recommended), " +
-				"or use \"--global_domain false\" to create legacy local domain. Global domain of single cluster has zero overhead, but only advantages for future migration and fail over. Please check Cadence documentation for more details."}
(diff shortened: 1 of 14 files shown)


longquanzheng:
@yycptt @yux0 this is based on what we discussed offline. 
Uber has been using global domain as default for a long time. I think we should do the same for OSS users.

common/domain/handler.go
```diff
@@ -127,7 +127,8 @@ func (d *handlerImpl) RegisterDomain(
 
 	if !d.clusterMetadata.IsGlobalDomainEnabled() {
 		if registerRequest.IsGlobalDomain {
-			return &types.BadRequestError{Message: "Cannot register global domain when not enabled"}
+			return &types.BadRequestError{Message: "Cannot register global domain when not enabled. Please update config to enable global domain(recommended), " +
```
yux0, line 130: Is this too detail info? the --global_domain belongs to domain CLI?

yux0 (approved):

common/domain/handler.go
longquanzheng: yeah good point. I will remove `--global_domain` from the error message.

PR #4283, opened 2021-06-23
Remove unnecessary nonzero requirements in some fields of NoSQL config struct

longquanzheng:
**What changed?**
Remove unnecessary nonzero fields in NoSQL config struct

**Why?**
1. PluginName requirement will break existing users
2. Keyspace will not be generically applicable for all NoSQL DBs

**How did you test it?**
N/A

**Potential risks**
No

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -170,7 +170,7 @@ type (
 	// NoSQL contains configuration to connect to NoSQL Database cluster
 	NoSQL struct {
 		// PluginName is the name of NoSQL plugin, default is "cassandra". Supported values: cassandra
-		PluginName string `yaml:"pluginName" validate:"nonzero"`
+		PluginName string `yaml:"pluginName"`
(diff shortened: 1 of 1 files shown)


yycptt (approved):

Groxx:
was this actually necessary? with #4284 , the cassandra config is correctly ported into the new structure.

or more generally: is it a problem if someone doesn't define a plugin name? I'm not sure what that'd result in elsewhere.

yycptt:
> was this actually necessary? with #4284 , the cassandra config is correctly ported into the new structure.
> 
> or more generally: is it a problem if someone doesn't define a plugin name? I'm not sure what that'd result in elsewhere.

I think the zero value error happens when creating the config struct using yaml library which happens before the logic in #4284.

Groxx:
Ah, yes, good point.  It'd still be necessary in that case.

PR #4286, opened 2021-06-28
Add client and handler implementation for cross-cluster task APIs

yycptt:
**What changed?**
- Add client implementation for cross-cluster task APIs
- Add handler implementation for admin GetCrossClusterTasks
- Add dummy implementation for history Get/RespondCrossClusterTasks

**Why?**
- For supporting cross cluster operations

**How did you test it?**
- The fact that code builds means the all wiring up and type conversions are done correctly. 
- Unit test for GetCrossClusterTasks in history client impl will be added after the todo in the impl is resolved.

**Potential risks**

**Release notes**

**Documentation Changes**


client/history/client.go
```diff
@@ -1020,6 +1015,122 @@ func (c *clientImpl) NotifyFailoverMarkers(
...
+			if common.IsServiceBusyError(futureErr) {
+				err = futureErr
+			}
+		} else {
+			for shardID, tasks := range resp.TasksByShard {
+				response.TasksByShard[shardID] = tasks
+			}
+		}
+	}
+	wg.Wait()
```
yux0, line 1081: should this before future get?

yycptt: Both will work I think?

I guess I can probably just remove the `wg`, once all futures are unblocked those goroutines will eventually be completed.

yycptt: Updated. Removed the wg usage and move the logic for reducing ctx timeout before creating the futures.

client/history/client.go
```diff
@@ -1020,6 +1015,121 @@ func (c *clientImpl) NotifyFailoverMarkers(
...
+			//
+			// for _, failedShardID := range requestByClient[client].ShardIDs {
+			// 	response.FailedCauseByShard[failedShardID] = ...
+			// }
+			//
+			// for now following the pattern for getting replication tasks:
+			// ignore errors other than service busy, so that task fetcher in target
+			// cluster can slow down.
+			if common.IsServiceBusyError(futureErr) {
+				err = futureErr
```
demirkayaender, line 1071: The behavior here is potentially to return a non-empty reponse and an error together since we aren't assigning nil to response when the error is set. If this is the intended behavior it would be good to put a comment on top of the `GetCrossClusterTasks` function to explain the contract. 

If you wanted to return nil when the error is set, then you can just say the for loop continues so we can make sure all the futures complete.

demirkayaender (approved):
My comment is more like a nit, the rest of the changes look go to me, so approving to unblock but it's nice to address the comment.

client/history/client.go
yycptt: It will wait until all futures complete since I am not breaking the for loop here I think. I just set the err value here when the error type is service busy. 

Yeah, so the plan for this diff is to return an error in this case. I will soon update the definition for the response struct to return per shard error as part of the response (as indicated by the TODO comment). Then we can return nil error and put error details in the response.

PR #4290, opened 2021-07-02
Clean up linting warnings

yux0:
**What changed?**
Clean up linting warnings

This is the part 1. Peresistence fold linting is pending after the nosql refactoring is done.

**Why?**
Too many linting warnings

**How did you test it?**
N/A

**Potential risks**
N/A

**Release notes**
N/A

**Documentation Changes**


host/ndc/integration_test.go
```diff
@@ -87,15 +89,34 @@ var (
 	}
 )
 
-func NewNDCIntegrationTestSuite(params NDCIntegrationTestSuiteParams) *NDCIntegrationTestSuite {
```
yycptt, line 90: I am not quite sure if we are using this function and the two struct definitions above in our internal repo. If so, I think we need to keep them exported.

service/matching/task.go
```diff
@@ -42,10 +42,10 @@ type (
 		decisionTaskInfo *types.MatchingPollForDecisionTaskResponse
 		activityTaskInfo *types.PollForActivityTaskResponse
 	}
-	// internalTask represents an activity, decision, query or started (received from another host).
+	// InternalTask represents an activity, decision, query or started (received from another host).
 	// this struct is more like a union and only one of [ query, event, forwarded ] is
 	// non-nil for any given task
-	internalTask struct {
+	InternalTask struct {
```
yycptt, line 48: Trying to understand why we need to export this type. Looks like an internal type definition for matching package to me?

host/client_integration_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
yux0: updated

host/ndc/integration_test.go
yux0: Revert it to be exported

service/matching/task.go
yux0: Yes. Other components in matching are already exported so I make this struct exported as well. My plan is to do a refactor in matching (similar to the history, to group the components in sub packages and export them)

yycptt (approved):

PR #3975, opened 2021-02-10
Fix some racy reads in tests

Groxx:
`PollForDecisionTask` is executed on a different goroutine, and these reads are not synchronized,
so these race and cause fairly-common test failures.

Using atomics is a pretty simple way around this, as is simply removing the strange `Do(...).Return(...)` cases and replacing them with `DoAndReturn`.

Unfortunately there's still a category of races in here that are not too easy to resolve without risking deadlocks:
```
WARNING: DATA RACE
Write at 0x00c000e81310 by goroutine 124:
  github.com/stretchr/testify/suite.(*Suite).SetT()
      /home/user/gopath/pkg/mod/github.com/stretchr/testify@v1.5.1/suite/suite.go:36 +0xbc
  github.com/uber/cadence/service/matching.(*MatcherTestSuite).SetT()
      <autogenerated>:1 +0x4d
  github.com/stretchr/testify/suite.Run.func2.1()
      /home/user/gopath/pkg/mod/github.com/stretchr/testify@v1.5.1/suite/suite.go:135 +0xb9
  github.com/stretchr/testify/suite.Run.func2()
      /home/user/gopath/pkg/mod/github.com/stretchr/testify@v1.5.1/suite/suite.go:138 +0x38f
  testing.tRunner()
      /opt/go/root/src/testing/testing.go:1127 +0x202

Previous read at 0x00c000e81310 by goroutine 159:
  github.com/uber/cadence/service/matching.(*MatcherTestSuite).TestMustOfferRemoteMatch.func4()
      /home/user/gopath/src/github.com/uber/cadence/service/matching/matcher_test.go:359 +0xab
  github.com/uber/cadence/service/matching.ensureAsyncReady.func1()
      /home/user/gopath/src/github.com/uber/cadence/service/matching/matcher_test.go:482 +0xec
```
but this only happened after ~90-100 test runs, so it's at least MUCH less common.
Fundamentally it means tests need to wait on the async goroutines' completion, which could deadlock if a test fails.  That's probably fine (the go test timeout will still kill it eventually), but it's a bit more annoying than I want to combine in this one commit.

PR #4291, opened 2021-07-02
Fix a racy read in test

Groxx:
Unfortunately I seem to have missed this `wait()` call in #3975, and it decided to assert itself recently:
https://buildkite.com/uberopensource/cadence-server/builds/11015#bf06bca6-2ad3-49fd-a063-de1dd67655b9
(races between line 363 and the suite's `SetT()` method)

This should fix it.  There aren't any other missing-`wait()`s AFAICT.


mkolodezny (approved):
LGTM, thanks!

PR #4292, opened 2021-07-02
Target cluster cross cluster task processor

yycptt:
**What changed?**
- Add cross cluster task processor for processing those tasks at target cluster

**Why?**
- For supporting cross domain/cluster operations

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/fetcher.go
```diff
@@ -104,12 +108,31 @@ func NewCrossClusterTaskFetchers(
...
+	requestByShard map[int32]fetchRequest,
+) (map[int32]interface{}, error) {
+	adminClient := clientBean.GetRemoteAdminClient(sourceCluster)
+	shardIDs := make([]int32, 0, len(requestByShard))
+	for shardID := range requestByShard {
+		shardIDs = append(shardIDs, shardID)
+	}
+	request := &types.GetCrossClusterTasksRequest{
+		ShardIDs:      shardIDs,
+		TargetCluster: currentCluster,
```
demirkayaender, line 122: Are you planning to set the maxTaskID here sometime in the future? If we don't set, the response might be too big in case there are lots of tasks in the queue?

If you are planning to add that maybe you can add a comment here.

service/history/task/cross_cluster_task.go
```diff
@@ -73,9 +76,35 @@ type (
...
 	}
 )
 
+// NewCrossClusterTaskForTargetCluster creates a CrossClusterTask
+// for the processing logic at target cluster
+// the returned the Future will be unblocked when after the task
+// is processed. The future value has type types.CrossClusterTaskResponse
+// and there will be not error returned for this future. All errors will
+// be recorded by the FailedCause field in the response.
+func NewCrossClusterTaskForTargetCluster(
```
demirkayaender, line 89: Seems like not used, did you need to add this in this PR?

service/history/task/cross_cluster_task_processor_test.go
```diff
@@ -0,0 +1,292 @@
...
+			select {
+			case <-processor.shutdownCh:
+			default:
+				close(processor.shutdownCh)
+			}
+
+			s.Len(request.TaskResponses, completedTasks)
+			s.Equal(s.mockShard.Resource.GetClusterMetadata().GetCurrentClusterName(), request.TargetCluster)
+			s.Equal(s.mockShard.GetShardID(), int(request.GetShardID()))
+			// s.False(request.GetFetchNewTasks())
```
demirkayaender, line 159: Did you want to keep this?

demirkayaender (approved):
Looks like the initial work on the fetcher side. The changes look good to me. Left couple comments but none of them is blocking.

service/history/task/fetcher.go
yycptt: My currently plan is let the source cluster decide how many tasks should be returned in one response. So if there's lots of tasks in the queue, they will be fetched in batches.

service/history/task/cross_cluster_task.go
yycptt: It's used at L222 in `cross_cluster_task_processor.go`. Not sure if you review that file, it's collapsed by default I think.

service/history/task/cross_cluster_task_processor_test.go
yycptt: Yeah, will uncomment this after the idl change is landed.

service/history/task/cross_cluster_task_processor_test.go
```diff
@@ -0,0 +1,292 @@
...
+			return submitted, nil
+		},
+	).Times(numTasks - 1) // -1 since there's a duplicate task
+
+	s.mockShard.Resource.HistoryClient.EXPECT().RespondCrossClusterTasksCompleted(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(_ context.Context, request *types.RespondCrossClusterTasksCompletedRequest) (*types.RespondCrossClusterTasksCompletedResponse, error) {
+			s.Len(request.TaskResponses, completedTasks)
+			s.Equal(s.mockShard.Resource.GetClusterMetadata().GetCurrentClusterName(), request.TargetCluster)
+			s.Equal(s.mockShard.GetShardID(), int(request.GetShardID()))
+			// s.True(request.GetFetchNewTasks())
```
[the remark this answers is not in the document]
yycptt: Yes, will uncomment this after the idl change is landed.

service/history/task/cross_cluster_task_processor_test.go
```diff
@@ -0,0 +1,292 @@
...
+
+	completedTasks := 0
+	s.mockProcessor.EXPECT().TrySubmit(gomock.Any()).DoAndReturn(
+		func(t Task) (bool, error) {
+			submitted := rand.Intn(2) == 0
+			if submitted {
+				completedTasks++
+				t.Ack()
+			}
+			// redispatcher interval is set to 1hr, basically disabled
```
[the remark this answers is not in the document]
yycptt: If a task is not accepted by the mockProcessor, it will be submitted to the redispatcher so that the submission can be retried later. 

For this test I set the time interval for redispatch to 1hr which basically disabled redispatcher so that I can easily count how many tasks are submitted/completed by the mockProcessor. 
If the redispatcher is working and re-submitting tasks from time to time, then the logic for counting exactly how many tasks are submitted will be more complicated.

service/history/task/cross_cluster_task_processor_test.go
```diff
@@ -0,0 +1,292 @@
...
+			return submitted, nil
+		},
+	).Times(numTasks - 1) // -1 since there's a duplicate task
+
+	s.mockShard.Resource.HistoryClient.EXPECT().RespondCrossClusterTasksCompleted(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(_ context.Context, request *types.RespondCrossClusterTasksCompletedRequest) (*types.RespondCrossClusterTasksCompletedResponse, error) {
```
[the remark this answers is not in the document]
yycptt: Make sense. Done.

service/history/task/fetcher.go
demirkayaender: Thanks for the explanation. Could you add this as a comment.

service/history/task/cross_cluster_task.go
yux0: What do you think to have the future return in all NewCrossClusterXXXTask() and no need for the NewCrossClusterTaskForTargetCluster?

service/history/task/cross_cluster_task.go
```diff
@@ -73,9 +76,35 @@ type (
...
+	shard shard.Context,
+	taskRequest *types.CrossClusterTaskRequest,
+	logger log.Logger,
+	maxRetryCount dynamicconfig.IntPropertyFn,
+) (CrossClusterTask, future.Future) {
```
yux0, line 94: What do you think to have the future return in all NewCrossClusterXXXTask() and no need for the NewCrossClusterTaskForTargetCluster?

yux0 (approved):

service/history/task/fetcher.go
yycptt: Sure. Added it as comment.

service/history/task/cross_cluster_task.go
yycptt: Yeah, I agree that future can be returned in all NewCrossClusterXXXTask() functions. But I guess there should still be a dedicated constructor for tasks in target cluster as the input parameter will be quite different? Either the current `CrossClusterTaskRequest` or some more detailed fields like signal input, childworkflow start attribute, task state etc. For tasks in the source cluster, those won't be necessary as they can be loaded from mutable state or events cache in source cluster. But in target cluster, we have to pass them in when creating the task.

PR #4298, opened 2021-07-06
Refactoring Cassandra workflow persistence manager for NoSQL support-Part 3

longquanzheng:
**What changed?**
1. Create NoSQL interface for other remaining methods
2. Extract the logic of Cassandra to implement the interface
3. Replace executionStore with NoSQL plugin db style
4. Remove unused code

**Why?**
For https://github.com/uber/cadence/issues/3514

**How did you test it?**
existing tests. 

**Potential risks**
Low to medium risks.
Even though the goal of the PR is to keep the same behavior, it's still some uknown and this is the critical persistence for Cadence. 

**Release notes**
No

**Documentation Changes**
No


common/persistence/nosql/nosqlplugin/interfaces.go
```diff
@@ -49,13 +49,13 @@ type (
 	// Cassandra implementation cannot do it due to backward-compatibility. Any other NoSQL implementation should use datablob for non-significant columns.
 	// Follow the comment for each tableCRUD for what are 'significant' columns.
 	tableCRUD interface {
-		historyEventsCRUD
```
yux0, line 52: Could we add a mockgen for this interface file?

common/persistence/nosql/nosqlExecutionStore.go
```diff
@@ -0,0 +1,831 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
yux0, line 1: also update this?

common/persistence/nosql/nosqlExecutionStore.go
```diff
@@ -0,0 +1,831 @@
...
+	return nil
+}
+
+func (d *nosqlExecutionStore) DeleteCurrentWorkflowExecution(
+	ctx context.Context,
+	request *p.DeleteCurrentWorkflowExecutionRequest,
+) error {
+	err := d.db.DeleteCurrentWorkflow(ctx, d.shardID, request.DomainID, request.WorkflowID, request.RunID)
+	if err != nil {
+		return convertCommonErrors(d.db, "DeleteWorkflowCurrentRow", err)
```
yux0, line 492: nit: do we want to just use the same as the method name? DeleteCurrentWorkflow

common/persistence/nosql/nosqlExecutionStore.go
```diff
@@ -0,0 +1,831 @@
...
+	return nil
+}
+
+func (d *nosqlExecutionStore) GetReplicationTasksFromDLQ(
+	ctx context.Context,
+	request *p.GetReplicationTasksFromDLQRequest,
+) (*p.InternalGetReplicationTasksFromDLQResponse, error) {
+	tasks, nextPageToken, err := d.db.SelectReplicationDLQTasksOrderByTaskID(ctx, d.shardID, request.SourceClusterName, request.BatchSize, request.NextPageToken, request.ReadLevel, request.MaxReadLevel)
+	if err != nil {
+		return nil, convertCommonErrors(d.db, "GetReplicationTasks", err)
```
yux0, line 751: nit: GetReplicationTasksFromDLQ

common/persistence/nosql/nosqlplugin/interfaces.go
longquanzheng: We can add it when we are writing unit tests, it's pretty easy to have. 
But I don't think we should do it now as we don't have any unit tests here.
Because having mockgen would just become an unnecessary burden to maintain the code.

common/persistence/nosql/nosqlExecutionStore.go
longquanzheng: yeah sure.

common/persistence/nosql/nosqlExecutionStore.go
longquanzheng: ohh right, good catch

common/persistence/nosql/nosqlExecutionStore.go
longquanzheng: yeah we should do that.

common/persistence/nosql/nosqlplugin/interfaces.go
yux0: I will try to add some unit tests later

yux0 (approved):

common/persistence/nosql/nosqlExecutionStore.go
```diff
@@ -0,0 +1,831 @@
...
+)
+
+// NewExecutionStore is used to create an instance of ExecutionStore implementation
+func NewExecutionStore(
+	shardID int,
+	client gocql.Client,
+	session gocql.Session,
+	logger log.Logger,
+) (p.ExecutionStore, error) {
+	db := cassandra.NewCassandraDBFromSession(client, session, logger)
```
yycptt, line 70: nit: shall we add some TODO here indicating we need to support other nosql plugin as well here?

yycptt (approved):

common/persistence/nosql/nosqlExecutionStore.go
longquanzheng: yeah good catch.

Groxx:
Hmm.  Could this be causing problems with the docker image?  client integration tests are failing for what seem to be server reasons: https://github.com/uber-go/cadence-client/pull/1111
```
2021-07-09T23:47:26.190186742Z 2021/07/09 23:47:26 Loading config; env=docker,zone=,configDir=/etc/cadence/config
2021-07-09T23:47:26.190200929Z 2021/07/09 23:47:26 Loading configFiles=[/etc/cadence/config/docker.yaml]
2021-07-09T23:47:26.191787925Z 2021/07/09 23:47:26 Config file corrupted.yaml: unmarshal errors:
2021-07-09T23:47:26.191799021Z   line 71: field protocolVersion not found in type config.NoSQL
2021-07-09T23:47:26.191802380Z   line 79: field protocolVersion not found in type config.NoSQL
```

longquanzheng:
> field protocolVersion not found in type config.NoSQL

Not this PR, but from another one. I have a fix here: https://github.com/uber/cadence/pull/4310

Sorry for breaking it. I think we need some nightly integ test run int client repo to catch this issue earlier(it's hard to do it in server because the docker image is only uploaded after the PR is merged. What do you think?

Groxx:
Yea, there's some missing coupling between client and server tests that'd be good to fix, since they inherently rely on each other.  E.g. server changes should probably run the whole client suite of tests as well.

I'm not sure how to achieve that tho ¯\_(ツ)_/¯  I assume it's possible since it's just a bunch of docker containers, but I haven't poked at buildkite before.

PR #4301, opened 2021-07-07
Dynamic replication batch size

yux0:
**What changed?**
Dynamic replication batch size

**Why?**
The replication batch size should be calculated based on the backlog in a particular shard. This could help with hot shards

**How did you test it?**
Unit tests
TODO: bench tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/replication/task_ack_manager.go
+++ b/service/history/replication/task_ack_manager.go
@@ -29,6 +29,7 @@ import (
 	ctx "context"
 	"errors"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/uber/cadence/common"
@@ -52,6 +53,8 @@ var (
 	errUnknownQueueTask       = errors.New("unknown task type")
 	errUnknownReplicationTask = errors.New("unknown replication task")
 	defaultHistoryPageSize    = 1000
+	minReadTaskSize           = 20
+	maxReplicationLatency     = int64(40)
 )
 
 type (
@@ -76,6 +79,9 @@ type (
 		historyManager   persistence.HistoryManager
 		rateLimiter      *quotas.DynamicRateLimiter
 		retryPolicy      backoff.RetryPolicy
+
+		taskLock             sync.Mutex
+		lastTaskCreationTime time.Time
 
 		metricsClient metrics.Client
 		logger        log.Logger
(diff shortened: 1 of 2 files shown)


service/history/replication/task_ack_manager.go
```diff
@@ -52,6 +53,8 @@ var (
 	errUnknownQueueTask       = errors.New("unknown task type")
 	errUnknownReplicationTask = errors.New("unknown replication task")
 	defaultHistoryPageSize    = 1000
+	minReadTaskSize           = 20
+	maxReplicationLatency     = int64(40)
```
vytautas-karpavicius, line 57: I suggest adding time unit suffix (I guess `Seconds`) or even better use time.Duration instead. As of now it is not clear what 40 means.

service/history/replication/task_ack_manager.go
```diff
@@ -625,6 +640,24 @@ func (t *taskAckManagerImpl) generateHistoryReplicationTask(
 	)
 }
 
+func (t *taskAckManagerImpl) getBatchSize() int {
+
+	t.taskLock.Lock()
+	taskLatency := int64(time.Now().Sub(t.lastTaskCreationTime) / time.Second)
+	t.taskLock.Unlock()
```
demirkayaender, line 647: Trying to understand how useful this lock is: it feels like having and not having the lock here are the same to me. Could you you explain a bit what it is achieving here?

Are the time assignments in go not atomic that if we try to read its value while assigning somewhere else we get a crash?

yux0: Most likely it will not be used. I added it to prevent concurrent calls to the method. Updated it with atomic.Value

common/dynamicconfig/constants.go
```diff
@@ -1258,6 +1258,12 @@ const (
 	// Default value: FALSE
 	// Allowed filters: N/A
 	ReplicatorProcessorEnablePriorityTaskProcessor
+	// ReplicatorUpperLatency is indicates the max allowed replication latency between clusters
```
yycptt, line 1261: nit: also indicate the unit for the number? I guess default 40 means 40 seconds?

service/history/config/config.go
```diff
@@ -459,6 +460,7 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
 		ReplicatorProcessorMaxRedispatchQueueSize:             dc.GetIntProperty(dynamicconfig.ReplicatorProcessorMaxRedispatchQueueSize, 10000),
 		ReplicatorProcessorEnablePriorityTaskProcessor:        dc.GetBoolProperty(dynamicconfig.ReplicatorProcessorEnablePriorityTaskProcessor, false),
 		ReplicatorProcessorFetchTasksBatchSize:                dc.GetIntPropertyFilteredByShardID(dynamicconfig.ReplicatorTaskBatchSize, 25),
+		ReplicatorUpperLatency:                                dc.GetIntProperty(dynamicconfig.ReplicatorTaskBatchSize, 40),
```
yycptt, line 463: `dynamicconfig.ReplicatorUpperLatency`?

also, shall we use `dc.GetDurationProperty` instead?

service/history/replication/task_ack_manager.go
```diff
@@ -101,16 +106,21 @@ func NewTaskAckManager(
 	retryPolicy.SetMaximumAttempts(config.ReplicatorReadTaskMaxRetryCount())
 	retryPolicy.SetBackoffCoefficient(1)
 
+	taskCreatedTime := atomic.Value{}
+	taskCreatedTime.Store(time.Now())
```
yycptt, line 110: nit: use `shard.GetTimesource().Now()`?

If there's a replication backlog on startup, the batch size for the first response will likely to be small I think? Probably we can always use defaultBatchSize for the first response by storing an empty time here?

service/history/replication/task_ack_manager.go
```diff
@@ -625,6 +640,22 @@ func (t *taskAckManagerImpl) generateHistoryReplicationTask(
 	)
 }
 
+func (t *taskAckManagerImpl) getBatchSize() int {
+
+	taskLatency := int(time.Now().Sub(t.lastTaskCreationTime.Load().(time.Time)) / time.Second)
```
yycptt, line 645: nit: `t.shard.GetTimesource().Now()`

yux0: updated

service/history/replication/task_ack_manager.go
yux0: make sense. use default batch size on first response

service/history/config/config.go
yux0: updated

common/dynamicconfig/constants.go
yux0: updated

PR #4302, opened 2021-07-07
Skip updating rows which are deleted in the same transaction

Shaddoll:
**What changed?**
Skip updating activityInfo, timerInfo, childexecutionInfo, signalInfo, requestCancelInfo, signalRequestSets rows which are deleted in the same transaction.

**Why?**
The row being updated is deleted in the same transaction, so we actually don't need the update.

**How did you test it?**
existing tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/sql/workflowStateMaps.go
+++ b/common/persistence/sql/workflowStateMaps.go
@@ -45,8 +45,12 @@ func updateActivityInfos(
 ) error {
 
 	if len(activityInfos) > 0 {
-		rows := make([]sqlplugin.ActivityInfoMapsRow, len(activityInfos))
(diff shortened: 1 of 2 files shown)


yux0:
Could we just remove the request ID from updateSignalRequestedIDs when deleting the request?

Shaddoll:
> Could we just remove the request ID from updateSignalRequestedIDs when deleting the request?

What do you mean? Not deleting the request?

yycptt:
Shall we perform this logic when closing the transaction in mutable state builder instead of the persistence layer for sql, so that all persistence implementation can benefit from it?

PR #4304, opened 2021-07-07
Update Mutable State to reduce unnecessary update to DB

Shaddoll:
**What changed?**
Picking https://github.com/temporalio/temporal/pull/1463

- Reduce unnecessary update to DB if deletion is specified
- Refactor Mutable State Builder

**Why?**
Perf optimization

**How did you test it?**
existing tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -83,28 +83,28 @@ var (
 
 type (
 	mutableStateBuilder struct {
-		pendingActivityInfoIDs     map[int64]*persistence.ActivityInfo    // Schedule Event ID -> Activity Info.
-		pendingActivityIDToEventID map[string]int64                       // Activity ID -> Schedule Event ID of the activity.
-		updateActivityInfos        map[*persistence.ActivityInfo]struct{} // Modified activities from last update.
-		deleteActivityInfos        map[int64]struct{}                     // Deleted activities from last update.
-		syncActivityTasks          map[int64]struct{}                     // Activity to be sync to remote
+		pendingActivityInfoIDs     map[int64]*persistence.ActivityInfo // Schedule Event ID -> Activity Info.
+		pendingActivityIDToEventID map[string]int64                    // Activity ID -> Schedule Event ID of the activity.
+		updateActivityInfos        map[int64]*persistence.ActivityInfo // Modified activities from last update.
+		deleteActivityInfos        map[int64]struct{}                  // Deleted activities from last update.
+		syncActivityTasks          map[int64]struct{}                  // Activity to be sync to remote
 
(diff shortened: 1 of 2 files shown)


service/history/execution/mutable_state_builder.go
```diff
@@ -4707,3 +4709,41 @@ func (e *mutableStateBuilder) logDataInconsistency() {
 		tag.WorkflowRunID(runID),
 	)
 }
+
+func (e *mutableStateBuilder) closeTransactionAsMutationRemoveDeletedInfos() {
```
demirkayaender, line 4713: I think we are fixing the symptom here; do we know why we added updates when we also wanted to delete in the first place?

Shaddoll: I think it's just because we didn't delete it from update info when we added delete info.
So this would gurantee that.

yux0:
Took a look at the CloseTransactionAsMutation. My understanding is that the same record in update and delete is not expected. should we just add a sanity check? do you see any case that has the record in both update and delete?

yycptt:
> Took a look at the CloseTransactionAsMutation. My understanding is that the same record in update and delete is not expected. should we just add a sanity check? do you see any case that has the record in both update and delete?

Wenquan told me they found one case related to activity retry that will result in updating/deleting the same record. I haven't look into it though.

yycptt:
https://github.com/temporalio/temporal/pull/1463

yux0:
> [temporalio/temporal#1463](https://github.com/temporalio/temporal/pull/1463)

Yeah. Could we do something similar to update the list in mutable state builder?

yycptt:
> do you see any case that has the record in both update and delete?

I am thinking about workflow reset, when reset, we apply all events in history and when encounter an activity start event, we add an entry in update list, later when applying activity completed event, we add an entry in delete list for the same activity. Then after applying all event, we do one db operation to reset the workflow which will result in this case I think. 

so my understanding is there're cases this situation will happen and it's expected, and we need this change to do the optimization. 

If replication apply multiple event batches in one transaction then the same will happen I think?

demirkayaender:
Probably you want to do it as how it was done in the temporal PR but still keep the changes here as a sanity check and log error anytime we encounter a duplicate?

service/history/execution/mutable_state_builder.go
```diff
@@ -1360,6 +1363,7 @@ func (e *mutableStateBuilder) DeleteActivity(
 		e.logDataInconsistency()
 	}
 
+	delete(e.updateActivityInfos, scheduleEventID)
```
demirkayaender, line 1366: Assuming DeleteActivity should always be called after UpdateActivity (otherwise it would be a bug).

demirkayaender (approved):
lgtm

PR #4315, opened 2021-07-14
Optimization for start child workflow task

yux0:
**What changed?**
Optimization for start child workflow task

**Why?**
Pick https://github.com/temporalio/temporal/pull/1397

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


yycptt (approved):

service/history/task/transfer_active_task_executor.go
```diff
@@ -698,6 +703,11 @@ func (t *transferActiveTaskExecutor) processStartChildExecution(
 	if err != nil {
 		return err
 	}
+
+	// NOTE: do not access anything related mutable state after this lock release
+	// release the context lock since we no longer need mutable state builder and
+	// the rest of logic is making RPC call, which takes time.
+	release(nil)
```
Groxx, line 710: is there any way to detect (and probably fail?) if this occurs?

yux0: no logic between the lock release and return. Should be easy to detect. In worst case, it will return data inconsistency error

PR #4317, opened 2021-07-19
Add cross cluster target task executor

yycptt:
**What changed?**
- Implemented executor for cross-cluster target task
- Refactored definition for cross-cluster tasks and categorize them by source/target instead of task type

**Why?**
- For supporting cross-cluster task

**How did you test it?**
- Unit tests.

**Potential risks**
N/A, not enabled yet

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -0,0 +1,321 @@
...
+	var err error
+	switch task.GetTaskType() {
+	case persistence.CrossClusterTaskTypeStartChildExecution:
+		response.StartChildExecutionAttributes, err = t.executeStartChildExecutionTask(ctx, targetTask)
+	case persistence.CrossClusterTaskTypeCancelExecution:
+		response.CancelExecutionAttributes, err = t.executeCancelExecutionTask(ctx, targetTask)
+	case persistence.CrossClusterTaskTypeSignalExecution:
+		response.SignalExecutionAttributes, err = t.executeSignalExecutionTask(ctx, targetTask)
+	default:
+		err = errUnknownTransferTask
```
demirkayaender, line 124: Can we use our own errors? `errUnknownCrossClusterTask`

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	var retryable bool
+	response.FailedCause, retryable = t.convertErrorToFailureCause(err)
+	targetTask.response = response
+
+	// only return retryable/unexpected error so that
```
demirkayaender, line 131: Can we log the errors here since we are swallowing it if it's not retryable?

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -0,0 +1,321 @@
...
+func (t *crossClusterTargetTaskExecutor) executeStartChildExecutionTask(
+	ctx context.Context,
+	task *crossClusterTargetTask,
+) (*types.CrossClusterStartChildExecutionResponseAttributes, error) {
+	attributes := task.request.StartChildExecutionAttributes
+	if attributes == nil {
+		return nil, errMissingTaskRequestAttributes
+	}
+
+	targetDomainName, err := t.verifyAndTargetDomainName(attributes.TargetDomainID)
```
demirkayaender, line 149: Is this a naming convention we use instead of `verifyAndGetTargetDomainName `?

Also, you can probably remove `Target` from the function name? It doesn't seem to compare source and target domains.

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -0,0 +1,321 @@
...
+	if err == errDomainStandby {
+		return types.CrossClusterTaskFailedCauseDomainNotActive.Ptr(), false
+	}
+	if err == ErrTaskPendingActive {
+		return types.CrossClusterTaskFailedCauseDomainNotActive.Ptr(), true
+	}
+	if err == errUnknownTaskProcessingState || err == errMissingTaskRequestAttributes {
+		return types.CrossClusterTaskFailedCauseUncategorized.Ptr(), false
+	}
+	return types.CrossClusterTaskFailedCauseUncategorized.Ptr(), true
```
demirkayaender, line 296: nit: you can use switch/case here.

service/history/task/cross_cluster_task.go
```diff
@@ -37,136 +40,146 @@ import (
...
+			resubmitTaskFn,
+			maxRetryCount,
+		),
 	}
 }
 
-// NewCrossClusterCancelWorkflowTask initialize cross cluster cancel workflow task and task future
-func NewCrossClusterCancelWorkflowTask(
+// NewCrossClusterTaskForTargetCluster creates a cross cluster task
+// for the processing it at the target cluster
```
demirkayaender, line 114: This function is called at the target all the time right? Can you change the comment like 

```
// NewCrossClusterTaskForTargetCluster is called at the target cluster to 
// process the cross cluster task
```

Confusion in the original comment is that we don't know if we are going to send it from the source or if we are already at the target cluster.

demirkayaender: nit: Similarly for the name, you might want to change it to `NewCrossClusterTargetTask`

service/history/task/cross_cluster_task_processor.go
```diff
@@ -222,7 +228,12 @@ func (p *crossClusterTaskProcessor) processTaskRequests(
 			crossClusterTask, future := NewCrossClusterTaskForTargetCluster(
 				p.shard,
 				taskRequest,
+				p.taskExecutor,
 				p.logger,
+				func(t Task) {
+					// the only possible error is processor shutdown which is safe to discard
```
demirkayaender, line 234: looks like there are some other errors but they were logged in processor_base.go? Even if it's a different error, we don't want to change how this function behaves right?

demirkayaender:
Overall, changes look good to me but this will conflict with my RecordChildCompletion task changes. Let's get this one in whenever we can.

service/history/task/cross_cluster_target_task_executor.go
yycptt: Updated.

service/history/task/cross_cluster_target_task_executor.go
yycptt: Renamed to `verifyDomainActive`

service/history/task/cross_cluster_target_task_executor.go
yycptt: I think logging all non-retryable errors will add too much noise to the logs, as some errors types like entity not exists or already started are kind of expected when the same task is processed multiple times. So I added a log for all uncategorized errors. For expected error types source cluster will get the failure cause and log the error if needed.

service/history/task/cross_cluster_task.go
yycptt: Done.

service/history/task/cross_cluster_task_processor.go
yycptt: Yes, `p.submitTask()` only return error during shutdown. Other errors will be logged and the task will be push to a re-dispatch queue for retrying the submission. Here we just need to make sure the `submitTask` is invoked so that task is either submitted or pushed to re-dispatch queue.

demirkayaender (approved):
Looks good to me!

PR #4320, opened 2021-07-20
Add Permissions to Attributes and reading Public/Private key from disk instead of reading it from yaml file

noiarek:
**What changed?**
Adding Permissions to attributes and reading public / private key from path instead of adding it to the config file

**Why?**
Proposal was defined here https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#heading=h.iar4if2w5gus. Continuation of the first step https://github.com/uber/cadence/pull/4306

**How did you test it?**
Unit Test

**Potential risks**
Not AFAIK

**Release notes**
Yes when the whole proposal is implemented

**Documentation Changes**
Yes when the whole proposal is implemented


noiarek:
@longquanzheng part II of the changes we agreed to move to a different diff

service/frontend/accessControlledHandler.go
```diff
@@ -326,6 +337,7 @@ func (a *AccessControlledWorkflowHandler) PollForActivityTask(
 		APIName:    "PollForActivityTask",
 		DomainName: request.GetDomain(),
 		TaskList:   request.TaskList,
+		Permission: authorization.PermissionRead,
```
longquanzheng, line 340: This is PermissionWrite. 
PollForActivity will get the activity task and also start the activity, which trigger to write the event "ActivityTaskStarted" event in the history.

service/frontend/accessControlledHandler.go
```diff
@@ -350,6 +362,7 @@ func (a *AccessControlledWorkflowHandler) PollForDecisionTask(
 		APIName:    "PollForDecisionTask",
 		DomainName: request.GetDomain(),
 		TaskList:   request.TaskList,
+		Permission: authorization.PermissionRead,
```
longquanzheng, line 365: Same here, this is PermissionWrite

service/frontend/accessControlledHandler.go
```diff
@@ -420,6 +434,7 @@ func (a *AccessControlledWorkflowHandler) RegisterDomain(
 	attr := &authorization.Attributes{
 		APIName:    "RegisterDomain",
 		DomainName: request.GetName(),
+		Permission: authorization.PermissionWrite,
```
longquanzheng, line 437: This should be PermissionAdmin :D

service/frontend/accessControlledHandler.go
```diff
@@ -722,6 +746,7 @@ func (a *AccessControlledWorkflowHandler) GetTaskListsByDomain(
 	attr := &authorization.Attributes{
 		APIName:    "GetTaskListsByDomain",
 		DomainName: request.GetDomain(),
+		Permission: authorization.PermissionAdmin,
```
longquanzheng, line 749: This should be Read :D It's reading the tasklists for the domain.

service/frontend/accessControlledHandler.go
```diff
@@ -745,6 +770,7 @@ func (a *AccessControlledWorkflowHandler) UpdateDomain(
 	attr := &authorization.Attributes{
 		APIName:    "UpdateDomain",
 		DomainName: request.GetName(),
+		Permission: authorization.PermissionRead,
```
longquanzheng, line 773: Same as register domain, any write on domain requires Admin permission

common/util.go
```diff
@@ -931,8 +932,12 @@ func SleepWithMinDuration(desired time.Duration, available time.Duration) time.D
...
 }
 
-func StringToRSAPublicKey(publicKeyString string) (*rsa.PublicKey, error) {
-	block, _ := pem.Decode([]byte(publicKeyString))
+func LoadRSAPublicKey(path string) (*rsa.PublicKey, error) {
+	key, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("invalid public key path %s", path)
+	}
+	block, _ := pem.Decode(key)
```
longquanzheng, line 940: What is the second return value? If it’s error then we should check and return if not nil

longquanzheng (approved):
Lgtm except for one comment about return value

common/util.go
noiarek: https://pkg.go.dev/encoding/pem#Decode it returns the rest of the input after the key

An example here https://golang.org/src/encoding/pem/example_test.go  you can see it returns `"and some more"`  from line 31

PR #4321, opened 2021-07-21
Update admin CLI commands for cross-cluster queue

yycptt:
**What changed?**
- Update admin remove task / describe queue / reset queue command for cross-cluster queue.
- IDL change: https://github.com/uber/cadence-idl/pull/79

**Why?**
- Admin tooling support for cross-cluster queue for dealing with potential issues.

**How did you test it?**
- Tested locally
- Test will pass after IDL change is landed and IDL submodule is updated.

**Potential risks**
- N/A, currently bad request error will be returned for describe/reset queue commands as cross-cluster queue is not initialized.

**Release notes**

**Documentation Changes**


service/history/historyEngine.go
```diff
@@ -2674,6 +2684,18 @@ func (e *historyEngineImpl) ResetTimerQueue(
 	return err
 }
 
+func (e *historyEngineImpl) ResetCrossClusterQueue(
+	ctx context.Context,
+	clusterName string,
+) error {
+	if e.crossClusterProcessor == nil {
```
Shaddoll, line 2691: Do you want to add a comment to remind removing this check when the processor is implemented?

Shaddoll (approved):

service/history/historyEngine.go
yycptt: Makes sense. I added a TODO at the place where`errQueueNotInitialized` is defined.

PR #4325, opened 2021-07-23
Fix admin workflow re-replicate command

yycptt:
**What changed?**
- Fix admin workflow re-replicate command

**Why?**
- It's broken right now

**How did you test it?**
- Tested locally

**Potential risks**

**Release notes**

**Documentation Changes**


service/frontend/adminHandler.go
```diff
@@ -1048,13 +1047,6 @@ func (adh *adminHandlerImpl) validateGetWorkflowExecutionRawHistoryV2Request(
 		return &types.BadRequestError{Message: "Invalid PageSize."}
 	}
 
-	if request.StartEventID == nil &&
-		request.StartEventVersion == nil &&
-		request.EndEventID == nil &&
-		request.EndEventVersion == nil {
-		return &types.BadRequestError{Message: "Invalid event query range."}
-	}
```
yycptt, line 1056: @yux0 Any idea why this case is not allowed? 
From code (https://github.com/uber/cadence/blob/0.22.x/service/frontend/adminHandler.go#L1063) looks like it can be handled. I also tried locally by re-replicating a workflow without specifying any of the four fields and it can work.
Please let me know if I missed anything. Thanks!

tools/cli/adminKafkaCommands.go
```diff
@@ -502,23 +501,23 @@ func AdminRereplicate(c *cli.Context) {
...
+	}
+	if c.IsSet(FlagEndEventVersion) {
+		endVersion = common.Int64Ptr(c.Int64(FlagEndEventVersion))
+	}
 	domainID := getRequiredOption(c, FlagDomainID)
 	wid := getRequiredOption(c, FlagWorkflowID)
 	rid := getRequiredOption(c, FlagRunID)
 	shardID := common.WorkflowIDToHistoryShard(wid, numberOfShards)
 	contextTimeout := defaultResendContextTimeout
 	executionManager := initializeExecutionStore(c, shardID, 0)
```
yycptt, line 518: @yux0 Do we have to do the 2dc workflow check in the command? 
Currently the re-replication command should be issued to the target cluster. But when we use re-replication, typically the workflow doesn't exist in target cluster and we have to specify the DB option for the source cluster in order to get and verify the version history. 
I wonder if the admin GetRawHistory handler already catches the case where the workflow is 2dc and returns a bad request error? If so, I guess we can remove the check here?

yux0: Yes. We don't need it

service/frontend/adminHandler.go
yux0: the start event id/version or end event id/version is required to identify a branch. So this is needed

yux0 (approved):

service/frontend/adminHandler.go
yycptt: Synced offline. This check is not necessary.

tools/cli/adminKafkaCommands.go
yycptt: Removed the check.

PR #4327, opened 2021-07-25
CLI sending authorized request

noiarek:
**What changed?**
`--jwt` flag added to CLI

**Why?**
Requested in [this proposal](https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#heading=h.iar4if2w5gus)

**How did you test it?**
Tested locally
```
./cadence-server --env development_oauth start
```
and
```
./cadence --do samples-domain --jwt eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwicGVybWlzc2lvbiI6InJlYWQiLCJkb21haW4iOiJzYW1wbGVzLWRvbWFpbiIsImlhdCI6MTYyNjMzNjQ2MywiVFRMIjoxMDAwMDAwfQ.ELjxoLPBIIBf3eUEQy-sl0yarhZJ4oj0Bc9ozs79I-5cvg5VKBULxZSomTkc0fZS4en91ZwmLpckXbUzlnEk3WWbKvHqWYBqfuf7DxxxiH8OMaB4wu5NPJ24lz3lYmBbJZGI333FxX9AweC2-Ac3TmMp-E7U1go6LaLSELaXLse__fk8vQMC61COnwu5YGc9fOhaIQ3U9N7qIIGv_oCp3UFOBcUkwQdMHjk3zTrQ60bboGQBMOCBUpe9nz_m_2Bxx4B39qB9DH3fgMPG_-Rsp4M5YT01cW4WujP1QqeGW8rhkUDE2GqjGQi9yla0M7mjJmNjFeRsVTPtBTly1fHANQ domain desc
```

**Potential risks**
Nothing as long as oauth file is not loaded

**Release notes**
When the whole proposal is implemented

**Documentation Changes**
When the whole proposal is implemented


bench/lib/context.go
```diff
@@ -41,6 +41,9 @@ const (
 
 	// CtxKeyCadenceClient is the name of the context key for the cadence client this cadence worker listens to
 	CtxKeyCadenceClient = ContextKey("ctxKeyCadenceClient")
+
+	// CtxKeyJWT is the name of the context key for the JWT
+	CtxKeyJWT = ContextKey("ctxKeyJWT")
```
noiarek, line 46: is `bech/lib` the right place to kave a context key ? I suspect it might be better in the `cli` folder but just placed it there because I found the other ones

tools/cli/app.go
```diff
@@ -59,6 +59,11 @@ func NewCliApp() *cli.App {
 			Usage:  "optional timeout for context of RPC call in seconds",
 			EnvVar: "CADENCE_CONTEXT_TIMEOUT",
 		},
+		cli.StringFlag{
```
noiarek, line 62: This is a global flag. Is it expected to have the flag global?

bench/lib/context.go
longquanzheng: Yeah should not be here. You can put it int factory.go under cli/

common/config/config.go
```diff
@@ -63,7 +63,7 @@ type (
 		// Blobstore is the config for setting up blobstore
 		Blobstore Blobstore `yaml:"blobstore"`
 		// Authorization is the config for setting up authorization
-		Authorization Authorization `yaml:authorization`
+		Authorization Authorization `yaml:"authorization""`
```
longquanzheng, line 66: Seems like there is one more double quote at the end. Probably this cause the config doesn’t work

tools/cli/app.go
longquanzheng: https://github.com/uber/cadence/blob/bec009abeb29fc92a564477efdeb33efe5426170/tools/cli/app.go#L44

When you put a flag here, it’s a global flag. Global means you declare it once for all commands of the CLI tool

tools/cli/domainUtils.go
```diff
@@ -453,3 +453,7 @@ func getConfigDir(c *cli.Context) string {
 	}
 	return dirPath
 }
+
+func getJWT(c *cli.Context) string {
```
longquanzheng, line 457: Maybe in factory.go or others as this will be shared for all commands

config/development_oauth.yaml
```diff
@@ -0,0 +1,8 @@
+authorization:
```
longquanzheng, line 1: Like others, you need to copy the whole file from https://github.com/uber/cadence/blob/master/config/development.yaml and then add this authorization section. 

The `config/development_oauth.yaml` will only override `base.yaml` which only have a section of logging config.

longquanzheng: actually I have to remove some section in 
```
archival:
  history:
    status: "enabled"
    enableRead: true
    provider:
      filestore:
        fileMode: "0666"
        dirMode: "0766"
      gstorage:
        credentialsPath: "/tmp/gcloud/keyfile.json"
  visibility:
    status: "enabled"
    enableRead: true
    provider:
      filestore:
        fileMode: "0666"
        dirMode: "0766"

domainDefaults:
  archival:
    history:
      status: "enabled"
      URI: "file:///tmp/cadence_archival/development"
    visibility:
      status: "enabled"
      URI: "file:///tmp/cadence_vis_archival/development"

```
And also add 
```
system.enableParentClosePolicyWorker:
  - value: false
    constraints: {}
system.enableWorkflowShadower:
  - value: false
    constraints: {}
system.enableFailoverManager:
  - value: false
    constraints: {}
```

So that we don't start any system workflow for now. Because we don't have a way to authorize those request unitil golang client authorizer is implemented. 
I have address those comments myself and pushed to your branch.

tools/cli/flags.go
```diff
@@ -287,6 +288,10 @@ var flagsForExecution = []cli.Flag{
 		Name:  FlagRunIDWithAlias,
 		Usage: "RunID",
 	},
+	cli.StringFlag{
+		Name:  FlagJWT,
+		Usage: "JWT token for authentication",
```
longquanzheng, line 293: this is authorization not authentication :D 

We have a different story for authentication: https://github.com/uber/cadence/issues/4095

config/development_oauth.yaml
```diff
@@ -0,0 +1,116 @@
...
+  filestore:
+    outputDirectory: "/tmp/blobstore"
+
+authorization:
+  oauthAuthorizer:
+    enable: true
+    maxJwtTTL: 600000000
+    jwtCredentials:
+      algorithm: "RS256"
+      publicKey: "common/authorization/keytest.pub"
```
longquanzheng, line 115: Let’s move the two files into a new folder under config/ , eg “credentials”. Because docker and home brew will automatically copy everything under config/ and will not keep the source code after binaries are built. That way users and test this feature with docker compose and homebrew with these test credentials

longquanzheng (approved):

PR #4332, opened 2021-07-28
[WIP] Update OAuth implementation to use domainCache to authorize

longquanzheng:
**What changed?**

**Why?**
Based on discussion with @just-at-uber , we need to make change on the proposal to support multiple domains operation on WebUI with one logging.
The best way to do it is letting OAuthAuthorizer to get the group permission info by domainData. 

Ideally domainCache should be passed in on initialization. However, some implementation may not be able to get the domainCache(especially Uber's internal implementation). So we are changing the Authorizer interface to pass it in so that it will work for all implementation. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/authorization/oauthAutorizer_test.go
+++ b/common/authorization/oauthAutorizer_test.go
@@ -26,6 +26,7 @@ import (
 	"testing"
 
 	"github.com/cristalhq/jwt/v3"
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/suite"
 	"go.uber.org/yarpc/api/encoding"
@@ -33,6 +34,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/config"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
@@ -46,6 +48,8 @@ type (
 		att             Attributes
 		token           string
 		tokenExpiredIat string
+		controller      *gomock.Controller
+		domainCache     cache.DomainCache
 		ctx             context.Context
 	}
 )
@@ -99,6 +103,9 @@ func (s *oauthSuite) SetupTest() {
 		TaskList:   nil,
 		Permission: PermissionRead,
 	}
+
+	s.controller = gomock.NewController(s.T())
+	s.domainCache = cache.NewMockDomainCache(s.controller)
 	s.ctx = ctx
 }
 
@@ -107,7 +114,7 @@ func (s *oauthSuite) TearDownTest() {
 }
 
 func (s *oauthSuite) TestCorrectPayload() {
-	authorizer := NewOAuthAuthorizer(s.cfg, s.logger)
+	authorizer := NewOAuthAuthorizer(s.cfg, s.logger, s.domainCache)
 	result, err := authorizer.Authorize(s.ctx, &s.att)
 	s.NoError(err)
 	s.Equal(result.Decision, DecisionAllow)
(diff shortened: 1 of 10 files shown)


just-at-uber:
You have some failing unit tests that need fixing

common/authorization/oauthAutorizer_test.go
```diff
@@ -108,23 +108,23 @@ func (s *oauthSuite) TearDownTest() {
 
 func (s *oauthSuite) TestCorrectPayload() {
 	authorizer := NewOAuthAuthorizer(s.cfg, s.logger)
-	result, err := authorizer.Authorize(s.ctx, &s.att)
+	result, err := authorizer.Authorize(s.ctx, &s.att, nil)
```
just-at-uber, line 111: You might want to add a test which passes the domain cache

longquanzheng: yeah I will ping you again once ready.

service/frontend/service.go
```diff
@@ -254,9 +254,7 @@ func (s *Service) Start() {
 	// Additional decorations
 	var handler Handler = s.handler
 	handler = NewDCRedirectionHandler(handler, s, s.config, s.params.DCRedirectionPolicy)
-	if s.params.Authorizer != nil {
-		handler = NewAccessControlledHandlerImpl(handler, s, s.params.Authorizer)
-	}
+	handler = NewAccessControlledHandlerImpl(handler, s, s.params.Authorizer, s.params.AuthorizationConfig)
```
noiarek, line 257: So it seems handler is being replaced straight away, we might need to remove either this line or the prev one?

service/frontend/accessControlledHandler.go
```diff
@@ -42,7 +43,10 @@ type AccessControlledWorkflowHandler struct {
 var _ Handler = (*AccessControlledWorkflowHandler)(nil)
 
 // NewAccessControlledHandlerImpl creates frontend handler with authentication support
-func NewAccessControlledHandlerImpl(wfHandler Handler, resource resource.Resource, authorizer authorization.Authorizer) *AccessControlledWorkflowHandler {
+func NewAccessControlledHandlerImpl(wfHandler Handler, resource resource.Resource, authorizer authorization.Authorizer, cfg config.Authorization) *AccessControlledWorkflowHandler {
+	if authorizer == nil {
```
noiarek, line 47: In the old code this is called by [this file](https://github.com/uber/cadence/blob/master/service/frontend/service.go#L258) that is checking already if the Authorizer is nil. This is being built [here](https://github.com/uber/cadence/blob/master/service/frontend/service.go#L182) and in both places where `NewService` is being called [here](https://github.com/uber/cadence/blob/master/host/onebox.go#L428) and [here](https://github.com/uber/cadence/blob/master/cmd/server/cadence/server.go#L232) the Authorizer is being populated. I'm wondering if we need this validation at all (and I was the one whoe removed it last time, sorry if it was needed)

common/authorization/oauthAutorizer_test.go
```diff
@@ -181,7 +188,7 @@ func (s *oauthSuite) TestIncorrectPermissionInAttributes() {
 
 func (s *oauthSuite) TestIncorrectDomainInAttributes() {
 	s.att.DomainName = "myotherdomain"
-	authorizer := NewOAuthAuthorizer(s.cfg, s.logger)
+	authorizer := NewOAuthAuthorizer(s.cfg, s.logger, s.domainCache)
```
noiarek, line 191: Is domain cache something that changes per user? If that's the case it might be better inside of Attributes (when being called using Authorization). 

I'm implementing Authentication and if I need to use this variable, probably this is the right place then.

service/frontend/accessControlledHandler.go
longquanzheng: This will also be called by Uber’s internal repo which may not have authorizer. It’s just to be compatible

common/authorization/oauthAutorizer_test.go
longquanzheng: This is a component that we are going to get some attributes by domainName. 

something like
```
./cadence --do samples-domain domain desc
Name: samples-domain
UUID: cfebaf91-d58d-46fe-9951-345c357332bf
Description:
OwnerEmail:
DomainData: map[ WRITE_GROUPS:group3@indeed.com group4@indeed.com serviceA serviceB READ_GROUPS:group1@indeed.com group2@indeed.com]
```
Get the read or write groups that have the permission, and then compare with the `groups` in the JWT token. 
It's better to do it inside this authroizer because not every implementation needs it. Getting the domain data may have some overhead(very low though, since it's a cache).

service/frontend/service.go
longquanzheng: oh you may misunderstand. It's passing the old value and then replace to a new one.

longquanzheng:
Close this one in favor of https://github.com/uber/cadence/pull/4333

PR #4333, opened 2021-07-28
Update OAuth implementation to use domainCache to authorize 

longquanzheng:
**What changed?**
Update OAuth implementation to use domainCache to authorize

**Why?**
Based on discussion with @just-at-uber , we need to make change on the proposal to support multiple domains operation on WebUI with one logging.
The best way to do it is letting OAuthAuthorizer to get the group permission info by domainData. 

Ideally domainCache should be passed in on initialization. However, some implementation may not be able to get the domainCache(especially Uber's internal implementation). So we are changing the Authorizer interface to pass it in so that it will work for all implementation. 

See proposal https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#

**How did you test it?**
new and existing tests. Working with @noiarek to fix the tests. 

**Potential risks**
No

**Release notes**
No, feature not ready yet.

**Documentation Changes**
No, feature not ready yet.


service/frontend/service.go
```diff
@@ -254,9 +254,7 @@ func (s *Service) Start() {
 	// Additional decorations
 	var handler Handler = s.handler
 	handler = NewDCRedirectionHandler(handler, s, s.config, s.params.DCRedirectionPolicy)
-	if s.params.Authorizer != nil {
```
longquanzheng, line 257: The reason we remove this check is to allow creating Authorizer within the service, so that we can have the `domainCache` entity. 
We have to allow boostraping(server.go) passing in a nil as Authorizer, and then create an Authorizer based on the config. 
Uber's internal repo doing the customized boostraping which will pass in a non-nil or nil Authorizer will not have any behavior change. By default this AccessControllerHandler is noop.

common/authorization/oauthAuthorizer.go
```diff
@@ -107,19 +125,36 @@ func (a *oauthAuthority) parseToken(tokenStr string, verifier jwt.Verifier) (*jw
...
-		return fmt.Errorf("token doesn't have the right permission")
+	return nil
+}
+
+func (a *oauthAuthority) validatePermission(claims *jwtClaims, attributes *Attributes, data map[string]string) error {
+	groups := ""
+	switch attributes.Permission {
+	case PermissionRead:
+		groups = data[common.DomainDataKeyForReadGroups]
+	case PermissionWrite:
```
noiarek, line 143: Is write still write + read? We might want to change back the type of permissionwrite to string

common/authorization/oauthAuthorizer.go
```diff
@@ -107,19 +125,36 @@ func (a *oauthAuthority) parseToken(tokenStr string, verifier jwt.Verifier) (*jw
...
+		groups = data[common.DomainDataKeyForWriteGroups]
+	default:
+		return fmt.Errorf("code bug, this shouldn't happen")
```
noiarek, line 146: I don't know if this is necessarily true for `PermissionAdmin`

If the claims (the token) has `IsAdmin = false` but the Attributes requires admin, this is not a code bug, is an insufficient permission in the token (I think)

common/authorization/oauthAuthorizer.go
longquanzheng: yeah you are right. Fixed it.

longquanzheng: Yes, you are right. Fixed it.

just-at-uber:
Looks good to me. I'm reaching out to another Cadence team member to review too. Thanks!

common/authorization/oauthAuthorizer.go
```diff
@@ -107,19 +127,40 @@ func (a *oauthAuthority) parseToken(tokenStr string, verifier jwt.Verifier) (*jw
...
+	return nil
+}
+
+func (a *oauthAuthority) validatePermission(claims *jwtClaims, attributes *Attributes, data map[string]string) error {
+	groups := ""
+	switch attributes.Permission {
+	case PermissionRead:
+		groups = data[common.DomainDataKeyForReadGroups] + groupSeparator + data[common.DomainDataKeyForWriteGroups]
+	case PermissionWrite:
+		groups = data[common.DomainDataKeyForWriteGroups]
```
noiarek, line 146: I think is the other way around:

If read, read grous
if write, read + write

common/authorization/oauthAuthorizer.go
```diff
@@ -107,19 +127,40 @@ func (a *oauthAuthority) parseToken(tokenStr string, verifier jwt.Verifier) (*jw
...
+	default:
+		if claims.Admin {
```
noiarek, line 148: This is test in line 91 so maybe we should only leave 

`return fmt.Errorf("token doesn't have permission for %s API", attributes.Permission)` (in case we add more in the future)

longquanzheng: yeah I am fine with that.

common/authorization/oauthAuthorizer.go
longquanzheng: This read is attribute meaning that API require read permission, so either a read or write permission on this domain should work. So we need to read all the groups on this domain.

If this is write, meaning that API is requiring write permission, then we have to only looking at the groups that have write permission. 

Can you double check? (It's sorta easy to think it wrong).

yycptt (approved):

yycptt (approved):
The server start up logic change looks good to me

PR #4336, opened 2021-07-30
Add X-Cluster Child Workflow Completion Tasks

demirkayaender:
**What changed?**
Defining and persisting cross cluster tasks for RecordChildWorkflowExeuctionComplete task. 

**Why?**
With this diff, child workflow notify the remote workflow about its completion.

**How did you test it?**
transfer_active_task_executor_test

**Potential risks**

**Release notes**

**Documentation Changes**


common/types/shared.go
```diff
@@ -10652,12 +10652,58 @@ func (v *CrossClusterSignalExecutionRequestAttributes) GetControl() (o []byte) {
 type CrossClusterSignalExecutionResponseAttributes struct {
 }
 
+type CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes struct {
+	TargetDomainID     string             `json:"targetDomainID,omitempty"`
+	TargetWorkflowID   string             `json:"targetWorkflowID,omitempty"`
+	TargetRunID        string             `json:"targetRunID,omitempty"`
+	InitiatedEventID   int64              `json:"initiatedEventID,omitempty"`
+	CompletedExecution *WorkflowExecution `json:"workflowExecution,omitempty"`
+	CompletionEvent    *HistoryEvent      `json:"scheduledEvent,omitempty"`
```
demirkayaender, line 10661: will delete these.

service/history/execution/mutable_state_task_generator.go
```diff
@@ -83,6 +83,7 @@ type (
 			targetCluster string,
 		) error
 
+		// create corresponding tasks in persistence applyTasks
```
demirkayaender, line 86: will delete this

service/history/task/transfer_active_task_executor_test.go
```diff
@@ -706,6 +707,89 @@ func (s *transferActiveTaskExecutorSuite) TestProcessCloseExecution_HasParent()
...
+		WorkflowExecution:  &parentExecution,
+		InitiatedID:        parentInitiatedID,
+		CompletedExecution: &workflowExecution,
+		CompletionEvent:    event,
+	}).Return(nil).AnyTimes()
+	s.mockVisibilityMgr.On("RecordWorkflowExecutionClosed", mock.Anything, mock.Anything).Return(nil).Once()
+	s.mockArchivalMetadata.On("GetVisibilityConfig").Return(archiver.NewDisabledArchvialConfig())
+	s.mockExecutionMgr.On("UpdateWorkflowExecution", mock.Anything, mock.Anything).Return(
+		&persistence.UpdateWorkflowExecutionResponse{MutableStateUpdateSessionStats: nil}, nil)
+	s.mockClusterMetadata.EXPECT().ClusterNameForFailoverVersion(int64(-24)).Return(s.mockClusterMetadata.GetCurrentClusterName()).AnyTimes()
```
demirkayaender, line 787: Not sure where -24 is coming from, any idea? This is s.version until we close the execution

service/history/task/transfer_active_task_executor_test.go
```diff
@@ -628,6 +628,45 @@ func (s *transferActiveTaskExecutorSuite) TestProcessDecisionTask_Duplication()
...
+func (s *transferActiveTaskExecutorSuite) TestProcessCloseExecution_HasParentCrossCluster() {
+	s.testProcessCloseExecution_HasParent(
+		s.remoteTargetDomainEntry,
+		func(
+			mutableState execution.MutableState,
+			workflowExecution, targetExecution types.WorkflowExecution,
+		) {
+			s.mockVisibilityMgr.On("RecordWorkflowExecutionClosed", mock.Anything, mock.Anything).Return(nil).Once()
+			s.mockArchivalMetadata.On("GetVisibilityConfig").Return(archiver.NewDisabledArchvialConfig())
+			s.mockClusterMetadata.EXPECT().ClusterNameForFailoverVersion(int64(-24)).Return(s.mockClusterMetadata.GetCurrentClusterName()).AnyTimes()
```
demirkayaender, line 652: We are consistently returning -24 here after the child execution is closed. Not sure where it's coming from

common/types/mapper/proto/shared.go
```diff
@@ -1360,6 +1360,47 @@ func ToCrossClusterSignalExecutionResponseAttributes(t *sharedv1.CrossClusterSig
...
 
+// FromCrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes converts internal CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes type to proto
+func FromCrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes(t *types.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes) *sharedv1.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
+	if t == nil {
+		return nil
+	}
+	return &sharedv1.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes{
+		TargetDomainId:          t.TargetDomainID,
+		TargetWorkflowExecution: FromWorkflowRunPair(t.TargetWorkflowID, t.TargetRunID),
+		InitiatedEventId:        t.InitiatedEventID,
```
yycptt, line 1371: Missing `completionEvent` here?

common/types/mapper/proto/shared.go
```diff
@@ -1360,6 +1360,47 @@ func ToCrossClusterSignalExecutionResponseAttributes(t *sharedv1.CrossClusterSig
...
+// ToCrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes converts proto CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes type to internal
+func ToCrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes(t *sharedv1.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes) *types.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
+	if t == nil {
+		return nil
+	}
+	return &types.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes{
+		TargetDomainID:   t.TargetDomainId,
+		TargetWorkflowID: ToWorkflowID(t.TargetWorkflowExecution),
+		TargetRunID:      ToRunID(t.TargetWorkflowExecution),
+		InitiatedEventID: t.InitiatedEventId,
```
yycptt, line 1384: Same here?

common/types/shared.go
```diff
@@ -10652,12 +10652,56 @@ func (v *CrossClusterSignalExecutionRequestAttributes) GetControl() (o []byte) {
 type CrossClusterSignalExecutionResponseAttributes struct {
 }
 
+type CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes struct {
+	TargetDomainID   string `json:"targetDomainID,omitempty"`
+	TargetWorkflowID string `json:"targetWorkflowID,omitempty"`
+	TargetRunID      string `json:"targetRunID,omitempty"`
+	InitiatedEventID int64  `json:"initiatedEventID,omitempty"`
```
yycptt, line 10659: `CompletionEvent`?

proto/internal/uber/cadence/shared/v1/queue.proto
```diff
@@ -109,12 +109,22 @@ message CrossClusterSignalExecutionRequestAttributes {
 message CrossClusterSignalExecutionResponseAttributes {
 }
 
+message CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
+  string target_domain_id = 1;
+  api.v1.WorkflowExecution target_workflow_execution = 2;
+  int64 initiated_event_id = 3;
```
yycptt, line 115: add `completion_event`?

service/history/task/transfer_active_task_executor_test.go
yycptt: -24 is the failover version we used for local domains. https://github.com/uber/cadence/blob/master/common/constants.go#L35

service/history/task/transfer_active_task_executor.go
```diff
@@ -337,21 +338,30 @@ func (t *transferActiveTaskExecutor) processCloseExecution(
...
-			},
-			CompletionEvent: completionEvent,
-		})
+		targetDomainEntry, err := t.shard.GetDomainCache().GetDomainByID(parentDomainID)
+		if err != nil {
+			return err
+		}
+		if targetCluster, isCrossCluster := t.isCrossClusterTask(task.DomainID, targetDomainEntry); isCrossCluster {
+			// TODO: consider moving this logic to GenerateWorkflowCloseTasks and uxxse here as a back-up to save latency
+			err = t.generateCrossClusterTask(ctx, wfContext, task, targetCluster)
```
yycptt, line 347: I think we probably need to do some refactoring here. 
Currently, the impl of `generateCrossClusterTask` will close the transaction for the mutable state and perform a DB update and is intended to be called as the last step when processing a task. It works for the other three types as for them, we just need to generate one new cross-cluster task, and then the task processing is done. 

For close execution, however, it's possible that another task needs to be generated for applying parent close policy and we can't call `generateCrossClusterTask` twice to close the transaction twice and update DB twice (maybe we don't have such check, but I don't think we should do it). Also we still need the mutable state for applying parent close policy.

I guess the idea here is that we only generate the new task but do not update the workflow. Then at the very end of `processCloseExecution`, we will have a complete list of tasks need to be generated, call the update workflow execution method to close the mutable state transaction, and do a DB update.

Similarly, `taskGenerator.GenerateCrossClusterTaskFromTransferTask` probably also need to be refactored as it's possible that we need to generate two cross-cluster tasks from one transfer task (with probably different target cluster), while the current impl assumes a 1:1 relationship. Sorry I didn't plan for that when writing that method.

service/history/task/transfer_active_task_executor.go
```diff
@@ -1219,7 +1229,12 @@ func (t *transferActiveTaskExecutor) generateCrossClusterTask(
 		wfContext,
 		false,
 		func(ctx context.Context, mutableState execution.MutableState) error {
-			if !mutableState.IsWorkflowExecutionRunning() {
+			if task.TaskType == persistence.TransferTaskTypeCloseExecution {
+				if !mutableState.IsWorkflowCompleted() {
```
yycptt, line 1233: Do we need this new method? Is `!mutableState.IsWorkflowCompleted` the same as `mutableState.IsWorkflowExecutionRunning`?

demirkayaender: Not exactly same, do we want to add the new x-cluster task for the following states? 

```
persistence.WorkflowStateZombie:
persistence.WorkflowStateCorrupted:
```

service/history/task/transfer_active_task_executor.go
```diff
@@ -364,6 +379,17 @@ func (t *transferActiveTaskExecutor) processCloseExecution(
 		return err
 	}
 
+	// TODO: pass crossClusterTaskGenerators to processParentClosePolicy to add new cross
+	// cluster tasks then move the call below after processParentClosePolicy call
+	if len(crossClusterTaskGenerators) > 0 {
+		t.generateCrossClusterTasks(
```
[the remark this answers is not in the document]
demirkayaender: Added the error handling. 

I was planning move this down after the parent close policy changes. Could be better to keep the changes smaller and for one API. In the PR after this one it will be after the `processParentClosePolicy`.

service/history/task/transfer_active_task_executor.go
demirkayaender: If it's not expected, why should we include those cases here? It's probably better to only have the valid case. Having other states can also confuse other people reading here. We can talk today to see if you have other concerns.

service/history/task/transfer_active_task_executor.go
```diff
@@ -1208,18 +1234,22 @@ func (t *transferActiveTaskExecutor) isCrossClusterTask(
...
 ) error {
 	return t.updateWorkflowExecution(
 		ctx,
 		wfContext,
 		false,
 		func(ctx context.Context, mutableState execution.MutableState) error {
-			if !mutableState.IsWorkflowExecutionRunning() {
+			if task.TaskType == persistence.TransferTaskTypeCloseExecution {
+				if !mutableState.IsWorkflowCompleted() {
+					return &types.BadRequestError{Message: "Workflow execution is still running, should have been completed."}
```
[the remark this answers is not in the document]
demirkayaender: Good catch!

yycptt (approved):

PR #4339, opened 2021-08-04
Parallelize GetWorkflowExecution SQL calls

demirkayaender:
**What changed?**
During DocStore migration we realized that GetWorkflowExecution API calls are the slowest (3x-4x slower at best). With this PR I'm parallelizing the SQL calls so we can save some latency for this API

This API makes GetWorkflowExecution 2.6 times faster (decreased sql latency by 62%) for SQL plugins based on bench test runs. Even though, it's still slightly slower compared to nosql, the difference is ignorable: <5ms.

**Why?**
To improve SQL performance

**How did you test it?**
Current unit tests should pass. This PR doesn't change any functionality

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/sql/sqlExecutionStore.go
+++ b/common/persistence/sql/sqlExecutionStore.go
@@ -342,8 +342,13 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
 
 	err := g.Wait()
 	if err != nil {
-		return nil, &types.InternalServiceError{
-			Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
+		switch err := err.(type) {
+		case *types.EntityNotExistsError:
+			return nil, err
+		default:
+			return nil, &types.InternalServiceError{
+				Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
+			}
 		}
 	}
 
--- a/common/persistence/sql/sqlExecutionStore.go
+++ b/common/persistence/sql/sqlExecutionStore.go
@@ -216,14 +216,13 @@ func (m *sqlExecutionStore) createWorkflowExecutionTx(
 	return &p.CreateWorkflowExecutionResponse{}, nil
 }
 
-func (m *sqlExecutionStore) GetWorkflowExecution(
+func (m *sqlExecutionStore) getExecutions(
 	ctx context.Context,
 	request *p.InternalGetWorkflowExecutionRequest,
-) (*p.InternalGetWorkflowExecutionResponse, error) {
-
-	domainID := serialization.MustParseUUID(request.DomainID)
-	runID := serialization.MustParseUUID(request.Execution.RunID)
-	wfID := request.Execution.WorkflowID
+	domainID serialization.UUID,
+	wfID string,
+	runID serialization.UUID,
+) ([]sqlplugin.ExecutionsRow, error) {
 	executions, err := m.db.SelectFromExecutions(ctx, &sqlplugin.ExecutionsFilter{
 		ShardID: m.shardID, DomainID: domainID, WorkflowID: wfID, RunID: runID})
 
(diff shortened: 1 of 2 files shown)


Groxx:
I'll pretty much always recommend using an errgroup.WithContext instead of rolling these by hand: https://pkg.go.dev/golang.org/x/sync/errgroup

Then error-returning (first only, others ignored) and canceling is handled for you, and much simpler.  and as long as you `Wait()`, you can just use local vars rather than chans to read the results.

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -255,6 +254,103 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
+	wfID := request.Execution.WorkflowID
+
+	// get executions
+	executionsChan := make(chan []sqlplugin.ExecutionsRow)
+	executionsErrChan := make(chan error)
+	go func() {
+		executions, err := m.getExecutions(ctx, request, domainID, wfID, runID)
+		executionsChan <- executions
+		executionsErrChan <- err
+	}()
```
Groxx, line 276: with an errgroup this would probably be:
```go
var executions []sqlplugin.ExecutionsRow
g.Go(func() error {
  var err error
  executions, err = m.getEsxecutions(ctx, request, domainID, wfID, runID)
  return err
})
...
err := g.Wait()
if err != nil {
  return nil, err
}
// executions is populated and not racy, use it
```

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -255,116 +256,107 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
+		var err error
+		signalsRequested, err = getSignalsRequested(
+			ctx, m.db, m.shardID, domainID, wfID, runID)
+		return err
+	})
+
+	err := g.Wait()
+	if executionsError != nil {
+		return nil, executionsError
+	}
```
Groxx, line 340: I'm guessing this err is more likely to be relevant, hence it's preferred?

If so: I don't think that'll actually work.  Once anything in the errgroup fails, the context will be canceled, so ^ the executions error is very likely to be either A) the same as the errgroup err, because it failed first, or B) a context-cancelled err, because it failed after some other request.  Context-cancelled is almost always going to be worse than whatever the errgroup returns, since there's no information about the cause of the cancellation.

If we don't mind _always waiting_ for executions, it could just use the non-errgroup context.  `g.Wait()` will still wait for it to complete (it waits for all funcs to return), but that way it should have a relevant error, at the cost of potentially-increased run-time.

go.mod
```diff
@@ -69,14 +69,15 @@ require (
 	go.uber.org/zap v1.13.0
 	golang.org/x/net v0.0.0-20201031054903-ff519b6c9102
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
+	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
```
Groxx, line 72: wow, really?  that's the first time we've used sync?  quite surprising

go.mod
```diff
@@ -69,14 +69,15 @@ require (
...
 	golang.org/x/text v0.3.4 // indirect
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
 	golang.org/x/tools v0.1.0
 	gonum.org/v1/gonum v0.7.0
 	google.golang.org/api v0.26.0
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20201201144952-b05cb90ed32e // indirect
-	google.golang.org/grpc v1.29.1
+	google.golang.org/grpc v1.29.1 // indirect
```
Groxx, line 80: not harmful if this is incorrect for some reason... but odd.  do we really not import grpc anywhere?  I wonder when that stopped happening, since this PR can't be the cause.

Groxx (commented):
with a general caution for "panics will kill the process" still: looks pretty good, though check that executionsErr note.

for panics: do we have any logs implying these might be happening?  if not, it may be fine, they really should be roughly-nonexistent.

common/persistence/sql/sqlExecutionStore.go
demirkayaender: Yeah this is a bit racy but dangerous IMO. We are paying for missing some of these errors (but get new ones) to get better performance. Given we will still get "some" error and this function is supposed to succeed 99+% of the time. I'm not really worried about it. I believe this is a good trade-off. Let me know if you'd like discuss about this.

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -252,119 +254,119 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
-			m.shardID,
-			domainID,
-			wfID,
-			runID)
-		if err != nil {
-			return nil, err
+	state, err := m.populateWorkflowMutableState(executions[0])
+	if err != nil {
+		return nil, &types.InternalServiceError{
+			Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
```
Groxx, line 360: it's probably worth making these two error messages different, so we can tell them apart.  though the causes are quite possibly different between the two.

common/persistence/sql/sqlExecutionStore.go
Groxx: I don't mean to imply it's racy (it isn't, in the data-safety sense at least).  I mean that this should _almost always_ be the same as **or worse** than not tracking `executionsError` separately at all.

When something else fails, `executionsError` will _almost always_ be a context-canceled error because much of its time is spent doing RPC calls... which would be returned here, instead of the error from the thing-that-failed-first.  So this will _mostly_ suppress other useful errors, and return useless ones.

...with a very small window for races where two goroutines both fail after their associated requests are completed, but before they return (on the order of microseconds).  We might be able to capture those with a "... and executions err is not a canceled-context error", but I'd only expect it to hit that branch _much_ less than 1% of the time.

Groxx (commented):
I think there's a change coming for the executionsError, but otherwise LGTM.  Should be a quick 👍

common/persistence/sql/sqlExecutionStore.go
demirkayaender: just merged them, I think the impact of separating them should be pretty low

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -252,119 +254,119 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
-	if err != nil {
-		return nil, &types.InternalServiceError{
-			Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
+func (m *sqlExecutionStore) GetWorkflowExecution(
+	ctx context.Context,
+	request *p.InternalGetWorkflowExecutionRequest,
+) (resp *p.InternalGetWorkflowExecutionResponse, e error) {
+	recoverPanic := func(err *error) {
+		if r := recover(); r != nil {
+			*err = fmt.Errorf("DB operation panicked: %v %s", r, debug.Stack())
```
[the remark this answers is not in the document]
demirkayaender: I have a task about capturing the correct callstacks. Right now all the errors have the metricsClient callstacks unfortunately.

Groxx (approved):

PR #4351, opened 2021-08-12
Introduce throttling retry

Shaddoll:
**What changed?**
Introduce throttling retry using a different retry policy when throttling happens

**Why?**
When throttling happens, retry should also be throttled 

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**


common/backoff/retry.go
```diff
@@ -125,6 +125,45 @@ func Retry(operation Operation, policy RetryPolicy, isRetryable IsRetryable) err
 	}
 }
 
+// ThrottleRetry uses a different retry policy when throttling happens
+func ThrottleRetry(operation Operation, policy RetryPolicy, isRetryable IsRetryable, throttlePolicy RetryPolicy, isThrottle IsRetryable) error {
```
yux0, line 129: Can we add some unit tests for this new retrier?

longquanzheng:
It would be nice to add more words to the title so that the release notes will be easier to read. E.g retry for what

Shaddoll:
> Do you plan to have a separate PR for adding ThrottlePolicy?

This PR has a default ThrottlePolicy, which is enough for our use case.

PR #4357, opened 2021-08-13
Added config store functionality (initial implementation)

aliue:
**What changed?**
Adds an alternative dynamic config system other than file-based client. Cadence service takes over responsibility of storing and propagating changes to dynamic configs. Added admin cli commands for operators to manage them. Currently only supports Cassandra as data store.

**Why?**
Offers a potentially better solution to dynamic config as file-based system can be difficult to use in production environment as all hosts need a local copy. Currently, it is the operator's responsibility to manage syncing this file.

**How did you test it?**
Tested locally, added unit tests. Made sure it passes canary with both Cassandra and SQL.

**Potential risks**
Currently, configStore is not enabled by default, so it shouldn't causes any issues that we did not check for already.

**Release notes**
Updated cassandra schema and cli. Should be backwards compatible with the previous dynamicConfigClient struct in development.yaml if new dynamicconfig struct is not specified.

**Documentation Changes**


yycptt (approved):

yycptt:
Great job! Thanks for the contribution!

PR #4364, opened 2021-08-14
Adding middleware to inject Auth token for internal requests to frontend

noiarek:
**What changed?**

**Why?**
For https://docs.google.com/document/d/1a-Xo9FSCLubYlIQHB_eX5j4Yv2GDv_82J94At2dAL3o/edit#

**How did you test it?**
Local test:
```
docker-compose -f docker/dev/cassandra-esv7-kafka.yml up
```
and start server with
```
./cadence-server --zone es_v7 start
```

then register a domain and start two workflows
```
$cadence --do sample d re
Domain sample successfully registered.
```
then use batch to terminate the two workflows
```
qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf start --wt wftype --et 100 --tl tasklist
Started Workflow Id: d202b1ef-69ec-4025-ac05-3b24ff58e7b9, run Id: 3d2a7f80-e046-4ba7-ad8b-4ae86c6033bf
qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf start --wt wftype --et 100
qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf start --wt wftype --et 100000 --tl abc
Started Workflow Id: 8873df2a-4bb7-4ff5-aa63-be089ddafbca, run Id: da1757bb-19bb-4e8e-86a0-3f415ca6a111

qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf list --op
  WORKFLOW TYPE |             WORKFLOW ID              |                RUN ID                | TASK LIST | START TIME | EXECUTION TIME
  wftype        | 8873df2a-4bb7-4ff5-aa63-be089ddafbca | da1757bb-19bb-4e8e-86a0-3f415ca6a111 | abc       | 14:50:47   | 14:50:47
  wftype        | d202b1ef-69ec-4025-ac05-3b24ff58e7b9 | 3d2a7f80-e046-4ba7-ad8b-4ae86c6033bf | tasklist  | 14:50:30   | 14:50:30

qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf batch start -q "WorkflowType='wftype'" --re test --bt terminate
This batch job will be operating on 2 workflows.
Please confirm[Yes/No]:yes
{
  "jobID": "da1b9b01-790f-4051-8ccc-8aa360aed5ee",
  "msg": "batch job is started"
}

```
and confirm it succeeded
```
qlong@~/cadence:
(adding-middleware-for-frontend) $cadence --do sample wf batch desc --jid da1b9b01-790f-4051-8ccc-8aa360aed5ee
{
  "msg": "batch job is finished successfully"
}
```

**Potential risks**

**Release notes**

**Documentation Changes**


common/config/config.go
```diff
@@ -100,8 +100,12 @@ type (
 		Algorithm string `yaml:"algorithm"`
 		// Public Key Path for verifying JWT token passed in from external clients
 		PublicKey string `yaml:"publicKey"`
+		// Public Key loaded once
+		PublicKeyLoaded []byte `yaml:"-"`
 		// Private Key Path for creating JWT token
 		PrivateKey string `yaml:"privateKey"`
+		// Private Key loaded once
+		PrivateKeyLoaded []byte `yaml:"-"`
```
longquanzheng, line 108: Let's not adding these two entries here. The loaded logic can be done outside to leave the config clean.

common/config/config.go
```diff
@@ -321,6 +325,8 @@ type (
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service address(Host:Port). Host can be DNS name.
 		RPCAddress string `yaml:"rpcAddress"`
+		// Private Key Path
+		PrivateKey string `yaml:"privateKey"`
```
longquanzheng, line 329: Let's add a new `AuthorizationProvider` struct here
AuthorizationProvider struct{
  Enable bool
  Type string  // only support OAuthAuthorization
  PrivateKey string
}

Also we can remove the private key from JwtCredentials and only use this one: https://github.com/uber/cadence/blob/a58b8b9670417ccbf2cd021882e750f95d6fc508/common/config/config.go#L104

cmd/server/cadence/server.go
```diff
@@ -220,8 +221,13 @@ func (s *server) startService() common.Daemon {
 		}
 	}
 
-	dispatcher, err := params.DispatcherProvider.Get(common.FrontendServiceName, s.cfg.PublicClient.HostPort)
+	// will return empty array if not enabled
+	privateKey, err := s.cfg.Authorization.OAuthAuthorizer.GetPrivateKey()
```
longquanzheng, line 225: like the comment, use clusterMetadata.clusterInfomation.<currentCluster>.AuthorizationProvider 

You may need to create a new wrapper to create `AuthProvider` based on the config as it will be used in `clientBean.go`

common/config/config.go
```diff
@@ -321,6 +321,15 @@ type (
...
 		// Address indicate the remote service address(Host:Port). Host can be DNS name.
 		RPCAddress string `yaml:"rpcAddress"`
+		// AuthorizationProvider contains the information to authorize the cluster
+		AuthorizationProvider struct {
+			// Enable indicates if the auth provider is enabled
+			Enable bool `yaml:"enable"`
+			// Type auth provider type
+			Type string `yaml:"type"` // only supports OAuthAuthorization
+			// PrivateKey is the private key path
+			PrivateKey string `yaml:"privateKey"`
```
noiarek, line 331: If this config is per cluster, does that mean they would have to write this per cluster? Is this the most efficient way @longquanzheng ? (Not sure how bigs clusters can be)

client/clientBean.go
```diff
@@ -256,7 +273,7 @@ func NewDNSYarpcDispatcherProvider(logger log.Logger, interval time.Duration) Di
 	}
 }
 
-func (p *dnsDispatcherProvider) Get(serviceName string, address string) (*yarpc.Dispatcher, error) {
+func (p *dnsDispatcherProvider) Get(serviceName string, address string, options *DispatcherOptions) (*yarpc.Dispatcher, error) {
```
noiarek, line 276: Do I need to run something to get the latest mock for this?  Quoting you @longquanzheng 

```There are two implementation, one is mock. The real one is dnsYarpcDispatcherProvider.```

File: client/clientBean.go [hunk 6]
```diff
@@ -116,8 +123,18 @@ func NewClientBean(factory Factory, dispatcherProvider DispatcherProvider, clust
 		if !info.Enabled {
 			continue
 		}
+		var authProvider clientworker.AuthorizationProvider
+		if info.AuthorizationProvider.Enable {
+			privateKey, err := ioutil.ReadFile(info.AuthorizationProvider.PrivateKey)
```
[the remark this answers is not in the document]
longquanzheng: It's only created during startup.

client/clientBean.go
```diff
@@ -285,6 +302,9 @@ func (p *dnsDispatcherProvider) Get(serviceName string, address string) (*yarpc.
 				ServiceName: serviceName,
 			},
 		},
+		OutboundMiddleware: yarpc.OutboundMiddleware{
+			Unary: &outboundMiddleware{authProvider: options.AuthProvider},
```
longquanzheng, line 306: Need to add a check if options is nil

cmd/server/cadence/server.go
```diff
@@ -220,7 +222,16 @@ func (s *server) startService() common.Daemon {
...
+	// will return empty array if not enabled
+	var authProvider clientworker.AuthorizationProvider
+	if s.cfg.Authorization.OAuthAuthorizer.Enable {
+		privateKey, err := ioutil.ReadFile(s.cfg.Authorization.OAuthAuthorizer.JwtCredentials.PrivateKey)
+		if err != nil {
+			log.Fatalf("invalid private key path %s", s.cfg.Authorization.OAuthAuthorizer.JwtCredentials.PrivateKey)
+		}
+		authProvider = clientworker.NewJwtAuthorizationProvider(privateKey)
+	}
+	dispatcher, err := params.DispatcherProvider.Get(common.FrontendServiceName, s.cfg.PublicClient.HostPort, &client.DispatcherOptions{AuthProvider: &authProvider})
```
longquanzheng, line 234: This code is duplicated. Can we move it to a helper func in common/authorization pkg?

common/config/config.go
longquanzheng: I never see we do in this way. 
Can you move the struct outside to be independent? 

So that:
```golang
	ClusterInformation struct {
		Enabled                bool  `yaml:"enabled"`
		RPCName string `yaml:"rpcName"`
		// Address indicate the remote service address(Host:Port). Host can be DNS name.
		RPCAddress string `yaml:"rpcAddress"`

		AuthorizationProvider AuthorizationProvider `yaml:"authorizationProvider"`
...
	}
		AuthorizationProvider struct {
			// Enable indicates if the auth provider is enabled
			Enable bool `yaml:"enable"`
			// Type auth provider type
			Type string `yaml:"type"` // only supports OAuthAuthorization
			// PrivateKey is the private key path
			PrivateKey string `yaml:"privateKey"`
		}
```

longquanzheng:
Also, don't forget to remove the private key from Authorizer config. 
https://github.com/uber/cadence/blob/a58b8b9670417ccbf2cd021882e750f95d6fc508/common/config/config.go#L104

I thought we will use it in this PR but we changed our mind to make it separate(which is much better)

host/xdc/elasticsearch_test.go
```diff
@@ -18,8 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// +build !race
-// +build esintegration
+//go:build !race && esintegration
```
noiarek, line 21: These file was modified after I ran `make bins`, not sure if it will break something. Same with the next modification

longquanzheng: Did you try “make fly”?

noiarek: I specifically made this new changes in a new commit just in case I need to revert :) I ran `make fmt` and I got `make: Nothing to be done for 'fmt'`. I reverted the last commit (removing all this changes) and tried `make fmt` and got the same message. Should I revert them manually?

common/config/config.go
```diff
@@ -323,8 +323,18 @@ type (
 		// RPCName indicate the remote service name
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service address(Host:Port). Host can be DNS name.
-		// For currentCluster, it's usually the same as publicClient.hostPort
```
longquanzheng, line 326: Why this change? Do you run into some conflicts with master?

noiarek: I did actually, deleted it by mistake trying to fix it. I'll roll it back

.gen/proto/api/v1/common.pb.go
```diff
@@ -1732,10 +1732,7 @@ func (m *WorkflowExecution) Unmarshal(dAtA []byte) error {
 			if err != nil {
 				return err
 			}
-			if skippy < 0 {
-				return ErrInvalidLengthCommon
-			}
-			if (iNdEx + skippy) < 0 {
+			if (skippy < 0) || (iNdEx+skippy) < 0 {
```
longquanzheng, line 1735: It seems like this is due to some version upgrade in the proto tool, so I believe that's fine. 
Cc @Groxx

host/xdc/elasticsearch_test.go
Groxx: Seems like a go 1.17 thing: https://golang.org/doc/go1.17#build-lines

Since we're not officially on 1.17... tbh I'm not sure if we should do this or not.  And I'm fairly surprised that `go fmt` does this since our stated version is 1.13, and older versions of go will not maintain them - if we ever change these build conditions and are *not* running <1.17, go fmt will ignore the new format, and 1.17+ will use the old conditions because it's now preferred.

longquanzheng: Are you referring to https://github.com/uber/cadence/blob/38881a8f996ce3230a77b7acc946438fbc532895/Dockerfile#L7 

I think we should change it to 1.16. 
@noiarek can you try using 1.16? If you use homebrew, there is a command to switch the go version. 

Thanks @Groxx !

Groxx: I meant https://github.com/uber/cadence/blob/master/go.mod#L3 (actually go 1.12 apparently), but that works too / I imagine we build the tests from that dockerfile?  If so, it wouldn't detect out-of-sync format directives without upgrading, yeah.

.gen/proto/api/v1/common.pb.go
Groxx: This shouldn't be affected unless the makefile's versioned proto tool was changed, which this diff doesn't do.

The two branches that are merged are identical though, possibly a newer `go fmt -s` did this?

host/xdc/elasticsearch_test.go
noiarek: @Groxx  I've installed go 1.16.7 and ran `make fmt` after it and I'm still getting this modifications. Any advice?

longquanzheng (approved):

PR #4365, opened 2021-08-15
Fill currentCluster RpcAddress with publicClient as default

longquanzheng:
**What changed?**
1. Fill currentCluster RpcAddress with publicClient as default
2. Improve config comments 
3. Remove redundant config in development_oauth.yaml

**Why?**
Improvement for https://github.com/uber/cadence/issues/3984
It's been a pain for OSS users to configure this value for a long long time. 
Also the Helmchart doesn't have an easy way to make it correct. (see https://github.com/banzaicloud/banzai-charts/pull/1278#discussion_r681587829 )

**How did you test it?**
New unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -55,6 +55,7 @@ type (
 		// Archival is the config for archival
 		Archival Archival `yaml:"archival"`
 		// PublicClient is config for sys worker service connecting to cadence frontend
+		// Default to currentCluster's RPCAddress in ClusterInformation
 		PublicClient PublicClient `yaml:"publicClient"`
 		// DynamicConfigClient is the config for setting up the file based dynamic config client
 		// Filepath would be relative to the root directory when the path wasn't absolute.
@@ -324,7 +325,6 @@ type (
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service address(Host:Port). Host can be DNS name.
 		// For currentCluster, it's usually the same as publicClient.hostPort
-		// Default to publicClient.hostPort if empty
 		RPCAddress string `yaml:"rpcAddress" validate:"nonzero"`
 	}
 
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -324,7 +324,7 @@ type (
 		RPCName string `yaml:"rpcName"`
 		// Address indicate the remote service address(Host:Port). Host can be DNS name.
 		// For currentCluster, it's usually the same as publicClient.hostPort
-		// Default to publicClient.hostPort if empty 
+		// Default to publicClient.hostPort if empty
 		RPCAddress string `yaml:"rpcAddress"`
 	}
 
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -54,7 +54,7 @@ type (
 		Kafka KafkaConfig `yaml:"kafka"`
 		// Archival is the config for archival
 		Archival Archival `yaml:"archival"`
-		// PublicClient is config for connecting to cadence frontend
+		// PublicClient is config for sys worker service connecting to cadence frontend
 		PublicClient PublicClient `yaml:"publicClient"`
 		// DynamicConfigClient is the config for setting up the file based dynamic config client
 		// Filepath would be relative to the root directory when the path wasn't absolute.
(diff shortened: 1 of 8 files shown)


common/config/config.go
```diff
@@ -502,15 +510,21 @@ func (c *Config) fillDefaults() error {
...
-				cluster.RPCName = "cadence-frontend"
-				c.ClusterMetadata.ClusterInformation[k] = cluster
+
+	for name, cluster := range c.ClusterMetadata.ClusterInformation {
+		if cluster.RPCName == "" {
+			// filling RPCName with a default value if empty
+			cluster.RPCName = "cadence-frontend"
+		}
+		if name == c.ClusterMetadata.CurrentClusterName {
+			if cluster.RPCAddress == "" {
```
yycptt, line 520: nit: merge into one `if`?

config/development.yaml
```diff
@@ -80,8 +80,6 @@ clusterMetadata:
     active:
       enabled: true
       initialFailoverVersion: 0
-      rpcName: "cadence-frontend"
-      rpcAddress: "localhost:7933"
```
yycptt, line 84: shall we leave them here? People may not know we have the logic for filling default values for those fields and get confused.

common/config/config.go
```diff
@@ -502,15 +510,21 @@ func (c *Config) fillDefaults() error {
...
+				// filling current cluster's RPC address from publicClient if empty
+				cluster.RPCAddress = c.PublicClient.HostPort
```
yycptt, line 522: In the linked issue I saw you mentioned that 
> the one in clusterInformation is more generic
> we should deprecate publicClient
so shall we reverse the operation, i.e. automatically fill in `publicClient.HostPort` if not specified?

longquanzheng: yeah, that's right. Will change to that. 
(side story, there was an issue in HelmChart that the HelmChart maintainer told me it's too difficult to provide a right value in RpcAddress. But they finally figured out this morning Lol. so I think we should be able to fill in publicClient like you said

config/development.yaml
longquanzheng: I feel like anything we have filled by default, should leave empty. 
This is already a convention in our config. 
Leaving them here will just make more work for users , and potentially, they will make it wrong. 
For example, a lot of users putting "localhost" here in the cluster setup, which would never work. 

So I still think we should remove them. But by addressing your another comment, I will remove the one in publicClient instead.

common/config/config.go
```diff
@@ -502,15 +507,21 @@ func (c *Config) fillDefaults() error {
...
-				cluster.RPCName = "cadence-frontend"
-				c.ClusterMetadata.ClusterInformation[k] = cluster
-			}
+
+	for name, cluster := range c.ClusterMetadata.ClusterInformation {
+		if cluster.RPCName == "" {
+			// filling RPCName with a default value if empty
+			cluster.RPCName = "cadence-frontend"
 		}
+		c.ClusterMetadata.ClusterInformation[name] = cluster
```
yycptt, line 516: nit: I guess we can move this assignment back, into the if.

config/development.yaml
```diff
@@ -80,8 +80,7 @@ clusterMetadata:
...
+      rpcAddress: "localhost:7933" # this is to let system worker connected to the fronted end service. In cluster setup, localhost will not work
```
yycptt, line 83: nit: I think this comment is bit confusing as this address will also be used by other clusters for xdc requests. Maybe say something like: this will be used to populate publicClient rpcAddress to let system worker connect to .......

yycptt (approved):

config/development.yaml
longquanzheng: rpcAddress is not only for xdc. 
see https://github.com/uber/cadence/blob/a58b8b9670417ccbf2cd021882e750f95d6fc508/service/worker/batcher/workflow.go#L240
https://github.com/uber/cadence/blob/a58b8b9670417ccbf2cd021882e750f95d6fc508/service/worker/parentclosepolicy/client.go#L63
https://github.com/uber/cadence/blob/a58b8b9670417ccbf2cd021882e750f95d6fc508/service/worker/parentclosepolicy/workflow.go#L114

That's why a lot of OSS users already run into issues because they use localhost for this config. 

I will make it more clear. SO system worker should include XDC worker(I thought XDC processor is still in worker service looks like i am wrong, we have move history replicator to history service). 

So 
"this is to let worker service and XDC replicator connected to the frontend service"

yycptt: yeah I understand. Sorry, probably I didn't express myself clearly. I was trying to say the original comment may make people fill the address is **ONLY** for sys worker. 
New new comment looks great! :)

common/config/config.go
longquanzheng: ohh right.

PR #4366, opened 2021-08-15
Improve archival config validation

longquanzheng:
**What changed?**
Improve archival config validation

**Why?**
For https://github.com/uber/cadence/issues/4361

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/config/archival.go
+++ b/common/config/archival.go
@@ -28,28 +28,17 @@ import (
 
 // Validate validates the archival config
 func (a *Archival) Validate(domainDefaults *ArchivalDomainDefaults) error {
-	if !isArchivalConfigValid(a.History.Status, a.History.EnableRead, domainDefaults.History.Status, domainDefaults.History.URI, a.History.Provider != nil) {
-		return errors.New("Invalid history archival config")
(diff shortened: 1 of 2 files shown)


common/config/archival.go
```diff
@@ -28,28 +28,17 @@ import (
 
 // Validate validates the archival config
 func (a *Archival) Validate(domainDefaults *ArchivalDomainDefaults) error {
-	if !isArchivalConfigValid(a.History.Status, a.History.EnableRead, domainDefaults.History.Status, domainDefaults.History.URI, a.History.Provider != nil) {
-		return errors.New("Invalid history archival config")
+	if a.History.Status == common.ArchivalEnabled {
```
yycptt, line 31: If archival is disabled, I think we still need to check if "read" is disabled or not. Since read also requires the archiver to be enabled and initialized. 
For other two for domain default status and URI, I agree that we don't have to check them when archival is disabled.

longquanzheng: Okay, will add it back.

PR #4375, opened 2021-08-16
Try detecting timer and activity resurrection

yycptt:
**What changed?**
When processing user timer or activity timeout task, if a timer is overdue for a long time (by default, 24 hours), scan workflow history to see if that timer has already been fired or if that activity has ready finished.

**Why?**
For detecting timer/activity resurrection and prevent duplicated timer fired or activity completion.

**How did you test it?**
- Unit test. 

**Potential risks**
N/A in the worst case, we can increase the threshold for detecting resurrection to be a large number to turn the feature off.

**Release notes**

**Documentation Changes**


yux0 (approved):
Great changes. I think we can also have a similar thing to detect the mutable state and history not sync issue.

PR #4376, opened 2021-08-16
Fix ndc reset workflow replication bug

Shaddoll:
**What changed?**

1. ndc transaction manager calls `PersistNonStartWorkflowBatchEvents` when the workflow is a reset workflow
2. rename `PersistFirstWorkflowEvents` to `PersistStartWorkflowBatchEvents` and `PersistNonFirstWorkflowEvents` to `PersistNonStartWorkflowBatchEvents`

**Why?**
for reset workflows, we forked a branch in existing history, and don't need to create a new branch

rename the functions to make it less confusing for reset workflows

**How did you test it?**
existing test & new integration test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/ndc/new_workflow_transaction_manager.go
+++ b/service/history/ndc/new_workflow_transaction_manager.go
@@ -157,10 +157,20 @@ func (r *transactionManagerForNewWorkflowImpl) createAsCurrent(
 		return err
 	}
 
-	targetWorkflowHistorySize, err := targetWorkflow.GetContext().PersistFirstWorkflowEvents(
-		ctx,
-		targetWorkflowEventsSeq[0],
-	)
+	var targetWorkflowHistorySize int64
+	if len(targetWorkflowEventsSeq[0].Events) > 0 {
+		if targetWorkflowEventsSeq[0].Events[0].GetEventType() == types.EventTypeWorkflowExecutionStarted {
+			targetWorkflowHistorySize, err = targetWorkflow.GetContext().PersistStartWorkflowBatchEvents(
+				ctx,
+				targetWorkflowEventsSeq[0],
+			)
+		} else { // reset workflows fall into else branch
+			targetWorkflowHistorySize, err = targetWorkflow.GetContext().PersistNonStartWorkflowBatchEvents(
+				ctx,
+				targetWorkflowEventsSeq[0],
+			)
+		}
+	}
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


service/history/ndc/new_workflow_transaction_manager.go
```diff
@@ -227,10 +237,20 @@ func (r *transactionManagerForNewWorkflowImpl) createAsZombie(
...
-	)
+	var targetWorkflowHistorySize int64
+	if len(targetWorkflowEventsSeq[0].Events) > 0 {
+		if targetWorkflowEventsSeq[0].Events[0].GetEventType() == types.EventTypeWorkflowExecutionStarted {
+			targetWorkflowHistorySize, err = targetWorkflow.GetContext().PersistStartWorkflowBatchEvents(
+				ctx,
+				targetWorkflowEventsSeq[0],
+			)
+		} else { // reset workflows fall into else branch
+			targetWorkflowHistorySize, err = targetWorkflow.GetContext().PersistNonStartWorkflowBatchEvents(
```
longquanzheng, line 248: Do you know where does ndc create the branch record in history_tree table? 
On active side, it calls `ForkHistoryBranch` make that. 
https://github.com/uber/cadence/blob/44392dc5c7c4e0866c0cd73dc43709d5762532fb/service/history/reset/resetter.go#L421

Without that record, the branch info will be corrupted.

Shaddoll: it's created in this function https://github.com/uber/cadence/blob/master/service/history/ndc/history_replicator.go#L275

longquanzheng (approved):

service/history/ndc/new_workflow_transaction_manager.go
yux0: I may not have the full context here. But the createAs... is for startWorkflowBatchEvents. Is there a bug in reset to call this function with non start batch?

service/history/ndc/new_workflow_transaction_manager.go
```diff
@@ -157,10 +157,20 @@ func (r *transactionManagerForNewWorkflowImpl) createAsCurrent(
...
+				ctx,
+				targetWorkflowEventsSeq[0],
+			)
+		}
+	}
```
mkolodezny, line 173: can you provide please some context here?

service/history/ndc/new_workflow_transaction_manager.go
mkolodezny, line 248: idem

service/history/ndc/new_workflow_transaction_manager.go
Shaddoll: for reset workflows, the first event is not workflow started, but decisiontaskfailed or decisiontasktimedout. and the history of reset workflows are created from a forked branch of an existing workflows and they don't require creating a new branch.

longquanzheng: Just a note that this bug should only exists for sql because Cassandra uses upsert command for inserting

mkolodezny: makes sense, thanks @longquanzheng!

yux0 (approved):

PR #4379, opened 2021-08-17
Emit logs with workflow execution tag for timedout frontend requests

yycptt:
**What changed?**
- Emit log with domain/workflow/runID tag for timedout frontend requests
- For worker related APIs, count the request toward the domain level rate limiter

**Why?**
- Make investigation of request timeout easier

**How did you test it?**
- Unit test

**Potential risks**
- Pressure on logging
- Customer may likely encounter more service busy errors

**Release notes**

**Documentation Changes**


service/frontend/workflowHandler.go
```diff
@@ -722,18 +727,24 @@ func (wh *WorkflowHandler) RecordActivityTaskHeartbeat(
...
+		domainWrapper,
 	)
 	defer sw.Stop()
 
-	if wh.isShuttingDown() {
-		return nil, errShuttingDown
-	}
+	// Count the request in the RPS towards the requested domain,
+	// but we still accept it even if RPS is exceeded
+	wh.allow(domainWrapper)
```
yycptt, line 742: previously this `wh.allow(nil)` which won't consume RPS token for the requested domain, only token for the global RPS is consumed.

yux0 (approved):

PR #4385, opened 2021-08-18
Refactor ClusterMetadata defaults and validation

vytautas-karpavicius:
**What changed?**
Refactored ClusterMetadata configuration validation and filling in defaults.

**Why?**
To follow structure already established with other config sections. They have `validate` function that checks for invalid values and `fillDefaults` which sets default values if they are not provided.

ClusterMetadata had this logic scattered in a few other places. Collected them and put into separate `cluster.go` file.

I'm planning to add more validation/defaults to this section later, that is why I decided to refactor it first.

**How did you test it?**
Added new unit tests for all validation logic. Instead of panics, normal errors are returned that are easier to tests against.

I also suggest returning them as `multierr` as this would allow listing all the problems at once, instead of fix -> start server -> discover another error -> fix cycle.

**Potential risks**

**Release notes**

**Documentation Changes**


common/config/cluster.go
```diff
@@ -0,0 +1,139 @@
...
+		// PrimaryClusterName is the primary cluster name, only the primary cluster can register / update domain
+		// all clusters can do domain failover
+		PrimaryClusterName string `yaml:"primaryClusterName"`
+		// MasterClusterName is deprecated. Please use PrimaryClusterName.
+		MasterClusterName string `yaml:"masterClusterName"`
+		// CurrentClusterName is the name of the cluster of current deployment
+		CurrentClusterName string `yaml:"currentClusterName"`
+		// ClusterInformation contains information for each cluster within the replication group
+		// Key is the clusterName
+		ClusterInformation map[string]ClusterInformation `yaml:"clusterInformation"`
```
longquanzheng, line 47: Can we change this to "ReplicationGroup"? 

e.g. `ReplicationGroup map[string]ClusterInformation`

vytautas-karpavicius: Yeah, having the same name for both map name and item within that map does not seem right.

`ClusterGroup` seems reasonable. Or maybe just simply `Clusters`?
I think it would also make sense to rename `ClusterMetadata` to (maybe) `ClusterGroupMetadata` as it contains config for multiple clusters not just single one.

longquanzheng: I think having "Group" explicitly is more friendly as the "s' is quite hard to read/recognize. 

Yes, I love the idea of `ClusterGroupMetadata`

vytautas-karpavicius: Renamed:
- `ClusterInformation` to `ClusterGroup`
- `ClusterMetadata` to `ClusterGroupMetadata`
This also means introducing new fields in yaml deprecating old ones. Otherwise having them different could lead to some confusion.

common/config/config.go
```diff
@@ -44,8 +43,10 @@ type (
 		Persistence Persistence `yaml:"persistence"`
 		// Log is the logging config
 		Log Logger `yaml:"log"`
-		// ClusterMetadata is the config containing all valid clusters and active cluster
-		ClusterMetadata *ClusterMetadata `yaml:"clusterMetadata"`
+		// ClusterGroupMetadata is the config containing all valid clusters and active cluster
+		ClusterGroupMetadata *ClusterGroupMetadata `yaml:"clusterGroupMetadata"`
+		// ClusterMetadata is deprecated. Please use ClusterGroupMetadata
```
longquanzheng, line 48: To make it deprecated, start with `"Deprecated:" `
https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go/36360323

common/config/cluster.go
```diff
@@ -0,0 +1,147 @@
...
+		// all clusters can do domain failover
+		PrimaryClusterName string `yaml:"primaryClusterName"`
+		// MasterClusterName is deprecated. Please use PrimaryClusterName.
+		MasterClusterName string `yaml:"masterClusterName"`
+		// CurrentClusterName is the name of the cluster of current deployment
+		CurrentClusterName string `yaml:"currentClusterName"`
+		// ClusterGroup contains information for each cluster within the replication group
+		// Key is the clusterName
+		ClusterGroup map[string]ClusterInformation `yaml:"clusterGroup"`
+		// ClusterInformation is deprecated. Please use ClusterGroup.
```
longquanzheng, line 48: Same here: // Deprecated: ...

longquanzheng (approved):
Just a comment about the deprecation notation. Otherwise LGTM.

common/config/config.go
vytautas-karpavicius: Good tip! Updated.

PR #4390, opened 2021-08-19
gRPC for cross DC traffic

vytautas-karpavicius:
**What changed?**
- Updated `DispatcherProvider` interface to support creating dispatchers for gRPC. Updated `dnsDispatcherProvider` implementation to accordingly.
- Client factory will instantiate either Thrift or GRPC client based on the given dispatcher. These clients are used for:
  - admin client: for cross dc replication
  - frontend client: for request forwarding
- Updated config section within `ClusterGroupMetadata` to allow configuring cross DC transport. If not specified, it defaults to `tchannel`, as currently used.

**Why?**
To support gRPC for cross DC traffic.

**How did you test it?**
Integration tests for cross DC replication.

**Potential risks**
TChannel is still used by default, and gRPC is currently opt in to minimize risk.

**Release notes**
Updated CHANGELOG.md

**Documentation Changes**

### DIFF
--- a/client/clientBean.go
+++ b/client/clientBean.go
@@ -119,7 +119,14 @@ func NewClientBean(factory Factory, dispatcherProvider DispatcherProvider, clust
 			continue
 		}
 
-		dispatcher, err := dispatcherProvider.Get(info.RPCName, info.RPCAddress)
+		var dispatcher *yarpc.Dispatcher
+		var err error
+		switch info.RPCTransport {
+		case tchannel.TransportName:
+			dispatcher, err = dispatcherProvider.Get(info.RPCName, info.RPCAddress)
+		case grpc.TransportName:
+			dispatcher, err = dispatcherProvider.GetGRPC(info.RPCName, info.RPCAddress)
+		}
 		if err != nil {
 			return nil, err
 		}
--- a/client/clientBean.go
+++ b/client/clientBean.go
@@ -68,7 +68,7 @@ type (
 
 	// DispatcherProvider provides a dispatcher to a given address
 	DispatcherProvider interface {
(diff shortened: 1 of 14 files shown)


client/clientBean.go
```diff
@@ -65,9 +66,10 @@ type (
 		SetRemoteFrontendClient(cluster string, client frontend.Client)
 	}
 
-	// DispatcherProvider provides a diapatcher to a given address
+	// DispatcherProvider provides a dispatcher to a given address
 	DispatcherProvider interface {
 		Get(name string, address string) (*yarpc.Dispatcher, error)
```
Shaddoll, line 71: Can you rename it to GetTchannel to make it explicit?

client/clientfactory.go
```diff
@@ -358,3 +374,8 @@ func (cf *rpcClientFactory) newFrontendGRPCClient(hostAddress string) (frontend.
 		apiv1.NewVisibilityAPIYARPCClient(config),
 	), nil
 }
+
+func isGRPCOutbound(config transport.ClientConfig) bool {
+	namer, ok := config.GetUnaryOutbound().(transport.Namer)
```
Shaddoll, line 379: should we panic/log.fatal here if `ok==false`?

Shaddoll (approved):

client/clientBean.go
vytautas-karpavicius: Yes, updated.

client/clientfactory.go
vytautas-karpavicius: Unless old version of yarpc is used, this will not happen. But I guess explicit early panic is better. Updated.

PR #4392, opened 2021-08-19
Support ApplyParentClosePolicy Cross Cluster Tasks

demirkayaender:
**What changed?**
With these changes we can persist cross cluster requests to apply parent close policy

**Why?**
When the parent workflow closes, we need to apply the parent close policy on children that are under different domains and clusters.

**How did you test it?**
transferActiveTaskExecutorSuite

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/dataManagerInterfaces.go
```diff
@@ -610,6 +618,13 @@ type (
 		TargetCluster string
 	}
 
+	// CrossClusterApplyParentClosePolicyTask is the cross-cluster version of ApplyParentClosePolicyTask
+	CrossClusterApplyParentClosePolicyTask struct {
+		ApplyParentClosePolicyTask
```
demirkayaender, line 623: We can actually flatten this struct and only use `CrossClusterApplyParentClosePolicyTask`; not sure if we wanted to keep the convention. Let me know what you think.

common/types/shared.go
```diff
@@ -10696,13 +10696,70 @@ func (v *CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes) GetI
...
+func (v *AppyParentClosePolicyAttributes) GetParentClosePolicy() (o *ParentClosePolicy) {
+	if v != nil {
+		return v.ParentClosePolicy
+	}
+	return
+}
+
+type CrossClusterApplyParentClosePolicyRequestAttributes struct {
+	AppyParentClosePolicyAttributes []*AppyParentClosePolicyAttributes `json:"appyParentClosePolicyAttributes,omitempty"`
+	ParentClosePolicy               *ParentClosePolicy                 `json:"parentClosePolicy,omitempty"`
```
yycptt, line 10740: we can remove this field I think?

service/history/execution/mutable_state_task_generator.go
```diff
@@ -82,6 +82,10 @@ type (
 			transferTask *persistence.TransferTaskInfo,
 			targetCluster string,
 		) error
+		GenerateCrossClusterApplyParentClosePolicyTask(
```
yycptt, line 85: nit: probably add a TODO here saying ideally we should merge this method with `GenerateCrossClusterTaskFromTransferTask`?

yycptt (approved):
LGTM. Some minor comments.

PR #4398, opened 2021-08-20
Implement cross-cluster source task

yycptt:
**What changed?**
- Implement cross-cluster source task.
- The source task executor implementation will be in another diff. (Executor contains the logic for processing the task after a response is received from target cluster.)

**Why?**
- For supporting cross domain/cluster operations.

**How did you test it?**
- Unit test
- Test on buildkite should pass once IDL change is landed and IDL submodule is updated.

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_task.go
```diff
@@ -208,47 +240,401 @@ func newCrossClusterTaskBase(
...
+	if t.state != ctask.TaskStatePending || t.processingState != processingState(response.TaskState) {
+		// this might happen when:
+		// 1. duplicated response (shard movement in target cluster)
+		// 2. shard movement in source cluster causing task to be re-processed
+		// 3. task got invalidated during domain failover callback
+		return fmt.Errorf("unexpected task state, expected: %v, actual: %v", t.processingState, response.TaskState)
+	}
+
+	if t.GetTaskID() != response.GetTaskID() {
+		return fmt.Errorf("unexpected taskID, expected: %v, actual: %v", t.GetTaskID(), response.GetTaskID())
```
yux0, line 590: Should it just ignore the update?

yux0 (approved):

service/history/task/cross_cluster_task.go
yycptt: the contract for this method is returning an error is response is not recorded. So it needs to return an error here and leave the task state unchanged. How to handle the error is up to the queue processor, and I think we just need to log the error.

For this particular case I think it should not happen otherwise it will be a bug in the queue processor impl, so logging the error is necessary I think.

service/history/task/cross_cluster_task.go
```diff
@@ -35,14 +37,45 @@ import (
...
+// NOTE: DO NOT change the value for each state
+// new states must be added at the end to ensure backward compatibility
+// across clusters.
 const (
-	processingStateInitialized processingState = iota + 1
-	processingStateResponseReported
-	processingStateResponseRecorded
+	// processingStateInitialized is the initial state after a task is loaded
+	// task is available for poll at this state
+	processingStateInitialized processingState = 1
```
demirkayaender, line 59: Probably ALL_CAPS would be a better const naming.

service/history/task/cross_cluster_task.go
```diff
@@ -208,47 +265,396 @@ func newCrossClusterTaskBase(
...
+	}
 }
 
 func (t *crossClusterSourceTask) Nack() {
-	panic("Not implement")
+	// Nack will be called when we are unable to move the next step for processing
+	// basically add the task to redispatch queue, so it can be retried later.
+	// wether the task state is setting to Nacked or not is not important, since
+	// task will be retried forever
+	logEvent(t.eventLogger, "Nacked task")
```
demirkayaender, line 316: Are we going to have non-retryable errors? We talked about domains but as an example, workflow id might not exist; there could be some other cases we might know that retry will fail.

service/history/task/cross_cluster_task.go
```diff
@@ -208,47 +265,396 @@ func newCrossClusterTaskBase(
...
+	}
+
+	switch t.GetTaskType() {
+	case persistence.CrossClusterTaskTypeStartChildExecution:
+		attributes, taskState, err := t.getRequestForStartChildExecution(ctx, taskInfo, mutableState)
+		if err != nil || attributes == nil {
+			return nil, err
+		}
+		request.TaskInfo.TaskType = types.CrossClusterTaskTypeStartChildExecution.Ptr()
+		request.TaskInfo.TaskState = int16(taskState)
```
demirkayaender, line 445: nit: you can consider moving the common assignments like this to the end.

service/history/task/cross_cluster_task.go
```diff
@@ -208,47 +265,396 @@ func newCrossClusterTaskBase(
...
+
+func (t *crossClusterSourceTask) getRequestForStartChildExecution(
+	ctx context.Context,
+	taskInfo *persistence.CrossClusterTaskInfo,
+	mutableState execution.MutableState,
+) (*types.CrossClusterStartChildExecutionRequestAttributes, processingState, error) {
+	initiatedEventID := taskInfo.ScheduleID
+	childInfo, ok := mutableState.GetChildExecutionInfo(initiatedEventID)
+	if !ok {
+		return nil, t.processingState, nil
```
demirkayaender, line 478: are you intentionally returning nil error here? Is it expected? Same for the below ones.

demirkayaender (approved):

service/history/task/cross_cluster_task.go
yycptt: I think ALL_CAPS is against golang naming convention and will result in a lint warning like the following.
```
service/history/task/cross_cluster_task.go:59:2: [var-naming] don't use ALL_CAPS in Go names; use CamelCase
```
so will keep the existing naming.

service/history/task/cross_cluster_task.go
yycptt: sure. updated.

PR #4408, opened 2021-08-24
Add getFailoverInfo API

yux0:
**What changed?**
Add getFailoverInfoByDomainID API

**Why?**
To fetch on-going graceful failover info

**How did you test it?**
TODO

**Potential risks**
New API

**Release notes**

**Documentation Changes**


service/history/failover/coordinator.go
```diff
@@ -331,6 +364,7 @@ func (c *coordinatorImpl) notifyRemoteCoordinator(
 		if err != nil {
 			c.scope.IncCounter(metrics.FailoverMarkerNotificationFailure)
 			c.logger.Error("Failed to notify failover markers", tag.Error(err))
+			return
 		}
 
 		for marker := range requestByMarker {
```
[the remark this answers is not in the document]
yux0: > this for loop just for clearing `requestByMarker`? Looks like we can just return an error from this function, and let the caller clear the map by assigning a new empty map to it instead of using a for loop?

Yes. Remove the loop here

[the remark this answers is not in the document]
yux0: > Another question unrelated to this change: in the `aggregateNotificationRequests` func (L376) below, is the purpose to group `notificationRequest` by marker? But it looks like the key type of the map is a pointer to the marker instead of the marker itself. So if the marker used by different shards are different objects (I guess that's the case?), those markers won't actually be aggregated?

Yes. This is a bug. Changed it

service/history/failover/coordinator.go
```diff
@@ -53,23 +54,30 @@ const (
...
+)
+
 type (
 	// Coordinator manages the failover markers on sending and receiving
 	Coordinator interface {
 		common.Daemon
 
 		NotifyFailoverMarkers(shardID int32, markers []*types.FailoverMarkerAttributes)
 		ReceiveFailoverMarkers(shardIDs []int32, marker *types.FailoverMarkerAttributes)
+		GetFailoverInfo(domainID string) (int, []int32, error)
```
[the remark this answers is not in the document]
yux0: updated

client/history/grpcClient.go
```diff
@@ -25,6 +25,8 @@ import (
 
 	"go.uber.org/yarpc"
 
+	"github.com/uber/cadence/common/types/mapper/thrift"
+
```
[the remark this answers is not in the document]
yux0: updated

client/history/grpcClient.go
```diff
@@ -242,3 +243,8 @@ func (g grpcClient) TerminateWorkflowExecution(ctx context.Context, request *typ
 	_, err := g.c.TerminateWorkflowExecution(ctx, proto.FromHistoryTerminateWorkflowExecutionRequest(request), opts...)
 	return proto.ToError(err)
 }
+
+func (g grpcClient) GetFailoverInfo(ctx context.Context, request *types.GetFailoverInfoRequest, opts ...yarpc.CallOption) (*types.GetFailoverInfoResponse, error) {
+	response, err := g.c.GetFailoverInfo(ctx, proto.FromHistoryGetFailoverInfoRequest(request), opts...)
+	return proto.ToHistoryGetFailoverInfoResponse(response), thrift.ToError(err)
```
vytautas-karpavicius, line 249: Please use `proto.ToError(err)` here, as this is grpc client.

common/types/mapper/proto/history.go
```diff
@@ -249,6 +249,44 @@ func ToHistoryGetDLQReplicationMessagesResponse(t *historyv1.GetDLQReplicationMe
 	}
 }
 
+func FromHistoryGetFailoverInfoRequest(t *types.GetFailoverInfoRequest) *historyv1.GetFailoverInfoRequest {
```
yycptt, line 252: nit: shall we add some tests in `history_test.go`?

common/types/mapper/thrift/history.go
```diff
@@ -1245,3 +1245,45 @@ func ToProcessingQueueStateArrayMap(t map[string][]*history.ProcessingQueueState
 	}
 	return v
 }
+
+// FromGetFailoverInfoRequest converts internal GetFailoverInfoRequest type to thrift
+func FromGetFailoverInfoRequest(t *types.GetFailoverInfoRequest) *history.GetFailoverInfoRequest {
```
yycptt, line 1250: nit: I created a `thrift-tests` folder for testing thrift mappers. For newly added types, shall we also add some tests there?

yycptt (approved):

client/history/grpcClient.go
yux0: Good catch

common/types/mapper/proto/history.go
yux0: Add unit tests

common/types/mapper/thrift/history.go
yux0: added unit tests

PR #4429, opened 2021-08-29
Allow replicate history generated from same cluster

yux0:
**What changed?**
This covers the case when the history has replicated to remote clusters. Allow the history to replicate back to the source cluster if the history is missing in the source cluster.

**Why?**
1. Some DLQ messages are due to the ban of this replication. This will reduce DLQ messages.
2. Help disaster recover.
3. Update the LoadWorkflowExecute. Not all the calls need to check buffer events.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/context.go
```diff
@@ -351,31 +339,6 @@ func (c *contextImpl) LoadWorkflowExecution(
 		)
 	}
 
-	flushBeforeReady, err := c.mutableState.StartTransaction(domainEntry)
```
yycptt, line 354: We no longer need to `StartTransaction` after loading persistence mutable state? I thought it will also update the current version for the workflow?

yycptt (approved):

PR #4433, opened 2021-08-31
Matching per domain RPS limit

yycptt:
**What changed?**
- Add per domain rps limit for matching service

**Why?**
- For better multi-tenancy. Sometimes one domain generates lots of activity tasks and consumes all matching rps quota on the host and causing delays for other domains decision/activity task.

**How did you test it?**
Integration test.

**Potential risks**
 - N/A, by default the domain rps is not enabled and domain rps is equal to host rps limit.

**Release notes**

**Documentation Changes**


yycptt:
Changes in `.gen` folder is auto-generated after running  `rm -fr .gen` and `make clean && make bins`

service/matching/config.go
```diff
@@ -100,6 +101,7 @@ func NewConfig(dc *dynamicconfig.Collection) *Config {
 		PersistenceGlobalMaxQPS:         dc.GetIntProperty(dynamicconfig.MatchingPersistenceGlobalMaxQPS, 0),
 		EnableSyncMatch:                 dc.GetBoolPropertyFilteredByTaskListInfo(dynamicconfig.MatchingEnableSyncMatch, true),
 		RPS:                             dc.GetIntProperty(dynamicconfig.MatchingRPS, 1200),
+		DomainRPS:                       dc.GetIntProperty(dynamicconfig.MatchingDomainRPS, 0),
```
yux0, line 104: Need to filter by domain

service/matching/handler.go
```diff
@@ -80,9 +80,19 @@ func NewHandler(
...
 		config:        config,
 		metricsClient: resource.GetMetricsClient(),
-		rateLimiter: quotas.NewDynamicRateLimiter(func() float64 {
-			return float64(config.RPS())
-		}),
+		rateLimiter: quotas.NewMultiStageRateLimiter(
+			func() float64 {
+				return float64(config.RPS())
+			},
+			func(domain string) float64 {
```
yux0, line 87: is the domain used

service/matching/config.go
yycptt: 👍  Thanks for catching. Updated.

service/matching/handler.go
yycptt: Used after changed the config to be filtered by domain.

PR #4434, opened 2021-08-31
Add a default all for signal name

yux0:
**What changed?**
Add a default all for signal name

**Why?**
The signal name may create too many metrics series

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/metrics/context.go
+++ b/common/metrics/context.go
@@ -35,6 +35,15 @@ func TagContext(ctx context.Context, tag Tag) context.Context {
 	return context.WithValue(ctx, contextTagsKey, tags)
 }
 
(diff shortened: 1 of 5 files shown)


demirkayaender (approved):

Groxx:
ehhh... domains are not all _that_ high cardinality, and with this change we lose domain-level rollups.

if anything, we might want to triple-emit.  or do something in the scope implementation to allow custom rollups, for systems with different orders of magnitude of cardinality support.

Groxx (approved):
yea, false is probably the safe option.  and 👍 for a debugging-oriented config section.

PR #4435, opened 2021-08-31
Revert ratelimiting behavior for frontend worker related APIs

yycptt:
**What changed?**
- Revert ratelimiting behavior for frontend worker related APIs

**Why?**
- https://github.com/uber/cadence/pull/4379 changes for ratelimiting behavior for frontend worker related APIs so that those calls will be counted towards domain rps limit. This might break existing configs we set for customers and limit customer's capability of freely scaling their workers. So revert to old behavior.
- We will have the capability to specify domain rps limit on matching service which can help us prevent one domain's open workflows from making progress too fast.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -737,9 +737,9 @@ func (wh *WorkflowHandler) RecordActivityTaskHeartbeat(
 	)
 	defer sw.Stop()
 
-	// Count the request in the RPS towards the requested domain,
+	// Count the request in the host RPS,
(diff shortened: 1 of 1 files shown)


mkolodezny (commented):
should add the global rps domain limit for matching first?

mkolodezny (approved):
assuming the change just rollbacks the https://github.com/uber/cadence/pull/4379/files

PR #4445, opened 2021-09-02
Implement cross-cluster source task executor

yycptt:
**What changed?**
- Implement cross-cluster source task executor

**Why?**
For supporting cross-cluster operations

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -22,30 +22,410 @@
...
+	}
+
+	// handle common errors
+	failedCause := task.response.FailedCause
+	if failedCause != nil {
+		switch *failedCause {
+		case types.CrossClusterTaskFailedCauseDomainNotActive:
+			return t.generateNewTask(ctx, wfContext, mutableState, task)
+		case types.CrossClusterTaskFailedCauseWorkflowNotExists:
+			return &types.EntityNotExistsError{}
```
yux0, line 152: what about provide some error detail

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -22,30 +22,410 @@
...
+
+		childRunID := task.response.StartChildExecutionAttributes.GetRunID()
+		err := recordChildExecutionStarted(ctx, taskInfo, wfContext, attributes, childRunID, now)
+		if err != nil {
+			return err
+		}
+
+		// advance the task state so that it will be available for polling again
+		t.setTaskState(task, ctask.TaskStatePending, processingStateResponseRecorded)
+		return errContinueExecution
```
yux0, line 186: return error here to prevent task from acked?

yycptt: `errContinueExecution` is returned here and then ignored on L107. The whole Execute() don't have to return an error and it's ok to call Ack() method on the task since Ack() method won't actually set the task state to acked, so the task is still pending and available for polling.

service/history/task/cross_cluster_source_task_executor.go
yycptt: Sure I can add the details here, but I think the entire entity not exists error will simply be ignored in the handleError() method of cross cluster task type as this is an expected error.

yux0 (approved):

PR #4458, opened 2021-09-03
Set limit on range queries to prevent bad queries causing degradation

Shaddoll:
**What changed?**

- Refactored SQL plugins to split point deletes and range deletes
- Limit range deletes on timer/transfer/replication/cross-cluster tasks and run range deletes by batch (SQL only)
- Limit range queries on transfer/cross-cluster tasks query (NoSQL already supports this, add support for SQL)
- Limit range deletes on history nodes and run range deletes by batch within a single transaction (SQL only)
- Introduce TaskDeleteBatchSize dynamic config

**Why?**
It's a good practice to set limit on range queries to avoid a single bad query causing degradation of database.

**How did you test it?**
Existing tests

**Potential risks**

**Release notes**
There is some breaking change in sqlplugin interface.

**Documentation Changes**


common/persistence/sql/sqlExecutionStore.go
```diff
@@ -791,12 +794,23 @@ func (m *sqlExecutionStore) RangeCompleteTransferTask(
 	ctx context.Context,
 	request *p.RangeCompleteTransferTaskRequest,
 ) error {
-
-	if _, err := m.db.DeleteFromTransferTasks(ctx, &sqlplugin.TransferTasksFilter{
-		ShardID:   m.shardID,
-		MinTaskID: &request.ExclusiveBeginTaskID,
-		MaxTaskID: &request.InclusiveEndTaskID}); err != nil {
-		return convertCommonErrors(m.db, "RangeCompleteTransferTask", "", err)
+	for {
```
Shaddoll, line 797: It doesn't need to be done within a single transaction. It guarantees that if the function returns no error, all target rows will be deleted.

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -740,9 +740,20 @@ func (m *sqlExecutionStore) GetTransferTasks(
...
+			return nil, convertCommonErrors(m.db, "GetTransferTasks", "failed to deserialize page token", err)
+		}
+		minReadLevel = readLevel
+	}
 	rows, err := m.db.SelectFromTransferTasks(ctx, &sqlplugin.TransferTasksFilter{
-		ShardID: m.shardID, MinTaskID: &request.ReadLevel, MaxTaskID: &request.MaxReadLevel})
+		ShardID:   m.shardID,
+		MinTaskID: minReadLevel,
+		MaxTaskID: request.MaxReadLevel,
+		PageSize:  request.BatchSize,
```
demirkayaender, line 755: There are 2 upper bound constraints here. Did you check if this works fine? I'm wondering if it would be a problem when the last row's taskID < maxReadLevel (i.e. if there's any component in the stack depending on that logic)

Even if this is fine, it would be useful to write a unittest to make sure this scenario won't be broken in the future.

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -770,6 +781,12 @@ func (m *sqlExecutionStore) GetTransferTasks(
 			Version:                 info.GetVersion(),
 		}
 	}
+	if len(rows) > 0 {
+		lastTaskID := rows[len(rows)-1].TaskID
+		if lastTaskID < request.MaxReadLevel {
+			resp.NextPageToken = serializePageToken(lastTaskID)
```
demirkayaender, line 787: regarding the comment above, will this adjustment take care of any potential issues?

common/persistence/sql/sqlExecutionStore.go
```diff
@@ -859,15 +897,22 @@ func (m *sqlExecutionStore) CompleteCrossClusterTask(
 func (m *sqlExecutionStore) RangeCompleteCrossClusterTask(
 	ctx context.Context,
 	request *p.RangeCompleteCrossClusterTaskRequest,
-) error {
-	if _, err := m.db.DeleteFromCrossClusterTasks(ctx, &sqlplugin.CrossClusterTasksFilter{
+) (*p.RangeCompleteCrossClusterTaskResponse, error) {
+	result, err := m.db.RangeDeleteFromCrossClusterTasks(ctx, &sqlplugin.CrossClusterTasksFilter{
```
demirkayaender, line 901: This is our first time introducing pagination during deletes. How are we going to continue deleting?

common/persistence/sql/sqlHistoryStore.go
```diff
@@ -434,9 +440,20 @@ func (m *sqlHistoryStore) DeleteHistoryBranch(
...
+				if err != nil {
+					return err
+				}
+				rowsAffected, err := result.RowsAffected()
+				if err != nil {
+					return err
+				}
+				if rowsAffected < _defaultHistoryNodeDeleteBatch ||
+					rowsAffected == persistence.UnknownNumRowsAffected ||
+					rowsAffected > _defaultHistoryNodeDeleteBatch {
```
demirkayaender, line 454: nit, align the condition lines with the same indentation.

demirkayaender: How's `rowsAffected > _defaultHistoryNodeDeleteBatch` possible?

common/persistence/sql/sqlHistoryStore.go
```diff
@@ -434,9 +440,20 @@ func (m *sqlHistoryStore) DeleteHistoryBranch(
 				// No any branch is using this range, we can delete all of it
 				nodeFilter.MinNodeID = br.BeginNodeID
 			}
-			_, err := tx.DeleteFromHistoryNode(ctx, nodeFilter)
-			if err != nil {
-				return err
+			for {
```
demirkayaender, line 443: okay so this for loop is supposed to handle the delete pagination

service/history/queue/timer_queue_processor.go
```diff
@@ -414,11 +414,21 @@ func (t *timerQueueProcessor) completeTimer() error {
...
+			InclusiveBeginTimestamp: t.ackLevel,
+			ExclusiveEndTimestamp:   newAckLevelTimestamp,
+			PageSize:                pageSize, // pageSize may or may not be honored
+		})
+		if err != nil {
+			return err
+		}
+		if resp.TasksCompleted < pageSize || // all target tasks are deleted
+			resp.TasksCompleted == persistence.UnknownNumRowsAffected || // underlying database does not support rows affected, so pageSize is not honored and all target tasks are deleted
+			resp.TasksCompleted > pageSize { // pageSize is not honored and all tasks are deleted
```
demirkayaender, line 429: why would it not be honored?

demirkayaender (commented):
I think the changes look good overall, but we need to add some testing for the new changes.

service/history/queue/timer_queue_processor.go
Shaddoll: We allow them not to be honored.  In case, there is some database doesn't support `Delete Order By Limit`.

common/persistence/sql/sqlHistoryStore.go
Shaddoll: We allow pagesize not to be honored. In case, there is some database doesn't support Delete Order By Limit.

common/persistence/sql/sqlExecutionStore.go
Shaddoll: There are loops in queue processor.

common/persistence/sql/sqlExecutionStore.go
Shaddoll: We already have test cases.

Shaddoll: see https://github.com/uber/cadence/blob/master/common/persistence/persistence-tests/persistenceTestBase.go#L1274

common/persistence/sql/sqlExecutionStore.go
Shaddoll: I've updated the tests to cover the new cases.

common/persistence/persistence-tests/persistenceTestBase.go
```diff
@@ -1353,9 +1353,21 @@ Loop:
...
+		resp, err := s.ExecutionManager.RangeCompleteReplicationTask(ctx, &p.RangeCompleteReplicationTaskRequest{
+			InclusiveEndTaskID: inclusiveEndTaskID,
+			PageSize:           1,
+		})
+		if err != nil {
+			return err
+		}
+		if resp.TasksCompleted < 1 ||
+			resp.TasksCompleted == p.UnknownNumRowsAffected ||
+			resp.TasksCompleted > 1 {
```
demirkayaender, line 1366: nit: please keep the indentation for `if` the same

common/persistence/sql/sqlplugin/mysql/execution.go
```diff
@@ -329,15 +344,20 @@ func (mdb *db) SelectFromTimerTasks(ctx context.Context, filter *sqlplugin.Timer
...
+// RangeDeleteFromTimerTasks deletes multi rows from timer_tasks table
+func (mdb *db) RangeDeleteFromTimerTasks(ctx context.Context, filter *sqlplugin.TimerTasksFilter) (sql.Result, error) {
+	filter.MinVisibilityTimestamp = mdb.converter.ToMySQLDateTime(filter.MinVisibilityTimestamp)
+	filter.MaxVisibilityTimestamp = mdb.converter.ToMySQLDateTime(filter.MaxVisibilityTimestamp)
+	if filter.PageSize > 0 {
+		return mdb.conn.ExecContext(ctx, rangeDeleteTimerTaskByBatchQuery, filter.ShardID, filter.MinVisibilityTimestamp, filter.MaxVisibilityTimestamp, filter.PageSize)
 	}
-	*filter.VisibilityTimestamp = mdb.converter.ToMySQLDateTime(*filter.VisibilityTimestamp)
-	return mdb.conn.ExecContext(ctx, deleteTimerTaskQuery, filter.ShardID, *filter.VisibilityTimestamp, filter.TaskID)
+	return mdb.conn.ExecContext(ctx, rangeDeleteTimerTaskQuery, filter.ShardID, filter.MinVisibilityTimestamp, filter.MaxVisibilityTimestamp)
```
demirkayaender, line 360: shouldn't this be `deleteTimerTaskQuery` ?

common/persistence/sql/sqlplugin/postgres/execution.go
```diff
@@ -274,10 +284,15 @@ func (pdb *db) SelectFromTransferTasks(ctx context.Context, filter *sqlplugin.Tr
 
 // DeleteFromTransferTasks deletes one or more rows from transfer_tasks table
 func (pdb *db) DeleteFromTransferTasks(ctx context.Context, filter *sqlplugin.TransferTasksFilter) (sql.Result, error) {
-	if filter.MinTaskID != nil {
-		return pdb.conn.ExecContext(ctx, rangeDeleteTransferTaskQuery, filter.ShardID, *filter.MinTaskID, *filter.MaxTaskID)
```
demirkayaender, line 278: was this intentional delete?
it does look like we have the wrong filter here if this condition is true though...

demirkayaender (commented):
There seems to be some updates to functions like `RangeDeleteFromTimerTasks`, could you confirm that those changes were intentional?

common/persistence/sql/sqlplugin/mysql/execution.go
Shaddoll: I split delete function into point delete and range delete

common/dynamicconfig/constants.go
```diff
@@ -917,6 +917,13 @@ const (
 	// Allowed filters: N/A
 
 	TimerTaskBatchSize
+	// TimerTaskDeleteBatchSize is batch size for timer processor to delete timer tasks
+	// KeyName: history.timerTaskDeleteBatchSize
+	// Value type: Int
+	// Default value: 4000
+	// Allowed filters: N/A
+
```
yycptt, line 931: nit: I guess the empty line should on L913 and 920 instead of 918 and 925?

nit: I am considering group the read batch size for all task types into one config and probably we can do the same for other configs as well. Not sure how useful it is to have separate configs.

demirkayaender (approved):

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -306,14 +306,23 @@ func (c *crossClusterQueueProcessorBase) completeAckTasks(nextAckLevel int64) (i
...
+			InclusiveEndTaskID:   nextAckLevel,
+			PageSize:             pageSize, // pageSize may or may not be honored
+		})
+		if err != nil {
+			return c.ackLevel, err
+		}
+		if resp.TasksCompleted < pageSize || // all target tasks are deleted
+			resp.TasksCompleted == persistence.UnknownNumRowsAffected || // underlying database does not support rows affected, so pageSize is not honored and all target tasks are deleted
+			resp.TasksCompleted > pageSize { // pageSize is not honored and all tasks are deleted
+			break
```
yycptt, line 323: nit: can we create some util method for this logic, since it's a bit complicated.  Duplicating it for each queue processor is kind of error-prone when we need to modify it. 

I guess if the only why we use the returned response is through this check, we can also just return if the deletion is completed?

service/history/replication/dlq_handler.go
```diff
@@ -219,7 +219,7 @@ func (r *dlqHandlerImpl) MergeMessages(
 		}
 	}
 
-	err = r.shard.GetExecutionManager().RangeDeleteReplicationTaskFromDLQ(
+	_, err = r.shard.GetExecutionManager().RangeDeleteReplicationTaskFromDLQ(
```
yycptt, line 222: we don't need to specify the batch size and check the result here?

Shaddoll: this is called in an api handler, i don't want to make it fail in the middle. it make break the semantics of the api

PR #4461, opened 2021-09-05
Validate shard ID range

yux0:
**What changed?**
Validate shard ID range

**Why?**
Invalid shard could start in history which the total shard number is incorrect.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/controller.go
+++ b/service/history/shard/controller.go
@@ -260,7 +260,7 @@ func (c *controller) shardClosedCallback(shardID int, shardItem *historyShardsIt
 }
 
 func (c *controller) getOrCreateHistoryShardItem(shardID int) (*historyShardsItem, error) {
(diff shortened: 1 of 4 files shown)


service/history/shard/controller.go
```diff
@@ -255,6 +260,15 @@ func (c *controller) shardClosedCallback(shardID int, shardItem *historyShardsIt
 }
 
 func (c *controller) getOrCreateHistoryShardItem(shardID int) (*historyShardsItem, error) {
+	if shardID >= c.config.NumberOfShards { // zero based shard ID
```
yycptt, line 263: nit: will shardID < 0?

yycptt (approved):

service/history/shard/controller.go
yux0: good call out. We should check it as well

PR #4467, opened 2021-09-07
Fix workflow retry policy overriding workflow timeout

fenghan89:
**What changed? **
Updated CreateHistoryStartWorkflowRequest common/util.go to only override workflow expiration timestamp with the retry policy timeout if the retry policy timeout exceeds workflow timeout

**Why?**
https://t3.uberinternal.com/browse/CDNC-660

**How did you test it?**
Added TestCreateHistoryStartWorkflowRequest_RetryWorkflowExpirationTimeout

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/host/retry_policy_workflow_test.go
+++ b/host/retry_policy_workflow_test.go
@@ -46,8 +46,8 @@ func (s *IntegrationSuite) TestWorkflowRetryPolicyTimeout() {
 	retryPolicy := &types.RetryPolicy{
 		InitialIntervalInSeconds:    1,
 		BackoffCoefficient:          2.0,
-		MaximumIntervalInSeconds:    5,
-		ExpirationIntervalInSeconds: 5,
+		MaximumIntervalInSeconds:    1,
+		ExpirationIntervalInSeconds: 1,
 		MaximumAttempts:             0,
 		NonRetriableErrorReasons:    []string{"bad-error"},
 	}
@@ -58,8 +58,8 @@ func (s *IntegrationSuite) TestWorkflowRetryPolicyTimeout() {
 		WorkflowType:                        workflowType,
 		TaskList:                            taskList,
 		Input:                               nil,
-		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(10),
-		TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(10),
+		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(3),
+		TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(3),
 		Identity:                            identity,
 		RetryPolicy:                         retryPolicy,
 	}
@@ -68,7 +68,7 @@ func (s *IntegrationSuite) TestWorkflowRetryPolicyTimeout() {
 	s.Nil(err0)
 
 	s.Logger.Info("StartWorkflowExecution", tag.WorkflowRunID(we.RunID))
-	time.Sleep(11 * time.Second) // wait 1 second for timeout
+	time.Sleep(3 * time.Second) // wait 1 second for timeout
 
 	historyResponse, err := s.engine.GetWorkflowExecutionHistory(createContext(), &types.GetWorkflowExecutionHistoryRequest{
 		Domain: s.domainName,
@@ -78,24 +78,19 @@ func (s *IntegrationSuite) TestWorkflowRetryPolicyTimeout() {
 	})
 	s.Nil(err)
 	history := historyResponse.History
-	s.NotNil(history.Events[0])
-	s.Equal(history.Events[0].EventType, types.EventTypeWorkflowExecutionStarted.Ptr())
-	expirationTime := history.Events[0].WorkflowExecutionStartedEventAttributes.ExpirationTimestamp
+	firstEvent := history.Events[0]
+	s.NotNil(firstEvent)
+	s.Equal(firstEvent.EventType, types.EventTypeWorkflowExecutionStarted.Ptr())
+	expirationTime := firstEvent.WorkflowExecutionStartedEventAttributes.ExpirationTimestamp
 	s.NotNil(expirationTime)
 	delta := time.Unix(0, *expirationTime).Sub(now)
-	s.True(delta > 5*time.Second)
-	s.True(delta < 6*time.Second)
+	s.True(delta > 0*time.Second)
+	s.True(delta < 2*time.Second)
 
-	wfExecution, err := s.engine.DescribeWorkflowExecution(createContext(), &types.DescribeWorkflowExecutionRequest{
-		Domain: s.domainName,
-		Execution: &types.WorkflowExecution{
-			WorkflowID: id,
-		},
-	})
-	s.NotNil(wfExecution)
-	delta = time.Unix(0, *wfExecution.WorkflowExecutionInfo.CloseTime).Sub(time.Unix(0, *wfExecution.WorkflowExecutionInfo.StartTime))
-	s.True(delta > 9*time.Second)
-	s.True(delta < 11*time.Second)
+	lastEvent := history.Events[len(history.Events)-1]
+	delta = time.Unix(0, common.Int64Default(lastEvent.Timestamp)).Sub(time.Unix(0, common.Int64Default(firstEvent.Timestamp)))
+	s.True(delta > 2*time.Second)
+	s.True(delta < 4*time.Second)
 }
 
 func (s *IntegrationSuite) TestWorkflowRetryPolicyContinueAsNewAsRetry() {
@@ -113,8 +108,8 @@ func (s *IntegrationSuite) TestWorkflowRetryPolicyContinueAsNewAsRetry() {
 	retryPolicy := &types.RetryPolicy{
 		InitialIntervalInSeconds:    1,
 		BackoffCoefficient:          2.0,
-		MaximumIntervalInSeconds:    5,
-		ExpirationIntervalInSeconds: 5,
+		MaximumIntervalInSeconds:    3,
+		ExpirationIntervalInSeconds: 3,
 		MaximumAttempts:             0,
 		NonRetriableErrorReasons:    []string{"bad-error"},
 	}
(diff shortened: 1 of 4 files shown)


common/util.go
```diff
@@ -521,8 +521,12 @@ func CreateHistoryStartWorkflowRequest(
 	histRequest.FirstDecisionTaskBackoffSeconds = Int32Ptr(firstDecisionTaskBackoffSeconds)
 
 	if startRequest.RetryPolicy != nil && startRequest.RetryPolicy.GetExpirationIntervalInSeconds() > 0 {
-		expirationInSeconds := startRequest.RetryPolicy.GetExpirationIntervalInSeconds() + firstDecisionTaskBackoffSeconds
+		expirationInSeconds := Int32Default(startRequest.ExecutionStartToCloseTimeoutSeconds)
+		if startRequest.RetryPolicy.GetExpirationIntervalInSeconds() > expirationInSeconds {
+			expirationInSeconds = startRequest.RetryPolicy.GetExpirationIntervalInSeconds()
+		}
```
yycptt, line 527: - If execution timeout is 10h, retry expiration interval is 1h and workflow fails after 2h. With the new logic, I think the workflow will be retried as the expiration is 10h? But looks like we should not do the retry since the retry expiration has passed?

- For later attempts of the workflow, I think we will still encounter the situation where workflow timeout time is after retry expiration time and workflow will timeout earlier by the retry expiration.

- The second point means we kinda have an inconsistency of behavior for first attempts and later attempts since we guarantee the first attempt won't be cut off by retry expiration but later attempts will. I guess it would be better to keep the behavior consistent?

fenghan89: @yycptt 

`If execution timeout is 10h, retry expiration interval is 1h and workflow fails after 2h. With the new logic, I think the workflow will be retried as the expiration is 10h? But looks like we should not do the retry since the retry expiration has passed?`
Hm, maybe I misunderstood what the RetryPolicy.ExpirationIntervalInSeconds means; My interpretation is that if the workflow retries, but in the retry takes longer than 2h, it will timeout. Are you saying that the retry expiration means in the case when 2h is pass we will not attempt a retry? 

`For later attempts of the workflow, I think we will still encounter the situation where workflow timeout time is after retry expiration time and workflow will timeout earlier by the retry expiration.`
I think related to my first point -- if the workflow is retried, should we not timeout earlier by retry expiration? And a followup question -- should the the RetryPolicy's expiration interval be absolute or per retry (i.e. if its 1h does that mean a) each retry should timeout after 1h or b) if 1h is past from StartWorkflow no retries will be attempted)

As a followup question for subsequent retries, I'm seeing that the method I modified `CreateHistoryStartWorkflowRequest` is called in `startWorkflowWithRetry`, `StartWorkflowExecution`, `SignalWithStartWorkflowExecution` and `terminateAndStartWorkflow`, If i were to make it consistent such that retries will not time out earlier -- do you know which other method should I make the modification in?

yycptt: - yeah, that's my understanding of the ticket.  
- For the second and third comments, my main point is it looks like the behavior will be inconsistent between the first and later attempts. I am not leaning towards any of them though, both sounds reasonable to me. 
- This is just my thought. We can discuss with the team to reach a conclusion.

My understanding it the expiration interval is absolute and is calculated and converted to expiration timestamp once when starting the workflow (or the beginning of each run for cron). Then the expiration timestamp will be used across all retires.

You may want to look into how workflow timeout task is generated: https://github.com/uber/cadence/blob/master/service/history/execution/mutable_state_task_generator.go#L138
and how the expiration timestamp is carried over to the next run:
https://github.com/uber/cadence/blob/master/service/history/execution/mutable_state_builder.go#L1700

common/util.go
```diff
@@ -519,13 +519,16 @@ func CreateHistoryStartWorkflowRequest(
 	}
 
 	histRequest.FirstDecisionTaskBackoffSeconds = Int32Ptr(firstDecisionTaskBackoffSeconds)
-
+	expirationInSeconds := Int32Default(startRequest.ExecutionStartToCloseTimeoutSeconds)
```
yycptt, line 522: If the user doesn't specify retry expiration timeout, only the max retry attempts, assigning this value will add an additional retry condition and limit the retry to happen within the workflow timeout I think.

fenghan89: I think if this isn't set then `GenerateWorkflowStartTasks` will perform the same logic so this will be a no-op in that case? since expiration timestamp will be now + Int32Default(startRequest.ExecutionStartToCloseTimeoutSeconds) and in GenerateWorkflowStartTasks we are calculating `workflowTimeoutTimestamp := startTime.Add(workflowTimeoutDuration)` which should give us the same values if retry expiration is not set

fenghan89:
@yycptt  I updated the PR according to what we discussed, with 1 change. I updated the expiration time according to the maximum of retry timeout and wf timeout. Since in our discussions i think the understanding was that retry timeout should always be at least as large as wf timeout and we shouldn't allow "irrational" configurations. (i.e. when wf timeout is > retry timeout).

service/history/execution/mutable_state_builder.go
```diff
@@ -1661,21 +1661,22 @@ func (e *mutableStateBuilder) addWorkflowExecutionStartedEventForContinueAsNew(
...
 	}
+
+	// if ContinueAsNew as Cron or decider, recalculate the expiration timestamp and set attempts to 0
+	req.Attempt = 0
+	if attributes.RetryPolicy != nil && attributes.RetryPolicy.GetExpirationIntervalInSeconds() > 0 {
+		// expirationTime calculates from first decision task schedule to the end of the workflow
+		expirationInSeconds := attributes.RetryPolicy.GetExpirationIntervalInSeconds() + req.GetFirstDecisionTaskBackoffSeconds()
+		expirationTime := e.timeSource.Now().Add(time.Second * time.Duration(expirationInSeconds))
+		req.ExpirationTimestamp = common.Int64Ptr(expirationTime.UnixNano())
+	}
```
yycptt, line 1672: Is the logic the same as the old one? There's no behavior change and it's just a code refactoring?

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,319 @@
...
+		TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(10),
+		Identity:                            identity,
+		RetryPolicy:                         retryPolicy,
+	}
+	now := time.Now()
+	we, err0 := s.engine.StartWorkflowExecution(createContext(), request)
+	s.Nil(err0)
+
+	s.Logger.Info("StartWorkflowExecution", tag.WorkflowRunID(we.RunID))
+	time.Sleep(11 * time.Second) // wait 1 second for timeout
```
yycptt, line 71: Can we use a smaller workflow timeout like 3s? Wait 10s here seems a bit long for a test, and all three tests need 35s to complete.

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,319 @@
...
+			WorkflowID: id,
+		},
+	})
+	s.NotNil(wfExecution)
+	delta = time.Unix(0, *wfExecution.WorkflowExecutionInfo.CloseTime).Sub(time.Unix(0, *wfExecution.WorkflowExecutionInfo.StartTime))
+	s.True(delta > 9*time.Second)
+	s.True(delta < 11*time.Second)
+}
+
+func (s *IntegrationSuite) TestWorkflowRetryPolicyContinueAsNewAsRetry() {
```
yycptt, line 101: nit: Do we still need the following two tests given that there's no logic change in the continue as new path?

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,319 @@
...
+		return []byte(strconv.Itoa(0)), []*types.Decision{{
+			DecisionType: types.DecisionTypeContinueAsNewWorkflowExecution.Ptr(),
+			ContinueAsNewWorkflowExecutionDecisionAttributes: &types.ContinueAsNewWorkflowExecutionDecisionAttributes{
+				WorkflowType:                        workflowType,
+				TaskList:                            &types.TaskList{Name: tl},
+				Input:                               buf.Bytes(),
+				ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(10),
+				TaskStartToCloseTimeoutSeconds:      common.Int32Ptr(10),
+				RetryPolicy:                         retryPolicy,
+				Initiator:                           types.ContinueAsNewInitiatorRetryPolicy.Ptr(),
```
yycptt, line 150: Continue as new is an implementation detail inside cadence server and `Initiator` should not be specified by caller. The decision task handler only needs to return a fail workflow decision and the server will convert it to a continue as new event with the right initiator.

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,319 @@
...
+			s.Logger.Warn("Execution not timedout yet.")
+			time.Sleep(200 * time.Millisecond)
+			continue GetHistoryLoop
+		}
+
+		timeoutEventAttributes := lastEvent.WorkflowExecutionTimedOutEventAttributes
+		s.Equal(types.TimeoutTypeStartToClose, *timeoutEventAttributes.TimeoutType)
+		break GetHistoryLoop
+	}
+	historyResponse, err := s.engine.GetWorkflowExecutionHistory(createContext(), &types.GetWorkflowExecutionHistoryRequest{
```
yycptt, line 192: We can reuse the history we get in the for loop?

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,319 @@
...
+		},
+	})
+	s.Nil(err)
+	history := historyResponse.History
+	s.NotNil(history.Events[0])
+	s.Equal(history.Events[0].EventType, types.EventTypeWorkflowExecutionStarted.Ptr())
+	numAttempts := history.Events[0].WorkflowExecutionStartedEventAttributes.Attempt
+	s.Equal(numAttempts, int32(1))
+
+	wfExecution, err := s.engine.DescribeWorkflowExecution(createContext(), &types.DescribeWorkflowExecutionRequest{
```
yycptt, line 205: We can simply use the timestamp for the last event here?

service/history/execution/mutable_state_builder.go
fenghan89: yup, this is just refactoring

host/retry_policy_workflow_test.go
fenghan89: I don't think we have a test for these flows as far as i know, so thought i would add some

host/retry_policy_workflow_test.go
```diff
@@ -0,0 +1,290 @@
...
+
+	we, err0 := s.engine.StartWorkflowExecution(createContext(), request)
+	s.Nil(err0)
+
+	s.Logger.Info("StartWorkflowExecution", tag.WorkflowRunID(we.RunID))
+	dtHandler := func(execution *types.WorkflowExecution, wt *types.WorkflowType,
+		previousStartedEventID, startedEventID int64, history *types.History) ([]byte, []*types.Decision, error) {
+		buf := new(bytes.Buffer)
+		return []byte(strconv.Itoa(0)), []*types.Decision{{
+			DecisionType: types.DecisionTypeContinueAsNewWorkflowExecution.Ptr(),
```
yycptt, line 233: similar to retry policy, continue as new for cron is also create by server, decision task handler is only needs to return workflow completed here. In start workflow request, you'll need to specify a cron schedule. like `@every 3s` and you may need to recalculate the sleep time below.

yycptt (approved):
LGTM overall. One more comment for integration test. Feel free to land after address it. Thanks!

PR #4472, opened 2021-09-09
Fix workflow refresh for closed workflows

yycptt:
**What changed?**
- Allow workflow refresh for closed workflows
- Remove duplicated/buggy GetCloseEvent() method in mutableState
- Remove unused parameter from workflow refresher, task generator, mutableState and timerSequence
- Handle get history client error when calling RefreshWorkflowTask 
- Remove defer() in CapturePanic() as this function is already called in defer

**Why?**
- Bug fix for refresh closed workflows

**How did you test it?**
- Tested locally

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_task_generator.go
+++ b/service/history/execution/mutable_state_task_generator.go
@@ -40,6 +40,7 @@ type (
 	// MutableStateTaskGenerator generates workflow transfer and timer tasks
 	MutableStateTaskGenerator interface {
 		GenerateWorkflowStartTasks(
+			now time.Time,
 			startEvent *types.HistoryEvent,
 		) error
 		GenerateWorkflowCloseTasks(
@@ -131,6 +132,7 @@ func NewMutableStateTaskGenerator(
 }
 
 func (r *mutableStateTaskGeneratorImpl) GenerateWorkflowStartTasks(
+	now time.Time,
 	startEvent *types.HistoryEvent,
 ) error {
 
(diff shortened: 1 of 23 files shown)


longquanzheng (approved):

yux0 (approved):

service/history/execution/mutable_state_task_generator.go
```diff
@@ -145,10 +136,11 @@ func (r *mutableStateTaskGeneratorImpl) GenerateWorkflowStartTasks(
 
 	executionInfo := r.mutableState.GetExecutionInfo()
 	startVersion := startEvent.GetVersion()
+	startTimestamp := time.Unix(0, startEvent.GetTimestamp())
 
 	workflowTimeoutDuration := time.Duration(executionInfo.WorkflowTimeout) * time.Second
 	workflowTimeoutDuration = workflowTimeoutDuration + firstDecisionDelayDuration
-	workflowTimeoutTimestamp := now.Add(workflowTimeoutDuration)
+	workflowTimeoutTimestamp := startTimestamp.Add(workflowTimeoutDuration)
```
yycptt, line 143: Looks like we still need to pass in a timestamp here since workflow reset needs to use reset time instead of start event time for calculating the timeout timestamp. 
Will revert the changes for this function to preserve the now parameter.

longquanzheng: How about change the parameter name to something else. It’s not really now. Also a comment would help understand why we don’t use start event timestamp

longquanzheng: Eg: timeoutStartTime

yycptt: sure. I renamed it to `startTime` and added a comments says for reset the workflow start time should be the reset time in order to re-calculated the timeout timestamp.

ceearrashee:
Hello!
The fix that you made did not include in the last releases. 
Can you guide me on when this fix will be released and in which release?

longquanzheng:
@ceearrashee Hello, it's included in the next minor release branch: https://github.com/uber/cadence/commits/0.23.x
v0.23.0 is being baked and will be out soon. Stay tuned.

PR #4474, opened 2021-09-09
Execution logic for RecordChildCompletion and ApplyParentClosePolicy

demirkayaender:
**What changed?**
Added source cluster execution logic for RecordChildCompletion and ApplyParentClosePolicy

Also added new cross cluster task types here: https://github.com/uber/cadence-idl/pull/86

**Why?**
We need to be able to notify cross cluster parents and children upon closing a workflow.

**How did you test it?**
Buildkite pass

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_task.go
```diff
@@ -470,6 +486,88 @@ func (t *crossClusterSourceTask) GetCrossClusterRequest() (request *types.CrossC
 	return request, nil
 }
 
+func (t *crossClusterSourceTask) VerifyTaskFailoverVersion(
```
yycptt, line 489: nit: since this util function specifically checks for last write version, maybe rename to something like `verifyLastWriteVersion`?

service/history/task/cross_cluster_task.go
```diff
@@ -470,6 +486,88 @@ func (t *crossClusterSourceTask) GetCrossClusterRequest() (request *types.CrossC
...
+) error {
+	lastWriteVersion, err := mutableState.GetLastWriteVersion()
+	if err != nil {
+		return err
+	}
+	ok, err := verifyTaskVersion(t.shard, t.logger, taskInfo.DomainID, lastWriteVersion, taskInfo.Version, taskInfo)
+	if err != nil || !ok {
+		return err
+	}
+	return nil
```
yycptt, line 501: if `ok == false` and `err == nil`, the function will return `nil`. Then the execution for the caller will continue, while we should skip the task by returning `nil, processingState, nil` in caller to indicate the task is a no-op I think.

service/history/task/cross_cluster_task.go
```diff
@@ -470,6 +486,88 @@ func (t *crossClusterSourceTask) GetCrossClusterRequest() (request *types.CrossC
...
+	completionEvent, err := mutableState.GetCompletionEvent(ctx)
+	if err != nil {
+		return nil, t.processingState, err
+	}
+
+	attributes := &types.CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes{
+		TargetDomainID:   taskInfo.TargetDomainID,
+		InitiatedEventID: initiatedEventID,
+		CompletionEvent:  completionEvent,
+	}
```
yycptt, line 566: we also need to specify `TargetWorkflowID` and `TargetRunID` here I guess?

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -222,6 +226,97 @@ func (t *crossClusterTargetTaskExecutor) executeCancelExecutionTask(
...
+	ctx context.Context,
+	task *crossClusterTargetTask,
+) (*types.CrossClusterApplyParentClosePolicyResponseAttributes, error) {
+	attributes := task.request.ApplyParentClosePolicyAttributes
+	if attributes == nil {
+		return nil, errMissingTaskRequestAttributes
+	}
+
+	for _, childAttrs := range attributes.AppyParentClosePolicyAttributes {
+		tagetDomainName, err := t.verifyDomainActive(childAttrs.ChildDomainID)
```
yycptt, line 239: nit: typo `targetDomainName`

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -222,6 +226,97 @@ func (t *crossClusterTargetTaskExecutor) executeCancelExecutionTask(
...
+			scope.IncCounter(metrics.ParentClosePolicyProcessorFailures)
+			return nil, err
+		}
+		scope.IncCounter(metrics.ParentClosePolicyProcessorSuccess)
+	}
+
+	// TODO: Consider going through all the children, even if some fail to apply the parent policy,
+	// and add the policy application status by a child identifier (domain-wf-run id)
+	// Right now, return value is all or none even if we were able apply the policy on some children successfuly
+	return &types.CrossClusterApplyParentClosePolicyResponseAttributes{}, nil
```
yycptt, line 268: 👍

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -222,6 +226,97 @@ func (t *crossClusterTargetTaskExecutor) executeCancelExecutionTask(
...
+func (t *crossClusterTargetTaskExecutor) executeApplyParentClosePolicyTask(
+	ctx context.Context,
+	task *crossClusterTargetTask,
+) (*types.CrossClusterApplyParentClosePolicyResponseAttributes, error) {
+	attributes := task.request.ApplyParentClosePolicyAttributes
+	if attributes == nil {
+		return nil, errMissingTaskRequestAttributes
+	}
+
+	for _, childAttrs := range attributes.AppyParentClosePolicyAttributes {
```
yycptt, line 238: shall we also add a sanity check for `task.processingState` here?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -96,6 +97,10 @@ func (t *crossClusterSourceTaskExecutor) Execute(
 		err = t.executeCancelExecutionTask(ctx, sourceTask)
 	case persistence.CrossClusterTaskTypeSignalExecution:
 		err = t.executeSignalExecutionTask(ctx, sourceTask)
+	case persistence.CrossClusterTaskTypeRecordChildWorkflowExeuctionComplete:
+		err = t.executeRecordChildWorkflowExecutionCompleteTask(ctx, sourceTask)
+	case persistence.CrossClusterTaskTypeApplyParentPolicy:
+		err = t.executeApplyParentClosePolicyTask(ctx, sourceTask)
 	// TODO: implement the execution logic for recordChildCompleted and applyParentClosePolicy task
```
yycptt, line 104: nit: remove `TODO`?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -269,6 +265,113 @@ func (t *crossClusterSourceTaskExecutor) executeCancelExecutionTask(
...
+	task *crossClusterSourceTask,
+) (retError error) {
+	taskInfo := task.GetInfo().(*persistence.CrossClusterTaskInfo)
+	wfContext, mutableState, release, err := loadWorkflowForCrossClusterTask(ctx, t.executionCache, taskInfo, t.metricsClient, t.logger)
+	if err != nil || mutableState == nil {
+		return err
+	}
+	defer func() { release(retError) }()
+
+	err = task.VerifyTaskFailoverVersion(mutableState, taskInfo)
```
yycptt, line 310: same comment as the one in `cross_cluster_task.go`, if `verifyTaskVersion` returned inside this util method returned `false, nil`, we should also return nil here to skip the task.

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -222,6 +227,105 @@ func (t *crossClusterTargetTaskExecutor) executeCancelExecutionTask(
...
+		return nil, err
+	}
+
+	// child info is in taskInfo
+	taskInfo, ok := task.Info.(*persistence.CrossClusterTaskInfo)
+	if !ok {
+		return nil, errUnexpectedTask
+	}
+
+	switch task.processingState {
```
yycptt, line 301: nit: we can remove this switch since we've verified the processingState on L280?

service/history/task/cross_cluster_task.go
```diff
@@ -470,6 +486,90 @@ func (t *crossClusterSourceTask) GetCrossClusterRequest() (request *types.CrossC
...
+) (bool, error) {
+	lastWriteVersion, err := mutableState.GetLastWriteVersion()
+	if err != nil {
+		return false, err
+	}
+	ok, err := verifyTaskVersion(t.shard, t.logger, taskInfo.DomainID, lastWriteVersion, taskInfo.Version, taskInfo)
+	if err != nil || !ok {
+		return false, err
+	}
+	return true, nil
```
yycptt, line 501: nit: this is the same as `return ok, err`?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -142,6 +151,8 @@ func (t *crossClusterSourceTaskExecutor) executeStartChildExecutionTask(
 		return t.generateNewTask(ctx, wfContext, mutableState, task)
 	}
 
+	// handle common errors
+
```
yycptt, line 155: nit: remove duplicated comment?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -269,6 +281,84 @@ func (t *crossClusterSourceTaskExecutor) executeCancelExecutionTask(
...
+	}
+
+	// handle common errors
+	failedCause := task.response.FailedCause
+	if failedCause != nil {
+		if _, ok := expectedCloseErrors[*failedCause]; !ok {
+			return errUnexpectedErrorFromTarget
+		}
+	}
+
```
yycptt, line 311: We still need to handle `CrossClusterTaskFailedCauseDomainNotActive` here right? by generating new transfer/cross cluster tasks. Same for recordChildCompletion method below.

service/history/task/cross_cluster_task.go
```diff
@@ -470,6 +486,90 @@ func (t *crossClusterSourceTask) GetCrossClusterRequest() (request *types.CrossC
...
+
+	attributes := &types.CrossClusterApplyParentClosePolicyRequestAttributes{}
+	children := mutableState.GetPendingChildExecutionInfos()
+	for _, childInfo := range children {
+		targetDomainEntry, err := t.shard.GetDomainCache().GetDomain(childInfo.DomainName)
+		if err != nil {
+			return nil, t.processingState, err
+		}
+		targetCluster := targetDomainEntry.GetReplicationConfig().ActiveClusterName
+		if targetCluster == t.targetCluster {
```
yycptt, line 525: Looks like if they are not equal, that child will be dropped? 

I am thinking the case where there're two childworkflows in two different domains, one of them did a failover while we are fetching the task and causing `targetCluster != t.targetCluster`, I guess in this case, we will need to create a new task for this failed over domain. 
We can still send the child to the target cluster (by removing the check) and when receiving the domain not active error, source task executor will create the new task. Or mark the task as invalid here and return an error, again the source task executor will split the task into two.

Another idea is that when converting close execution task to apply parent close task, we can generate one applyParentClose task for each target (child) domain, so we'll never need to split the task when processing it.

service/history/task/transfer_active_task_executor.go
```diff
@@ -1700,11 +1695,14 @@ func (t *transferActiveTaskExecutor) processParentClosePolicy(
 
 	for initiatedID, childDomainID := range sameClusterChildDomainIDs {
 		childInfo := childInfos[initiatedID]
-		if err := t.applyParentClosePolicy(
+		if err := applyParentClosePolicy(
```
yycptt, line 1698: I thought we agreed on sending a signal to the system workflow for processing parent close policy if the child domain may potentially be cross-cluster? Or that will be in another pr? 

I am totally fine with having another pr for it. Actually, you can add a TODO here and I can implement that logic. Let's get this pr landed first.

yycptt (approved):

service/history/task/transfer_active_task_executor.go
demirkayaender: Let me put a comment here, I'm not 100% sure if we need to use a system workflow to apply parent policy.

PR #4476, opened 2021-09-10
Emit metrics during shardscanner fix and scan activities

mantas-sidlauskas:
**What changed?**
Tagged shardscanner metrics will be emitted during every scan/fix. 
Tags include: 
* domain name
* Invariant name
* Operation
* Result

**Why?**
This should replace all aggregated metrics for shards. Aggregated metrics are emitted only after workflow is finished and must be queried to extract specific stats for a domain.

Using tagged metrics can answer all questions from metric aggregation side (ex: group invariants by domain name, etc.) 

**How did you test it?**

### DIFF
--- a/common/metrics/tags.go
+++ b/common/metrics/tags.go
@@ -31,19 +31,21 @@ const (
 	buildVersionTag = "build_version"
 	goVersionTag    = "go_version"
 
-	instance       = "instance"
-	domain         = "domain"
-	targetCluster  = "target_cluster"
-	activeCluster  = "active_cluster"
-	taskList       = "tasklist"
-	taskListType   = "tasklistType"
-	workflowType   = "workflowType"
-	activityType   = "activityType"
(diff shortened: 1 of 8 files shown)


common/metrics/tags.go
```diff
@@ -135,6 +137,16 @@ func DecisionTypeTag(value string) Tag {
 	return metricWithUnknown(decisionType, value)
 }
 
+// ShardscannerScanResult returns a new shardscanner scan result type tag.
+func ShardscannerScanResult(value string) Tag {
```
just-at-uber, line 141: small thing but perhaps capitalize scanner (and all other relevant areas within PR):
ShardScannerScanResult

just-at-uber (approved):

mkolodezny (approved):
lgtm

mkolodezny (commented):
Does it apply to the overall scanner framework (including executions ones)?

PR #4483, opened 2021-09-12
Update get tasklist response

yux0:
**What changed?**
Update get tasklist response include poller info

**Why?**
Save one call to get poller info from describe domain

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


client/matching/client.go
```diff
@@ -258,24 +258,29 @@ func (c *clientImpl) GetTaskListsByDomain(
...
 	for _, future := range futures {
 		var resp *types.GetTaskListsByDomainResponse
 		if err = future.Get(ctx, &resp); err != nil {
 			return nil, err
 		}
-		for _, tl := range resp.TaskListNames {
-			if _, ok := taskListMap[tl]; !ok {
-				taskListMap[tl] = struct{}{}
+		for name, tl := range resp.GetDecisionTaskListMap() {
+			if _, ok := decisionTaskListMap[name]; !ok {
```
yycptt, line 269: if `ok`, do we need to merge the results for pollerInfo?

common/types/mapper/proto/api.go
```diff
@@ -1766,18 +1766,52 @@ func FromGetTaskListsByDomainResponse(t *types.GetTaskListsByDomainResponse) *ap
...
+	resp := &apiv1.GetTaskListsByDomainResponse{}
+	decisionTaskList := t.GetDecisionTaskListMap()
+	activityTaskList := t.GetActivityTaskListMap()
+	var decisionTaskListMap, activityTaskListMap map[string]*apiv1.DescribeTaskListResponse
+	if len(decisionTaskList) > 0 {
+		decisionTaskListMap = make(map[string]*apiv1.DescribeTaskListResponse, len(decisionTaskList))
+		for name, tl := range decisionTaskList {
+			decisionTaskListMap[name] = FromDescribeTaskListResponse(tl)
+		}
+	}
```
yycptt, line 1778: nit: can we create another set of From/To methods for the response map and reuse it for both activity and decision task list map? Then this method can also be simplified to something like
```
if t == nil {
    return nil
}

return &apiv1.GetTaskListsByDomainResponse{
    DecisionTaskListMap: FromDescribeTaskListResponseMap(t.GetDecisionTaskListMap()),
    ActivityTaskListMap: FromDescribeTaskListResponseMap(t.GetActivityTaskListMap()),
}
```

common/types/mapper/proto/matching.go
```diff
@@ -222,18 +222,52 @@ func FromMatchingGetTaskListsByDomainResponse(t *types.GetTaskListsByDomainRespo
 	if t == nil {
 		return nil
 	}
-	return &matchingv1.GetTaskListsByDomainResponse{
-		TaskListNames: t.GetTaskListNames(),
+	resp := &matchingv1.GetTaskListsByDomainResponse{}
+	decisionTaskList := t.GetDecisionTaskListMap()
+	activityTaskList := t.GetActivityTaskListMap()
+	var decisionTaskListMap, activityTaskListMap map[string]*matchingv1.DescribeTaskListResponse
+	if len(decisionTaskList) > 0 {
```
yycptt, line 229: nit: same here

common/types/shared.go
```diff
@@ -4636,13 +4636,22 @@ func (v *GetTaskListsByDomainRequest) GetDomain() (o string) {
 
 // GetTaskListsByDomainResponse is an internal type (TBD...)
 type GetTaskListsByDomainResponse struct {
-	TaskListNames []string `json:"taskListNames,omitempty"`
```
yycptt, line 4639: Just want to confirm that making breaking changes here is ok as no users (internal/open source) are actively using it?

tools/cli/adminTaskListCommands.go
```diff
@@ -87,8 +87,31 @@ func AdminListTaskList(c *cli.Context) {
 	}
 
 	fmt.Println("Task Lists for domain " + domain + ":")
-	for _, taskList := range response.GetTaskListNames() {
-		fmt.Println(taskList)
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetBorder(true)
+	table.SetColumnSeparator("|")
+	table.SetHeader([]string{"Task List Name", "Type", "Poller Number"})
```
yycptt, line 93: nit: `Poller Count`?

tools/cli/adminTaskListCommands.go
```diff
@@ -87,8 +87,31 @@ func AdminListTaskList(c *cli.Context) {
...
+	table.SetHeaderLine(true)
+	table.SetHeaderColor(tableHeaderBlue, tableHeaderBlue, tableHeaderBlue)
+	for name, taskList := range response.GetDecisionTaskListMap() {
+		table := tablewriter.NewWriter(os.Stdout)
+		table.SetBorder(true)
+		table.SetColumnSeparator("|")
+		table.SetHeader([]string{"Task List Name", "Type", "Poller Number"})
+		table.SetHeaderLine(true)
+		table.SetHeaderColor(tableHeaderBlue, tableHeaderBlue, tableHeaderBlue)
```
yycptt, line 102: I may miss something. why do we need to create a new table and set the header, color, border again here?

common/types/shared.go
yux0: Yes. No user is using this

tools/cli/adminTaskListCommands.go
yux0: No need. removed those.

tools/cli/adminTaskListCommands.go
yux0: updated

client/matching/client.go
yux0: Yes. It should merge the result

common/types/mapper/proto/api.go
yux0: good point. Update it

tools/cli/adminTaskListCommands.go
```diff
@@ -87,8 +87,19 @@ func AdminListTaskList(c *cli.Context) {
...
-		fmt.Println(taskList)
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetBorder(true)
+	table.SetColumnSeparator("|")
+	table.SetHeader([]string{"Task List Name", "Type", "Poller Count"})
+	table.SetHeaderLine(true)
+	table.SetHeaderColor(tableHeaderBlue, tableHeaderBlue, tableHeaderBlue)
+	for name, taskList := range response.GetDecisionTaskListMap() {
+		table.Append([]string{name, strconv.Itoa(len(taskList.GetPollers()))})
+		table.Render()
```
yycptt, line 98: we only need to `Render()` once at the very end I think?

client/matching/client.go
```diff
@@ -258,24 +258,31 @@ func (c *clientImpl) GetTaskListsByDomain(
...
+				decisionTaskListMap[name] = tl
 			}
+			decisionTaskListMap[name].Pollers = append(decisionTaskListMap[name].Pollers, tl.GetPollers()...)
```
yycptt, line 272: Do we need to put this append in `else`? looks like otherwise the pollers will be duplicated if `name` is not in the map previously?

yux0: yes

tools/cli/adminTaskListCommands.go
yux0: yes

yycptt (approved):

client/matching/client.go
```diff
@@ -258,24 +258,33 @@ func (c *clientImpl) GetTaskListsByDomain(
...
 			return nil, err
 		}
-		for _, tl := range resp.TaskListNames {
-			if _, ok := taskListMap[tl]; !ok {
-				taskListMap[tl] = struct{}{}
+		for name, tl := range resp.GetDecisionTaskListMap() {
+			if _, ok := decisionTaskListMap[name]; !ok {
+				decisionTaskListMap[name] = tl
+			} else {
+				decisionTaskListMap[name].Pollers = append(decisionTaskListMap[name].Pollers, tl.GetPollers()...)
```
yycptt, line 272: It would be great if we can add some unit tests for client implementation that has special logic.

yux0: We don't have any right now. I created a task to cover this folder.

PR #4486, opened 2021-09-13
Fix parent close policy for cross-domain childs

yycptt:
**What changed?**
- Allow applying parent close policy for cross-domain child workflows.

**Why?**
- Current implementation of parent close policy assumes all child workflows are in the same domain as the parent.

**How did you test it?**
- Added unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/task/transfer_active_task_executor.go
+++ b/service/history/task/transfer_active_task_executor.go
@@ -1634,8 +1634,6 @@ func (t *transferActiveTaskExecutor) processParentClosePolicy(
 	childInfos map[int64]*persistence.ChildExecutionInfo,
 	crossClusterTaskGenerators *[]generatorF,
 ) error {
-	domainID := task.DomainID
-
 	if len(childInfos) == 0 {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


service/history/task/transfer_active_task_executor.go
```diff
@@ -1696,11 +1698,12 @@ func (t *transferActiveTaskExecutor) processParentClosePolicy(
 			})
 	}
 
-	for _, childInfo := range sameClusterChildInfos {
+	for initiatedID, childDomainID := range sameClusterChildDomainIDs {
+		childInfo := childInfos[initiatedID]
```
demirkayaender, line 1702: I thought keeping the child info in the value field was just a reference. Was it taking a full copy of childInfo? If so, we could just use the reference to the value to keep here simpler to avoid querying the info with eventID again?

demirkayaender (approved):
Changes look good to me, left a small comment

service/history/task/transfer_active_task_executor.go
yycptt: yes, childInfo is a reference. I changed the type of `sameClusterChildXXX` to record the child domainID, so that we don't need to query domain cache again here. 
I considered storing both information, which requires either defining a new type to hold both domainID and childInfo, or use two slices and match them by index. Both approaches kinda make the code more complicated so I ended up using this approach. 
What's your thought on this? Happy to change the implementation.

demirkayaender: I see, this looks good to me. I don't have a strong opinion.

PR #4489, opened 2021-09-14
Remove aggregated metrics from shard scanner

mantas-sidlauskas:
**What changed?**
Removing aggregated metrics from shard scanner

**Why?**
With recent metrics addition #4476, aggregated metrics are not needed anymore. 

**How did you test it?**
Unit test/local test

**Potential risks**

**Release notes**

**Documentation Changes**


service/worker/scanner/shardscanner/activities.go
```diff
@@ -383,39 +381,3 @@ func fixShard(
 	}
 	return &report, nil
 }
-
-// ScannerEmitMetricsActivity will emit metrics for a complete run of ShardScanner
-func ScannerEmitMetricsActivity(
```
emrahs, line 388: Thanks for the clean-up @mantas-sidlauskas! Overall, I like it but I think we should keep **ShardSuccessCount** and **ShardControlFlowFailureCount** so that we can keep an eye on the health of the scanner and perhaps create alerts for.  I'd suggest either keeping these as aggregated metrics or emitting from the scan activity just like you did in your PR #4476. What do you think?

service/worker/scanner/shardscanner/scanner_workflow.go
```diff
@@ -169,16 +165,7 @@ func (wf *ScannerWorkflow) Start(ctx workflow.Context) error {
 		}
 	}
 
-	activityCtx = getShortActivityContext(ctx)
-	summary := wf.Aggregator.GetStatusSummary()
-
-	return workflow.ExecuteActivity(activityCtx, ActivityScannerEmitMetrics, ScannerEmitMetricsActivityParams{
```
emrahs, line 175: This makes the scanner workflow non-deterministic. Even though we can manage this easily by restarting the workflow, it would be an annoyance for open-source users. We should either use GetVersion API or document this as a breaking change in the release notes to let users know that they should restart the scanner if it's running it. Given the small impact, I'd go with the release notes option.

mantas-sidlauskas: Good point Emrah, 
I will revert back this activity to not break the determinism here, but will reduce the list of metrics we emit instead

PR #4490, opened 2021-09-15
Add a hard limit for number of decision retry attempts with a dynamic…

emrahs:
Add a hard limit for number of decision retry attempts with a dynamic config

**What changed?**
Added a new dynamic config to limit the maximum number of times a decision can be retried, and added code to automatically terminate the workflows that exceed this limit.

**Why?**
Unlimited retries, combined with poorly designed workflows and/or persistence issues were causing a snowball effect on the server-side and worsening degradations. This new limit will stop such workflows from continuously adding load to the server.

**How did you test it?**
Unit tests, and manual local tests using a sample workflow and artificially introduced decision completion failure. 

**Potential risks**
Since this commit is replacing an unlimited retry flow to a fixed number of retries, some poorly designed Cadence workflows may start being terminated although they had a chance to succeed in the past.

**Release notes**

**Documentation Changes**


service/history/config/config.go
```diff
@@ -502,6 +503,7 @@ func New(dc *dynamicconfig.Collection, numberOfShards int, storeType string, isA
 		StickyTTL:                         dc.GetDurationPropertyFilteredByDomain(dynamicconfig.StickyTTL, time.Hour*24*365),
 		DecisionHeartbeatTimeout:          dc.GetDurationPropertyFilteredByDomain(dynamicconfig.DecisionHeartbeatTimeout, time.Minute*30),
 		DecisionRetryCriticalAttempts:     dc.GetIntProperty(dynamicconfig.DecisionRetryCriticalAttempts, 10), // about 30m
+		DecisionRetryMaxAttempts:          dc.GetIntPropertyFilteredByDomain(dynamicconfig.DecisionRetryMaxAttempts, 1000),
```
yux0, line 506: should we allow infinite retry if this set to 0?

emrahs: That's a good idea. Will do it.

service/history/decision/handler.go
```diff
@@ -841,6 +843,25 @@ func failDecisionHelper(
...
+		executionInfo := mutableState.GetExecutionInfo()
+		handler.logger.Error("Decision attempt exceeds limit.",
+			tag.WorkflowDomainID(executionInfo.DomainID),
+			tag.WorkflowID(executionInfo.WorkflowID),
+			tag.WorkflowRunID(executionInfo.RunID))
+
+		if _, err := mutableState.AddWorkflowExecutionTerminatedEvent(
+			mutableState.GetNextEventID(),
+			common.FailureReasonDecisionAttemptsExceedsLimit,
+			[]byte("Decision attempt exceeded limit."),
```
yycptt, line 858: nit: shall we use `cause` and/or `details` parameter here?

service/history/decision/handler.go
```diff
@@ -480,7 +480,8 @@ Update_History_Loop:
 				tag.WorkflowID(token.WorkflowID),
 				tag.WorkflowRunID(token.RunID),
 				tag.WorkflowDomainID(domainID))
-			msBuilder, err = failDecisionHelper(ctx, wfContext, scheduleID, startedID, failCause, []byte(failMessage), request)
+			msBuilder, err = handler.failDecisionHelper(ctx, wfContext, scheduleID, startedID, failCause,
```
yycptt, line 483: If workflow is terminated here, do we need to exit the method and return an error like what we did for `TransactionSizeLimitError`? I am not sure if the rest of the logic here and in the frontend handler still work for terminated workflow.

service/history/decision/handler.go
emrahs: These two are in the DecisionTaskFailedEvent already, so I don't think it's very useful to add them termination event, too.

service/history/decision/handler.go
emrahs: It's too early to terminate here since the mutable state needs to be updated (line 547). All other code is safe to execute since the state of the workflow transitions to Terminated. Having said that, the very last section in the function about dispatching activity/decision tasks is pointless if the workflow is no longer running. I'll add an early return there.

service/history/decision/handler.go
```diff
@@ -841,6 +848,26 @@ func failDecisionHelper(
 		return nil, err
 	}
 
+	domainName := domainEntry.GetInfo().Name
+	maxAttempts := handler.config.DecisionRetryMaxAttempts(domainName)
+	if maxAttempts > 0 && mutableState.GetExecutionInfo().DecisionAttempt > int64(maxAttempts) {
+		message := "Decision attempt exceeds limit"
+		executionInfo := mutableState.GetExecutionInfo()
+		handler.logger.Error(message,
```
yycptt, line 856: nit: not sure if we want to emit a metric for the force termination here.

yycptt (approved):
LGTM! Sorry for the delay.

service/history/decision/handler.go
yycptt: hmmm, in most cases yes. But if we are unlucky and the first decision failure is due to decision timeout then later failed events and reason won't be recorded in history I think. 

A similar issue is when using the `--non-deterministic-only` flag when doing a batch reset. Most workflow's last event is decisionTaskFailed event which contains the detailed info for the non-deterministic error. But a small number of them are end by decisionTaskTimeout event, which can't be caught by the `--non-deterministic-only` flag.

PR #4491, opened 2021-09-15
Managed failover workflow improvements

yux0:
**What changed?**
1. Support graceful failover in managed failover workflow.
2. Check remote task list poller number prior to failing over a domain.

**Why?**
1. Support batch failover
2. Prevent dis-config and no worker on the to-be-active cluster.

**How did you test it?**
Unit  tests update

**Potential risks**

**Release notes**

**Documentation Changes**


service/worker/failovermanager/workflow.go
```diff
@@ -461,3 +486,22 @@ func cleanupChannel(channel workflow.Channel) {
...
 	}
 }
+
+func validateTaskListPollerInfo(ctx context.Context, feClient frontend.Client, domain string) error {
+	resp, err := feClient.GetTaskListsByDomain(ctx, &types.GetTaskListsByDomainRequest{Domain: domain})
+	if err != nil {
+		return fmt.Errorf("failed to get task list for domain %s", domain)
+	}
+	for name, tl := range resp.GetDecisionTaskListMap() {
+		if len(tl.GetPollers()) == 0 {
```
yycptt, line 496: so we only need to check if there's a poller or not in the target cluster, instead of verifying it for all active task lists in source cluster?

yycptt (approved):

service/worker/failovermanager/workflow.go
yux0: The active task lists will be put into the cache in passive matching. So I guess we don't need to

yycptt: Interesting... I thought activity/decision will be pushed to standby matching only when those two tasks are discarded by transfer queue processing logic. Is that what you're referring to, or it's something else?

yux0: Oh, I see. then we need to query both side.

PR #4493, opened 2021-09-16
Refactor cross cluster queue implementation

yycptt:
**What changed?**
- Implemented common ordered map collection
- Manage all ready-for-poll tasks in cross-cluster queue using an ordered map
- During failover, only validates tasks whose source/target domain did a failover
- Misc bug fixes & improvement in cross-cluster queue implementation

**Why?**
Having a separate component for managing all ready-for-poll tasks can:
- Serve as an extension point for improve multi-tenancy in cross-cluster queue implementation
- On polling, we don't have to get all tasks from ack managed, check if they are ready for poll, etc.

**How did you test it?**
- Unit test

**Potential risks**
- N/A, new feature, not enabled yet.

**Release notes**

**Documentation Changes**


service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -353,62 +349,57 @@ func (c *crossClusterQueueProcessorBase) readTasks(
 
 func (c *crossClusterQueueProcessorBase) pollTasks() []*types.CrossClusterTaskRequest {
 
-	// TODO: control the number of tasks returned
-	var result []*types.CrossClusterTaskRequest
-	for _, queueTask := range c.getTasks() {
-		crossClusterTask, ok := queueTask.(task.CrossClusterTask)
+	batchSize := 100
```
yycptt, line 352: This will be merged with https://github.com/uber/cadence/pull/4487 after it's landed and batch size will be controlled by a dynamic config

common/collection/ordered_map.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	m.list.Remove(mapValue.listElement)
+	delete(m.items, key)
+}
+
+func (m *orderedMap) Iter() MapIterator {
+	iterator := newMapIterator()
+
+	go orderedMapIteratorLoop(nil, iterator, m.list)
```
yux0, line 114: could it iterate the data channel while the iterator is not fully initialized

yycptt: I don't think so? Channels for the iterator is initialized in `newMapIterator` which happens before this go routine is created and before the method returns.

yux0 (approved):

PR #4494, opened 2021-09-17
Update persistence test to verify delete history branch actually delete related nodes

Shaddoll:
**What changed?**
Update persistence tests to verify that deleting history branch operation actually deletes targeted nodes from history node table

**Why?**
Improve test coverage

**How did you test it?**
persistence tests

**Potential risks**

**Release notes**

**Documentation Changes**


longquanzheng:
Is it only updating the test? 
If so can we update the title to "Update persistence test to verify ..." so that the release will know this is not functional changes. Otherwise LGTM

Shaddoll:
> Is it only updating the test?
> If so can we update the title to "Update persistence test to verify ..." so that the release will know this is not functional changes. Otherwise LGTM
Yes, it only update the test

longquanzheng (approved):

PR #4498, opened 2021-09-17
Refactor to add a SQL driver layer for multiple SQL databases support as sharded SQL

longquanzheng:
**What changed?**
Refactor to add a SQL driver layer for multiple SQL databases support as sharded SQL

**Why?**
To support using multiple SQL as sharded MySQL. 
Idea from a Cadence OSS user and some suggestion from @meiliang86 
This allows user to have a Cadence cluster with a much larger scale. (then using XDC to add more DB instances)

This is refactor only. Next PR will implement the driver interface for multiple sql 

**How did you test it?**
Existing tests. 

**Potential risks**
Low risk. This is refactoring only. No behavior change. 
It added a parameter to a method in sqlplugin interfaces -- the internal implementation for docStore should just ignore it.

**Release notes**

**Documentation Changes**


common/persistence/sql/sqlHistoryStore.go
```diff
@@ -117,7 +117,8 @@ func (m *sqlHistoryStore) AppendHistoryNodes(
 			DataEncoding: string(blob.Encoding),
 		}
 
-		return m.txExecute(ctx, "AppendHistoryNodes", func(tx sqlplugin.Tx) error {
+		dbShardID := sqlplugin.GetDBShardIDFromHistoryShardID(request.ShardID, m.db.GetTotalNumDBShards())
```
Shaddoll, line 120: This will cause large partition issue. We originally used `shard_id` as the partition key of sharded SQL, but the latency of read history operation kept growing as the bench test ran.
When you do continue as new, all workflows with the same workflow id will be stored in the same shard. So instead, we use `shard_id, tree_id` as the partition key

common/persistence/sql/sqlTaskStore.go
```diff
@@ -70,7 +68,8 @@ func (m *sqlTaskStore) LeaseTaskList(
 ) (*persistence.LeaseTaskListResponse, error) {
 	var rangeID int64
 	var ackLevel int64
-	shardID := m.shardID(request.DomainID, request.TaskList)
+	shardID := sqlplugin.GetDBShardIDFromDomainIDAndTasklist(request.DomainID, request.TaskList, m.db.GetTotalNumDBShards())
```
Shaddoll, line 71: It breaks our logic in Docstore. the shard id is currently ignore by mysql and postgres, but it's used in our internal docstore db.

Shaddoll: Just create a new variable dbShardID like in sqlExecutionStore.go and keep the old shardID the same  as before

common/persistence/sql/sqldriver/interface.go
```diff
@@ -0,0 +1,56 @@
...
+	Driver interface {
+
+		// shared methods of both non-transactional sqlx.DB and transactional sqlx.Tx
+		commonOfDbAndTx
+
+		// From sqlx.DB
+		Exec(dbShardID int, query string, args ...interface{}) (sql.Result, error)
+		Select(dbShardID int, dest interface{}, query string, args ...interface{}) error
+		Get(dbShardID int, dest interface{}, query string, args ...interface{}) error
+		BeginTxx(ctx context.Context, dbShardID int, opts *sql.TxOptions) (*sqlx.Tx, error)
```
Shaddoll, line 42: Should it be `BeginTx`?

common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -191,7 +191,7 @@ type (
 
 	// TasksRow represents a row in tasks table
 	TasksRow struct {
-		ShardID      int
+		ShardID      int // this is DBShardID, not historyShardID (TODO: maybe rename it for clarification)
```
Shaddoll, line 194: It's history shard ID

common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -220,7 +220,7 @@ type (
 	// TasksFilter contains the column names within tasks table that
 	// can be used to filter results through a WHERE clause
 	TasksFilter struct {
-		ShardID              int
+		ShardID              int // this is DBShardID, not historyShardID (TODO: maybe rename it for clarification)
```
Shaddoll, line 223: It's history ShardID.

common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -239,7 +239,7 @@ type (
 
 	// TaskListsRow represents a row in task_lists table
 	TaskListsRow struct {
-		ShardID      int
+		ShardID      int // this is DBShardID, not historyShardID (TODO: maybe rename it for clarification)
```
Shaddoll, line 242: HistoryShardID

common/persistence/sql/sqlplugin/interfaces.go
```diff
@@ -257,7 +257,7 @@ type (
 	// TaskListsFilter contains the column names within task_lists table that
 	// can be used to filter results through a WHERE clause
 	TaskListsFilter struct {
-		ShardID             int
+		ShardID             int // this is DBShardID, not historyShardID (TODO: maybe rename it for clarification)
```
Shaddoll, line 260: HistoryShardID

common/persistence/sql/sqlplugin/interfaces.go
longquanzheng: https://github.com/uber/cadence/blob/master/common/persistence/sql/sqlTaskStore.go#L550

Can you confirm? I think this is different from execution store

common/persistence/sql/sqlHistoryStore.go
longquanzheng: Yeah good point. Will change it.

common/persistence/sql/sqlTaskStore.go
longquanzheng: Yeah that will work too but that will be confusing. 
This shardID is calculated by 
`id := farm.Hash32([]byte(domainID+"_"+name)) % uint32(m.nShards)`
which is exactly the same thing that we are trying to get as dbShardID. 

It's not historyShardID.

common/persistence/sql/sqlplugin/mysql/visibility.go
```diff
@@ -80,7 +80,9 @@ var errCloseParams = errors.New("missing one of {closeStatus, closeTime, history
 // its left as such and no update will be made
 func (mdb *db) InsertIntoVisibility(ctx context.Context, row *sqlplugin.VisibilityRow) (sql.Result, error) {
 	row.StartTime = mdb.converter.ToMySQLDateTime(row.StartTime)
-	return mdb.conn.ExecContext(ctx,
+	dbShardID := sqlplugin.GetDBShardIDFromDomainID(row.DomainID, mdb.GetTotalNumDBShards())
```
Shaddoll, line 83: Shard by domain ID is not a good implementation, because it will cause large partition issue

longquanzheng: I don't think we have better option based on the schema and our query pattern: https://github.com/uber/cadence/blob/master/schema/mysql/v57/visibility/schema.sql

It's okay for here. A Cadence cluster with large scale should use advanced visibility anyway. We just need to do our best here based on the current architecture

common/persistence/sql/sqlHistoryStore.go
longquanzheng: Changed to GetDBShardIDFromTreeID :D thanks for pointing it out.

common/persistence/sql/sqlTaskStore.go
longquanzheng: > It breaks our logic in Docstore
I just moved the logic to `dbSharding.go` and the behavior the exactly the same. It shouldn't break anything.

Shaddoll: Ah, you're right. I just noticed that in our internal config, we set the nShards to be the same as numHistoryShards.

common/persistence/sql/sqlplugin/mysql/visibility.go
Shaddoll: what about `run_id` or `domain_id, run_id`?

longquanzheng: It will not work. Because visibilityStore needs to query by workflowID or closeStatus etc. 
The indexes are also built with that pattern: https://github.com/uber/cadence/blob/master/schema/mysql/v57/visibility/schema.sql#L19

If the partition is based on runID or `domainID+runID`(no difference because runID is already random), a single query/index will need to be sent to all shards and then merge back.

Also, Cassandra uses domainID as partitionKey as well. (domainPatition is calculated by domainID). https://github.com/uber/cadence/blob/master/schema/cassandra/visibility/schema.cql

common/persistence/sql/sqlHistoryStore.go
Shaddoll: I think you didn't change all of them?

Shaddoll (approved):

common/persistence/sql/common.go
```diff
@@ -53,8 +53,8 @@ func (m *sqlStore) Close() {
 	}
 }
 
-func (m *sqlStore) txExecute(ctx context.Context, operation string, f func(tx sqlplugin.Tx) error) error {
-	tx, err := m.db.BeginTx(ctx)
+func (m *sqlStore) txExecute(dbShardID int, ctx context.Context, operation string, f func(tx sqlplugin.Tx) error) error {
```
antstorm, line 56: nit: As far as I know idiomatic Golang suggests passing context as the first argument (if present)

antstorm (commented):
Looks great!

common/persistence/sql/common.go
longquanzheng: Yeah not sure how I missed this one. Good catch. Thanks!

PR #4514, opened 2021-09-27
Reshuffle service.Config and resource.Params

vytautas-karpavicius:
**What changed and why?**
This PR reshuffles few structures between packages.

1. `service.BootstrapParams` was moved to `resource.Params`
`resource` package already contains big container with many dependencies used between all Cadence services. These parameters were even used to initialize it. Parameters also contain many dependencies.

2. `resource/config.ResourceConfig` was moved to `service.Config`
It various dynamic config settings shared between all services. They do not contains any dependencies.

3. Persistence layer used `service.BootstrapParams` and `resource/config.ResourceConfig` directly. Now it will use `service.Config` and additionally introduces `persistence/client.Params` structure where all dependencies required for persistence are passed.

**Justification**
Current package layout was suboptimal because with certain changes it easily results in circular dependencies. This was already a case for `resource/config.ResourceConfig` which was moved to separate package for this reason.

The goal-state of this change is as follows:
- `resource` package contains service related things, with dependencies. This is the package that uses other things, but itself is not used by others.
- `service` package contains service related things, without other _significant_ dependencies. This is the package that others can use.
Thus: `service` <- used by other packages <- used by `resource`

Additionally there was a case where persistence layer used whole `service.BootstrapParams` struct. After reshuffle this violated above goal-state, therefore it will accept its own `persistence/client.Params` structure with explicit dependencies it uses. 

**How did you test it?**
Existing tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/server/cadence/server.go
+++ b/cmd/server/cadence/server.go
@@ -40,6 +40,7 @@ import (
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/messaging/kafka"
 	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/resource"
 	"github.com/uber/cadence/common/rpc"
 	"github.com/uber/cadence/common/service"
 	"github.com/uber/cadence/service/frontend"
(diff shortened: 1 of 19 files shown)


mantas-sidlauskas (approved):
lgtm

PR #4515, opened 2021-09-27
Fix record child completion error handling

yycptt:
**What changed?**
- Fix record child completion error handling

**Why?**
- Fix record child completion error handling

**How did you test it?**
- Unit test

**Potential risks**
- N/A

**Release notes**

**Documentation Changes**


service/history/task/transfer_active_task_executor.go
```diff
@@ -341,12 +343,13 @@ func (t *transferActiveTaskExecutor) processCloseExecution(
 	var crossClusterTaskGenerators []generatorF
 	// Communicate the result to parent execution if this is Child Workflow execution
 	if replyToParentWorkflow {
-		targetDomainEntry, err := t.shard.GetDomainCache().GetDomainByID(parentDomainID)
```
demirkayaender, line 344: how was this shadowing the errors? if we get error here we return the error, if the error was set before we would return it before getting into this block.

yycptt: The error from `RecordChildExecutionCompleted` on L357 is assigned to the error variable created here inside the if statement. But the error checking logic is on L379 which is outside the if statement.

demirkayaender (approved):

mkolodezny (approved):

yux0 (approved):

PR #4517, opened 2021-09-28
Introduce rpc.Params

vytautas-karpavicius:
**What changed?**
Introduced `rpc.Params` structure that contains required parameters to initialize `rpc.Factory`.
This decouples Factory from using config directly. It is beneficial for 2 reasons:
- Allows unit test config part easier
- Allows reusing the same `rpc.Factory` in integration tests. Before we had a separate implementation there, with minor differences. Now we can pass those differences via Params.

Also refactored service config retrieval into a function. This is now used in couple of places.

**Why?**
To reduce code duplication and increase test coverage.

**How did you test it?**
Additional unit tests for refactored code.
Existing tests.
Starting server locally with removed config section.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/server/cadence/cadence.go
+++ b/cmd/server/cadence/cadence.go
@@ -82,9 +82,6 @@ func startHandler(c *cli.Context) {
 	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, syscall.SIGTERM)
 	for _, svc := range services {
-		if _, ok := cfg.Services[svc]; !ok {
-			log.Fatalf("`%v` service missing config", svc)
-		}
 		server := newServer(svc, &cfg)
 		daemons = append(daemons, server)
 		server.Start()
--- a/cmd/server/cadence/server.go
+++ b/cmd/server/cadence/server.go
@@ -70,9 +70,6 @@ func newServer(service string, cfg *config.Config) common.Daemon {
 
 // Start starts the server
 func (s *server) Start() {
-	if _, ok := s.cfg.Services[s.name]; !ok {
-		log.Fatalf("`%v` service missing config", s)
-	}
 	s.daemon = s.startService()
 }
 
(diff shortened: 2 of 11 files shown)


cmd/server/cadence/cadence.go
```diff
@@ -82,9 +82,6 @@ func startHandler(c *cli.Context) {
 	sigc := make(chan os.Signal, 1)
 	signal.Notify(sigc, syscall.SIGTERM)
 	for _, svc := range services {
-		if _, ok := cfg.Services[svc]; !ok {
```
vytautas-karpavicius, line 85: No need to check here. This is checked down the line in `server.startService` when config is actually needed.

cmd/server/cadence/server.go
```diff
@@ -70,9 +70,6 @@ func newServer(service string, cfg *config.Config) common.Daemon {
 
 // Start starts the server
 func (s *server) Start() {
-	if _, ok := s.cfg.Services[s.name]; !ok {
```
vytautas-karpavicius, line 73: No need to check here. This is checked down the line in `server.startService` when config is actually needed.

mantas-sidlauskas (approved):
left one minor comment

common/rpc/grpc.go
```diff
@@ -29,7 +29,13 @@ import (
 	"github.com/uber/cadence/common/service"
 )
 
-type GRPCPorts map[string]int
+type (
+	GRPCPortResolver interface {
+		GetGRPCAddress(service, hostAddress string) (string, error)
+	}
```
mantas-sidlauskas, line 35: Is this interface really needed?

If so, I would make it more general, as it does nothing specific to `grpc`

vytautas-karpavicius: The interface is needed, as integration tests pass different implementation.
But I agree that the name could be more general. Updated it.

PR #4518, opened 2021-09-28
Increase archival initial retry interval

demirkayaender:
**What changed?**
Increased the initial retry interval for history archival request. 

**Why?**
Current retries were exhausting the storage we use for archival.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


yycptt (commented):
I think we should fix the retry policy in our internal repo for uploading blob, where there's basically no backoff? The backoff here looks kinda ok to me.

longquanzheng:
maybe we should expose as configuration instead of hardcoded.

demirkayaender:
Closing this one, will handle this with a config or internally.

PR #4521, opened 2021-09-30
Add size checker when replication messages return

yux0:
**What changed?**
Add size checker when replication messages return

**Why?**
If the size is above the grpc config, drop some messages.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


client/clientfactory.go
```diff
@@ -156,8 +157,13 @@ func (cf *rpcClientFactory) NewHistoryClientWithTimeout(timeout time.Duration) (
 		return cf.newHistoryThriftClient(clientKey)
 	}
 
+	defaultMessageSize := cf.rpcFactory.GetGRPCMaxSizeInBytes()
+	if cf.enableGRPCOutbound && defaultMessageSize == 0 {
+		defaultMessageSize = common.DefaultGRPCSizeLimit
```
yycptt, line 162: I guess we don't have to make this specific to grpc?

client/history/client.go
```diff
@@ -854,16 +859,22 @@ func (c *clientImpl) GetReplicationMessages(
...
+	if len(errChan) > 0 {
+		err = <-errChan
+	}
+
 	response := &types.GetReplicationMessagesResponse{MessagesByShard: make(map[int32]*types.ReplicationMessages)}
+	responseSize := int64(0)
 	for resp := range respChan {
 		for shardID, tasks := range resp.MessagesByShard {
+			responseSize += tasks.GetSizeInByte()
+			if responseSize > int64(c.rpcMaxSizeInBytes()) {
```
yycptt, line 872: Is `responseSize` here for a batch of tasks? If so, if batch size is too high, we still need to manually adjust the number otherwise it will still get blocked?

service/history/replication/task_ack_manager.go
```diff
@@ -196,6 +200,7 @@ TaskInfoLoop:
 		}
 		readLevel = taskInfo.GetTaskID()
 		if replicationTask != nil {
+			taskSize += replicationTask.GetSizeInByte()
```
yycptt, line 203: Will we have the same size limit issue when returning the aggregated response from history to frontend?

client/clientfactory.go
vytautas-karpavicius: +1
I also think the entire logic with defaults and retrieval from dynamic could be within rpcFactory iself. Here we could just use the final resolved size `rpcFactory.GetMaxMessageSize()`

service/history/replication/task_ack_manager.go
```diff
@@ -233,6 +238,7 @@ TaskInfoLoop:
 		ReplicationTasks:       replicationTasks,
 		HasMore:                hasMore,
 		LastRetrievedMessageID: readLevel,
+		Size:                   taskSize,
```
vytautas-karpavicius, line 241: I believe the whole size calculation could be done more universally within a middleware. It already knows serialized content in bytes. Then is become the question how to use it. One way I can think - is to set a value in the passed context. I will try to come up with some PR for that.

vytautas-karpavicius: Please check whether this could be suitable: https://github.com/uber/cadence/pull/4522

It is not exactly the same - it contains serialized message size, but after gathering all shards it would get serialized again, so it is more or less the same. I guess we could also reserve some additional buffer, for the margin of error. Lets say, stop accumulating tasks once we pass 90% of max message size.

client/clientfactory.go
yux0: Do you mean we should also limit the size with yarpc?

client/history/client.go
yux0: we can adjust the dynamic config rpcMaxSizeInBytes

service/history/replication/task_ack_manager.go
yux0: Thank you. Yes. This makes more sense

yux0: I will see how can I get the response info from the context

client/clientfactory.go
vytautas-karpavicius: I meant that this `if` statement together with `common.DefaultGRPCSizeLimit` could go inside `rpc` package. It is not strictly required, just some code organization.

service/history/replication/task_ack_manager.go
yux0: The aggregated response operation happens in history. If the size exceeds, the response will be dropped

service/history/replication/task_ack_manager.go
```diff
@@ -50,10 +50,13 @@ import (
...
-	errUnknownReplicationTask = errors.New("unknown replication task")
-	defaultHistoryPageSize    = 1000
-	minReadTaskSize           = 20
+	errUnknownQueueTask             = errors.New("unknown task type")
+	errUnknownReplicationTask       = errors.New("unknown replication task")
+	defaultHistoryPageSize          = 1000
+	minReadTaskSize                 = 20
+	defaultSyncActivitySizeInByte   = 480
+	defaultFailoverMarkerSizeInByte = 150
+	defaultHistorySizeInByte        = 300
```
yycptt, line 59: We can remove those definitions?

client/history/client.go
```diff
@@ -846,24 +856,37 @@ func (c *clientImpl) GetReplicationMessages(
...
+		}(ctx, client, req)
 	}
 
 	wg.Wait()
 	close(respChan)
 	close(errChan)
 
+	var err error
+	if len(errChan) > 0 {
+		err = <-errChan
```
yycptt, line 873: we can return early here, and skip the logic for preparing the result for the rest of the shards?

client/clientfactory.go
```diff
@@ -156,8 +156,10 @@ func (cf *rpcClientFactory) NewHistoryClientWithTimeout(timeout time.Duration) (
 		return cf.newHistoryThriftClient(clientKey)
 	}
 
+	defaultMessageSize := cf.rpcFactory.GetMaxMessageSize()
 	client := history.NewClient(
 		cf.numberOfHistoryShards,
+		cf.dynConfig.GetIntProperty(dynamicconfig.GRPCMaxSizeInByte, defaultMessageSize),
```
yycptt, line 162: nit: Does it mean the dynamic value we set should always be less than the `MaxMessageSize` from rpc factory? If so, shall we do a check to ensure that?

yycptt (approved):
LGTM. Some minor comments.

client/history/client.go
```diff
@@ -846,24 +856,37 @@ func (c *clientImpl) GetReplicationMessages(
 				}
 				return
 			}
-			respChan <- resp
-		}(requestContext, client, req)
+			_, responseInfo := rpc.ContextWithResponseInfo(requestContext)
```
vytautas-karpavicius, line 859: I'm afraid this will not work. You need to use context returned from `ContextWithResponseInfo` when making an rpc call, otherwise the responseInfo structure will remain empty.

yux0: Just realize this use as a callback. Update d the code.

PR #4522, opened 2021-09-30
Added response info middleware

vytautas-karpavicius:
**What changed?**
Added outbound middleware for filling `rpc.ResponseInfo` structure.

**Why?**
This can be useful for a general way to obtain information about the response before it gets deserialized and returned to the client. One such example - obtaining serialized size of the message.

Intended usage example:
```
ctx, responseInfo := rpc.ContextWithResponseInfo(ctx)
resp, err := client.GetReplicationMessages(ctx, request, opts...)
messageSize := responseInfo.Size
...
```

**How did you test it?**
Added unit test, verified locally with print statement that message size can be obtained that way.

**Potential risks**

**Release notes**

**Documentation Changes**


yux0 (approved):

common/rpc/middleware.go
```diff
@@ -47,3 +47,32 @@ func (m *authOutboundMiddleware) Call(ctx context.Context, request *transport.Re
...
+type ResponseInfo struct {
+	Size int
+}
+
+type responseInfoMiddleware struct{}
+
+func (m *responseInfoMiddleware) Call(ctx context.Context, request *transport.Request, out transport.UnaryOutbound) (*transport.Response, error) {
+	response, err := out.Call(ctx, request)
+
+	if value := ctx.Value(_responseInfoContextKey); value != nil {
```
yux0, line 71: could we make this a public function so that we can get the response size in client?

PR #4523, opened 2021-09-30
Use common dispatcher for public client outbound

vytautas-karpavicius:
**What changed?**
Stop using `DispatcherProvider` for creating public client. Instead use common dispatcher from rpc.Factory.

Builder for public client outbound was created which also includes auth middleware initialization. It was previously located within server.go itself. Moving it here allows consolidating all logic related logic in one place and covering it with unit tests.

**Why?**
To eventually delete `DispatcherProvider` entirely.

**How did you test it?**
Existing tests and new unit tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/rpc/outbounds.go
+++ b/common/rpc/outbounds.go
@@ -21,12 +21,57 @@
 package rpc
 
 import (
+	"fmt"
+
+	"github.com/uber/cadence/common/authorization"
+	"github.com/uber/cadence/common/config"
+	"github.com/uber/cadence/common/service"
+
 	"go.uber.org/yarpc"
+	"go.uber.org/yarpc/api/middleware"
 	"go.uber.org/yarpc/transport/grpc"
 	"go.uber.org/yarpc/transport/tchannel"
+)
+
+const (
+	// OutboundPublicClient is the name of configured public client outbound
+	OutboundPublicClient = "public-client"
 )
 
 // OutboundsBuilder allows defining outbounds for the dispatcher
 type OutboundsBuilder interface {
 	Build(*grpc.Transport, *tchannel.Transport) (yarpc.Outbounds, error)
+}
+
+func newPublicClientOutbound(config *config.Config) (publicClientOutbound, error) {
+	if len(config.PublicClient.HostPort) == 0 {
+		return publicClientOutbound{}, fmt.Errorf("need to provide an endpoint config for PublicClient")
+	}
+
+	var authMiddleware *authOutboundMiddleware
+	if config.Authorization.OAuthAuthorizer.Enable {
+		clusterName := config.ClusterGroupMetadata.CurrentClusterName
+		clusterInfo := config.ClusterGroupMetadata.ClusterGroup[clusterName]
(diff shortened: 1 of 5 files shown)


common/rpc/params_test.go
```diff
@@ -45,6 +47,9 @@ func TestNewParams(t *testing.T) {
 	_, err = NewParams(serviceName, makeConfig(config.Service{RPC: config.RPC{BindOnIP: "invalidIP"}}))
 	assert.EqualError(t, err, "failed to get listen IP: unable to parse bindOnIP value or it is not an IPv4 address: invalidIP")
 
+	_, err = NewParams(serviceName, &config.Config{Services: map[string]config.Service{"frontend": {}}})
+	assert.EqualError(t, err, "failed to create public client outbound: need to provide an endpoint config for PublicClient")
```
mantas-sidlauskas, line 51: nit: error messages can be shorter. Ref: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

common/rpc/outbounds.go
```diff
@@ -21,12 +21,57 @@
...
+
+func newPublicClientOutbound(config *config.Config) (publicClientOutbound, error) {
+	if len(config.PublicClient.HostPort) == 0 {
+		return publicClientOutbound{}, fmt.Errorf("need to provide an endpoint config for PublicClient")
+	}
+
+	var authMiddleware *authOutboundMiddleware
+	if config.Authorization.OAuthAuthorizer.Enable {
+		clusterName := config.ClusterGroupMetadata.CurrentClusterName
+		clusterInfo := config.ClusterGroupMetadata.ClusterGroup[clusterName]
```
mantas-sidlauskas, line 54: do we need to check if `clusterName` is in map?

common/rpc/outbounds.go
```diff
@@ -21,12 +21,57 @@
...
 	Build(*grpc.Transport, *tchannel.Transport) (yarpc.Outbounds, error)
 }
+
+func newPublicClientOutbound(config *config.Config) (publicClientOutbound, error) {
+	if len(config.PublicClient.HostPort) == 0 {
+		return publicClientOutbound{}, fmt.Errorf("need to provide an endpoint config for PublicClient")
+	}
+
+	var authMiddleware *authOutboundMiddleware
+	if config.Authorization.OAuthAuthorizer.Enable {
```
mantas-sidlauskas, line 52: nit/optional: reduce nesting with early return when `Enabled == false`

mantas-sidlauskas (approved):
lgtm, All suggestions are optional

common/rpc/outbounds.go
vytautas-karpavicius: This is validated here: https://github.com/uber/cadence/blob/master/common/config/cluster.go#L111
So we should be ok.

common/rpc/params_test.go
vytautas-karpavicius: Agree, will update.

common/rpc/outbounds.go
vytautas-karpavicius: For this particular case, I would like to keep it. As this is adding optional feature rather than error'ing out early.

PR #4527, opened 2021-10-01
Add metrics for cross cluster implementation

yycptt:
**What changed?**
- Add metrics for cross cluster components
- Add missing parts in cross-cluster components
- Handle shard level error when getting cross cluster tasks

**Why?**
- Cross-domain operations support

**How did you test it?**
- Existing test since it's mostly adding metrics

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/cmd/server/tools.go
+++ b/cmd/server/tools.go
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+//go:build tools
 // +build tools
 
 package tools
--- a/common/util.go
+++ b/common/util.go
@@ -982,3 +982,17 @@ func ConvertErrToGetTaskFailedCause(err error) types.GetTaskFailedCause {
 	}
 	return types.GetTaskFailedCauseUncategorized
 }
+
+// ConvertGetTaskFailedCauseToErr converts GetTaskFailedCause to error
+func ConvertGetTaskFailedCauseToErr(failedCause types.GetTaskFailedCause) error {
+	switch failedCause {
+	case types.GetTaskFailedCauseServiceBusy:
+		return &types.ServiceBusyError{}
+	case types.GetTaskFailedCauseTimeout:
+		return context.DeadlineExceeded
+	case types.GetTaskFailedCauseShardOwnershipLost:
+		return &types.ShardOwnershipLostError{}
+	default:
+		return &types.InternalServiceError{Message: "uncategorized error"}
+	}
+}
--- a/service/history/queue/cross_cluster_queue_processor_base.go
+++ b/service/history/queue/cross_cluster_queue_processor_base.go
@@ -276,7 +276,9 @@ processorPumpLoop:
 				continue processorPumpLoop
 			}
 
-			if c.readyForPollTasks.Len() > c.options.MaxPendingTaskSize() {
+			numReadyForPoll := c.readyForPollTasks.Len()
+			c.metricsScope.RecordTimer(metrics.CrossClusterTaskPendingTimer, time.Duration(numReadyForPoll))
+			if numReadyForPoll > c.options.MaxPendingTaskSize() {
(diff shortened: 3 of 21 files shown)


cmd/server/tools.go
```diff
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+//go:build tools
```
demirkayaender, line 21: nit: is this generated?

common/util.go
```diff
@@ -982,3 +982,17 @@ func ConvertErrToGetTaskFailedCause(err error) types.GetTaskFailedCause {
...
+func ConvertGetTaskFailedCauseToErr(failedCause types.GetTaskFailedCause) error {
+	switch failedCause {
+	case types.GetTaskFailedCauseServiceBusy:
+		return &types.ServiceBusyError{}
+	case types.GetTaskFailedCauseTimeout:
+		return context.DeadlineExceeded
+	case types.GetTaskFailedCauseShardOwnershipLost:
+		return &types.ShardOwnershipLostError{}
+	default:
+		return &types.InternalServiceError{Message: "uncategorized error"}
```
demirkayaender, line 996: Do you have the message from the failedCause so we can emit here?

cmd/server/tools.go
yycptt: Yes auto-generated by editor when running `make bins`

service/history/queue/cross_cluster_queue_processor_base.go
```diff
@@ -276,7 +276,9 @@ processorPumpLoop:
 				continue processorPumpLoop
 			}
 
-			if c.readyForPollTasks.Len() > c.options.MaxPendingTaskSize() {
+			numReadyForPoll := c.readyForPollTasks.Len()
+			c.metricsScope.RecordTimer(metrics.CrossClusterTaskPendingTimer, time.Duration(numReadyForPoll))
+			if numReadyForPoll > c.options.MaxPendingTaskSize() {
 				c.logger.Warn("too many outstanding ready for poll tasks in cross cluster queue.")
```
demirkayaender, line 282: Should we make this an error log?

service/history/shard/context.go
```diff
@@ -180,11 +180,12 @@ var (
...
 const (
-	conditionalRetryCount    = 5
-	logWarnTransferLevelDiff = 3000000 // 3 million
-	logWarnTimerLevelDiff    = time.Duration(30 * time.Minute)
-	historySizeLogThreshold  = 10 * 1024 * 1024
-	minContextTimeout        = 1 * time.Second
+	conditionalRetryCount       = 5
+	logWarnTransferLevelDiff    = 3000000 // 3 million
+	logWarnTimerLevelDiff       = time.Duration(30 * time.Minute)
+	logWarnCrossClusterLevelLag = 3000000
```
demirkayaender, line 186: is this a duration? if so, do you mind commenting the unit of this?

common/util.go
yycptt: The failed cause is an enum so we don't have the message here (the target cluster). The detailed error message will be logged at the source cluster during fetch. The target side only needs to know it's an unexpected error and retry.

service/history/queue/cross_cluster_queue_processor_base.go
yycptt: hmmm, I don't have a strong opinion but probably not? This is expected if the target cluster is under high load and slow down the polling.

service/history/shard/context.go
yycptt: It's not duration, but an absolute number based on for taskID which is calculated based on rangeID. So it's a very rough estimate, during shard movement rangeID will increase by 1 and taskID will be increased by around 1M. I basically used the same number as `logWarnTransferLevelDiff`.

demirkayaender: Can we use a more descriptive name like minTaskIDToLogWarn or add a comment here to explain what this is?

yycptt: Added a comment for transfer and cross-cluster level

PR #4531, opened 2021-10-03
Add Redirect policy to forward all domain APIs

longquanzheng:
**What changed?**
Add Redirect policy to forward all domain APIs

**Why?**
In cluster migration case, forwarding all domain APIs is preferred because this will simplify the worker setup(no need to start worker set in each cluster).

**How did you test it?**
Unit tests

**Potential risks**
Low.

**Release notes**
Added comment to godocs. 
Also working on documentation along with https://github.com/uber/cadence-docs/pull/62 

**Documentation Changes**
https://github.com/uber/cadence-docs/pull/62


service/frontend/dcRedirectionPolicy.go
```diff
@@ -38,15 +35,18 @@ const (
...
 	// 3. SignalWorkflowExecution
 	// 4. RequestCancelWorkflowExecution
 	// 5. TerminateWorkflowExecution
 	// 6. QueryWorkflow
-	// please also reference selectedAPIsForwardingRedirectionPolicyWhitelistedAPIs
+	// 7. ResetWorkflow
+	// please also reference selectedAPIsForwardingRedirectionPolicyAPIAllowlist
 	DCRedirectionPolicySelectedAPIsForwarding = "selected-apis-forwarding"
+	// DCRedirectionPolicyAllDomainAPIsForwarding means forwarding all the worker and non-worker APIs based domain
+	DCRedirectionPolicyAllDomainAPIsForwarding = "all-domain-apis-forwarding"
```
yux0, line 49: what about all-apis-forwarding? all-domains seems only domain related apis?

service/frontend/dcRedirectionPolicy.go
```diff
@@ -172,14 +178,16 @@ func (policy *SelectedAPIsForwardingRedirectionPolicy) getTargetClusterAndIsDoma
...
-	enforceDCRedirection := call.Header(common.EnforceDCRedirection)
-	if !policy.config.EnableDomainNotActiveAutoForwarding(domainEntry.GetInfo().Name) && enforceDCRedirection != "true" {
-		// do not do dc redirection if auto-forwarding dynamic config and EnforceDCRedirection context flag is not enabled
+	if !policy.config.EnableDomainNotActiveAutoForwarding(domainEntry.GetInfo().Name) {
+		// do not do dc redirection if auto-forwarding dynamicconfig is not enabled
 		return policy.currentClusterName, false
 	}
 
-	_, ok := selectedAPIsForwardingRedirectionPolicyWhitelistedAPIs[apiName]
+	if policy.allDomainAPIs {
```
yux0, line 186: Is it possible to have a per domain config?

service/frontend/dcRedirectionPolicy.go
longquanzheng: yeah we don't(probably shouldn't) forward non-domain related api. 
For example, describe shard, add search attributes, etc

service/frontend/dcRedirectionPolicy.go
longquanzheng: that's a good idea.

PR #4535, opened 2021-10-04
Add server startup test to prevent docker image master-auto-setup getting broken

longquanzheng:
**What changed?**

**Why?**
This has been an issue for several times. Currently there is no test to protect the startup code. 
This will prevent https://github.com/uber/cadence/pull/4534 from happening again. 

**How did you test it?**
Test added 
Screen Shot 2021-10-04 at 4 37 34 PM

Screen Shot 2021-10-04 at 4 38 35 PM

**Potential risks**
No

**Release notes**

**Documentation Changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -463,11 +463,11 @@ GOCOVERPKG_ARG := -coverpkg="$(subst $(SPACE),$(COMMA),$(addprefix $(PROJECT_ROO
 test: bins ## Build and run all tests
 	@rm -f test
 	@rm -f test.log
+	# special test case without race detector
+    go test -v ./cmd/server/cadence/
 	@for dir in $(PKG_TEST_DIRS); do \
 		go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log; \
 	done;
-	# special test case without race detector
-	go test -v ./cmd/server/cadence/
 
 test_e2e: bins
 	@rm -f test
--- a/Makefile
+++ b/Makefile
@@ -466,6 +466,8 @@ test: bins ## Build and run all tests
 	@for dir in $(PKG_TEST_DIRS); do \
 		go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log; \
 	done;
+	# special test case without race detector
+	go test -v ./cmd/server/cadence/
 
 test_e2e: bins
 	@rm -f test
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -531,12 +531,12 @@ cover_ci: $(COVER_ROOT)/cover.out $(BIN)/goveralls
 	$(BIN)/goveralls -coverprofile=$(COVER_ROOT)/cover.out -service=buildkite || echo Coveralls failed;
 
 install-schema: cadence-cassandra-tool
-	./cadence-cassandra-tool --ep 127.0.0.1 create -k cadence --rf 1
-	./cadence-cassandra-tool --ep 127.0.0.1 -k cadence setup-schema -v 0.0
-	./cadence-cassandra-tool --ep 127.0.0.1 -k cadence update-schema -d ./schema/cassandra/cadence/versioned
-	./cadence-cassandra-tool --ep 127.0.0.1 create -k cadence_visibility --rf 1
-	./cadence-cassandra-tool --ep 127.0.0.1 -k cadence_visibility setup-schema -v 0.0
-	./cadence-cassandra-tool --ep 127.0.0.1 -k cadence_visibility update-schema -d ./schema/cassandra/visibility/versioned
```
longquanzheng, line 539: to reviewer: this is compatible because the default value is already `127.0.0.1` see https://github.com/uber/cadence/blob/d19cae1128b43da87fe437c582bce07ffd237f9e/tools/cassandra/main.go#L68

Removing this here so that I can use env variable to override in buildkite test

cmd/server/cadence/server.go
```diff
@@ -58,9 +58,9 @@ type (
 	}
 )
 
-// newServer returns a new instance of a daemon
+// NewServer returns a new instance of a daemon
 // that represents a cadence service
-func newServer(service string, cfg *config.Config) common.Daemon {
+func NewServer(service string, cfg *config.Config) common.Daemon {
```
vytautas-karpavicius, line 63: Any reason why this became public? I do not see this being used outside of this package.

cmd/server/cadence/server_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	var cfg config.Config
+	err := config.Load(env, configDir, zone, &cfg)
+	if err != nil {
+		log.Fatal("Config file corrupted.", err)
+	}
+	// replace local host to docker network
+	if runInBuildKite {
+		ds := cfg.Persistence.DataStores[cfg.Persistence.DefaultStore]
+		ds.NoSQL.Hosts = "cassandra"
```
vytautas-karpavicius, line 76: I wonder if this should be taken from env variable? I see `CASSANDRA_HOST` being set buildkite pipeline.

vytautas-karpavicius (approved):
Overall look good, just a couple of clarifications.

cmd/server/cadence/server.go
longquanzheng: Yeah I will change it back. Initially I was writing `main_test.go` but then I moved it to  `server_test.go`

PR #4536, opened 2021-10-04
Support isGlobal in Workflow Metrics

demirkayaender:
**What changed?**
Added isGlobal dimension to worklow metrics related to following operations:
- StartWorkflowExecution
- SignalWorkflowExecution
- SignalWithStartWorkflowExecution
- TerminateWorkflowExecution
- ResetWorkflowExecution
- RequestCancelWorkflowExecution

**Why?**
Without this way, we don't have a way to count total number of workflows in an environment with multiple regions. We need to be able to differentiate between local vs global domain so we can get the total number of workflows with 
all_workflows(region_1) + local_workflows(region_2) + ... + local_workflows(region_n)

**How did you test it?**
workflowHandlerTest unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx:
How is this intended to be used?  Since global domains are only active in one cluster at a time, it doesn't seem like this would do much for differentiating many-local vs single-global; one has metrics everywhere, one has metrics only in one at a time.

Except for forwarded calls.  But will those get this far into code?  I think they're forwarded earlier in the handler-wrappers.

demirkayaender:
If I remember right, global domain counters were counted twice; certainly for ElasticSearch (visibility records) but not 100% sure for the metrics. There will be another PR for ElasticSearch. My goal is to get accurate number for total number of workflows running/closed in an environment given time range.

Let's double check in tomorrow's scrum.

demirkayaender:
Abandoning this since issue only exists for ElasticSearch (handled in another PR)

PR #4534, opened 2021-10-04
Revert "Use common dispatcher for public client outbound (#4523)"

vytautas-karpavicius:
This reverts commit a094a330b5ebd40dde60a5661cf35faec0eaf24d.

**What changed?**
Issue was found, reverting to unblock Java client release.

**Why?**
```
{"level":"info","ts":"2021-10-04T12:07:22.965-0700","msg":"Received a ring changed event","service":"cadence-frontend","component":"service-resolver","logging-call-at":"rpServiceResolver.go:199"}
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x59faca6]

goroutine 2231 [running]:
go.uber.org/yarpc/internal/observability.(*Middleware).handlePanicForCall(0xc0001ad680, 0xc000702460, 0x6ae5ec0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/qlong/go/pkg/mod/go.uber.org/yarpc@v1.56.0/internal/observability/middleware.go:315 +0x18d
panic(0x6597860, 0x822cea0)
        /usr/local/go-dists/1.16.2/go/src/runtime/panic.go:965 +0x1b9
github.com/uber/cadence/common/rpc.(*authOutboundMiddleware).Call(0x0, 0x6ef1cf8, 0xc00140c780, 0xc00184a900, 0x6efae78, 0xc00124ae20, 0x0, 0x0, 0x0)
        /Users/qlong/cadence/common/rpc/middleware.go:37 +0x26
go.uber.org/yarpc/api/middleware.unaryOutboundWithMiddleware.Call(...)
        /Users/qlong/go/pkg/mod/go.uber.org/yarpc@v1.56.0/api/middleware/outbound.go:113
go.uber.org/yarpc/internal/outboundmiddleware.unaryChainExec.Call(0xc00124af10, 0x0, 0x0, 0x6efbeb8, 0xc000b234d0, 0x6ef1cf8, 0xc00140c780, 0xc00184a900, 0x825de40, 0x6ef1cf8, ...)
        /Users/qlong/go/pkg/mod/go.uber.org/yarpc@v1.56.0/internal/outboundmiddleware/chain.go:103 +0x8d
[...]

PR #4537, opened 2021-10-05
Use common dispatcher for public client outbound #2

vytautas-karpavicius:
**What changed?**
Same as #4523, but with fixed middleware pointer, which required the revert #4534.

authMiddleware was stored as a pointer, which is not equal to nil which is checked when wrapping a middleware here: https://github.com/yarpc/yarpc-go/blob/dev/api/middleware/outbound.go#L59

Also updated unit test to ensure true nil is used when auth is not configured.

---

Stop using `DispatcherProvider` for creating public client. Instead use common dispatcher from rpc.Factory.

Builder for public client outbound was created which also includes auth middleware initialization. It was previously located within server.go itself. Moving it here allows consolidating all logic related logic in one place and covering it with unit tests.

**Why?**
To eventually delete `DispatcherProvider` entirely.

**How did you test it?**
Existing tests and new unit tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/rpc/outbounds.go
+++ b/common/rpc/outbounds.go
@@ -43,6 +43,11 @@ type OutboundsBuilder interface {
 	Build(*grpc.Transport, *tchannel.Transport) (yarpc.Outbounds, error)
 }
 
+type publicClientOutbound struct {
+	address        string
+	authMiddleware middleware.UnaryOutbound
+}
+
 func newPublicClientOutbound(config *config.Config) (publicClientOutbound, error) {
 	if len(config.PublicClient.HostPort) == 0 {
 		return publicClientOutbound{}, fmt.Errorf("need to provide an endpoint config for PublicClient")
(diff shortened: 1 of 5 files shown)


common/rpc/outbounds.go
```diff
@@ -21,12 +21,57 @@
...
+		authMiddleware = &authOutboundMiddleware{authProvider}
+	}
+
+	return publicClientOutbound{config.PublicClient.HostPort, authMiddleware}, nil
+}
+
+type publicClientOutbound struct {
+	address        string
+	authMiddleware middleware.UnaryOutbound
+}
```
mantas-sidlauskas, line 68: move this up

mantas-sidlauskas (approved):

common/rpc/outbounds.go
vytautas-karpavicius: done

PR #4545, opened 2021-10-06
Add inbound metrics middleware

vytautas-karpavicius:
**What changed?**
1. Added `caller` metrics tag for inbound requests. This is achieved by adding `inboundMetricsMiddleware` that attaches additional tags on the context.
2. This middleware greatly simplifies such request based tags, thus in addition to `caller`, `transport` tagging was moved here as well. Previously it was scattered across all handlers. Wow we can keep it in only one place.
3. Refactored `metrics.TagContext`. There was two variants of this: one accepting one tag `metrics.TagContext`, other - multiple tags via var args params `metrics.TagContexts`. The latter one can cover both cases.

**Why?**
`caller` tag is needed to distinguish xdc traffic from regular traffic. This can be used to track clients gRPC migration status.

**How did you test it?**
Added unit test for new middleware.
Tested with local server and local grafana to ensure that new `caller` tag is emited.

**Potential risks**
Previously tchannel traffic was emited with `thrift` tag. This is not entirely correct. Now transport name is set correctly. This change may require dashboard updating.

**Release notes**

**Documentation Changes**


mantas-sidlauskas (approved):
👍

PR #4605, opened 2021-10-27
Update CLI client factory to use grpc clients

fenghan89:
Updated CLI client factory to use grpc clients instead of thrift clients

Thrift interfaces are being replaced with gRPC 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


tools/cli/util.go
```diff
@@ -743,26 +808,32 @@ func parseTimeDuration(duration string) (dur time.Duration, err error) {
...
+	return fmt.Sprintf("cadence-cli@%s", getHostName())
 }
 
-func getCliIdentity() string {
+// GetWorkerIdentity gets a default identity for the worker.
+//
+// This contains a random UUID, generated each time it is called, to prevent identity collisions when workers share
+// other host/pid/etc information.  These alone are not guaranteed to be unique, especially when Docker is involved.
+// Take care to retrieve this only once per worker.
+func getWorkerIdentity(tasklistName string) string {
```
vytautas-karpavicius, line 827: Why do we need `getWorkerIdentity`? IMO `getCliIdentity` is enough for this application and could be used everywhere.

tools/cli/adminClusterCommands.go
```diff
@@ -87,27 +87,43 @@ func AdminRebalanceStart(c *cli.Context) {
...
+		Identity:                            getWorkerIdentity(""),
+		WorkflowIDReusePolicy:               types.WorkflowIDReusePolicyAllowDuplicate.Ptr(),
+		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(60),
+		Input:                               input,
+		TaskList: &types.TaskList{
+			Name: failovermanager.TaskListName,
+		},
+		Memo: &types.Memo{
+			Fields: map[string][]byte{
+				common.MemoKeyForOperator: []byte(getOperator()),
```
vytautas-karpavicius, line 112: This will likely need conversion to json as well. https://github.com/uber-go/cadence-client/blob/master/internal/internal_workflow_client.go#L1165

tools/cli/adminFailoverCommands.go
```diff
@@ -171,18 +200,34 @@ func AdminFailoverList(c *cli.Context) {
...
+
+	request := &types.QueryWorkflowRequest{
+		Domain: getSystemLocalDomainName(),
+		Execution: &types.WorkflowExecution{
+			WorkflowID: workflowID,
+			RunID:      runID,
+		},
+		Query: &types.WorkflowQuery{
+			QueryType: failovermanager.QueryType,
+			// might need args here?
```
vytautas-karpavicius, line 215: Should be fine. Go client is passing nil here when no var args are provided.

tools/cli/workflowBatchCommands.go
```diff
@@ -202,13 +212,38 @@ func StartBatchJob(c *cli.Context) {
...
+		RequestID:                           uuid.New(),
+		WorkflowID:                          workflowId,
+		ExecutionStartToCloseTimeoutSeconds: common.Int32Ptr(int32(batcher.InfiniteDuration.Seconds())),
+		TaskList:                            &types.TaskList{Name: batcher.BatcherTaskListName},
+		Memo: &types.Memo{
+			Fields: map[string][]byte{
+				"Reason": []byte(reason),
+			},
+		},
+		SearchAttributes: &types.SearchAttributes{
```
vytautas-karpavicius, line 230: Same as Memo, Search attributes are subject to data converter in the sdk: https://github.com/uber-go/cadence-client/blob/master/internal/internal_workflow_client.go#L1181

tools/cli/workflowBatchCommands.go
```diff
@@ -202,13 +212,38 @@ func StartBatchJob(c *cli.Context) {
...
+			IndexedFields: map[string][]byte{
+				"CustomDomain": []byte(domain),
+				"Operator":     []byte(operator),
+			},
+		},
+		WorkflowType: &types.WorkflowType{Name: batcher.BatchWFTypeName},
+		Input:        input,
+	}
+	// which domain does this run in?
```
vytautas-karpavicius, line 239: This should be `common.BatcherLocalDomainName`.
While we query for domains provided by user `--domain`, the actual workflow runs in `common.BatcherLocalDomainName`. Previous client used this when it was initialized at the start of the function, therefore we need to pass it in start request now.

vytautas-karpavicius:
> Hey @fenghan89 , Thank you so much for making this for us!
> 
> I would suggest adding a optional global param `--grpc` (with shortcut `-g` to switch to gRPC.
> 
> Most users including my companies are still on TChannel. Making this hard switch will cause lots of problems.
> 
> cc @meiliang86 @vytautas-karpavicius

+1
Yes, global flag to switch transport is easy to implement and would be beneficial. Although I would argue that it may be better to provide `--transport` flag with default value `tchannel` that will be flipped to `grpc` in some future release. Also this flag could come together with environment variable, so that users do not need to specify it on every command. Such setup will is less likely to break stuff if Cadence CLI is used in some user scripts.

tools/cli/defs.go
```diff
@@ -26,11 +26,17 @@ import (
...
+	"github.com/uber/cadence/common/types"
 )
 
 const (
-	localHostPort = "127.0.0.1:7933"
+	localHost    = "127.0.0.1"
+	tchannelPort = localHost + ":7933"
+	grpcPort     = localHost + ":7833"
+
+	grpcChannel   = "grpc"
```
vytautas-karpavicius, line 38: Channel is specific to TChannel. Maybe just call it grpcTransport, tchannelTransport intead?

tools/cli/taskListCommands.go
```diff
@@ -23,22 +23,32 @@ package cli
...
-	response, err := wfClient.DescribeTaskList(ctx, taskList, taskListType)
+
+	request := &types.DescribeTaskListRequest{
+		Domain: domain,
+		TaskList: &types.TaskList{
+			Name: taskList,
+		},
+		TaskListType: taskListType,
+	}
+	response, err := wfClient.DescribeTaskList(ctx, request, cc.GetDefaultCLIYarpcCallOptions()...)
```
vytautas-karpavicius, line 51: Do we set `GetDefaultCLIYarpcCallOptions` on every call?

If so, those can be set using common middleware on the dispatcher. It would reduce such boilerplate code. But this may as well be a separate PR.

tools/cli/workflowBatchCommands.go
```diff
@@ -27,16 +27,19 @@ import (
...
 	"github.com/uber/cadence/common"
 	cc "github.com/uber/cadence/common/client"
 	"github.com/uber/cadence/service/worker/batcher"
+
+	"github.com/uber/cadence/common/types"
 )
 
 var (
+	//TODO convert feature flagd into yarpc options
 	DefaultClientOptions = cclient.Options{
```
vytautas-karpavicius, line 43: Is this still needed?

tools/cli/taskListCommands.go
fenghan89: ya, so the default yarpc options are being passed in cadence-client when `newChannelContext` is called

longquanzheng (commented):
Remove myself from blocker

longquanzheng (approved):

tools/cli/adminClusterCommands.go
```diff
@@ -87,27 +87,45 @@ func AdminRebalanceStart(c *cli.Context) {
...
+		ErrorAndExit("Failed to serialize params for failover workflow", err)
+	}
+	memo, err := getWorkflowMemo(map[string]interface{}{
+		common.MemoKeyForOperator: getOperator(),
+	})
+	if err != nil {
+		ErrorAndExit("Failed to serialize memo", err)
+	}
+	request := &types.StartWorkflowExecutionRequest{
+		Domain:                              common.SystemLocalDomainName,
```
vytautas-karpavicius, line 106: nit: use `getSystemLocalDomainName()` for consistency with other places

tools/cli/workflowBatchCommands.go
```diff
@@ -202,13 +216,39 @@ func StartBatchJob(c *cli.Context) {
 		},
 		RPS: rps,
 	}
-	wf, err := client.StartWorkflow(tcCtx, options, batcher.BatchWFTypeName, params)
+	input, err := json.Marshal(params)
+	if err != nil {
```
vytautas-karpavicius, line 220: This if and couple more below are empty.

PR #4610, opened 2021-10-30
Support partial retries for ApplyParentPolicy

demirkayaender:
**What changed?**
Before this change any failure in applyParentPolicy operation would lead to a full retry where we retry even the successful policy applications. With this change we are only going to retry the failed ones.

Depends on https://github.com/uber/cadence-idl/pull/95

**Why?**
So we can save resources and operate more efficiently.

**How did you test it?**
Unit tests.

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -239,37 +239,69 @@ func (t *crossClusterTargetTaskExecutor) executeApplyParentClosePolicyTask(
...
+				anyErr = err
+			}
+		}
+		// Optimize SOURCE SIDE RETRIES by returning each child status separately
+		if err != nil {
+			response.FailedChildren = append(
+				response.FailedChildren,
+				&types.ApplyParentClosePolicyResult{
+					Child:       childAttrs,
+					FailedCause: failedCause,
```
demirkayaender, line 286: talked offline, it might be worth to have another iteration to add status to the request and update the request for the target side retries. 

this way we avoid false positive success.

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -302,29 +303,64 @@ func (t *crossClusterSourceTaskExecutor) executeApplyParentClosePolicyTask(
...
+	}
+	if len(failedDomains) > 0 {
+		taskInfo.TargetDomainIDs = failedDomains
+		t.setTaskState(task, ctask.TaskStatePending, processingStateInitialized)
+		return err
+	}
+
+	// When processing state = processingStateInitialized, there's nothing we need to do
+	// task is already complete, ack the task. All the other states are invalid
+	if processingState(task.response.TaskState) != processingStateInitialized {
```
yycptt, line 359: we should do this sanity check before we process the results I think?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -302,29 +303,61 @@ func (t *crossClusterSourceTaskExecutor) executeApplyParentClosePolicyTask(
...
+	if len(domainsToRegenerateTask) > 0 {
+		// since there are domains we can retry, we need to combine them with failed ones too for the retry
+		for domainID := range failedDomains {
+			domainsToRegenerateTask[domainID] = struct{}{}
+		}
+		taskInfo := task.GetInfo().(*persistence.CrossClusterTaskInfo)
+		taskInfo.TargetDomainIDs = domainsToRegenerateTask
+		return t.generateNewTask(ctx, wfContext, mutableState, task)
+	} else if len(failedDomains) > 0 {
+		taskInfo.TargetDomainIDs = failedDomains
```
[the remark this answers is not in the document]
yycptt: return `errContinueExecution` in this case

service/history/task/cross_cluster_target_task_executor.go
```diff
@@ -239,37 +239,83 @@ func (t *crossClusterTargetTaskExecutor) executeApplyParentClosePolicyTask(
...
-			*childAttrs.ParentClosePolicy,
-		)
 		switch err.(type) {
-		case *types.EntityNotExistsError, *types.WorkflowExecutionAlreadyCompletedError, *types.CancellationAlreadyRequestedError:
+		case *types.EntityNotExistsError,
+			*types.WorkflowExecutionAlreadyCompletedError,
+			*types.CancellationAlreadyRequestedError:
 			// expected error, no-op
-			break
+			scope.IncCounter(metrics.ParentClosePolicyProcessorSuccess)
```
yycptt, line 294: I am thinking maybe we should:
- emit those metrics at the source cluster
- let the source cluster determine how to handle those errors. The target cluster just reports the result/failed cause as is.
What's your thought? 
Those are not blocking, we can do them later if we want to.

yycptt (commented):
Looks good overall. Mostly nit/non-blocking comments

service/history/task/cross_cluster_source_task_executor.go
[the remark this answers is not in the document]
demirkayaender: Should this be the same for record child completion task?

common/types/shared.go
```diff
@@ -10816,20 +10816,71 @@ func (v *ApplyParentClosePolicyAttributes) GetParentClosePolicy() (o *ParentClos
 	return
 }
 
+type ApplyParentClosePolicyRequest struct {
+	Child     *ApplyParentClosePolicyAttributes `json:"child,omitempty"`
+	Completed bool                              `json:"completed,omniempty"`
+}
+
+// GetChild is an internal getter (TBD...)
+func (v *ApplyParentClosePolicyRequest) GetParentClosePolicy() (o *ApplyParentClosePolicyAttributes) {
```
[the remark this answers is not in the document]
yycptt: Rename to `GetChild`?

service/history/task/cross_cluster_source_task_executor.go
```diff
@@ -361,7 +404,7 @@ func (t *crossClusterSourceTaskExecutor) executeRecordChildWorkflowExecutionComp
 			types.CrossClusterTaskFailedCauseWorkflowAlreadyCompleted:
 			// Do nothing, these errors are expected if the target workflow is already closed
 		default:
-			return errUnexpectedErrorFromTarget
+			return errContinueExecution
```
yycptt, line 407: To fix this issue we also need to reset the task processing state before returning the error. Otherwise, it will be `reported` and the task is not available for polling.
`t.setTaskState(task, ctask.TaskStatePending, processingState(task.response.TaskState))`
I will land my pr after this one, so I can fix it.

yycptt (approved):

yycptt (approved):

PR #4612, opened 2021-11-01
Parent close policy should apply to child workflow only

yycptt:
**What changed?**
- Add childWorkflowOnly flag to history terminate workflow execution request
- Use childworkflowOnly flag when applying parent close policy
- In parent close policy workflow use history client and signal remote frontend if child domain is not active
- IDL change: https://github.com/uber/cadence-idl/pull/97

**Why?**
- Currently, it's possible that parent close policy is applied to a non-childworkflow. As an example, if workflow closes when a child workflow is still in init state and there's another workflow with the same workflowID, we will apply the close policy to the current open non-childworkflow
- In terms of fix, two approaches
   1. Do not apply parent close policy if child is in Init state. This can work in most cases, however, we don't know the child is actually started or not just by looking at the mutable state of the parent. (e.g. start child timeout -> workflow close -> tx task retry). So we may fail to apply the policy to some started children.
   2. The fix implemented in this PR. Use the childWorkflowOnly flag when terminating or cancelling childworkflows to make sure the workflow is a child of the workflow that is closing. The fix includes three parts:
      - Currently, there's no childWorkflowOnly option when terminating workflow, so add related fields and logic for that
      - Use the field when processing transfer and cross-cluster task for applying parent close policy.
      - Since childWorkflowOnly field is only defined in history request (not frontend) and parent close policy workflow needs to handle the case where the child domain is not active in the current cluster, we can no longer relying on the auto-forwarding logic provided by the frontend (since childWorkflowOnly field is not defined). Instead, once we found a child is not active, we need to signal the workflow (system domain) in the remote cluster to process it.
Using this approach, we can avoid the issue in approach 1. 

   However, neither approach can prevent the case where childworkflow is started after the parent close policy is applied. For example:
  - Processing transfer task, found workflow is running, send the start child request
  - Start child request timeout
  - Workflow close (due to timeout, etc.)
  - Process parent close policy (child still in Init state), no-op since child is not started yet
  - Start child request got applied.

**How did you test it?**
- WIP

**Potential risks**

**Release notes**

**Documentation Changes**


longquanzheng:
Just read through it. So far overall looks good to me

proto/internal/uber/cadence/history/v1/service.proto
```diff
@@ -274,6 +274,8 @@ message ResetWorkflowExecutionResponse {
 message TerminateWorkflowExecutionRequest {
   api.v1.TerminateWorkflowExecutionRequest request = 1;
   string domain_id = 2;
+  api.v1.WorkflowExecution external_workflow_execution = 3;
+  bool child_workflow_only = 4;
```
longquanzheng, line 278: Can you add a comment for this field? (also would be nice for other `child_workflow_only` similarly). 
This name makes me think it's terminating the childWFs of the wf in the request. I think having comments will help a lot.

service/worker/parentclosepolicy/workflow.go
```diff
@@ -107,52 +115,166 @@ func ProcessorWorkflow(ctx workflow.Context) error {
...
+	remoteExecutions map[string][]RequestDetail,
+	numWorkflows int,
+) error {
+	for cluster, executions := range remoteExecutions {
+		remoteClient := clientBean.GetRemoteFrontendClient(cluster)
+		signalCtx, cancel := context.WithTimeout(ctx, signalTimeout)
+		// TODO: we may want to specify a custom data converter,
+		// instead of relying on the fact that the default data converter
+		// in client library is json
+		signalInput, err := json.Marshal(Request{
```
longquanzheng, line 249: I think you can use the SDK's dataConverter directly?

longquanzheng (approved):
Some minor comments

service/worker/parentclosepolicy/workflow.go
yycptt: Good point. I didn't know the default data converter is exported in the `encoded` package. Thanks!

proto/internal/uber/cadence/history/v1/service.proto
yycptt: Done. Added comments for signal, cancel and terminate.

PR #4614, opened 2021-11-03
Support DNS SRV Records within Ringpop

lindleywhite:
This PR adds the ability to configure an additional DNS method to reference DNS SRV records. This allows the initial service discovery of services to lookup hosts based on SRV records instead of having fixed ports within the host configuration. 

Cadence is required to have fixed ports when configuring it mainly because it is expected to be hosted on primary VMs. However, when you deploy Cadence to a containerized environment you may not be able to fix ports to hosts and would rather have dynamic discoverability.

The tests are supplied with the code. The code just extends what is already there.

The worst case would be that the net.DefaultResolver can not lookup SRV records and thus the server components erring out on startup.

**Release notes**
Cadence can now support DNS SRV records within the configuration. 

**Documentation Changes**
This could be added as an option within the ringpop configuration. Please advise.


emrahs:
@mantas-sidlauskas

common/config/ringpop.go
```diff
@@ -121,6 +123,8 @@ func parseBootstrapMode(
 		return BootstrapModeCustom, nil
 	case "dns":
 		return BootstrapModeDNS, nil
+	case "dns-srv":
```
longquanzheng, line 126: can you help update this comment: https://github.com/uber/cadence/blob/8510816511e1532514bd11aadc09966278db3975/common/config/config.go#L160

to 
`BootstrapMode is a enum that defines the ringpop bootstrap method, currently supports: hosts, file, custom, dns and dns-srv`

common/config/ringpop.go
```diff
@@ -56,6 +56,8 @@ const (
 	// BootstrapModeDNS represents a list of hosts passed in the configuration
 	// to be resolved, and the resulting addresses are used for bootstrap
 	BootstrapModeDNS
+	// BootstrapModeDNSSRV represents a list of DNS hosts passed in the configuration to resolve host and ports, resulting in a list of addresses
```
longquanzheng, line 59: Nit: add more details about a use case -- it allows using different ports as each SRV record can use a different port.

common/config/ringpop.go
```diff
@@ -295,6 +300,84 @@ func (provider *dnsProvider) Hosts() ([]string, error) {
...
+}
+
+func (provider *dnsSRVProvider) Hosts() ([]string, error) {
+	var results []string
+	resolvedHosts := map[string][]string{}
+
+	for _, service := range provider.UnresolvedHosts {
+		serviceParts := strings.Split(service, ".")
+		if len(serviceParts) <= 2 {
+			provider.Logger.Warn("could seperate service name from domain", tag.Address(service))
```
longquanzheng, line 339: Do you mean "could not"? 
Also should we error out this case? this could mean some wrong config value.

common/config/ringpop.go
```diff
@@ -295,6 +300,84 @@ func (provider *dnsProvider) Hosts() ([]string, error) {
...
+			continue
+		}
+		serviceName := serviceParts[0]
+		domain := strings.Join(serviceParts[1:], ".")
+		resolved, exists := resolvedHosts[serviceName]
+		if !exists {
+			_, addrs, err := provider.Resolver.LookupSRV(context.Background(), serviceName, "tcp", domain)
+
+			if err != nil {
+				provider.Logger.Warn("could not resolve host", tag.Address(serviceName), tag.Error(err))
```
longquanzheng, line 349: We should error out this case too. Ringpop requires all bootstrap hosts to be discovered, otherwise there can be network partition issue: 
E.g. In a 10 node cluster -- we configure two nodes to be discovered as bootstrap, but only one of them is returned in different times. 
Now it's possible that 5 nodes can discover the same node A, but other 5 nodes discover node B. Then Cadence will start running in a coruppted modes. There are two rings competing each other. They will compete for acquiring Cadence history shards and will constantly having errors.

This is rare, though. Unless you think this error is normal enough that we should ignore.

common/config/ringpop.go
```diff
@@ -236,6 +240,7 @@ func (factory *RingpopFactory) createRingpop() (*membership.RingPop, error) {
 
 type dnsHostResolver interface {
 	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+	LookupSRV(ctx context.Context, service string, proto string, name string) (cname string, addrs []*net.SRV, err error)
```
mantas-sidlauskas, line 243: nit: 
```suggestion
	LookupSRV(ctx context.Context, service, proto, name string) (cname string, addrs []*net.SRV, err error)
```

common/config/ringpop.go
```diff
@@ -295,6 +300,84 @@ func (provider *dnsProvider) Hosts() ([]string, error) {
...
+	Logger          log.Logger
+}
+
+func newDNSSRVProvider(
+	hosts []string,
+	resolver dnsHostResolver,
+	logger log.Logger,
+) *dnsSRVProvider {
+
+	set := map[string]struct{}{}
```
mantas-sidlauskas, line 315: What is happening here? Can we just pass `hosts` slice to `UreslovedHosts`?

lindleywhite: Ah yeah, I was confused by this as well but it seems the purpose is to remove any duplicates. See: https://github.com/uber/cadence/pull/1941#discussion_r290788721

common/config/ringpop.go
lindleywhite: I've changed this to be an error instead and modified the tests to expect that bad domains are an error case.

common/config/ringpop.go
lindleywhite: I think that this is a welcome change but it does however change the expected behavior from the DNS method. The previous method just assumes that any non resolvable host is non valid and therefore removed from the true host list. I think your suggestion is expecting DNS to be unavailable for a brief time. I've committed your suggestion.

longquanzheng (approved):

PR #4619, opened 2021-11-05
Use direct outbound for history client

vytautas-karpavicius:
**What changed?**
Client factory used to initialize and cache history clients for each peer. For each client a separate dispatcher was created via rpc factory. This is rather complicated setup and is not idiomatic on YARPC framework.

In order to route requests to different peers, yarpc provides [direct](https://pkg.go.dev/go.uber.org/yarpc/peer/direct#New) peer chooser. It uses `ShardKey` set on the request to route this particular call to the given host. We do not need to initialize multiple dispatchers and multiple clients with caching on top.

This PR:
- Introduces direct outbound builder
- Refactors out `history.PeerChooser`, documents it and adds unit tests.
- Refactors `history.Client` to drop client cache and instead adds `WithShardKey` for routing to appropriate instance.
- Updates `client.Factory` to wire up new setup.

**Why?**
To move forward towards use of single yarpc dispatcher.

The similar effort is needed for matching client as well. After that we can basically drop rpc.Factory and use single Dispatcher everywhere.

**How did you test it?**
New unit tests.
Integration tests.
Locally with cadence canary.

**Potential risks**
This also brings one observable change. History client will no longer use caller name `history-service-client`, but rather use the actual service name where the call is originating, for example: `cadence-frontend`.
Even though the we can override callerName, at the moment I do not think this is needed. I'm not aware where this could be used other than yarpc metrics.
IMO, the actual name of caller service is more useful than a constant string.

**Release notes**

**Documentation Changes**

### DIFF
--- a/client/history/client.go
+++ b/client/history/client.go
@@ -50,7 +50,7 @@ type (
 		tokenSerializer   common.TaskTokenSerializer
 		timeout           time.Duration
 		client            Client
-		peer              PeerResolver
+		peerResolver      PeerResolver
 		logger            log.Logger
 	}
 
@@ -75,7 +75,7 @@ func NewClient(
 		tokenSerializer:   common.NewJSONTaskTokenSerializer(),
 		timeout:           timeout,
 		client:            client,
-		peer:              peerResolver,
+		peerResolver:      peerResolver,
 		logger:            logger,
 	}
 }
@@ -85,7 +85,7 @@ func (c *clientImpl) StartWorkflowExecution(
 	request *types.HistoryStartWorkflowExecutionRequest,
 	opts ...yarpc.CallOption,
 ) (*types.StartWorkflowExecutionResponse, error) {
-	peer, err := c.peer.FromWorkflowID(request.StartRequest.WorkflowID)
+	peer, err := c.peerResolver.FromWorkflowID(request.StartRequest.WorkflowID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


client/history/peerResolver.go
```diff
@@ -0,0 +1,89 @@
...
+	addressMapper  AddressMapperFn
+}
+
+type AddressMapperFn func(string) (string, error)
+
+// NewPeerResolver creates a new history peer resolver.
+func NewPeerResolver(numberOfShards int, membership membership.Monitor, addressMapper AddressMapperFn) PeerResolver {
+	return PeerResolver{
+		numberOfShards: numberOfShards,
+		membership:     membership,
```
mantas-sidlauskas, line 44: If you need only `history` resolver, you can replace membership with `resolver` and pass history resolver here

vytautas-karpavicius: Agree, updated the PR.

client/history/client.go
```diff
@@ -64,15 +65,17 @@ func NewClient(
...
 	logger log.Logger,
 ) Client {
 	return &clientImpl{
 		numberOfShards:    numberOfShards,
 		rpcMaxSizeInBytes: rpcMaxSizeInBytes,
 		tokenSerializer:   common.NewJSONTaskTokenSerializer(),
 		timeout:           timeout,
-		clients:           clients,
+		client:            client,
+		peer:              peerResolver,
```
mantas-sidlauskas, line 78: nit: rename `peer` to `peerResolver` to make it more obvious

mantas-sidlauskas (approved):

PR #4627, opened 2021-11-10
Remove Membership Factory

mantas-sidlauskas:
**What changed?**
`membership.Monitor` is used instead of `MembershipMonitorFactory`

**Why?**
To simplify upcoming changes for pluggable `membership` options.

**How did you test it?**
Unit tests/local test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/membership/interfaces.go
+++ b/common/membership/interfaces.go
@@ -23,20 +23,12 @@
 package membership
 
 import (
-	"errors"
-
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/types"
 )
 
-// ErrUnknownService is thrown for a service that is not tracked by this instance
-var ErrUnknownService = errors.New("Service not tracked by Monitor")
-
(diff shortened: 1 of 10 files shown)


common/membership/interfaces.go
```diff
@@ -23,21 +23,13 @@
...
-
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/types"
 )
 
-// ErrUnknownService is thrown for a service that is not tracked by this instance
-var ErrUnknownService = errors.New("Service not tracked by Monitor")
-
 // ErrInsufficientHosts is thrown when there are not enough hosts to serve the request
 var ErrInsufficientHosts = &types.InternalServiceError{Message: "Not enough hosts to serve the request"}
```
vytautas-karpavicius, line 31: What is the reason for removing up/below errors but not this one? All three of them does not seem to be used outside of this package.

mantas-sidlauskas: will address this in next PR

PR #4630, opened 2021-11-11
Use generated proto types from cadence-idl repo

vytautas-karpavicius:
**What changed?**
- `go get github.com/uber/cadence-idl`
- Update Makefile to exclude generated types in idls from formatting
- Update Makefile to not generate public API proto types
- Removed existing generated public API proto types
- Replaced `github.com/uber/cadence/.gen/proto/api/v1` with `github.com/uber/cadence-idl/go/proto/api/v1` everywhere
- Regenerate internal proto types (import path updates, binary descriptor changes)

**Why?**
Public API proto types are generated on cadence-idl repo (https://github.com/uber/cadence-idl/pull/94) so that they could be shared between client/server. Otherwise it results in duplicated type registrations when using both.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Makefile
```diff
@@ -225,7 +226,7 @@ $(THRIFT_GEN): $(THRIFT_FILES) $(BIN)/thriftrw $(BIN)/thriftrw-plugin-yarpc | $(
 PROTO_ROOT := proto
 # output location is defined by `option go_package` in the proto files, all must stay in sync with this
 PROTO_OUT := .gen/proto
```
mantas-sidlauskas, line 228: is this still relevant?

mantas-sidlauskas (approved):

Makefile
vytautas-karpavicius: Yes, since we still generate all internal types here. Those that are used for history/matching services.
Only public API types were moved to cadence-idl repo, internal types stays here.

PR #4639, opened 2021-11-13
Apply ConnectAttributes for postgres

wenerme:
**What changed?**

apply ConnectAttributes to pg dsn

**Why?**

override conn params, allowed override sslmode without enable tls, enable tls will check ca.
allowed add extra params. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


longquanzheng:
@wenerme thank you so much for the contribution! Just to confirm, have you tested this in your environment? Because we don’t have integration test to test every connection setup.

wenerme:
@longquanzheng Yes, tested on my local, due to https://github.com/uber/cadence/blob/master/tools/sql/handler.go#L239 can not --tls to use the TLS.SSLMode, so modified the code, finished schema setup.

longquanzheng (approved):

PR #4641, opened 2021-11-15
Fix build by switching to golang:1.17-alpine3.13

vytautas-karpavicius:
**What changed?**
Use different golang:1.17 variant that is based on alpine3.13 instead.
The newer version (3.14) broke our build with `Operation not permitted` errors.
Others are having similar issues: https://github.com/docker-library/php/issues/1177

**Why?**
master build is failing after #4632

**How did you test it?**
`docker build .`

**Potential risks**

**Release notes**

**Documentation Changes**


wenerme:
You better upgrade docker instead.

longquanzheng:
> You better upgrade docker instead.

Thanks for looking @wenerme 
I didn't look at lots of the context. The build does work for me as I am using the newer version. 
Do you any cons if using this `alpine3.13` in the PR?

wenerme:
No obviously cons, but it's always better to use newer version, beside, alpine is about to release 3.15.

longquanzheng:
> No obviously cons, but it's always better to use newer version, beside, alpine is about to release 3.15.

Actually not just issue for local docker build, the master image releasing pipeline is also broken:
https://buildkite.com/uberopensource/cadence-server/builds/12484#7b0314ab-02c0-4231-bf50-0d1c62a176b4

I just approved it and let it go through. We can have a ticket to improve it later on.

PR #4654, opened 2021-11-22
Fix remote sync match for standby domains and task creation time

yycptt:
**What changed?**
- Fix remote sync match check for standby domains
- Fix persisted task creation time for cassandra

**Why?**
- Currently, for standby domains, matching will persist the task no matter it's forwarded from child partition or not. But forwarded tasks has no TTL (since they are not supposed to be persisted) and if we persist them in the standby tasklist, they will live in DB forever if the user never does domain failover. This will cause the tasklist partition to grow infinitely and incur high db latency.

**How did you test it?**
- Unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlTaskStore.go
+++ b/common/persistence/nosql/nosqlTaskStore.go
@@ -226,6 +226,7 @@ func (t *nosqlTaskStore) CreateTasks(
 	ctx context.Context,
 	request *p.InternalCreateTasksRequest,
 ) (*p.CreateTasksResponse, error) {
+	now := time.Now()
 	var tasks []*nosqlplugin.TaskRowForInsert
 	for _, t := range request.Tasks {
 		task := &nosqlplugin.TaskRow{
@@ -236,7 +237,7 @@ func (t *nosqlTaskStore) CreateTasks(
 			WorkflowID:   t.Execution.GetWorkflowID(),
 			RunID:        t.Execution.GetRunID(),
 			ScheduledID:  t.Data.ScheduleID,
-			CreatedTime:  t.Data.CreatedTime,
+			CreatedTime:  now,
 		}
 		ttl := int(t.Data.ScheduleToStartTimeout.Seconds())
 		tasks = append(tasks, &nosqlplugin.TaskRowForInsert{
--- a/common/persistence/nosql/nosqlTaskStore.go
+++ b/common/persistence/nosql/nosqlTaskStore.go
@@ -226,7 +226,6 @@ func (t *nosqlTaskStore) CreateTasks(
 	ctx context.Context,
 	request *p.InternalCreateTasksRequest,
 ) (*p.CreateTasksResponse, error) {
-	now := time.Now()
 	var tasks []*nosqlplugin.TaskRowForInsert
 	for _, t := range request.Tasks {
 		task := &nosqlplugin.TaskRow{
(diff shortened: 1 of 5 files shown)


demirkayaender (approved):

common/persistence/nosql/nosqlTaskStore.go
```diff
@@ -237,7 +236,7 @@ func (t *nosqlTaskStore) CreateTasks(
 			WorkflowID:   t.Execution.GetWorkflowID(),
 			RunID:        t.Execution.GetRunID(),
 			ScheduledID:  t.Data.ScheduleID,
-			CreatedTime:  now,
+			CreatedTime:  t.Data.CreatedTime,
```
longquanzheng, line 239: Is there no test for this field? how about separating a different PR and add the test assertion?

common/persistence/nosql/nosqlplugin/cassandra/tasks.go
```diff
@@ -373,7 +373,7 @@ func (db *cdb) InsertTasks(
 				task.WorkflowID,
 				task.RunID,
 				scheduleID,
-				tasklistCondition.LastUpdatedTime,
+				task.CreatedTime,
```
longquanzheng, line 376: same as above.

service/matching/taskListManager.go
```diff
@@ -236,18 +236,26 @@ func (c *taskListManagerImpl) AddTask(ctx context.Context, params addTaskParams)
 			return nil, err
 		}
 
+		isForwarded := params.forwardedFrom != ""
+
 		if domainEntry.GetDomainNotActiveErr() != nil {
-			r, err := c.taskWriter.appendTask(params.execution, params.taskInfo)
+			// standby task, only persist when task is not forwarded from child partition
 			syncMatch = false
+			if isForwarded {
```
longquanzheng, line 244: seems like we could also adding code to not try forwarding to root when the domain is not active. (but this is needed anyway, due to racing condition)

yycptt: yes, I will create a separate for not forwarding on standby

common/persistence/nosql/nosqlTaskStore.go
yycptt: I do, just fixed the test and I realized we need similar changes for SQL. 
so I will revert those changes and do it in another pr.

common/persistence/nosql/nosqlplugin/cassandra/tasks.go
yycptt: This is different from the above one and needs to be fixed & released. 

It should use the same value as the no ttl case above.

longquanzheng: yeah, I meant we need to add an assertion in the persistence test so that it wont be broken again.

yycptt: so this is a bit complex. there's nothing we need to add in the test since there's already a check for task create time. The problem is that in tests, when calling the CreateTask API, we provided a tasklist info with LastUpdateTime, but in application code when calling CreateTask, we don't have the LastUpdateTime so the value is empty. That's why previously the test is also passing. 

For now, I changed the test and cleared the tasklist LastUpdatedTime field when calling CreateTask in persistence tests.

service/matching/taskListManager.go
yycptt: https://github.com/uber/cadence/issues/4656

common/persistence/nosql/nosqlplugin/cassandra/tasks.go
longquanzheng: I see. So the test was cheating..

longquanzheng (approved):

PR #4662, opened 2021-11-30
Improve cross cluster components shutdown logic

yycptt:
**What changed?**
- Improve cross cluster task processor and fetcher shutdown logic so that on shutdown, they won't block on pending requests. This is done by cancelling the context used for making calls.
- Changes in the cross-cluster task processor is not required since we currently shut down the task fetcher first and the fetcher will fail all task processors' requests to it. But if in the future we shutdown task processors first, without this fix, they will be blocked since they're waiting for results from the task fetcher. 

**Why?**
- Improve shutdown logic

**How did you test it?**
- Tested locally

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/cross_cluster_task_processor.go
```diff
@@ -219,9 +225,13 @@ func (p *crossClusterTaskProcessor) processLoop() {
 		sw := p.metricsScope.StartTimer(metrics.CrossClusterFetchLatency)
 
 		var taskRequests []*types.CrossClusterTaskRequest
-		err := p.taskFetcher.Fetch(p.shard.GetShardID()).Get(context.Background(), &taskRequests)
```
demirkayaender, line 222: Do you mind creating a Jira task for this to examine all the context.Background() calls we have in the codebase?

demirkayaender (approved):

service/history/task/cross_cluster_task_processor.go
```diff
@@ -303,6 +314,10 @@ func (p *crossClusterTaskProcessor) processTaskRequests(
 		}
 		cancel()
 
+		if p.ctx.Err() != nil {
+			return
+		}
```
Groxx, line 319: would this be better above, in the loop?  or is that unsafe to interrupt?  (though if anything under it checks the context, it may still be interrupted)

Groxx:
I don't follow the set of changes as a whole, but `context.Background()--` is very very much a good thing.

service/history/task/cross_cluster_task_processor.go
yycptt: Sure. I don't have a strong opinion on this. Moving it into the for loop also works and probably can save us some iterations during the shutdown. 

I was thinking the for loop is already non-blocking once the context is cancelled on shutdown, and we just need one check before sending the response back instead of checking it in each iteration.

service/history/task/cross_cluster_task_processor.go
yycptt: https://github.com/uber/cadence/issues/4664

PR #4676, opened 2021-12-10
Add support for custom cassandra authenticators

johndelcastillo:
** What changed **
Propagated support for custom authenticators from the gocql driver to the cadence application, this also required updating the version of the gocql driver.

** Why **
Depending on how you configure your Cassandra cluster, it may expose a custom authenticator, which if not expected by the gocql driver, will result in an error and the connection being rejected.

** Verification **
To verify, I updated the cassandra test cluster to use password authentication and introduced a negative unit test to ensure the AllowedAuthenticators config is being passed down to the gocql driver correctly.

**Potential risks**
gocql driver was updated to the latest via `go get`, minimal risk should be included there.

**Release notes**
Added AllowedAuthenticator support in the gocql driver


docker/dev/cassandra-config.yml
```diff
@@ -0,0 +1,1302 @@
+
```
longquanzheng, line 1: Is it possible to not to use this file? I think you only want to change some of the config opts so is that possible to override via other way like env?

johndelcastillo: No, I had a look. You can't make this Cassandra docker image (https://hub.docker.com/_/cassandra) work with Auth without updating the yaml.

Its not strictly necessary if we are comfortable not testing the auth path of the gocql client.

longquanzheng: yeah I was about to say the same thing. Let's skip the test for this. 
Cadence doesn't have a convention to test every config like those. It's okay not to test because if they don't work, we will know :P

johndelcastillo: sure, i'll make the updates

environment/env.go
```diff
@@ -36,6 +36,16 @@ const (
 	CassandraPort = "CASSANDRA_DB_PORT"
 	// CassandraDefaultPort Cassandra default port
 	CassandraDefaultPort = "9042"
+	// CassandraUsername env
+	CassandraUsername = "CASSANDRA_DB_USERNAME"
+	// CassandraUsername env
+	CassandraDefaultUsername = "cassandra"
```
longquanzheng, line 42: let's make it back to empty. It's not just unit test, there are other integ test, pipeline and even dev envs already use empty username/password.

Changing it is going to be painful but unnecessary. I will keep it empty for simple

environment/env.go
```diff
@@ -36,6 +36,16 @@ const (
...
+	// CassandraPassword env
+	CassandraPassword = "CASSANDRA_DB_PASSWORD"
+	// CassandraPassword env
+	CassandraDefaultPassword = "cassandra"
+	// CassandraPassword env
+	CassandraAllowedAuthenticators = "CASSANDRA_DB_ALLOWEDAUTHENTICATORS"
```
longquanzheng, line 48: `CASSANDRA_DB_ALLOWED_AUTHENTICATORS` may look better

environment/env.go
emrahs: nit: comments smell like copy & paste :)

longquanzheng (approved):

PR #4678, opened 2021-12-11
Emit persistence latency histograms

Shaddoll:
**What changed?**

- Update metrics config to make reporting Interval configurable (default to 1s)
- Add RecordHistogramDuration method metrics client
- Refactor persistence metrics client
- Update persistence config to make latency histogram configurable
- Update persistence metrics client to emit persistence latency histograms (by default disabled, can be enabled by setting persistence config `enablePersistenceLatencyHistogramMetrics: true`)

**Why?**

- latency histogram is more accurate and can be aggregated, but it's more expensive, so add a config to allow users to not enable it
- metrics reporting uses lots of CPU, 1 sec interval may be too short for us, make it configurable

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**


common/config/metrics.go
```diff
@@ -66,6 +70,9 @@ var (
 // NewScope builds a new tally scope for this metrics configuration
 // Only one reporter type is allowed
 func (c *Metrics) NewScope(logger log.Logger, service string) tally.Scope {
+	if c.ReportingInterval <= 0 {
+		c.ReportingInterval = _defaultInterval
```
demirkayaender, line 74: nit: change the name to `_defaultReportingInterval`

demirkayaender (approved):

PR #4683, opened 2021-12-15
Add jitter for starting failover queue

yycptt:
**What changed?**
- Add some jitter for starting failover queue processing

**Why?**
- on domain failover, we will create a failover timer/transfer queue for every shard and all of them will start to load tasks at almost the same time. This caused a spike load and CAS latency in our underlying DB to become very high. As a short-term fix, added some jitter so that those failover queues won't start at the same time. 
- By default, there's no jitter for failover queue. 

**How did you test it?**
- Tested locally

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/queue/timer_queue_processor_base.go
+++ b/service/history/queue/timer_queue_processor_base.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"math/rand"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


service/history/queue/timer_queue_processor_base.go
```diff
@@ -154,8 +155,13 @@ func (t *timerQueueProcessorBase) Start() {
 
 	t.redispatcher.Start()
 
+	newPollTime := time.Time{}
+	if startJitter := t.options.MaxStartJitterInterval(); startJitter > 0 {
+		now := t.shard.GetTimeSource().Now()
+		newPollTime = now.Add(time.Duration(rand.Int63n(int64(startJitter))))
+	}
 	for _, queueCollections := range t.processingQueueCollections {
-		t.upsertPollTime(queueCollections.Level(), time.Time{})
```
Shaddoll, line 158: From the outcome perspective, is there any difference between `time.Time{}` and `time.Now()`?

yycptt: there's no difference between those two. Both make the timerGate fire as soon as possible.

Shaddoll (approved):

PR #4695, opened 2022-01-04
Improve standby task processing

Shaddoll:
**What changed?**

- Refactor standby task filter logic to move the special cases handling logic to task filter
- Improve standby task processing for the special cases to only process those tasks if the domain is replicated to that cluster
- Fix task_requests metrics

**Why?**
It's not necessary to process standby tasks if the domain is not replicated to the cluster. And it will reduce load to ES and history archival.

**How did you test it?**

**Potential risks**
Standby task which is supposed to be processed is not processed causing workflow not closed in visibility and history not deleted.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -521,8 +521,22 @@ func newTimerQueueStandbyProcessor(
 		}
 		if timer.TaskType == persistence.TaskTypeWorkflowTimeout ||
 			timer.TaskType == persistence.TaskTypeDeleteHistoryEvent {
-			// guarantee the processing of workflow execution history deletion
-			return true, nil
+			domainEntry, err := shard.GetDomainCache().GetDomainByID(timer.DomainID)
+			if err == nil {
+				if domainEntry.HasReplicationCluster(clusterName) {
+					// guarantee the processing of workflow execution history deletion
+					return true, nil
+				}
+			} else {
+				if _, ok := err.(*types.EntityNotExistsError); !ok {
+					// retry the task if failed to find the domain
+					logger.Warn("Cannot find domain", tag.WorkflowDomainID(timer.DomainID))
+					return false, err
+				} else {
+					logger.Warn("Cannot find domain, default to not process task.", tag.WorkflowDomainID(timer.DomainID), tag.Value(timer))
+					return false, nil
+				}
+			}
 		}
 		return taskAllocator.VerifyStandbyTask(clusterName, timer.DomainID, timer)
 	}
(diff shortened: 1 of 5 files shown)


service/history/queue/timer_queue_processor.go
```diff
@@ -519,6 +519,18 @@ func newTimerQueueStandbyProcessor(
...
 			return false, errUnexpectedQueueTask
 		}
+		if timer.TaskType == persistence.TaskTypeWorkflowTimeout ||
+			timer.TaskType == persistence.TaskTypeDeleteHistoryEvent {
+			domainEntry, err := shard.GetDomainCache().GetDomainByID(timer.DomainID)
+			if err == nil {
+				for _, cluster := range domainEntry.GetReplicationConfig().Clusters {
+					if cluster.ClusterName == clusterName {
+						// guarantee the processing of workflow execution history deletion
+						return true, nil
```
emrahs, line 529: Returning true here bypasses the VerifyStandbyTask call below. We should check for the negative case only here and let the next step decide if the task is good to go.

Shaddoll: Bypassing verifyStandbyTask is by design. Previously we always execute these 2 type of tasks even if verifyStandbyTask is false. The improvement is to not execute the tasks if the domain is not replicated to the cluster. For example, previously a task from local domain is executed in standby queue, but with this change it won't.

service/history/queue/timer_queue_processor.go
```diff
@@ -519,6 +519,18 @@ func newTimerQueueStandbyProcessor(
 		if !ok {
 			return false, errUnexpectedQueueTask
 		}
+		if timer.TaskType == persistence.TaskTypeWorkflowTimeout ||
+			timer.TaskType == persistence.TaskTypeDeleteHistoryEvent {
+			domainEntry, err := shard.GetDomainCache().GetDomainByID(timer.DomainID)
+			if err == nil {
```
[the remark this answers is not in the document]
Shaddoll: If err != nil, just fallback to VerifyStandbyTask

service/history/queue/timer_queue_processor.go
emrahs: What I mean is that returning true here is a definitive decision to execute the task; although it may be invalid according to the logic in VerifyStandbyTask. 

I think what we should do is to check the timer.TaskType only and return **false** as needed (regardless of the replication config). Then do the replication config check in the VerifyStandbyTask, since it applies to both timer and transfer tasks. Finally, return true in only one place, which is at the end of the VerifyStandbyTask.

emrahs: Chatted offline and decided to keep this as-is for now.

emrahs (approved):

PR #4704, opened 2022-01-19
WIP: Changes error handling for ServiceIsBusyError to use go 1.13 error conventions

davidporter-id-au:
As per https://go.dev/blog/go1.13-errors using error unwrapping with
Is/As allows for richer information to be embedded in returned errors,
and would (in the future) allow developers to add contextual information
without loss of fidelity in control flow.

This PR is 1/n for this particular error. It's intended as a spike to
determine how difficult such a change is to make, and a path for future
error handling.

**What changed?**
- Control flow which relied on type assertions for this type of error is now using errors.As
- Added some unit-tests to verify

**Why?**

**How did you test it?**
WIP

**Potential risks**

**Release notes**

**Documentation Changes**


client/history/client.go
```diff
@@ -820,9 +821,10 @@ func (c *clientImpl) GetReplicationMessages(
 			if err != nil {
 				c.logger.Warn("Failed to get replication tasks from client", tag.Error(err))
 				// Returns service busy error to notify replication
-				if _, ok := err.(*types.ServiceBusyError); ok {
+				var e *types.ServiceBusyError
+				if errors.As(err, &e) {
 					select {
-					case errChan <- err:
+					case errChan <- e:
```
davidporter-id-au, line 827: this... might be the desired behaviour? I figure if it 's a wrapped error, the channel won't expect having a wrapper being passed along, but it does mean the wrap is lost in this instance

PR #4708, opened 2022-01-24
Hashring: return Hostinfo struct instead of string

mantas-sidlauskas:
**What changed?**
Membership will return enhanced Hostinfo data. 
Additionally, Ringpop will add port labels to propagate supported transport ports across hashring members.

**Why?** 
Additional port data in `Hostinfo` will open easier RPC transport migrations. 

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


common/membership/resolver.go
```diff
@@ -145,12 +146,12 @@ func (rpo *MultiringResolver) Stop() {
 
 // WhoAmI asks to provide current instance address
 func (rpo *MultiringResolver) WhoAmI() (*HostInfo, error) {
-	address, err := rpo.provider.WhoAmI()
+	host, err := rpo.provider.WhoAmI()
```
vytautas-karpavicius, line 149: I think we can simply: `return rpo.provider.WhoAmI()` here

common/peerprovider/ringpopprovider/provider.go
```diff
@@ -129,8 +136,16 @@ func (r *Provider) Start() {
 		r.logger.Fatal("unable to get ring pop labels", tag.Error(err))
 	}
 
+	// set tchannel port to labels
+	_, port, err := net.SplitHostPort(r.channel.PeerInfo().HostPort)
+	if err == nil {
```
vytautas-karpavicius, line 141: What if `err != nil` ?

common/peerprovider/ringpopprovider/provider.go
```diff
@@ -171,13 +186,53 @@ func (r *Provider) SelfEvict() error {
...
+	memberData := func(member swim.Member) bool {
+		portMap := make(membership.PortMap)
+		if v, ok := member.Label(roleKey); !ok || v != service {
+			return false
+		}
+
+		if v, ok := member.Label(portTchannel); ok {
+			port, err := labelToPort(v)
+			if err != nil {
+				r.logger.Warn("port cannot be converted", tag.Error(err), tag.Value(v))
```
vytautas-karpavicius, line 202: Maybe `tchannel port cannot be converted` and `grpc port cannot be converted` below since there two identical warnings currently, it may not be obvious which one is logged.

vytautas-karpavicius (approved):

common/membership/hashring.go
```diff
@@ -207,12 +201,19 @@ func (r *ring) MemberCount() int {
 }
 
 func (r *ring) Members() []*HostInfo {
```
davidporter-id-au, line 203: Question: Why is this an array of pointers?

mantas-sidlauskas: We use `*HostInfo` in our code base everywhere, so this led to `[]*HostInfo`

I'm about to commit changes to get rid of all pointers to `Hostinfo`

common/membership/resolver_test.go
```diff
@@ -55,24 +55,29 @@ func TestNewCreatesAllRings(t *testing.T) {
 
 func TestMethodsAreRoutedToARing(t *testing.T) {
 	var changeCh = make(chan *ChangedEvent)
-	a, _ := newTestResolver(t)
+	a, pp := newTestResolver(t)
+
 	// add members to this ring
-	r, _ := a.getRing("test-worker")
-	ring := emptyHashring()
+	//ring := emptyHashring()
```
davidporter-id-au, line 61: not used?

davidporter-id-au (approved):

common/membership/resolver_test.go
mantas-sidlauskas: yep, thanks

PR #4713, opened 2022-01-26
Add Watchdog Workflow with Corrupt Workflow Fix

demirkayaender:
**What changed?**
Added a watchdog workflow to auto fix known issues without requiring oncall to interfere

Depends on https://github.com/uber/cadence-idl/pull/99

**Why?**
There are some known issues outside of Cadence's codebase. For example, some deleted records may resurrect when using Cassandra as storage. This causes inconsistency in server, create infinite task retries and eventually causes oncall alerts. The more workflows we run the more likely this will happen. So we should automate the solution to avoid oncall disruptions 

**How did you test it?**
```
go test -v github.com/uber/cadence/service/frontend -run TestAdminHandlerSuite -testify.m TestMaintainCorruptWorkflow*
```
Also with local C*, ESv6 and ESv7

**Potential risks**

**Release notes**

**Documentation Changes**


demirkayaender:
Looks like my repo is actually contaminated with my previous change. I will update this PR when I cleanup

demirkayaender:
this is now ready to early-review

common/persistence/nosql/nosqlplugin/cassandra/visibility.go
```diff
@@ -80,6 +80,21 @@ const (
 		`AND start_time <= ? ` +
 		`AND workflow_id = ? `
 
+	// Intended for admin operations only
+	templateGetOpenWorkflowExecutionsByRunID = `SELECT ` + openExecutionsColumnsForSelect +
```
davidporter-id-au, line 84: with the quotes you should be able to do multiline without using the + for each line I should think. 

Given the `?` I take it this is a prepared statement?

common/persistence/nosql/nosqlplugin/cassandra/visibility.go
```diff
@@ -351,6 +366,33 @@ func (db *cdb) SelectOneClosedWorkflow(
 
 // Noop for Cassandra as it already handle by TTL
 func (db *cdb) DeleteVisibility(ctx context.Context, domainID, workflowID, runID string) error {
+	// Normally we only depend on TTL for Cassandra visibility deletion but
+	// we explicitly delete from open executions when an admin command is issued
+	key := persistence.VisbilityAdminDeletionKey("visibilityAdminDelete")
+	if v := ctx.Value(key); v != nil && v.(bool) {
```
davidporter-id-au, line 372: Question: Is this use of the context object a parameter we're using elsewhere?

common/types/admin.go
```diff
@@ -473,6 +473,35 @@ func (v *RestoreDynamicConfigRequest) GetFilters() (o []*DynamicConfigFilter) {
 	return
 }
 
+// AdminDeleteWorkflowRequest is an internal type (TBD...)
+type AdminDeleteWorkflowRequest struct {
+	Domain     string             `json:"domain,omitempty"`
```
vytautas-karpavicius, line 478: You should drop json annotations here. AFAICT this is not serialized directly.
We should come with a separate PR to clean those up on other types as well eventually.

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -112,6 +112,12 @@ service AdminAPI {
 
   // ListDynamicConfig returns all dynamic config parameters and associated values.
   rpc ListDynamicConfig(ListDynamicConfigRequest) returns (ListDynamicConfigResponse);
+
+  // DeleteWorkflow is used to delete a workflow with given domainName, workflowID and runID
+  rpc DeleteWorkflow(AdminDeleteWorkflowRequest) returns (DeleteWorkflowResponse);
+
+  // MaintainCorruptWorkflow is used to delete a workflow with given domainName, workflowID and runID
```
vytautas-karpavicius, line 119: Looking at the comment alone this seems identical to `DeleteWorkflow`. If it is different maybe mention if the comment how it it is different?

service/frontend/adminHandler.go
```diff
@@ -314,6 +327,239 @@ func (adh *adminHandlerImpl) RemoveTask(
...
+			_, err := client.GetWorkflowExecutionHistory(ctx, &types.GetWorkflowExecutionHistoryRequest{
+				Domain:    request.Domain,
+				Execution: request.Execution,
+			})
+			return err
+		},
+	}
+}
+
+func (adh *adminHandlerImpl) MaintainCorruptWorkflow(
```
vytautas-karpavicius, line 352: Could you explain why `MaintainCorruptWorkflow` is exposed via admin API?
It seems that with the introduction of `DeleteWorkflow` API we should have enough primitives to construct such logic in the CLI alone. Or is it planned to be used in some other way as well?
I'm bringing this up because, IMHO our API surface is already huge. It would be nice if we could contain it to smaller primitive building blocks, that could be used in higher level tools on top.

common/persistence/nosql/nosqlplugin/cassandra/visibility.go
davidporter-id-au: fair!

service/frontend/adminHandler.go
demirkayaender: MaintainCorruptWorkflow will have 2 purposes:
- It's used by the corrupt workflow cleaning system workflow
- It's moving the internal dlq-fix logic to open source so the community can use

service/history/task/task.go
```diff
@@ -224,6 +242,7 @@ func (t *taskImpl) HandleErr(
 				t.scope.RecordTimer(metrics.TaskAttemptTimerPerDomain, time.Duration(t.attempt))
 				t.logger.Error("Critical error processing task, retrying.",
 					tag.Error(err), tag.OperationCritical, tag.TaskType(t.GetTaskType()))
+				t.ReportCorruptWorkflowToWatchDog()
```
emrahs, line 245: If the watchdog isn't able to repair in a couple of retries, its chances of succeeding later are too slim, so we should back off from retrying this too many times in order to free up the workflow and the DB. Calling this for each attempt after the `criticalRetryCount()` is too much. Perhaps run it for every N multiples of the `criticalRetryCount` to retry more than once but less than a reasonable limit.

service/worker/watchdog/workflow.go
```diff
@@ -0,0 +1,169 @@
...
+		_ = workflow.ExecuteActivity(opt, handleCorrputedWorkflowActivity, request).Get(ctx, nil)
+	}
+}
+
+func (w *Workflow) getProcessedID(request *CorruptWFRequest) string {
+	return fmt.Sprintf("%s-%s", request.Workflow.WorkflowID, request.Workflow.RunID)
+}
+
+// handleCorrputedWorkflowActivity is activity to handle corrupted workflows in DB
+func (w *Workflow) handleCorrputedWorkflow(ctx context.Context, request *CorruptWFRequest) error {
```
emrahs, line 130: nit: typo in function name

service/worker/watchdog/workflow.go
```diff
@@ -0,0 +1,169 @@
...
+	_, err = adminClient.MaintainCorruptWorkflow(ctx, maintainWFRequest)
+	if err != nil {
+		return err
+	}
+
+	tagged.AddCounter(metrics.WatchDogNumCorruptWorkflowProcessed, 1)
+
+	// We couldn't decide if we should delete this workflow. That means the same logs will continue
+	// and the workflow will keep getting signals. So adding workflow to a cache to skip processing
+	w.processed.Put(w.getProcessedID(request), struct{}{})
```
emrahs, line 166: I didn't quite get this part. Could you elaborate a bit more?

service/worker/watchdog/client.go
```diff
@@ -0,0 +1,80 @@
...
+	signalCtx, cancel := context.WithTimeout(context.Background(), SignalTimeout)
+	defer cancel()
+	request := CorruptWFRequest{
+		DomainName: domainName,
+		Workflow: types.WorkflowExecution{
+			WorkflowID: workflowID,
+			RunID:      runID,
+		},
+	}
+	return c.cadenceClient.SignalWorkflow(signalCtx, WatchdogWFID, "", CorruptWorkflowWatchdogChannelName, request)
```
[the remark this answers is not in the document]
demirkayaender: Good point! I can actually move the LRU cache here. That way we wouldn't spam the signal API.

service/worker/watchdog/workflow.go
demirkayaender: This was to avoid spamming with retries. I will move this to the client (mentioned above).

demirkayaender (commented):
Thanks for the review. Replied some of the comments, will address the rest.

service/frontend/adminHandler.go
```diff
@@ -1391,10 +1660,10 @@ func (adh *adminHandlerImpl) ListDynamicConfig(ctx context.Context, request *typ
...
 }
 
-func checkValidKey(keyName string) (dynamicconfig.Key, error) {
-	keyVal, ok := dynamicconfig.KeyNames[keyName]
-	if !ok || keyVal == dynamicconfig.UnknownKey {
-		return dynamicconfig.UnknownKey, errors.New("invalid dynamic config parameter name")
+func checkValidKey(keyName string) (dc.Key, error) {
+	keyVal, ok := dc.KeyNames[keyName]
+	if !ok || keyVal == dc.UnknownKey {
+		return dc.UnknownKey, errors.New("invalid dynamic config parameter name")
```
mantas-sidlauskas, line 1666: nit: not directly related to this change, but it is always good to know, what exactly was passed, we can improve error message like this:
```suggestion
		return dc.UnknownKey, fmt.Errorf("invalid dynamic config parameter name: %q")
```

service/frontend/adminHandler.go
```diff
@@ -314,6 +327,259 @@ func (adh *adminHandlerImpl) RemoveTask(
...
+
+	} else {
+		logger.Error("Describe workflow failed", tag.Error(err))
+		if !skipErrors {
+			return nil, adh.error(err, scope)
+		}
+	}
+
+	return &types.AdminDeleteWorkflowResponse{
+		Deleted: deletedFromExecutions && deletedFromHistory && deletedFromVisibility,
```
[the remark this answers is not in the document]
mantas-sidlauskas: We can extend response here and add additional fields with the overall result:
`
&types.AdminDeleteWorkflowResponse{
		RemovedExecution: bool,
                RemovedHistory: bool,
                RemovedVisibility: bool,

                Success: bool
}
`

service/worker/watchdog/watchdog.go
```diff
@@ -0,0 +1,124 @@
...
+		logger:             logger,
+		scopedMetricClient: getScopedMetricsClient(metricsClient),
+		tallyScope:         tallyScope,
+		resource:           resource,
+		domainCache:        domainCache,
+		config:             config,
+	}
+}
+
+func getScopedMetricsClient(metricsClient metrics.Client) metrics.Scope {
```
mantas-sidlauskas, line 93: is this function needed?

service/worker/watchdog/workflow.go
```diff
@@ -0,0 +1,169 @@
...
+	CorruptWorkflowWatchdogChannelName = "CorruptWorkflowWatchdogChannelName"
+)
+
+type (
+	Workflow struct {
+		watchdog  *WatchDog
+		processed cache.Cache
+	}
+
+	// Request defines the request for corruptWorkflow maintenance
```
mantas-sidlauskas, line 58: not: comment

service/worker/watchdog/workflow.go
```diff
@@ -0,0 +1,169 @@
...
+	handleCorruptWorkflowOptions = workflow.ActivityOptions{
+		ScheduleToStartTimeout: time.Minute,
+		StartToCloseTimeout:    5 * time.Minute,
+		RetryPolicy:            &retryPolicy,
+	}
+
+	wfOptions = cclient.StartWorkflowOptions{
+		ID:                           WatchdogWFID,
+		TaskList:                     taskListName,
+		WorkflowIDReusePolicy:        cclient.WorkflowIDReusePolicyTerminateIfRunning,
```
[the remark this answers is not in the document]
demirkayaender: My main goal was to restart the workflow at each deployment. Not sure if there's a better way to do it.

emrahs (approved):

PR #4718, opened 2022-01-31
Drop pointers on internal history branch types

vytautas-karpavicius:
**What changed?**
- Drop pointers with internal types: `types.HistoryBranch` and `types.HistoryBranchRange`
- Drop json annotations for them. Those types are never serialized directly.
- Drop helper functions for them. We can use fields directly once pointers are gone.
- Within persistence layer reuse `thrift.[To/From]HistoryBranchRangeArray`
- Within thrift mapper deserialize fields directly with `*` (no zero values fallback). I did this to preserve current functionality - we would get NPE for those cases as it is currently. Although they should never be nil.

**Why?**
To simplify and drop unused code

**How did you test it?**
Existing tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/types/shared.go
+++ b/common/types/shared.go
@@ -3646,64 +3646,16 @@ func (v *History) GetEvents() (o []*HistoryEvent) {
 
 // HistoryBranch is an internal type (TBD...)
 type HistoryBranch struct {
-	TreeID    *string               `json:"treeID,omitempty"`
-	BranchID  *string               `json:"branchID,omitempty"`
-	Ancestors []*HistoryBranchRange `json:"ancestors,omitempty"`
-}
-
-// GetTreeID is an internal getter (TBD...)
-func (v *HistoryBranch) GetTreeID() (o string) {
-	if v != nil && v.TreeID != nil {
-		return *v.TreeID
-	}
-	return
-}
-
-// GetBranchID is an internal getter (TBD...)
-func (v *HistoryBranch) GetBranchID() (o string) {
-	if v != nil && v.BranchID != nil {
-		return *v.BranchID
-	}
-	return
-}
-
-// GetAncestors is an internal getter (TBD...)
-func (v *HistoryBranch) GetAncestors() (o []*HistoryBranchRange) {
-	if v != nil && v.Ancestors != nil {
-		return v.Ancestors
-	}
-	return
+	TreeID    string
+	BranchID  string
+	Ancestors []*HistoryBranchRange
 }
 
 // HistoryBranchRange is an internal type (TBD...)
 type HistoryBranchRange struct {
-	BranchID    *string `json:"branchID,omitempty"`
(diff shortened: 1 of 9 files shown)


common/types/shared.go
```diff
@@ -3646,64 +3646,16 @@ func (v *History) GetEvents() (o []*HistoryEvent) {
 
 // HistoryBranch is an internal type (TBD...)
 type HistoryBranch struct {
-	TreeID    *string               `json:"treeID,omitempty"`
-	BranchID  *string               `json:"branchID,omitempty"`
-	Ancestors []*HistoryBranchRange `json:"ancestors,omitempty"`
```
davidporter-id-au, line 3651: Is it ok to drop the JSON tags? I dont' know if this data is persisted, but I'd be worried that if it is, that'd be a problem

common/types/shared.go
```diff
@@ -3646,64 +3646,16 @@ func (v *History) GetEvents() (o []*HistoryEvent) {
...
-	}
-	return
+	TreeID    string
+	BranchID  string
+	Ancestors []*HistoryBranchRange
 }
 
 // HistoryBranchRange is an internal type (TBD...)
 type HistoryBranchRange struct {
-	BranchID    *string `json:"branchID,omitempty"`
```
davidporter-id-au, line 3680: ^ as before

common/types/shared.go
vytautas-karpavicius: It is safe. As mentioned in the description of PR, these types are not serialized directly.
We do serialize thrift counterparts in several places, but not those types.

common/types/shared.go
```diff
@@ -3646,64 +3646,16 @@ func (v *History) GetEvents() (o []*HistoryEvent) {
...
-
-// GetAncestors is an internal getter (TBD...)
-func (v *HistoryBranch) GetAncestors() (o []*HistoryBranchRange) {
-	if v != nil && v.Ancestors != nil {
-		return v.Ancestors
-	}
-	return
+	TreeID    string
+	BranchID  string
+	Ancestors []*HistoryBranchRange
```
Groxx, line 3651: I don't really think it should delay this PR, but: can we drop the pointers here?  I doubt a nil item in the slice is desirable.
(unless it is.  then LGTM)

Groxx (approved):
better types is always nice :)  and changes seem nicely straightforward.

maybe one day we can make (more of) these uuids into uuids...

common/types/shared.go
davidporter-id-au: Sorry, 🤦  didn't read

davidporter-id-au (approved):

common/types/shared.go
vytautas-karpavicius: In general I agree, but this would end up touching a lot more code, therefore I will keep it for now.

PR #4722, opened 2022-02-01
Fix parsing domain_id in child_info_maps for backward compatibility

Shaddoll:
**What changed?**
If domain_id from child_info_maps is empty UUID, we convert it to an empty string.

**Why?**
For backward compatibility. null uuids loaded from DB are converted to empty UUID, which is a breaking change for us, because the domain_id field is a new field.

**How did you test it?**
existing test, verify with child workflows created from 0.23.2 release

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx (approved):
yeah, string comparison may actually be preferable to comparing values here - we do not have or want any zero-valued UUIDs in the database.

PR #4725, opened 2022-02-03
Update cadence batch command to receive more input

Shaddoll:
**What changed?**
Update cadence batch command to receive more input

**Why?**
To make batch activity more configurable
Currently, if we have lots of workflows to process in the batch activity, the activity frequently times out and we have no way to change the default heart beat timeout via CLI.

**How did you test it?**
manual local test

**Potential risks**

**Release notes**

**Documentation Changes**


mkolodezny (approved):
lgtm, assuming the new flags are optional

PR #4726, opened 2022-02-04
Drop redundant pointer conversions

vytautas-karpavicius:
**What changed?**
Search for pattern `Ptr(*` and dropped redundant conversions.

**Why?**
Redundant.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


mkolodezny (approved):

Groxx (approved):
strictly speaking this may not be a safe change, since before this was making copies of the values... but I sincerely doubt we're modifying any of these input-int-pointers.

PR #4728, opened 2022-02-04
Drop unused flags for cli rereplicate command

vytautas-karpavicius:
**What changed?**
Dropped unused flags for rereplicate command in CLI.
- `FlagNumberOfShards` not used anywhere. It is annoying as it requires this flag to be present and > 0.
- All database flags. This command does not connect to DB directly. It uses admin client.

**Why?**
Cleanup.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/adminKafkaCommands.go
+++ b/tools/cli/adminKafkaCommands.go
@@ -479,11 +479,6 @@ func doRereplicate(
 
 // AdminRereplicate parses will re-publish replication tasks to topic
 func AdminRereplicate(c *cli.Context) {
-	numberOfShards := c.Int(FlagNumberOfShards)
-	if numberOfShards <= 0 {
-		ErrorAndExit("numberOfShards is must be > 0", nil)
-		return
-	}
 	sourceCluster := getRequiredOption(c, FlagSourceCluster)
 
 	adminClient := cFactory.ServerAdminClient(c)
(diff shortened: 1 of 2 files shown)


tools/cli/adminKafkaCommands.go
```diff
@@ -479,11 +479,6 @@ func doRereplicate(
 
 // AdminRereplicate parses will re-publish replication tasks to topic
 func AdminRereplicate(c *cli.Context) {
-	numberOfShards := c.Int(FlagNumberOfShards)
```
mkolodezny, line 482: maybe make it optional (default=16k) here https://github.com/uber/cadence/blob/af307533b00b050cf62ced0f788697328ec5ac14/tools/cli/adminCommands.go#L395-L394

mkolodezny (approved):

Groxx:
Thank you!  I was just wondering about this command recently, glad it's less unnecessarily complicated-looking now 😄

tools/cli/adminKafkaCommands.go
vytautas-karpavicius: I think we should leave it explicit. If user configures cluster with different number of shards, such silent default can result in very unexpected results.

PR #4731, opened 2022-02-08
Protect membership member keys concurrent access

mantas-sidlauskas:
**What changed?**
Adding read lock for member keys

**Why?**
There was a race condition when `Lookup` is called during periodic member `Refresh`

**How did you test it?**
Wrote a separate test case to be checked with -race flag 

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/membership/hashring.go
+++ b/common/membership/hashring.go
@@ -68,7 +68,7 @@ type ring struct {
 	value atomic.Value // this stores the current hashring
 
 	members struct {
-		sync.Mutex
+		sync.RWMutex
 		refreshed time.Time
 		keys      map[string]HostInfo // for mapping ip:port to HostInfo
 	}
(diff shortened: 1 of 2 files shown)


vytautas-karpavicius:
Not directly related to your current changes, but can `members, err := r.peerProvider.GetMembers(r.service)` in refresh be time consuming?
Maybe it is better to call it before taking the write lock?

vytautas-karpavicius (approved):

mantas-sidlauskas:
> Not directly related to your current changes, but can `members, err := r.peerProvider.GetMembers(r.service)` in refresh be time consuming? Maybe it is better to call it before taking the write lock?

Good call. If peer provider is not keeping members in-memory and is doing network call, this might be a problem. Will address this in a separate PR

PR #4743, opened 2022-02-21
Add rate limits for inline archival

vytautas-karpavicius:
**What changed?**
Added dynamic rate limiting for inline archival (both history and visibility). Added metrics to record throttled count when exceeding rate limit.

**Why?**
Bursty workflow deletion & archival may put load on the underlying archival provider when done inline. Added additional measures to limit RPS (with global option as archival provider is a shared resource). If such limit is reached inline archival will not even be tried and will be delegated to archival workflow instead.

Default values are intentionally set high, to not affect current load. Only to be tightened if needed.

**How did you test it?**
Additional unit tests to test for inline throttling and metric emission.

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (commented):
What is the default behaviour for systems where the configuration for this feature is not explicitly turned on?

davidporter-id-au:
> What is the default behaviour for systems where the configuration for this feature is not explicitly turned on?

Sorry, I didn't read the description closely enough, you've already answered. lgtm!

davidporter-id-au (approved):

PR #4749, opened 2022-02-23
Use named port to select transport for outbound calls

mantas-sidlauskas:
**What changed?**
AddressMapper is changed to named port setting.

Some additional changes in this PR:

* `CreateShardOwnershipLostError` is returning `TChannel` address for backwards compatibility
* `LookupByAddress` returns `membership.HostInfo` which is handy to find peer based on IP:port without knowing transport details
* Use `zaptest` as logger in unit tests

**Why?**
Named port (`tchannel` or `grpc`) will now define what transport is to be used for outbound calls.

**How did you test it?**
Unit tests, integration tests

**Potential risks**

**Release notes**

**Documentation Changes**


host/onebox.go
```diff
@@ -774,30 +795,34 @@ func newPublicClient(dispatcher *yarpc.Dispatcher) cwsc.Interface {
 	)
 }
 
-func (c *cadenceImpl) newRPCFactory(serviceName string, tchannelHostPort string) common.RPCFactory {
-	grpcPortResolver := grpcPortResolver{}
-	grpcHostPort, err := grpcPortResolver.GetGRPCAddress("", tchannelHostPort)
+func (c *cadenceImpl) newRPCFactory(serviceName string, host membership.HostInfo) common.RPCFactory {
+	tchannelAddres, err := host.GetNamedAddress(membership.PortTchannel)
```
vytautas-karpavicius, line 799: nit: typo `tchannelAddres`

client/matching/peerResolver.go
```diff
@@ -29,22 +29,25 @@ import (
...
+		namedPort: namedPort,
+		service:   service.Matching,
+	}
 }
 
 // FromTaskList resolves the matching peer responsible for the given task list name.
 // It uses our membership provider to lookup which instance currently owns the given task list.
 // FromHostAddress is used for further resolving.
 func (pr PeerResolver) FromTaskList(taskListName string) (string, error) {
-	host, err := pr.resolver.Lookup(service.Matching, taskListName)
```
vytautas-karpavicius, line 47: Any reason why this is a struct member now, instead of using `service.Matching` like before? It is still initialized to the same const in the constructor.

common/membership/hostinfo.go
```diff
@@ -37,15 +39,19 @@ type PortMap map[string]uint16
...
+	addr     string // ip:port returned by peer provider
+	ip       string // @todo should we set this to net.IP ?
 	identity string
 	portMap  PortMap // ports host is listening to
 }
 
 // NewHostInfo creates a new HostInfo instance
 func NewHostInfo(addr string) HostInfo {
+	ip, _, _ := net.SplitHostPort(addr)
+	net.ParseIP(addr)
```
vytautas-karpavicius, line 51: This is line is not need if we are not using net.IP

common/membership/hostinfo.go
```diff
@@ -72,6 +81,39 @@ func (hi HostInfo) GetAddress() string {
 	return hi.addr
 }
 
+// GetNamedAddress returns the ip:port address
+func (hi HostInfo) GetNamedAddress(port string) (string, error) {
+	if port, set := hi.portMap[port]; set {
+		return fmt.Sprintf("%s:%d", hi.ip, port), nil
```
vytautas-karpavicius, line 87: I suggest using https://pkg.go.dev/net#JoinHostPort here. It is symmetrical to our usage `net.SplitHostPort` and also plays nicely with IPv6

common/rpc/factory.go
```diff
@@ -37,16 +37,15 @@ const defaultGRPCSizeLimit = 4 * 1024 * 1024
...
+	maxMessageSize int
+	channel        tchannel.Channel
+	dispatcher     *yarpc.Dispatcher
 }
 
 // NewFactory builds a new rpc.Factory
 func NewFactory(logger log.Logger, p Params) *Factory {
 	inbounds := yarpc.Inbounds{}
-
+	logger = logger.WithTags(tag.Service(p.ServiceName))
```
vytautas-karpavicius, line 48: I believe we tag service name here already: https://github.com/uber/cadence/blob/7084679d54d983c7667d319ac3321c97d61d3e3f/common/resource/params.go#L74

Although I don't particularly like such tagging within Update... function.

common/membership/hostinfo.go
mantas-sidlauskas: +1

common/rpc/factory.go
mantas-sidlauskas: removed `UpdateLoggerWithServiceName` and logger tagging in rpc factory

PR #4751, opened 2022-02-24
Staticcheck cleanup

vytautas-karpavicius:
**What changed?**
Cleaned up lots of code according to `staticcheck` suggestions. Each commits deals with one suggestion at a time. I have included suggestion code which can be looked up here: https://staticcheck.io/docs/checks

Those changes are mostly mechanical. There are more warning the tool reports, that may affect functionality. I will try to address those separately.

**Why?**
To clean up our codebase and make more consistent with best go practices.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/archiver/gcloud/queryParser.go
```diff
@@ -233,21 +231,6 @@ func (p *queryParser) convertComparisonExpr(compExpr *sqlparser.ComparisonExpr,
...
-		}
-		if err := p.convertCloseTime(timestamp, "<=", parsedQuery); err != nil {
-			return err
-		}
-	default:
-		return fmt.Errorf("operator %s is not supported for close time", op)
-	}
-	return nil
-}
-
```
Groxx, line 250: unused or something?  afaik this is correct for BigQuery-style datastores

common/archiver/s3store/historyArchiver.go
```diff
@@ -68,7 +68,6 @@ type (
 		s3cli     s3iface.S3API
 		// only set in test code
 		historyIterator archiver.HistoryIterator
-		config          *config.S3Archiver
```
Groxx, line 71: apparently this config is now contained in the `s3cli`, so 👍

common/archiver/util.go
```diff
@@ -30,7 +30,7 @@ import (
 
 var (
 	errEmptyDomainID         = errors.New("DomainID is empty")
-	errEmptyDomainName       = errors.New("Domain name is empty")
+	errEmptyDomainName       = errors.New("DomainName is empty")
```
Groxx, line 33: staticcheck complained about this?  or did it just complain about uppercase -> now it matches the param name, and staticcheck is happy because it looks like a ProperNoun or something?

common/config/metrics.go
```diff
@@ -116,7 +116,13 @@ func (c *Metrics) newStatsdScope(logger log.Logger) tally.Scope {
 	if len(config.HostPort) == 0 {
 		return tally.NoopScope
 	}
-	statter, err := statsd.NewBufferedClient(config.HostPort, config.Prefix, config.FlushInterval, config.FlushBytes)
+	statter, err := statsd.NewClientWithConfig(&statsd.ClientConfig{
```
Groxx, line 119: ah.  deprecated.  👍

common/elasticsearch/esql/globals.go
```diff
@@ -85,11 +83,3 @@ const (
...
 	StartTimeOrder       = "desc"
 )
-
-// DEBUG usage
-//nolint
-func print(v interface{}) {
-	fmt.Println("==============")
-	fmt.Println(v)
-	fmt.Println("==============")
-}
```
Groxx, line 95: honestly I wonder if we can find a linter to just ban `fmt.Print*` across the board.  it's basically always better-served by some injected thing, like a logger

common/log/loggerimpl/logger_test.go
```diff
@@ -49,8 +49,7 @@ func TestDefaultLogger(t *testing.T) {
 		outC <- buf.String()
 	}()
 
-	var zapLogger *zap.Logger
-	zapLogger = zap.NewExample()
+	zapLogger := zap.NewExample()
```
Groxx, line 52: would it cause any problems to change these to `zaptest.New(t)` instead?  then logs will be scoped to the test.

common/persistence/dataStoreInterfaces.go
```diff
@@ -716,7 +716,7 @@ type (
 		CloseTimestamp     time.Time
 		Status             types.WorkflowExecutionCloseStatus
 		HistoryLength      int64
-		RetentionSeconds   time.Duration
+		RetentionPeriod    time.Duration
```
Groxx, line 719: minor honestly, but do we have any broad pattern for this kind of thing for "period" vs "interval"?  since it's a bit odd to do it two ways in the same file.

(not that I think either is _wrong_, they kinda are different meanings.  just wondering out loud if consistency would be better.)

common/persistence/sql/sqlExecutionStoreUtil.go
```diff
@@ -1224,54 +1224,6 @@ func assertRunIDAndUpdateCurrentExecution(
...
-				"assertAndUpdateCurrentExecution failed. Current state %v, expected %v",
-				currentRow.State,
-				previousState,
-			)}
-		}
-		return nil
-	}
-	if err := assertCurrentExecution(ctx, tx, shardID, domainID, workflowID, assertFn); err != nil {
-		return err
-	}
```
Groxx, line 1270: since this does a DB lock: is this something we _should_ be calling, or is it being handled somewhere else?

common/persistence/sql/sqlplugin/postgres/db.go
```diff
@@ -61,17 +61,11 @@ func (pdb *db) IsDupEntryError(err error) bool {
 }
 
 func (pdb *db) IsNotFoundError(err error) bool {
-	if err == sql.ErrNoRows {
-		return true
-	}
-	return false
+	return err == sql.ErrNoRows
```
Groxx, line 64: since we're working on wrapped errors: are you on a recent master?  I'm not sure what the sequencing of these changes is.  since this'd be a fairly obvious `errors.Is(err, sql.ErrNoRows)` transformation.
(but probably not worth doing in this PR, keep this one full of obviously-noops)

common/testing/event_generator.go
```diff
@@ -356,12 +349,6 @@ func (g *EventGenerator) pickRandomVertex(
 	return endVertex.DeepCopy()
 }
 
-func (g *EventGenerator) shouldBumpVersion() bool {
-	// 1//1000 to bump the version
-	//return g.dice.Intn(1000) == 500
-	return false
-}
-
```
Groxx, line 364: huh.  wonder if this would be useful to enable in tests.  (but if unused: 👍)

host/integration_test.go
```diff
@@ -1613,12 +1613,12 @@ func (s *IntegrationSuite) TestRateLimitBufferedEvents() {
 			// Buffered Signals
 			for i := 0; i < 100; i++ {
 				buf := new(bytes.Buffer)
-				binary.Write(buf, binary.LittleEndian, i)
+				binary.Write(buf, binary.LittleEndian, int32(i))
```
Groxx, line 1616: any reason for int32 vs int64?  I think this was behaving as int64, as int is architecture-dependent and basically nobody uses 32-bit for servers

host/reset_workflow_test.go
```diff
@@ -404,7 +404,7 @@ func (s *IntegrationSuite) TestResetWorkflow_NoDecisionTaskCompleted() {
 	s.Logger.Info("PollAndProcessWorkflowTask", tag.Error(err))
 	s.NoError(err)
 
-	events = s.getHistory(s.domainName, &types.WorkflowExecution{
+	s.getHistory(s.domainName, &types.WorkflowExecution{
```
Groxx, line 407: With errcheck (or similar), I guess this is probably preferable over `_ = s.getHistory(...)`, isn't it... since that could be ignoring an error, but `s.getHistory(...)` cannot.  Interesting.

host/xdc/integration_failover_test.go
```diff
@@ -1,2054 +0,0 @@
...
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-
-//go:build !race
-// +build !race
```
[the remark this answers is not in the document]
Groxx: that aside: could that be marking this as unused?  and/or is there anything worth saving in here?  we should just fix the races.

service/frontend/clusterRedirectionHandler_test.go
```diff
@@ -708,6 +708,7 @@ func (s *clusterRedirectionHandlerSuite) TestRespondQueryTaskCompleted() {
 	token, err := s.handler.tokenSerializer.SerializeQueryTaskToken(&common.QueryTaskToken{
 		DomainID: s.domainID,
 	})
+	s.Nil(err)
```
Groxx, line 711:
```suggestion
	s.NoError(err)
```
is probably preferable

service/history/historyEngine2_test.go
```diff
@@ -848,6 +848,7 @@ func (s *engine2Suite) createExecutionStartedState(we types.WorkflowExecution, t
 }
 
 //nolint:unused
+//lint:ignore U1000 for printing within tests
```
Groxx, line 851: huh.  just because of the name?  maybe better to just change the name...

service/matching/taskListManager_test.go
```diff
@@ -329,6 +329,6 @@ func TestAddTaskStandby(t *testing.T) {
 
 	addTaskParam.forwardedFrom = "from child partition"
 	syncMatch, err = tlm.AddTask(context.Background(), addTaskParam)
-	require.Error(t, errRemoteSyncMatchFailed) // should not persist the task
+	require.Error(t, err) // should not persist the task
```
Groxx, line 332: lol.  well that's a more useful test now

service/worker/scanner/data_corruption_workflow_test.go
```diff
@@ -43,6 +43,8 @@ import (
 
 const testWorkflowName = "default-test-workflow-type-name"
 
+var validBranchToken = []byte{89, 11, 0, 10, 0, 0, 0, 12, 116, 101, 115, 116, 45, 116, 114, 101, 101, 45, 105, 100, 11, 0, 20, 0, 0, 0, 14, 116, 101, 115, 116, 45, 98, 114, 97, 110, 99, 104, 45, 105, 100, 0}
```
Groxx, line 46: 👍👍 for more valid values, though is there a more human-friendly way to build this?  as it apparently includes "test-tree-id" and "test-branch-id" strings

tools/cli/util.go
```diff
@@ -714,7 +714,7 @@ func parseTime(timeStr string, defaultValue int64) int64 {
 func parseTimeRange(timeRange string) (time.Time, error) {
 	match, err := regexp.MatchString(defaultDateTimeRangeShortRE, timeRange)
 	if !match { // fallback on to check if it's of longer notation
-		match, err = regexp.MatchString(defaultDateTimeRangeLongRE, timeRange)
+		_, err = regexp.MatchString(defaultDateTimeRangeLongRE, timeRange)
 	}
```
Groxx, line 718: not necessarily relevant as this doesn't change behavior, but:  what if the second doesn't match?  looks like `regexp.MatchString` will only error if the regex fails to compile.

more generally, it seems pretty obvious that we should only be compiling these once, rather than on every call :\  think that's worth doing here, or later?

Groxx:
notes are worth checking, but LGTM either way, and they're nearly all just observations / questions.  Let me know if you'd like me to check any changes you do end up making :)

Groxx (approved):

common/archiver/gcloud/queryParser.go
vytautas-karpavicius: Yeah this is unused.

common/archiver/util.go
vytautas-karpavicius: `staticcheck` complained about upper case. Since other errors here include exact param names, I changed this for consistency as well.

Out of curiosity checked their source code: https://github.com/dominikh/go-tools/blob/f41ebe766a412094c088d7d32598c19442db66c1/stylecheck/lint.go#L418

Seems like if it detects more upper letters in the first word - it is ok:
```
// Word is probably an initialism or
// multi-word function name
```

common/elasticsearch/esql/globals.go
vytautas-karpavicius: Quickly searched for such usages. Majority of the comes in CLI or in tests. Those are probably valid uses.

common/log/loggerimpl/logger_test.go
vytautas-karpavicius: Tried changing this - tests fail.

I think `zaptest.New(t)` is good for places when you just want to pass some logger when testing things and do no pollute output for passed tests. There are some places that could be updated to use it.

Here however, we testing the logger implementation itself.

common/persistence/dataStoreInterfaces.go
vytautas-karpavicius: Yeah I thought about it.

But if we decide to do it, we should change it everywhere. `Period` mostly goes with workflow retention, `Interval` is very common for retry policy fields. We also have lots of `Timeout` fields as well.

This would be big change. I'm not against it, but IMHO this is very low priority item.

host/integration_test.go
vytautas-karpavicius: Good point, apparently I'm living in the past :)
Updated.

service/worker/scanner/data_corruption_workflow_test.go
vytautas-karpavicius: Copied it from some other place, as the test actually started failing after checking for errors :)

This is thrift serialized structure underneath. I guess we could have helpers in our test to generate those.

tools/cli/util.go
vytautas-karpavicius: Good observation. Few problems here:
- regexes are not compiled as you mentioned (although not very impactful as those are used only on CLI commands)
- these matches are not used anywhere. Error simply check whether they compile.
- Match groups in those regexes are not utilized, we parse things manually below

I think I will address this in a separate PR. Also checked that there exists a library for such functionality: https://github.com/karrick/tparse
What is opinion on taking this dependency? Or just simply fixing is enough?

service/history/historyEngine2_test.go
vytautas-karpavicius: This is currently not used anywhere, so we could remove it entirely.
Since I saw previous `nolint` I assumed this may be sometimes used for debugging in tests, so added additional declaration for staticcheck.
Not sure whether it is worth keeping such helpers though...

host/xdc/integration_failover_test.go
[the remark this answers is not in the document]
vytautas-karpavicius: This entire test suite is unused and disabled for 10 months already.
AFAICT we have another similar suite in `host/ndc`. Might be worth double checking with the team.

common/persistence/sql/sqlplugin/postgres/db.go
vytautas-karpavicius: This is on recent master.
Yeah, lets leave it a separate PR, that addresses such cases.

PR #4761, opened 2022-03-03
Added history corruption checks for workflow signaling

vytautas-karpavicius:
**What changed?**
Added history corruption check when signaling workflows.
It ensures that the workflow being signaled has the first event in the history. Otherwise:
- an error is returned for `SignalWorkflowExecution`
- new workflow is started for `SignalWithStartWorkflowExecution`

**Why?**
After some analysis for common messages in DLQ, it appears that many of them fail with `Resend events due to missing mutable state` error. Even resending history does not help.

Checking workflow in the source cluster reveals, that is contains corrupted history. Basically only signal events stored. No starting event.

I do not have exact explanation why it is the case. One speculation is that execution record is resurrected (in Cassandra DB), and some customers signal workflow long after their retention period. In normal case (at least for SignalWithStart) simply new workflow would be created. But since we have resurrected record, it simply updates it and records that one (or more) signal events. We do not check for history validity at that point.

So the idea here is to try checking whether we have a first event recorded. If not simply reject the signal or in signalWithStart case start the new one instead.

**How did you test it?**
Tested locally with sample [signalcounter](https://github.com/uber-common/cadence-samples/tree/master/cmd/samples/recipes/signalcounter):
- Started workflow
- Signaled workflow - works
- Deleted all entries from history_node table for that tree_id
- Restarted server (to clean events cache)
- Signaled workflow - returns EntityNotFound error
- SignalWithStart - creates new run
- Signaled workflow - works

**Potential risks**
To minimize risk, wrapping this with a feature flag (off by default), and planning to enable it incrementally per domain basis.

**Release notes**

**Documentation Changes**


service/history/historyEngine.go
```diff
@@ -2407,6 +2425,32 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(
 	)
 }
 
+func (e *historyEngineImpl) checkForHistoryCorruptions(ctx context.Context, mutableState execution.MutableState) error {
+	domainName := mutableState.GetDomainEntry().GetInfo().Name
+	if !e.config.EnableHistoryCorruptionCheck(domainName) {
```
davidporter-id-au, line 2430: I know this is demonstration code only, but perhaps a log/metric would help the rollout of this feature. If it's not enabled, at least we could get a rough idea of the number of events.

davidporter-id-au (approved):

service/history/historyEngine.go
vytautas-karpavicius: We already have a rough estimates, as we see those errors once events get replicated and put into DLQ.

PR #4773, opened 2022-03-18
Added presentation layer for rendering workflow list tables

vytautas-karpavicius:
**What changed?**
Added generic table rendering function `RenderTable`. It accepts a slice of structs annotated with header names and potentially other tags relevant for presentation. It knows how to render different types (including our internal types such as types.Memo or types.SearchAttributes).

Refactored workflow list commands to use generic table rendering instead of constructing tables along the way.

Refactored workflow list fetch functions to return pagination function which can be composed in more flexible ways.

**Why?**
Currently CLI commands are sprinkled with table rendering logic which makes it difficult add additional output formats or pipe cadence commands together.

This is part of a larger effort to separate CLI logic from presentation layer.

**How did you test it?**
- Added new unit tests for rendering tables.
- Tested locally that commands work as before:
  - `worfklow list`
  - `workflow list --more`
  - `workflow listall`
  - `workflow scan`
  - `workflow listarchived`
  - `workflow listarchived --all`

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/table.go
+++ b/tools/cli/table.go
@@ -43,8 +43,8 @@ type TableOptions struct {
 	// Border specified whether to render table border
 	Border bool
 
-	// NoColor will not use color characters while printing table
-	NoColor bool
+	// Color will use coloring characters while printing table
+	Color bool
 
 	// PrintRawTime will print time as int64 unix nanos
 	PrintRawTime bool
@@ -56,7 +56,7 @@ type TableOptions struct {
 func RenderTable(w io.Writer, slice interface{}, opts TableOptions) {
 	sliceValue := reflect.ValueOf(slice)
 	if sliceValue.Kind() != reflect.Slice {
-		panic("table must be a slice")
+		panic(fmt.Errorf("table must be a slice, provided: %s", sliceValue.Kind()))
 	}
 
 	// No elements - nothing to render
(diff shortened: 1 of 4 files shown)


tools/cli/table.go
```diff
@@ -0,0 +1,207 @@
...
+	PrintRawTime bool
+	// PrintDateTime will print both date & time
+	PrintDateTime bool
+}
+
+// RenderTable is generic function for rendering a slice of structs as a table
+func RenderTable(w io.Writer, slice interface{}, opts TableOptions) {
+	sliceValue := reflect.ValueOf(slice)
+	if sliceValue.Kind() != reflect.Slice {
+		panic("table must be a slice")
```
mantas-sidlauskas, line 59: Two questions here: 
* can we avoid panic? 
* Can we provide more data to the result, for better understanding, what data was passed

The same applies to the next panic

tools/cli/table.go
```diff
@@ -0,0 +1,207 @@
...
+			if r == 0 {
+				headers = append(headers, header)
+				colors = append(colors, tableHeaderBlue)
+			}
+
+			row = append(row, formatValue(elem.Field(f).Interface(), opts, tag))
+		}
+		if r == 0 {
+			table.SetHeader(headers)
+			if !opts.NoColor {
```
mantas-sidlauskas, line 99: NIT:
for better readability, I would do the opposite.
If `NoColor` is set, then remove colors.

mantas-sidlauskas (approved):
lgtm, added a couple of nit level suggestions

tools/cli/table.go
vytautas-karpavicius: Agree, updated.

tools/cli/table.go
vytautas-karpavicius: I think panics are ok here, as they could only happen during development process.
Agree on second point, added additional info what what actually provided.

PR #4775, opened 2022-03-21
As-simple-as-possible addition of background-processing ratelimits

Groxx:
## What / Why

"User" requests (start, signal, query) have been ratelimited for basically
forever, but requests from workers (e.g. polling for tasks, responding) have
not, except at the most coarse level (per-host RPS).

This has left us with no effective way to clamp down on heavy-weight domains
when systems are becoming overloaded.  While we can limit how fast they
*begin* new workflows, they continue *processing* as quickly as possible,
particularly if there is a backlog (e.g. they started several million
workflows at once to work through a bulk job of some kind).

Since we are in need of a quick "stop the bleeding" patch, this is built
roughly as simply as I could manage.  All apparently-relevant ratelimits
now split "user" and "worker" requests apart, and throttle them individually.
The defaults for "worker" requests are `0` (unlimited) so this should be a
no-op until configured to do otherwise.

This also moves many "counted but not limited" endpoints to the per-domain
"worker" limit (previously global), so they can be controlled.  Since the
default is unlimited, and the per-domain limiter also increments the global
counts, this should also be a no-op but makes things more flexible in configuration.

To both simplify identifying things that need changing, and to make the
user/worker separation more clear, I've renamed all relevant "RPS" to
"UserRPS" and added "WorkerRPS" (and similar renames).

I am not confident that this covers all possible endpoints, but it should
at least be no *worse* than it was before.

## Config value selection concerns

Since background-processing *response* requests are counted but not limited,
along with some other endpoints, these worker-responses can starve workers from
getting new tasks.  
... but this should self-resolve as these responses are completed with no new
tasks being distributed.  So I'm not concerned about these.  But it may mean that
polling will become "spiky" if many responses occur at the same time.

`go.uber.org/cadence` clients do appear to have exponential-backoff retries on
polling endpoints, growing from 10ms to 10s, but ratelimiting requests will likely
cause a small burst of unnecessary (but cheap to re-reject) retries.
We could reduce the request impact of this by sleeping a bit before returning, but
I'm inclined to just let these bursts occur.  Scaling the frontend up is relatively easy
if necessary.

The pieces of the Go client I'm looking at are these.  There is other poll-like code
though, so I'm not completely confident that this is proof:
- this retrier `Throttle()` call: https://github.com/uber-go/cadence-client/blob/473aaacd07e49c2c416d01021e5d6813640b3705/internal/internal_worker_base.go#L261
- this retrier config (values are immediately above): https://github.com/uber-go/cadence-client/blob/473aaacd07e49c2c416d01021e5d6813640b3705/internal/internal_worker_base.go#L163

I have not yet checked the behavior of the java client on ServiceBusyError, but
hopefully it's similar.

In the worst case of either client not doing backoff, doing a
```go
time.Sleep(min(10*time.Second, time.Until(context.Deadline())))
```
before returning the ServiceBusyError is probably sufficient.

## Design

At a high level, the design is "treat user / worker orthogonally, so there's no
new fallback logic" plus "copy user limit/paste/add `worker` to things".

It's intentionally extremely simplistic and only adds one new disjoint
dimension for limiting things.  There are no other new concepts or config
structures, only new config *keys* (which are structured like / named similarly
to their current "user" ones).

I spent some time thinking about ways to make fallbacks smarter, share limits
between endpoints, etc... and ended up deciding it was going to be both a
complete mess to change quickly, and would require substantially different
config.  So for now I think keeping things orthogonal-but-identical is the
right approach.

In the future I think we'd almost certainly benefit from some kind of system that
would let us control ratelimits by many fine-grained values (like per endpoint,
endpoint-type (user/worker/etc), domain, tasklist, etc).  I think I've come up
with a *feasible* option, but before getting too deep into that we should
probably hunt for existing options.  Both because this may be an already-solved
problem that we can just adopt, and because reviewing some may give us
better ideas for building our own.

## Migration concerns

Assuming I didn't screw something up horrifically, little to none.
Fallbacks fall back to unlimited, which matches current behavior.

The main concern is that servers wrapping all this will likely not be able to
configure these limits until they update their code.  And anyone building
custom ratelimiter logic will very likely have breaking changes, but I doubt
it'll be too complex to update-or-abandon these changes for them.


service/history/queue/transfer_queue_processor.go
```diff
@@ -601,6 +601,7 @@ func newTransferQueueFailoverProcessor(
 
 	currentClusterName := shardContext.GetService().GetClusterMetadata().GetCurrentClusterName()
 	failoverUUID := uuid.New()
+	// TODO: unused, find log and restore or remove this With
```
just-at-uber, line 604: if unused, can consider removing code?

just-at-uber (approved):

service/history/queue/transfer_queue_processor.go
Groxx: yea, I honestly just forgot that I added this as I was reading through.
lemme see what the history of this logger is, I wonder if it used to log something that would still be useful sometimes.

service/frontend/workflowHandler.go
```diff
@@ -758,19 +780,19 @@ func (wh *WorkflowHandler) RecordActivityTaskHeartbeat(
 		return nil, wh.error(err, scope)
 	}
 
-	domainWrapper := domainWrapper{
```
Groxx, line 761: this shadowing got annoying a couple times while figuring out what I wanted the end result to be, so I just got rid of it everywhere.

service/frontend/workflowHandler.go
```diff
@@ -4358,6 +4387,7 @@ func checkPermission(
 	return nil
 }
 
+// this is a clear sign that we should accept a func rather than an interface with a single method
```
Groxx, line 4390: or just a string.

I'm debating changing it everywhere to just be `allow(true, theRequest.GetDomain())` because that's the same but doesn't involve passing a giant object to something that only needs a tiny piece.

common/quotas/multistageratelimiter.go
```diff
@@ -20,32 +20,53 @@
...
+	domainWorkerLimiters *Collection
+
+	globalUserLimiter   Limiter
+	globalWorkerLimiter Limiter
 }
 
 // NewMultiStageRateLimiter returns a new domain quota rate limiter. This is about
 // an order of magnitude slower than
-func NewMultiStageRateLimiter(global Limiter, domainLimiters *Collection) *MultiStageRateLimiter {
+func NewMultiStageRateLimiter(globalUser, globalWorker Limiter, domainUser, domainWorker *Collection) *MultiStageRateLimiter {
```
vytautas-karpavicius, line 38: Could we maybe add user/worker split as a separate Limiter implementation?

Right now (at least from the arguments) it seems like we have 4 stages. I would suggest adding
`NewSplitLimiter` or something with like:
```
type SplitLimiter struct {
	userLimiter Limiter
	workerLimiter  Limiter
}
func (sl *SplitLimiter) Allow(info Info) bool {
	if info.IsUser {
		return sl.userLimiter(info)
	} else {
		return sl.workerLimiter(info)
	}
}
```

This would add orthogonal splitting concern as a separate implementation. Encapsulating splitted limiters just as multistage encapsules its stages.

vytautas-karpavicius (approved):
I have one suggestion regarding code organization, otherwise LGTM.

common/quotas/multistageratelimiter.go
Groxx: Yeah, that'd be a pretty simple change, and pulling it out simplifies things a fair bit. Will do 👍

common/quotas/limiter_test.go
```diff
@@ -94,6 +94,31 @@ func TestMultiStageRateLimitingMultipleDomains(t *testing.T) {
...
+	}
+	check := func(suffix string) {
+		// per-domain has one set of limits
+		checkDomain("withdomain", " for domains"+suffix)
+		// global has an independent set of limits
+		checkDomain("", " for global"+suffix)
+	}
+
+	check("")
+	time.Sleep(time.Second)
```
mantas-sidlauskas, line 118: Can we use mocked clock here? For example this one: https://github.com/benbjohnson/clock

Groxx: we have mock clocks in other packages, but not for this - the `golang.org/x/time.Limiter` uses `time.Now()` internally and we have no way to change that.

so unfortunately no.  but this doesn't slow down this package by 1 second, since they're all run in parallel.

Groxx: and with the rewrite to just use two limiters: this goes away anyway.

service/history/queue/transfer_queue_processor.go
Groxx: nope, nvm, it's just wrong.  the logger is being passed to the `newTransferQueueProcessorBase` at the end.

PR #4783, opened 2022-03-29
[WIP] Make target domain not active non-retryable

demirkayaender:
**What changed?**
With current logic, we retry tasks if the target domain is not active (failed over). This causes retry storms and it's not fixable unless the domain fails over back to the previous active region. Making this non-retryable will help regenerate the task.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


taylanisikdemir:

PR #4784, opened 2022-03-30
Use errgroup.Group for fanout style workfloads

vytautas-karpavicius:
**What changed?**
Use `errgroup.Group` for places where some kind of parallel fanout is used.

**Why?**
Simplifies code for running go routines and handling first error.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/client/history/client.go
+++ b/client/history/client.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"go.uber.org/yarpc"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/dynamicconfig"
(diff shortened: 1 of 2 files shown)


mantas-sidlauskas (approved):

client/history/client.go
```diff
@@ -804,44 +805,37 @@ func (c *clientImpl) GetReplicationMessages(
 		req.Tokens = append(req.Tokens, token)
 	}
 
-	var wg sync.WaitGroup
-	wg.Add(len(requestsByPeer))
+	g := &errgroup.Group{}
```
Groxx, line 808: I'm not 100% sure it's an appropriate choice here, but since any error is treated as fatal to the whole group, it may be worth using a `g, ctx := errgroup.WithContext(ctx)` - it'll cancel the context on any error, so concurrent requests will be stopped asap rather than attempt to complete.

_usually_ I think WithContext produces better behavior, but there are definitely exceptions.

vytautas-karpavicius: I did not use it here, because previous logic also would not cancel other requests. But yeah, adding WithContext might be worth here.

PR #4786, opened 2022-03-31
Feature/cdnc 2263 Add toggle which can block domain failovers

allenchen2244:
**What Changed?**
Added support for locking down a cluster when we don't want to allow domains to be failed over to it if the cluster is already at capacity

**Why?**
In certain cases where clusters are already running at capacity failing over additional traffic can have harmful effects.

**How did you test it?**
Unit tests for both a successful failover when cluster isn't in lockdown and unit tests for failed when cluster is in lockdown.
Was also able to spin up 2 clusters using docker ui voodoo to verify end to end on local machine.

**Potential Risks**
Nothing should happen initially since default value is false and the blocking code will never be reached.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -133,6 +133,7 @@ var (
 	errClusterNameNotSet                          = &types.BadRequestError{Message: "Cluster name is not set."}
 	errEmptyReplicationInfo                       = &types.BadRequestError{Message: "Replication task info is not set."}
 	errEmptyQueueType                             = &types.BadRequestError{Message: "Queue type is not set."}
+	errDomainInLockdown                           = &types.BadRequestError{Message: "Domain is not accepting fail overs at this time."}
 	errShuttingDown                               = &types.InternalServiceError{Message: "Shutting down"}
 
 	// err for archival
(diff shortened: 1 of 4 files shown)


service/frontend/workflowHandler.go
```diff
@@ -4387,6 +4393,13 @@ func checkPermission(
 	return nil
 }
 
+func checkFailOverPermission(config *Config, domainName string) error {
+	if config.Lockdown(domainName) {
```
Shaddoll, line 4397: Should we allow super users to perform failover when there is a lockdown?

service/frontend/workflowHandler.go
```diff
@@ -133,6 +133,7 @@ var (
 	errClusterNameNotSet                          = &types.BadRequestError{Message: "Cluster name is not set."}
 	errEmptyReplicationInfo                       = &types.BadRequestError{Message: "Replication task info is not set."}
 	errEmptyQueueType                             = &types.BadRequestError{Message: "Queue type is not set."}
+	errDomainInLockdown                           = &types.BadRequestError{Message: "Domain is not accepting fail overs at this time."}
```
Shaddoll, line 136: "Domain is not accepting fail overs due to lockdown." I think it's better to tell users there is a lockdown.

service/frontend/workflowHandler.go
forshakuras: We don't plan on making this auth based. It will purely be a flipr config

allenchen2244: We talked about this yesterday. The only plan right now is to have this as a flipr config not any sort of auth based solution

Shaddoll (approved):

davidporter-id-au (commented):
Nice!

PR #4787, opened 2022-04-04
Added API for retrieving DLQ message count

vytautas-karpavicius:
**What changed?**
Added API for retrieving DLQ message count.

DLQ count periodic loops that emit metrics were moved to corresponding handlers (for domain and replication DLQs). In there they cache periodic results that can later be retrieved with and API call. API request also has an option to provide `force` option to fetch DB directly within the call instead of using latest cached values.

This API call may take a while as it fans out to every history host which then further fan out in every shard.
Ideally this could be optimized to a single persistence operation if all underlying datastores would support `GROUP BY`, but currently this is not possible.

**Why?**
Currently the only way to retrieve shards that contain DLQ messages is via emitted metrics. Which is rather inconvenient for building automation on to deal with DLQ messages. 

**How did you test it?**
Deployed new version on our staging environment which contained DLQ messages.
Run CLI command `admin dlq count` against that cluster, which returned affected shards and counts.

**Potential risks**

**Release notes**

**Documentation Changes**


common/domain/dlqMessageHandler.go
```diff
@@ -50,12 +64,46 @@ func NewDLQMessageHandler(
...
+	d.logger.Info("Domain DLQ handler started.")
+}
+
+// Stop stops the DLQ handler
+func (d *dlqMessageHandlerImpl) Stop() {
+	if !atomic.CompareAndSwapInt32(&d.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
+		return
+	}
+
+	d.logger.Debug("Domain DLQ handler shutting down.")
```
mantas-sidlauskas, line 95: as start has `Info` level, I think this should be `Info` as well

mantas-sidlauskas (approved):
couple of "nit" level comments, otherwise lgtm

client/history/client.go
```diff
@@ -903,6 +903,46 @@ func (c *clientImpl) ReapplyEvents(
...
+	}
+
+	var mu sync.Mutex
+	responses := make([]*types.HistoryCountDLQMessagesResponse, 0, len(peers))
+
+	g := &errgroup.Group{}
+	for _, peer := range peers {
+		peer := peer
+		g.Go(func() error {
+			response, err := c.client.CountDLQMessages(ctx, request, append(opts, yarpc.WithShardKey(peer))...)
```
davidporter-id-au, line 924: might need a panic handler?

common/types/testdata/service_history.go
```diff
@@ -75,6 +75,8 @@ var (
 	}
 	HistoryGetReplicationMessagesRequest  = AdminGetReplicationMessagesRequest
 	HistoryGetReplicationMessagesResponse = AdminGetReplicationMessagesResponse
+	HistoryCountDLQMessagesRequest        = types.CountDLQMessagesRequest{ForceFetch: true}
+	HistoryCountDLQMessagesResponse       = types.HistoryCountDLQMessagesResponse{Entries: map[types.HistoryDLQCountKey]int64{types.HistoryDLQCountKey{1, "A"}: 10}}
```
davidporter-id-au, line 79: What are these values?

proto/internal/uber/cadence/admin/v1/service.proto
```diff
@@ -268,6 +271,15 @@ message DescribeClusterResponse {
   map<string,shared.v1.PersistenceInfo> persistence_info = 3;
 }
 
+message CountDLQMessagesRequest {
+  bool forceFetch = 1;
+}
+
+message CountDLQMessagesResponse {
+  repeated shared.v1.HistoryDLQCountEntry history = 1;
+  int64 domain = 2;
```
davidporter-id-au, line 280: I don't understand this being an int, sorry?

service/history/engine/interface.go
```diff
@@ -67,6 +67,7 @@ type (
 		RespondCrossClusterTasksCompleted(ctx context.Context, targetCluster string, responses []*types.CrossClusterTaskResponse) error
 		QueryWorkflow(ctx context.Context, request *types.HistoryQueryWorkflowRequest) (*types.HistoryQueryWorkflowResponse, error)
 		ReapplyEvents(ctx context.Context, domainUUID string, workflowID string, runID string, events []*types.HistoryEvent) error
+		CountDLQMessages(ctx context.Context, forceFetch bool) (map[string]int64, error)
```
davidporter-id-au, line 70: Question: Do you think there's any chance of this needing to return more than just a count value? Would a response struct be worthwhile in the event that there is potentially further stuff to return?

service/history/handler.go
```diff
@@ -1753,6 +1756,50 @@ func (h *handlerImpl) ReapplyEvents(
...
+	if h.isShuttingDown() {
+		return nil, errShuttingDown
+	}
+
+	g := &errgroup.Group{}
+	var mu sync.Mutex
+	entries := map[types.HistoryDLQCountKey]int64{}
+	for _, shardID := range h.controller.ShardIDs() {
+		shardID := shardID
+		g.Go(func() error {
```
davidporter-id-au, line 1778: Should the panic handler not be in here?

davidporter-id-au (approved):
This will be excellent

proto/internal/uber/cadence/admin/v1/service.proto
vytautas-karpavicius: We have two broad types of replication:
- History (which is sharded) therefore have multiple entries.
- Domain. This is "global" and has entirely different replication mechanism. Also there only single bucked for DLQs, therefore just a single number here.

Do you think we should have a different structure/naming here?

common/types/testdata/service_history.go
vytautas-karpavicius: Those are just dummy non-empty values used for testing type mappers.

service/history/engine/interface.go
vytautas-karpavicius: I have added more general structure in proto, so that additional fields could be added into key later if needed.

This interface is internal, so we can refactor it later if needed. For now I wanted to keep it simple.

service/history/handler.go
vytautas-karpavicius: This and also few other places did not have it. But yeah, it is probably a good idea.
Added log.CapturePanic for this and other missing places. It logs the panic and returns it as error.

PR #4789, opened 2022-04-05
Add jittered workflow deletion configuration

allenchen2244:
**What changed?**
Added a toggled parameter that spreads out workflow deletions over a day. We still maintain the same retention period as defined in domain configurations. This will randomly add up to the configured int amount in minutes.

**Why?**
When customers trigger a large number of workflows at once these workflows would finish at the same time. Depending on domain retention period configurations this could create alerts where all those workflows need to be deleted at once.

**How did you test it?**
Tested it end to end locally and verified visibility timestamp on local cassandra. Path coverage unit test without error

**Potential risks**
Nothing should happen as no time would be added since all the current domains would not have the new flag active and default is no jittering

**Release notes**

**Documentation Changes**


common/util.go
```diff
@@ -390,6 +390,11 @@ func GenerateRandomString(n int) string {
 	return string(b)
 }
 
+// GenerateRandomInt generates a random int between 2 numbers
+func GenerateRandomInt(num1 int, num2 int) int {
```
mantas-sidlauskas, line 394: for a simple rand.Intn, there's no need to add new function to `util`.

common/cache/domainCache.go
```diff
@@ -952,3 +955,10 @@ func (entry *DomainCacheEntry) IsSampledForLongerRetention(
 	}
 	return false
 }
+
+func (entry *DomainCacheEntry) HoursToAddToRetentionDuration() time.Duration {
+	if _, ok := entry.info.Data[JitteredWorkflowDeletionKey]; ok {
```
mantas-sidlauskas, line 960: we can make this more flexible/controllable:

instead of setting it to "true/false", let's add actual hours. This way we allow operator to define jitter period, 1 or 12 hours, for example.

If data provided is not a valid int, then use predefined default value

common/cache/domainCache.go
```diff
@@ -952,3 +955,10 @@ func (entry *DomainCacheEntry) IsSampledForLongerRetention(
 	}
 	return false
 }
+
+func (entry *DomainCacheEntry) HoursToAddToRetentionDuration() time.Duration {
```
mantas-sidlauskas, line 959: based on a method name, expected output is duration in hours, but no time.hour is used.
options: 
* change name to not mention hours
* add time.Hour

common/cache/domainCache.go
```diff
@@ -901,6 +902,11 @@ var SampleRetentionKey = "sample_retention_days"
 // SampleRateKey is key to specify sample rate
 var SampleRateKey = "sample_retention_rate"
 
+// JitteredWorkflowDeletionHoursKey is key to determine the number of hours
+// generated deletion tasks should be jittered over. Should be a
+// parsable number or else it defaults to 0
+var JitteredWorkflowDeletionHoursKey = "jittered_workflow_deletion_hours"
```
mantas-sidlauskas, line 908: no reason to keep it as `var`, move to `const()` declaration plase.

common/cache/domainCache.go
```diff
@@ -952,3 +958,14 @@ func (entry *DomainCacheEntry) IsSampledForLongerRetention(
 	}
 	return false
 }
+
+func (entry *DomainCacheEntry) HoursToAddToRetentionDuration() time.Duration {
```
mantas-sidlauskas, line 962: Add comment

mantas-sidlauskas, line 962: we need to write tests for this method, test cases:
 * no key -> 0
 * not a number -> 0
 * a number -> number

common/cache/domainCache.go
```diff
@@ -901,6 +902,11 @@ var SampleRetentionKey = "sample_retention_days"
 // SampleRateKey is key to specify sample rate
 var SampleRateKey = "sample_retention_rate"
 
+// JitteredWorkflowDeletionHoursKey is key to determine the number of hours
+// generated deletion tasks should be jittered over. Should be a
+// parsable number or else it defaults to 0
+const JitteredWorkflowDeletionHoursKey = "jittered_workflow_deletion_hours"
```
davidporter-id-au, line 908: Sorry for being late to the party here, but if we need the jitter to be any more granular this might be difficult. I *don't* have a huge amount of context on the feature, but perhaps seconds or minutes might be a better level of granularity?

mantas-sidlauskas: a thought: declaration of jitter in hours is OK, but imho, we should calculate jitter in minute (or even second) resolution to avoid deletion spikes. Does it make sense?

allenchen2244: I can make it randomly jitter over a minute instead 👍👍

mantas-sidlauskas: on the second thought: if we declare everything in minutes, this will bring more clarity on what to expect from this jitter

common/cache/domainCache.go
```diff
@@ -952,3 +958,15 @@ func (entry *DomainCacheEntry) IsSampledForLongerRetention(
 	}
 	return false
 }
+
+// HoursToAddToRetentionDuration return number of hours to add to jittered wf deletion time
+func (entry *DomainCacheEntry) HoursToAddToRetentionDuration() time.Duration {
+	if strHours, ok := entry.info.Data[JitteredWorkflowDeletionHoursKey]; ok {
```
vytautas-karpavicius, line 964: Just a general wondering.
We usually use dynamicconfig with some default value and overrides per domain if needed. While this would work, it does not seem to provide a configurable default for all domains.

allenchen2244: It seems that the other retention configuration is done in the same way. IMHO we should either move it all into dynamic config somehow or not at all. It seems random to have 1 field be dynamic config then another in the info map since they're both related to retention duration calculation. Wdyt?

vytautas-karpavicius: IMO the issue with this approach, is there is no convenient way to apply a default for the whole cluster, which is what we want to do. With maybe some exceptions for a few domains if needed.

allenchen2244: Yes I can change the default to be jittered instead and use the flag to not jitter or a different jitter config 👍

common/cache/domainCache.go
```diff
@@ -952,3 +958,15 @@ func (entry *DomainCacheEntry) IsSampledForLongerRetention(
...
 }
+
+// MinutesToAddToRetentionDuration return number of hours to add to jittered wf deletion time
+func (entry *DomainCacheEntry) MinutesToAddToRetentionDuration() time.Duration {
+	if strMinutes, ok := entry.info.Data[JitteredWorkflowDeletionMinutesKey]; ok {
+		intVar, err := strconv.Atoi(strMinutes)
+		if err != nil {
+			return 0
+		}
+		return time.Minute * time.Duration(rand.Intn(intVar))
```
davidporter-id-au, line 969: Do you have a seed value specified for this btw? This will be deterministic if not, and you'll not end up with any jitter

allenchen2244: There is a rand.seed call inside host/service.go which is called in the service.go of each of the 4 actual services. The e2e tests I ran always had reasonably random numbers so I don't think it's deterministic.

service/history/execution/mutable_state_task_generator.go
```diff
@@ -247,7 +249,7 @@ func (r *mutableStateTaskGeneratorImpl) GenerateWorkflowCloseTasks(
 	}
 
 	closeTimestamp := time.Unix(0, closeEvent.GetTimestamp())
-	retentionDuration := time.Duration(retentionInDays) * time.Hour * 24
+	retentionDuration := (time.Duration(retentionInDays) * time.Hour * 24) + (time.Duration(rand.Intn(workflowDeletionTaskJitterRange)) * time.Minute)
```
davidporter-id-au, line 252: Just checking, do we set a seed value?

davidporter-id-au (approved):

PR #4792, opened 2022-04-06
Update SQL implementation of UpdateExecution to support async transaction

Shaddoll:
**What changed?**
Update SQL implementation of UpdateExecution to support async transaction

**Why?**
To improve the latency of UpdateExecution for SQL implementation. There are multiple writes to different tables in the same transaction which we don't care about their order. Using async transaction can reduce network latency.

**How did you test it?**
run persistence test, run integration test, run bench test

**Potential risks**
It may break SQL implementation of UpdateExecution. The feature is controlled by a dynamic config. If we something goes wrong, we can turn off the feature by updating the dynamic config.

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
nice!

PR #4795, opened 2022-04-08
Prototype of a "dump all dynamic configs" system

Groxx:
### What

A human-oriented "get all dynamic config values for X, Y, and Z" API.  Internally I imagine we'll expose this on some of our debug UIs, so we can quickly compare configs between clusters / between domains.  For testing, I just printed periodically to stdout, and it correctly showed my config-file-based overrides.

I do not ever intend for this to be exposed as an API that clients could touch.  It has far too high risk if the config client does not operate entirely in memory.  If cluster-owners decide it is safe and want to expose it, they should create their own way to do so.

### Why

Because we've repeatedly wondered what a value is for a domain / cluster / etc.  The answer is spread across many source files + our dynamic config system, and we have no way to quickly gather them all or query for specific values.  
I figured something like this would be technically possible, and good news!  It is!  ...with quite a few caveats!  But I think those are acceptable in practice.

Longer term I think we need some bigger changes to our dynamic configs, particularly "register everything up-front", "pass context", and "let the client provide a full list if possible" (rather than this combinatorial hack-job).  I'm not yet sure what any of that looks like though, and this was pretty easy to implement.


Groxx:
Made a quick-n-dirty test with:
```go
go func() {
	for {
		<-time.After(5 * time.Second)
		all := dynamicconfig.GetAllDynamicConfigs(context.Background(), dc, nil, nil, nil, nil, nil, nil, nil)
		keys := make([]string, 0, len(all))
		for k := range all {
			keys = append(keys, dynamicconfig.Keys[k])
		}
		sort.Strings(keys)
		for _, k := range keys {
			this := all[dynamicconfig.KeyNames[k]]
			fmt.Println("Config:", k, "len:", len(this))
			for _, v := range this {
				if reflect.DeepEqual(v.Current, v.Default) {
					fmt.Printf("\t%v: %v\n", v.Filters, v.Current)
				} else {
					fmt.Printf("\t%v: %v (default %v)\n", v.Filters, v.Current, v.Default)
				}
			}
		}
	}
}()
```
Which outputs rather handily:
```
Config: frontend.domainFailoverRefreshInterval len: 1
	map[]: 10s
Config: frontend.domainFailoverRefreshTimerJitterCoefficient len: 1
	map[]: 0.1
Config: frontend.enableClientVersionCheck len: 1
	map[]: true (default false)
Config: frontend.esIndexMaxResultWindow len: 1
	map[]: 10000
Config: frontend.historyMgrNumConns len: 1
	map[]: 10
Config: frontend.persistenceGlobalMaxQPS len: 1
	map[]: 0
Config: frontend.persistenceMaxQPS len: 1
	map[]: 2000
Config: frontend.rps len: 1
	map[]: 1200
Config: frontend.shutdownDrainDuration len: 1
	map[]: 0s
Config: frontend.throttledLogRPS len: 1
	map[]: 20
Config: frontend.validSearchAttributes len: 1
	map[]: map[BinaryChecksums:KEYWORD CadenceChangeVersion:KEYWORD CloseStatus:INT CloseTime:INT CustomBoolField:BOOL CustomDatetimeField:DATETIME CustomDoubleField:DOUBLE CustomIntField:INT CustomKeywordField:KEYWORD CustomStringField:STRING DomainID:KEYWORD ExecutionTime:INT HistoryLength:INT IsCron:BOOL NumClusters:INT RunID:KEYWORD StartTime:INT TaskList:KEYWORD WorkflowID:KEYWORD WorkflowType:KEYWORD]
Config: frontend.visibilityArchivalQueryMaxPageSize len: 1
	map[]: 10000
Config: frontend.workerrps len: 1
	map[]: 2147483647
Config: history.adminOperationToken len: 1
	map[]: CadenceTeamONLY
Config: history.enableAdminProtection len: 1
	map[]: false
Config: limit.maxIDWarnLength len: 1
	map[]: 128
Config: system.enableGracefulFailover len: 1
	map[]: true
Config: system.enableReadFromClosedExecutionV2 len: 1
	map[]: false
Config: system.enableVisibilitySampling len: 1
	map[]: true
Config: system.maxRetentionDays len: 1
	map[]: 30
Config: system.minRetentionDays len: 1
	map[]: 0 (default 1)
Config: system.requiredDomainDataKeys len: 1
	map[]: map[]
```
which ain't bad, I think.  might even be worth finishing up and exposing.

davidporter-id-au (approved):

common/dynamicconfig/debug.go
```diff
@@ -0,0 +1,343 @@
...
+			list = unfilterable(k, v.defaultValue, prop())
+		case FloatPropertyFn:
+			list = unfilterable(k, v.defaultValue, prop())
+		case DurationPropertyFn:
+			list = unfilterable(k, v.defaultValue, prop())
+
+		// all shard IDs are treated the same
+		case IntPropertyFnWithShardIDFilter:
+			for _, s := range shardIDs {
+				if ctx.Err() != nil {
```
vytautas-karpavicius, line 181: Just curious. Why are you testing for context error in every case? Is this that sensitive that we need to catch it and return immediately (more over, this ctx is not even propagated down)? What would happen if this check would only be left in outer loop?

Groxx: Our dynamic config client API doesn't have any context argument, so we can't pass timeouts / a context to it.  It's something we should change.

Re why every case: because the clients are pluggable, and we don't pass context.  So we have no way to know how long / costly / etc it is to retrieve any given value.  Particularly for the 3-dimensional values, the number of queries involved could be *huge*, so cutting it off ASAP when asked seemed important.

In most cases, yea, I expect it's not necessary.  E.g. flipr just dumps disk files into memory periodically, so we can slam through it as hard as we want, and that's a good thing for a large number of reasons.  It's more that we have no way to ensure that a client has that quality, and if we don't provide a way to cancel these queries, they can't be canceled.

taylanisikdemir:

PR #4797, opened 2022-04-08
Wrap underlying cause for conditional update error

vytautas-karpavicius:
**What changed?**
Instead of using value error `execution.ErrConflict` lets use error type `execution.conflictError` which wraps underlying cause.

**Why?**
Currently it can be very hard to debug certain failures when only thing that is visible in logs or returned via some endpoints is 
`code:internal message:conditional update failed`

Now, underlying cause will be included in error message as well.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/context.go
```diff
@@ -45,10 +45,29 @@ const (
...
-	// ErrConflict is exported temporarily for integration test
-	ErrConflict = errors.New("conditional update failed")
-)
+type conflictError struct {
+	Cause error
+}
+
+func (e *conflictError) Error() string {
+	return fmt.Sprintf("conditional update failed: %s", e.Cause.Error())
+}
```
Groxx, line 54: Since I haven't actually looked into it in detail before: this should probably be `... %v", e.Cause`.

Because:
`fmt.Errorf` does this: https://github.com/golang/go/blob/master/src/fmt/errors.go#L17-L30
and, with the `Error()` impl which just returns that message, is basically "eagerly create the full message".  
TIL.  
That `wrapErrs` flag later reaches here in the fmt printer: https://github.com/golang/go/blob/master/src/fmt/print.go#L574-L587
which just swaps it with `%v`, and holds onto the error value (which is ignored by Errorf).  
TIL 2.

That's... simple and fairly backwards-compatible, but yow.  That's worse than I expected.  Potentially unnecessarily costly and destroys a lot of information for no good reason (doesn't check for or support `%+v` or `%#v` because it can't).

service/history/execution/context.go
```diff
@@ -45,10 +45,29 @@ const (
...
-var (
-	// ErrConflict is exported temporarily for integration test
-	ErrConflict = errors.New("conditional update failed")
-)
+type conflictError struct {
+	Cause error
+}
+
+func (e *conflictError) Error() string {
+	return fmt.Sprintf("conditional update failed: %s", e.Cause.Error())
```
Groxx, line 53:
```suggestion
	return fmt.Sprintf("conditional update failed: %v", e.Cause)
```

Since I haven't actually looked into it in detail before: this should probably be `... %v", e.Cause`.

Because:
`fmt.Errorf` does this: https://github.com/golang/go/blob/master/src/fmt/errors.go#L17-L30
and, with the `Error()` impl which just returns that message, is basically "eagerly create the full message".  
TIL.  
That `wrapErrs` flag later reaches here in the fmt printer: https://github.com/golang/go/blob/master/src/fmt/print.go#L574-L587
which just swaps it with `%v` for stringifying, and holds onto the error value (which is ignored by Errorf).  
TIL 2.

That's... simple and fairly backwards-compatible, but yow.  Worse than I expected.  Potentially unnecessarily costly and destroys a lot of information for no good reason (doesn't check for or support `%+v` or `%#v` because it can't).

service/history/execution/context.go
```diff
@@ -45,10 +45,29 @@ const (
...
+
+func (e *conflictError) Unwrap() error {
+	return e.Cause
+}
+
+// NewConflictError is only public because it used in workflow/util_test.go
+// TODO: refactor those tests
+func NewConflictError(cause error) error {
+	return &conflictError{cause}
+}
```
Groxx, line 64: personal preference: require a `*testing.T` argument for any test-only code.  even if it's unused, it makes it rather unambiguous that it's not intended for any other kind of use.

Groxx (approved):
Small stuff worth thinking about, but LGTM.

Hiding the value (and type if feasible) and using an `IsThing` func is definitely my personal preference - it's more future-flexible and you can't miss any of the old references, so 👍👍

service/history/execution/context.go
vytautas-karpavicius: That is actually a neat idea 👍 . I'm going to do that from now on.

PR #4799, opened 2022-04-15
Only update maxReadLevel after successful re-acquire of shard

Shaddoll:
**What changed?**
cherry pick https://github.com/temporalio/temporal/pull/2272 from temporal

> we shouldn't update maxReadLevel immediately when we attempt to reacquire the rangeid lock in the background, we should do it only after a successful re-acquisition (which on cassandra implies that the previous unknown transaction definitively succeeded or failed).

**Why?**
see https://github.com/temporalio/temporal/pull/2272

> If maxReadLevel is updated, readers catch up to it, and then the original transaction actually gets applied and becomes visible later, then we'd lose tasks.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -623,7 +623,6 @@ func (s *contextImpl) CreateWorkflowExecution(
 	); err != nil {
 		return nil, err
 	}
-	defer s.updateMaxReadLevelLocked(transferMaxReadLevel)
 
 Create_Loop:
 	for attempt := 0; attempt < conditionalRetryCount && ctx.Err() == nil; attempt++ {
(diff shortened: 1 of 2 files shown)


service/history/shard/context.go
```diff
@@ -634,55 +633,62 @@ Create_Loop:
...
 
 		response, err := s.executionManager.CreateWorkflowExecution(ctx, request)
-		if err != nil {
-			switch err.(type) {
-			case *types.WorkflowExecutionAlreadyStartedError,
-				*persistence.WorkflowExecutionAlreadyStartedError,
-				*persistence.CurrentWorkflowConditionFailedError,
-				*types.ServiceBusyError,
-				*persistence.TimeoutError,
-				*types.LimitExceededError:
```
Shaddoll, line 644: CreateWorkflowExecution never returns LimitExceededError.

service/history/shard/context.go
```diff
@@ -634,55 +633,62 @@ Create_Loop:
 		request.RangeID = currentRangeID
 
 		response, err := s.executionManager.CreateWorkflowExecution(ctx, request)
-		if err != nil {
-			switch err.(type) {
-			case *types.WorkflowExecutionAlreadyStartedError,
-				*persistence.WorkflowExecutionAlreadyStartedError,
-				*persistence.CurrentWorkflowConditionFailedError,
-				*types.ServiceBusyError,
-				*persistence.TimeoutError,
```
Shaddoll, line 643: When timeout happens, we should close the shard.

demirkayaender (approved):

PR #4800, opened 2022-04-17
Updating 'domain-not-found' error messages

davidporter-id-au:
Adds more information to some unhelpful error messages which don't specify what's occurring when history gets stuck in a loop complaining about not being able to load form the domain cache. 

**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/cache/domainCache.go
```diff
@@ -617,8 +618,7 @@ func (c *domainCache) getDomain(
 	if cacheHit {
 		return c.getDomainByID(id, true)
 	}
-	// impossible case
-	return nil, &types.InternalServiceError{Message: "domainCache encounter case where domain exists but cannot be loaded"}
+	return nil, fmt.Errorf("failure to load domain %s: domain exists but cannot be loaded - %w", name, &types.InternalServiceError{})
```
vytautas-karpavicius, line 621: Any reason why we are wrapping the error this way?
It will no longer be InternalServiceError at the top, and we will end up reporting it as "Uncategorized error" for example [here](https://github.com/uber/cadence/blob/master/service/frontend/adminHandler.go#L1548).

davidporter-id-au: ah! good catch. I need to change that code too to use errors.Is/As (https://go.dev/blog/go1.13-errors) 

the %w directive should ensure that the internalServiceError is wrapped and propogated along with the aforementioned message. 

Do I need to do this here? Maybe not, I could do the the is/As change separately, however, considering this was not supposed to be possible I figured it wasn't high risk

davidporter-id-au: I think you're entirely right and I fully support this. The same thing had occurred to me a while back when I was looking at the RPC types.

PR #4806, opened 2022-04-27
Check for resurrected activities during RecordActivityTaskStarted

vytautas-karpavicius:
**What changed?**
- Moved out resurrected activity check code out of task processing.
- Added additional resurrected activity check in `historyEngine.RecordActivityTaskStarted` if it is already past scheduleToClose timeout.

**Why?**
Similarly to timer task processing, resurrected activity can sometimes be observed when matching triggers RecordActivityTaskStarted. In those cases activity gets started twice which result in non-deterministic errors, failures to replicate to remote cluster followed by DLQ messages.

To mitigate that we could reuse logic used in timer task processing and scan for previous history if such activity was already completed before. If so, delete it from mutable state and return that it does not exist.

As history scan may be expensive - only do it if timing of RecordActivityTaskStarted is suspicious. That is - if it is attempted to start after it should already be completed or timed out.

**How did you test it?**
Added additional unit test for history engine simulating resurrected activity condition.

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
What's the release plan? Do you intend to have a 'monitor-only' mode?

service/history/historyEngine.go
```diff
@@ -1702,6 +1703,38 @@ func (e *historyEngineImpl) RecordActivityTaskStarted(
 			requestID := request.GetRequestID()
 			ai, isRunning := mutableState.GetActivityInfo(scheduleID)
 
+			// RecordActivityTaskStarted is already past scheduleToClose timeout.
+			// If at this point pending activity is still in mutable state it may be resurrected.
+			// Otherwise it would be completed or timed out already.
+			if isRunning && e.timeSource.Now().After(ai.ScheduledTime.Add(time.Duration(ai.ScheduleToCloseTimeout)*time.Second)) {
```
davidporter-id-au, line 1709: What's the risk of this being a false-positive due to clock-skew or GC or something?

vytautas-karpavicius: > What's the risk of this being a false-positive due to clock-skew or GC or something?

This is just a trigger check. We don't want to run those checks often as they are potentially expensive with large histories. Worst case - wasted resources, increased latencies.

service/history/execution/integrity.go
```diff
@@ -0,0 +1,177 @@
...
+		minTimerStartedID,
+		mutableState.GetNextEventID(),
+		branchToken,
+	))
+	for iter.HasNext() {
+		item, err := iter.Next()
+		if err != nil {
+			return nil, err
+		}
+		event := item.(*types.HistoryEvent)
```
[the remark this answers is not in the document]
vytautas-karpavicius: This code was moved (from `service/history/task/timer_active_task_executor.go`) as is. Going to leave it for now. Ideally we should introduce generics here with go 1.18.

PR #4811, opened 2022-05-04
Improve failover coordinator error logging

mantas-sidlauskas:
**What changed?**
Adding err tags to log messages

**Why?**
This will help diagnose log messages as actual error now will be visible

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/domain/failover_watcher.go
+++ b/common/domain/failover_watcher.go
@@ -24,6 +24,7 @@ package domain
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


common/domain/failover_watcher.go
```diff
@@ -183,13 +184,12 @@ func CleanPendingActiveState(
 	// this call has to be made
 	metadata, err := domainManager.GetMetadata(context.Background())
 	if err != nil {
-		return err
+		return fmt.Errorf("getting metadata: %w", err)
```
davidporter-id-au, line 187: I don't have a great picture in my head for what might be listening to this error. Is there anything that might be doing an exact match on this which needs to be changed to use `errors.As` ?

davidporter-id-au (approved):

common/domain/failover_watcher.go
mantas-sidlauskas: good point. but no, this error is just being used for logging purposes in `coordinator.go:309` and `failower_watcher.go:160`

davidporter-id-au: sounds good then

PR #4817, opened 2022-05-05
Fix error type returned from GetWorkflowExecution and DeleteWorkflowExecution

Shaddoll:
**What changed?**
Do not convert error returned from GetWorkflowExecution and DeleteWorkflowExecution to InternalServiceError

**Why?**
We have logic check error type from persistence layer and converting those errors to InternalServiceError results in unexpected behavior such as incorrect retry policy

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
I assume the persistence errors are handling the DB errors and wrapping them correctly, in which case this makes sense

PR #4823, opened 2022-05-10
Forward activity responses and heartbeats on failover as well

Groxx:
We have a user desiring this, and in general it seems like a good idea.
Activities are generally assumed to be "high cost" to lose, or at least potentially.

Longer term, we should probably consider making this a per-domain config,
rather than something that is hardcoded for a whole cluster.  Nothing
about this seems like it would be cluster-bound.  For the current planned use
though, that's fine.

### DIFF
--- a/service/frontend/clusterRedirectionPolicy.go
+++ b/service/frontend/clusterRedirectionPolicy.go
@@ -43,10 +43,34 @@ const (
 	// 5. TerminateWorkflowExecution
 	// 6. QueryWorkflowStrongConsistency
 	// 7. ResetWorkflow
-	// please also reference selectedAPIsForwardingRedirectionPolicyAPIAllowlist
+	// please also reference selectedAPIsForwardingRedirectionPolicyAPIAllowlist and DCRedirectionPolicySelectedAPIsForwardingV2
 	DCRedirectionPolicySelectedAPIsForwarding = "selected-apis-forwarding"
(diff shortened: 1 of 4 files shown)


mkolodezny (approved):

service/frontend/clusterRedirectionPolicy.go
```diff
@@ -83,6 +109,28 @@ var selectedAPIsForwardingRedirectionPolicyAPIAllowlist = map[string]struct{}{
 	"ResetWorkflowExecution":           {},
 }
 
+// selectedAPIsForwardingRedirectionPolicyAPIAllowlistV2 contains a list of non-worker APIs which can be redirected.
+// This is paired with DCRedirectionPolicySelectedAPIsForwardingV2 - keep both lists up to date.
```
davidporter-id-au, line 113: do we have a unit test for this maybe? (assert.Equal(t, len(selectedAPIsForwardingRedirectionPolicyAPIAllowlistV2), len(DCRedirectionPolicySelectedAPIsForwardingV2))

davidporter-id-au (approved):

service/frontend/clusterRedirectionPolicy.go
Groxx: yea, probably a good idea.  lemme write one.  I can also ensure it's a superset with one.

PR #4824, opened 2022-05-10
Test output cleanup

Groxx:
While working on another PR, I got into the weeds of trying to figure out why my local tests
were failing, getting frustrated at unrelated output making it confusing, and decided to do something about it.

So in a nutshell, this PR contains:
- use the global `log.*` methods less
- thread `*testing.T` through some (but not all) of the noisier tests, mostly persistence-test-based and things that call it
- as I did ^ that, `loggerimpl.NewDevelopment` became unused!  yay!
- noticed domain/dlqMessageHandler_test.go tests were failing...
  - ... and noticed some obviously-incorrect expectations being set, but _not_ causing failures in CI...
  - ... so these have been wrong since they were introduced in https://github.com/uber/cadence/pull/3010
  - ... and the new failures were likely because I was using 1.17, which has `t.Cleanup(func())`, which gomock uses to automatically check assertions nowadays.
  - so I fixed them and deleted some unnecessary copypasta
- noticed races in the tasklist/scavenger tests, struggled to understand the intent behind the mutexes, gave up and changed it
  - this would _likely_ be better if it shut down synchronously, and as far as I can tell that should work fine.
    I didn't do it here though, as it's a larger semantic/behavioral change.  I'm entirely open to making it sync though.
- last but not least, `make test TEST_ARG=-race` no longer spews a ton of output on the main server process test.
  - but it still does by default.  if your TEST_ARG includes `-v`, you'll get `-v` there too, and that's enabled by default.

---

This had MUCH fewer new races than the last attempt I made in other code (nearly all due to async shutdown),
and `make test TEST_ARG=-race` is now pleasantly terse outside of failures.

You can almost notice tests failing now _just by looking_!


service/history/historyEngine2_test.go
```diff
@@ -177,7 +178,7 @@ func (s *engine2Suite) TestRecordDecisionTaskStartedSuccessStickyExpired() {
 
 	msBuilder := execution.NewMutableStateBuilderWithEventV2(
```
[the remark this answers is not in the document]
Groxx: yea, this one only takes like 6 changes.  the function above it is in _89_, tbh I think that needs its own commit or something :|

tools/cli/render_test.go
```diff
@@ -78,9 +78,9 @@ func Test_RenderTable(t *testing.T) {
 			data: testTable,
 			opts: RenderOptions{OptionalColumns: map[string]bool{"map": true, "slice": false}, PrintDateTime: true},
 			expectOutput: "" +
-				"        STRING        | INTEGER | BOOL  |         TIME         |    MAP      \n" +
-				"  text                |     123 | true  | 2000-01-02T03:04:05Z | A:AA, B:BB  \n" +
-				"  ...g long long long |     456 | false | 2000-11-12T13:14:15Z |             \n",
+				"        STRING        | INTEGER | BOOL  |           TIME            |    MAP      \n" +
```
mantas-sidlauskas, line 81: what is happening here?

common/domain/dlqMessageHandler_test.go
```diff
@@ -210,25 +202,9 @@ func (s *dlqMessageHandlerSuite) TestMergeMessages_ThrowErrorOnGetDLQAckLevel()
...
 
-	tasks := []*types.ReplicationTask{
-		{
-			TaskType:             types.ReplicationTaskTypeDomain.Ptr(),
-			SourceTaskID:         messageID,
-			DomainTaskAttributes: domainAttribute,
-		},
-	}
 	s.mockReplicationQueue.EXPECT().GetDLQAckLevel(gomock.Any()).Return(int64(-1), testError).Times(1)
-	s.mockReplicationQueue.EXPECT().GetMessagesFromDLQ(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
```
davidporter-id-au, line 227: do we know the intention of this being removed?

PR #4828, opened 2022-05-15
Add visibility operation types to Kafka message 

duoertai:
**What changed?**
Add a new field to indicate what visibility operations is for the message 

**Why?**
For https://github.com/cadence-oss/cadence-notification/issues/17

**How did you test it?**
Local tested 

**Potential risks**
No, everything is completely compatible 

**Release notes**

**Documentation Changes**


duoertai:
@longquanzheng

common/constants.go
```diff
@@ -219,3 +219,9 @@ const MemoKeyForOperator = "operator"
 
 // ReservedTaskListPrefix is the required naming prefix for any task list partition other than partition 0
 const ReservedTaskListPrefix = "/__cadence_sys/"
+
+// Enum for visibility message type
```
longquanzheng, line 223: Type is probably to generic name to understand. How about call it "visibilityOperation"? and we will have three operations: RecordStarted, RecordClosed and UpsertSearchAttributes

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -92,7 +92,7 @@ func (v *esVisibilityStore) RecordWorkflowExecutionStarted(
 	request *p.InternalRecordWorkflowExecutionStartedRequest,
 ) error {
 	v.checkProducer()
-	msg := getVisibilityMessage(
+	msg := createVisibilityMessage(
```
longquanzheng, line 95: Seems like you can delete the existing methods getVisibilityMessage and getVisibilityMessageForCloseExecution

longquanzheng (commented):
Some other comments:

* To setup local env, see  https://github.com/uber/cadence/blob/master/CONTRIBUTING.md and use ./docker/dev/cassandra-esv7-kafka.yml
* also need to update tests in esVisibilityStore_test and run the test locally 
* to verify the field is published to Kafka locally, you can use a tool to read the message, or you can change this code to emit a Info log: https://github.com/uber/cadence/blob/master/service/worker/indexer/esProcessor.go#L215 (or debug log if you want to check it in)

longquanzheng (approved):

longquanzheng:
@duoertai can you also update the test `esVisibilityStore_test` and also sign the CLA

duoertai:
@longquanzheng tried local testing, it seems the new field is added, but the new field is not valid to ES. here is the log:
`{"level":"error","ts":"2022-06-12T17:07:04.548-0700","msg":"Unregistered field.","service":"cadence-worker","component":"indexer","component":"indexer-processor","es-field":"NotificationType","logging-call-at":"processor.go:250","stacktrace":"github.com/uber/cadence/common/log/loggerimpl.(*loggerImpl).Error\n\t/Users/ertaiduo/Projects/cadence/common/log/loggerimpl/logger.go:136\ngithub.com/uber/cadence/service/worker/indexer.(*indexProcessor).dumpFieldsToMap\n\t/Users/ertaiduo/Projects/cadence/service/worker/indexer/processor.go:250\ngithub.com/uber/cadence/service/worker/indexer.(*indexProcessor).generateESDoc\n\t/Users/ertaiduo/Projects/cadence/service/worker/indexer/processor.go:230\ngithub.com/uber/cadence/service/worker/indexer.(*indexProcessor).addMessageToES\n\t/Users/ertaiduo/Projects/cadence/service/worker/indexer/processor.go:213\ngithub.com/uber/cadence/service/worker/indexer.(*indexProcessor).process\n\t/Users/ertaiduo/Projects/cadence/service/worker/indexer/processor.go:179\ngithub.com/uber/cadence/service/worker/indexer.(*indexProcessor).messageProcessLoop\n\t/Users/ertaiduo/Projects/cadence/service/worker/indexer/processor.go:161"}`

duoertai:
and it was due to the check here https://github.com/uber/cadence/blob/master/service/worker/indexer/processor.go#L249-L250:
`func (p *indexProcessor) isValidFieldToES(field string) bool {
	if _, ok := p.config.ValidSearchAttributes()[field]; ok {
		return true
	}
	if field == definition.Memo || field == definition.KafkaKey || field == definition.Encoding {
		return true
	}
	return false
}
`

do we need change the logic for this check? and do we need to update ES schema?

longquanzheng:
> and it was due to the check here https://github.com/uber/cadence/blob/master/service/worker/indexer/processor.go#L249-L250: `func (p *indexProcessor) isValidFieldToES(field string) bool { if _, ok := p.config.ValidSearchAttributes()[field]; ok { return true } if field == definition.Memo || field == definition.KafkaKey || field == definition.Encoding { return true } return false } `
> 
> do we need change the logic for this check? and do we need to update ES schema?

Yes let’s update the check. I don’t think we need to update ES schema because this field is only for Kafka

duoertai:
tested locally with kafka-console-consumer:
--------------------------------------------
NotificationType

RecordStarted
--------------------------------------------
NotificationType

UpsertSearchAttributes
--------------------------------------------
NotificationType


RecordClosed

common/constants.go
```diff
@@ -219,3 +219,15 @@ const MemoKeyForOperator = "operator"
 
 // ReservedTaskListPrefix is the required naming prefix for any task list partition other than partition 0
 const ReservedTaskListPrefix = "/__cadence_sys/"
+
+type (
+	// EncodingType is an enum that represents various data encoding types
```
longquanzheng, line 224: Comment is inaccurate

common/elasticsearch/defs.go
```diff
@@ -24,20 +24,21 @@ import "github.com/uber/cadence/.gen/go/indexer"
...
+	ExecutionTime    = "ExecutionTime"
+	CloseTime        = "CloseTime"
+	CloseStatus      = "CloseStatus"
+	HistoryLength    = "HistoryLength"
+	Memo             = "Memo"
+	Encoding         = "Encoding"
+	TaskList         = "TaskList"
+	IsCron           = "IsCron"
+	NumClusters      = "NumClusters"
+	NotificationType = "NotificationType"
```
longquanzheng, line 41: Also need to update this name

duoertai:
tested in local env again, field name is updated to VisibilityOperation

service/worker/indexer/processor.go
```diff
@@ -252,6 +252,11 @@ func (p *indexProcessor) dumpFieldsToMap(fields map[string]*indexer.Field) map[s
 			continue
 		}
 
+		// skip processing NotificationType since it's for kafka
```
longquanzheng, line 255: Skip VisibilityOperation since it’s not being used for advanced visibility

PR #4834, opened 2022-05-18
initial 1.18 upgrade

Groxx:
Just testing out of curiosity.
Contains both 1.18 and a full `go get -u all` (which is WAY more likely to cause problems)


Groxx:
yea, lots of problems.  heh.

PR #4838, opened 2022-05-20
Fix error conversion for WorkflowExecutionAlreadyStartedError

Shaddoll:
**What changed?**
Convert perisistence.WorkflowExecutionAlreadyStartedError to types.WorkflowExecutionAlreadyStartedError instead of types.InternalServiceError

**Why?**
We treat types.InternalServiceError as a server error in frontend, but WorkflowExecutionAlreadyStartedError should be an application error. And this affects our availability when customers keeps retrying on WorkflowExecutionAlreadyStartedError

**How did you test it?**
existing test

**Potential risks**
It's possible that customers at client side check the error type, and since this change changes the error type, it could be a breaking change for the customers.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -2061,7 +2061,7 @@ func (h *handlerImpl) convertError(err error) error {
 
 		return shard.CreateShardOwnershipLostError(h.GetHostInfo(), info)
 	case *persistence.WorkflowExecutionAlreadyStartedError:
-		return &types.InternalServiceError{Message: err.Msg}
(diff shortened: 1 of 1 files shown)


Groxx:
Yeah, for purely-internal this certainly makes sense.  Even publicly we've had this in our APIs since the reuse-policy was introduced: https://github.com/uber/cadence/pull/1185

Groxx (approved):

mkolodezny (approved):

PR #4844, opened 2022-05-24
Log replication messages that did not fit

vytautas-karpavicius:
**What changed?**
During fanout to multiple peers and collection of replication messages we collect responses up until the RPC message size limit. Remaining responses are discarded. They are expected to be picked up with next call.

However, there is not visibility whether (or how often) this is happening.

- Adding additional logging to emit which shards were discarded as well as info about sizes (max size, total size accumulated so far and current size that is out of limits).
- Also, instead of terminating early and returning partial response. Continue the loop and try other peer responses. They may be smaller and could still fit into the response.

**Why?**
For better visibility.

**How did you test it?**
Deployed in our staging environment, reduced max rpc limit via dynamic config `system.grpcMaxSizeInByte` and checked that logs are emitted.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/client/history/client.go
+++ b/client/history/client.go
@@ -59,6 +59,7 @@ type (
 	getReplicationMessagesWithSize struct {
 		response *types.GetReplicationMessagesResponse
 		size     int
+		peer     string
 	}
 )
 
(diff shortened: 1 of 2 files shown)


mkolodezny (approved):

davidporter-id-au (approved):
This would be awesome, it's been a problem for debugging prod before.

PR #4852, opened 2022-05-27
Removed unused internal type getters

vytautas-karpavicius:
**What changed?**
Removed all unused internal type getters.

**Why?**
Dead code. We should move away from using them at all. Those are legacy from previously used Thrift types.

**How did you test it?**
Successful compilation is enough to ensure that they are not needed. 

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
Red diffs are good diffs 

Also appreciate the move away from Java-isms

PR #4853, opened 2022-05-31
Fix CLI rendering long workflow types

vytautas-karpavicius:
**What changed?**
Fixed long workflow type rendering in CLI.

**Why?**
For certain strings it was panic'ing with `error calling table: runtime error: slice bounds out of range [-9:]`

**How did you test it?**
Unit test & locally to verify it is fixed

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/render.go
+++ b/tools/cli/render.go
@@ -292,7 +292,7 @@ func trimString(str string, maxLength int) string {
 
 	items := strings.Split(str, "/")
 	lastItem := items[len(items)-1]
-	if len(str) < maxLength {
+	if len(lastItem) < maxLength {
 		return ".../" + lastItem
 	}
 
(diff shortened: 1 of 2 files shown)


tools/cli/render.go
```diff
@@ -292,7 +292,7 @@ func trimString(str string, maxLength int) string {
 
 	items := strings.Split(str, "/")
```
demirkayaender, line 293: Your change itself is good but is "/" a standard delimiter in workflow types that we expect/document? If not, this kind of looks arbitrary.

demirkayaender (approved):

tools/cli/render.go
vytautas-karpavicius: It is quite common as workflow type name default to fully qualified name if not overriden. So they have form of `path/to/workflow/file.WorkflowName` to it is get shortened to `.../file.WorkflowName`

PR #4855, opened 2022-06-01
Add metrics for empty reads

neil-xie:
Added new empty response metrics for persistence operation read calls, also added new counter and scope for empty response.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/persistenceMetricClients.go
+++ b/common/persistence/persistenceMetricClients.go
@@ -437,8 +437,7 @@ func (p *workflowExecutionPersistenceClient) ListCurrentExecutions(
 		var err error
 		resp, err = p.persistence.ListCurrentExecutions(ctx, request)
 		if err == nil && len(resp.Executions) == 0 {
-			p.metricClient.Scope(metrics.PersistenceEmptyResponseScope,
-				metrics.OperationTag("ListCurrentExecutions")).IncCounter(metrics.PersistenceEmptyResponseCounter)
+			p.metricClient.IncCounter(metrics.PersistenceListCurrentExecutionsScope, metrics.PersistenceEmptyResponseCounter)
 		}
 		return err
 	}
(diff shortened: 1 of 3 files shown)


common/persistence/persistenceMetricClients.go
```diff
@@ -436,6 +436,10 @@ func (p *workflowExecutionPersistenceClient) ListCurrentExecutions(
 	op := func() error {
 		var err error
 		resp, err = p.persistence.ListCurrentExecutions(ctx, request)
+		if err == nil && len(resp.Executions) == 0 {
+			p.metricClient.Scope(metrics.PersistenceEmptyResponseScope,
```
vytautas-karpavicius, line 440: I suggest reusing existing scopes, such as `metrics.PersistenceListCurrentExecutionsScope`. No need for new `metrics.OperationTag` as well since existing scope already indicates that.

neil-xie: > I suggest reusing existing scopes, such as `metrics.PersistenceListCurrentExecutionsScope`. No need for new `metrics.OperationTag` as well since existing scope already indicates that.
Got it, I can reuse the scopes just need the new counter. Thanks for the advice!

neil-xie: Got it, I can reuse the scopes just need the new counter. Thanks for the advice!

common/persistence/persistenceMetricClients.go
```diff
@@ -436,6 +436,9 @@ func (p *workflowExecutionPersistenceClient) ListCurrentExecutions(
 	op := func() error {
 		var err error
 		resp, err = p.persistence.ListCurrentExecutions(ctx, request)
+		if err == nil && len(resp.Executions) == 0 {
+			p.metricClient.IncCounter(metrics.PersistenceListCurrentExecutionsScope, metrics.PersistenceEmptyResponseCounter)
```
demirkayaender, line 440: Instead of creating a new scope, you need to add `operation` tags to your metrics. That would make the implementation simpler and we can both aggregate and separate empty responses by operations

demirkayaender (dismissed):

common/persistence/persistenceMetricClients.go
neil-xie: We have operation tags binding with the existing scopes in defs.go, for this one, it has `PersistenceListCurrentExecutionsScope: {operation: "ListCurrentExecutions"},`

demirkayaender: Oh I see, these are existing scopes.

vytautas-karpavicius (approved):

PR #4858, opened 2022-06-06
Update the default values of dynamic config to not depend on static config

Shaddoll:
**What changed?**

Update the default values of dynamic config to be static values, and to not depend on values read from static config file

**Why?**

- Dynamic config and static config should be independent of each other
- With this change we can create a source of truth to store the default values of dynamic config

**How did you test it?**
existing test

**Potential risks**

**Release notes**
If there is no `system.grpcMaxSizeInByte` config, it needs to be created. And the value should be the same as the `maxPayloadSize` in the static config.

**Documentation Changes**

### DIFF
--- a/client/clientfactory.go
+++ b/client/clientfactory.go
@@ -117,7 +117,7 @@ func (cf *rpcClientFactory) NewHistoryClientWithTimeout(timeout time.Duration) (
 	peerResolver := history.NewPeerResolver(cf.numberOfHistoryShards, cf.resolver, namedPort)
 
 	supportedMessageSize := cf.rpcFactory.GetMaxMessageSize()
-	maxSizeConfig := cf.dynConfig.GetIntProperty(dynamicconfig.GRPCMaxSizeInByte, supportedMessageSize)
+	maxSizeConfig := cf.dynConfig.GetIntProperty(dynamicconfig.GRPCMaxSizeInByte, 4*1024*1024)
 	if maxSizeConfig() > supportedMessageSize {
 		return nil, fmt.Errorf(
 			"GRPCMaxSizeInByte dynamic config value %v is larger than supported value %v",
--- a/common/archiver/archivalMetadata.go
+++ b/common/archiver/archivalMetadata.go
@@ -86,18 +86,18 @@ func NewArchivalMetadata(
 ) ArchivalMetadata {
 	historyConfig := NewArchivalConfig(
 		historyStatus,
-		dc.GetStringProperty(dynamicconfig.HistoryArchivalStatus, historyStatus),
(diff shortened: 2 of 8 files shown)


common/archiver/archivalMetadata.go
```diff
@@ -177,7 +183,7 @@ func (a *archivalConfig) ReadEnabled() bool {
 	if !a.ClusterConfiguredForArchival() {
 		return false
 	}
-	return a.enableRead()
+	return a.staticEnableRead && a.dynamicEnableRead()
```
davidporter-id-au, line 186: This configuration might be surprising as some folks might expect either `||` or `if dynamicConfig != nil return dynamicConfig else staticConfig`

davidporter-id-au: I *dont* have a strong opinion on what it should be as I don't know enough about how this and similar config values are used in production, but I think maybe we should call out this both in the head of the diff and maybe in comments?

common/archiver/archivalMetadata.go
```diff
@@ -55,7 +55,8 @@ type (
 	archivalConfig struct {
 		staticClusterStatus  ArchivalStatus
 		dynamicClusterStatus dynamicconfig.StringPropertyFn
-		enableRead           dynamicconfig.BoolPropertyFn
+		staticEnableRead     bool
```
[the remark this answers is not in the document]
Shaddoll: yes, it's hided through `ReadEnabled() bool`. if both are true, read is enabled

common/archiver/archivalMetadata.go
Shaddoll: This is to keep the behavior the same as GetClusterStatus function

davidporter-id-au (approved):
Looks good. I'm mildly worried this might manifest in some behaviour changes, but maybe we can use Steven's config dump to detect this.

client/clientfactory.go
```diff
@@ -117,7 +117,7 @@ func (cf *rpcClientFactory) NewHistoryClientWithTimeout(timeout time.Duration) (
 	peerResolver := history.NewPeerResolver(cf.numberOfHistoryShards, cf.resolver, namedPort)
 
 	supportedMessageSize := cf.rpcFactory.GetMaxMessageSize()
-	maxSizeConfig := cf.dynConfig.GetIntProperty(dynamicconfig.GRPCMaxSizeInByte, supportedMessageSize)
+	maxSizeConfig := cf.dynConfig.GetIntProperty(dynamicconfig.GRPCMaxSizeInByte, 4*1024*1024)
```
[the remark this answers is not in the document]
Shaddoll: will be addressed in https://github.com/uber/cadence/pull/4863

PR #4861, opened 2022-06-08
Move some proto definitions to admin package

vytautas-karpavicius:
**What changed?**
Move some proto type definitions from `shared` package to `admin` package.

Additionally added `.gitattributes` with `.gen linguist-generated=true` to collapse generated files in PRs by default.

**Why?**
This is a preparation step before moving out `admin/v1` to [cadence-idl](https://github.com/uber/cadence-idl/) repository. This is in turn need for WebUI grpc migration.

This change is safe, as types themselves did not change and thus they remain wire compatible.

**How did you test it?**
Deployed branch to staging2 environment.
Used CLI with grpc transport to issue `admin cluster describe` command.

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au:
Not at all opposed to this change, but I'm slightly scared that some of the monorepo tooling might expect IDL files to never be moved or renamed. Might I suggest just getting a stamp from @moisesvega  or some folks from the IDL group?

just-at-uber (approved):

davidporter-id-au:
> Not at all opposed to this change, but I'm slightly scared that some of the monorepo tooling might expect IDL files to never be moved or renamed. Might I suggest just getting a stamp from @moisesvega or some folks from the IDL group?

ignore me, I didn't know these weren't imported

davidporter-id-au (approved):

PR #4863, opened 2022-06-10
Refactor dynamic config

Shaddoll:
**What changed?**

- Refactored dynamic config to enforce type of dynamic properties
- Refactored dynamic config to provide a source of truth for default values of dynamic properties

**Why?**
The default values and types of dynamic properties are declared in comments and therefore are not enforced in our code. People can make a mistake when using these dynamic properties.

**How did you test it?**
unit test

**Potential risks**
the default value of `system.enableVisibilitySampling` was incorrectly set to `true` in frontend, and we fixed it in this PR.
If the frontend was depending on the default value of this property, the behavior is changed in this PR, which means read from elasticsearch is not sampled.

**Release notes**

**Documentation Changes**


common/dynamicconfig/constants.go
```diff
@@ -2120,456 +1224,3736 @@ const (
...
+		DefaultValue: common.ConvertIntMapToDynamicConfigMapProperty(map[int]int{0: 100, 1: 10000}),
+	},
+}
+
+var _keyNames map[string]Key
+
+func init() {
+	_keyNames = make(map[string]Key)
+	for k, v := range IntKeys {
+		_keyNames[v.KeyName] = k
```
Groxx, line 4942: should verify that they're all unique here, since it's kinda our only chance to do so

Shaddoll: yes, we should do it, will make an update

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -251,8 +253,8 @@ func (csc *configStoreClient) RestoreValue(name dc.Key, filters map[dc.Filter]in
 		return dc.NotFoundError
```
Shaddoll, line 253: add validation to check if key type and value type are matched

common/dynamicconfig/config.go
```diff
@@ -171,53 +171,51 @@ type IntPropertyFnWithWorkflowTypeFilter func(domainName string, workflowType st
...
 		val, err := c.client.GetIntValue(
 			key,
 			filters,
-			defaultValue,
 		)
 		if err != nil {
 			c.logError(key, filters, err)
 		}
-		c.logValue(key, filters, val, defaultValue, intCompareEquals)
+		c.logValue(key, filters, val, key.DefaultInt(), intCompareEquals)
```
Groxx, line 196: should be `key.DefaultValue()` (plus later ones)

common/dynamicconfig/config.go
```diff
@@ -171,53 +171,51 @@ type IntPropertyFnWithWorkflowTypeFilter func(domainName string, workflowType st
...
 
 // GetIntProperty gets property and asserts that it's an integer
-func (c *Collection) GetIntProperty(key Key, defaultValue int) IntPropertyFn {
+func (c *Collection) GetIntProperty(key IntKey) IntPropertyFn {
 	return func(opts ...FilterOption) int {
 		filters := c.toFilterMap(opts...)
 		val, err := c.client.GetIntValue(
 			key,
 			filters,
-			defaultValue,
```
Groxx, line 192: pretty sure this is necessary - without it I believe our dynamic config client will return zero values if we haven't overridden / defined the key, rather than the configured default value.

common/dynamicconfig/config_benchmark_test.go
```diff
@@ -33,7 +33,7 @@ func BenchmarkGetIntProperty(b *testing.B) {
 	cln := NewCollection(client, log.NewNoop())
 	key := MatchingMaxTaskBatchSize
 	for i := 0; i < b.N; i++ {
-		size := cln.GetIntProperty(key, 10)
-		assert.Equal(b, 10, size())
+		size := cln.GetIntProperty(key)
+		assert.Equal(b, key.DefaultInt(), size())
```
Groxx, line 37: this should be the default configured for the MatchingMaxTaskBatchSize:
```go
	MatchingMaxTaskBatchSize: DynamicInt{
		DynamicBase: DynamicBase{
			KeyName:     "matching.maxTaskBatchSize",
			Description: "MatchingMaxTaskBatchSize is max batch size for task writer",
		},
		DefaultValue: 100,
	},
```

common/dynamicconfig/config_test.go
```diff
@@ -181,8 +181,33 @@ func (s *configSuite) TestUpdateConfig() {
 }
 
 func TestDynamicConfigKeyIsMapped(t *testing.T) {
-	for i := UnknownKey; i < LastKeyForTest; i++ {
-		key, ok := Keys[i]
+	for i := UnknownIntKey + 1; i < LastIntKey; i++ {
+		key, ok := IntKeys[i]
+		require.True(t, ok)
+		require.NotEmpty(t, key)
```
Groxx, line 187: since these are a real pain to identify when they fail (you just see "not true" and the line, not the iteration number): add `i`/`key` to the message?

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -133,15 +133,16 @@ func (csc *configStoreClient) startUpdate() error {
...
 
-func (csc *configStoreClient) GetValueWithFilters(name dc.Key, filters map[dc.Filter]interface{}, defaultValue interface{}) (interface{}, error) {
-	return csc.getValueWithFilters(name, filters, defaultValue)
+func (csc *configStoreClient) GetValueWithFilters(name dc.Key, filters map[dc.Filter]interface{}) (interface{}, error) {
+	return csc.getValueWithFilters(name, filters, name.DefaultValue())
 }
 
-func (csc *configStoreClient) GetIntValue(name dc.Key, filters map[dc.Filter]interface{}, defaultValue int) (int, error) {
+func (csc *configStoreClient) GetIntValue(name dc.IntKey, filters map[dc.Filter]interface{}) (int, error) {
+	defaultValue := name.DefaultInt()
```
Groxx, line 145: fairly sure these should be `name.DefaultValue()` like the other cases (GetValue and earlier config comments)

common/dynamicconfig/configstore/config_store_client_test.go
```diff
@@ -86,7 +86,7 @@ func (s *configStoreClientSuite) SetupTest() {
 			SchemaVersion: 1,
 			Entries: []*types.DynamicConfigEntry{
 				{
-					Name: dc.Keys[dc.TestGetBoolPropertyKey],
+					Name: dc.TestGetBoolPropertyKey.String(),
```
Groxx, line 89: kinda wonder if this could be changed to hold a `Key` instead...
... but string is the same as it was, so if it's not trivial, nope.  it's fine.

common/dynamicconfig/configstore/config_store_client_test.go
```diff
@@ -328,16 +328,31 @@ func defaultTestSetup(s *configStoreClientSuite) {
 
 func (s *configStoreClientSuite) TestGetValue() {
 	defaultTestSetup(s)
-	v, err := s.client.GetValue(dc.TestGetBoolPropertyKey, true)
+	v, err := s.client.GetValue(dc.TestGetBoolPropertyKey)
 	s.NoError(err)
 	s.Equal(false, v)
```
Groxx, line 333: huh.  why wasn't this `true` before?

common/dynamicconfig/configstore/config_store_client_test.go
```diff
@@ -390,33 +405,24 @@ func (s *configStoreClientSuite) TestGetIntValue_FilterNotMatch() {
...
-	defaultValue := 2000
 	filters := map[dc.Filter]interface{}{
 		dc.DomainName: "global-samples-domain",
 	}
-	v, err := s.client.GetIntValue(dc.TestGetIntPropertyKey, filters, defaultValue)
-	s.Error(err)
-	s.Equal(defaultValue, v)
-}
-
-func (s *configStoreClientSuite) TestGetIntValue_WrongTypeKey() {
```
Groxx, line 409: these should probably come back when the value type checks are added

common/dynamicconfig/constants.go
```diff
@@ -48,170 +231,20 @@ const UnlimitedRPS = math.MaxInt32
 * So that our go-docs can have the full [documentation](https://pkg.go.dev/github.com/uber/cadence@v0.19.1/common/service/dynamicconfig#Key).
 ***/
 const (
-	UnknownKey Key = iota
+	UnknownIntKey IntKey = iota
 
 	// key for tests
-	TestGetPropertyKey
```
Groxx, line 54: this key completely disappeared, might mean mistakenly missing tests?

common/dynamicconfig/constants.go
```diff
@@ -2120,456 +1224,3736 @@ const (
...
+		},
+		DefaultValue: true,
+	},
+	MatchingEnableTaskInfoLogByDomainID: DynamicBool{
+		DynamicBase: DynamicBase{
+			KeyName:     "MatchingEnableTaskInfoLogByDomainID is enables info level logs for decision/activity task based on the request domainID",
+			Description: "matching.enableTaskInfoLogByDomainID",
+		},
+		DefaultValue: false,
+	},
```
Groxx, line 3763: key/description swapped

common/dynamicconfig/file_based_client.go
```diff
@@ -95,15 +95,16 @@ func NewFileBasedClient(config *FileBasedClientConfig, logger log.Logger, doneCh
...
 
-func (fc *fileBasedClient) GetValueWithFilters(name Key, filters map[Filter]interface{}, defaultValue interface{}) (interface{}, error) {
-	return fc.getValueWithFilters(name, filters, defaultValue)
+func (fc *fileBasedClient) GetValueWithFilters(name Key, filters map[Filter]interface{}) (interface{}, error) {
+	return fc.getValueWithFilters(name, filters, name.DefaultValue())
 }
 
-func (fc *fileBasedClient) GetIntValue(name Key, filters map[Filter]interface{}, defaultValue int) (int, error) {
+func (fc *fileBasedClient) GetIntValue(name IntKey, filters map[Filter]interface{}) (int, error) {
+	defaultValue := name.DefaultInt()
```
Groxx, line 107: same DefaultValue() note

common/dynamicconfig/file_based_client.go
```diff
@@ -166,9 +169,8 @@ func (fc *fileBasedClient) GetMapValue(
 	return defaultValue, errors.New("value type is not map")
```
Groxx, line 169: not actually relevant to this PR unless you feel like it, but:
might be useful to change this error message to include the actual type, via `%T`

common/dynamicconfig/file_based_client.go
```diff
@@ -187,7 +189,7 @@ func (fc *fileBasedClient) GetDurationValue(
 }
 
 func (fc *fileBasedClient) UpdateValue(name Key, value interface{}) error {
-	keyName := Keys[name]
+	keyName := name.String()
 	currentValues := make(map[string][]*constrainedValue)
 
 	confContent, err := ioutil.ReadFile(fc.config.Filepath)
```
Groxx, line 195: at a glance this seems like it's missing the same-type check as well, like the configstore client

common/dynamicconfig/file_based_client_test.go
```diff
@@ -62,38 +62,37 @@ func (s *fileBasedClientSuite) SetupTest() {
...
 	s.Equal(false, v)
 }
 
 func (s *fileBasedClientSuite) TestGetValue_NonExistKey() {
-	defaultValue := true
-	v, err := s.client.GetValue(LastKeyForTest, defaultValue)
+	v, err := s.client.GetValue(EnableVisibilitySampling)
 	s.Error(err)
-	s.Equal(defaultValue, v)
+	s.Equal(EnableVisibilitySampling.DefaultBool(), v)
```
Groxx, line 73: same zero-value -> `DefaultValue()` suspicion

common/dynamicconfig/file_based_client_test.go
```diff
@@ -183,17 +179,16 @@ func (s *fileBasedClientSuite) TestGetMapValue() {
 }
 
 func (s *fileBasedClientSuite) TestGetMapValue_WrongType() {
-	var defaultVal map[string]interface{}
```
Groxx, line 186: may be a sign we shouldn't use default values (for trivial ones, bool is fine) in these tests, to ensure we're not hitting some other kind of fallback branch

common/dynamicconfig/inMemoryClient.go
```diff
@@ -48,90 +48,87 @@ func (mc *inMemoryClient) SetValue(key Key, value interface{}) {
...
-func (mc *inMemoryClient) GetIntValue(name Key, filters map[Filter]interface{}, defaultValue int) (int, error) {
+func (mc *inMemoryClient) GetIntValue(name IntKey, filters map[Filter]interface{}) (int, error) {
 	mc.RLock()
 	defer mc.RUnlock()
 
 	if val, ok := mc.globalValues[name]; ok {
 		return val.(int), nil
 	}
-	return defaultValue, NotFoundError
+	return name.DefaultInt(), NotFoundError
```
Groxx, line 75: `DefaultValue()`

common/dynamicconfig/inMemoryClient.go
```diff
@@ -48,90 +48,87 @@ func (mc *inMemoryClient) SetValue(key Key, value interface{}) {
...
 
 	if val, ok := mc.globalValues[name]; ok {
 		return val.(time.Duration), nil
 	}
-	return defaultValue, NotFoundError
+	return name.DefaultDuration(), NotFoundError
 }
 
 func (mc *inMemoryClient) UpdateValue(key Key, value interface{}) error {
+	if err := ValidateKeyValuePair(key, value); err != nil {
```
Groxx, line 129: 👍

common/dynamicconfig/inMemoryClient.go
```diff
@@ -48,90 +48,87 @@ func (mc *inMemoryClient) SetValue(key Key, value interface{}) {
...
+		return err
+	}
 	mc.SetValue(key, value)
 	return nil
 }
```
Groxx, line 134: 👇 maybe also change those errors to mention in-mem, not file client

common/dynamicconfig/nopClient.go
```diff
@@ -31,45 +31,39 @@ import (
...
-	name Key, filters map[Filter]interface{}, defaultValue interface{},
-) (interface{}, error) {
+func (mc *nopClient) GetValueWithFilters(name Key, filters map[Filter]interface{}) (interface{}, error) {
 	return nil, NotFoundError
 }
 
-func (mc *nopClient) GetIntValue(name Key, filters map[Filter]interface{}, defaultValue int) (int, error) {
-	return defaultValue, NotFoundError
+func (mc *nopClient) GetIntValue(name IntKey, filters map[Filter]interface{}) (int, error) {
+	return name.DefaultInt(), NotFoundError
```
Groxx, line 43: here I think `DefaultInt()` is defensible, since `GetValue` returns nil (as it did before).

... though I think "all operations will always return default values" implies it should have been returning `defaultValue` like other clients.  and like some of these methods return, but not others.

... this client was/is kinda messed up when you look closely, huh.

common/dynamicconfig/nopClient.go
```diff
@@ -31,45 +31,39 @@ import (
...
 
-func (mc *nopClient) GetDurationValue(
-	name Key, filters map[Filter]interface{}, defaultValue time.Duration,
-) (time.Duration, error) {
-	return defaultValue, NotFoundError
+func (mc *nopClient) GetDurationValue(name DurationKey, filters map[Filter]interface{}) (time.Duration, error) {
+	return name.DefaultDuration(), NotFoundError
 }
 
 func (mc *nopClient) UpdateValue(name Key, value interface{}) error {
```
Groxx, line 69: 👇 same error message note, also totally optional

common/task/parallelTaskProcessor_test.go
```diff
@@ -210,8 +210,10 @@ func (s *parallelTaskProcessorSuite) TestMonitor() {
 
 	s.processor.shutdownWG.Add(1) // for monitor
 	dcClient := dynamicconfig.NewInMemoryClient()
+	err := dcClient.UpdateValue(dynamicconfig.TaskSchedulerWorkerCount, workerCount)
```
Groxx, line 213: yea, this makes sense for fixing tests that had different values before 👍 .  simple and accurate to what prod semantically has to do: overwrite it somehow.

common/util.go
```diff
@@ -1009,3 +1009,11 @@ func ConvertGetTaskFailedCauseToErr(failedCause types.GetTaskFailedCause) error
 		return &types.InternalServiceError{Message: "uncategorized error"}
 	}
 }
+
+// GetTaskPriority returns priority given a task's priority class and subclass
```
Groxx, line 1013: tbh I'm inclined to keep this next to the definitions of the priorities, since it's directly related to them.

semi-related though: is it difficult to make these different types?  relying on the iota'd consts as ints feels kinda wrong.
(if it's not trivial or you don't feel like it: nope!  this is how it was, leave it.)

host/dynamicconfig.go
```diff
@@ -151,7 +145,7 @@ func (d *dynamicClient) GetDurationValue(
 		}
 	}
 	d.RUnlock()
-	return d.client.GetDurationValue(name, filters, defaultValue)
+	return d.client.GetDurationValue(name, filters)
 }
 
 func (d *dynamicClient) UpdateValue(name dynamicconfig.Key, value interface{}) error {
```
Groxx, line 151: may also be missing type checks?  though with a "deferred to other system" impl it's a bit more vague.

Groxx, line 151: separately: 
```go
func newIntegrationConfigClient(client dynamicconfig.Client) *dynamicClient {
```
ick.  can we move that to the test that uses it?  (as with others, 100% optional)

service/frontend/workflowHandler_test.go
```diff
@@ -485,7 +485,8 @@ func (s *workflowHandlerSuite) TestStartWorkflowExecution_Failed_InvalidTaskStar
 
 func (s *workflowHandlerSuite) TestRegisterDomain_Failure_MissingDomainDataKey() {
 	dynamicClient := dc.NewInMemoryClient()
-	dynamicClient.UpdateValue(dc.RequiredDomainDataKeys, map[string]interface{}{"Tier": true})
+	err := dynamicClient.UpdateValue(dc.RequiredDomainDataKeys, map[string]interface{}{"Tier": true})
+	s.NoError(err)
```
Groxx, line 489: 👍

service/history/config/config.go
```diff
@@ -313,262 +312,236 @@ type Config struct {
...
+		TaskCriticalRetryCount:                  dc.GetIntProperty(dynamicconfig.TaskCriticalRetryCount),
+		ActiveTaskRedispatchInterval:            dc.GetDurationProperty(dynamicconfig.ActiveTaskRedispatchInterval),
+		StandbyTaskRedispatchInterval:           dc.GetDurationProperty(dynamicconfig.StandbyTaskRedispatchInterval),
+		TaskRedispatchIntervalJitterCoefficient: dc.GetFloat64Property(dynamicconfig.TaskRedispatchIntervalJitterCoefficient),
+		StandbyTaskReReplicationContextTimeout:  dc.GetDurationPropertyFilteredByDomainID(dynamicconfig.StandbyTaskReReplicationContextTimeout),
+		EnableDropStuckTaskByDomainID:           dc.GetBoolPropertyFilteredByDomainID(dynamicconfig.EnableDropStuckTaskByDomainID),
+		ResurrectionCheckMinDelay:               dc.GetDurationPropertyFilteredByDomain(dynamicconfig.ResurrectionCheckMinDelay),
+
+		QueueProcessorEnableSplit:                          dc.GetBoolProperty(dynamicconfig.QueueProcessorEnableSplit),
+		QueueProcessorSplitMaxLevel:                        dc.GetIntProperty(dynamicconfig.QueueProcessorSplitMaxLevel), // 3 levels, start from 0
```
Groxx, line 379:
```
// 3 levels, start from 0
```
lets copy that to the `QueueProcessorSplitMaxLevel` definition, it's good info.

service/history/config/config.go
```diff
@@ -313,262 +312,236 @@ type Config struct {
...
+
+		ReplicatorTaskDeleteBatchSize:          dc.GetIntProperty(dynamicconfig.ReplicatorTaskDeleteBatchSize),
+		ReplicatorReadTaskMaxRetryCount:        dc.GetIntProperty(dynamicconfig.ReplicatorReadTaskMaxRetryCount),
+		ReplicatorProcessorFetchTasksBatchSize: dc.GetIntPropertyFilteredByShardID(dynamicconfig.ReplicatorTaskBatchSize),
+		ReplicatorUpperLatency:                 dc.GetDurationProperty(dynamicconfig.ReplicatorUpperLatency),
+
+		ExecutionMgrNumConns:       dc.GetIntProperty(dynamicconfig.ExecutionMgrNumConns),
+		HistoryMgrNumConns:         dc.GetIntProperty(dynamicconfig.HistoryMgrNumConns),
+		MaximumBufferedEventsBatch: dc.GetIntProperty(dynamicconfig.MaximumBufferedEventsBatch),
 		// 10K signals should big enough given workflow execution has 200K history lengh limit. It needs to be non-zero to protect continueAsNew from infinit loop
```
Groxx, line 460: probably similar, worth copying to the definition / what if we made it based on the hardcoded value of max history length?

service/history/config/config.go
```diff
@@ -313,262 +312,236 @@ type Config struct {
...
-		MaximumSignalsPerExecution:      dc.GetIntPropertyFilteredByDomain(dynamicconfig.MaximumSignalsPerExecution, 10000),
-		ShardUpdateMinInterval:          dc.GetDurationProperty(dynamicconfig.ShardUpdateMinInterval, 5*time.Minute),
-		ShardSyncMinInterval:            dc.GetDurationProperty(dynamicconfig.ShardSyncMinInterval, 5*time.Minute),
-		ShardSyncTimerJitterCoefficient: dc.GetFloat64Property(dynamicconfig.TransferProcessorMaxPollIntervalJitterCoefficient, 0.15),
+		MaximumSignalsPerExecution:      dc.GetIntPropertyFilteredByDomain(dynamicconfig.MaximumSignalsPerExecution),
+		ShardUpdateMinInterval:          dc.GetDurationProperty(dynamicconfig.ShardUpdateMinInterval),
+		ShardSyncMinInterval:            dc.GetDurationProperty(dynamicconfig.ShardSyncMinInterval),
+		ShardSyncTimerJitterCoefficient: dc.GetFloat64Property(dynamicconfig.TransferProcessorMaxPollIntervalJitterCoefficient),
 
 		// history client: client/history/client.go set the client timeout 30s
```
Groxx, line 466: any idea what this is implying / should we copy it to definition?

service/history/config/config.go
```diff
@@ -313,262 +312,236 @@ type Config struct {
...
-		ThrottledLogRPS:   dc.GetIntProperty(dynamicconfig.HistoryThrottledLogRPS, 4),
-		EnableStickyQuery: dc.GetBoolPropertyFilteredByDomain(dynamicconfig.EnableStickyQuery, true),
-
-		ValidSearchAttributes:                    dc.GetMapProperty(dynamicconfig.ValidSearchAttributes, definition.GetDefaultIndexedKeys()),
-		SearchAttributesNumberOfKeysLimit:        dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesNumberOfKeysLimit, 100),
-		SearchAttributesSizeOfValueLimit:         dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesSizeOfValueLimit, 2*1024),
-		SearchAttributesTotalSizeLimit:           dc.GetIntPropertyFilteredByDomain(dynamicconfig.SearchAttributesTotalSizeLimit, 40*1024),
-		StickyTTL:                                dc.GetDurationPropertyFilteredByDomain(dynamicconfig.StickyTTL, time.Hour*24*365),
-		DecisionHeartbeatTimeout:                 dc.GetDurationPropertyFilteredByDomain(dynamicconfig.DecisionHeartbeatTimeout, time.Minute*30),
-		DecisionRetryCriticalAttempts:            dc.GetIntProperty(dynamicconfig.DecisionRetryCriticalAttempts, 10), // about 30m
```
Groxx, line 525:
```
// about 30m
```
maybe worth copying

service/history/config/config.go
```diff
@@ -581,22 +554,48 @@ func NewForTest() *Config {
 
 // NewForTestByShardNumber create new history service config for test
 func NewForTestByShardNumber(shardNumber int) *Config {
-	dc := dynamicconfig.NewNopCollection()
+	inMem := dynamicconfig.NewInMemoryClient()
+	var err error
+	err = inMem.UpdateValue(dynamicconfig.HistoryLongPollExpirationInterval, 10*time.Second)
+	panicIfErr(err)
```
Groxx, line 560: purely as an option:
```go
panicIfErr(inMem.UpdateValue(...))`
```
and no need for the var.

service/history/config/config.go
```diff
@@ -581,22 +554,48 @@ func NewForTest() *Config {
...
+	config.NormalDecisionScheduleToStartMaxAttempts = dc.GetIntPropertyFilteredByDomain(dynamicconfig.NormalDecisionScheduleToStartMaxAttempts)
 	return config
 }
 
 // GetShardID return the corresponding shard ID for a given workflow ID
 func (config *Config) GetShardID(workflowID string) int {
 	return common.WorkflowIDToHistoryShard(workflowID, config.NumberOfShards)
 }
+
+func panicIfErr(err error) {
```
Groxx, line 597: tbh I generally just define these inside the func that uses them, so they're not visible for reuse by other things.
but also meh, it hardly matters, this is fine.

common/constants.go
```diff
@@ -231,3 +229,33 @@ const (
...
+const (
+	numBitsPerLevel = 3
+
+	// NoPriority is the value returned if no priority is ever assigned to the task
+	NoPriority = -1
+)
+
+const (
+	// HighPriorityClass is the priority class for high priority tasks
+	HighPriorityClass = iota << numBitsPerLevel
```
Groxx, line 242: I'm fuzzy on why these had to move, but also I don't really mind either way.  just curious / forgetful.

service/worker/scanner/history/scavenger_test.go
```diff
@@ -70,7 +69,7 @@ func (s *ScavengerTestSuite) createTestScavenger(rps int) (*mocks.HistoryV2Manag
 	controller := gomock.NewController(s.T())
 	workflowClient := history.NewMockClient(controller)
 
-	maxWorkflowRetentionInDays := dynamicconfig.GetIntPropertyFn(domain.DefaultMaxWorkflowRetentionInDays)
+	maxWorkflowRetentionInDays := dynamicconfig.GetIntPropertyFn(dynamicconfig.MaxRetentionDays.DefaultInt())
```
Groxx, line 72: 30 -> 0 seems likely wrong  (and all the tests below do `-0 * 2`)

service/worker/scanner/tasklist/scavenger.go
```diff
@@ -137,7 +134,7 @@ func NewScavenger(
 	getOrphanTasksPageSize := opts.GetOrphanTasksPageSizeFn
 	if getOrphanTasksPageSize == nil {
 		getOrphanTasksPageSize = func(opts ...dynamicconfig.FilterOption) int {
-			return DefaultScannerGetOrphanTasksPageSize
+			return dynamicconfig.ScannerGetOrphanTasksPageSize.DefaultInt()
```
Groxx, line 137: 1000 -> 0

service/worker/scanner/workflow_test.go
```diff
@@ -68,7 +68,7 @@ func (s *scannerWorkflowTestSuite) TestScavengerActivity() {
 		resource: mockResource,
 		cfg: Config{
 			TaskListScannerOptions: tasklist.Options{
-				GetOrphanTasksPageSizeFn: dynamicconfig.GetIntPropertyFn(tasklist.DefaultScannerGetOrphanTasksPageSize),
+				GetOrphanTasksPageSizeFn: dynamicconfig.GetIntPropertyFn(dynamicconfig.ScannerGetOrphanTasksPageSize.DefaultInt()),
```
Groxx, line 71: another 1k -> 0 change

service/worker/service.go
```diff
@@ -130,78 +128,77 @@ func NewConfig(params *resource.Params) *Config {
...
 	advancedVisWritingMode := dc.GetStringProperty(
 		dynamicconfig.AdvancedVisibilityWritingMode,
-		common.AdvancedVisibilityWritingModeOn,
 	)
 	if common.IsAdvancedVisibilityWritingEnabled(advancedVisWritingMode(), params.PersistenceConfig.IsAdvancedVisibilityConfigExist()) {
 		config.IndexerCfg = &indexer.Config{
-			IndexerConcurrency:       dc.GetIntProperty(dynamicconfig.WorkerIndexerConcurrency, 1000),
-			ESProcessorNumOfWorkers:  dc.GetIntProperty(dynamicconfig.WorkerESProcessorNumOfWorkers, 1),
-			ESProcessorBulkActions:   dc.GetIntProperty(dynamicconfig.WorkerESProcessorBulkActions, 1000),
-			ESProcessorBulkSize:      dc.GetIntProperty(dynamicconfig.WorkerESProcessorBulkSize, 2<<24), // 16MB
```
Groxx, line 202:
```
// 16MB
```
I suppose there's a bit of a duplication between here / default documented value / default actual value 🤔 
... copy everywhere maybe?  it's in the comment, but not in the source that sets it.

tools/cli/adminConfigStoreCommands.go
```diff
@@ -186,9 +185,7 @@ func AdminListDynamicConfig(c *cli.Context) {
 	ctx, cancel := newContext(c)
 	defer cancel()
 
-	req := &types.ListDynamicConfigRequest{
-		ConfigName: dynamicconfig.UnknownKey.String(),
-	}
+	req := &types.ListDynamicConfigRequest{}
```
Groxx, line 188: I'm a little inclined to say "use `ConfigName: nil`"... but also kinda meh since there's only the one field.
an explicit nil in a field "feels" a bit more intentional to me tho, so I lean that way since the nil is intended, not just "an empty request".

tools/cli/domainUtils.go
```diff
@@ -276,9 +276,9 @@ func initializeDomainHandler(
 ) domain.Handler {
 
 	domainConfig := domain.Config{
-		MinRetentionDays:  dynamicconfig.GetIntPropertyFn(domain.DefaultMinWorkflowRetentionInDays),
-		MaxBadBinaryCount: dynamicconfig.GetIntPropertyFilteredByDomain(domain.MaxBadBinaries),
-		FailoverCoolDown:  dynamicconfig.GetDurationPropertyFnFilteredByDomain(domain.FailoverCoolDown),
+		MinRetentionDays:  dynamicconfig.GetIntPropertyFn(dynamicconfig.MinRetentionDays.DefaultInt()),
+		MaxBadBinaryCount: dynamicconfig.GetIntPropertyFilteredByDomain(dynamicconfig.MaxRetentionDays.DefaultInt()),
+		FailoverCoolDown:  dynamicconfig.GetDurationPropertyFnFilteredByDomain(dynamicconfig.FrontendFailoverCoolDown.DefaultDuration()),
```
Groxx, line 281: pretty significant changes here, both different names and now-zero values

common/dynamicconfig/config.go
Shaddoll: I think it's something we want. We can prevent from using undefined keys.

common/dynamicconfig/file_based_client.go
Shaddoll: we need to return int instead of interface{} in this function, cannot use DefaultValue() here

common/dynamicconfig/nopClient.go
Shaddoll: it can be deleted later. inMemoryClient can be used to replace nopClient.

common/dynamicconfig/configstore/config_store_client_test.go
Shaddoll: no, we cannot change it, because we store the key name in dynamic storage.

common/dynamicconfig/constants.go
Shaddoll: it's an untyped key. deleted it because we're enforcing type

common/dynamicconfig/nopClient.go
Groxx: yeah, that sounds pretty reasonable.  that'd also allow updates, which seems like a good thing for any place where nopClient is used.

common/dynamicconfig/configstore/config_store_client_test.go
Shaddoll: Wrong key type won't happen after introducing typed key.

common/dynamicconfig/file_based_client.go
Shaddoll: what's the benefit here?

common/constants.go
Shaddoll: common/task package imports common/dynamicconfig package.
we had to move it to avoid circular dependency

host/dynamicconfig.go
Shaddoll: the value parameter of UpdateValue is more generic, and the validation logic also depends on the implementation of client (see config_store_client), so we cannot do checks here

tools/cli/domainUtils.go
Shaddoll: synced offline, the DefaultXXX functions do not return zero values

common/dynamicconfig/config.go
Groxx: (from IRL chat) ah, yeah, I'm getting confused by remembering the old API.  the key contains default now, so there's no need to pass it a second time.

👍

common/dynamicconfig/configstore/config_store_client.go
Groxx: Nope, I was misinterpreting things.  this is good 👍

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -235,7 +237,11 @@ func (csc *configStoreClient) GetDurationValue(
 }
 
 func (csc *configStoreClient) UpdateValue(name dc.Key, value interface{}) error {
-	return csc.updateValue(name, value, csc.config.UpdateRetryAttempts)
+	dcValues, ok := value.([]*types.DynamicConfigValue)
```
Groxx, line 240: tbh this feels kinda odd, but I suppose you've run it and it's correct?

explicitly declaring the type is at least MUCH clearer about what's happening, so 👍👍  on that anyway.  I just wonder if converting to the correct type should be a responsibility of the caller of UpdateValue....

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -334,11 +352,6 @@ func (csc *configStoreClient) updateValue(name dc.Key, value interface{}, retryA
 			}
 		}
 	} else {
-		dcValues, ok := value.([]*types.DynamicConfigValue)
-		if !ok {
-			return errors.New("invalid value")
-		}
-
```
Groxx, line 341: ah, they moved from here.  great, looks correct then, and 👍👍 to clearer type safety.

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -593,3 +606,52 @@ func convertFromDataBlob(blob *types.DataBlob) (interface{}, error) {
...
+	case dc.DurationKey:
+		if _, ok := value.(time.Duration); !ok {
+			durationStr, ok := value.(string)
+			if !ok {
+				return err
+			}
+			if _, err = time.ParseDuration(durationStr); err != nil {
+				return errors.New("value string encoding cannot be parsed into duration")
+			}
+		}
```
Groxx, line 648: feels kinda odd to do this parse-and-validate here, and then do it again every time the value is retrieved in GetDurationValue, rather than just parsing it and using the parsed result...

... but that matches current behavior, so for this PR seems fine.  more of a "maybe we should change this later" though.

common/dynamicconfig/constants.go
```diff
@@ -1921,94 +1175,16 @@ const (
...
 	// Value type: Int
-	// Default value: false
+	// Default value: 5
 	// Allowed filters: N/A
-	CurrentExecutionsScannerInvariantCollectionHistory
-	// CurrentExecutionsScannerInvariantCollectionMutableState is indicates if mutable state invariant checks should be run
-	// KeyName: worker.currentExecutionsInvariantCollectionMutableState
+	TimersScannerConcurrency
+	// TimersScannerPersistencePageSize is the page size of timers persistence fetches in timers scanner
+	// KeyName: worker.timersScannerPersistencePageSize
```
Groxx, line 1185: heh, wow, that's a _terrible_ diff-algorithm result

common/constants.go
Groxx: ah, due to combining the types / keys / etc probably.  yea, that's fine.

Groxx (approved):
Overall much improved I think, thanks a lot for doing this!

PR #4864, opened 2022-06-10
Fix status check for visibility and archival

Shaddoll:
**What changed?**

- Update status check for visibility to not only check dynamic config, but also check static config
- Update read status check for archival to not only check dynamic config, but also check static config

**Why?**
- For visibility, it's possible that the dynamic config says the advanced visibility is enabled, but in static config we don't provide advanced visibility. In this case, the server might break, and we shouldn't allow that happen.
- For archival, it's possible that the dynamic config says read is enabled, but in static config we mark read as disabled. 

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
I think it's important to call out that this is preserving existing behaviour, but ty for doing this

PR #4866, opened 2022-06-14
CDNC-1746: cadence-history does not emit continue-as-new metrics

ZackLK:
- Emit metric when completing via ContinueAsNew
- Emit metric and logs when completing via unknown type, so that we'll
  know in the future if there are more types we should emit metrics for

It's good to have metrics.

Tested locally via docker-compose with local server changes and cadence-samples/pso which uses ContinueAsNew.
Plan to test in staging2 env (just to experience that workflow.)


service/history/execution/context_util.go
```diff
@@ -155,5 +155,15 @@ func emitWorkflowCompletionStats(
 			tag.WorkflowRunID(runID),
 			tag.WorkflowDomainName(domainName),
 		)
+	case types.EventTypeWorkflowExecutionContinuedAsNew:
+		scope.IncCounter(metrics.WorkflowContinuedAsNew)
+	default:
+		scope.IncCounter(metrics.WorkflowCompletedUnknownType)
+		logger.Info("NOT emitting completion stats for event type",
```
demirkayaender, line 162: Probably we should make this a `.Warn` and messaging around not knowing the completion event could be better:
`Workflow completed with an unknown even type`

service/history/execution/context_util.go
```diff
@@ -155,5 +155,15 @@ func emitWorkflowCompletionStats(
...
+			tag.WorkflowEventType(event.EventType.String()),
```
demirkayaender, line 163: We might just put this in the message string using `fmt.Sprintf`. Tags are mostly useful when we want to filter different type of logs.

service/history/execution/context_util.go
```diff
@@ -155,5 +155,15 @@ func emitWorkflowCompletionStats(
 			tag.WorkflowRunID(runID),
 			tag.WorkflowDomainName(domainName),
 		)
+	case types.EventTypeWorkflowExecutionContinuedAsNew:
+		scope.IncCounter(metrics.WorkflowContinuedAsNew)
```
emrahs, line 159: Also add a log here?

emrahs (approved):
Lgtm with a minor comment. Thanks!

service/history/execution/context_util.go
emrahs: +1

service/history/execution/context_util.go
ZackLK: I'm happy to if you think we should have one. It felt like WorkflowContinuedAsNew was normal enough that it didn't warrant a log (similar to the Completed / Cancelled / Failed cases), but perhaps they should all log?

service/history/execution/context_util.go
ZackLK: I had Emrah clarify his comment and I agree that constant strings are in general nicer for querying the "message" in logs. Ender, let me know if you feel strongly about using fmt.Sprintf instead of a tag and we can discuss more.

PR #4868, opened 2022-06-16
Adds cross-cluster canary

davidporter-id-au:
This adds a canary feature, config and workflow to test the cross-cluster and cross-domain feature. 

**Why?**

**How did you test it?**
- running locally

**Potential risks**
- This introduces some hopefully backwards compatable but otherwise additional config to the canary, so I'll do a deploy to check that it doesn't break without the config enabled. 

**Release notes**

**Documentation Changes**


canary/config.go
```diff
@@ -49,6 +49,9 @@ const (
 	CadenceServiceName = "cadence-frontend"
 	// CanaryServiceName is the default service name for cadence canary
 	CanaryServiceName = "cadence-canary"
+	// CrossClusterCanaryModeFull is a canary testing mode which tests all permutations of
+	// the cross-cluster/domain feature
+	CrossClusterCanaryModeFull = "test-all"
```
just-at-uber, line 54: Maybe list all possible options in a comment in the case developers want to change this value in the future?

just-at-uber (approved):

canary/config.go
davidporter-id-au: This is the only option so far, but I just didn't want to make it a bool in case there's future options

PR #4871, opened 2022-06-17
Fix query workflow high latency after a long inactive time

Shaddoll:
**What changed?**
see https://github.com/temporalio/temporal/pull/2811

**Why?**
see https://github.com/temporalio/temporal/pull/2811

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/matching/matchingEngine.go
```diff
@@ -48,6 +48,10 @@ import (
 	"github.com/uber/cadence/common/types"
 )
 
+// If sticky poller is not seem in last 10s, we treat it as sticky worker unavailable
+// This seems aggressive, but the default sticky schedule_to_start timeout is 5s, so 10s seems reasonable.
+const _stickyPollerUnavailableWindow = 10 * time.Second
```
Groxx, line 53: kinda implies it should be based on the default/configured sticky timeout.

if a user did have a very-long sticky schedule_to_start, would it be worth changing this time?  like say they set it to 5 days, to allow sequentially processing a large batch job on a single worker (say it's CPU bound).

Groxx:
I'm also curious about this:
>tchannel doesn't propagate context cancel from client to server, which fails the new integration test

How is that addressed in this PR?  And how does it not-pass cancellation?  I assume the client canceling an in-flight request would disconnect, which in most cases should close the connection on the server -> I'd expect yarpc or tchannel to close the context at that point, since a response is known to be impossible.

host/query_workflow_test.go
```diff
@@ -148,6 +148,7 @@ func (s *IntegrationSuite) TestQueryWorkflow_Sticky() {
 	}
 	queryResultCh := make(chan QueryResult)
 	queryWorkflowFn := func(queryType string) {
+		time.Sleep(100 * time.Millisecond) // sleep for a short time, otherwise sticky tasklist will not have workers available
```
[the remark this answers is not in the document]
Shaddoll: this is to make sure when the query hit matching, the sticky tasklist had a poller in the past 10 seconds
and in this test, there was no sticky tasklist poller before this line

service/history/task/transfer_active_task_executor.go
```diff
@@ -261,7 +261,21 @@ func (t *transferActiveTaskExecutor) processDecisionTask(
 	// release the context lock since we no longer need mutable state builder and
 	// the rest of logic is making RPC call, which takes time.
 	release(nil)
-	return t.pushDecision(ctx, task, taskList, decisionTimeout)
+	err = t.pushDecision(ctx, task, taskList, decisionTimeout)
+	if _, ok := err.(*types.StickyWorkerUnavailableError); ok {
+		// sticky worker is unavailable, switch to non-sticky task list
+		taskList = &types.TaskList{
+			Name: mutableState.GetExecutionInfo().TaskList,
+		}
```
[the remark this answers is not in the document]
Shaddoll: By default, it uses normal tasklist. I went through our code, and we don't set the tasklist type explicitly if it's not sticky.

Shaddoll:
> I'm also curious about this:
> 
> > tchannel doesn't propagate context cancel from client to server, which fails the new integration test
> 
> How is that addressed in this PR? And how does it not-pass cancellation? I assume the client canceling an in-flight request would disconnect, which in most cases should close the connection on the server "soon" -> I'd expect yarpc or tchannel to close the context at that point, since a response is known to be impossible.

resolved by using grpc.
If the client cancel the context, the server context is not cancelled and the long poll will not stop immediately. That's what I observed when using tchannel. And it looks like tchannel doesn't propagate cancel: https://github.com/uber/tchannel-go/blob/dev/relay_test.go#L895

service/matching/matchingEngine.go
Shaddoll: Is sticky schedule to start configurable?

Shaddoll: Does that affect polling interval? If they set it to 5 days, and the tasklist didn't receive any poller for a period, we assume the worker is down and return the error. 10s should be fine even for long timeout.
Is the polling interval something configurable?

service/history/task/transfer_standby_task_executor.go
```diff
@@ -186,7 +192,7 @@ func (t *transferStandbyTaskExecutor) processDecisionTask(
 		if decisionInfo.StartedID == common.EmptyEventID {
 			return newPushDecisionToMatchingInfo(
 				decisionTimeout,
-				types.TaskList{Name: transferTask.TaskList},
+				types.TaskList{Name: executionInfo.TaskList}, // at standby, always use non-sticky tasklist
```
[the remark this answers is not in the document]
Shaddoll: I didn't understand your question here.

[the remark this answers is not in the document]
davidporter-id-au: +1 to incomprehension :D

davidporter-id-au (approved):
lgtm more or less, with the caveat that I'm not at all familiar with this code

service/history/historyEngine.go
```diff
@@ -1291,7 +1291,19 @@ func (e *historyEngineImpl) queryDirectlyThroughMatching(
 			scope.IncCounter(metrics.DirectQueryDispatchStickySuccessCount)
 			return &types.HistoryQueryWorkflowResponse{Response: matchingResp}, nil
 		}
-		if yarpcError, ok := err.(*yarpcerrors.Status); !ok || yarpcError.Code() != yarpcerrors.CodeDeadlineExceeded {
+		switch v := err.(type) {
+		case *types.StickyWorkerUnavailableError:
```
davidporter-id-au, line 1295: I don't know what should occur here, should we be logging this scenario?

davidporter-id-au:
contingent stamp based on results of https://github.com/uber/cadence/pull/4871/files#r906360008

service/history/historyEngine.go
Shaddoll: this case will be logged in line 1347

service/matching/matchingEngine.go
Groxx: ah, yea, I guess "time in queue" doesn't really affect "time spent after being started" does it.  👍

service/history/task/transfer_standby_task_executor.go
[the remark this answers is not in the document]
Groxx: `transferTask` is "verified" a couple lines above.
and then a presumably-not-verified value is used instead.

why is a not-verified value better?  and/or should `executionInfo` be verified too?

PR #4874, opened 2022-06-22
Add support for changing the gocql connect timeout

johndelcastillo:
**What changed?**
While testing the cadence-cassandra-tool locally against a remote cassandra cluster, I was observing some timeouts due to the default value being too small for clusters far away (Australia -> US_WEST).

To avoid that, we need to be able to overwrite the connect timeout value that the gocql driver exposes.

**Why?**
Otherwise the tool doesn't work for clusters that take longer than the 600ms default.

**How did you test it?**
Built locally and tested from Aus -> US cluster

**Potential risks**
Shouldn't really be any, I made the default timeout 2s instead of 600ms, and you can make it whatever you want.

**Release notes**
Added support for setting connect timeout for the cassandra connection

**Documentation Changes**


tools/cassandra/cqlclient.go
```diff
@@ -56,9 +57,10 @@ type (
 var errGetSchemaVersion = errors.New("failed to get current schema version from cassandra")
 
 const (
-	DefaultTimeout       = 30 // Timeout in seconds
-	DefaultCassandraPort = 9042
-	SystemKeyspace       = "system"
+	DefaultTimeout        = 30 // Timeout in seconds
+	DefaultConnectTimeout = 2  // Connet timeout in seconds
```
just-at-uber, line 61: typo in comment

davidporter-id-au (approved):
lgtm

PR #4876, opened 2022-06-23
Removed TaskID from types.HistoryTaskV2Attributes

vytautas-karpavicius:
**What changed?**
Removed `TaskID` field from `types.HistoryTaskV2Attributes`

**Why?**
Not actually used anywhere. Instead we have another `SourceTaskID` field which is set on `types.ReplicationTask` that is actually used. It caught my attention as this was set to suspicious value in TaskAckManager:
```
    TaskID:              task.FirstEventID,
```
After checking, it looks like do not care about this field in the other side of replication stack.

**How did you test it?**
Existing tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/types/mapper/proto/shared.go
+++ b/common/types/mapper/proto/shared.go
@@ -592,7 +592,6 @@ func FromHistoryTaskV2Attributes(t *types.HistoryTaskV2Attributes) *adminv1.Hist
 		return nil
 	}
 	return &adminv1.HistoryTaskV2Attributes{
-		TaskId:              t.TaskID,
 		DomainId:            t.DomainID,
 		WorkflowExecution:   FromWorkflowRunPair(t.WorkflowID, t.RunID),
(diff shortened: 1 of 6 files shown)


Shaddoll (approved):

davidporter-id-au (approved):

davidporter-id-au:
Is there a corresponding IDL change being made? I'm not super familiar with this type, sorry

Groxx:
yeah, should probably be marked as deprecated at the very least.

and I assume(?) this is server-side only?  or do we have to worry about old clients / old servers during an upgrade?

vytautas-karpavicius:
> yeah, should probably be marked as deprecated at the very least.

Yeah, good idea: https://github.com/uber/cadence-idl/pull/113

> and I assume(?) this is server-side only? or do we have to worry about old clients / old servers during an upgrade?

It is server side only. Specifically part of replication stack.

PR #4879, opened 2022-06-27
[codegen] introduce gowrap for generating retryableClient

shijiesheng:
**What changed?**

* added gowrap package
* use codegen for retryableClient (file name is unchanged for diff comparison)
* added the tool in the Makefile

**Why?**
avoid mechanical changes for decorators. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/archiver/s3store/mocks/S3API_mock.go
```diff
@@ -0,0 +1,25 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package mocks
+
+//go:generate mockery --dir ../../../../vendor/github.com/aws/aws-sdk-go/service/s3/s3iface --name S3API --output .
```
Groxx, line 25: seems kinda unrelated?  also may be better to stick to a single mock generator tool, unless there some major benefit of mockery vs mockgen. 

I can never remember which does what, maybe mockery would be worth adopting?  just doesn't seem related to gowrap / should probably be a separate thing.

client/frontend/retryableClient.go
```diff
@@ -51,584 +58,409 @@ func NewRetryableClient(
 	}
 }
 
-func (c *retryableClient) DeprecateDomain(
-	ctx context.Context,
-	request *types.DeprecateDomainRequest,
-	opts ...yarpc.CallOption,
-) error {
+func (c *retryableClient) CountWorkflowExecutions(ctx context.Context, cp1 *types.CountWorkflowExecutionsRequest, p1 ...yarpc.CallOption) (cp2 *types.CountWorkflowExecutionsResponse, err error) {
```
Groxx, line 61: I kinda wonder if it'd be worth sorting these before regenerating so they mostly line up...

common/archiver/s3store/mocks/S3API_mock.go
shijiesheng: I'll put it in a separate diff. Unfortunately, we use mockery in multiple places in the code.

docker/buildkite/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13.3
+FROM golang:1.17
```
davidporter-id-au, line 1: do you mind if we pull these out into separate diffs too? sorry to be annoying

shijiesheng: Yes. https://github.com/uber/cadence/pull/4889 I'm fixing some test failures.

common/archiver/s3store/mocks/S3API_mock.go
shijiesheng: It's merged in https://github.com/uber/cadence/pull/4887

docker/buildkite/Dockerfile
shijiesheng: it's merged.

client/frontend/template/retry
```diff
@@ -0,0 +1,70 @@
...
+            backoff.WithRetryableError(isRetryable),
+        ),
+    }
+}
+
+{{range $method := .Interface.Methods}}
+    {{if or
+        (eq $method.Name "CountWorkflowExecutions")
+        (eq $method.Name "GetSearchAttributes")
+        (eq $method.Name "RefreshWorkflowTasks")
```
Groxx, line 40: alrighty, from a manual check: yea, only these 3 changed, now no longer do retries.

do we not want retries on these for some reason though?  or are these maybe just leftovers from developing the template?

go.mod
```diff
@@ -56,119 +57,290 @@ require (
...
-	gopkg.in/yaml.v2 v2.2.8
+	gopkg.in/yaml.v2 v2.4.0
 )
 
 require (
-	cloud.google.com/go v0.56.0 // indirect
-	cloud.google.com/go/bigquery v1.6.0 // indirect
-	cloud.google.com/go/pubsub v1.3.1 // indirect
-	github.com/BurntSushi/toml v0.3.1 // indirect
+	4d63.com/gochecknoglobals v0.1.0 // indirect
```
Groxx, line 74: (to pick a random line)
this kinda seems like a ridiculous increase in dependencies D:  are they all necessary, or could we re-do this without a `-u` / whatever pulled in so many?

either that or we should probably find out what dependency is leading to this bloat, because wow is that a lot just for doing automatic code wrapping.  +50%, though it probably has no impact on our built binaries:
```
❯ du -h vendor | tail -n1
 96M	vendor
❯ git checkout master
❯ rm -rf vendor
❯ go mod vendor
❯ du -h vendor | tail -n1
 64M	vendor
```

client/frontend/retryableClient.go
Groxx: sorted and checked by hand: all changes look good / obviously identical, tho those 3 explicit exceptions are a bit of a❓.

go.mod
Groxx: tackled in https://github.com/uber/cadence/pull/4895
should be easy to adopt those changes here whenever that's done.

client/frontend/template/retry
shijiesheng: There used to have exceptions but now it seams they are all gone.

go.mod
shijiesheng: It's much cleaner now!

internal/tools/go.mod
```diff
@@ -6,63 +6,272 @@ require (
 	github.com/dmarkham/enumer v1.5.5
 	github.com/gogo/protobuf v1.3.2
 	github.com/golang/mock v1.6.0
+	github.com/hexdigest/gowrap v1.2.5
 	github.com/mattn/goveralls v0.0.11
 	github.com/mgechev/revive v1.2.1
 	github.com/uber/tcheck v1.1.0
-	github.com/vektra/mockery/v2 v2.14.0
+	github.com/vektra/mockery/v2 v2.0.3
```
shijiesheng, line 13: This was modified in https://github.com/uber/cadence/pull/4895 and actually v2.14.0 requires go version 1.18 which we don't support yet. So I'm downgrading it here.

davidporter-id-au (approved):

client/frontend/template/retry
Groxx: yea.  can't really see why these would be special tbh.  seems harmless to redo them on failures.

internal/tools/go.mod
Groxx: thank you!  I was using 1.18, didn't notice.

PR #4881, opened 2022-06-27
Move visibility operation from search attributes to indexer message

duoertai:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


duoertai:
@longquanzheng

longquanzheng (approved):

common/persistence/elasticsearch/esVisibilityStore.go
```diff
@@ -772,6 +781,10 @@ func createVisibilityMessage(
 		Version:     common.Int64Ptr(taskID),
 		Fields:      fields,
 	}
+
```
longquanzheng, line 784: How about add a default branch in the switch ? Also I think we should just panic in this case

longquanzheng (approved):

PR #4883, opened 2022-06-27
Removed replication mocks

vytautas-karpavicius:
**What changed?**
Removed all replication mocks.

**Why?**
They are either not used at all or used internally in a couple of places.
For them we can use light-weigh fakes instead and have less code.

**How did you test it?**
Existing tests should pass.

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx:
While I'm largely in favor of this, one advantage of lines like
```go
	s.taskExecutor.EXPECT().execute(replicationTask, true).Return(0, nil).Times(1)
```
that is being lost is the ability to _require_ that those mocks are used.  the current fakes are silent about non-use.

Which may be entirely reasonable for those tests, I haven't read closely enough yet.  Would there be any simple way to require them though?  Unused fakes leading to gradual drift and misunderstanding of tests has been a pretty big problem in past projects in my experience, and the "must be used" check is a pretty simple way to lock that down.

davidporter-id-au (approved):
Tbh I'm not sure this is significantly easier to maintain fakes in the long-term, but 👍

Groxx:
> Some thoughts in general. Cadence seems to be build in a very "Java" like style. Interfaces comes first, necessary or not. Along with mocks. It seems to be contrary to many best practices offered by go-lang. Specifically to this discussion: return structs, accept interfaces that are often defined where they are used and only with a subset of actually needed methods. Smaller interfaces are easily faked, making unit-tests easier to write. IMO it is very counter-productive to write unit-tests where 90% of the test is mocking complex dependencies. Small interfaces along with fakes, makes unit testing much easier. I would very much like us go into that direction.
> 
> Of course mocks still have their place. Especially for Facade style interfaces exposed by packages. One good example - all persistence managers and their corresponding mocks. However, all dependencies that can be modeled as a state or used completely within a single package internally can probably be simplified as fakes instead.
> 
> Feel free to argue on those topics, I'm interested in other opinions as well.

I pretty much completely agree, and yeah - the up-front interface-and-mock habits lead to very artificial tests.  I'd much prefer to mock less, further away from the code actually being tested, because it means you're getting additional checks of "this leads to behavior X which is the actual goal of this method, even though it's 3 calls away".

---

Re fakes and "drift": largely two things.

1. fakes tend to become less and less accurate in being *accurate* mimics of the thing they're faking, unless there's some way to _automatically_ check that they produce identical(-enough) behavior.
2. fakes *frequently* do not require that they are actually used / that data you put in them is actually relevant to the test.  E.g. if you put record X in a fake storage, it *should* (usually) be a test failure if X is not ever looked at.  It implies your test code is not accurately describing what happens.  I've consistently seen heavy use of fakes turn into "this test prepares hundreds of fake records, less than 1% of which are relevant, and ~10% are not used by any test in the entire project".

^ and those often combine to create a gradual loss of accuracy.  A lot of which can be traced back to their frequent inability to ensure they're relevant.  
Mocks have the same issues (and some that are much worse), but the automated check of "expected X, X must happen" prevents a fairly large category of mistakes very simply.

None of which is unfixable.  I just want us to establish a strong habit of e.g. "if you have a fake in a test, it must be relevant", with only rare exceptions.  
And ideally a way to automatically check that the fake behaves the same as the real thing, though I realize that's frequently not worth the effort.

PR #4886, opened 2022-06-29
Add JitterDelay option when creating workflows.

ZackLK:
**What changed?**
Added a JitterStart parameter for the customer to specify a # of seconds to jitter the start of their workflow.

**Why?**
When customers create a large # of workflows and specify they should all start at the same time, it can cause a thundering herd problem. This parameter allows the start time to be jittered so that the large # of workflows can be started at slightly different times.

**How did you test it?**
Unit test and tested manually via CLI:
  ./cadence --domain samples-domain workflow start --tl helloWorldGroup --wt main.helloWorldWorkflow --et 30 --input '"Jittered"' --jitter_start_seconds 5

**Potential risks**

**Release notes**

**Documentation Changes**


common/util.go
```diff
@@ -530,6 +530,12 @@ func CreateHistoryStartWorkflowRequest(
 		firstDecisionTaskBackoffSeconds += delayStartSeconds
 	}
 
+	// Add a random jitter to start time, if requested.
+	jitterStartSeconds := startRequest.GetJitterStartSeconds()
+	if jitterStartSeconds > 0 {
+		firstDecisionTaskBackoffSeconds += rand.Int31n(jitterStartSeconds + 1)
```
davidporter-id-au, line 536: rand has a seed value somewhere?

davidporter-id-au (commented):
No worries with this diff as it's self-evident what's going on, but I'd normally suggest pulling the IDL changes into a separate diff/PR so as to not obscure any changes

common/util.go
```diff
@@ -530,6 +530,12 @@ func CreateHistoryStartWorkflowRequest(
 		firstDecisionTaskBackoffSeconds += delayStartSeconds
 	}
 
+	// Add a random jitter to start time, if requested.
```
emrahs, line 533: Jitter should also apply to each occurrence of a workflow when there is a cron schedule. Otherwise, only the first run of a cron workflow will benefit from the jitter.

common/util.go
ZackLK: Good call - turns out we already seed rand in host/service.go for every service.

davidporter-id-au: perfect

common/util.go
ZackLK: I believe this is now working - see my latest commit.

ZackLK:
Tested new cron changes via:
./cadence --domain samples-domain workflow start --tl helloWorldGroup --wt main.helloWorldWorkflow --et 30 --input '"Jittered"' --jitter_start_seconds 5 --cron '@every 10s'

(in cadence-samples) bin/helloworld -m worker 

And noting that the workflow doesn't run every 10 seconds, but rather runs between 10-15 seconds every 10 seconds.

CONTRIBUTING.md
```diff
@@ -215,6 +215,12 @@ go test -v <path> -run <TestSuite> -testify.m <TestSpercificTaskName>
 go test -v github.com/uber/cadence/common/persistence/persistence-tests -run TestVisibilitySamplingSuite -testify.m TestListClosedWorkflowExecutions
 ```
 
+## IDL Changes
+
+If you make changes in the idls submodule and want to test them locally, you can easily do that by using go mod to use the local idls directory instead of github.com/uber/cadence-idl. Temporarily add the following to the bottom of go.mod:
+
+```replace github.com/uber/cadence-idl => ./idls```
```
demirkayaender, line 222: 👍

common/backoff/cron.go
```diff
@@ -63,13 +69,24 @@ func GetBackoffForNextSchedule(cronSchedule string, startTime time.Time, closeTi
 	}
 	backoffInterval := nextScheduleTime.Sub(closeUTCTime)
 	roundedInterval := time.Second * time.Duration(math.Ceil(backoffInterval.Seconds()))
-	return roundedInterval
+
+	var jitter time.Duration
+	if jitterStartSeconds > 0 {
```
demirkayaender, line 74: Could be good to validate jitterStartSeconds somewhere to make sure its max value is lower than the time till next schedule

demirkayaender: Similarly, we should return an error when it's negative.

common/backoff/cron_test.go
```diff
@@ -61,8 +61,39 @@ func TestCron(t *testing.T) {
...
+var cronWithJitterStartTests = []struct {
+	cron               string
+	startTime          string
+	jitterStartSeconds int32
+	result             time.Duration
+}{
+	{"* * * * *", "2018-12-17T08:00:00+00:00", 10, time.Second * 60},
+	{"* * * * *", "2018-12-17T08:00:10+00:00", 30, time.Second * 50},
+	{"* * * * *", "2018-12-17T08:00:25+00:00", 5, time.Second * 35},
+	{"* * * * *", "2018-12-17T08:00:59+00:00", 45, time.Second * 1},
```
demirkayaender, line 79: We should try 60+ here and expect error

common/backoff/cron.go
```diff
@@ -45,7 +46,12 @@ func ValidateSchedule(cronSchedule string) error {
 
 // GetBackoffForNextSchedule calculates the backoff time for the next run given
 // a cronSchedule, workflow start time and workflow close time
-func GetBackoffForNextSchedule(cronSchedule string, startTime time.Time, closeTime time.Time) time.Duration {
+func GetBackoffForNextSchedule(
+	cronSchedule string,
+	startTime time.Time,
+	closeTime time.Time,
+	jitterStartSeconds int32,
+) time.Duration {
```
demirkayaender, line 54: if you are moving the jitter to this function, you might want to change the return type to `(error, time.Duration)`

service/frontend/workflowHandler.go
```diff
@@ -2084,6 +2085,10 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 		return nil, wh.error(errInvalidDelayStartSeconds, scope, tags...)
 	}
 
+	if startRequest.GetJitterStartSeconds() < 0 {
+		return nil, wh.error(errInvalidJitterStartSeconds, scope, tags...)
```
demirkayaender, line 2089: 👍

service/frontend/workflowHandler.go
```diff
@@ -2084,6 +2085,10 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
...
+	}
+
```
demirkayaender, line 2091: We can add the `jitter < cron_interval` check below

demirkayaender (commented):
Could you check if we have a test for next cron start? Based on the code it looks like next cron schedule should consider jitter again (expected behavior) as opposed to delay_start not being considered (expected behavior).

If this is easy to test, we should test; it can simply be breakable with a code change later.

davidporter-id-au (approved):

PR #4889, opened 2022-06-30
upgrade go version to 1.17 in go mod and Buildkite dockerfile

shijiesheng:
**What changed?**
* go version for buildkite docker
* go mod tidy

For some packages, it requires newer go version. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,8 @@ test_v =
 endif
 
 # a literal space value, for makefile purposes
-SPACE :=
-SPACE +=
+null  :=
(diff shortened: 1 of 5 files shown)


Groxx (approved):
Would prefer to include the comment in the makefile, since it's a weird construct and there are other common ways to do it.  But feel free to reword or whatever.

Otherwise 👍 looks good!  I'm kinda surprised we only had those 3 mocks wrong tbh.

PR #4891, opened 2022-07-02
Add helper function to list all dynamic config keys used in production

Shaddoll:
**What changed?**
Add helper function to list all dynamic config keys used in production

**Why?**
To provide a way to get all keys used in production excluding keys created for unit tests

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/dynamicconfig/constants.go
```diff
@@ -81,6 +81,30 @@ type (
 	}
 )
 
+// ListAllProductionKeys returns all key used in production
+func ListAllProductionKeys() []Key {
+	result := make([]Key, 0, len(IntKeys)+len(BoolKeys)+len(FloatKeys)+len(StringKeys)+len(DurationKeys)+len(MapKeys))
+	for i := TestGetIntPropertyFilteredByTaskListInfoKey + 1; i < LastIntKey; i++ {
```
davidporter-id-au, line 87: What do you think about also exposing methods for each of these as well which hides this implementation detail of the Last key and similar? 

`ListIntConfigKeys()` etc?

davidporter-id-au (approved):

common/dynamicconfig/constants.go
Shaddoll: I don't find a use case for listing keys by type now. I can add that later if I find a use case

PR #4894, opened 2022-07-05
Refactor task ack manager

vytautas-karpavicius:
**What changed?**
Completed refactoring of `replication.TaskAckManager` with the intention to separate message hydration part out of it. I’m planning to experiment with other replication models that will reuse some hydration logic but use different replication message distribution/delivery mechanisms.
- Moved message hydration part out to `task_hydrator.go`. It depends on mutable state provider and history provider. Currently they are implemented as loaders from cache and database respectively. With plans to also have direct in-memory providers.
- Moved replication task reading logic out to `task_reader.go` along with dynamic batch size logic.
- Replaced some of the dependencies with small private interfaces, so that each component could be easily and fully tested. All refactored code in now 100% unit-test covered.

Changes outside of refactoring:
- Removed `ReplicationTaskProcessorReadHistoryBatchSize` dynamic config. Default value was 5. It was used to change page size when loading history during message hydration. Typically there should always be one record returned. We also have a sanity check for that. So in theory it is always enough to fetch at most two records. Also no pagination loop is needed (removed as well).
- Logic to determine dynamic batch size never worked. Noticed that while writing new unit tests. It used `taskInfo.GetVisibilityTimestamp()` which always returned `time.Time{}`. Preventing an update of `lastTaskCreationTime` field and in turn always using defaultBatchSize. Fixed that by referencing `CreatedTime` directly.

**How did you test it?**
- Full unit test coverage for all refactored parts.
- Existing integration tests.

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/replication/task_hydrator.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// Hydrate will enrich replication task with additional information from mutable state and history events.
+func (h TaskHydrator) Hydrate(ctx context.Context, task persistence.ReplicationTaskInfo) (*types.ReplicationTask, error) {
+	switch task.TaskType {
+	case persistence.ReplicationTaskTypeFailoverMarker:
+		return hydrateFailoverMarkerTask(task), nil
+	}
+
+	ms, release, err := h.msProvider.GetMutableState(ctx, task.DomainID, task.WorkflowID, task.RunID)
```
Shaddoll, line 79: release function isn't called in all code paths. Should we do something like
```
defer func() {
  if release != nil {
    release(...)
  }
}()
```

service/history/config/config.go
```diff
@@ -272,7 +272,6 @@ type Config struct {
 	ReplicationTaskProcessorNoTaskRetryWait            dynamicconfig.DurationPropertyFnWithShardIDFilter
 	ReplicationTaskProcessorCleanupInterval            dynamicconfig.DurationPropertyFnWithShardIDFilter
 	ReplicationTaskProcessorCleanupJitterCoefficient   dynamicconfig.FloatPropertyFnWithShardIDFilter
-	ReplicationTaskProcessorReadHistoryBatchSize       dynamicconfig.IntPropertyFn
```
Shaddoll, line 275: Could you delete the definition of the dynamic property from `common/dynamicconfig/constants.go`?

service/history/replication/task_reader.go
```diff
@@ -0,0 +1,106 @@
...
+		BatchSize:    r.getBatchSize(),
+	}
+	response, err := r.executionManager.GetReplicationTasks(ctx, &request)
+	if err != nil {
+		return nil, false, err
+	}
+
+	var lastTaskCreationTime time.Time
+	for _, task := range response.Tasks {
+		creationTime := time.Unix(0, task.CreationTime)
```
Shaddoll, line 79: nice catch

service/history/replication/task_hydrator.go
```diff
@@ -0,0 +1,265 @@
...
+	// only get the first batch
+	return h.getEventsBlob(ctx, task.NewRunBranchToken, common.FirstEventID, common.FirstEventID+1)
+}
+
+func (h historyLoader) getEventsBlob(ctx context.Context, branchToken []byte, minEventID, maxEventID int64) (*types.DataBlob, error) {
+	resp, err := h.history.ReadRawHistoryBranch(ctx, &persistence.ReadHistoryBranchRequest{
+		BranchToken: branchToken,
+		MinEventID:  minEventID,
+		MaxEventID:  maxEventID,
+		PageSize:    2, // Load more than one to check for data inconsistency errors
```
Shaddoll, line 229: nice!

Shaddoll (approved):

PR #4895, opened 2022-07-06
Separating tools dependencies from main dependencies

Groxx:
Adding gowrap in #4879 pulled in an additional *30MB* of dependency source files, which is
pretty ridiculous.  It's happening because, while gowrap uses a tools-tagged `tools.go`,
it's in the same module as everything else, so its dependencies are pulled in too, because
Go has no way to tell that it's not needed by callers (it needs to consider ALL tags).

We're doing the same thing.  Granted, we're an application and not a library, but it still
seems kinda unnecessary.  And it forces quite a large number of upgrades we may not want to
force (yet).

So this moves `tools.go` to a new package, sets up a new `go.{mod,sum}` just for it, and adds
a bit of automation to prevent it from drifting too far from the main module.
With this, `go mod tidy` in the top level does not consider tools-dependencies any more,
so we actually *remove* some, and do not add tons more or force upgrades to pull in gowrap/etc
in the future.

---

internal/tools is just because it needs to be in a separate package, and "tools" was taken.
plus nobody should ever import it, so it should be a clear "dependency dead-end".


Makefile
```diff
@@ -110,7 +111,19 @@ LINT_SRC := $(filter-out %_test.go ./.gen/%, $(ALL_SRC))
 # downloads and builds a go-gettable tool, versioned by go.mod, and installs
 # it into the build folder, named the same as the last portion of the URL.
 define go_build_tool
-@echo "building $(or $(2), $(notdir $(1))) from $(1)..."
+@echo "building $(or $(2), $(notdir $(1))) from internal/tools/go.mod..."
+@go build -mod=readonly -modfile=internal/tools/go.mod -o $(BIN)/$(or $(2), $(notdir $(1))) $(1)
```
davidporter-id-au, line 115: won't this fail your `@` rule?

go.sum
```diff
@@ -20,7 +20,6 @@ cloud.google.com/go/bigquery v1.6.0/go.mod h1:hyFDG0qSGdHNz8Q6nDN8rYIkld0q/+5uBZ
 cloud.google.com/go/datastore v1.0.0/go.mod h1:LXYbyblFSglQ5pkeyhO+Qmw7ukd3C+pD7TKLgZqpHYE=
 cloud.google.com/go/datastore v1.1.0 h1:/May9ojXjRkPBNVrq+oWLqmWCkr4OU5uRY29bu0mRyQ=
 cloud.google.com/go/datastore v1.1.0/go.mod h1:umbIZjpQpHh4hmRpGhH4tLFup+FVzqBi1b3c64qFpCk=
-cloud.google.com/go/firestore v1.1.0/go.mod h1:ulACoGHTpvq5r8rxGJ4ddJZBZqakUQqClKRT5SZwBmk=
```
davidporter-id-au, line 23: Biiig red diff, this is nice!

scripts/check-gomod-version.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+if [[ $2 = "-v" ]]; then
+    v () {
+        >&2 echo "$@"
```
davidporter-id-au, line 6: you didn't want to just `set -x`?

davidporter-id-au (approved):
nice!

Makefile
Groxx: if merged without updating it, yes.
stacked PRs are not a thing github supports though, so I have to either include its changes in this PR, or wait for it to merge first.

scripts/check-gomod-version.sh
Groxx: I think I was initially thinking that 🤔
lemme check.  if it shows the output, yea, I'll just `-x` it.  it's even better for troubleshooting, and it's not a ton of output.

Groxx: yep, `-x` ftw.  thanks!

internal/tools/tools.go
```diff
@@ -37,6 +37,8 @@ import (
 	_ "golang.org/x/tools/cmd/goimports"
 	// mockgen for generating mocks
 	_ "github.com/golang/mock/mockgen"
+	// gowrap for generating decorators for interface
+	_ "github.com/hexdigest/gowrap"
 	// mockery for generating mocks
```
Groxx, line 42: oh, actually, I didn't mean to include this one.  it was just for testing.
lemme undo that.

common/thrift_util.go
```diff
@@ -1,117 +0,0 @@
...
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-
-package common
-
-import (
-	"reflect"
-
-	"github.com/apache/thrift/lib/go/thrift"
```
Groxx, line 26: surprisingly, this is the only import of apache/thrift in the server.  the rest comes from importing the client.

might be worth exploring that, to see if we can get rid of it entirely.... but perhaps not.  either way, this file was totally unused, and it moved a problematic dependency to only-indirect, so I figured I'd include it.

Makefile
Groxx: updated!

davidporter-id-au (approved):

PR #4896, opened 2022-07-06
Minor makefile cleanup, verbose CI, fmt with a recent Go version

Groxx:
Due to some unnecessary complexity troubleshooting a buildkite failure recently,
I'm declaring "no more @ use", period.  Knowing the full command that has failed
is extremely useful, and this is an easy way to achieve it.

Also:
- made some vars slightly more readable?
- make the Makefile verbose in all buildkite jobs
- `make fmt` with Go 1.18 because it apparently changes goimports' behavior.
  older Go versions do not undo this, so it should be stable.

---

@-failures are detected immediately on any makefile use, and look pretty good imo:
```
❯ make
Makefile:75: Makefile cannot use @ to silence commands, use $Q instead:
Makefile:76: found on line 543:	@echo Running integration test for 3+ dc with $(PERSISTENCE_TYPE) $(PERSISTENCE_PLUGIN)
Makefile:77: *** fix that line and try again.  Stop.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,7 +275,7 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 # Rule-breaking targets intended ONLY for special cases with no good alternatives.
 # ====================================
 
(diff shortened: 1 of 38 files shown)


davidporter-id-au (approved):
seems reasonable

PR #4898, opened 2022-07-08
Standardizing cancellation behavior: a canceled workflow never starts a new run

Groxx:
# Summary for busy people

Workflow cancellation was kinda weird and confusing, and left some awful, unrecoverable, and un-*preventable* edge cases (particularly with child workflows).  It also left users with no way to reliably stop work, aside from termination.  Termination is inherently "unclean" and risky, so it should not be required to achieve something outside exceptional circumstances where recovery is not possible.

This commit changes that: cancellation is now "sticky", and a canceled workflow does not ever trigger a new run after it completes, regardless of how it completes, so it can be used as a reliable "stop processing after cleanup" tool.  The final state of a canceled workflow's run is now *always* a successful completion with a value, canceled, or timed out. (termination remains always "terminated")  
A canceled workflow can still start and abandon child workflows, so all current behavior with retries / continue as new / etc can be replicated with child workflows if desired.

A fair bit of (not very complex) additional work here and in nearly all other repos is required to truly complete this, but it is *functional* and non-optional with this commit alone.  
In particular, adding a dynamic config to (temporarily!) restore old behavior should be fairly easy if it proves to be needed.

# More details and motivation

Part 1 of [many, tbd, in multiple repos] involved in changing workflow cancellation to reliably end workflows.
Tests will be coming soon, for now I'm using a fairly simple set of workflows and checking the resulting histories exhaustively by hand.

The primary motivation for these changes is to address some impossible-to-recover-from scenarios when canceling child workflows.  After further exploration and discussion we've realized that, without these changes, there is *no reliable way* to stop a sequence of workflows without relying on termination, which we consistently treat as a fallback / impure-but-necessary ultimate hammer.

Workflows should not *need* to rely on termination to achieve a desired behavior.  With these changes, cancellation becomes capable of *guaranteeing* that workflows end within some finite time, which is a unique ability and makes it much more consistent and reliable.  
Turning this into a "complete" change will require quite a few tests, documentation changes, client-side changes (to allow recording more info, and likely changing test suites), and some smallish database and maybe RPC changes (to hold/return more data in cancellation errors).

We are also not currently planning on making this configurable.  It's seen as a correction of an under-specified and somewhat flawed chunk of behavior, more than "a change".  
Existing workflows will not experience replay errors, but it is still a substantial *semantic* change, though from what we have seen cancellation is relatively rarely used (partly due to its complex behavior).  If issues are encountered / if users end up needing it, it should be fairly easy to add a per-domain/tasklist/workflow type configuration value, but it will be opt-*out*, not opt-*in*.

# What was happening

Previously, workflow behavior on cancellation was pretty frequently surprising to our users, arguably inconsistent, and not very well documented:

| **PREVIOUS**  | **simple**               | **retry**                                 | **cron**                                | **retry+cron**                                    |
|--------------:|--------------------------|-------------------------------------------|-----------------------------------------|---------------------------------------------------|
| **success**   | success                  | success                                   | success<br>continue cron<br>cron        | success<br>continue cron<br>cron<br>retry         |
| **cancel**    | canceled                 | canceled                                  | canceled                                | canceled                                          |
| **retryable** | (n/a, fatal)             | continue retry<br>retry<br>recorded error | (n/a, fatal)                            | continue retry<br>cron<br>retry<br>recorded error |
| **fatal**     | failed<br>recorded error | failed<br>recorded error                  | continue cron<br>cron<br>recorded error | continue cron<br>cron<br>retry<br>recorded error  |
| **continue**  | continue immediately     | continue immediately<br>retry             | continue immediately                    | continue immediately<br>retry                     |
| **timeout**   | timeout                  | continue retry<br>retry<br>recorded error | continue cron<br>cron<br>recorded error | continue retry<br>cron<br>retry<br>recorded error |

A legend is:
- success / etc shows the final state of the canceled run (success = completed with a value that can be retrieved)
- "continue X" covers what source is used to compute the next run's starting delay (cron, retry, or no delay)
- "cron" / "retry" shows whether or not cron/retry configuration is carried over to the new run
  - note that cron is lost by default with continue-as-new
- and "recorded error" is whether or not the returned error is saved in its entirety (type + reason + details)

This largely summarizes as "cancellation works when you end with the canceled-context error", say from `ctx.Err()`, otherwise it behaves like normal (or nearly) and many scenarios will start a new run.
That's somewhat reasonable, but it's fairly "fragile" (it depends on what you return, and there are *many* ways for code to return some other error), and most importantly it means *there is no reliable way to stop a workflow* except to terminate it.

That has severe consequences in at least two scenarios:
1. When termination is *unsafe*
2. When a parent workflow cancels a child by canceling its context

For 1, for manual cancellations it's potentially reasonable to just terminate a run that begins after a successful cancel... but in principle if you're using cancellation it implies that termination is *not* desired, and potentially not safe to do.  Canceling may result in a brand new run that immediately starts new behavior, leaving you with no safe window to terminate and not leave bad state lingering.  
So users wanting a safe way to stop a sequence of workflows have no reliable way to do so.

For 2, it puts parent+child workflows in an extremely awkward, and essentially unrecoverable scenario.  Cancellation is a *one time event*, and as far as the parent is concerned, if the child/its context is canceled, the child is canceled...  
...but if the child then starts a new run for *any* reason (retry, cron, reset, etc), that new run is no longer canceled.  The parent has no way to know this has happened, and has no way to *re*-cancel the new child, so it can easily lead to the collection of workflows getting into an impossible state that it never recovers from.

Both cases are able to lead to unreliable behavior which can only use termination to stop, and for which no "safe" option exists.

After reviewing some customer issues and desires and thinking about things, we've settled on "cancel should guarantee that things stop".  Not necessarily in a timely manner, but that's fine.  And if a workflow wants to run behavior longer or larger than its current run can achieve, it has a workaround: start a new (likely child) workflow to do the cleanup.

# What happens now

So that's what this PR does, in a minimal / to-be-polished way so we can start running it for our stuck users while we flesh out tests and change other behaviors.

Currently that means our cancellation behavior is now:

| **CURRENT**    | **simple**                                | **retry**                                 | **cron**                                  | **retry+cron**                            |
|--------------:|-------------------------------------------|-------------------------------------------|-------------------------------------------|-------------------------------------------|
| **success**   | success                                   | success                                   | success                                   | success                                   |
| **cancel**    | canceled                                  | canceled                                  | canceled                                  | canceled                                  |
| **retryable** | (n/a, fatal)                              | canceled<br>recorded error (details only) | (n/a, fatal)                              | canceled<br>recorded error (details only) |
| **fatal**     | canceled<br>recorded error (details only) | canceled<br>recorded error (details only) | canceled<br>recorded error (details only) | canceled<br>recorded error (details only) |
| **continue**  | canceled<br>(no details)                  | canceled<br>(no details)                  | canceled<br>(no details)                  | canceled<br>(no details)                  |
| **timeout**   | timeout                                   | timeout                                   | timeout                                   | timeout                                   |

And the new "details" entries cover whether or not an error's "details" (the custom encoded data, not reason or type) are saved.  Unfortunately the current cancellation event (and clients' API) does not allow recording all data, or any in some cases, so the original reason/message and error type are lost and are replaced with a canceled error.

Now, cancellation *always* ends workflows with the current run.  Returning a value will return that value, including in cron scenarios, timeouts are still timeouts (and they imply a possibly un-clean termination), and _all_ errors or attempts to continue-as-new will instead result in a canceled state.

# Future changes to make to finish this effort

With further changes to the clients and RPC/storage models, canceled errors will store more details about what was returned.  E.g. continue-as-new does not record what was *attempted* to be started, and other error types lose their "reason" (i.e. the message) and type but not details.  Pretty clearly this is sub-par, and we should be capable of reporting the actual return in full so it can be retrieved if needed.  This is also why returning a value now always ends in a completed state, so successful completions do not lose those values.

Prior to merging into master / a release, we may end up making this configurable (likely with a default of opt-out), to address both the sub-par information recording and the semantically-breaking behavior change.  Docs changes are also due, as well as some integration tests, client library changes (e.g. to make sure the test suite reflects the new behavior), etc.

Another gap to plug is that resetting a workflow does not "forward" the canceled state to the new run.  We should probably be treating cancellation like we do signals: cancel the new run if the current run is canceled.  This will ensure that you can reset a child and retain the parent's cancellation, so it'll very likely become the default behavior, but we'll allow overriding it.  Resets are manual actions, they can break the rules if desired.  And they can just manually cancel later if they decide they do want it.

And last and perhaps least: it's quite strange that continue-as-new does not retain cron config.  At least from the Go client.  I suspect it's just not adding to / pulling from the context correctly.


service/history/execution/mutable_state_builder.go
```diff
@@ -2659,6 +2659,28 @@ func (e *mutableStateBuilder) AddFailWorkflowEvent(
 	return event, nil
 }
 
+func (e *mutableStateBuilder) AddCancelWorkflowEvent(
```
Shaddoll, line 2662: we don't have this function before?

Groxx: Not that I saw anyway.  A bit surprisingly.
I can go search more comprehensively though.

Shaddoll: we have `AddWorkflowExecutionCanceledEvent`, just noticed that you added this but didn't add `ReplicateWorkflowExecutionCanceledEvent `

davidporter-id-au, line 2662: Is this different to `AddWorkflowExecutionCanceledEvent` on line 2757?

service/history/execution/history_builder.go
```diff
@@ -314,6 +314,18 @@ func (b *HistoryBuilder) AddFailWorkflowEvent(decisionCompletedEventID int64,
 	return b.addEventToHistory(event)
 }
 
+// AddCancelWorkflowEvent adds WorkflowExecutionCanceled event to history
+func (b *HistoryBuilder) AddCancelWorkflowEvent(decisionCompletedEventID int64,
```
davidporter-id-au, line 318: Is this different to `AddWorkflowExecutionCanceledEvent` on line 491?

davidporter-id-au:
Looks good, though:
- some minor duplication that I think needs to be cleaned up maybe? 
- It'd be nice to maybe pull out those mocks and formatting changes out, as they obscure the change a bit

Broadly, the diff description is excellent, and of a level the rest of can only aspire to. However, a one-line summary at the top about cancellation behaviour being esoteric and understood by noone might be worthwhile.

service/history/execution/mutable_state_builder.go
Groxx: yeah, we have these methods paired between historybuilder and mutablestatebuilder.  not entirely sure why, I just followed the patterns

service/history/execution/history_builder.go
Groxx: looks like no!  just didn't search for the method names thoroughly enough.
removing this one.

service/history/execution/mutable_state_builder.go
Groxx: ah hah.  yeah, that should've been a sign 🤔 

thanks + @davidporter-id-au , good catches :)

Shaddoll (approved):

service/history/decision/task_handler.go
```diff
@@ -789,6 +806,19 @@ func (handler *taskHandlerImpl) handleDecisionContinueAsNewWorkflow(
 		return nil
 	}
 
+	if is, _ := handler.mutableState.IsCancelRequested(); is {
+		// cancellation must be sticky, as it's telling things to stop.
+		// this is particularly important for child workflows, as if they restart themselves after the parent
+		// cancels its context, there is no way for the parent to cancel the new run.
+		cancelAttrs := types.CancelWorkflowExecutionDecisionAttributes{
+			Details: nil, // TODO: serialize continue-as-new data somehow, may deserve a new field
```
davidporter-id-au, line 814: No need to get fancy, maybe even just a string here or something might be nice to indicate what has occurred

davidporter-id-au (approved):
nice!

service/history/decision/task_handler.go
Groxx: I was considering that, yeah.  think it's worth just putting in a json string for now?  `"ContinueAsNew ignored because workflow is canceled"`?

Groxx: For now, I think I'm going to leave this empty.  Nil is a bit "safer" since we'll be storing structured data here in the future.

If we get enough confusion around this behavior though, even just hardcoding an explanation may be fine.  
When we get an "underlying reason"-like field added, that can be used to differentiate the type of data stored, which will be necessary anyway - a CanceledError could store a ContinueAsNew-details or a FailedError-details or possibly other things too.  An empty "underlying reason" seems like a fairly safe way to say "this is old data and cannot be deserialized that way".

PR #4899, opened 2022-07-09
Add new es record for uninitialized workflow

neil-xie:
Added new visibility record for uninitialized workflow execution to detect stuck workflow.
Added bulk create request to insert the new record so it wouldn't change the status of existing workflows.

**Why?**

**How did you test it?**
Manual test by breaking the start workflow execution process and unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/lib/pq v1.2.0
 	github.com/m3db/prometheus_client_golang v0.8.1
 	github.com/olekukonko/tablewriter v0.0.4
-	github.com/olivere/elastic v6.2.21+incompatible
+	github.com/olivere/elastic v6.2.37+incompatible
 	github.com/olivere/elastic/v7 v7.0.21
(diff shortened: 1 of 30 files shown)


go.mod
```diff
@@ -26,7 +26,7 @@ require (
 	github.com/lib/pq v1.2.0
 	github.com/m3db/prometheus_client_golang v0.8.1
 	github.com/olekukonko/tablewriter v0.0.4
-	github.com/olivere/elastic v6.2.21+incompatible
+	github.com/olivere/elastic v6.2.37+incompatible
```
davidporter-id-au, line 29: In general it'd be better if we do these version upgrades in a standalone diff, as it makes it clearer and easier to see. Probably ok in this diff however.

neil-xie: It is good to know! I will follow the way next time.

davidporter-id-au (approved):

PR #4900, opened 2022-07-12
CDNC-1781 Add restart command/api

allenchen2244:
Added an api and a cli command that restarts a previous workflow with the same parameters

Simple restart and terminate that's different from reset

Tested it locally

Nothing that bad can happen. Only caller is CLI.


tools/cli/app_test.go
```diff
@@ -339,6 +339,20 @@ func (s *cliAppSuite) TestShowHistory_PrintDateTime() {
 	s.Nil(err)
 }
 
+func (s *cliAppSuite) TestRestartWorkflow() {
```
ZackLK, line 342: Are these used somewhere?

ZackLK (commented):
Can you write a test for the RestartWorkflowExecution?

tools/cli/app_test.go
allenchen2244: this is a unit test for the cli

ZackLK: Oh wow, not sure what I was thinking when I wrote that comment. They sure are unit tests for the cli. :D

ZackLK:
> Can you write a test for the RestartWorkflowExecution?

Just to be a bit more clear about this comment of mine - it would be nice if there was a test that made sure that a Restart did what you expect it to do, that wasn't completely mocked out and exercised the new code you wrote.

allenchen2244:
> Can you write a test for the RestartWorkflowExecution?

Test is added

davidporter-id-au (approved):

PR #4903, opened 2022-07-16
Add new CI step for lint validation

shijiesheng:
**What changed?**

Add a new CI step to validate code is linted. 

**Why?**

commands like `make go-generate` could be ignored and cause the build to fail. The new CI step aims to address such issues. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


.buildkite/pipeline-master.yml
```diff
@@ -19,6 +19,21 @@ steps:
...
+    command: "./scripts/buildkite/golint.sh"
+    artifact_paths:
+      - ".build/golint/*.out"
+    retry:
+      automatic:
+        limit: 1
+    plugins:
+      - docker-compose#v3.0.0:
+          run: unit-test
+          config: docker/buildkite/docker-compose.yml
```
davidporter-id-au, line 35: do we need the compose plugins when running the script?

.buildkite/pipeline-master.yml
```diff
@@ -19,6 +19,21 @@ steps:
           run: unit-test
           config: docker/buildkite/docker-compose.yml
 
+  - label: ":golangci-lint: validate code is clean"
+    agents:
+      queue: "workers"
+      docker: "*"
+    command: "./scripts/buildkite/golint.sh"
+    artifact_paths:
+      - ".build/golint/*.out"
```
davidporter-id-au, line 28: do we have any artifacts?

.buildkite/pipeline-master.yml
shijiesheng: Yes, we need the golang docker to run the makefile.

.buildkite/pipeline-master.yml
shijiesheng: Currently no. It's reserved for future usage.

davidporter-id-au (approved):

PR #4905, opened 2022-07-18
Fix flaky retrypolicy tests.

ZackLK:
Fix flaky retry tests.

- Use random seed in test suite to hit the flakiness easier.
- Adjust 2 tests to use the correct remaining times (which I found by
  instrumenting the code to print out the actual remaining time for
  the expiration interval.)

### DIFF
--- a/.buildkite/pipeline-master.yml
+++ b/.buildkite/pipeline-master.yml
@@ -19,6 +19,20 @@ steps:
           run: unit-test
           config: docker/buildkite/docker-compose.yml
 
+  - label: ":golangci-lint: validate code is clean"
+    agents:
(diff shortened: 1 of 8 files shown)


common/backoff/retrypolicy_test.go
```diff
@@ -44,6 +45,7 @@ func TestRetryPolicySuite(t *testing.T) {
 }
 
 func (s *RetryPolicySuite) SetupTest() {
+	rand.Seed(int64(time.Now().Nanosecond()))
```
davidporter-id-au, line 48: oof

davidporter-id-au (approved):

PR #4910, opened 2022-07-21
[testing before updating branch] 0.24.x hotfix

Groxx:
running in buildkite as it's more reliable than locally


Groxx:
pipeline is no longer compatible with 0.24.x apparently 😶 
oh well.  99% of it worked locally, onward to canary environments.

PR #4911, opened 2022-07-22
Fix remaining server lint warnings and make lint error by default.

ZackLK:
- Minor text changes while reading this file
- Ignore lint warning: [defer] recover must be called inside a deferred function
- Fix lint warning: [var-naming] func parameter startedShardId should be startedShardID
- Fix lint warnings: [context-as-argument] context.Context should be the first parameter of a function
- Fix lint warning: [receiver-naming] receiver name mdb should be consistent with previous receiver name pdb for db
- Fix lint warning: [indent-error-flow] if block ends with a return statement, so drop this else and outdent its block
- Fix lint warning: [context-as-argument] context.Context should be the first parameter of a function
- Fix lint warning: [unexported-return] exported func NewDNSPeerChooserFactory returns unexported type rpc.dnsPeerChooserFactory, which can be annoying to use
- Make lint (revive) error by default.

### DIFF
--- a/common/persistence/sql/sqlExecutionStoreUtil.go
+++ b/common/persistence/sql/sqlExecutionStoreUtil.go
@@ -422,6 +422,7 @@ func applyWorkflowMutationAsyncTx(
 	runID := serialization.MustParseUUID(executionInfo.RunID)
 
 	recoverPanic := func(err *error) {
+		// revive:disable-next-line:defer Func is being called using defer().
 		if r := recover(); r != nil {
 			*err = fmt.Errorf("DB operation panicked: %v %s", r, debug.Stack())
 		}
@@ -769,6 +770,7 @@ func (m *sqlExecutionStore) applyWorkflowSnapshotAsyncTxAsNew(
 	runID := serialization.MustParseUUID(executionInfo.RunID)
 
 	recoverPanic := func(err *error) {
+		// revive:disable-next-line:defer Func is being called using defer().
 		if r := recover(); r != nil {
 			*err = fmt.Errorf("DB operation panicked: %v %s", r, debug.Stack())
 		}
--- a/common/rpc/peer_chooser.go
+++ b/common/rpc/peer_chooser.go
@@ -41,7 +41,7 @@ type (
 	}
 )
 
-func NewDNSPeerChooserFactory(interval time.Duration, logger log.Logger) *dnsPeerChooserFactory {
+func NewDNSPeerChooserFactory(interval time.Duration, logger log.Logger) PeerChooserFactory {
 	if interval <= 0 {
 		interval = defaultDNSRefreshInterval
 	}
--- a/docs/non-deterministic-error.md
+++ b/docs/non-deterministic-error.md
@@ -37,13 +37,13 @@ In Cadence, we use "close" to describe the opposite status of “open”. For ac
 
 * The decision state machine must be **Scheduled->Started->Closed** 
 
-* The first decision task is triggered by server. Starting from that, the rest decision task is triggered by some events of workflow itself -- when those events mean something the workflow could be waiting for. For example, Signaled, ActivityClosed, ChildWorkflowClosed, TimerFired events. Events like ActivityStarted won’t trigger decisions.  
+* The first decision task is triggered by server. Starting from that, the rest of the decision tasks are triggered by some events of the workflow itself -- when those events mean something the workflow could be waiting for. For example, Signaled, ActivityClosed, ChildWorkflowClosed, TimerFired events. Events like ActivityStarted won’t trigger decisions.
 
 * When a decision is started(internally called in flight), there cannot be any other events written into history before a decision is closed. Those events will be put into a buffer until the decision is closed -- flush buffer will write the events into history.
 
 * **In executing mode**, a decision task will try to complete with some entities: Activities/Timers/Childworkflows/etc Scheduled. 
 
-* **In history replay mode**, decisions use all of those above to rebuild a stack. **Activities/ChildWorkflows/Timers/etc will not be re-executed during history replay.** 
+* **In history replay mode**, decisions use all of those above to rebuild a stack. **Activities/Timers/ChildWorkflows/etc will not be re-executed during history replay.** 
 
 ### Activity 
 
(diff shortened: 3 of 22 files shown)


common/persistence/sql/sqlExecutionStoreUtil.go
```diff
@@ -422,6 +422,7 @@ func applyWorkflowMutationAsyncTx(
 	runID := serialization.MustParseUUID(executionInfo.RunID)
 
 	recoverPanic := func(err *error) {
+		// revive:disable-next-line:defer Func is being called using defer().
 		if r := recover(); r != nil {
```
Groxx, line 426: this is unfortunately not always true.  there are definitely some incorrect-and-not-working uses of `recoverPanic()` because Go is extremely picky about when and how `recover()` is called.

ZackLK: All current uses of recoverPanic() are being called with defer() - I didn't mean to imply that you couldn't create future cases where you were using it wrong. The issue is that this lint warning doesn't alert at the call site, it alerts at the call to recover(), so it seems pretty useless right now. Do you have a better suggestion than ignoring it?

Groxx: ah, yeah, those look fine.

the one I was thinking of was this, which has apparently already been suppressed: 
https://github.com/uber/cadence/blob/8b42a6dccb725957b61c73dc10902f854c271ee2/common/log/panic.go#L35-L37

which is incorrect, it's used wrong in multiple places.  e.g. 
https://github.com/uber/cadence/blob/8b42a6dccb725957b61c73dc10902f854c271ee2/service/frontend/clusterRedirectionHandler.go#L1219-L1228

which is why I don't like hiding this lint, even though it's rather annoying in this file's correct usage.  it's an _extremely_ error-prone thing to do, and allowing it leaves room for both bad copy-pasta without understanding the details, and incorrect usage later without any warning.

ZackLK: Right, but I don't think what we've got today is the solution. What we've got right now is a lint warning that everyone ignores, that doesn't point at line 1226 above, but rather just points at line 37 and does it regardless of the callers.

After reading and playing more with defer/recover, I'm certainly coming around to your point about not using recover-in-a-helper functions. In the case of CapturePanic(), that's 122 call sites to update and maintain. Ugh.

How about we add our own linter :P
`ag 'CapturePanic\(' | grep -v defer | wc -l
       1`

Groxx: because then we'd have
```go
func... () {
  defer afterCall(...)
}
func afterCall(...) {
  defer log.CapturePanic(..)
```
which is equally wrong :)

but yeah, 100% ugh but it's _so_ error-prone and _so_ hard to notice and _so_ potentially damaging that it's one where I think nearly any amount of pain is worth it.

I'm game to suppress it for now to enforce the other lints, and then follow up with restoring this one later / soon / as long as _it happens_, I'm happy.  I just fear forgetting it forever and then we encounter an easily-prevented outage.

common/rpc/peer_chooser.go
```diff
@@ -41,7 +41,7 @@ type (
 	}
 )
 
-func NewDNSPeerChooserFactory(interval time.Duration, logger log.Logger) *dnsPeerChooserFactory {
+func NewDNSPeerChooserFactory(interval time.Duration, logger log.Logger) PeerChooserFactory {
```
Groxx, line 44: I'm guessing it complained about "private types are annoying to use"?

Any chance this func is actually only used in this package, and can be made private?

docs/non-deterministic-error.md
```diff
@@ -37,25 +37,25 @@ In Cadence, we use "close" to describe the opposite status of “open”. For ac
...
+* The first decision task is triggered by server. Starting from that, the rest of the decision tasks are triggered by some events of the workflow itself -- when those events mean something the workflow could be waiting for. For example, Signaled, ActivityClosed, ChildWorkflowClosed, TimerFired events. Events like ActivityStarted won’t trigger decisions.
 
 * When a decision is started(internally called in flight), there cannot be any other events written into history before a decision is closed. Those events will be put into a buffer until the decision is closed -- flush buffer will write the events into history.
 
 * **In executing mode**, a decision task will try to complete with some entities: Activities/Timers/Childworkflows/etc Scheduled. 
 
-* **In history replay mode**, decisions use all of those above to rebuild a stack. **Activities/ChildWorkflows/Timers/etc will not be re-executed during history replay.** 
+* **In history replay mode**, decisions use all of those above to rebuild a stack. **Activities/Timers/ChildWorkflows/etc will not be re-executed during history replay.**
```
Groxx, line 46: no complaint, just curious: do we use the new order elsewhere?  or just trying to make it flow more nicely?

docs/non-deterministic-error.md
```diff
@@ -65,7 +65,7 @@ In Cadence, we use "close" to describe the opposite status of “open”. For ac
 
 * Local activity is only recorded with DecisionCompleted -- no state machine needed.
 
-* Local activity completed with trigger another decision
+* Local activity completed will trigger another decision
```
Groxx, line 68: not confident this is true actually... but eh.  that's an issue for another time / a rewrite of this doc, probably.

service/history/execution/history_builder_test.go
```diff
@@ -1039,7 +1039,7 @@ func (s *historyBuilderSuite) addActivityTaskScheduledEvent(
 	retryPolicy *types.RetryPolicy,
 	requestLocalDispatch bool,
 ) (*types.HistoryEvent, *persistence.ActivityInfo, *types.ActivityLocalDispatchInfo) {
-	event, ai, activityDispatchInfo, _, _, err := s.msBuilder.AddActivityTaskScheduledEvent(decisionCompletedID,
+	event, ai, activityDispatchInfo, _, _, err := s.msBuilder.AddActivityTaskScheduledEvent(nil, decisionCompletedID,
```
Groxx, line 1042: ick.  nil context 😢
(PR doesn't change anything, the knowledge is just sadness-inducing)

Groxx (approved):
yay enforcement!
thank you!!!

docs/non-deterministic-error.md
ZackLK: Hah good question - if you look at line 44, it uses the ordering that I changed it to, so I was just trying to make lines 44 and 46 line up.

common/rpc/peer_chooser.go
ZackLK: NewDNSPeerChooserFactory() is used by cmd/server/cadence/server.go, so no :(

Yeah I agree about returning a specific type instead of the interface being preferable. Better to ignore the lint warning? Does this code even need an interface that has one impl?

Groxx: the repo as a whole is fairly interface-happy / java-flavored, so in a lot of cases I think the answer is no - no need for an interface.

........ but Go makes that kinda hard with tests.
which is probably more reason to reduce our mocks overall and run more "real" code though :|

PR #4913, opened 2022-07-23
Replace unsafe usage of recover() in helper functions

ZackLK:
We decided that our usage of helper functions when using defer/recover was too brittle, since one
too many functions causes recover() not to do the right thing in Go. Switching to this idiom, where
the defer caller also calls recover() ensures that we will get the right result from recover(), but
also lets us continue to use the helper functions to deduplicate the code that runs after recover().

### DIFF
--- a/common/log/panic.go
+++ b/common/log/panic.go
@@ -31,7 +31,9 @@ import (
 // If the panic value is not error then a default error is returned
 // We have to use pointer is because in golang: "recover return nil if was not called directly by a deferred function."
 // And we have to set the returned error otherwise our handler will return nil as error which is incorrect
-// errPanic MUST be the result from calling recover()
+// errPanic MUST be the result from calling recover, which MUST be done in a single level deep
+// deferred function. The usual way of calling this is:
+// - defer func() { log.CapturePanic(recover(), logger, &err) }()
 func CapturePanic(errPanic interface{}, logger Logger, retError *error) {
 	if errPanic != nil {
 		err, ok := errPanic.(error)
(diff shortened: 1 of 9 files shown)


Groxx:
Now that it's merged, want to see if https://github.com/mgechev/revive/pull/719 can be pulled in easily enough?  `go get -modfile internal/tools/go.mod ...` iirc.  That plus adding `"immediate-recover"` to the defer list in the toml will (hopefully) require that pattern.

common/log/panic.go
```diff
@@ -31,10 +31,11 @@ import (
...
-	// revive:disable-next-line:defer Caller must call from a deferred function
-	if errPanic := recover(); errPanic != nil {
+// errPanic MUST be the result from calling recover, which MUST be done in a single level deep
+// deferred function. The usual way of calling this is:
+// - defer func() { log.CapturePanic(recover(), logger, &err) }()
+func CapturePanic(errPanic interface{}, logger Logger, retError *error) {
+	if errPanic != nil {
 		err, ok := errPanic.(error)
 		if !ok {
 			err = fmt.Errorf("panic object is not error: %#v", errPanic)
```
Groxx, line 41: I'll lightly vote for a rename, but either's clear enough I think.  and there is _some_ benefit in implying something may be panicking error objects and they're handled specially...

eh.  your pick.
```suggestion
// "recovered" MUST be the result from calling recover, which MUST be done in a single level deep
// deferred function. The usual way of calling this is:
// - defer func() { log.CapturePanic(recover(), logger, &err) }()
func CapturePanic(recovered interface{}, logger Logger, retError *error) {
	if recovered != nil {
		err, ok := recovered.(error)
		if !ok {
			err = fmt.Errorf("panic object is not error: %#v", recovered)
```

Groxx (approved):
👍 all LGTM.  And I am glad that gofmt is permissive about this being one line...

PR #4916, opened 2022-07-26
Hotfix for our lack of List*Execution API ratelimiters

Groxx:
ElasticSearch is struggling to scale, this is to give us some more breathing room
because our biggest querier will take longer to reduce their rates.

This is very similar to ea4d16525c2210104c58eea8ab57d46821868a5b , but slightly easier because it is only changing the frontend.
A more complete fix of removing or correcting the now-deprecated fields would be good.

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -378,6 +378,12 @@ const (
 	// Default value: UnlimitedRPS
 	// Allowed filters: N/A
 	FrontendWorkerRPS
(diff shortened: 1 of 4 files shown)


Groxx:
moving to master

davidporter-id-au (approved):

Groxx:
Will rewrite for master and open a new PR, as there are too many conflicting changes to make merging this to both worth the effort.

PR #4917, opened 2022-07-26
Update revive to catch more defer/recover badness

ZackLK:
- Update to latest version of revive that includes Steven's change to find more defer/recover badness.
- Ignore 158 new warnings about package-comments.
- Add Steven's new immediate-recover linter. 
  - Verified this would have caught the previous mistakes I made (woohoo!)


Groxx (approved):
🥳

PR #4918, opened 2022-07-26
downgrade testify to fix monorepo

allenchen2244:
Downgrading testify to fix monorepo error


davidporter-id-au (approved):

Groxx:
iirc this was due to https://github.com/stretchr/testify/commit/a409ccf19e17d8a90156521ce71c5e8ef8f6bca8 adding mutexes, which caused a bunch of "copies lock" vet failures.

PR #4922, opened 2022-07-28
Adding DomainTag to the Persistence metrics client

abhishekj720:
**What changed?**
Added DomainTag to the Persistence metrics client

**Why?**
For creation of metrics to demonstrate customer load

**How did you test it?**
Locally.

**Potential risks**
None

**Release notes**
N/A

**Documentation Changes**
N/A


common/persistence/persistenceMetricClients.go
```diff
@@ -224,7 +224,8 @@ func NewConfigStorePersistenceMetricsClient(
 	}
 }
 
-func (p *persistenceMetricsClientBase) updateErrorMetric(scope int, err error) {
+func (p *persistenceMetricsClientBase) updateErrorMetric(scope int, err error, scopeWithDomainTag ...metrics.Scope) {
+
```
Shaddoll, line 228: It seems that you added a new parameter, but didn't use the parameter in this function.

common/persistence/persistenceMetricClients.go
```diff
@@ -258,7 +259,28 @@ func (p *persistenceMetricsClientBase) updateErrorMetric(scope int, err error) {
 	}
 }
 
-func (p *persistenceMetricsClientBase) call(scope int, op func() error) error {
+func (p *persistenceMetricsClientBase) call(scope int, op func() error, domainName ...string) error {
```
Shaddoll, line 262: For extendability, instead of adding domainName, it's better to add metricsTag, so that if we want to add other tags in the future we don't need to change this

common/reconciliation/invariant/historyExists.go
```diff
@@ -66,13 +68,19 @@ func (h *historyExists) Check(
 			Info:            "failed to check: expected concrete execution",
 		}
 	}
+	domainID := concreteExecution.GetDomainID()
+	domainName, err := h.DomainCache.GetDomainName(domainID)
+	if err != nil {
+		return CheckResult{}
```
Shaddoll, line 74: I think it should not return an empty result, but should return a failed result

common/persistence/dataManagerInterfaces.go
```diff
@@ -979,6 +979,7 @@ type (
 	// WorkflowEvents is used as generic workflow history events transaction container
 	WorkflowEvents struct {
 		DomainID    string
+		DomainName  string
```
Shaddoll, line 982: It seems that this change is not necessary

service/history/events/cache.go
```diff
@@ -60,6 +60,7 @@ type (
 
 	cacheImpl struct {
 		cache.Cache
+		DomainCache    cache.DomainCache
```
Shaddoll, line 63: We don't want to export this field. So it should be domainCache

service/history/execution/context.go
```diff
@@ -943,6 +943,7 @@ func (c *contextImpl) PersistStartWorkflowBatchEvents(
 		}
 	}
 
+	domainName := workflowEvents.DomainName
```
Shaddoll, line 946: we should get the domainName from domainID instead of workflowEvents

service/history/ndc/history_replicator.go
```diff
@@ -521,6 +523,7 @@ func (r *historyReplicatorImpl) applyNonStartEventsToNoneCurrentBranchWithoutCon
 		),
 		&persistence.WorkflowEvents{
 			DomainID:    task.getDomainID(),
+			DomainName:  domainName,
```
Shaddoll, line 526: revert this change, we don't want to change the data we stored in the database

service/history/replication/task_ack_manager.go
```diff
@@ -518,6 +525,7 @@ func (t *taskAckManagerImpl) generateHistoryReplicationTask(
 					task.NewRunBranchToken,
 					common.FirstEventID,
 					common.FirstEventID+1,
+					task.DomainID,
```
Shaddoll, line 528: nit: this line is not consistent with line 514

common/persistence/persistenceMetricClients.go
abhishekj720: Yes, I had a doubt regarding in the function: whether to put everything in if/else as per domainName is coming or not?

common/persistence/persistenceMetricClients.go
```diff
@@ -224,52 +224,61 @@ func NewConfigStorePersistenceMetricsClient(
...
 	default:
 		p.logger.Error("Operation failed with internal error.", tag.Error(err), tag.MetricScope(scope))
-		p.metricClient.IncCounter(scope, metrics.PersistenceFailures)
+		scopeWithDomainTag.IncCounter(metrics.PersistenceFailures)
 	}
 }
 
-func (p *persistenceMetricsClientBase) call(scope int, op func() error) error {
-	p.metricClient.IncCounter(scope, metrics.PersistenceRequests)
+func (p *persistenceMetricsClientBase) call(scope int, op func() error, tags ...string) error {
```
Shaddoll, line 262: tags can also be int type, so instead of `tags ...string`, we want `tags ...metrics.Tag`

common/persistence/persistenceMetricClients.go
```diff
@@ -224,52 +224,61 @@ func NewConfigStorePersistenceMetricsClient(
...
+	var scopeWithDomainTag metrics.Scope
+	if tags != nil {
+		domainName := tags[0]
```
Shaddoll, line 265: If we pass more than 1 tag, it won't work.

common/reconciliation/invariant/historyExists_test.go
```diff
@@ -132,10 +134,15 @@ func (s *HistoryExistsSuite) TestCheck() {
 	for _, tc := range testCases {
 		execManager := &mocks.ExecutionManager{}
 		historyManager := &mocks.HistoryV2Manager{}
+		ctrl := gomock.NewController(s.T())
 		execManager.On("GetWorkflowExecution", mock.Anything, mock.Anything).Return(tc.getExecResp, tc.getExecErr)
 		historyManager.On("ReadHistoryBranch", mock.Anything, mock.Anything).Return(tc.getHistoryResp, tc.getHistoryErr)
-		i := NewHistoryExists(persistence.NewPersistenceRetryer(execManager, historyManager, c2.CreatePersistenceRetryPolicy()))
+		
+		domainCache := cache.NewMockDomainCache(ctrl)
```
Shaddoll, line 141: great!

service/history/execution/context.go
```diff
@@ -944,6 +944,10 @@ func (c *contextImpl) PersistStartWorkflowBatchEvents(
 	}
 
 	domainID := workflowEvents.DomainID
+	domainName, err := c.shard.GetDomainCache().GetDomainName(domainID)
+	if err != nil {
+		return -1, err
```
Shaddoll, line 949: let's return 0 instead of -1 here

common/reconciliation/invariant/historyExists.go
```diff
@@ -37,16 +38,18 @@ const (
...
-		pr persistence.Retryer
+		pr          persistence.Retryer
+		domainCache cache.DomainCache
 	}
 )
 
 // NewHistoryExists returns a new history exists invariant
 func NewHistoryExists(
-	pr persistence.Retryer,
+	pr persistence.Retryer, dc cache.DomainCache,
```
Shaddoll, line 48: Some tests failed because you didn't change the reference of this function.

common/persistence/persistenceMetricClients.go
```diff
@@ -224,52 +224,61 @@ func NewConfigStorePersistenceMetricsClient(
...
 	}
 }
 
-func (p *persistenceMetricsClientBase) call(scope int, op func() error) error {
-	p.metricClient.IncCounter(scope, metrics.PersistenceRequests)
+func (p *persistenceMetricsClientBase) call(scope int, op func() error, tags ...metrics.Tag) error {
+	var scopeWithDomainTag metrics.Scope
+	if tags != nil {
+		domainName := tags[0]
+		scopeWithDomainTag = p.metricClient.Scope(scope, metrics.DomainTag(domainName.Value()))
```
Shaddoll, line 266: you still only use the first tag and ignore the others
should be `scopeWithDomainTag = p.metricClient.Scope(scope, tags...)`

service/frontend/adminHandler.go
```diff
@@ -749,6 +749,10 @@ func (adh *adminHandlerImpl) GetWorkflowExecutionRawHistoryV2(
 		return nil, adh.error(err, scope)
 	}
 	scope = scope.Tagged(metrics.DomainTag(request.GetDomain()))
+	domainName, err := adh.GetDomainCache().GetDomainName(domainID)
```
Shaddoll, line 752: request.GetDomain() is the domain name, you don't need to use domainCache to get the domain name from domainID.

service/worker/scanner/executions/types.go
```diff
@@ -49,7 +50,8 @@ type ScanType int
 
 type (
 	//InvariantFactory represents a function which returns Invariant
-	InvariantFactory func(retryer persistence.Retryer) invariant.Invariant
+	InvariantFactory  func(retryer persistence.Retryer) invariant.Invariant
+	InvariantsFactory func(retryer persistence.Retryer, dc cache.DomainCache) invariant.Invariant
```
Shaddoll, line 54: Could you remove this change in this PR? It seems this is not used in this PR

service/history/events/cache.go
```diff
@@ -60,6 +60,7 @@ type (
 
 	cacheImpl struct {
 		cache.Cache
+		domainCache    cache.DomainCache
```
Shaddoll, line 63: This field is added but not assigned

service/frontend/adminHandler_test.go
```diff
@@ -310,6 +310,7 @@ func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2_FailedOnDomain
 func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2() {
 	ctx := context.Background()
 	s.mockDomainCache.EXPECT().GetDomainID(s.domainName).Return(s.domainID, nil).AnyTimes()
+	s.mockDomainCache.EXPECT().GetDomainName(s.domainID).Return(s.domainName, nil).AnyTimes()
```
Shaddoll, line 313: this change seems unnecessary

service/frontend/adminHandler_test.go
```diff
@@ -348,6 +349,7 @@ func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2() {
 func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2_SameStartIDAndEndID() {
 	ctx := context.Background()
 	s.mockDomainCache.EXPECT().GetDomainID(s.domainName).Return(s.domainID, nil).AnyTimes()
+	s.mockDomainCache.EXPECT().GetDomainName(s.domainID).Return(s.domainName, nil).AnyTimes()
```
Shaddoll, line 352: same for this change

Shaddoll (approved):

PR #4934, opened 2022-08-10
adding reason to cancel workflow.

mindaugasbarcauskas:
Adding ability to specify cancellation reason when cancelling through cli

To have more invisibility on why workflow was cancelled.

**How did you test it?**
tested locally.

**Potential risks**

don't see a lot of risk with this change.

**Release notes**

**Documentation Changes**


Groxx:
Since proto generated files changed, this probably means that you changed IDL?  If accurate, this should probably have an idl submodule and cadence-idl go-module update as well

go.mod
```diff
@@ -35,7 +35,7 @@ require (
 	github.com/sirupsen/logrus v1.6.0
 	github.com/stretchr/testify v1.7.1
 	github.com/uber-go/tally v3.3.15+incompatible
-	github.com/uber/cadence-idl v0.0.0-20220727202351-6774502e78f4
+	github.com/uber/cadence-idl v0.0.0-20220810110900-c8cf5f3a9720
```
mindaugasbarcauskas, line 38: Hi @Groxx did you mean this change for cadence-idl?

Groxx: yeah, I suppose this PR was probably waiting on that one being merged, wasn't it.  will read today 👍

common/types/mapper/thrift/shared.go
```diff
@@ -4234,6 +4234,7 @@ func FromRequestCancelWorkflowExecutionRequest(t *types.RequestCancelWorkflowExe
 		WorkflowExecution: FromWorkflowExecution(t.WorkflowExecution),
 		Identity:          &t.Identity,
 		RequestId:         &t.RequestID,
+		Cause:             &t.Cause,
```
Groxx, line 4237: I don't know if it'll matter or not, and we're doing this for other fields too, but:
should this actually be "if empty, nil, else `&t.Cause`"?

thrift differentiates between `nil` and `""` but proto doesn't, so I wonder if this means we're passing back a bunch of empty-but-not-nil fields.  might explain why some of our logs / CLI outputs / etc are showing zero valued data...

---

but anyway, not blocking at all, just an observation / side question, and maybe something for someone to tackle in a later PR of some kind.

Groxx (approved):

davidporter-id-au (approved):

common/types/mapper/thrift/shared.go
mindaugasbarcauskas: Cause is string field not pointer so it will still have emtpy value and can't be nil, it is good question in general, but I think leaving cause as empty string is okay either way since it is just metadata

PR #4935, opened 2022-08-10
Partial response of GetReplicationMessages on history service

vytautas-karpavicius:
**What changed?**
Added code to track the size of response returned by `GetReplicationMessages` in history handler. If it reaches max grpc response size, skip responses from individual shards and return partial response.

**Why?**
If we happen to use large `history.replicatorTaskBatchSize` value, multiple shards on a history host can contribute to a large response being returned. This can either cause:
- an immediate error if grpc transport is used between frontend/history
- or can cause a stuck shard if tchannel is used between frontend/history but grpc between different clusters.

We already have similar logic on a fanout to history hosts within frontend.
Here we need a similar thing on a fanout to shards on single host.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/handler.go
```diff
@@ -1624,11 +1625,28 @@ func (h *handlerImpl) GetReplicationMessages(
...
+
 	messagesByShard := make(map[int32]*types.ReplicationMessages)
 	result.Range(func(key, value interface{}) bool {
 		shardID := key.(int32)
 		tasks := value.(*types.ReplicationMessages)
-		messagesByShard[shardID] = tasks
+
+		size := proto.FromReplicationMessages(tasks).Size()
+		if (responseSize + size) >= maxResponseSize {
+			// Log shards that did not fit for debugging purposes
```
davidporter-id-au, line 1638: Would you mind adding a metric here too? 

I assume it's safe for a partial response to come back? These'll be picked up next iteration of the replication fetch?

vytautas-karpavicius: Yes it is safe, remaining bits will be returned with next request.

vytautas-karpavicius: IMHO such metric is not strictly necessary as we already log it. And for alerting this will be detected by new ly added large replication backlog alert.

davidporter-id-au: we can set alerts based on metrics, but not logs though

davidporter-id-au (approved):

service/history/handler.go
davidporter-id-au: ok, sounds good

PR #4939, opened 2022-08-11
Add funcorder linter

shijiesheng:
**What changed?**

We need to sort the order of access methods in preparation for add codegen components. So I wrote a simple ast-based linter that automatically scan through the package and apply the changes. 

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/linter/funcorder/cmd/funcorder/main.go
+++ b/tools/linter/funcorder/cmd/funcorder/main.go
@@ -1,3 +1,25 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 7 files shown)


tools/linter/funcorder/cmd/funcorder/main.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import (
+	"github.com/uber/cadence/tools/linter/funcorder"
+	"golang.org/x/tools/go/analysis/singlechecker"
+)
+
+func main() {
```
Shaddoll, line 8: How is the tool used? Do we need to run it manually? Do we need to pass the file we want to fix?

shijiesheng: Currently it's a script that runs manually. go run main.go will output the command line.

Shaddoll (approved):

tools/linter/funcorder/cmd/funcorder/main.go
Groxx: I'm guessing this isn't hooked into the `make lint` step yet?  should that be done?

Groxx: ah.  and we should probably move this to internal/tools/go.mod

PR #4949, opened 2022-08-16
Simplify history engine task read ID logic

Shaddoll:
**What changed?**
Cherry pick https://github.com/temporalio/temporal/pull/2724
- Do not update max readable task ID if transaction to DB is rejected
- Remove redundant logic which try to update max readable task ID after shard range ID increase

**Why?**
Simplify logic

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -644,8 +644,6 @@ Create_Loop:
 			*types.ServiceBusyError:
 			// No special handling required for these errors
 			// We know write to DB fails if these errors are returned
-			// Updating MaxReadLevel doesn't matter in this case
-			s.updateMaxReadLevelLocked(transferMaxReadLevel)
 			return nil, err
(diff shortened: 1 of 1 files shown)


service/history/shard/context.go
```diff
@@ -680,10 +678,6 @@ Create_Loop:
 					)
 					s.closeShard()
 					break Create_Loop
-				} else {
-					// Shard is re-acquired successfully, and we know write to DB is guaranteed to
-					// complete, it's safe to update MaxReadLevel
-					s.updateMaxReadLevelLocked(transferMaxReadLevel)
```
Shaddoll, line 686: It's safe to remove this because if we hit this branch, `renewRangeLocked` succeeds and `transferMaxReadLevel ` is already updated inside `renewRangeLocked` function.

vytautas-karpavicius (approved):

PR #4952, opened 2022-08-16
Replication cache for sharing hydrated messages

vytautas-karpavicius:
**What changed?**
Added replication cache to store hydrated replication messages.

Two components were introduced.
- `replication.Cache` - Data structure only. This is in memory implementation of a cache. Replication tasks always come in order, thus internally heap is utilized to achieve faster `Ack` (range removals from cache). Overall complexity:
  - Get: O(1)
  - Put: O(log N)
  - Ack: O(log N)
where N is the cache size. Cache capacity is configure via new dynamic config property `history.replicatorCacheCapacity`. For now default is 10K. This may be changed later. Increasing capacity can be done dynamically, reducing - requires a restart to reclaim memory.
- `replication.TaskStore` - "manager" component for the cache. It manages separate cache for each cluster. Putting new message will put it to all caches which accept it. Ack/Get work with individual caches. It also deals with filtering out irrelevant task for domains not participating in that cluster. Rate limiting logic and retries were also moved here out from `TaskAckManager`. 

**Why?**
This is an optimization attempt to reduce replication caused load on the database.

In particular every time we fetch replication messages, they go through hydration process where raw replication tasks are enriched with history event data (at least for `ReplicationTaskTypeHistoryV2` messages, which are the majority). This results in `readhistorybranch` persistence calls.

When we have multiple polling clusters in nDC setup, each of them result is fetching and hydrating same replication messages. This can be optimized, by introducing this cache. Once one cluster hydrates the message its pointer is put to caches for other clusters, so that they do not have to repeat this process.

We can even take a step further. After a successful transaction for creating/updating workflow, we already have persisted history blobs at that point. Thus on success we could cheaply hydrate the messages and put them into cache immediately. #4953 - preparation step for that. Another PR will follow after, to complete the integration.

On a healthy system this should significantly reduce `readhistorybranch` persistence calls. If replication on one cluster starts lagging, the cache will fill up and we will stop accepting new messages for it. That means that later it will need to read them from the database as it currently done today.

**How did you test it?**
- 100% unit test coverage for new components: `replication.Cache` and `replication.TaskStore`
- Passing existing integration tests
- Deployed branch on staging2, to verify replication still works and cache metrics are emitted.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/replication/cache.go
+++ b/service/history/replication/cache.go
@@ -1,0 +1,150 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2022 Uber Technologies Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package replication
+
+import (
+	"container/heap"
+	"errors"
+	"sync"
+
+	"github.com/uber/cadence/common/dynamicconfig"
+	"github.com/uber/cadence/common/types"
+)
+
+var (
+	errCacheFull    = errors.New("cache is full")
+	errAlreadyAcked = errors.New("already acked")
+)
+
+// Cache is an in-memory implementation of a cache for storing hydrated replication messages.
+// Messages can come out of order as long as their task ID is higher than last acknowledged message.
+// Out of order is expected as different source clusters will share hydrated replication messages.
+//
+// Cache utilizes heap to keep replication messages in order. This is needed for efficient acknowledgements in O(log N).
+//
+// Cache capacity can be increased dynamically. Decrease will require a restart, as new tasks will not be accepted, but memory will not be reclaimed either.
+//
+// Cache methods are thread safe. It is expected to have writers and readers from different go routines.
+type Cache struct {
+	mu sync.RWMutex
+
+	capacity dynamicconfig.IntPropertyFn
+
+	order int64Heap
+	cache map[int64]*types.ReplicationTask
+
+	lastAck int64
+}
+
+// NewCache create a new instance of replication cache
+func NewCache(capacity dynamicconfig.IntPropertyFn) *Cache {
+	initialCapacity := capacity()
+	return &Cache{
+		capacity: capacity,
+		order:    make(int64Heap, 0, initialCapacity),
+		cache:    make(map[int64]*types.ReplicationTask, initialCapacity),
+	}
+}
+
+// Size returns current size of the cache
+func (c *Cache) Size() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return len(c.order)
+}
+
+// Put stores replication task in the cache.
+// - If cache is full, it will return errCacheFull
+// - If given task has ID lower than previously acknowledged task, it will errOutOfOrder
+func (c *Cache) Put(task *types.ReplicationTask) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Check for full cache
+	if len(c.order) >= c.capacity() {
+		return errCacheFull
+	}
+
+	taskID := task.SourceTaskID
+
+	// Reject task as it was already acknowledged
+	if c.lastAck >= taskID {
+		return errAlreadyAcked
+	}
+
+	// Do not add duplicate tasks
+	if _, exists := c.cache[taskID]; exists {
+		return nil
+	}
+
+	c.cache[taskID] = task
+	heap.Push(&c.order, taskID)
+
+	return nil
+}
+
+// Get will return a stored task having a given taskID.
+// If task is not cache, nil is returned.
+func (c *Cache) Get(taskID int64) *types.ReplicationTask {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return c.cache[taskID]
+}
+
+// Ack is used to acknowledge replication messages.
+// Meaning they will be removed from the cache.
+func (c *Cache) Ack(level int64) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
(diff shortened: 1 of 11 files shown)


davidporter-id-au:
Stupid question: Why would these tasks be read multiple times during replication? 

Excellent description by the way

service/history/replication/task_store.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+
+		scope := m.scope.Tagged(metrics.SourceClusterTag(cluster))
+
+		err := cache.Put(task)
+		switch err {
+		case errCacheFull:
+			scope.IncCounter(metrics.CacheFullCounter)
+		case errAlreadyAcked:
+			// No action, this is expected.
```
Shaddoll, line 187: Is this branch needed? How could a new history event be acked by a remote cluster?

Shaddoll (approved):

service/history/replication/cache.go
```diff
@@ -0,0 +1,150 @@
...
+// Cache is an in-memory implementation of a cache for storing hydrated replication messages.
+// Messages can come out of order as long as their task ID is higher than last acknowledged message.
+// Out of order is expected as different source clusters will share hydrated replication messages.
+//
+// Cache utilizes heap to keep replication messages in order. This is needed for efficient acknowledgements in O(log N).
+//
+// Cache capacity can be increased dynamically. Decrease will require a restart, as new tasks will not be accepted, but memory will not be reclaimed either.
+//
+// Cache methods are thread safe. It is expected to have writers and readers from different go routines.
+type Cache struct {
```
Shaddoll, line 48: Looks like it's worth making it an interface, considering we probably will introduce disk cache later.
But it can be done later, not necessarily in this PR

vytautas-karpavicius:
> Stupid question: Why would these tasks be read multiple times during replication?
> 
> Excellent description by the way

In nDC scenario, when we have >1 polling cluster, each is reading replication tasks and hydrating them. Once hydrated it is cached so that others could reuse it.

service/history/replication/task_store.go
vytautas-karpavicius: It could be removed, but I kept it to explicitly cover all possible errors returned by the cache.

Suppose we have current cluster X and two polling clusters A and B.  Both have cache size 3. Reads are in batch size 2.
1. X commit tasks 1, 2.
2. A reads until 2, hydrates them and puts to cache A and cache B. size(A) = 2, size(B) = 2
3. A acks until 2, size(A) = 0, ack(A) = 2
4. X commit task 3, 4.
5. A reads until 4, hydrates them and puts to cache A and cache B. size(A) = 2, size(B) = 3. Task4 is skipped with errCacheFull in cache B.
6. A acks until 4. size(A) = 0, ack(A) = 4
7. B reads until 2, returns from cache. size(B) = 3
8. B acks until 2, size(B) = 1, ack(B) = 2
9. B reads until 4, returns Task3 from cache, hydrates Task4 and puts to cache A and cache B. Cache A skips it with errAlreadyAcked, since ack(A) = 4 > Task3. size(B) = 2
10. B acks until 4, size(B) = 0, ack(B) = 4

To sum up. Some polling cluster may be too slow and gets it cache full. After hydrating caches for faster polling clusters will reject them, since they were already acked by them. Let me know if you need further clarification.

service/history/replication/cache.go
vytautas-karpavicius: Generally in GoLang (as opposed to some other languages) we should avoid introducing interfaces until we actually need them. It is also a good practice to define them in consumer package instead (unless they are very common interfaces, used in multiple places).

https://github.com/golang/go/wiki/CodeReviewComments#interfaces
https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b

service/history/replication/task_store.go
```diff
@@ -0,0 +1,223 @@
...
+			// Some cluster(s) may be already past this, due to different fetch rates.
+		}
+
+		scope.UpdateGauge(metrics.CacheSize, float64(cache.Size()))
```
mantas-sidlauskas, line 191: there's always caveat with gauges. Let's say we have 10 history instances and all of those will report gauge metric with different value. Usually, last one wins, but it will be not meaningful at all.

Add additional, "shard" tag to expose cache size per shard

service/history/replication/cache_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	require.NoError(t, cache.Put(task4))
+	assert.Equal(t, 1, cache.Size())
+	assert.Nil(t, cache.Get(100))
+	assert.Nil(t, cache.Get(200))
+	assert.Nil(t, cache.Get(300))
+	assert.Equal(t, task4, cache.Get(400))
+}
+
+func BenchmarkCache(b *testing.B) {
```
davidporter-id-au, line 102: nice!

service/history/replication/task_ack_manager.go
```diff
@@ -121,25 +96,8 @@ func (t *TaskAckManager) GetTasks(ctx context.Context, pollingCluster string, la
...
-		// construct replication task from DB
-		_ = t.rateLimiter.Wait(ctx)
-		var replicationTask *types.ReplicationTask
-		op := func() error {
-			var err error
-			replicationTask, err = t.hydrator.Hydrate(ctx, *task)
-			return err
-		}
-		err = t.throttleRetry.Do(ctx, op)
 		switch err.(type) {
```
davidporter-id-au, line 101: consider using `errors.Is` here

service/history/replication/task_store.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/common/quotas"
+	"github.com/uber/cadence/common/types"
+	"github.com/uber/cadence/service/history/config"
+)
+
+var errUnknownCluster = errors.New("unknown cluster")
```
davidporter-id-au, line 43: Consider exporting errors like this so we can match on them elsewhere without doing string matching

service/history/replication/task_store.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	domain, err := m.domains.GetDomainByID(info.DomainID)
+	if err != nil {
+		return nil, fmt.Errorf("resolving domain: %w", err)
+	}
+
+	// Domain does not exist in this cluster, do not replicate the task
+	if !domain.HasReplicationCluster(cluster) {
+		return nil, nil
```
davidporter-id-au, line 123: Is this an expected scenario ever?

service/history/replication/task_store.go
davidporter-id-au: Could m3 tolerate that kind of cardinality? I'm not sure it could. But your point about this being therefore not helpful if it's last-write-wins is a good one

davidporter-id-au (approved):

service/history/replication/task_store.go
vytautas-karpavicius: It is. Domain can be configured on only on a subset of clusters. No need to replicate them everywhere. Although this is probably not very common.

service/history/replication/task_store.go
vytautas-karpavicius: It can, but barely. We already have a few places that add shard tag. In particular DLQ size, and replication backlog, fetched tasks, etc. in replication/task_ack_manager.go.

service/history/replication/cache.go
```diff
@@ -0,0 +1,150 @@
...
+	"sync"
+
+	"github.com/uber/cadence/common/dynamicconfig"
+	"github.com/uber/cadence/common/types"
+)
+
+var (
+	errCacheFull    = errors.New("cache is full")
+	errAlreadyAcked = errors.New("already acked")
+)
```
Groxx, line 37: probably worth adding "replication cache: ..." to these, since the message itself is pretty generic / could conceivably be from other cache / ack code.  it'd take fairly rich error wrapping to differentiate them, currently, and we don't yet have _any_ noteworthy wrapping.

PR #4953, opened 2022-08-16
Return persisted history events blob

vytautas-karpavicius:
**What changed?**
Return whole data blob when persisting history events, instead of just size.

**Why?**
To be able to use persisted data blob when hydrating replication messages on successful transaction. #4952 For more context.

**How did you test it?**
Existing test.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/context.go
+++ b/service/history/execution/context.go
@@ -97,11 +97,11 @@ type (
 		PersistStartWorkflowBatchEvents(
 			ctx context.Context,
 			workflowEvents *persistence.WorkflowEvents,
-		) (int64, error)
+		) (persistence.DataBlob, error)
 		PersistNonStartWorkflowBatchEvents(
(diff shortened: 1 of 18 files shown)


service/history/execution/context.go
```diff
@@ -983,16 +983,19 @@ func (c *contextImpl) PersistStartWorkflowBatchEvents(
 			// TransactionID is set by shard context
 		},
 	)
-	return size, err
+	if err != nil {
+		return persistence.DataBlob{}, err
```
mindaugasbarcauskas, line 987: this should be len(resp.Data) no?

vytautas-karpavicius: resp is most likely nil when an error is returned

ZackLK (approved):

PR #4954, opened 2022-08-16
Added changes to archiver to add domain Tag

abhishekj720:
**What changed?**
Added changes to archiver to add domain Tag

**Why?**
To create metrics with Domain Tag for Domain Cost Attribution

**How did you test it?**
Tested locally and added mocks as well.

**Potential risks**
No potential risks

**Release notes**
N/A

**Documentation Changes**
N/A

### DIFF
--- a/service/worker/archiver/client.go
+++ b/service/worker/archiver/client.go
@@ -89,7 +89,7 @@ type (
 	}
 
 	client struct {
-		metricsScope                metrics.Scope
+		metricsClient               metrics.Client
 		logger                      log.Logger
 		cadenceClient               cclient.Client
 		numWorkflows                dynamicconfig.IntPropertyFn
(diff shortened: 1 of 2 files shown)


service/worker/archiver/client.go
```diff
@@ -148,12 +148,13 @@ func NewClient(
 
 // Archive starts an archival task
 func (c *client) Archive(ctx context.Context, request *ClientRequest) (*ClientResponse, error) {
+	scopeWithDomainTag := c.metricsClient.Scope(metrics.ArchiverClientScope, metrics.DomainTag(request.ArchiveRequest.DomainName))
```
Shaddoll, line 151: There is a better approach. Check this method: https://sourcegraph.uberinternal.com/github.com/uber/cadence/-/blob/common/metrics/scope.go#L122

abhishekj720: You mean like using Tagged function, right? For example- 
c.metricsScope.Tagged(metrics.DomainTag(request.ArchiveRequest.DomainName))
instead of all other changes.

Shaddoll: yes

Shaddoll (commented):
The change is done in another PR. could you close this one?

PR #4955, opened 2022-08-17
Simplify shard write operations

Shaddoll:
**What changed?**
Remove retry for shard write operations

**Why?**
To simplify logic, the code currently handles a case which can never be hit

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -175,12 +175,7 @@ var (
 	ErrShardClosed = errors.New("shard closed")
 )
 
-var (
-	errMaxAttemptsExceeded = errors.New("maximum attempts exceeded to update history")
-)
-
 const (
-	conditionalRetryCount = 5
 	// transfer/cross cluster diff/lag is in terms of taskID, which is calculated based on shard rangeID
 	// on shard movement, taskID will increase by around 1 million
 	logWarnTransferLevelDiff    = 3000000 // 3 million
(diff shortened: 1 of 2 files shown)


service/history/shard/context.go
```diff
@@ -624,67 +619,55 @@ func (s *contextImpl) CreateWorkflowExecution(
...
+				tag.Error(err),
+			)
+			s.closeShard()
 			return nil, err
-		case *persistence.ShardOwnershipLostError:
-			{
-				// RangeID might have been renewed by the same host while this update was in flight
-				// Retry the operation if we still have the shard ownership
-				if currentRangeID != s.getRangeID() {
-					continue Create_Loop
```
Shaddoll, line 653: This path can never be hit, because the rangeID is protected by a lock, and the lock is held during the write operation

service/history/shard/context.go
```diff
@@ -624,67 +619,55 @@ func (s *contextImpl) CreateWorkflowExecution(
...
-				}
-			}
-		default:
-			{
-				// We have no idea if the write failed or will eventually make it to
-				// persistence. Increment RangeID to guarantee that subsequent reads
-				// will either see that write, or know for certain that it failed.
-				// This allows the callers to reliably check the outcome by performing
-				// a read.
-				err1 := s.renewRangeLocked(false)
```
Groxx, line 671: this updates range ID.  seems to imply it was used?

Shaddoll: this is kept and the code is below

Groxx: oooh, I see.  you meant that the range isn't updated in this line:
```go
response, err := s.executionManager.CreateWorkflowExecution(ctx, request)
```
so it can't be different than when it was loaded immediately before that call / it's now immediately after.

yeah, that makes sense.  I want to dig in a bit and verify that, but that does seem very likely true.
any idea if there were other code changes that led to this no longer being reachable?  or has it always been dubious?

Shaddoll: I did the same change to UpdateWorkflowExeuction and ConflictResolveWorkflowExecution

service/history/shard/context.go
```diff
@@ -624,67 +619,55 @@ func (s *contextImpl) CreateWorkflowExecution(
...
-			return response, nil
-		case *types.WorkflowExecutionAlreadyStartedError,
-			*persistence.WorkflowExecutionAlreadyStartedError,
-			*persistence.CurrentWorkflowConditionFailedError,
-			*types.ServiceBusyError:
-			// No special handling required for these errors
-			// We know write to DB fails if these errors are returned
+	if s.isClosed() {
+		return nil, ErrShardClosed
+	}
```
Groxx, line 624: I'm guessing `s.isClosed()` is no longer possible now?

service/history/shard/context.go
```diff
@@ -624,67 +619,55 @@ func (s *contextImpl) CreateWorkflowExecution(
...
-				err1 := s.renewRangeLocked(false)
-				if err1 != nil {
-					// At this point we have no choice but to unload the shard, so that it
-					// gets a new RangeID when it's reloaded.
-					s.logger.Warn(
-						"Closing shard: CreateWorkflowExecution failed due to unknown error.",
-						tag.Error(err),
-					)
-					s.closeShard()
-					break Create_Loop
```
Groxx, line 680: oof.  this means it was returning `errMaxAttemptsExceeded` rather than the actual error :\

I'm not sure if that'll change any semantics (seems like probably no), but this certainly seems like a likely improvement at least.

service/history/shard/context.go
```diff
@@ -746,65 +729,53 @@ func (s *contextImpl) UpdateWorkflowExecution(
...
-				} else {
-					// Shard is stolen, trigger shutdown of history engine
-					s.logger.Warn(
-						"Closing shard: UpdateWorkflowExecution failed due to stolen shard.",
-						tag.Error(err),
-					)
-					s.closeShard()
-					break Update_Loop
-				}
-			}
```
Groxx, line 783: this `break Update_Loop` seems particularly bad :\

service/history/shard/context.go
Shaddoll: I need to double confirm. Probably not needed anymore now

service/history/shard/context.go
Groxx:
```
// RangeID might have been renewed by the same host while this update was in flight
// Retry the operation if we still have the shard ownership
```
hmmmmmmmmm.  this seems to imply an async update in this process, and `s.rangeID` is an atomic var...
... which is never read from.  only written to.

I agree that this only changes in renewRangeLocked and:
- before this should never be hit because it set `currentRangeID := s.getRangeID()` before entering this switch
- now this can never be hit because it doesn't loop at all, so it releases and re-acquires the lock before re-reading (no worse than before)

but this does make me wonder if the whole `contextImpl` may be flawed / using the wrong field or something.  or that atomic var is just another piece of dead code perhaps.

Groxx: ok, so rangeID was probably effectively removed in #2634

Groxx:
ok, yeah, I think I follow all this now.

Since `s.isClosed()` is an atomic check rather than something we can trust, I think my bias is to keep all of those initial checks currently.  Atomic = concurrently modified = it may save a request occasionally?  at least until it's ruled out.

Groxx (approved):
Sounds good.  Error values seem almost-definitely improved, and agreed that these are non-reachable paths.

PR #4959, opened 2022-08-18
Adding limit for amount of pending activties in mutable state.

mindaugasbarcauskas:
**What changed?**

Added limit to how many pending activties mutable state can have.

**Why?**

We need to limit mutable state size because otherwise we encounter timeouts when fetching mutable state from database.

**How did you test it?**

Tested locally + unit test.

**Potential risks**

If we have some clients that are above the threshold (default value of pending activities is 1024) their workflows would get stuck. In order to avoid this I added warn and error level logging and added feature flag for throwing the error so first we can see how many users will be affected.

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -302,6 +302,16 @@ const (
 	// Default value: 51200 (50*1024)
 	// Allowed filters: DomainName
 	HistoryCountLimitWarn
+	// PendingActivitiesCountLimitError is the limit of how many pending activities a workflow can have at a point in time
+	// KeyName: limit.pendingActivityCount.error
+	// Value type: Int
+	// Default value: 1024
+	PendingActivitiesCountLimitError
+	// PendingActivitiesCountLimitWarn is the limit of how many activities a workflow can have before a warning is logged
+	// KeyName: limit.pendingActivityCount.warn
+	// Value type: Int
+	// Default value: 512
+	PendingActivitiesCountLimitWarn
 	// DomainNameMaxLength is the length limit for domain name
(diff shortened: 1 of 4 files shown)


common/dynamicconfig/constants.go
```diff
@@ -3688,6 +3714,11 @@ var BoolKeys = map[BoolKey]DynamicBool{
 		Description:  "Lockdown defines if we want to allow failovers of domains to this cluster",
 		DefaultValue: false,
 	},
+	EnablePendingActivityValidation: DynamicBool{
+		KeyName:      "limit.pendingActivityCount.enabled",
+		Description:  "",
```
vytautas-karpavicius, line 3719: Could you add a description here as well?

mindaugasbarcauskas: added.

vytautas-karpavicius (approved):

service/history/execution/mutable_state_builder.go
```diff
@@ -2132,6 +2135,28 @@ func (e *mutableStateBuilder) AddActivityTaskScheduledEvent(
 		return nil, nil, nil, false, false, e.createCallerError(opTag)
 	}
 
+	pendingActivitiesCount := len(e.pendingActivityInfoIDs)
+
+	if pendingActivitiesCount >= e.config.PendingActivitiesCountLimitError() {
+		e.logger.Error("Pending activity count exceeds error limit",
```
Groxx, line 2141: definitely something we need metrics on, not just logs - this'll need automated alerts.

PR #4961, opened 2022-08-19
Improve log for transfer task validator

Shaddoll:
**What changed?**
- Add a tag to the logs of transfer task validator to indicate whether the lost task can be a false positive case

**Why?**
- Currently, the transfer task validator logs transfer tasks which can be lost, but when there is a timeout in persistence layer, the transfer tasks may not be persisted in the database, and these can potentially be false positive cases. We want to find these cases from the logs to improve the debugging experience.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/engine/interface.go
+++ b/service/history/engine/interface.go
@@ -26,8 +26,8 @@ import (
 	"context"
 
 	"github.com/uber/cadence/common"
-	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/types"
+	hcommon "github.com/uber/cadence/service/history/common"
 	"github.com/uber/cadence/service/history/events"
 )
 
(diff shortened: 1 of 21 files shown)


service/history/engine/interface.go
```diff
@@ -80,8 +80,8 @@ type (
 		DescribeCrossClusterQueue(ctx context.Context, clusterName string) (*types.DescribeQueueResponse, error)
 
 		NotifyNewHistoryEvent(event *events.Notification)
-		NotifyNewTransferTasks(executionInfo *persistence.WorkflowExecutionInfo, tasks []persistence.Task)
-		NotifyNewTimerTasks(executionInfo *persistence.WorkflowExecutionInfo, tasks []persistence.Task)
-		NotifyNewCrossClusterTasks(executionInfo *persistence.WorkflowExecutionInfo, tasks []persistence.Task)
+		NotifyNewTransferTasks(executionInfo *persistence.WorkflowExecutionInfo, tasks []persistence.Task, persistenceError bool)
```
shijiesheng, line 83: Should we add an entity that contains this flag instead? This will help extend in the future without modifying the interface.

Shaddoll: That brings a question of where to put the definition of the struct. It's not a simple change because it might introduce circular dependencies.

shijiesheng (approved):

PR #4962, opened 2022-08-19
Add update time to ES visibility requests and records

neil-xie:
**What changed?**
Added update time field to the workflow execution visibility requests and records, to track the workflows which don't make progress for a while.

**Why?**

**How did you test it?**
Tested with elastic search locally, was able to see the records with update time.

**Potential risks**

**Release notes**

**Documentation Changes**


shijiesheng (approved):
LGTM

PR #4969, opened 2022-08-23
Adding first scheduled time metadata field for cron workflows.

mindaugasbarcauskas:
**What changed?**

Adding first schedule time field in order to show when cron workflow was scheduled.

**Why?**
In case workflow which first scheduled the cron is old and is not retained anymore we will still be able to see when it first was scheduled.

**How did you test it?**

Tested locally + updated existing unit tests.

**Potential risks**

Low risk change.

**Release notes**

**Documentation Changes**


common/types/shared.go
```diff
@@ -6999,6 +6999,7 @@ type WorkflowExecutionStartedEventAttributes struct {
 	OriginalExecutionRunID              string                  `json:"originalExecutionRunId,omitempty"`
 	Identity                            string                  `json:"identity,omitempty"`
 	FirstExecutionRunID                 string                  `json:"firstExecutionRunId,omitempty"`
+	FirstScheduleTimeNano               *int64                  `json:"firstScheduleTimeNano,omitempty"`
```
vytautas-karpavicius, line 7002: Please use time.Time for this.

mindaugasbarcauskas: updated.

common/types/mapper/proto/helpers.go
```diff
@@ -79,6 +79,15 @@ func timeToUnixNano(t *gogo.Timestamp) *int64 {
 	return common.Int64Ptr(timestamp.UnixNano())
 }
 
+func timeStampToTime(t *gogo.Timestamp) *time.Time {
```
vytautas-karpavicius, line 82: I suggest writing helpers for both conversion in both directions. Both thrift and proto.

mindaugasbarcauskas: wrote helper for thrift + added test for said helper.

vytautas-karpavicius (approved):

PR #4977, opened 2022-08-26
Fix docker build.

ZackLK:
Fix docker image build, which has been broken since July 5th.
Tested by building normally & via docker - both locally and in buildkite.
- tcheck wasn't being found, but David and I agree that it seems unnecessary and doesn't seem to be used any more.
- Bash wasn't found in the image, but is needed to run scripts/check-gomod-version.sh, so I added it to the alpine image.
- During docker image build, the protoc binary had 644 permissions but needed to be executed so I added +x
- During docker image build, the PROTO_OUT/uber/cadence directory doesn't exist, so we skip the cp/rm commands that fail otherwise.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,6 +280,7 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 	$(call ensure_idl_submodule)
 	$Q mkdir -p $(PROTO_OUT)
 	$Q echo "protoc..."
+	$Q chmod +x $(BIN)/$(PROTOC_VERSION_BIN)
 	$Q $(foreach PROTO_DIR,$(PROTO_DIRS),$(EMULATE_X86) $(BIN)/$(PROTOC_VERSION_BIN) \
 		--plugin $(BIN)/protoc-gen-gogofast \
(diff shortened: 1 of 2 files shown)


davidporter-id-au (approved):

Groxx:
>Bash wasn't found in the image, but is needed to run scripts/check-gomod-version.sh, so I added it to the alpine image.

do we use any bash-isms?  we can very likely change that to use `sh` instead and get rid of bash.

Makefile
```diff
@@ -290,8 +291,11 @@ $(BUILD)/protoc: $(PROTO_FILES) $(BIN)/$(PROTOC_VERSION_BIN) $(BIN)/protoc-gen-g
 		--yarpc-go_out=$(PROTO_OUT) \
 		$$(find $(PROTO_DIR) -name '*.proto');\
 	)
-	$Q cp -R $(PROTO_OUT)/uber/cadence/* $(PROTO_OUT)/
-	$Q rm -r $(PROTO_OUT)/uber
+	$Q # This directory exists for local/buildkite but not for docker builds.
```
Groxx, line 294: hmm.  should this be running for docker builds actually?  we should probably not re-generate the committed files when publishing - better to publish a known-good state.

PR #4978, opened 2022-08-27
Run docker build on commits, to prevent docker build from breaking in the future

ZackLK:


davidporter-id-au (approved):
ty!

Groxx:
ooh.  yeah, great idea and thanks!

PR #4980, opened 2022-08-30
Immediate replication task hydration after successful transaction

vytautas-karpavicius:
**What changed?**
- Added `NewImmediateTaskHydrator` with corresponding inner bits for providing readily available data for replication message hydration.
- After successful transactions notify replication tasks with related data to immediately hydrate them and put them into cache.

**Why?**
After successful transaction we already have all necessary bits available for replication message hydration. We can take advantage of that - preemptively prepare replication messages and put them into cache, which will later save additional database calls (`readhistorybranch` and `getworkflowexecution`).

This should reduce overall load on database and also reduce replication latency.

**How did you test it?**
- Unit tests.
- Staging2

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/context.go
+++ b/service/history/execution/context.go
@@ -98,16 +98,16 @@ type (
 		PersistStartWorkflowBatchEvents(
 			ctx context.Context,
 			workflowEvents *persistence.WorkflowEvents,
-		) (persistence.DataBlob, error)
+		) (events.PersistedBlob, error)
 		PersistNonStartWorkflowBatchEvents(
 			ctx context.Context,
 			workflowEvents *persistence.WorkflowEvents,
-		) (persistence.DataBlob, error)
+		) (events.PersistedBlob, error)
 
 		CreateWorkflowExecution(
 			ctx context.Context,
 			newWorkflow *persistence.WorkflowSnapshot,
-			historySize int64,
+			persistedHistory events.PersistedBlob,
 			createMode persistence.CreateWorkflowMode,
 			prevRunID string,
 			prevLastWriteVersion int64,
@@ -337,7 +337,7 @@ func (c *contextImpl) LoadWorkflowExecution(
 func (c *contextImpl) CreateWorkflowExecution(
 	ctx context.Context,
 	newWorkflow *persistence.WorkflowSnapshot,
-	historySize int64,
+	persistedHistory events.PersistedBlob,
 	createMode persistence.CreateWorkflowMode,
 	prevRunID string,
 	prevLastWriteVersion int64,
@@ -359,6 +359,7 @@ func (c *contextImpl) CreateWorkflowExecution(
 		DomainName:          domainName,
 	}
 
+	historySize := int64(len(persistedHistory.Data))
 	historySize += c.GetHistorySize()
 	c.SetHistorySize(historySize)
 	createRequest.NewWorkflowSnapshot.ExecutionStats = &persistence.ExecutionStats{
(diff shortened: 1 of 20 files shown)


service/history/replication/task_hydrator.go
```diff
@@ -275,3 +283,47 @@ func (l mutableStateLoader) GetMutableState(ctx context.Context, domainID, workf
...
+}
+
+type immediateMutableState struct {
+	activities       map[int64]*persistence.ActivityInfo
+	versionHistories *persistence.VersionHistories
+}
+
+func (ms immediateMutableState) IsWorkflowExecutionRunning() bool {
+	// Immediate mutable state is always running
+	return true
```
Shaddoll, line 321: What is the impact if the workflow get closed immediately before replication?

vytautas-karpavicius: After some consideration I've update this.

This is used as early return when hydrating sync activity tasks. I think this check is somewhat redundant, since there will be no pending activity left after completion, and replication task will not be hydrated either way.

But for now let's keep it here for consistency with regular hydration from loaded mutable state. I may remove this entirely from both immediate and regular hydration later.

service/history/replication/task_hydrator.go
```diff
@@ -275,3 +283,47 @@ func (l mutableStateLoader) GetMutableState(ctx context.Context, domainID, workf
 func timeToUnixNano(t time.Time) *int64 {
 	return common.Int64Ptr(t.UnixNano())
 }
+
+type immediateHistoryProvider struct {
+	blob     *persistence.DataBlob
+	nextBlob *persistence.DataBlob
+}
+
+func (h immediateHistoryProvider) GetEventBlob(ctx context.Context, task persistence.ReplicationTaskInfo) (*types.DataBlob, error) {
```
mantas-sidlauskas, line 292: nit/optional: explicitly tell that args are not used
```suggestion
func (h immediateHistoryProvider) GetEventBlob(_ context.Context, _ persistence.ReplicationTaskInfo) (*types.DataBlob, error) {
```

service/history/replication/task_hydrator.go
```diff
@@ -60,6 +60,14 @@ type (
 	}
 )
 
+// NewImmediateTaskHydrator will enrich replication tasks with additional information that is immediately available.
+func NewImmediateTaskHydrator(isRunning bool, vh *persistence.VersionHistories, activities map[int64]*persistence.ActivityInfo, blob, nextBlob *persistence.DataBlob) TaskHydrator {
+	return TaskHydrator{
+		history:    immediateHistoryProvider{blob, nextBlob},
```
mantas-sidlauskas, line 66: nit, use field names for initialisation

service/history/historyEngine.go
```diff
@@ -2946,6 +2947,59 @@ func (e *historyEngineImpl) NotifyNewCrossClusterTasks(
...
+	default:
+		return nil, errors.New("unknown replication task")
+	}
+
+	hydrator := replication.NewImmediateTaskHydrator(
+		exec.IsRunning(),
+		versionHistories,
+		activities,
+		history.Find(info.BranchToken, info.FirstEventID),
+		history.Find(info.NewRunBranchToken, 1),
```
mantas-sidlauskas, line 2997: will this always be 1?

mantas-sidlauskas (approved):
looks good to me, couple of nit level suggestions

service/history/execution/context.go
```diff
@@ -368,12 +369,12 @@ func (c *contextImpl) CreateWorkflowExecution(
 	resp, err := c.createWorkflowExecutionWithRetry(ctx, createRequest)
 	if err != nil {
 		if c.isPersistenceTimeoutError(err) {
-			c.notifyTasksFromWorkflowSnapshot(newWorkflow, true)
+			c.notifyTasksFromWorkflowSnapshot(newWorkflow, events.PersistedBlobs{persistedHistory}, true)
```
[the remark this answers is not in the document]
vytautas-karpavicius: Notification are not call when there is an error but it is not isPersistenceTimeoutError.
This can probably be simplified, but I will leave as is for now.

service/history/historyEngine.go
vytautas-karpavicius: Changed magic number 1 to a constant `common.FirstEventID` used elsewhere as well.
This is always 1 at least in the current implementation: https://github.com/uber/cadence/blob/master/service/history/replication/task_hydrator.go#L227

service/history/execution/context.go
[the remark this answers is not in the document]
mantas-sidlauskas: yes, missed that return. thanks

PR #4987, opened 2022-09-01
Added logs for domainName empty situation

abhishekj720:
**What changed?**
-- added logs before returning to the base case. 

**Why?**
-- Added logs to understand where the program failed, and reason of failing.

**How did you test it?**
-- Tested Locally

**Potential risks**
No potential risks

**Release notes**
N/A

**Documentation Changes**
N/A


service/frontend/adminHandler.go
```diff
@@ -447,6 +447,7 @@ func (adh *adminHandlerImpl) deleteWorkflowFromHistory(
 		}
 		domainName, err := adh.GetDomainCache().GetDomainName(mutableState.ExecutionInfo.DomainID)
 		if err != nil {
+			logger.Error("Unexpected: Cannot fetch domain name", tag.Error(err))
 			return false
```
Shaddoll, line 451: Can we do `continue` here?

service/worker/scanner/history/scavenger.go
```diff
@@ -262,6 +262,8 @@ func (s *Scavenger) startTaskProcessor(
 					}
 					domainName, err := s.domainCache.GetDomainName(task.domainID)
 					if err != nil {
+						s.logger.Error("Not able to fetch domain name",
+							getTaskLoggingTags(err, task)...)
 						return
```
Shaddoll, line 267: we need to send the error to the channel, and also do `continue` instead of `return`.
See how we handle other errors in this func.

Shaddoll (approved):

PR #4991, opened 2022-09-06
Decouple StateBuilder from TaskGenerator

vytautas-karpavicius:
**What changed?**
Decoupled `StateBuilder` from `TaskGenerator` by moving task generation to `Replicate<...>` functions on mutable state. Task generation is needed on both active and passive side, so there is no need to call them separately. There were only a few places where active/passive side differs slightly. For them we can simply pass additional flag modifying that behavior. In particular, more attention is needed for:
- `ReplicateWorkflowExecutionStartedEvent` will `GenerateDelayedDecisionTasks` only for passive side. Active side does that separately during `AddFirstDecisionTaskScheduled`.  
-  After rebuilding `GenerateActivityTimerTasks`/`GenerateUserTimerTasks` was called. This can be skipped here, and instead call when closing a transaction. Previously it was not done due to `transactionPolicy == TransactionPolicyPassive`, now this is removed. And they will be generated for both active/passive cases.

**Why?**
- Reduced tech debt as indicates by `// TODO merge active & passive task generation`.
- This also decouples StateBuilder which makes it possible to further refactor and simplify.
- Task generation is more centralized.

**How did you test it?**
Existing tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/mutable_state_builder.go
```diff
@@ -2174,12 +2166,6 @@ func (e *mutableStateBuilder) AddActivityTaskScheduledEvent(
 		activityStartedScope.IncCounter(metrics.CadenceRequests)
 		return event, ai, nil, true, true, nil
 	}
-	// TODO merge active & passive task generation
```
Shaddoll, line 2177: This will change the behavior of activity local dispatch.

Before, we try dispatch activity task and if it succeeds, we skip the transfer task creation.
After, we always create the transfer task.

service/history/execution/mutable_state_builder.go
```diff
@@ -4690,8 +4628,7 @@ func (e *mutableStateBuilder) closeTransactionHandleActivityUserTimerTasks(
 	transactionPolicy TransactionPolicy,
```
Shaddoll, line 4628: It seems that we can remove this parameter

vytautas-karpavicius: Thanks, updated.

service/history/execution/mutable_state_builder.go
vytautas-karpavicius: Yes, updated. Good catch.

Shaddoll (approved):

PR #4993, opened 2022-09-08
Convert client peer resolving errors to service transient errors

Shaddoll:
**What changed?**
Convert client peer resolving errors to service transient errors

**Why?**
When processing signal and cancellation transfer tasks, we check if some errors are transient errors to determine whether we should retry or record a failure event in workflow history.
Peer resolving errors should be treated as transient errors and during deployment and restart we can see these errors.

**How did you test it?**
existing test

**Potential risks**
If we have a bad config file or network issue which causes peer resolving to always fail, those transfer tasks will be stuck. 
But I think that's the behavior we want.

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -263,6 +263,14 @@ func CheckDecisionResultLimit(
 	return nil
 }
 
+// ToServiceTransientError converts an error to ServiceTransientError
+func ToServiceTransientError(err error) error {
+	if err == nil || IsServiceTransientError(err) {
+		return err
+	}
+	return yarpcerrors.Newf(yarpcerrors.CodeUnavailable, err.Error())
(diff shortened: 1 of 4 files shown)


common/util.go
```diff
@@ -263,6 +263,14 @@ func CheckDecisionResultLimit(
 	return nil
 }
 
+// ToServiceTransientError converts an error to ServiceTransientError
+func ToServiceTransientError(err error) error {
```
davidporter-id-au, line 267: Do you mind adding a unit test roundtripping this through `IsServiceTransientError`?

davidporter-id-au (approved):

PR #4994, opened 2022-09-08
Emitting version metrics also adding a githook that runs a script the automatically updates the version unless the user manually skips

allenchen2244:
Adding version into the build_info metric gauge. 

Played around with metrics on grafana should be able to add a graph in pretty easily once this is done.

Also added a script that updates a version number with a timestamp if there's no manual update the major or patch version.

Added a githook that ensures the user is aware that they're updating the version or skipping it entirely if it's something really small


common/metrics/runtime.go
```diff
@@ -52,6 +52,8 @@ var (
 
 	// goVersion is the current runtime version.
 	goVersion = runtime.Version()
+	// cadenceVersion is the current version of cadence
+	cadenceVersion = "0.25.x"
```
davidporter-id-au, line 56: Can we add a makefile step or some other test step to ensure this doesn't get out of sync with the tagged release? @Groxx  does this seem insane?

allenchen2244: I think Stephen/Ender was saying that the tag version being the last and all the way to do it is to make sure we update the release steps to update this version with every branch since the version itself isn't accessible.

davidporter-id-au: I mean something like `make test` also doing a grep for this value and ensuring it's not out of sync with the tag

davidporter-id-au:
Please update the readme to include: 

1. The proposal you're putting here - tl;dr that we're adding a timestamp-based patch-version which is exposed at runtime which is hooked in to commits and bumped automatically (perhaps surprising) 
2. A callout that this requires `git config core.hooksPath .githooks` to manually be run

davidporter-id-au (approved):

common/metrics/runtime.go
Groxx: late to the game, but:
unfortunately I don't think there's a way to automate this, short of automating releases entirely... which brings other major complications too.  though we might do that eventually.

for environments that use go modules (i.e. not our internal monorepo, at least currently), we could pull this information from the module's build-time metadata, which is probably the best option.  that's always up to date.

for non-module environments... dunno.

PR #4996, opened 2022-09-14
Drop stuck workflow timeout timer task

Shaddoll:
**What changed?**
- Drop workflow timeout timer task if the error is conditional error or workflow start event missing error
- Drop stuck tasks only after a few retries
- Update related metrics

**Why?**
Lots of resurrected workflow history.

**How did you test it?**
This feature is controlled by a feature-flag and we can enable it domain by domain. Some of the logic is already tested in production. The new logic can be tested in a staging environment.

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/task.go
```diff
@@ -277,13 +277,20 @@ func (t *taskImpl) HandleErr(
...
-		return nil
+	if err == execution.ErrMissingWorkflowStartEvent {
+		if t.shouldDropTaskOnError() {
+			t.scope.IncCounter(metrics.TaskMissingEventCounterPerDomain)
+			t.logger.Error("Drop task due to corrupted workflow history", tag.Error(err), tag.TaskType(t.GetTaskType()), tag.LifeCycleProcessingFailed)
+			return nil
+		}
+	}
+
+	if execution.IsConflictError(err) {
```
emrahs, line 288: I don't think it's safe to drop a task based on a conflict error, which can happen for non-corruption reasons.

service/history/task/task.go
```diff
@@ -447,6 +454,21 @@ func (t *taskImpl) shouldResubmitOnNack() bool {
 		(t.queueType == QueueTypeActiveTransfer || t.queueType == QueueTypeActiveTimer)
 }
 
+func (t *taskImpl) shouldDropTaskOnError() bool {
+	if t.shard.GetConfig().EnableDropStuckTaskByDomainID(t.Info.GetDomainID()) && // use domainID here to avoid accessing domainCache
+		t.GetAttempt() > t.criticalRetryCount()/2 {
```
emrahs, line 459: I like that you are adding the attempt count here to ensure we don't drop transfer tasks on the first occurrence of an ErrMissingWorkflowStartEvent, since it can happen due to some transient DB issue. But, why divide the critical retry count by 2 instead of using it as is?

service/history/task/task.go
```diff
@@ -447,6 +454,21 @@ func (t *taskImpl) shouldResubmitOnNack() bool {
...
+		if transferTask, ok := t.Info.(*persistence.TransferTaskInfo); ok &&
+			transferTask.TaskType == persistence.TransferTaskTypeCloseExecution {
+			return true
+		}
+		if timerTask, ok := t.Info.(*persistence.TimerTaskInfo); ok &&
+			timerTask.TaskType == persistence.TaskTypeWorkflowTimeout {
```
emrahs, line 465: Can you elaborate more on why you think this task needs/can be dropped?

Note that workflow close/timeout isn't a low-priority event. If we fail to close a workflow, users won't be able to reuse the workflow ID and have other issues. Therefore, we should treat workflow close/timeout events as high priority as start workflow requests.

service/history/task/task.go
```diff
@@ -447,6 +454,21 @@ func (t *taskImpl) shouldResubmitOnNack() bool {
 		(t.queueType == QueueTypeActiveTransfer || t.queueType == QueueTypeActiveTimer)
 }
 
+func (t *taskImpl) shouldDropTaskOnError() bool {
```
emrahs, line 457: This helper is a little bit dangerous without the err being validated inside. I'd pass the error as an arg and check it, so the function can't be misused in the future.

service/history/task/task.go
Shaddoll: It doesn't have to be half of that value. It just needs to be smaller than that value to make sure we don't generate logs for the stuck tasks.

service/history/task/task.go
Shaddoll: Most of the stuck tasks are close/timeout tasks. 

> If we fail to close a workflow, users won't be able to reuse the workflow ID and have other issues.

We check the error to make sure the workflow is a resurrected workflow.

service/history/task/task.go
Shaddoll: even if we have retried multiple times?

PR #5001, opened 2022-09-23
Adds fix for domain ack level issue

davidporter-id-au:
It seems possible for ack levels to be higher than the last message. During this scenario domain updates aren't propagated (if the ack-levels are already in memory) because the replication won't fetch them. This is a guardrail against this. 

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Shaddoll (commented):
How does sql implementation handle this?
I'm wondering there might be a general big issue breaking assumptions we make on this component.

davidporter-id-au:
No idea, let me check

davidporter-id-au:
It's making the same assumption. You want me to add the check for both?

Shaddoll (approved):

PR #5009, opened 2022-10-03
Fix ndc flush buffered events

Shaddoll:
**What changed?**
Fix ndc flush buffered events
Fix https://github.com/uber/cadence/issues/5006

**Why?**
workflow can get stuck after failover because of no scheduled decision

**How did you test it?**
local test

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/workflow.go
```diff
@@ -232,22 +233,13 @@ func (r *workflowImpl) failDecision(
...
-		"",
-		0,
-	); err != nil {
+	if err := FailDecision(r.mutableState, decision, types.DecisionTaskFailedCauseFailoverCloseDecision); err != nil {
 		return err
 	}
-
-	return r.mutableState.FlushBufferedEvents()
+	if scheduleNewDecision {
+		return ScheduleDecision(r.mutableState)
```
davidporter-id-au, line 240: What will happen if this crashes partway through? It'll still get lost I assume?

davidporter-id-au (approved):

service/history/execution/workflow.go
Shaddoll: what do you mean by crash?

Shaddoll: these are in-memory operations, so it doesn't matter.

PR #5015, opened 2022-10-31
Feature/min initial failover version

davidporter-id-au:
## Problem 

The Initial Failover version is a somewhat overloaded monotonic counter, which uses the first couple of bits to identify the origin cluster for where an action was performed. 

However, when joining separate clusters, its necessary for these to be unique, migration between clusters isn't otherwise possible. 

Unfortunately, if you're joining two clusters with fairly high traffic together, it's necessary to move one of them off to a unique initial failover version. 

## Solution 
This change allows a dynamic configuration guarded change to issue a second initialFailoverVersion from a cluster, such that traffic can gradually be moved off the previous value. 

## Testing Plan 

Testing will be done by:
[✓] Verifying locally
[ ] Migrating in test environments
[ ] Migrating against non-production traffic.


common/cluster/metadata.go
```diff
@@ -151,3 +171,45 @@ func (m Metadata) ClusterNameForFailoverVersion(failoverVersion int64) string {
...
+			cluster,
+			m.allClusters,
+		))
+	}
+
+	// if using the minFailover Version during a cluster config, then return this from config
+	// (assuming it's safe to do so). This is not the normal state of things and intended only
+	// for when migrating versions.
+	usingMinFailoverVersion := m.useMinFailoverVersionOverride(domainName)
+	if usingMinFailoverVersion && info.MinInitialFailoverVersion != nil && *info.MinInitialFailoverVersion > info.InitialFailoverVersion {
```
Shaddoll, line 191: tbh, I don't think `*info.MinInitialFailoverVersion > info.InitialFailoverVersion` is necessary

common/cluster/metadata.go
```diff
@@ -138,7 +159,6 @@ func (m Metadata) ClusterNameForFailoverVersion(failoverVersion int64) string {
 	if failoverVersion == common.EmptyVersion {
 		return m.currentClusterName
 	}
-
 	initialFailoverVersion := failoverVersion % m.failoverVersionIncrement
 	clusterName, ok := m.versionToClusterName[initialFailoverVersion]
```
Shaddoll, line 163: This doesn't look correct if the domain is currently using minInitialFailoverVersion

common/config/cluster.go
```diff
@@ -58,6 +58,12 @@ type (
 		Enabled bool `yaml:"enabled"`
 		// InitialFailoverVersion is the identifier of each cluster. 0 <= the value < failoverVersionIncrement
 		InitialFailoverVersion int64 `yaml:"initialFailoverVersion"`
+		// MinInitialFailover is a new failover version for an initialFailoverVersion migration
+		// for when it's necessary to migrate between two values.
+		//
+		// this is a pointer to imply optionality, it's an optional field and its lack
+		// is indicated by a nil pointer. Zero is a valid field
+		MinInitialFailoverVersion *int64 `yaml:"minInitialFailoverVersion"`
```
Shaddoll, line 66: IMHO, MinInitialFailoverVersion is confusing, how about naming it NewInitialFailoverVersion

common/cluster/metadata.go
```diff
@@ -54,6 +63,9 @@ func NewMetadata(
 	primaryClusterName string,
 	currentClusterName string,
 	clusterGroup map[string]config.ClusterInformation,
+	useMinFailoverVersionOverrideConfig dynamicconfig.BoolPropertyFnWithDomainFilter,
+	metricsClient metrics.Client,
+	logger log.Logger,
 ) Metadata {
 	versionToClusterName := make(map[int64]string)
 	for clusterName, info := range clusterGroup {
```
Shaddoll, line 71: I think we also need to add MinInitalFailoverVersion to the map

common/config/cluster.go
davidporter-id-au: Sure, reasonable, will do

common/cluster/metadata.go
davidporter-id-au: Good catch 🤦

config/dynamicconfig/development.yaml
```diff
@@ -10,3 +10,9 @@ history.EnableConsistentQueryByDomain:
 history.enableCrossClusterOperations:
 - value: true
   constraints: {}
+history.enableCrossClusterOperations:
```
Shaddoll, line 13: history.enableCrossClusterOperations is duplicated

Shaddoll (approved):

PR #5025, opened 2022-11-10
Fix workflow cancellation

Shaddoll:
**What changed?**
- Don't return error if the workflow is already cancelled by the request with the same request ID

**Why?**
- Make the behavior idempotent 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


mindaugasbarcauskas (approved):
Looks good to me.

vytautas-karpavicius (approved):

PR #5027, opened 2022-11-17
Add cli tls support

charlese-instaclustr:
Added the option to use TLS via the Cadence CLI, by use of a `tls_cert_path` flag which accepts a file path to a PEM formatted certificate file.

**Why?**

A user may want to make a CLI call to the grpc port of a TLS enabled Cadence cluster. In this case, the CLI needs to be able to construct the correct TLS outbound configuration, which requires enabling the user to pass the relevant cert file.

**How did you test it?**

- Built the Cadence CLI locally
- Provisioned one TLS-enabled and one non-TLS-enabled Cadence cluster
- Tested new feature by making `cadence admin cluster describe` and `cadence domain register` CLI calls to the grpc port of the TLS-enabled cluster, while passing the relevant cert file via the new `tls_cert_path` flag.
- Regression tested by calling the same commands against the grpc port of the non-TLS-enabled cluster
- Regression tested with the same commands against the tchannel ports of both clusters

**Potential risks**

This change does not affect the behaviour of existing transport options. So, in the worst case (e.g. the new feature does not allow connection to a TLS-enabled cluster in some case), it should not break any existing deployments/configurations.

**Release notes**

No.

**Documentation Changes**

No.


tools/cli/app.go
```diff
@@ -79,6 +79,12 @@ func NewCliApp() *cli.App {
 			Usage:  "optional argument for transport protocol format, either 'grpc' or 'tchannel'. Defaults to tchannel if not provided",
 			EnvVar: "CADENCE_CLI_TRANSPORT_PROTOCOL",
 		},
+		cli.StringFlag{
+			Name:   FlagTLSCertPathWithAlias,
+			Value:  "",
```
shijiesheng, line 84: nit: delete this line to keep the style consistent

longquanzheng:
Quick question, Have you tested this working?

charlese-instaclustr:
> Quick question, Have you tested this working?

@longquanzheng Hi, yes, please see the testing procedure in my original comment:

- Built the Cadence CLI locally
-Provisioned one TLS-enabled and one non-TLS-enabled Cadence cluster
-Tested new feature by making cadence admin cluster describe and cadence domain register CLI calls to the grpc port of the TLS-enabled cluster, while passing the relevant cert file via the new tls_cert_path flag.
-Regression tested by calling the same commands against the grpc port of the non-TLS-enabled cluster
-Regression tested with the same commands against the tchannel ports of both clusters

Please let me know if you think any more should be conducted, thanks.

tools/cli/app.go
charlese-instaclustr: Done, thanks

shijiesheng (approved):

PR #5031, opened 2022-11-19
Add FirstExecutionRunID to mutable state

Shaddoll:
**What changed?**
Added FirstExecutionRunID to mutable state which gets persisted in DB. Now we don't rely on workflow execution started event for this information.

This partially picks https://github.com/temporalio/temporal/pull/365

**Why?**
See https://github.com/temporalio/temporal/pull/365

**How did you test it?**
unit test, integration test and local test on laptop

**Potential risks**
There might be backward compatibility issue, and we must make sure the old mutable state can still be read from database.

**Release notes**

**Documentation Changes**


common/persistence/serialization/persistence_mapper.go
```diff
@@ -152,6 +153,7 @@ func FromInternalWorkflowExecutionInfo(executionInfo *persistence.InternalWorkfl
 		CompletionEventEncoding:            string(common.EncodingTypeEmpty),
 		VersionHistoriesEncoding:           string(common.EncodingTypeEmpty),
 		InitiatedID:                        common.EmptyEventID,
+		FirstExecutionRunID:                MustParseUUID(executionInfo.FirstExecutionRunID),
```
davidporter-id-au, line 156: Is there any possibility of this being an invalid UUID? I'm always a little scared of using MustParse for places like this

davidporter-id-au: Looking below at the builder - from a naïve perspective this looks maybe it could be an invalid uuid, or at least a nil one

davidporter-id-au (approved):

common/persistence/serialization/persistence_mapper.go
Shaddoll: MustParseUUID works with empty string

PR #5032, opened 2022-11-24
[WIP]Update parent close policy to terminate/cancel child workflows even after continue as new

Shaddoll:
**What changed?**
this picks https://github.com/temporalio/temporal/pull/696
and partially picks https://github.com/temporalio/temporal/pull/365

**Why?**
see the 2 cherry picks

**How did you test it?**
TBD

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/transfer_active_task_executor.go
```diff
@@ -2050,9 +2052,11 @@ func applyParentClosePolicy(
 				Domain: childDomainName,
 				WorkflowExecution: &types.WorkflowExecution{
 					WorkflowID: childStartedWorkflowID,
-					RunID:      childStartedRunID,
```
mindaugasbarcauskas, line 2053: I am wondering if not setting the runId could cause some issues since from client side anyone can access the WorkflowExecution, what do you think? 
Was it removed from here because now we will be using FirstExecutionRunID everywhere?

Shaddoll: we use FirstExecutionRunID only here, and it doesn't make sense to set run id if FirstExecutionRunID is set

mindaugasbarcauskas (approved):
Looks good.

PR #5036, opened 2022-11-28
Fix SignalWithStartWorkflow 

Shaddoll:
**What changed?**
pick https://github.com/temporalio/temporal/pull/719
and fix duplicate SignalWithStartWorkflow requests

**Why?**
see https://github.com/temporalio/temporal/pull/719

**How did you test it?**
Tested locally on laptop

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -24,6 +24,7 @@ package history
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"sync"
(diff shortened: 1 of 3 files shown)


service/history/handler.go
```diff
@@ -1144,12 +1144,36 @@ func (h *handlerImpl) SignalWithStartWorkflowExecution(
...
+		// Two simultaneous SignalWithStart requests might try to start a workflow at the same time.
+		// This can result in one of the requests failing with one of two possible errors:
+		//    1) If it is a brand new WF ID, one of the requests can fail with WorkflowExecutionAlreadyStartedError
+		//       (createMode is persistence.CreateWorkflowModeBrandNew)
+		//    2) If it an already existing WF ID, one of the requests can fail with a CurrentWorkflowConditionFailedError
+		//       (createMode is persisetence.CreateWorkflowModeWorkflowIDReuse)
+		// If either error occurs, just go ahead and retry. It should succeed on the subsequent attempt.
+		// For simplicity, we keep trying unless the context finishes or we get an error that is not one of the
+		// two mentioned above.
+		_, isExecutionAlreadyStartedErr := err2.(*persistence.WorkflowExecutionAlreadyStartedError)
```
davidporter-id-au, line 1161: nit: prefer `errors.Is` I think, that way it'll keep working even if we end up wrappign the error

service/history/handler.go
```diff
@@ -1144,12 +1144,36 @@ func (h *handlerImpl) SignalWithStartWorkflowExecution(
 		return nil, h.error(err1, scope, domainID, workflowID)
 	}
 
-	resp, err2 := engine.SignalWithStartWorkflowExecution(ctx, wrappedRequest)
-	if err2 != nil {
-		return nil, h.error(err2, scope, domainID, workflowID)
-	}
+	for {
```
davidporter-id-au, line 1147: this looks like it could retry more than twice, would it be possible to add a counter and just check we don't end up retrying more than N times?

Shaddoll: Yeah, it could at most do twice. But it doesn't hurt to retry until context timeout.

service/history/handler.go
Shaddoll: will do

davidporter-id-au (approved):

PR #5038, opened 2022-12-05
Improve Cassandra errors for schema check

mantas-sidlauskas:
**What changed?**
Return original error from `cql` client when doing schema check.

**Why?**
Make error self explanatory: from this
```
cassandra schema version compatibility check failed: unable to read schema version keyspace/database: cadence error: failed to get current schema version from cassandra
```
to this

```
cassandra schema version compatibility check failed: reading schema version keyspace/database: "cadence", error: reading schema: unconfigured table schema_version
```

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cassandra/handler.go
+++ b/tools/cassandra/handler.go
@@ -75,7 +75,7 @@ func verifyCompatibleVersion(
 	// However, this file will be refactor to support NoSQL soon. After the refactoring, cycle dependency issue
 	// should be gone and we can use constant at that time
 	if ds.NoSQL.PluginName != "cassandra" {
-		return fmt.Errorf("unknown NoSQL plugin name: %v", ds.NoSQL.PluginName)
+		return fmt.Errorf("unknown NoSQL plugin name: %q", ds.NoSQL.PluginName)
 	}
 
 	return CheckCompatibleVersion(*ds.NoSQL, expectedCassandraVersion)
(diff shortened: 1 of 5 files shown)


tools/cassandra/handler.go
```diff
@@ -75,7 +75,7 @@ func verifyCompatibleVersion(
 	// However, this file will be refactor to support NoSQL soon. After the refactoring, cycle dependency issue
 	// should be gone and we can use constant at that time
 	if ds.NoSQL.PluginName != "cassandra" {
-		return fmt.Errorf("unknown NoSQL plugin name: %v", ds.NoSQL.PluginName)
+		return fmt.Errorf("unknown NoSQL plugin name: %q", ds.NoSQL.PluginName)
```
davidporter-id-au, line 78: I guess this is at the handler layer, but wouldn't %w be preferable due to the unwrapping properties?

davidporter-id-au (approved):

tools/cassandra/handler.go
mantas-sidlauskas: We don't have error to wrap here, I changed `%v` to `%q` to have  errors like `unknown NoSQL plugin name: ""` instead of `unknown NoSQL plugin name:`. 
Or do I miss something?

davidporter-id-au: both %w, %v and %q should all print out the error string as far as I know, but %w will keep the error as being available for higher layers to match off. But, as you point out, this is the handler layer, so this is probably not an issue.

PR #5041, opened 2022-12-08
Wf version metrics

allenchen2244:
Cleaned up ESAnalyzer and added workflow version metrics

Clean up dead code and add workflow version metrics

Tested fully end to end in staging and was able to see the correct metrics in m3 against elasticsearch results. Also wrote a unit test to test against a raw es search query result.

Very little risk since all the deleted code was not running and the new code is just a read operation to elasticsearch wrapped behind a flipr.


Shaddoll (commented):
Do you mind splitting it to 2 separate PRs?

allenchen2244:
> Do you mind splitting it to 2 separate PRs?

Not sure how to do a split. I'm kind of reusing old stuff

davidporter-id-au (approved):

PR #5045, opened 2022-12-10
Fix error message

ZackLK:
Fixed error message to correctly include operation and error


Groxx (approved):
Whoops.

Thanks for the fix!

PR #5051, opened 2022-12-15
Exposes schema task

davidporter-id-au:
Exposes UpdateTask such that it may be used for testing and to determine Schema changes rather than just writing them to stdout

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Shaddoll (approved):

davidporter-id-au (commented):
bumping to force a rebuild because it's stuck

PR #5054, opened 2022-12-20
Do not return not exists error in history pagination function

Shaddoll:
**What changed?**
Do not return not exists error in history pagination function

**Why?**
For a history paginator, if there is no history events, we should not return entity not exists error but stop the iteration.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/historyManager.go
+++ b/common/persistence/historyManager.go
@@ -422,6 +422,7 @@ func (m *historyV2ManagerImpl) readRawHistoryBranch(
 	if err != nil {
 		return nil, nil, 0, nil, err
 	}
+	// TODO: consider if it's possible to remove this branch
 	if len(resp.History) == 0 && len(request.NextPageToken) == 0 {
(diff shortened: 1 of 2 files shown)


davidporter-id-au:
Can you provide a bit more detail as to this scenario? Under conditions is returning no history a valid state?

allenchen2244 (approved):
LGTM

Shaddoll:
> Can you provide a bit more detail as to this scenario? Under conditions is returning no history a valid state?

@davidporter-id-au We have some helper functions iterating workflow histories. The iterator provides `HasNext()` and `Next()` functions to iterate histories. But if the workflow history doesn't exist, the first iteration of `HasNext()` doesn't return `false`, and when we call `Next()`, we get an error. I think for an iterator, this behavior is not what we want, and it actually breaks the timer resurrection checker, because when we get an EntityNotExist error, the timer tasks stops retrying

davidporter-id-au (approved):

davidporter-id-au:
nice, thank you for the explanation

PR #5061, opened 2023-01-02
Hot shard detection

davidporter-id-au:
### Summary

Samples tasks being processed and issues a log and metric if a repeated entry is identified, with the intention of spotting hot-shards.

### Background

Based on a lot of operational experiences, it has been clear for a while that we need a cheap way to identify workflows which are running continue-as-new a lot or have very high task process thrashing on a single shard, such activities can take out a database's availbility. It's also operationally not always quite easy to identify such workflows without more subtle-clues such as looking at timeout correlations and guesswork. 

Initially I was considering a more sophisticated detector - something which would be able to identify if a workflow has ever been seen with some approximate count and considered count-min-sketch implementations for this. However, after fiddling with them for a while I decided against them in favour of a far simpler approach of just sampling. The reason for this is that count-min-sketch and similar probabilistic approximation algorithms overcount, making the risks of false-positives everpresent unless you have a moderate amount of memory allocated for such detection. 

Sampling on the other hand, is, if tuned correctly - by definition - undercounts, making it easier to control, and - I hope - offers a similarly modest memory footprint. 

### Testing 

So far it's only tuned to my laptop, and given it's all configured with absolute values, this is near-certainly wrong. I'll have to tune more once deployed.

**Potential risks**
Lots, a panic in this could do quite a bit of damage, it'll have to be tested well.

**Release notes**
Ready for release, no changes required.

**Documentation Changes**

### DIFF
--- a/common/hot-shard/hot-shard.go
+++ b/common/hot-shard/hot-shard.go
@@ -1,14 +1,38 @@
-package hot_shard
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


common/hot-shard/hot-shard.go
```diff
@@ -0,0 +1,212 @@
...
+	existingCount, ok := d.windowCount[hashedIdentifier]
+	d.mu.RUnlock()
+
+	// if the value has been seen before, and we know it's over limit don't sample this, reply immediately that
+	// we've seen this value before
+	if ok && existingCount > d.limit {
+		d.log.Warn("hot-shard detected", tag.WorkflowID(workflowID), tag.Dynamic("entries", dim))
+		d.metrics.IncCounter(metrics.HotShardRateLimitTriggered, 1)
+		// keep adding to counter
+		d.sampleAndCheck(now, hashedIdentifier, workflowID, dim)
```
davidporter-id-au, line 93: need to consider doing this async

common/hot-shard/hot-shard.go
```diff
@@ -0,0 +1,212 @@
...
+	metrics               metrics.Client
+	sampleRate            float64
+	totalHashmapSizeLimit int
+}
+
+func (d *detector) Check(now time.Time, workflowID string, dim ...string) bool {
+	// first, hash the inputs, and see
+	// then if we've seen this value before
+	d.mu.RLock()
+	hashedIdentifier := d.hashInputs(workflowID, dim)
```
davidporter-id-au, line 83: todo: redo this with atomics

davidporter-id-au:
Closing as this is being handled elsewhere by @jhtuber 's work

PR #5070, opened 2023-01-31
Add domain tag to unregistered field error

neil-xie:
**What changed?**
Add domain tag to unregister field error for ElasticSearch, so we can know where it come from and fix

**Why?**

**How did you test it?**
Existing tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/indexer/processor.go
+++ b/service/worker/indexer/processor.go
@@ -276,7 +276,7 @@ func (p *indexProcessor) dumpFieldsToMap(fields map[string]*indexer.Field, domai
 				attr[k] = p.decodeSearchAttrBinary(v.GetBinaryData(), k)
 			}
 		default:
-			// must be bug in code and bad deployment, check data sent from producer
+			// there must be bug in code and bad deployment, check data sent from producer
 			p.logger.Fatal("Unknown field type")
 		}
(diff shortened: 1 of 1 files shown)


davidporter-id-au (approved):

service/worker/indexer/processor.go
```diff
@@ -232,7 +232,7 @@ func (p *indexProcessor) addMessageToES(indexMsg *indexer.Message, kafkaMsg mess
 }
 
 func (p *indexProcessor) generateESDoc(msg *indexer.Message, keyToKafkaMsg string) map[string]interface{} {
-	doc := p.dumpFieldsToMap(msg.Fields)
+	doc := p.dumpFieldsToMap(msg.Fields, msg.GetDomainID())
```
davidporter-id-au, line 235: mind just ensuring this is nil safe? Might be worth a small test

mkolodezny (approved):

service/worker/indexer/processor.go
neil-xie: nil condition is handled in the GetDomainID() function. It will return default empty string it domain id is nil

PR #5074, opened 2023-02-04
Hotfix a replication panic causing crashes

davidporter-id-au:
There's an uncaught panic in this codepath which crashes in the presence of probably corrupt data. The issue raises the wider question of uncaught panics not being handled in this processor, but I'll address that in a followup


demirkayaender (approved):

Groxx (approved):
makes sense, though I am curious how this happens - out-of-order replication or corrupt data maybe?  I'm not familiar with the related code, but it does seem like data existing _should_ be a safe assumption in general.

PR #5075, opened 2023-02-06
Adding panic-handling

davidporter-id-au:
## Context 
As a follow-up to the panic [here](https://github.com/uber/cadence/pull/5074) I'm keen to see that uncaught panics don't cause instability. So with a small refactor, adding a recovery to the replication daemon here. 

## Testing
Unit-testing

### DIFF
--- a/service/history/replication/task_processor.go
+++ b/service/history/replication/task_processor.go
@@ -175,41 +175,50 @@ func (p *taskProcessorImpl) Stop() {
 }
 
 func (p *taskProcessorImpl) processorLoop() {
-	defer func() {
-		p.logger.Debug("Closing replication task processor.", tag.ReadLevel(p.lastRetrievedMessageID))
-	}()
-
-Loop:
 	for {
-		// for each iteration, do close check first
-		select {
-		case <-p.done:
-			p.logger.Debug("ReplicationTaskProcessor shutting down.")
-			return
-		default:
-		}
-
-		respChan := p.sendFetchMessageRequest()
-
-		select {
-		case response, ok := <-respChan:
-			if !ok {
-				p.logger.Debug("Fetch replication messages chan closed.")
-				continue Loop
-			}
-
-			p.logger.Debug("Got fetch replication messages response.",
-				tag.ReadLevel(response.GetLastRetrievedMessageID()),
-				tag.Bool(response.GetHasMore()),
-				tag.Counter(len(response.GetReplicationTasks())),
-			)
-
-			p.taskProcessingStartWait()
-			p.processResponse(response)
-		case <-p.done:
-			return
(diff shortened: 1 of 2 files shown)


service/history/replication/task_processor.go
```diff
@@ -175,41 +175,50 @@ func (p *taskProcessorImpl) Stop() {
...
-
-		select {
-		case response, ok := <-respChan:
-			if !ok {
-				p.logger.Debug("Fetch replication messages chan closed.")
-				continue Loop
-			}
+func (p *taskProcessorImpl) processReplicationTasks() (stopping bool) {
+	defer func() {
+		e := recover()
```
davidporter-id-au, line 188: this is the only addition, apart from the refactoring

Groxx (changes requested):
minor but important change needed, otherwise LGTM and yay fewer `goto`s!

service/history/replication/task_processor.go
```diff
@@ -175,41 +175,50 @@ func (p *taskProcessorImpl) Stop() {
...
+		if e != nil {
+			p.logger.Error("A panic was encountered in replication task processing", tag.Dynamic("error", e))
+		}
+	}()
+	// for each iteration, do close check first
+	select {
+	case <-p.done:
+		p.logger.Debug("ReplicationTaskProcessor shutting down.")
+		return false
```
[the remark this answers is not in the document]
davidporter-id-au: good catch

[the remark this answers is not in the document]
davidporter-id-au: urgh... I did something dumb and I thought this changed already

service/history/replication/task_processor.go
```diff
@@ -175,41 +175,50 @@ func (p *taskProcessorImpl) Stop() {
...
 
-			p.taskProcessingStartWait()
-			p.processResponse(response)
-		case <-p.done:
-			return
+	select {
+	case response, ok := <-respChan:
+		if !ok {
+			p.logger.Debug("Fetch replication messages chan closed.")
+			return true
```
Groxx, line 207:
```suggestion
			return false
```
I think?  I believe this changed since the last round

service/history/replication/task_processor.go
```diff
@@ -175,41 +175,50 @@ func (p *taskProcessorImpl) Stop() {
...
+			tag.Bool(response.GetHasMore()),
+			tag.Counter(len(response.GetReplicationTasks())),
+		)
+
+		p.taskProcessingStartWait()
+		p.processResponse(response)
+	case <-p.done:
+		return true
 	}
+	return false
```
Groxx, line 221: ah, actually, should this be true too?
```suggestion
	return true
```

PR #5077, opened 2023-02-06
Fix resurrection check for timer and activity

Shaddoll:
**What changed?**
Fix the beginning event of resurrection checks

**Why?**
Bug fix

**How did you test it?**
local test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/integrity.go
+++ b/service/history/execution/integrity.go
@@ -38,21 +38,17 @@ func GetResurrectedTimers(
 	shard shard.Context,
 	mutableState MutableState,
 ) (map[string]struct{}, error) {
-	// 1. find min timer startedID for all pending timers
+	// 1. check if there is any pending timer
 	pendingTimerInfos := mutableState.GetPendingTimerInfos()
-	minTimerStartedID := common.EndEventID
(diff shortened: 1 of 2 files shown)


service/history/execution/integrity.go
```diff
@@ -62,7 +58,7 @@ func GetResurrectedTimers(
 	iter := collection.NewPagingIterator(getHistoryPaginationFn(
 		ctx,
 		shard,
-		minTimerStartedID,
+		1,
```
Groxx, line 61: ah, yeah, I guess this would have missed things that were lower ID somehow.  or if it entirely lost a timer.

makes sense 👍

Shaddoll: This actually can include invalid nodes. For example, the history branch may look like this 
```
node_id, txn_id
------------------
1,             10
2,             20 (contains event 2 and 3)
2,             11
3,             12
4,             21
```
node_id =3 is an invalid node, because the txn_id is lower than previous node. But if we read events from 3, that node is included, because we don't know the previous txn_id and do not discard that node.

PR #5080, opened 2023-02-08
UpdateWorkflow ShardId based metrics

allenchen2244:
**Describe what has changed in this PR**

Add shardId to updateworkflow. Also created a reusable thing to add more metrics around shardIds if needed.

**Tell your future self why have you made these changes**
For hotshard detection

**How have you verified this change? Tested locally? Added a unit test? Checked in staging env?** 
Tested on staging
https://ugrafana.uberinternal.com/d/000058231/cadence-persistence?editPanel=60&orgId=1&from=now-30m&to=now&var-deployment=staging-dca&var-deployment=staging-phx&var-services=All&var-latency=p99&var-operation=All

**Assuming the worst case, what can be broken when deploying this change to production?**
Metrics for updateworkflow break

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -1839,8 +1839,10 @@ const (
 	PersistenceEmptyResponseCounter
 
 	PersistenceRequestsPerDomain
+	PersistenceRequestsPerShard
 	PersistenceFailuresPerDomain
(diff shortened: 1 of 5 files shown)


neil-xie (approved):
lg

PR #5082, opened 2023-02-08
Emit workflow counts per workflow type metrics

neil-xie:
**What changed?**
Emit workflow counts per workflow type metrics
Fix the es query to exclude uninitialized workflow execution records

**Why?**

**How did you test it?**
Tested at staging2, added new panel to dashboard to see the results

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/esanalyzer/workflow.go
+++ b/service/worker/esanalyzer/workflow.go
@@ -123,6 +123,7 @@ func (w *Workflow) getWorkflowVersionQuery(domainName string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// exclude uninitialized workflow executions by checking whether record has start time field
 	return fmt.Sprintf(`
 {
     "aggs" : {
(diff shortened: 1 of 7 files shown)


allenchen2244:
Sorry, we should be fixing the names if we plan on adding more workflows to esanalyzer. I feel like having 1 wf do both things is a gross violation of SRP

service/worker/esanalyzer/workflow.go
```diff
@@ -206,6 +213,9 @@ func (w *Workflow) emitWorkflowVersionMetrics(ctx context.Context) error {
 				return err
 			}
 			for _, workflowType := range domainWorkflowVersionCount.WorkflowTypes {
+				w.analyzer.tallyScope.Tagged(
```
mkolodezny, line 216: nit: maybe have` {domainTag: domainName, workflowTypeTag: workflowType.AggregateKey} ` wrapped into var to dedup with loop below.

mkolodezny (approved):

service/worker/esanalyzer/workflow.go
neil-xie: Thanks! This is removed and emitted in a new workflow

allenchen2244 (approved):
lgtm

PR #5083, opened 2023-02-08
Add min_event_id,max_event_id flags to admin workflow show

Shaddoll:
**What changed?**
Add 2 cli flags

**Why?**
for debugging

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx:
Ah, there already was a limit but it wasn't controllable.  yeah, this is a solid improvement.

Might be good to get a warning printed when the response is "full" so you know you may have missed some data, but we didn't have that before either.

Groxx (approved):

PR #5086, opened 2023-02-09
Add Canary TLS support

charlese-instaclustr:
**What changed?**

Added the option for Cadence Canary to connect to Cadence over TLS via the grpc port. This requires including the filepath of an appropriate CA file in the Canary config.

**Why?**

Cadence is moving towards primarily support communication over grpc. TLS-enabled Cadence clusters can only conform with this standard while using Canary if Canary supports TLS-encrypted connections.

**How did you test it?**

Verified this change by running my branch of canary against a TLS-enabled Cadence cluster. Confirmed that canary logs show a successful connection with the cluster, and that `workflow_success` counter metrics increase for the `cadence_canary` domain.

Regression tested by running my branch of canary against a non-TLS Cadence cluster, and without using the new `tlsCaFile` config parameter. Again, found that canary logs show a successful connection with the cluster, and that `workflow_success` counter metrics increase for the `cadence_canary` domain.

**Potential risks**

In the worst case, this change could affect existing connections between canary and cadence in production; however, changes to the current path have been minimised, making this unlikely.

If the new logic were to not work, only users who attempted to leverage the new TLS capability of canary would be impacted, by Canary being unable to connect.

**Release notes**

No configuration changes are _required_, so not notable.

**Documentation Changes**

None required.


Shaddoll (approved):

charlese-instaclustr:
Hi @Shaddoll @shijiesheng (just tagging for visibility as you have both been active in the thread -- thanks for that),
This seems to be approved and the tests passing, could you please give me some support in getting this change merged in? Thanks in advance!

PR #5087, opened 2023-02-09
Fix prometheus frontend label inconsistencies

charlese-instaclustr:
**What changed?**

- Added dummy `domain` labels to frontend metrics emitted by admin frontend, to conform with the formatting of normal frontend metrics.
- Introduced a metric scope for the `DescribeCluster` handler, namely `AdminDescribeClusterScope`

**Why?**

Our team identified a bug whereby prometheus would fail to report some frontend metrics due to the fact that they shared a common fully-qualified name (e.g. cadence_requests and cadence_latency) but had inconsistencies in whether or not they included a `domain` label. For context, Prometheus cannot handle these inconsistent schemas, and will only report metrics which fit the first schema it encounters after startup. Attempting to report metrics which don't fit the dominant schema will trigger the following warning:
```
{"level":"warn","ts":"2023-02-09T15:55:18.934Z","msg":"error in prometheus reporter","service":"cadence-frontend","error":"a previously registered descriptor with the same fully-qualified name as Desc{fqName: \"cadence_latency\", help: \"cadence_latency summary\", constLabels: {}, variableLabels: [transport cadence_service operation caller]} has different label names or a different help string","logging-call-at":"metrics.go:151"}
{"level":"warn","ts":"2023-02-09T15:55:18.934Z","msg":"error in prometheus reporter","service":"cadence-frontend","error":"a previously registered descriptor with the same fully-qualified name as Desc{fqName: \"cadence_requests\", help: \"cadence_requests counter\", constLabels: {}, variableLabels: [transport cadence_service operation caller]} has different label names or a different help string","logging-call-at":"metrics.go:151"}
```

We identified this error happening in two different places:
- Metrics reporting from `adminHandler`, which previously did not tag metrics with domain labels
- Metrics reporting from `accessControlledHandler#ListDomains` specifically, because the ListDomains function lacks a domain context unlike other functions in that handler

We also opportunistically introduced the `AdminDescribeClusterScope` metrics scope because DescribeCluster currently reports metrics to the `AdminGetWorkflowExecutionRawHistoryV2Scope` which appears to be a copy-paste error.

**How did you test it?**

Test steps conducted:
- Built Cadence off my branch
- Ran Cadence with Cassandra
- Ran `cadence --do testDomain domain register`
- Ran `cadence admin cluster describe`
- Ran `cadence admin domain list`
- Collected the following metrics from Prometheus:

`curl http://localhost:7739/metrics | grep cadence_requests`
Screenshot 2023-02-09 at 4 46 30 pm
(note that we can see an admin operation being counted (DescribeCluster) as well non-admin metrics that have both dummy e.g. RegisterDomain and actual e.g. PollForDecisionTask domain labels. Also note that this shows our new metric scope for DescribeCluster working successfully).

`curl http://localhost:7739/metrics | grep cadence_latency_sum`
Screenshot 2023-02-09 at 4 47 38 pm
(note same as above, but here we are looking at latency metrics)

`curl http://localhost:7739/metrics | grep cadence_authorization_latency_sum`
Screenshot 2023-02-09 at 4 48 48 pm
(note specifically the presence of the ListDomain operation, which would previously not have been possible)

**Potential risks**

In the worst case, an error in the change could affect availability of Cadence frontend metrics. However, the approach used replicates existing patterns from how the `workflowHandler` reports on non domain-specific functionality (we reuse the pattern of labelling domain as "_unknown_" from there), so is expected to be robust. There are also no substantial changes to any of the metric reporting logic.

**Release notes**

Not notable for release.

**Documentation Changes**

None


mantas-sidlauskas (approved):

allenchen2244 (approved):
Can't say no if Mantas say yes

PR #5090, opened 2023-02-10
CLI: print all available dynamic config keys

mantas-sidlauskas:
**What changed?**
Adding CLI command to print out all available Dynamic Config keys with all details: name, description and default value.
For usability, `--format` flag can be used to print the list in preferred format (json, table, go template)

**Why?**
Currently, there are 369 configuration keys. 
This command will help to find needed key and also check what is the default value.

Snippet of the output:
```
$ cadence admin config describe
+-----------------------------------------------------------------------+---------------------------------------------------------------+--------------------------------+
|                                 NAME                                  |                          DESCRIPTION                          |         DEFAULT VALUE          |
+-----------------------------------------------------------------------+---------------------------------------------------------------+--------------------------------+
| admin.errorInjectionRate                                              | dminErrorInjectionRate is                                     |                              0 |
|                                                                       | the rate for injecting random                                 |                                |
|                                                                       | error in admin client                                         |                                |
| frontend.decisionResultCountLimit                                     | FrontendDecisionResultCountLimit                              |                              0 |
|                                                                       | is max number of decisions per                                |                                |
|                                                                       | RespondDecisionTaskCompleted                                  |                                |
|                                                                       | request                                                       |                                |
| frontend.disableListVisibilityByFilter                                | DisableListVisibilityByFilter                                 |                          false |
|                                                                       | is config to disable list                                     |                                |
|                                                                       | open/close workflow using                                     |                                |
|                                                                       | filter                                                        |                                |
| frontend.domainFailoverRefreshInterval                                | DomainFailoverRefreshInterval                                 |                            10s |
|                                                                       | is the domain failover refresh                                |                                |
|                                                                       | timer                                                         |                                |
| frontend.domainFailoverRefreshTimerJitterCoefficient                  | DomainFailoverRefreshTimerJitterCoefficient                   |                            0.1 |
|                                                                       | is the jitter for domain failover refresh                     |                                |

```

### DIFF
--- a/tools/cli/admin.go
+++ b/tools/cli/admin.go
@@ -1286,12 +1286,12 @@ func newAdminConfigStoreCommands() []cli.Command {
 			},
 		},
 		{
-			Name:    "describe",
-			Aliases: []string{"d"},
-			Usage:   "Describe all available configuration values",
+			Name:    "listall",
+			Aliases: []string{"la"},
+			Usage:   "List all available configuration keys",
 			Flags:   []cli.Flag{getFormatFlag()},
 			Action: func(c *cli.Context) {
-				AdminDescribeDynamicConfig(c)
+				AdminListConfigKeys(c)
 			},
 		},
 	}
--- a/tools/cli/admin.go
+++ b/tools/cli/admin.go
@@ -1285,5 +1285,14 @@ func newAdminConfigStoreCommands() []cli.Command {
 				AdminListDynamicConfig(c)
 			},
 		},
+		{
+			Name:    "describe",
+			Aliases: []string{"d"},
+			Usage:   "Describe all available configuration values",
+			Flags:   []cli.Flag{getFormatFlag()},
+			Action: func(c *cli.Context) {
+				AdminDescribeDynamicConfig(c)
+			},
+		},
 	}
 }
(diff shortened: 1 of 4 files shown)


tools/cli/admin.go
```diff
@@ -1280,5 +1280,14 @@ func newAdminConfigStoreCommands() []cli.Command {
 				AdminListDynamicConfig(c)
 			},
 		},
+		{
+			Name:    "describe",
```
Shaddoll, line 1284: `list` seems to be a better name

mantas-sidlauskas: Yeah, naming is hard.

There is `list-dynamic-config` argument already, and I'm proposing to rename it to `list` [in this PR](https://github.com/uber/cadence/pull/5089/files#diff-44b0fcadee023fdd151d83020b84d464e980b288f49d165b1115f2dd43c075eaR1280)

Shaddoll (approved):

tools/cli/admin.go
Shaddoll: how about `listall`? `describe` makes me think that it is a command requiring a key as input and only shows the description of that config.

mantas-sidlauskas: done :+1:

Shaddoll (approved):
Stamped, but there is a unit test failure

PR #5092, opened 2023-02-13
Small refactoring of task writer

Shaddoll:
**What changed?**
Move some methods from taskListManagerImpl to taskWriter

**Why?**
These methods are only used by taskWriter, thus it can improve the readability of the code.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskWriter.go
+++ b/service/matching/taskWriter.go
@@ -22,11 +22,16 @@
 package matching
 
 import (
+	"context"
 	"errors"
+	"fmt"
 	"sync/atomic"
 
+	"github.com/uber/cadence/common/backoff"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/messaging"
+	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/types"
 )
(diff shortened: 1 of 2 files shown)


service/matching/taskWriter.go
```diff
@@ -67,19 +75,31 @@ var errShutdown = errors.New("task list shutting down")
...
+		taskAckManager: tlMgr.taskAckManager,
+		stopCh:         make(chan struct{}),
+		appendCh:       make(chan *writeTaskRequest, tlMgr.config.OutstandingTaskAppendsThreshold()),
+		logger:         tlMgr.logger,
+		scope:          tlMgr.scope,
 	}
 }
 
-func (w *taskWriter) Start(block taskIDBlock) {
+func (w *taskWriter) Start() error {
```
davidporter-id-au, line 90: this this called anywhere else? I would just be worried about calling SetAckLevel erroneously

davidporter-id-au: its not, so that's all good

service/matching/taskWriter.go
```diff
@@ -172,7 +225,7 @@ writerLoop:
 					maxReadLevel = taskIDs[i]
 				}
 
-				r, err := w.tlMgr.db.CreateTasks(tasks)
+				r, err := w.db.CreateTasks(tasks)
```
davidporter-id-au, line 228: nice, that bidirectional reference thing is confusing

service/matching/taskWriter.go
Shaddoll: SetAckLevel doesn't return error
it's called inside start method of tasklistmanager

davidporter-id-au (approved):

PR #5094, opened 2023-02-13
CDNC-2088

bowenxia:
**What changed?**
Add AttemptCount into log

**Why?**
Currently we have metrics for number-of-attempts and have alerts on them, but finding the associated workflows can be tricky due to other logs also existing / temporary issues that self-recover happening all the time.

**How did you test it?**
Unit test

**Potential risks**
no

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/task/task.go
+++ b/service/history/task/task.go
@@ -236,10 +236,12 @@ func (t *taskImpl) HandleErr(
 			t.attempt++
 			if t.attempt > t.criticalRetryCount() {
 				t.scope.RecordTimer(metrics.TaskAttemptTimerPerDomain, time.Duration(t.attempt))
-				t.logger.Error("Critical error processing task, retrying.",
+				t.logger.Error("Critical error processing task, retrying.", // TODO: there might be more. If there is, do the same thing
 					tag.Error(err),
 					tag.OperationCritical,
+					// fixme!!! Add a new tag attempt into TaskType.log the attempt
 					tag.TaskType(t.GetTaskType()),
+					tag.AttemptCount(t.GetAttempt()),
 				)
 			}
 		}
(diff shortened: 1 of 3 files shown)


service/history/task/task.go
```diff
@@ -236,10 +236,12 @@ func (t *taskImpl) HandleErr(
 			t.attempt++
 			if t.attempt > t.criticalRetryCount() {
 				t.scope.RecordTimer(metrics.TaskAttemptTimerPerDomain, time.Duration(t.attempt))
-				t.logger.Error("Critical error processing task, retrying.",
+				t.logger.Error("Critical error processing task, retrying.", // TODO: there might be more. If there is, do the same thing
 					tag.Error(err),
 					tag.OperationCritical,
+					// fixme!!! Add a new tag attempt into TaskType.log the attempt
```
neil-xie, line 242: It is better to create a new ticket for it and link to your current ticket or just fix it if it is a simple fix.

bowenxia: Ohh I forgot to delete those comments. Those were reminders for myself. Thanks Neil.

bowenxia: The comments were deleted.

neil-xie:
from your note, do you need to add the tag here https://github.com/uber/cadence/blob/master/service/history/task/cross_cluster_task.go#L362

bowenxia:
@neil-xie Yes you are right! Nice catch. I totally missed that one. I've searched all other uses of the metric TaskAttemptTimerPerDomain, and made sure that everything was covered.

neil-xie (approved):
lg

davidporter-id-au (approved):
nice!

PR #5095, opened 2023-02-14
Small refactoring of taskReader

Shaddoll:
**What changed?**
- Add some fields to taskReader from taskListManager
- Small improvement on retry logic

**Why?**
Improve readability
The goal is to remove `taskListManager` field from `taskReader`

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -110,10 +110,9 @@ type (
 		// prevent tasks being dispatched to zombie pollers.
 		outstandingPollsLock sync.Mutex
 		outstandingPollsMap  map[string]context.CancelFunc
-
-		shutdownCh chan struct{}  // Delivers stop to the pump that populates taskBuffer
-		startWG    sync.WaitGroup // ensures that background processes do not start until setup is ready
-		stopped    int32
+		shutdownCh           chan struct{}  // Delivers stop to the pump that populates taskBuffer
+		startWG              sync.WaitGroup // ensures that background processes do not start until setup is ready
+		stopped              int32
 	}
 )
 
@@ -476,12 +475,7 @@ func (c *taskListManagerImpl) executeWithRetry(
 
 	throttleRetry := backoff.NewThrottleRetry(
 		backoff.WithRetryPolicy(persistenceOperationRetryPolicy),
-		backoff.WithRetryableError(func(err error) bool {
-			if _, ok := err.(*persistence.ConditionFailedError); ok {
-				return false
-			}
-			return persistence.IsTransientError(err)
-		}),
+		backoff.WithRetryableError(persistence.IsTransientError),
 	)
 	err = throttleRetry.Do(context.Background(), op)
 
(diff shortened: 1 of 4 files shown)


service/matching/taskListManager.go
```diff
@@ -476,12 +473,7 @@ func (c *taskListManagerImpl) executeWithRetry(
 
 	throttleRetry := backoff.NewThrottleRetry(
 		backoff.WithRetryPolicy(persistenceOperationRetryPolicy),
-		backoff.WithRetryableError(func(err error) bool {
-			if _, ok := err.(*persistence.ConditionFailedError); ok {
-				return false
-			}
-			return persistence.IsTransientError(err)
-		}),
+		backoff.WithRetryableError(persistence.IsTransientError),
```
davidporter-id-au, line 476: hm.. not related to this change, but we need to change persistence.IsTransientError to use errors.is/as

service/matching/taskReader.go
```diff
@@ -114,16 +142,15 @@ dispatchLoop:
 }
 
 func (tr *taskReader) getTasksPump() {
-	tr.tlMgr.startWG.Wait()
```
davidporter-id-au, line 117: oh wow, that's confusing

davidporter-id-au: is there are replacement for this WG in terms of functionality? how do we ensure that this only starts when ready?

service/matching/taskReader.go
```diff
@@ -158,58 +185,61 @@ getTasksPumpLoop:
...
 func (tr *taskReader) getTaskBatchWithRange(readLevel int64, maxReadLevel int64) ([]*persistence.TaskInfo, error) {
-	response, err := tr.tlMgr.executeWithRetry(func() (interface{}, error) {
-		return tr.tlMgr.db.GetTasks(readLevel, maxReadLevel, tr.tlMgr.config.GetTasksBatchSize())
-	})
+	var response *persistence.GetTasksResponse
+	op := func() (err error) {
+		response, err = tr.db.GetTasks(readLevel, maxReadLevel, tr.config.GetTasksBatchSize())
+		return
+	}
+	err := tr.throttleRetry.Do(context.Background(), op)
```
davidporter-id-au, line 220: this `throttleRetry.Do` appears to be semantically different than the `tr.tlMgr.executeWithRetry` function in that it isn't - unless I'm confused - handling conditional failures. 

Do you want to make this change in the refactor?

Shaddoll: GetTasks doesn't return conditional failures

service/matching/taskReader.go
Shaddoll: this wait isn't necessary, because it's the last step of start

service/matching/taskReader.go
davidporter-id-au: 🤦  yes

service/matching/taskReader.go
davidporter-id-au: good observation... I didn't notice

davidporter-id-au (approved):

PR #5097, opened 2023-02-14
Resolve an infinite loop around impossible cron schedules

Groxx:
We got some infinite (or near enough to infinite) hot loops in cron-schedule-Next-ing code today.
While we don't yet have any concrete examples to investigate, I'm going to claim it's due to being 2 weeks away from the end of February, and February is weird.  People are probably accidentally scheduling things for Feb 30 or 31.

There does seem to be some invalid cron schedule parsing happening in robfig/cron, e.g. for `0 0 0 , 0`, but the bulk of the blame for this rests on our code: `.Next` returns a zero-valued time if there is no next time, and that's `.Before` other dates... by a very large margin.
There's a decent chance the next `.Next` will return another zero value too, leading to a truly infinite loop.

Since a lot of this code was kinda ridiculous about not returning errors, I've just changed that and all callers.
In most cases there was a simple place "nearby" that already returned errors, so it just now returns another kind of error.

---

Because these invalid schedules are likely *always* invalid (robfig/cron checks 5 years of values, e.g. to work around leap years), and this would *always* infinite loop on those values, I believe it's safe to assume that these invalid cron schedules have never been written to databases.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -519,7 +519,7 @@ func CreateHistoryStartWorkflowRequest(
 	domainID string,
 	startRequest *types.StartWorkflowExecutionRequest,
 	now time.Time,
-) *types.HistoryStartWorkflowExecutionRequest {
+) (*types.HistoryStartWorkflowExecutionRequest, error) {
 	histRequest := &types.HistoryStartWorkflowExecutionRequest{
 		DomainUUID:   domainID,
 		StartRequest: startRequest,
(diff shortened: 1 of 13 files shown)


common/util.go
```diff
@@ -530,8 +530,12 @@ func CreateHistoryStartWorkflowRequest(
 	firstDecisionTaskBackoffSeconds := delayStartSeconds
 	if len(startRequest.GetCronSchedule()) > 0 {
 		delayedStartTime := now.Add(time.Second * time.Duration(delayStartSeconds))
-		firstDecisionTaskBackoffSeconds = backoff.GetBackoffForNextScheduleInSeconds(
+		var err error
+		firstDecisionTaskBackoffSeconds, err = backoff.GetBackoffForNextScheduleInSeconds(
```
shijiesheng, line 534: nit: 
```suggestion
		firstDecisionTaskBackoffSeconds, err := backoff.GetBackoffForNextScheduleInSeconds(
```

shijiesheng, line 534:
```suggestion
		firstDecisionTaskBackoffSeconds, err := backoff.GetBackoffForNextScheduleInSeconds(
```

shijiesheng (approved):
LGTM with one nit

common/util.go
Groxx: that declares a new `firstDecisionTaskBackoffSeconds` variable, rather than assigning it to the one outside the if block.

I really dislike Go's `:=` semantics :|  honestly I didn't notice this the first time, but thankfully Goland colors variables differently when they shadow others, so it was pretty obvious.

PR #5101, opened 2023-02-15
Unload taskListManager by instance, not taskListID

Shaddoll:
**What changed?**
cherry pick https://github.com/temporalio/temporal/pull/1917

**Why?**
see https://github.com/temporalio/temporal/pull/1917

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -1,5 +1,7 @@
-// Copyright (c) 2017-2020 Uber Technologies, Inc.
-//
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Portions of the Software are attributed to Copyright (c) 2020 Temporal Technologies Inc.
+
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


service/matching/matchingEngine.go
```diff
@@ -253,11 +255,14 @@ func (e *matchingEngineImpl) updateTaskList(taskList *taskListID, mgr taskListMa
...
 
-func (e *matchingEngineImpl) removeTaskListManager(id *taskListID) {
+func (e *matchingEngineImpl) removeTaskListManager(tlMgr taskListManager) {
+	id := tlMgr.TaskListID()
 	e.taskListsLock.Lock()
-	defer e.taskListsLock.Unlock()
-
-	delete(e.taskLists, *id)
+	currentTlMgr, ok := e.taskLists[*id]
+	if ok && tlMgr == currentTlMgr {
```
davidporter-id-au, line 262: I would add a comment as to why your'e checking both ID and instance referencing the description at https://github.com/temporalio/temporal/pull/1917, since honestly it's a little confusing without looking at the PR and surrounding comment

service/matching/matchingEngine.go
```diff
@@ -253,11 +255,14 @@ func (e *matchingEngineImpl) updateTaskList(taskList *taskListID, mgr taskListMa
...
+		delete(e.taskLists, *id)
+	}
+	e.taskListsLock.Unlock()
```
davidporter-id-au, line 265: nit: Would there be any reason to not defer this above the lock?

it's currently fine I think now, but if there's any other control flow changes It'd be easy to miss

service/matching/matchingEngine.go
```diff
@@ -836,16 +841,17 @@ func (e *matchingEngineImpl) getTask(
...
-	tlMgr, ok := e.taskLists[*id]
-	if ok {
-		delete(e.taskLists, *id)
+	currentTlMgr, ok := e.taskLists[*id]
+	if !ok || tlMgr != currentTlMgr {
+		e.taskListsLock.Unlock()
+		return
 	}
+	delete(e.taskLists, *id)
 	e.taskListsLock.Unlock()
```
davidporter-id-au, line 853: maybe we can defer this too and only have a single instance of unlock?

davidporter-id-au (approved):

service/matching/matchingEngine.go
Groxx: I think we want/need it to unlock before `Stop()` runs :\

in a more general sense though, yeah - I've been hoping we can come up with a better structure for these kinds of things.  Now that we have generics, it might even be palatable to use an IIFE...

davidporter-id-au: ah, good point

PR #5103, opened 2023-02-16
Flexible / sane header forwarding

Groxx:
Blindly forwarding all inbound headers to outbound requests is unambiguously wrong, and causing problems.

This adds a semi-flexible system to allow controlling what headers are forwarded, via static and dynamic config.
Dynamic config is currently only loaded at startup, but at least internally it's still quicker than changing static config.  We should consider making this truly dynamic.

As this has historically been copying ?? to outbounds, leading to ?? total headers: the current version does not change anything.  We'll likely experiment internally to make sure our new sets of headers are safe with our setup, and will make a separate PR to tighten down the list somewhat when we've confirmed that it works + is likely to work in other clusters.

---

Why this kind of add/remove/regex list structure?  No strong reason, I've just used it in previous projects pretty happily.  Only-inclusive or only-exclusive has been insufficient every time, but I've had this in a couple services for years without needing more.

Longer-term we probably want this to be truly dynamically updatable and/or fully pluggable.  If we're lucky we might not need more than this though, and driving it via config is pretty convenient and easily understood.


common/rpc/params.go
```diff
@@ -84,6 +85,15 @@ func NewParams(serviceName string, config *config.Config, dc *dynamicconfig.Coll
 		return Params{}, fmt.Errorf("public client outbound: %v", err)
 	}
 
+	forwardingRules, err := getForwardingRules(dc)
```
davidporter-id-au, line 88: fail closed? I think this seems reasonable, but a little risky, do we have a log?

davidporter-id-au (approved):
hard to fault this,

PR #5108, opened 2023-02-21
Feature/draft zonal partitioner spike

davidporter-id-au:
## Context 

This is 1/n of changes for Zonal isolation - the ability for workers and work within tasklists to be subdivided into zones (a regional subdivision) and to be able to be drained and isolated independently. The intention of this change is to provide a set of interfaces for use within matching, where most of the logic to do this isolation will occur.

## Changes

- Adds a partitioner library interface and default implementation (as an reference sample) of the partitioner library
- Adds types to support this internally

## To follow up
- IDL changes for feature
- Partitioner library more complete implementation and integration into frontend
- Persistence changes.


common/types/matching.go
```diff
@@ -37,6 +37,7 @@ type AddActivityTaskRequest struct {
 	Source                        *TaskSource               `json:"source,omitempty"`
 	ForwardedFrom                 string                    `json:"forwardedFrom,omitempty"`
 	ActivityTaskDispatchInfo      *ActivityTaskDispatchInfo `json:"activityTaskDispatchInfo,omitempty"`
+	PartitionConfig               []byte
```
Shaddoll, line 40: The type should be `PartitionConfig`

common/types/matching.go
```diff
@@ -425,6 +430,7 @@ type MatchingQueryWorkflowRequest struct {
 	TaskList      *TaskList             `json:"taskList,omitempty"`
 	QueryRequest  *QueryWorkflowRequest `json:"queryRequest,omitempty"`
 	ForwardedFrom string                `json:"forwardedFrom,omitempty"`
+	WorkerZone    string
```
Shaddoll, line 433: not sure if it's necessary

common/types/partition.go
```diff
@@ -0,0 +1,28 @@
...
+
+// PartitionConfig is a key/value based set of configuration for partitioning traffic. Intended to be a key/value pair
+// of data encoded in JSON or whatever encoding suits. This is intentionally opaque and to be passed blindly
+// to the partitioner of choice as it may contain business-specific types.
+//
+// Example of the intent:
+// partitionCfg := []byte(`{"wf-start-zone": "zone123", "userid: "1234", "weighting": 0.5}`)
+// which, for example, may allow the partitioner to choose to split traffic based on where the workflow started, or
+// the user, or any arbitrary other configuration
+type PartitionConfig []byte
```
Shaddoll, line 23: How does the client do encoding if we're not exposing the structure?

common/types/matching.go
davidporter-id-au: thank you, good catch

common/types/matching.go
davidporter-id-au: remove for now

common/types/partition.go
davidporter-id-au: I think as discussed, this is expected to be opaque

Issue #5109, opened 2023-02-21
Dynamicconfig.Client interface does not enforce consistent input types to client functions


charlese-instaclustr:
**Version of Cadence server, and client(which language)**
This is very important to root cause bugs.
- Server version: Latest (0.25.0)

**Describe the bug**

Dynamicconfig.Client is the interface shared by the various dynamic config client implementations, e.g. `file_based_client.go` and `config_store_client.go`. However, the interface is quite unrestrictive; for example, it specifies an input type of `interface {}` for the `UpdateValue` function for the values being updated. This has permitted diverging implementations in the actual clients, such that `config_store_client.go` implicitly expects a different datatype to `file_based_client.go` here.

This means that calls to `UpdateValue` (and other dynamic config client functions with similar problems) will not succeed for all DC implementations. For one concrete example,` adminHandler#AddSearchAttribute` calls `UpdateValue` with the input format expected by `file_based_client`; calling `AddSearchAttribute` while running the config store client therefore corrupts the config store and renders the `frontend.ValidSearchAttributes` config value unusable.

**To Reproduce**
Is the issue reproducible? 
- Yes

**Steps to reproduce the behavior:**
A clear and concise description of the reproduce steps.

- Run Cadence server with the config store dynamic config client enabled (config available in `development.yaml`)
- Add an initial value to the dynamic config for `frontend.validSearchAttributes` in the CLI: `cadence admin config updc --dynamic_config_name frontend.validSearchAttributes --dynamic_config_value '{"Value": '{"testAttribute": 2}',"Filters":[]}'`
- Verify that the config value has been added: `cadence admin config listdc`
- Trigger the `AddSearchAttribute` handler: `cadence admin cluster asa --search_attr_key testAttribute2 --search_attr_type 2`

The command will 'succeed', but after waiting ~10 seconds for the config to update, you will now find that the `frontend.validSearchAttributes` can no longer be written to or read from on this server.

**Expected behavior**

The Dynamic Config client interface and the client implementations must be refactored to be truly consistent in their input / output types. All calls to Dynamic config client methods throughout the codebase should succeed regardless of the specific client implementation in use.

**Additional context**
Add any other context about the problem here, E.g. Stackstace, workflow history.

Some additional context on the UpdateValue function specifically. Note that the `config_store_client` clearly anticipates `value` having a type of `[]*types.DynamicConfigValue`:

```
func (csc *configStoreClient) UpdateValue(name dc.Key, value interface{}) error {
[...]

PR #5110, opened 2023-02-21
Fix type validation in configstore DC client value updating

charlese-instaclustr:
Fixed a misleading check that was allowing `config_store_client#UpdateValue` to return no error, despite actually failing the operation and corrupting the dynamic config storage in the process.

Added a more verbose log message when the type check fails, to aid users in understanding the issue.

**Why?**

This is a temporary change mitigating a more widespread issue described in #5109 -- in summary, inconsistent assumptions about input data types across different implementations of the functions outlined in the `Dynamicconfig.Client` interface lead to instances in the codebase where the success / failure of calls to a dynamic config client depends on its specific implementation.

By making this change, the `config_store_client` will no longer attempt to update its values when its`UpdateValue` implementation is passed data in the format currently expected by the `file_based_config` client. This is an improvement because it will stop users inadvertently corrupting their dynamic config, and by consulting the updated log message, they will be informed that they can update their config via the CL instead.

However, long-term, a priority should be addressing issue #5109 and eliminating the inconsistency in this DC client interface.

**How did you test it?**

Tested locally.

- Ran Advanced Visibility Cadence server with the config store dynamic config client enabled (config available in `development.yaml`)
- Added an initial value to the dynamic config for frontend.validSearchAttributes in the CLI: `cadence admin config updc --dynamic_config_name frontend.validSearchAttributes --dynamic_config_value '{"Value": '{"testAttribute": 2}',"Filters":[]}'`
- Verified that the config value has been added: `cadence admin config listdc`
- Triggered the AddSearchAttribute handler: `cadence admin cluster asa --search_attr_key testAttribute2 --search_attr_type 2`. Note that this handler calls the `UpdateValue` function with the problematic input data format.
-  Reran: `cadence admin config listdc`. The prior set value for `frontend.validSearchAttributes` remains the same; by constrast, in the current `master` implementation, the prior set value would be gone.
- Checked the Cadence logs, and confirmed that the updated log message shows.

**Potential risks**

No notable risks from this change; we are simply fixing a type validation so that it serves its intended purpose.

**Release notes**

Not notable.

**Documentation Changes**

N/A


allenchen2244 (approved):

allenchen2244:
@charlese-instaclustr maybe test is failing somehow?

charlese-instaclustr:
@allenchen2244 good pickup, apologies, the tests picked up an edge case where the update value is nil. Added handling for the edge case to my change if you could rereview please. Thanks

charlese-instaclustr:
Hi @allenchen2244 , would you be able to give me a hand by rerunning the build and tests please? Thanks

PR #5111, opened 2023-02-21
Remove circular dependencies between matching components

Shaddoll:
**What changed?**
- Remove circular dependencies between matching components
- Move completeTask function to taskReader
- Remove unused method

**Why?**
Improve readability

**How did you test it?**
existing test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -35,6 +35,7 @@ import (
 	"github.com/uber/cadence/common/backoff"
 	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/cluster"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/messaging"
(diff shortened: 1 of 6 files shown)


service/matching/taskListManager.go
```diff
@@ -515,12 +480,12 @@ func (c *taskListManagerImpl) executeWithRetry(
 }
 
 func (c *taskListManagerImpl) trySyncMatch(ctx context.Context, params addTaskParams) (bool, error) {
-	task := newInternalTask(params.taskInfo, c.completeTask, params.source, params.forwardedFrom, true, params.activityTaskDispatchInfo)
+	task := newInternalTask(params.taskInfo, nil, params.source, params.forwardedFrom, true, params.activityTaskDispatchInfo)
```
Shaddoll, line 483: sync task doesn't need the completeTask function, since the complete function can never be invoked

service/matching/taskReader.go
```diff
@@ -297,6 +296,35 @@ func (tr *taskReader) persistAckLevel() error {
 	return nil
 }
 
-func (tr *taskReader) isTaskAddedRecently(lastAddTime time.Time) bool {
-	return time.Since(lastAddTime) <= tr.config.MaxTasklistIdleTime()
+// completeTask marks a task as processed. Only tasks created by taskReader (i.e. backlog from db) reach
+// here. As part of completion:
+//   - task is deleted from the database when err is nil
+//   - new task is created and current task is deleted when err is not nil
+func (tr *taskReader) completeTask(task *persistence.TaskInfo, err error) {
```
emrahs, line 303: Chatted offline about moving this from `taskListManager` to `taskReader`. The plan is to refactor this further so that this is `handleError` that remains in `taskReader` and move the ack level update logic back to the `taskListManager`. It'll be done in a separate PR.

emrahs (approved):

service/matching/taskReader.go
Shaddoll: Chatted offline, the ack level update logic will still stay in `taskReader`, but we will split the completeTask function to `AckTask` and `NackTask`

PR #5113, opened 2023-02-22
Bench: possibility to pass frontend address using env

mantas-sidlauskas:
**What changed?**
Providing "expand" option to config library. 
Small changes:
* Use 5s context for creating bench domains
* Stop bench if failed to create bench domain 

**Why?**
This will allow to point to different cadence cluster without changing config files.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


config/bench/development.yaml
```diff
@@ -5,7 +5,7 @@ bench:
 
 cadence:
   service: "cadence-frontend"
-  host: "127.0.0.1:7933"
+  host: "${CADENCE_FRONTEND_ADDRESS:127.0.0.1:7933}"
```
Shaddoll, line 8: where should we put the quotes? You put the quotes at different places.

mantas-sidlauskas: thanks, fixed

Shaddoll (approved):

PR #5114, opened 2023-02-23
Fix deadlock issue

Shaddoll:
**What changed?**
Fix deadlock issue

**Why?**
Fix deadlock issue

**How did you test it?**
eyeball check

**Potential risks**

**Release notes**

**Documentation Changes**


neil-xie (approved):

Groxx (approved):
👍 looks correct, yea.  Good find!

Any idea how much work it would be to split this into locked vs unlocked objects, to reduce this chance in the future?

PR #5119, opened 2023-02-24
Fix InsertTasks query for Cassandra

Shaddoll:
**What changed?**
- Fix the conditional update query of InsertTasks operation for Cassandra
- Code cleanup

**Why?**
The conditional update query of InsertTasks should be a no-op. When inserting tasks, we do the conditional update to make sure the rangeID doesn't change and we don't want to update the metadata of tasklists.

**How did you test it?**
existing persistence test

**Potential risks**
There is no risk. The change only affect the LastUpdate metadata of tasklist, which is used for garbage collection. However, we only run the garbage collector of tasklists with database not supporting TTL.

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/tasks.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/tasks.go
@@ -136,6 +136,15 @@ const (
 		`and task_id = ? ` +
 		`IF range_id = ?`
 
+	templateUpdateTaskListRangeIDQuery = `UPDATE tasks SET ` +
+		`range_id = ? ` +
+		`WHERE domain_id = ? ` +
+		`and task_list_name = ? ` +
+		`and task_list_type = ? ` +
+		`and type = ? ` +
+		`and task_id = ? ` +
+		`IF range_id = ?`
+
 	templateDeleteTaskListQuery = `DELETE FROM tasks ` +
 		`WHERE domain_id = ? ` +
 		`AND task_list_name = ? ` +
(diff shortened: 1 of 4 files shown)


common/persistence/nosql/nosqlplugin/cassandra/tasks.go
```diff
@@ -380,17 +387,11 @@ func (db *cdb) InsertTasks(
 	}
 
 	// The following query is used to ensure that range_id didn't change
-	batch.Query(templateUpdateTaskListQuery,
+	batch.Query(templateUpdateTaskListRangeIDQuery,
```
emrahs, line 390: This change made me think whether it makes sense or not, so I thought I should dump my thought process here if anyone is curious.

With this change, `ackLevel` and `lastUpdated` are no longer updated at the time of task insertion. During steady state, that shouldn't make a difference since the task pump updates those fields periodically (every 1m). However, when tasklist lease ownership changes (due to host restart/add/remove, certain persistence errors) we will have a larger window of tasks that are re-dispatched due to ack-level reset during lease renewal. That makes me question whether this change is worth the trade-off or not. The main question is "do we get any value out of not updating the `tasklist` column from the Cassanda perspective?". 

Since this is a LWT query, the main cost of the query will be on paxos, so there compute or latency-related gains will be negligible. However, doing a partial update here (ie. not updating the `tasklist` column) reduces the amount of data stored in the sstable for each task insertion query. Therefore, there is some storage benefit as well as some compute benefit on the compaction side due to having less amount of data to compact. 

Given the benefits on the Cassandra side and the fact that we care a lot about storage optimization, I think this change is justified. While we are sacrificing the quality of the `ackLevel` freshness a little bit, we can remediate that to some degree by increasing the frequency of the periodic `ackLevel` updates using the `matching.updateAckInterval` config in clusters where we care more about efficiency at the Cadence layer rather than efficiency at the Cassandra layer.

emrahs (approved):

PR #5123, opened 2023-02-28
Visiblity shard id to open source stuff

allenchen2244:
**What changed?**
Add shard_id to open source visiblity.

**Why?**
For open source customers

**How did you test it?**
Unit/int tests pass

**Potential risks**
Open source break for open source users not uber

**Release notes**
Update schemas for database solution that is needed


neil-xie (approved):
Looks good

PR #5124, opened 2023-03-01
create RequestBody struct, implement it using requests, add requestbody into Attribute for auditing

bowenxia:
**What changed?**
create RequestBody struct, 
implement it using requests,
add RequestBody into Attribute
add a log tag for request body

**Why?**
create RequestBody struct -> as an interface
implement it using requests -> we can store different requests as RequestBody, and can use a bodyToString method
to transfer a struct into string. So that it is easier to log
add RequestBody into Attribute -> auditing manual accesses by logging the request body

**How did you test it?**
unit test

**Potential risks**
Leaking PII? -> But have already manually filtered fields to be logged. 

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/authorization/authorizer.go
+++ b/common/authorization/authorizer.go
@@ -87,7 +87,7 @@ func NewPermission(permission string) Permission {
 
 // Authorizer is an interface for authorization
 type Authorizer interface {
-	Authorize(ctx context.Context, attributes *Attributes) (Result, error)
+	Authorize(ctx context.Context, attributes *Attributes, auditInfo AuditInfo) (Result, error)
 }
 
 func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvider, error) {
@@ -96,4 +96,8 @@ func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvide
 		return nil, fmt.Errorf("invalid private key path %s", privateKey)
 	}
 	return clientworker.NewAdminJwtAuthorizationProvider(pk), nil
+}
+
+type AuditInfo interface {
+	ToAuditInfo() string
 }
--- a/common/authorization/authorizer.go
+++ b/common/authorization/authorizer.go
@@ -58,7 +58,7 @@ type (
 		WorkflowType *types.WorkflowType
 		TaskList     *types.TaskList
 		Permission   Permission
-		AuditInfo    AuditInfo
+		RequestBody  RequestBody
 	}
 
 	// Result is result from authority.
@@ -99,6 +99,6 @@ func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvide
 	return clientworker.NewAdminJwtAuthorizationProvider(pk), nil
 }
 
-type AuditInfo interface {
-	ToAuditInfo() string
+type RequestBody interface {
+	BodyToString() string
 }
--- a/common/authorization/authorizer.go
+++ b/common/authorization/authorizer.go
@@ -58,6 +58,7 @@ type (
 		WorkflowType *types.WorkflowType
 		TaskList     *types.TaskList
 		Permission   Permission
+		AuditInfo    AuditInfo
 	}
 
 	// Result is result from authority.
@@ -87,7 +88,7 @@ func NewPermission(permission string) Permission {
 
 // Authorizer is an interface for authorization
 type Authorizer interface {
-	Authorize(ctx context.Context, attributes *Attributes, auditInfo AuditInfo) (Result, error)
+	Authorize(ctx context.Context, attributes *Attributes) (Result, error)
 }
 
 func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvider, error) {
--- a/common/types/admin.go
+++ b/common/types/admin.go
@@ -26,6 +26,10 @@ type AddSearchAttributeRequest struct {
 	SecurityToken   string                      `json:"securityToken,omitempty"`
 }
 
+func (v *AddSearchAttributeRequest) BodyToString() string {
+	return StructToString(*v)
+}
+
 // GetSearchAttribute is an internal getter (TBD...)
 func (v *AddSearchAttributeRequest) GetSearchAttribute() (o map[string]IndexedValueType) {
 	if v != nil && v.SearchAttribute != nil {
@@ -45,6 +49,10 @@ type DescribeClusterResponse struct {
 type AdminDescribeWorkflowExecutionRequest struct {
 	Domain    string             `json:"domain,omitempty"`
 	Execution *WorkflowExecution `json:"execution,omitempty"`
+}
+
+func (v *AdminDescribeWorkflowExecutionRequest) BodyToString() string {
+	return StructToString(*v)
 }
 
 // GetDomain is an internal getter (TBD...)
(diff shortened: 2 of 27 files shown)


service/frontend/accessControlledHandler.go
```diff
@@ -882,3 +1022,9 @@ func (a *AccessControlledWorkflowHandler) getMetricsScopeWithDomainName(
 ) metrics.Scope {
 	return a.GetMetricsClient().Scope(scope).Tagged(metrics.DomainTag(domainName))
 }
+
+func isManual(ctx context.Context) bool {
+	callContext := yarpc.CallFromContext(ctx)
+	caller := callContext.Caller()
+	return caller == "cadence-cli" || caller == "cadence-web" || caller == "cadence-ui"
```
demirkayaender, line 1029: Could you add a comment above that sometime in the future we should also validate the caller.

common/types/shared.go
```diff
@@ -5572,6 +5688,50 @@ type SignalWithStartWorkflowExecutionRequest struct {
 	JitterStartSeconds                  *int32                 `json:"jitterStartSeconds,omitempty"`
 }
 
+func (v *SignalWithStartWorkflowExecutionRequest) BodyToString() string {
+	cloneRequest := SignalWithStartWorkflowExecutionRequest{
```
demirkayaender, line 5692: Add a comment here why you are manually assigning attributes. You are trying to prevent logging workflow params, i.e. user inputs.

common/types/shared.go
```diff
@@ -5962,6 +6142,45 @@ type StartWorkflowExecutionRequest struct {
...
 
+func (v *StartWorkflowExecutionRequest) BodyToString() string {
+	cloneRequest := StartWorkflowExecutionRequest{
+		Domain:       v.Domain,
+		WorkflowID:   v.WorkflowID,
+		Identity:     v.Identity,
+		RequestID:    v.RequestID,
+		CronSchedule: v.CronSchedule,
+	}
+	// deep copy the value within pointers
```
demirkayaender, line 6153: ditto

common/authorization/authorizer.go
```diff
@@ -58,6 +58,7 @@ type (
 		WorkflowType *types.WorkflowType
 		TaskList     *types.TaskList
 		Permission   Permission
+		RequestBody  ManualRequestBody
```
demirkayaender, line 61: Maybe let's call ManualRequestBody as FilteredRequestBody.

common/types/shared.go
```diff
@@ -5572,6 +5688,50 @@ type SignalWithStartWorkflowExecutionRequest struct {
...
+func (v *SignalWithStartWorkflowExecutionRequest) BodyToString() string {
+	cloneRequest := SignalWithStartWorkflowExecutionRequest{
+		Domain:       v.Domain,
+		WorkflowID:   v.WorkflowID,
+		SignalName:   v.SignalName,
+		Identity:     v.Identity,
+		RequestID:    v.RequestID,
+		CronSchedule: v.CronSchedule,
+	}
+	// deep copy the value within pointers
```
demirkayaender, line 5700: same here, explain why you are doing this manually

common/types/shared.go
```diff
@@ -5679,6 +5839,22 @@ type SignalWorkflowExecutionRequest struct {
...
 }
 
+func (v *SignalWorkflowExecutionRequest) BodyToString() string {
+	cloneSignalWorkflowExecutionRequest := SignalWorkflowExecutionRequest{
+		Domain:     v.Domain,
+		SignalName: v.SignalName,
+		Identity:   v.Identity,
+		RequestID:  v.RequestID,
+	}
+	// deep copy the value within a pointer
```
demirkayaender, line 5849: ditto

service/frontend/accessControlledHandler.go
```diff
@@ -706,6 +818,11 @@ func (a *AccessControlledWorkflowHandler) StartWorkflowExecution(
 		Permission:   authorization.PermissionWrite,
 		WorkflowType: request.WorkflowType,
 	}
+
+	if isManual(ctx) {
+		attr.RequestBody = request
```
demirkayaender, line 823: For all the requests that can contain PII, add a comment here that the authorizer should use the new function you introduced while logging requests to avoid revealing private information.

common/types/admin.go
```diff
@@ -26,6 +26,10 @@ type AddSearchAttributeRequest struct {
 	SecurityToken   string                      `json:"securityToken,omitempty"`
 }
 
+func (v *AddSearchAttributeRequest) BodyToString() string {
```
demirkayaender, line 29: The naming here is a bit misleading. It's not really stringifying the struct, it's outputting a string for logging. You probably rename this to something like "SerializeForLogging". Same for the util function.

service/frontend/accessControlledHandler.go
bowenxia: done

common/types/shared.go
bowenxia: done

common/types/shared.go
bowenxia: done

common/authorization/authorizer.go
bowenxia: done

common/types/shared.go
bowenxia: done

common/types/shared.go
bowenxia: done

service/frontend/accessControlledHandler.go
bowenxia: done

common/types/admin.go
bowenxia: done

common/types/shared.go
```diff
@@ -5572,6 +5688,53 @@ type SignalWithStartWorkflowExecutionRequest struct {
 	JitterStartSeconds                  *int32                 `json:"jitterStartSeconds,omitempty"`
 }
 
+func (v *SignalWithStartWorkflowExecutionRequest) SerializeForLogging() string {
+	// making a deep copy of the struct for manually assigning attributes
+	cloneRequest := SignalWithStartWorkflowExecutionRequest{
```
Shaddoll, line 5693: This is not a good pattern. These logic is hard to maintain.
Consider using json encoding to serialize the structs.

bowenxia: done

common/types/shared.go
```diff
@@ -7842,3 +8083,9 @@ type RespondCrossClusterTasksCompletedResponse struct {
 type StickyWorkerUnavailableError struct {
 	Message string `json:"message,required"`
 }
+
+// SerializeRequest Serialize an arbitrary request for logging
+func SerializeRequest(request interface{}) string {
+	res, _ := json.Marshal(request)
```
Shaddoll, line 8089: It's not good to ignore the error here.

common/types/shared.go
```diff
@@ -5925,6 +6109,10 @@ type RestartWorkflowExecutionRequest struct {
 	Identity          string             `json:"identity,omitempty"`
 }
 
+func (v *RestartWorkflowExecutionRequest) SerializeForLogging() (string, error) {
+	return SerializeRequest(*v)
```
Shaddoll, line 6113: you can pass the pointer as the parameter, which will save a deep copy

service/frontend/accessControlledAdminHandler.go
```diff
@@ -251,6 +306,11 @@ func (a *AccessControlledWorkflowAdminHandler) PurgeDLQMessages(ctx context.Cont
 		APIName:    "PurgeDLQMessages",
 		Permission: authorization.PermissionAdmin,
 	}
+
+	if isManual(ctx) {
```
Shaddoll, line 310: it seems that it's worth moving `isManual` check into `isAuthorized`

common/types/shared.go
bowenxia: done

service/frontend/accessControlledAdminHandler.go
bowenxia: Since we change to store a pointer instead of a struct in Attributes, as we discussed, we don't need to check if it is manual any more. Because a pointer won't occupy too much space.

Shaddoll (approved):

PR #5125, opened 2023-03-01
Scanner to purge deprecated domain workflows

agautam478:
**What changed?**
Added domain Status check for the deprecated domains in the invariants.

**Why?**
The changes were made to enable identification of the workflows that are associated with the deprecated domains and need deletion.

**How did you test it?**
Updated existing unit tests and tested locally.

**Potential risks**
The change is not a breaking change. worse case scenario is that some workflows might be missed from deletion. the risk of active workflows getting removed is minimal due to the existence of a hard condition on domain status.

**Release notes**
NA

**Documentation Changes**
NA

### DIFF
--- a/common/reconciliation/invariant/historyExists.go
+++ b/common/reconciliation/invariant/historyExists.go
@@ -24,6 +24,7 @@ package invariant
 
 import (
 	"context"
+	"fmt"
 
 	c "github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/cache"
@@ -69,7 +70,8 @@ func (h *historyExists) Check(
 		}
 	}
 	domainID := concreteExecution.GetDomainID()
-	domainName, errorDomainName := h.dc.GetDomainName(domainID)
+	domain, errorDomainName := h.dc.GetDomainByID(domainID)
+	domainName := domain.GetInfo().Name
 	if errorDomainName != nil {
 		return CheckResult{
 			CheckResultType: CheckResultTypeFailed,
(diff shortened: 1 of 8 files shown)


common/reconciliation/invariant/historyExists.go
```diff
@@ -78,7 +80,16 @@ func (h *historyExists) Check(
 			InfoDetails:     errorDomainName.Error(),
 		}
 	}
-
+	domainName := domain.GetInfo().Name
```
davidporter-id-au, line 83: @demirkayaender I was talking to Aditi about this: 

The premise of an invariant check is like an integration-test or an integrated assertion, it's a check that some set of things that should be true are indeed so. 

This particular assertion in this function is that history exists and is correct. It doesn't look to domains in any way nor would anyone running this invariant expect it to care about domain status. 

My opinion is that - insofar as code organisation goes - a completely new invariant here makes sense: An assertion that the workflow exists for a domain that is active and not deprecated. That'd be less surprising for other engineers looking at the contents of the invariants, as well as easier to control with flipr rollout and similar.

agautam478: I have a separate refactoring ticket since the code will require considerable changes before we can separate out the checks like this one. For now how about we remove the check from history invariant and let the other executions go through? I have placed the same check for concreteExecution and openCurrentExecutions as well.

common/reconciliation/invariant/inactiveDomainExists.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type (
+	inactiveDomainExists struct {
+		pr persistence.Retryer
+		dc cache.DomainCache
+	}
+)
+
+func NewInactiveDomainExists(
```
davidporter-id-au, line 45: you'll probably need to manually invoke this somewhere like the other constructors

davidporter-id-au (approved):

PR #5140, opened 2023-03-08
[history] refactor taskfilter logic in transfer and timer task queue processor

shijiesheng:
**What changed?**

* put task filtering logic into taskAllocator

**Why?**

* it's easy to miss the filtering logic if they exist in two places

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/queue/transfer_queue_processor.go
```diff
@@ -514,24 +514,6 @@ func newTransferQueueStandbyProcessor(
 		if !ok {
 			return false, errUnexpectedQueueTask
 		}
-		if task.TaskType == persistence.TransferTaskTypeCloseExecution ||
```
Shaddoll, line 517: These logic was put here on purpose, because we don't want to hide it

Shaddoll (approved):

service/history/queue/task_allocator.go
```diff
@@ -189,3 +191,20 @@ func (t *taskAllocatorImpl) Lock() {
 func (t *taskAllocatorImpl) Unlock() {
 	t.locker.Unlock()
 }
+
+// isDomainNotRegistered checks either if domain does not exist or is in deprecated or deleted status
+func isDomainNotRegistered(shard shard.Context, domainID string) (bool, error) {
+	domainEntry, err := shard.GetDomainCache().GetDomainByID(domainID)
+	if err != nil {
+		if _, ok := err.(*types.EntityNotExistsError); ok {
+			return true, nil
```
demirkayaender, line 200: Should we be more conservative here? Normally we never delete domains

demirkayaender: If we have a timer and we cannot find the domain, something must be wrong.

service/history/queue/task_allocator.go
```diff
@@ -189,3 +191,20 @@ func (t *taskAllocatorImpl) Lock() {
 func (t *taskAllocatorImpl) Unlock() {
 	t.locker.Unlock()
 }
+
+// isDomainNotRegistered checks either if domain does not exist or is in deprecated or deleted status
+func isDomainNotRegistered(shard shard.Context, domainID string) (bool, error) {
```
demirkayaender, line 196: Let's change this to isDomainDeprecated

PR #5142, opened 2023-03-09
add remaining persistence stuff that goes to a shard

allenchen2244:
**What changed?**
Add remaining persistence metrics that goes to a shard

**Why?**
Detect hotshards

**How did you test it?**
Unit tests/int test pass

**Potential risks**
I can't see anything that bad that could happen.


davidporter-id-au (approved):
Nice use of the sampleRate for logging. Hopefully dynamicConfig doesn't mind being hit that hard

PR #5146, opened 2023-03-10
Adds partitioning library

davidporter-id-au:
Adds partitioning library:

- Adds an interface for handling workflow isolation group redirection and failover
- Adds an interface for handling drain states for both global and domain-based drains
- Adds relevant internal types
- Adds a simple interface for handling global drains as stored in database-based dynamic config
- Adds a default implementation for the open-source implementation and implementation example


common/persistence/isolationGroupDrains.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"context"
+
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/types"
+)
+
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination zoneDrains_mock.go -self_package github.com/uber/cadence/common/persistence
+
+type GlobalIsolationGroupDrains interface {
```
Shaddoll, line 34: I don't think this interface is needed. We're using the dynamic config store to persist the drain config and should use the dynamic client to write and read the config.

common/types/partition.go
```diff
@@ -0,0 +1,61 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package types
+
+const (
+	IsolationGroupStatusInvalid IsolationGroupStatus = iota
```
Shaddoll, line 26: I think these definitions should be moved to `partition` package.
The `types` package should only contains the types defined in `idl`.

common/partition/default-partitioner.go
```diff
@@ -0,0 +1,181 @@
...
+// where the workflow was started, and is expected to be pinned, and a RunID for a fallback means
+// to partition data deterministically.
+type DefaultWorkflowPartitionConfig struct {
+	WorkflowStartIsolationGroup types.IsolationGroupName
+	RunID                       string
+}
+
+// DefaultPartitioner is a business-agnostic implementation of partitioning
+// which is used by the Cadence system to allocate workflows in matching by isolation-group
+type DefaultPartitioner struct {
```
Shaddoll, line 53: I think this should be unexported

common/persistence/isolationGroupDrains.go
davidporter-id-au: yes, this is for dynamicConfig, it's intended to use that in the DB with a different config type https://sourcegraph.uberinternal.com/github.com/uber/cadence/-/blob/common/persistence/dataManagerInterfaces.go#L224

common/types/partition.go
davidporter-id-au: ok! no worries

common/partition/partitioning.go
```diff
@@ -0,0 +1,65 @@
...
+	allIsolationGroups                []types.IsolationGroupName
+}
+
+// State is a convenience return type of a collection of IsolationGroup configurations
+type State struct {
+	Global types.IsolationGroupConfiguration
+	Domain types.IsolationGroupConfiguration
+}
+
+type Partitioner interface {
```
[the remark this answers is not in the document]
davidporter-id-au: let's sync to understand this

common/partition/default-isolation-group-state.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"fmt"
+
+	"github.com/uber/cadence/common/persistence"
+
+	"github.com/uber/cadence/common/cache"
+	"github.com/uber/cadence/common/log"
+)
+
+type DefaultIsolationGroupStateHandler struct {
```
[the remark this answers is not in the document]
davidporter-id-au: sure!

common/persistence/isolationGroupDrains.go
davidporter-id-au: removed as per discussion, this will live at a higher-level in `common/isolation-groups`

davidporter-id-au:
Closing this because I've apparently messed up the diff in favour of https://github.com/uber/cadence/pull/5169

PR #5151, opened 2023-03-10
Adding request body into Attributes for auditing purpose with PII fields are filtered

bowenxia:
**What changed?**
create FilteredRequestBody struct,
implement it using requests,
add FilteredRequestBody into Attribute
add a log tag for request body
add a email tag for actor emails
add nil checks/handles for nil pointers to PR: https://github.com/uber/cadence/pull/5124

**Why?**
create FilteredRequestBody struct -> as an interface
implement it using requests -> we can store different requests as FilteredRequestBody, and can use a SerializeForLogging method
to transfer a struct into string. So that it is easier to log
add FilteredRequestBody into Attribute -> auditing manual accesses by logging the request body
PR: https://github.com/uber/cadence/pull/5124 had bugs (Panic: pointer is nil) when sending startWorkflowRequest & other requests that filtered PII for auditing. 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


bowenxia:
The actual changes for this PR are only within the **first two** commits. Commit 4 has a lot of changes but that was a revert of a revert: https://github.com/uber/cadence/pull/5145.

common/types/shared.go
```diff
@@ -5572,6 +5689,76 @@ type SignalWithStartWorkflowExecutionRequest struct {
 	JitterStartSeconds                  *int32                 `json:"jitterStartSeconds,omitempty"`
 }
 
+func (v *SignalWithStartWorkflowExecutionRequest) SerializeForLogging() (string, error) {
```
Groxx, line 5692: As a broad bit of critique / feedback:
Fairly often this kind of thing gets missed / falls out of sync during future changes to a structure (though I admit many of these structs are _relatively_ unlikely to see changes that would be a significant problem).

A less-fragile option would be to use reflection and struct tags, as those are immediately visible when you make any modification to the struct.  Then all of these (a shallow + shallow-copy-slice-contents copy) could be built once and not need to be customized.

---

Otherwise, as this is not a more standard `x = y` copy or a deep-copy like "`json.Marshal` -> `json.Unmarshal`": why are values being selected / ignored, and why do semi-deep copies at all?

I don't mean to imply there is no reason or this should not be done, but diverging from a common operation should be justified, as it will be surprising to people in the future and needs to be maintained.
- Are we worried about mutation, and this non-mutated copy is more correct somehow?
- Would it be reasonable to do a shallow copy and then erase specific fields?  e.g. not logging `Input` seems completely reasonable, but `dup = *v; dup.Input = nil` is trivial compared to this code.
- Related to struct tags: much of this could be accomplished by `Input []byte 'json:"-"'` as this is immediately json-marshaled.  Would that be better?  Would it cause issues in some other logging / json-ifying location?
- Zap (the underlying logger) has a MarshalLogObject API pretty much entirely for this purpose.  Can we use it / do we have a mirror?  And can we generate it instead of writing it by hand?

Groxx (changes requested):
Blocking purely for high-level questions.  A relatively brief skim looked fine and I'm more than happy to do the whole thing, it just needs to make the need-for-complexity clear.

common/types/shared.go
bowenxia: After a great discussion, decided to use json tags to filter PII.

common/types/admin.go
```diff
@@ -26,6 +26,13 @@ type AddSearchAttributeRequest struct {
 	SecurityToken   string                      `json:"securityToken,omitempty"`
 }
 
+func (v *AddSearchAttributeRequest) SerializeForLogging() (string, error) {
+	if v == nil {
+		return "", nil
+	}
```
Groxx, line 32: tbh I might just put this in SerializeRequest since it really should be the same every time.

common/types/shared.go
```diff
@@ -5565,13 +5769,20 @@ type SignalWithStartWorkflowExecutionRequest struct {
 	Control                             []byte                 `json:"control,omitempty"`
```
Groxx, line 5769: may want to suppress `SignalInput` as well (one line above, but I can't comment there 😢)

common/types/shared_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+func TestSignalWithStartWorkflowExecutionRequestSerializeForLogging(t *testing.T) {
+	tests := map[string]struct {
+		input               *SignalWithStartWorkflowExecutionRequest
+		expectedOutput      string
+		expectedErrorOutput error
+	}{
+		"complete request without error": {
+			input:               createNewSignalWithStartWorkflowExecutionRequest(),
+			expectedOutput:      "{\"domain\":\"testDomain\",\"workflowId\":\"testWorkflowID\",\"workflowType\":{\"name\":\"testWorkflowType\"},\"taskList\":{\"name\":\"testTaskList\",\"kind\":\"STICKY\"},\"executionStartToCloseTimeoutSeconds\":1,\"taskStartToCloseTimeoutSeconds\":1,\"identity\":\"testIdentity\",\"requestId\":\"DF66E35D-A5B0-425D-8731-6AAC4A4B6368\",\"workflowIdReusePolicy\":\"AllowDuplicate\",\"signalName\":\"testRequest\",\"signalInput\":\"dGVzdFNpZ25hbElucHV0\",\"control\":\"dGVzdENvbnRyb2w=\",\"retryPolicy\":{\"initialIntervalInSeconds\":1,\"backoffCoefficient\":1,\"maximumIntervalInSeconds\":1,\"maximumAttempts\":1,\"nonRetriableErrorReasons\":[\"testArray\"],\"expirationIntervalInSeconds\":1},\"cronSchedule\":\"testSchedule\",\"header\":{},\"delayStartSeconds\":1,\"jitterStartSeconds\":1}",
```
Groxx, line 39: as these blobs are kinda large / opaque to skim (it's unclear what exactly is being tested / what this test cares about), another test that may be worth having: 
serialize something with a recognizable string somewhere in all the pii fields (e.g. "PII"), and assert that the output doesn't contain it.  since that's what we actually want to claim, not that the structure never changes.

common/types/shared.go
Groxx: for more context: we don't rely on these json tags for any important _behavior_ as far as we are aware (the codebase is quite consistent about explicitly mapping between types, not monkey-patching with serialization), and we can't find any logs including even "input" so it's quite likely these are all unused.

common/types/admin.go
bowenxia: Typed nil. Keep as-is.

common/types/shared.go
bowenxia: done

common/types/shared_test.go
bowenxia: Create a new struct that only contains PII. Would be much more clearer to see.

common/types/admin.go
Groxx: yea.  typed nils are worth avoiding on principle I think

common/types/shared.go
```diff
@@ -7842,3 +8095,18 @@ type RespondCrossClusterTasksCompletedResponse struct {
...
 	Message string `json:"message,required"`
 }
+
+// SerializeRequest Serialize an arbitrary request for logging
+// pass in a pointer as a parameter to save space
+func SerializeRequest(request interface{}) (string, error) {
+	if request == nil {
+		return "", nil
+	}
+
```
Groxx, line 8105:
```suggestion
```
probably just leftovers from trying the cleanup

common/types/shared_test.go
```diff
@@ -0,0 +1,317 @@
...
+		SearchAttributes:   &SearchAttributes{IndexedFields: map[string][]byte{}},
+		Header:             &Header{Fields: map[string][]byte{}},
+		DelayStartSeconds:  &testDelayStartSeconds,
+		JitterStartSeconds: &testJitterStartSeconds,
+	}
+	return testReq
+}
+
+type PiiSampleRequest struct {
+	Input            []byte            `json:"-"` // Filtering PII
```
Groxx, line 291: this is effectively testing `encoding/json`, which we can broadly assume is correct.

what we want is to ensure we don't start logging these fields _on the real objects_.  this won't protect against that at all.

that said: creating a pii-field-only instance of the struct and comparing against `{}` seems effective!  I think that'll work on the real things too, since everything(?) is `omitempty`.

bowenxia: deleted PiiSampleRequest

common/types/shared.go
bowenxia: done

Groxx (approved):
:shipit:

PR #5153, opened 2023-03-13
Update history to persist partition config

Shaddoll:
**What changed?**
Update history to persist partition config in mutable state and workflow started event and forward it to matching when pushing activity and decision tasks to matching

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/frontend/workflowHandler.go
+++ b/service/frontend/workflowHandler.go
@@ -2182,7 +2182,7 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 
 	wh.GetLogger().Debug("Start workflow execution request domainID", tag.WorkflowDomainID(domainID))
 	historyRequest, err := common.CreateHistoryStartWorkflowRequest(
-		domainID, startRequest, time.Now())
+		domainID, startRequest, time.Now(), nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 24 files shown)


service/frontend/workflowHandler.go
```diff
@@ -2182,7 +2182,7 @@ func (wh *WorkflowHandler) StartWorkflowExecution(
 
 	wh.GetLogger().Debug("Start workflow execution request domainID", tag.WorkflowDomainID(domainID))
 	historyRequest, err := common.CreateHistoryStartWorkflowRequest(
-		domainID, startRequest, time.Now())
+		domainID, startRequest, time.Now(), nil)
```
davidporter-id-au, line 2185: are you intending to add this later? 

I assume we're expecting this field to be optional and going to treat it as such?

service/history/historyEngine3_eventsv2_test.go
```diff
@@ -252,6 +256,7 @@ func (s *engine3Suite) TestStartWorkflowExecution_BrandNew() {
 			Identity:                            identity,
 			RequestID:                           requestID,
 		},
+		PartitionConfig: partitionConfig,
```
davidporter-id-au, line 259: I would not fill this in immediately unless we're sure it's going to be present in all cases.

davidporter-id-au (approved):

service/frontend/workflowHandler.go
Shaddoll: it will be added later

service/history/historyEngine3_eventsv2_test.go
Shaddoll: the change is backward compatible, so previous tests covered the case when it's nil

PR #5157, opened 2023-03-14
Feature/zonal partitioning admin mapping

davidporter-id-au:
- Adds internal partitioning types and constants to the types folder for IDL mapping
- Adds sketch outline of admin API mappers

### DIFF
--- a/common/types/admin.go
+++ b/common/types/admin.go
@@ -404,13 +404,16 @@ type UpdateGlobalIsolationGroupsRequest struct {
 
 type UpdateGlobalIsolationGroupsResponse struct{}
 
-type GetDomainIsolationGroupsRequest struct{}
+type GetDomainIsolationGroupsRequest struct {
+	Domain string
+}
 
 type GetDomainIsolationGroupsResponse struct {
 	IsolationGroups IsolationGroupConfiguration
 }
 
 type UpdateDomainIsolationGroupsRequest struct {
+	Domain          string
 	IsolationGroups IsolationGroupConfiguration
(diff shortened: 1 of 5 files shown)


common/types/admin.go
```diff
@@ -364,3 +364,54 @@ func (v *ListDynamicConfigRequest) SerializeForLogging() (string, error) {
...
+//
+// For example: This might be a global configuration persisted
+// in the config store and look like this:
+//
+//	IsolationGroupConfiguration{
+//	  "isolationGroup1234": {Name: "isolationGroup1234", Status: IsolationGroupStatusDrained},
+//	}
+//
+// Indicating that task processing isn't to occur within this isolationGroup anymore, but all others are ok.
+type IsolationGroupConfiguration map[string]IsolationGroupPartition
```
Shaddoll, line 393: it's quite different from idl definition

```
struct IsolationGroupConfiguration {
  10: optional list<IsolationGroupPartition> isolationGroups
}
```

davidporter-id-au: yes, this was intentional: Most of the checking of values in the isolation groups are set membership operations (is this zone drained) and representing them as a list would have meant iterating over them multiple times. Not a huge cost, but also not really worthwhile. 

However, representing them as a map by name introduces a redundant name, which I didn't want to expose in IDL, thus I was going to transform it here in the mapping, with the internal implementation better suited for membership checks, and the public IDL less confusing to understand

Shaddoll (approved):

PR #5158, opened 2023-03-14
ES: single interface for different ES/OpenSearch versions

mantas-sidlauskas:
**What changed?**
Adding interface for versioned ES/OS clients to implement

**Why?**
Leaking client structs up makes it impossible to implement business logic in a single place. Adding new ES or OpenSearch client is also very difficult. 

This interface uses `[][]byte` to pass raw data from the server, and client can marshal this data to appropriate struct type.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
🙏

PR #5160, opened 2023-03-15
Set poll interval for filebased dynamic config if not set

mantas-sidlauskas:
**What changed?**
When poll interval is less than default value, adjust the value instead of returning an error

**Why?**
When poll interval config entry is missing, cadence will use `NOOP` dynamic config. Instead of failing back to `NOOP` client, we are adjusting poll interval and have file based dynamic config client in use.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


mindaugasbarcauskas (approved):
Looks good.

PR #5162, opened 2023-03-15
Add thin ES clients

mantas-sidlauskas:
**What changed?**
* Adding `v6` and `v7` ES clients compatible with new interface
* Bulk processor moved to a separate package

**Why?**
Bulk processor was moved to avoid cyclic dependency between `elasticsearch` package and ES clients
New ES clients need to wired in a separate PR

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx:
Since there seems to be a substantial amount of new code in here:
what's the goal / purpose?  and/or where are these new requirements coming from, since no libraries have been upgraded?

(part of this is that I'm simply unfamiliar with this code, it might be obvious)

mantas-sidlauskas:
> Since there seems to be a substantial amount of new code in here: what's the goal / purpose? and/or where are these new requirements coming from, since no libraries have been upgraded?
> 
> (part of this is that I'm simply unfamiliar with this code, it might be obvious)

As this and https://github.com/uber/cadence/pull/5168 are related, let me add some more context here:

* `olivere v6/v7` libraries now are deprecated and author recommends to use "official" Elasticsearch clients. 
* There will be no `olivere v8`
* I'm adding support for Opensearch client

To reuse business logic while using different versions of Elasticsearch/OpenSearch, we have to keep "versioned" clients small and only doing client/server communication. 

https://github.com/uber/cadence/pull/5168 adds a possibility to build queries in identical way but without depending on external packages/structs. 

The code in this PR is not used right now, but needs to be wired in the next PR: `common/elasticsearch/client_{v6|v7}.go` will be replaced with a single `client.go` and will use code provided in this PR.

common/elasticsearch/client/v6/client.go
```diff
@@ -0,0 +1,205 @@
...
+
+	var byteResult [][]byte
+	if esResult != nil && esResult.Hits != nil {
+		for _, hit := range esResult.Hits.Hits {
+			byteResult = append(byteResult, *hit.Source)
+		}
+	}
+
+	result := &client.Response{
+		TookInMillis: esResult.TookInMillis,
```
Shaddoll, line 149: panic if `esResult` is `nil`. We should do a check on `esResult` and `err` and return early.

mantas-sidlauskas: 👍 thanks!

common/elasticsearch/client/v6/client.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	if esResult == nil {
+		return nil, err
+	}
+
+	var hits []*client.SearchHit
+	if esResult.Hits != nil {
+		for _, h := range esResult.Hits.Hits {
+			hits = append(hits, &client.SearchHit{Source: *h.Source})
```
Shaddoll, line 124: it is not safe dereferencing a pointer without checking if it's nil

common/elasticsearch/client/v6/client.go
```diff
@@ -0,0 +1,186 @@
...
+		TookInMillis: esResult.TookInMillis,
+		TotalHits:    esResult.TotalHits(),
+		Hits:         &client.SearchHits{Hits: hits},
+		ScrollID:     esResult.ScrollId,
+	}
+
+	if len(esResult.Aggregations) > 0 {
+		result.Aggregations = make(map[string]json.RawMessage, len(esResult.Aggregations))
+		for key, agg := range esResult.Aggregations {
+			result.Aggregations[key] = *agg
```
Shaddoll, line 138: it is not safe dereferencing a pointer without checking if it's nil

common/elasticsearch/client/v6/client.go
mantas-sidlauskas: done

common/elasticsearch/client/v6/client.go
mantas-sidlauskas: done

common/elasticsearch/client/v6/client.go
```diff
@@ -0,0 +1,190 @@
...
+			Message: fmt.Sprintf("ElasticSearch Error: Request timed out: %v ms", esResult.TookInMillis),
+		}
+	}
+
+	var sort []interface{}
+	var hits []*client.SearchHit
+
+	if esResult != nil && esResult.Hits != nil {
+		for _, h := range esResult.Hits.Hits {
+			hits = append(hits, &client.SearchHit{Source: *h.Source})
```
Shaddoll, line 170: the `nil` check is missing in `Search` func

common/elasticsearch/client/v6/client.go
```diff
@@ -0,0 +1,190 @@
...
+		TookInMillis: esResult.TookInMillis,
+		TotalHits:    esResult.TotalHits(),
+		Hits:         &client.SearchHits{Hits: hits},
+		Sort:         sort,
+	}
+
+	if len(esResult.Aggregations) > 0 {
+		result.Aggregations = make(map[string]json.RawMessage, len(esResult.Aggregations))
+		for key, agg := range esResult.Aggregations {
+			result.Aggregations[key] = *agg
```
Shaddoll, line 185: the `nil` check is missing in `Search` func

common/elasticsearch/client/v6/client.go
mantas-sidlauskas: added

common/elasticsearch/client/v6/client.go
mantas-sidlauskas: nil check is added

Shaddoll (approved):

PR #5163, opened 2023-03-15
Add Pinot docker files, table config and schema

neil-xie:
**What changed?**
Initial check in to add support for Pinot. 
Added pinot docker files, table config and schema files. 
Added instruction to create Pinot table.

**Why?**
ES will be deprecated internally at Uber, start supporting Pinot as an option. Existing ES features won't be affected. 

**How did you test it?**
Config changes, tested locally to create Pinot table.

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
I assume you mean to say that ES is looking to be deprecated internally to Uber's use-case, not generally @neil-xie ?

neil-xie:
> I assume you mean to say that ES is looking to be deprecated internally to Uber's use-case, not generally @neil-xie ?

@davidporter-id-au Oh yeah, my bad. It will only be deprecated internally and no new feature development for ES. Won't affect the existing feature.

PR #5164, opened 2023-03-16
[history] more cautious in deciding domain state to make decisions on…

shijiesheng:
… dropping queued tasks

**What changed?**

When domain cache returned entity not found error, don't drop queued tasks to be more conservative. 

**Why?**

In cases when the cache is dubious, we shouldn't drop the queued tasks.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/queue/task_allocator.go
+++ b/service/history/queue/task_allocator.go
@@ -196,10 +196,7 @@ func (t *taskAllocatorImpl) Unlock() {
 func isDomainNotRegistered(shard shard.Context, domainID string) (bool, error) {
 	domainEntry, err := shard.GetDomainCache().GetDomainByID(domainID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


davidporter-id-au (approved):
this might be unit-test worthy

Groxx:
Given the contents of #5140, I think the error(s) in this func won't be logged... tbh we probably do want to log and/or metric issues here, since they'd be unexpected.  Currently they're created and lost forever.

But yea, 👍 for this change.  Accidentally-merged clusters could also be dangerous otherwise, with not-founds.

Groxx (approved):

shijiesheng:
> this might be unit-test worthy

I'll add unit test (along with the log, metrics) in the follow-up

PR #5166, opened 2023-03-17
large workflow hot shard detection

allenchen2244:
**What changed?**
Added metrics to check if shardId has large blobs/historySize/historyCount

**Why?**
Identify shards that is under heavy load.

**How did you test it?**
Auto tests pass. Fully tested everything end to end in staging using cadence-killers to test different pathological scenarios 

**Potential risks**
Just metrics and logs. Can't think of anything

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1346,7 +1346,21 @@ const (
 	// Value type: Int
 	// Default value: 100
 	SampleLoggingRate
-
+	// LargeShardHistorySizeMetricThreshold defines the threshold for what consititutes a large history storage size to alert on
+	// KeyName: system.largeShardHistorySizeMetricThreshold
+	// Value type: Int
+	// Default value: 10485760
+	LargeShardHistorySizeMetricThreshold
+	// LargeShardHistoryEventMetricThreshold defines the threshold for what consititutes a large history event size to alert on
+	// KeyName: system.largeShardHistoryEventMetricThreshold
+	// Value type: Int
+	// Default value: 10000
+	LargeShardHistoryEventMetricThreshold
+	// LargeShardHistoryBlobMetricThreshold defines the threshold for what consititutes a large history blob size to alert on
+	// KeyName: system.largeShardHistoryBlobMetricThreshold
+	// Value type: Int
+	// Default value: 262144
+	LargeShardHistoryBlobMetricThreshold
 	// LastIntKey must be the last one in this const group
 	LastIntKey
 )
@@ -3453,6 +3467,21 @@ var IntKeys = map[IntKey]DynamicInt{
 		KeyName:      "system.sampleLoggingRate",
 		Description:  "The rate for which sampled logs are logged at. 100 means 1/100 is logged",
 		DefaultValue: 100,
+	},
+	LargeShardHistorySizeMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistorySizeMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history size to alert on, default is 10mb",
+		DefaultValue: 10485760,
+	},
+	LargeShardHistoryEventMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 10000",
+		DefaultValue: 10000,
+	},
+	LargeShardHistoryBlobMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history blob write to alert on, default is 1/4mb",
+		DefaultValue: 262144,
 	},
 }
 
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -3479,7 +3479,7 @@ var IntKeys = map[IntKey]DynamicInt{
 		DefaultValue: 10000,
 	},
 	LargeShardHistoryBlobMetricThreshold: DynamicInt{
-		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		KeyName:      "system.largeShardHistoryBlobMetricThreshold",
 		Description:  "defines the threshold for what consititutes a large history blob write to alert on, default is 1/4mb",
 		DefaultValue: 262144,
 	},
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1349,12 +1349,12 @@ const (
 	// LargeShardHistorySizeMetricThreshold defines the threshold for what consititutes a large history storage size to alert on
 	// KeyName: system.largeShardHistorySizeMetricThreshold
 	// Value type: Int
-	// Default value: 10485760
+	// Default value: 10485760 (10mb)
 	LargeShardHistorySizeMetricThreshold
 	// LargeShardHistoryEventMetricThreshold defines the threshold for what consititutes a large history event size to alert on
 	// KeyName: system.largeShardHistoryEventMetricThreshold
 	// Value type: Int
-	// Default value: 10000
+	// Default value: 50 * 1024
 	LargeShardHistoryEventMetricThreshold
 	// LargeShardHistoryBlobMetricThreshold defines the threshold for what consititutes a large history blob size to alert on
 	// KeyName: system.largeShardHistoryBlobMetricThreshold
@@ -3476,7 +3476,7 @@ var IntKeys = map[IntKey]DynamicInt{
 	LargeShardHistoryEventMetricThreshold: DynamicInt{
 		KeyName:      "system.largeShardHistoryEventMetricThreshold",
 		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 10000",
-		DefaultValue: 10000,
+		DefaultValue: 50 * 1024,
 	},
 	LargeShardHistoryBlobMetricThreshold: DynamicInt{
 		KeyName:      "system.largeShardHistoryBlobMetricThreshold",
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -3475,7 +3475,7 @@ var IntKeys = map[IntKey]DynamicInt{
 	},
 	LargeShardHistoryEventMetricThreshold: DynamicInt{
 		KeyName:      "system.largeShardHistoryEventMetricThreshold",
-		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 10000",
+		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 50k",
 		DefaultValue: 50 * 1024,
 	},
 	LargeShardHistoryBlobMetricThreshold: DynamicInt{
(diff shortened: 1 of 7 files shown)


common/dynamicconfig/constants.go
```diff
@@ -3454,6 +3468,21 @@ var IntKeys = map[IntKey]DynamicInt{
...
+	LargeShardHistorySizeMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistorySizeMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history size to alert on, default is 10mb",
+		DefaultValue: 10485760,
+	},
+	LargeShardHistoryEventMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 10000",
+		DefaultValue: 10000,
+	},
```
Groxx, line 3480: is this a per-workflow-history-count threshold?  would it make sense to use the same default value as the warning level?

common/dynamicconfig/constants.go
```diff
@@ -1346,7 +1346,21 @@ const (
 	// Value type: Int
 	// Default value: 100
 	SampleLoggingRate
-
+	// LargeShardHistorySizeMetricThreshold defines the threshold for what consititutes a large history storage size to alert on
+	// KeyName: system.largeShardHistorySizeMetricThreshold
+	// Value type: Int
+	// Default value: 10485760
```
Groxx, line 1352: could probably use the "10mb" comment here too

service/history/execution/context_util.go
```diff
@@ -21,15 +21,44 @@
...
+
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/types"
 )
 
+func emitLargeWorkflowShardIDStats(logger log.Logger, metricsClient metrics.Client, config *config.Config, shardID int,
+	domainName string, workflowID string, currentHistorySize int64, currentHistoryCount int64, oldHistorySize int64, oldHistoryCount int64, blobSize int64) {
```
Groxx, line 37: too many same-typed args for safety tbh.  make a temp struct, so it's essentially not possible to screw up the arg ordering?  it might also help the calling code be clearer.

also, since this is only used by the context and it has 5 args that come directly from the context: it'd probably be a lot simpler to just make this a method, so it can get those values from itself rather than as a bunch of arguments.  shouldn't change tests much either, since you can construct a minimal `contextImpl` pretty easily.

common/dynamicconfig/constants.go
allenchen2244: I think because we've been telling people 10k in cadence users but default is 25k which seems to be from over 2 years ago. We can always change in flipr as needed.

Groxx: I mean more that this is dependent on the datastore / other config, no one value is "correct".  and our internal values are largely irrelevant.

10k seems probably reasonable, and I'm not sure if we can actually make this a "the default matches the dynamic value of [another key]" without a lot more work.  just wondering if it'd make more sense to be hardcoded as the same as another, to show the implied relationship.

allenchen2244: How about I set both to the same hard coded value initially but take the lesser of the 2 when it comes to metric emission?

allenchen2244: yeah i think this is a good idea since those values are actually configurable on a domain level so doing a emitMetricWarning = generalWarn<domainWarn?generalWarn:domainWarn seems good.

common/dynamicconfig/constants.go
```diff
@@ -3454,6 +3468,21 @@ var IntKeys = map[IntKey]DynamicInt{
...
 	},
+	LargeShardHistorySizeMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistorySizeMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history size to alert on, default is 10mb",
+		DefaultValue: 10485760,
+	},
+	LargeShardHistoryEventMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 10000",
+		DefaultValue: 50 * 1024,
```
Groxx, line 3479:
```suggestion
		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 50k",
		DefaultValue: 50 * 1024,
```

service/history/execution/context.go
```diff
@@ -852,6 +858,10 @@ func (c *contextImpl) UpdateWorkflowExecutionWithNew(
 		domainName,
 		resp.MutableStateUpdateSessionStats,
 	)
+	c.stats.BlobSize = currentWorkflowSize - oldWorkflowSize
+	c.stats.OldHistoryCount = oldWorkflowHistoryCount
+	c.stats.OldHistorySize = oldWorkflowSize
```
Groxx, line 863: huh.  `Set/GetHistorySize` just forwards to c.stats like this.
tbh I think I like the bare field access better than the func, unless something external needs it (e.g. as an interface).  so 👍 maybe it's even worth removing those wrappers around HistorySize.

service/history/execution/context_util.go
```diff
@@ -30,6 +31,52 @@ import (
...
+func (c *contextImpl) emitLargeWorkflowShardIDStats() {
+	if c.shard.GetConfig().EnableShardIDMetrics() {
+		shardIDStr := strconv.Itoa(c.shard.GetShardID())
+
+		var blobSizeWarn int64
+		if c.shard.GetConfig().LargeShardHistoryBlobMetricThreshold() < c.shard.GetConfig().BlobSizeLimitWarn(c.GetDomainName()) {
+			blobSizeWarn = int64(c.shard.GetConfig().LargeShardHistoryBlobMetricThreshold())
+		} else {
+			blobSizeWarn = int64(c.shard.GetConfig().BlobSizeLimitWarn(c.GetDomainName()))
+		}
```
Groxx, line 43: just do it once and get the min of both values.  dynamic config can have arbitrary behavior under it, it's not (necessarily) free.

common/dynamicconfig/constants.go
Groxx: yea, min of both in code makes sense I think.  seems fairly likely that user-warning-level also means approaching "big".

worst case, we just change our minds later 🤷

service/history/execution/context_util.go
Groxx: extending the history-stats struct like you've done makes sense at a glance.  checking more thoroughly, but I think I like it - seems like it's all for the same purpose.

Groxx: neeeever mind - that's a persistence object.  that deserves more scrutiny.
particularly with the "old" values - passing those as an argument might be better, rather than saving them to the database.

allenchen2244: ohh those get saved? I'm guessing somewhere along the way it saves the new context? I'll just pass them then. Managed to cut it down to only 3 fields now though rest are just "c.blah"

Groxx: yea, I think all these need to be separate.  current count is stored in the database and this func relies on it to raise the count by the num of new events, and that ends up being written back to the database.

all the new metrics don't need that.  they're inferred from other recorded values and are just being held temporarily in memory to emit monitoring.  no need to duplicate them in the DB / risk them getting out of sync.

service/history/execution/context_util.go
```diff
@@ -30,6 +31,58 @@ import (
...
+		shardIDStr := strconv.Itoa(c.shard.GetShardID())
+
+		var blobSizeWarn int64
+		blobSizeGlobalWarn := c.shard.GetConfig().LargeShardHistoryBlobMetricThreshold()
+		blobSizeDomainWarn := c.shard.GetConfig().BlobSizeLimitWarn(c.GetDomainName())
+		if blobSizeGlobalWarn < blobSizeDomainWarn {
+			blobSizeWarn = int64(blobSizeGlobalWarn)
+		} else {
+			blobSizeWarn = int64(blobSizeDomainWarn)
+		}
```
Groxx, line 45: tbh it's probably easier to build a small
```go
func min64(a,b int) int64 {
  if a < b { return int64(a) }
  return int64(b)
}
```
and use it for all three.  it'd get rid of a bunch of not-intended-for-reuse temp vars too.

service/history/execution/context.go
```diff
@@ -720,11 +719,12 @@ func (c *contextImpl) UpdateWorkflowExecutionWithNew(
 	if err != nil {
 		return err
 	}
-
 	var persistedBlobs events.PersistedBlobs
-	currentWorkflowSize := c.GetHistorySize()
+	currentWorkflowSize, oldWorkflowSize := c.GetHistorySize(), c.GetHistorySize()
+	currentWorkflowHistoryCount, oldWorkflowHistoryCount := c.mutableState.GetNextEventID()-1, c.mutableState.GetNextEventID()-1
```
Groxx, line 724:
```suggestion
	currentWorkflowSize := c.GetHistorySize()
	oldWorkflowSize := currentWorkflowSize
	currentWorkflowHistoryCount := c.mutableState.GetNextEventID()-1
	oldWorkflowHistoryCount := currentWorkflowHistoryCount
```
since that's the intent, and it makes it impossible for it to mean anything different (e.g. say GetNextEventID() mutated something).

service/history/execution/context_util.go
allenchen2244: might have somethiing already. Let me poke around

Groxx: yea, it might require coercing to ints or int64s or something in a few places, but we probably do have a generic-min somewhere.  though I think they're simple enough that it's fine to simply rewrite it.

service/history/execution/context_util.go
Groxx: yeah, after turning it into a method I think it's small enough that it's less risky to have those args.  lots just have a bad track record of leading to accidentally-incorrect ordering.

service/history/execution/context_util.go
allenchen2244: found the min64. pretty clean now i think

service/history/execution/context.go
```diff
@@ -852,6 +854,7 @@ func (c *contextImpl) UpdateWorkflowExecutionWithNew(
 		domainName,
 		resp.MutableStateUpdateSessionStats,
 	)
+	c.emitLargeWorkflowShardIDStats(currentWorkflowSize-oldWorkflowSize, oldWorkflowHistoryCount, oldWorkflowSize)
```
Groxx, line 857: while I'm exploring the code in here:
would any of these be more-correct if they came from `newWorkflow`?  and/or would using `newWorkflow` mean we wouldn't need to maintain as much separately?

I'm not quite sure what the difference is tbh.  I'd have to read more carefully.  I don't think it'll be _dangerous_ to use the wrong one, just possibly misleading (due to subtly-wrong values, or due to unnecessary duplicate calculations that could drift)

allenchen2244: I don't think it matters. From what I can tell new workflow kind of derives the numbers the same way I am. Maybe it would be less duplicated code but at most it's saving 5 lines and makes it a bit more confusing to read imo.

Groxx: the main reason to prefer using the new-workflow's data is that it'd make it clear what the source of truth is, and get rid of any risk of the calculations drifting away from that source of truth.

which is an "if".  tbh I'm not sure if it's more-correct or not.

allenchen2244: I could try testing it later? The other stuff is kinda calculated the same way so i wanted to make it consistent. I don't wanna just change part of it then it's super confusing if we never change the rest.

Groxx: which ones are those?
but yea, later's fine (if ever)

PR #5169, opened 2023-03-20
Adds partitioning library

davidporter-id-au:
- Adds an interface for handling workflow isolation group redirection and failover
- Adds an interface for handling drain states for both global and domain-based drains
- Adds relevant internal types
- Adds a simple interface for handling global drains as stored in database-based dynamic config
- Adds a default implementation for the open-source implementation and implementation example


common/partition/partitioning.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+// Config is the base configuration for the partitioning library
+type Config struct {
+	zonalPartitioningEnabledGlobally  dynamicconfig.BoolPropertyFnWithDomainIDFilter
```
Shaddoll, line 35: Are the fields supposed to be exported? 
BTW, I don't see the value of having this struct.

common/partition/partitioning.go
Shaddoll, line 1: Do you mind renaming the filename to `interface.go`

common/partition/partitioning.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type Partitioner interface {
+	// GetIsolationGroupByDomainID gets where the task workflow should be executing. Largely used by Matching
+	// when determining which isolationGroup to place the tasks in.
+	// Implementations ought to return (nil, nil) for when the feature is not enabled.
+	GetIsolationGroupByDomainID(ctx context.Context, DomainID string, partitionKey PartitionConfig) (*string, error)
+	// IsDrained answers the question - "is this particular isolationGroup drained?". Used by startWorkflow calls
+	// and similar sync frontend calls to make routing decisions
+	IsDrained(ctx context.Context, Domain string, IsolationGroup string) (bool, error)
```
Shaddoll, line 47: It seems that this method is more suitable to be put inside `isolationgroup.State` interface

common/partition/partitioning.go
davidporter-id-au: whoops, thank you. yeah, let me fix that. I was going to export them.

common/partition/partitioning.go
davidporter-id-au: Sure

common/partition/default-partitioner.go
```diff
@@ -0,0 +1,183 @@
...
+// drain.
+func pickIsolationGroup(wfPartition defaultWorkflowPartitionConfig, available types.IsolationGroupConfiguration) string {
+	wfIG, isAvailable := available[wfPartition.WorkflowStartIsolationGroup]
+	if isAvailable && wfIG.State != types.IsolationGroupStateDrained {
+		return wfPartition.WorkflowStartIsolationGroup
+	}
+
+	// it's drained, fall back to picking a deterministic but random group
+	availableList := []string{}
+	for k, v := range available {
```
Shaddoll, line 165: `available` seems to only contain healthy isolation groups.

common/partition/default-partitioner.go
```diff
@@ -0,0 +1,183 @@
...
+		if v.State == types.IsolationGroupStateDrained {
```
Shaddoll, line 166: I suggest changing the condition to be `v.State != types.IsolationGroupStateHealthy`

common/partition/default-partitioner.go
```diff
@@ -0,0 +1,183 @@
...
+		WorkflowStartIsolationGroup: isolationGroup,
+		RunID:                       runID,
+	}
+}
+
+// picks an isolation group to run in. if the workflow was started there, it'll attempt to pin it, unless there is an explicit
+// drain.
+func pickIsolationGroup(wfPartition defaultWorkflowPartitionConfig, available types.IsolationGroupConfiguration) string {
+	wfIG, isAvailable := available[wfPartition.WorkflowStartIsolationGroup]
+	if isAvailable && wfIG.State != types.IsolationGroupStateDrained {
```
Shaddoll, line 159: I suggest changing the condition to be `wfIG.State == types.IsolationGroupStateHealthy`

common/isolationgroup/types.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// IsolationGroups is a convenience return type of a collection of IsolationGroup configurations
+type IsolationGroups struct {
+	Global types.IsolationGroupConfiguration
+	Domain types.IsolationGroupConfiguration
+}
+
+type ChangeEvent struct {
+	changed types.IsolationGroupConfiguration
```
Shaddoll, line 37: The field needs to be exported.

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -0,0 +1,161 @@
...
+	if !z.config.zonalPartitioningEnabledForDomain(domain) {
+		return nil, nil
+	}
+
+	domainData, err := z.domainCache.GetDomain(domain)
+	if err != nil {
+		return nil, fmt.Errorf("could not resolve domain in isolationGroup handler: %w", err)
+	}
+	domainState := domainData.GetInfo().IsolationGroupConfig
+	globalCfg, err := z.globalIsolationGroupDrains.FetchConfig(ctx, persistence.GlobalIsolationGroupConfig)
```
Shaddoll, line 93: I think we need some cache for `configStore`

common/partition/partitioning.go
davidporter-id-au: hm.. not sure... I think on the face of it that split makes more sense, however it does spread the responsibility a bit. Let me add it there now and we can see.

common/partition/default-partitioner.go
davidporter-id-au: This is a deny-based implementation. Isolation groups are explicitly assumed to be healthy unless specified as drained, so I can't rely on their being configured here.

common/partition/default-partitioner.go
davidporter-id-au: That is the intention, yes? I'm not quite sure I understand your comment.?

common/partition/default-partitioner.go
davidporter-id-au: This a deny-based implementation. we're only skipping zones which have been drained. Let me sync to make sure we're on the same page here.

common/isolationgroup/types.go
davidporter-id-au: thank you 🤦

common/isolationgroup/default-isolation-group-state.go
davidporter-id-au: Yes, but if it's ok let me follow it up. I'd like to do that in a subsequent PR

common/partition/interface.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+// Config is the base configuration for the partitioning library
+type Config struct {
+	ZonalPartitioningEnabledGlobally  dynamicconfig.BoolPropertyFnWithDomainIDFilter
```
Shaddoll, line 35: Why do we need both `ZonalPartitioningEnabledGlobally` and `ZonalPartitioningEnabledForDomain`? 
And we might need a better name, like `EnableTasklistIsolation` or something.

common/isolationgroup/types.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type ChangeEvent struct {
+	Changed types.IsolationGroupConfiguration
+}
+
+// Config is the base configuration
+type Config struct {
+	ZonalPartitioningEnabledGlobally  dynamicconfig.BoolPropertyFnWithDomainIDFilter
+	ZonalPartitioningEnabledForDomain dynamicconfig.BoolPropertyFnWithDomainFilter
```
Shaddoll, line 37: Why do we need both dynamic config?

common/partition/default-partitioner.go
Shaddoll: Because the state is not a boolean, we might introduce other state that is not healthy.

common/partition/interface.go
davidporter-id-au: yeah, that name is out of date, let me fix, sorry

common/isolationgroup/types.go
davidporter-id-au: How would we represent rollout on both a global and domain level? I'm totally open to the idea that maybe we don't care about this level of granularity, but I thought for rollout it's sufficiently high-risk that enabling it on a per-domain-basis might make sense.

common/isolationgroup/interface.go
```diff
@@ -0,0 +1,81 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination isolation_group_mock.go -self_package github.com/uber/cadence/common/isolationgroup
+
+// Controller is for handling the CRUD operations of isolation groups
+type Controller interface {
```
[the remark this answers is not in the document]
davidporter-id-au: ok, I'll remove for now

common/partition/default-partitioner.go
davidporter-id-au: That's possible, but I'm maybe not understanding your point. Actually explicitly healthy groups will be not included in this list at all under this implementation, so picking healthy explicitly is not an option.

common/isolationgroup/interface.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+
+	"github.com/uber/cadence/common"
+
+	"github.com/uber/cadence/common/types"
+)
+
+// The MIT License (MIT)
```
Shaddoll, line 33: License is duplicated.

common/partition/interface.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+// Config is the base configuration for the partitioning library
+type Config struct {
+	IsolationGroupEnabledGlobally  dynamicconfig.BoolPropertyFnWithDomainIDFilter
```
Shaddoll, line 35: This is not removed

Shaddoll (approved):

PR #5171, opened 2023-03-26
Adds a small test to catch issues with deadlocks

davidporter-id-au:
It's quite easy to call the getters on tasks in this function and cause a deadlock. Ensuring complete coverage on this branch to prevent this.

### DIFF
--- a/service/history/task/task_test.go
+++ b/service/history/task/task_test.go
@@ -22,6 +22,7 @@ package task
 
 import (
 	"errors"
+	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


service/history/task/task_test.go
```diff
@@ -266,6 +267,18 @@ func (s *taskSuite) TestTaskNack_ResubmitFailed() {
...
+		return true, nil
+	}, nil)
+
+	taskBase.criticalRetryCount = func(i ...dynamicconfig.FilterOption) int { return 0 }
+	s.mockTaskInfo.EXPECT().GetTaskType().Return(0)
+	assert.NotPanics(s.T(), func() {
+		taskBase.HandleErr(errors.New("err"))
+	})
+}
+
```
Groxx, line 281: Since a deadlock won't panic, I've done this in the past:
```go
done := make(chan struct{})
go func() {
  defer close(done)
  deadlocky()
}()
select {
case <-done:
case <-time.After(time.Second): t.Error("timed out == deadlock")
}
```
fast enough to be practical, without waiting for a full minute or whatever for the normal whole-package test timeout.

davidporter-id-au: while I appreciate that makes for a better performing test, it... feels slightly overspecialised with the benefit of hindsight to look exclusively for deadlocks, rather than just checking the path entirely (though, I suppose it would catch a panic all the same). 

It looks like there'll be a new `testing.T` method `SetTimeout(d time.Duration)` coming soon which should do exactly this and - imho, is a lot less noisy (https://github.com/golang/go/issues/48157)

I don't feel super strongly about it, but I'm wary about one-off testing infra setups such as this in that their complexity might dissuade people to maintain / think about them

Groxx (approved):
useful test either way, comments are just comments and not in any way meant as a requirement :)

PR #5173, opened 2023-03-28
Migrate matching to use tasksV2

Shaddoll:
**What changed?**
This is a draft PR to update matching to migrate from task v1 to task v2 in order to support isolation at persistence layer for async match.

task v2 has such schema in Cassandra:
```
CREATE TABLE tasks_v2 (
  domain_id        uuid,
  task_list_name   text,
  task_list_type   int, -- enum TaskListType {ActivityTask, DecisionTask}
+ isolation_group  text, -- always empty for tasklist
  type             int, -- enum rowType {Task, TaskList}
  task_id          bigint,  -- unique identifier for tasks, monotonically increasing
  range_id         bigint, -- Used to ensure that only one process can write to the table
  task             frozen<task>,
  task_list        frozen<task_list>,
  PRIMARY KEY ((domain_id, task_list_name, task_list_type), isolation_group, type, task_id)
) WITH COMPACTION = {
    'class': 'org.apache.cassandra.db.compaction.LeveledCompactionStrategy'
  };
```
task_list type is also updated
```
CREATE TYPE task_list (
  domain_id        uuid,
  name             text,
  type             int, -- enum TaskRowType {ActivityTask, DecisionTask}
  ack_level        bigint, -- task_id of the last acknowledged message
+ ack_levels       map<text, bigint>,
  kind             int, -- enum TaskListKind {Normal, Sticky}
  last_updated     timestamp
);
```

We add a new primary column `isolation_group` to the original table. We also separate the ack levels for each isolation group so that an isolation group will not prevent another isolation group from cleaning tasks from database.

We also introduce a new dynamic config `EnableTaskV2Migration` to control the migration at tasklist level. After this code is merged, matching will read tasks from both task v1 and v2 tables, and it will write to task v1 if the feature flag is disabled and to task v2 if it is enabled. Matching will consume more memory because we need more goroutines reading tasks from both table, but we also gain the ability to migrate task table in both directionn without reloading the tasklist from matching. 
Another thing is that v2 uses more memory than v1, because task reader has multiple goroutines for each isolation group, even if the tasklist does not need isolation, but we also gain the ability to disable or enable isolation without reloading tasklist from matching.

**Why?**

**How did you test it?**
Test is TBD.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/dataManagerInterfaces.go
+++ b/common/persistence/dataManagerInterfaces.go
@@ -446,6 +446,7 @@ type (
 		TaskType    int
 		RangeID     int64
 		AckLevel    int64
+		AckLevels   map[string]int64
 		Kind        int
 		Expiry      time.Time
 		LastUpdated time.Time
@@ -1276,9 +1277,10 @@ type (
 
 	// CreateTaskInfo describes a task to be created in CreateTasksRequest
 	CreateTaskInfo struct {
-		Execution types.WorkflowExecution
-		Data      *TaskInfo
-		TaskID    int64
+		Execution      types.WorkflowExecution
+		Data           *TaskInfo
+		IsolationGroup string
+		TaskID         int64
 	}
 
 	// CreateTasksResponse is the response to CreateTasksRequest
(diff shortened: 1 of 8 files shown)


taylanisikdemir:

PR #5178, opened 2023-03-29
Wires IsolationGroup Daemon and Partitioner into service start

davidporter-id-au:
Regrettably a somewhat-large PR which fundamentally is wiring the `isolationGroupState` and `partition.Partitioner` default implementations into the service `resource` runtime, and in doing so adding some necessary refactoring: 

1. Wires in the IsolationState library into the service:
  - Adds DynamicConfig keys
  - Adds a small amount of testing around the setup and configuration of the IsolationGroupState library

2. Updates DynamicConfigManager to be compatible with library's requirements:
  - Refactors UpdateDynamicConfig / FetchDynamicConfig of the `ConfigStoreManager` to take an input of the configuration type parameter
  - Updates the IsolationGroup to use the ConfigStoreManager instead of ConfigStore
  - Adds IsolationGroup, Partitioner to the service `ResourceImpl` params in preparation for where they'll be instantiated

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -75,6 +75,10 @@ type (
 		Authorization Authorization `yaml:"authorization"`
 		// HeaderForwardingRules defines which inbound headers to include or exclude on outbound calls
 		HeaderForwardingRules []HeaderRule `yaml:"headerForwardingRules"`
+		// IsolationGroups is for configuration the isolationGroupState handler
+		IsolationGroups IsolationGroups `yaml:"-" json:"-"`
+		// Partitioner configuration for subdivision of traffic
+		Partitioner Partitioner `yaml:"-" json:"-"`
 	}
 
 	HeaderRule struct {
@@ -514,6 +518,23 @@ type (
 		Status string `yaml:"status"`
 		// URI is the domain default URI for visibility archiver
 		URI string `yaml:"URI"`
+	}
+
+	// Partitioner is the configuration for the partitioning library, a library
+	// for segmenting portions of workflows into isolation-groups - a resiliency
+	// concept meant to help move workflows around and away from failure zones.
+	Partitioner struct {
+		IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
+		AllIsolationGroups    []string
+	}
+
+	// IsolationGroups refers to the configuration for subdivisions of workflows
+	// into a distinct and geographic grouping in which failure expected to be
+	// contained and able to be drained
+	IsolationGroups struct {
+		IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
+		UpdateFrequency       dynamicconfig.DurationPropertyFn
+		AllIsolationGroups    []string
 	}
 )
 
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -75,10 +75,6 @@ type (
 		Authorization Authorization `yaml:"authorization"`
 		// HeaderForwardingRules defines which inbound headers to include or exclude on outbound calls
 		HeaderForwardingRules []HeaderRule `yaml:"headerForwardingRules"`
-		// IsolationGroups is for configuration the isolationGroupState handler
-		IsolationGroups IsolationGroups `yaml:"-" json:"-"`
-		// Partitioner configuration for subdivision of traffic
-		Partitioner Partitioner `yaml:"-" json:"-"`
 	}
 
 	HeaderRule struct {
@@ -518,23 +514,6 @@ type (
 		Status string `yaml:"status"`
 		// URI is the domain default URI for visibility archiver
 		URI string `yaml:"URI"`
-	}
-
-	// Partitioner is the configuration for the partitioning library, a library
-	// for segmenting portions of workflows into isolation-groups - a resiliency
-	// concept meant to help move workflows around and away from failure zones.
-	Partitioner struct {
-		IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
-		AllIsolationGroups    []string
-	}
-
-	// IsolationGroups refers to the configuration for subdivisions of workflows
-	// into a distinct and geographic grouping in which failure expected to be
-	// contained and able to be drained
-	IsolationGroups struct {
-		IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
-		UpdateFrequency       dynamicconfig.DurationPropertyFn
-		AllIsolationGroups    []string
 	}
 )
 
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1346,7 +1346,21 @@ const (
 	// Value type: Int
 	// Default value: 100
 	SampleLoggingRate
-
+	// LargeShardHistorySizeMetricThreshold defines the threshold for what consititutes a large history storage size to alert on
+	// KeyName: system.largeShardHistorySizeMetricThreshold
+	// Value type: Int
+	// Default value: 10485760 (10mb)
+	LargeShardHistorySizeMetricThreshold
+	// LargeShardHistoryEventMetricThreshold defines the threshold for what consititutes a large history event size to alert on
+	// KeyName: system.largeShardHistoryEventMetricThreshold
+	// Value type: Int
+	// Default value: 50 * 1024
+	LargeShardHistoryEventMetricThreshold
+	// LargeShardHistoryBlobMetricThreshold defines the threshold for what consititutes a large history blob size to alert on
+	// KeyName: system.largeShardHistoryBlobMetricThreshold
+	// Value type: Int
+	// Default value: 262144 (1/4mb)
+	LargeShardHistoryBlobMetricThreshold
 	// LastIntKey must be the last one in this const group
 	LastIntKey
 )
@@ -3474,6 +3488,21 @@ var IntKeys = map[IntKey]DynamicInt{
 		KeyName:      "system.sampleLoggingRate",
 		Description:  "The rate for which sampled logs are logged at. 100 means 1/100 is logged",
 		DefaultValue: 100,
+	},
+	LargeShardHistorySizeMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistorySizeMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history size to alert on, default is 10mb",
+		DefaultValue: 10485760,
+	},
+	LargeShardHistoryEventMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryEventMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history event length to alert on, default is 50k",
+		DefaultValue: 50 * 1024,
+	},
+	LargeShardHistoryBlobMetricThreshold: DynamicInt{
+		KeyName:      "system.largeShardHistoryBlobMetricThreshold",
+		Description:  "defines the threshold for what consititutes a large history blob write to alert on, default is 1/4mb",
+		DefaultValue: 262144,
 	},
 }
 
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1818,13 +1818,13 @@ const (
 
 	EnableCassandraAllConsistencyLevelDelete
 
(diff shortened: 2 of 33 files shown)


common/persistence/persistenceErrorInjectionClients.go
```diff
@@ -2354,13 +2354,13 @@ func (p *configStoreErrorInjectionPersistenceClient) FetchDynamicConfig(ctx cont
...
 
-func (p *configStoreErrorInjectionPersistenceClient) UpdateDynamicConfig(ctx context.Context, request *UpdateDynamicConfigRequest) error {
+func (p *configStoreErrorInjectionPersistenceClient) UpdateDynamicConfig(ctx context.Context, request *UpdateDynamicConfigRequest, cfgType ConfigType) error {
 	fakeErr := generateFakeError(p.errorRate)
 
 	var persistenceErr error
 	var forwardCall bool
 	if forwardCall = shouldForwardCallToPersistence(fakeErr); forwardCall {
-		persistenceErr = p.persistence.UpdateDynamicConfig(ctx, request)
+		persistenceErr = p.persistence.UpdateDynamicConfig(ctx, request, DynamicConfig)
```
Shaddoll, line 2363: should be `cfgType`

common/resource/params.go
```diff
@@ -66,5 +69,7 @@ type (
 		ArchiverProvider         provider.ArchiverProvider
 		Authorizer               authorization.Authorizer // NOTE: this can be nil. If nil, AccessControlledHandlerImpl will initiate one with config.Authorization
 		AuthorizationConfig      config.Authorization     // NOTE: empty(default) struct will get a authorization.NoopAuthorizer
+		isolationGroupState      isolationgroup.State
```
Shaddoll, line 72: field should be exported

common/resource/resourceImpl.go
```diff
@@ -297,7 +303,9 @@ func New(
 			logger,
 			params.InstanceID,
 		),
-		rpcFactory: params.RPCFactory,
+		rpcFactory:      params.RPCFactory,
+		isolationGroups: params.isolationGroupState,
```
Shaddoll, line 307: I think we need some methods to access these fields

common/resource/params.go
Shaddoll: Could you also make change in `cmd/server.go`? If we do that, such error can be detected early.

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -56,8 +56,7 @@ func NewDefaultIsolationGroupStateWatcher(
 	logger log.Logger,
 	domainCache cache.DomainCache,
 	config Config,
-	globalIsolationGroupDrains persistence.ConfigStore,
-	done chan bool,
```
Shaddoll, line 60: this param is supposed to be removed?

common/persistence/persistenceErrorInjectionClients.go
davidporter-id-au: ? DynamicConfig is a type of value ConfigType
Screenshot 2023-03-29 at 5 35 49 PM

common/resource/params.go
davidporter-id-au: Sure, let me do that

common/persistence/persistenceErrorInjectionClients.go
Shaddoll: you should copy the param

davidporter-id-au: yes 🤦  thank you

common/isolationgroup/default-isolation-group-state.go
davidporter-id-au: no, I didn't intend to put that change here. once again, thank you

docker/buildkite/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN curl https://bootstrap.pypa.io/pip/2.7/get-pip.py -o get-pip.py
 RUN python get-pip.py
 RUN pip install -U 'pip<21'
 RUN pip install PyYAML==3.13 cqlsh==5.0.4
+RUN pip install futures
```
davidporter-id-au, line 25: this is a build-fixing cherry-pick

common/resource/resourceImpl.go
davidporter-id-au: I don't love getters like this, but good point

common/isolationgroup/types.go
```diff
@@ -23,21 +23,13 @@
...
-	"github.com/uber/cadence/common/dynamicconfig"
 	"github.com/uber/cadence/common/types"
 )
 
 type ChangeEvent struct {
 	Changed types.IsolationGroupConfiguration
 }
 
-// Config is the base configuration
-type Config struct {
```
davidporter-id-au, line 35: Moving this out to config

common/config/config.go
```diff
@@ -515,6 +519,23 @@ type (
 		// URI is the domain default URI for visibility archiver
 		URI string `yaml:"URI"`
 	}
+
+	// Partitioner is the configuration for the partitioning library, a library
+	// for segmenting portions of workflows into isolation-groups - a resiliency
+	// concept meant to help move workflows around and away from failure zones.
+	Partitioner struct {
+		IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
```
Shaddoll, line 527: dynamic configs don't belong to this package
if you search TODO, we're moving them out of this package

common/dynamicconfig/constants.go
```diff
@@ -1818,6 +1818,14 @@ const (
 
 	EnableCassandraAllConsistencyLevelDelete
 
+	// EnableIsolationGroupPartitioning Is a feature to enable subdivision of workflows by units called 'isolation-groups'
+	// and to control their movement and blast radius. This has some nontrivial operational overheads in management
+	// and a good understanding of poller distribution, so probably not worth enabling unless it's well understood.
+	// KeyName: system.enableIsolationGroupPartitioning
+	// Value type: bool
+	// Default value: false
+	EnableIsolationGroupPartitioning
```
Shaddoll, line 1827: How about naming it `system.enableTasklistIsolation`

common/isolationgroup/interface.go
```diff
@@ -48,4 +51,7 @@ type State interface {
 
 	Subscribe(domainID, key string, notifyChannel chan<- ChangeEvent) error
 	Unsubscribe(domainID, key string) error
+
+	ProvideDomainCache(cache cache.DomainCache) State
```
Shaddoll, line 55: I don't think these methods are needed, because `domainCache` and `ConfigStore` may not be used by other implementation of this interface.

common/dynamicconfig/constants.go
```diff
@@ -2510,6 +2518,12 @@ const (
 	// Default value: 30 minutes
 	ESAnalyzerBufferWaitTime
 
+	// IsolationGroupCfgUpdateFrequency
+	// KeyName: system.isolationGroupCfgUpdateFrequency
+	// Value type: Duration
+	// Default value: 2 minutes
+	IsolationGroupCfgUpdateFrequency
```
Shaddoll, line 2525: How about naming it `system.isolationGroupStateRefreshInterval`?

davidporter-id-au: That's a better name

common/isolationgroup/interface.go
davidporter-id-au: This is a sequencing thing where I don't have access to DomainCache or ConfigStore when I'm calling the constructor.

I need to pass in the IsolationGroupState library in params, but at that point I don't have either the domain cache or the Config store. So this is the only way I could think to provide them in after they've been passed in as params.

common/resource/resourceImpl.go
```diff
@@ -241,6 +247,13 @@ func New(
 		return nil, err
 	}
 
+	var isolationGroupState isolationgroup.State
+	if params.IsolationGroupState != nil {
+		isolationGroupState = params.IsolationGroupState.
+			ProvideDomainCache(domainCache).
```
Shaddoll, line 253: I realized that we cannot access `domainCache` and `configStoreManager` inside `cmd/server` package.
How about moving the initialization logic here? 
```
var isolationGroupState isolationgroup.State
if params.IsolationGroupState != nil {
    isolationGroupState = params.IsolationGroupState
} else {
    isolationGroupState = isolationgroup.NewDefaultIsolationGroupState(...)
}
```

davidporter-id-au: I think if you're passing in one in the params, you'll still need for it to have the Domain-cache, so I don't think that it's enough to just instantiate here. 

However, instead of passing in the module like `params.IsolationGroupState` I could pass in a factory function

```
NewDefaultIsolationGroupStateFac() func (
	logger log.Logger,
	config config.IsolationGroups,
	done chan struct{},
        domainCache domainCache,
) State {
  return func NewDefaultIsolationGroupStateWatcher( logger log.Logger,	config config.IsolationGroups, done chantruct{}, dc domainCache, cfgmgr configManager ) State { ... }
}
```
and then still instantiate that inside `resourceImpl`

common/config/config.go
Shaddoll: I think we can remove them. They don't seem to be necessary.
Just use the dynamic config by names directly.

common/dynamicconfig/constants.go
```diff
@@ -3821,6 +3842,11 @@ var BoolKeys = map[BoolKey]DynamicBool{
 		Description:  "EnableAuthorization is the key to enable authorization for a domain, only for extension binary:",
 		DefaultValue: false,
 	},
+	EnableTasklistIsolation: DynamicBool{
+		KeyName:      "system.enableIsolationGroupPartitioning",
```
Shaddoll, line 3846: keyName is not updated.

davidporter-id-au: ah, that was sloppy of me, sorry, let me fix

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -54,17 +54,18 @@ type defaultIsolationGroupStateHandler struct {
 
 func NewDefaultIsolationGroupStateWatcher(
 	logger log.Logger,
-	domainCache cache.DomainCache,
 	config Config,
-	globalIsolationGroupDrains persistence.ConfigStore,
-	done chan bool,
+	domainCache cache.DomainCache,
+	manager persistence.ConfigStoreManager,
+	done chan struct{},
```
Shaddoll, line 60: why do we need an external channel?

davidporter-id-au: This is the service shutdown channel, used by the daemon to stop polling and not leak goroutines I guess? It's just a pattern used elsewhere I was following

common/partition/interface.go
```diff
@@ -30,15 +30,17 @@ import (
...
-type Config struct {
-	IsolationGroupEnabled dynamicconfig.BoolPropertyFnWithDomainFilter
-	AllIsolationGroups    []string
-}
-
 type Partitioner interface {
 	// GetIsolationGroupByDomainID gets where the task workflow should be executing. Largely used by Matching
 	// when determining which isolationGroup to place the tasks in.
 	// Implementations ought to return (nil, nil) for when the feature is not enabled.
 	GetIsolationGroupByDomainID(ctx context.Context, DomainID string, partitionKey PartitionConfig) (*string, error)
```
Shaddoll, line 37: In this method, we're using `IsolationGroupEnabled` with domainName as filter, but what we pass is `domainID`

davidporter-id-au: yes, the domainID is being resolved to name in `AvailableIsolationGroupsByDomainID`

common/partition/interface.go
```diff
@@ -30,15 +30,17 @@ import (
...
 }
+
+// Config is the configuration for the partitioning library, a library
+// for segmenting portions of workflows into isolation-groups - a resiliency
+// concept meant to help move workflows around and away from failure zones.
+type Config struct {
```
Shaddoll, line 43: It can be removed

davidporter-id-au: I would prefer to keep given configuration tends to evolve and change and having a wrapper struct makes for a much easier set of changes in the future

Shaddoll (approved):

PR #5180, opened 2023-03-30
fix build

davidporter-id-au:
Adds small dependency which was missing to the build container which was breaking `cqlsh` which was breaking some integration tests. 

Should we not be using Python 2.7 for this stuff? absolutely, however, we might have to come back to that. Actually I wouldn't mind ripping out Cqlsh as a dependency entirely, after all there's probably better ways to verify things using pure go.


Groxx:
easy workaround for the moment anyway 👍

PR #5184, opened 2023-04-03
Feature/e2e isolation group testing

davidporter-id-au:
Happy to pull this apart as it's a rather mixed-together diff. The summary is that in implementing the CRUD endpoints for Global isolation-groups I realised I needed to use the higher-level configstore.Client (a dynamic config client) instead of the lower-level persistence manager.

Therefore, to make these endpoints match the existing pattern in the CRUD operations for the dynamic config endpoints this:

- Changes the Isolation-group library to use the higher-level dynamicconfig.Client over atop the manager
- refactors the config in resourceimpl significantly to take some of the mess out of there and confine it to the library
- Adds a number of small mappers
- Unexports some of the configuration portions of the isolation-group library default implementation, since it's not super useful outside this context.
- Adds Admin CLI commands

### E2E tests
```
./cadence admin isolation-groups update-global --config '[{"Name": "zone-1", "State": 2}, {"Name": "zone-2", "State": 1}]'
./cadence admin isolation-groups get-global
```


common/resource/resource_test.go
```diff
@@ -25,77 +25,12 @@ package resource
...
-	"github.com/uber/cadence/common/cluster"
-	"github.com/uber/cadence/common/dynamicconfig"
-	"github.com/uber/cadence/common/isolationgroup"
-	"github.com/uber/cadence/common/log/loggerimpl"
-	"github.com/uber/cadence/common/metrics"
-	"github.com/uber/cadence/common/persistence"
-	"github.com/uber/cadence/common/service"
 )
 
-func TestEnsureIsolationGroupImpl(t *testing.T) {
```
davidporter-id-au, line 41: Removing this because the logic for determining this is moved to the IsolationGroup package

common/isolationgroup/interface.go
```diff
@@ -50,16 +47,8 @@ type State interface {
 
 	Subscribe(domainID, key string, notifyChannel chan<- ChangeEvent) error
 	Unsubscribe(domainID, key string) error
-}
 
-// Config values for the partitioning library for segmenting portions of workflows into isolation-groups - a resiliency
-// concept meant to help move workflows around and away from failure zones.
-type Config struct {
```
davidporter-id-au, line 57: unexposing this and moving it to the default implmentation since there's no need for it to be exposed

PR #5186, opened 2023-04-04
Merge Activity and Decision matching tests

Shaddoll:
**What changed?**
Merge Activity and Decision matching tests

**Why?**
To reuse test logic for activity and decision tasks

**How did you test it?**
the test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/matchingEngine_test.go
+++ b/service/matching/matchingEngine_test.go
@@ -27,6 +27,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -280,9 +281,6 @@ func (s *matchingEngineSuite) PollForDecisionTasksResultTest() {
 	_, err = addTask(s.matchingEngine, s.handlerContext, addRequest)
 	s.NoError(err)
 
-	taskList := &types.TaskList{}
-	taskList.Name = tl
-
 	resp, err = pollTask(s.matchingEngine, s.handlerContext, pollReq)
 
 	expectedResp := &pollTaskResponse{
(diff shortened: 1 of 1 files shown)


service/matching/matchingEngine_test.go
```diff
@@ -545,265 +511,207 @@ func (s *matchingEngineSuite) TestAddThenConsumeActivities() {
...
+}
 
+func (s *matchingEngineSuite) SyncMatchTasks(taskType int) {
 	const taskCount = 10
 	const initialRangeID = 102
 	// TODO: Understand why publish is low when rangeSize is 3
 	const rangeSize = 30
+	var throttledTaskCount int64
+	if taskType == persistence.TaskListTypeActivity {
+		throttledTaskCount = 3
```
davidporter-id-au, line 538: Is this just some test value?

service/matching/matchingEngine_test.go
```diff
@@ -545,265 +511,207 @@ func (s *matchingEngineSuite) TestAddThenConsumeActivities() {
...
-		s.EqualValues(activityInput, result.Input)
-		s.EqualValues(workflowExecution, *result.WorkflowExecution)
-		token := &common.TaskToken{
-			DomainID:     domainID,
-			WorkflowID:   workflowID,
-			RunID:        runID,
-			ScheduleID:   scheduleID,
-			ActivityID:   activityID,
-			ActivityType: activityTypeName,
+	for i := int64(0); i < throttledTaskCount; i++ {
```
davidporter-id-au, line 601: It's not imediately clear what's going on here without being pretty steeped in the polling code. Do you mind either mind putting a comment or wrapping this in some function which indicates the intent. It's not immediately clear to me what throttledTaskCount is intended to be used for?

davidporter-id-au:
Nit: Non-blocking comment - More generally, I know the temptation for putting particularly messy code with lots of duplication into a way which is DRYer is tempting, and for test-code it's probably ok, but having code covered by switch or IF statements does make it tricky to follow. 

In this instance you're starting out with 
```
if task == Decision {
 ... 
} else {
 ...
}
```
which is pretty simple, and I think works here. But if there are parts that start to get shared it risks becoming quite hard to follow. I'd lean towards just leaving the code duplicated.

davidporter-id-au (approved):

PR #5187, opened 2023-04-04
Physical sharding for NoSQL-based persistence

emrahs:
**What changed?**
Added a sharding layer to the NoSQL persistence stack so that Cadence can use multiple Cassandra clusters at once in a physically sharded manner. 

Cadence is a heavily storage-bounded system, so the limits for the load per Cadence cluster is strictly limited by the underlying storage system. Given the massive adoption of Cadence at Uber, this scale limitation forces us to create more Cadence clusters than we want to operate. This capability will let us have one or two orders of magnitude larger Cadence clusters than we have today. 

Note that this feature only enables bootstrapping a brand-new cluster with multiple databases behind it. *Resharding* is designed but not implemented yet. 

**Why?**
So that a Cadence cluster can be bootstrapped with multiple Cassandra clusters powering it. 

**How did you test it?**
Added unit tests. Ran samples and tested bench tests in a staging environment.

**Potential risks**
Since this change significantly changes the low-level persistence logic, it can cause data loss if something goes terribly wrong. 

**Release notes**
The change is backward compatible. Existing Cadence cluster configurations can be updated, if desired, to use the sharded NoSQL config format. However, they *must* continue having a single shard since Cadence still doesn't have the ability to *reshard* data.

**Documentation Changes**
There is a sample config file included in this PR that shows how to make use of the feature in a new cluster.


common/config/persistence.go
```diff
@@ -136,6 +143,50 @@ func (c *Persistence) Validate() error {
...
+			}
+
+			// validate history sharding
+			historyShardMapping := cfg.ShardingPolicy.HistoryShardMapping
+			currentShardID := -1
+			for _, shardRange := range historyShardMapping {
+				if _, found := connections[shardRange.Shard]; !found {
+					return fmt.Errorf("ShardedNosql config: Unknown history shard name: %v", shardRange.Shard)
+				}
+				if shardRange.Start != currentShardID+1 {
```
Shaddoll, line 166: I prefer to make start inclusive but end exclusive, because that's a pretty common behavior across different areas.

common/persistence/nosql/factory.go
```diff
@@ -26,29 +26,28 @@ import (
...
 	p "github.com/uber/cadence/common/persistence"
-	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin"
 )
 
 type (
 	// Factory vends datastore implementations backed by cassandra
 	Factory struct {
 		sync.RWMutex
-		cfg              config.Cassandra
+		cfg              config.ShardedNoSQL
```
davidporter-id-au, line 35: is this strictly a feature superset?

common/persistence/nosql/shardedNosqlStore.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+
+	sn.Lock()
+	if shard, ok := sn.connectedShards[shardName]; ok { // read again to double-check
+		sn.Unlock()
+		return &shard, nil
+	}
+
+	s, err := sn.connectToShard(cfg.NoSQLPlugin, shardName)
```
davidporter-id-au, line 125: if we don't have any already (and I may have missed them) it'd be great to get some metrics around the number of connections occurring

common/persistence/nosql/shardingPolicy.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (sp *shardingPolicy) getTaskListShardName(domainID string, taskListName string, taskType int) string {
+	if !sp.hasShardedTasklist {
+		sp.logger.Debug("Selected default store shard for tasklist", tag.StoreShard(sp.defaultShard), tag.WorkflowTaskListName(taskListName))
+		return sp.defaultShard
+	}
+	tlShards := sp.config.ShardingPolicy.TaskListHashing.ShardOrder
+	tlShardCount := len(tlShards)
+	hash := farm.Hash32([]byte(domainID+"_"+taskListName)) % uint32(tlShardCount)
```
davidporter-id-au, line 106: having just made a silly mistake with this myself, just double-cehcking, are there any conditions under which tlShardCount could be zero? https://go.dev/play/p/Pv73y1KpbOZ

Shaddoll, line 106: taskType is passed in but not used, should it be included to compute the hash?

common/config/config.go
```diff
@@ -239,6 +241,47 @@ type (
...
+		// ShardingPolicy is the configuration for the sharding strategy used
+		ShardingPolicy ShardingPolicy `yaml:"shardingPolicy"`
+		// Connections is the collection of NoSQL DB plugins that are used to connect to the shard
+		Connections map[string]DBShardConnection `yaml:"connections"`
+	}
+
+	// ShardingPolicy contains configuration for physical DB sharding
+	ShardingPolicy struct {
+		// HistoryShardMapping defines the ranges of history shards stored by each DB shard. Ranges listed here *MUST*
+		// be continuous and non-overlapping, such that the first range in the list starts from Shard 0, each following
```
davidporter-id-au, line 257: Nonblocking question: I could be lazy to look for this, but i didn't see it. Do we have any runtime checks to ensure this is true?

common/persistence/nosql/shardedNosqlStore.go
```diff
@@ -0,0 +1,147 @@
...
+func newShardedNosqlStore(cfg config.ShardedNoSQL, logger log.Logger, dc *p.DynamicConfiguration) (*shardedNosqlStore, error) {
+	sn := shardedNosqlStore{
+		config: cfg,
+		dc:     dc,
+		logger: logger,
+	}
+
+	// Connect to the default shard
+	defaultShardName := cfg.DefaultShard
+	store, err := sn.connectToShard(cfg.Connections[defaultShardName].NoSQLPlugin, defaultShardName)
```
Shaddoll, line 58: it seems that `connectToShard` function only needs one parameter.

common/persistence/nosql/shardedNosqlStore.go
```diff
@@ -0,0 +1,147 @@
...
+	shardName := sn.shardingPolicy.getTaskListShardName(domainID, taskListName, taskType)
+	return sn.getShard(shardName)
+}
+
+func (sn *shardedNosqlStore) GetDefaultShard() nosqlStore {
+	return sn.defaultShard
+}
+
+func (sn *shardedNosqlStore) Close() {
+	sn.defaultShard.Close()
```
Shaddoll, line 91: nit: defaultShard is closed twice

common/config/config.go
Shaddoll: It's in `persistence.go`

common/config/config.go
```diff
@@ -198,6 +198,8 @@ type (
 		SQL *SQL `yaml:"sql"`
 		// NoSQL contains the config for a NoSQL based datastore
 		NoSQL *NoSQL `yaml:"nosql"`
```
Shaddoll, line 200: Are we deprecating `NoSQL` field? 
Add a comment to announce the deprecation?

Shaddoll (commented):
Are you adding new test to the buildkite pipeline?
I think we need to build an environment with shardedNoSQL to run all the persistence and integration tests.

Shaddoll (commented):
Another question: it seems that the sharding policy thing doesn't have any dependency on nosql, so probably the logic should not be put into `nosql` package. The logic can be reused for sql case, any plans for shardedSQL support?

common/persistence/nosql/shardingPolicy.go
emrahs: It should be impossible with the `if !sp.hasShardedTasklist` at the top of the func.

common/persistence/nosql/factory.go
emrahs: Cassandra is same as NoSQL and ShardedNoSQL is strictly a superset of it, so a NoSQL configuration can be parsed into a ShardedNoSQL config object.

common/config/persistence.go
emrahs: Yeah, I also debated the two options when modeling this part of the config. In a usual scenario, I agree that the start-inclusive & end-exclusive pattern works the best. However, when I was thinking about more arbitrary shard distributions (i.e. creating a mapping for a single shard) it felt like using inclusive start & end was making the config a little more readable.

Thinking about it again though, there are other ways to cleanly handle single shard pinning if it ever becomes a need. So I'll change this to the standard range pattern.

common/persistence/nosql/shardingPolicy.go
emrahs: `taskType` is in the interface, and the sharding implementation in SQL persistence uses it. I don't like the idea of using it though, because the more randomization is added to the decision making makes a single workflow depend on the more number of DBs being available. With the current implementation, a given workflow can continue running if only 3 databases are available: default DB for metadata, the DB that has the execution row, and the DB that has the tasklist row. If we take taskType into account, that'll increase this dependency factor from 3 to 4 without much value. Therefore, I decided to keep it out of the hash calculation.

common/persistence/nosql/shardedNosqlStore.go
emrahs: That's right! I forgot to clean this up while doing refactoring.

common/config/config.go
emrahs: We should not deprecate it right away. Most users outside Uber have very small Cadence deployments and the ShardedNoSQL config will just be unnecessarily more complex for them to maintain.

emrahs:
> Are you adding new test to the buildkite pipeline? I think we need to build an environment with shardedNoSQL to run all the persistence and integration tests.

Good idea but I'll leave that out as a separate task.

emrahs:
> Another question: it seems that the sharding policy thing doesn't have any dependency on nosql, so probably the logic should not be put into `nosql` package. The logic can be reused for sql case, any plans for shardedSQL support?

There is already a partitioning implementation for SQL and I think it has various weaknesses and needs to be refactored so that it can support resharding and does better job at colocation of related records across executions and history tables. Unifying the two sharding approaches by refactoring the SQL implementation is a good next step, and reusing some of the pieces in this PR might make sense at that time. I'll keep shardingPolicy in NoSQL implementation for now until it's needed outside the package.

Shaddoll (approved):

PR #5191, opened 2023-04-05
enables CRUD operations for global isolation-groups

davidporter-id-au:
**What changed?**

- Adds CRUD operations to isolation-group library
- Moves isolation-group-library to use the dynamic config client in line with other config-store providers
- Unexports some configuration no longer making sense to export for the IsolationGroup library

**Why?**

**How did you test it?**
```
select * from cluster_config;

@ Row 1
-----------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 row_type  | 1
 version   | 1
 encoding  | thriftrw
 timestamp | 2023-04-07 02:14:47.117000+0000
 values    | 0x590a000a00000000000000000f00140c000000010b000a0000003b73797374656d2e64656661756c7449736f6c6174696f6e47726f7570436f6e66696753746f72654d616e61676572476c6f62616c4d617070696e670f00140c000000010c000a08000a000000010b0014000000395b7b224e616d65223a227a6f6e652d31222c225374617465223a327d2c7b224e616d65223a227a6f6e652d32222c225374617465223a317d5d00000000

(1 rows)
```
```
cadence admin db decode_thrift -i '590a000a00000000000000000f00140c000000010b000a0000003b73797374656d2e64656661756c7449736f6c6174696f6e47726f7570436f6e66696753746f72654d616e61676572476c6f62616c4d617070696e670f00140c000000010c000a08000a000000010b0014000000395b7b224e616d65223a227a6f6e652d31222c225374617465223a327d2c7b224e616d65223a227a6f6e652d32222c225374617465223a317d5d00000000'
======= Decode into type config.DynamicConfigBlob ========
(*config.DynamicConfigBlob)(0x140001442e0)(DynamicConfigBlob{SchemaVersion: 0, Entries: [DynamicConfigEntry{Name: system.defaultIsolationGroupConfigStoreManagerGlobalMapping, Values: [DynamicConfigValue{Value: DataBlob{EncodingType: JSON, Data: [91 123 34 78 97 109 101 34 58 34 122 111 110 101 45 49 34 44 34 83 116 97 116 101 34 58 50 125 44 123 34 78 97 109 101 34 58 34 122 111 110 101 45 50 34 44 34 83 116 97 116 101 34 58 49 125 93]}}]}]})
======= As JSON ========
{"schemaVersion":0,"entries":[{"name":"system.defaultIsolationGroupConfigStoreManagerGlobalMapping","values":[{"value":{"EncodingType":"JSON","Data":"W3siTmFtZSI6InpvbmUtMSIsIlN0YXRlIjoyfSx7Ik5hbWUiOiJ6b25lLTIiLCJTdGF0ZSI6MX1d"}}]}]}
➜  cadence git:(feature/ig-lib-refactor) pbpaste | jq .
{
  "schemaVersion": 0,
  "entries": [
    {
      "name": "system.defaultIsolationGroupConfigStoreManagerGlobalMapping",
      "values": [
        {
          "value": {
            "EncodingType": "JSON",
            "Data": "W3siTmFtZSI6InpvbmUtMSIsIlN0YXRlIjoyfSx7Ik5hbWUiOiJ6b25lLTIiLCJTdGF0ZSI6MX1d"
          }
        }
      ]
    }
  ]
}
```
```
echo W3siTmFtZSI6InpvbmUtMSIsIlN0YXRlIjoyfSx7Ik5hbWUiOiJ6b25lLTIiLCJTdGF0ZSI6MX1d | base64 -D
[{"Name":"zone-1","State":2},{"Name":"zone-2","State":1}]
```

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/configstore/config_store_client.go
+++ b/common/dynamicconfig/configstore/config_store_client.go
@@ -677,6 +677,10 @@ func validateKeyDataBlobPair(key dc.Key, blob *types.DataBlob) error {
 		if _, ok := value.(map[string]interface{}); !ok {
 			return err
 		}
+	case dc.ListKey:
+		if _, ok := value.([]interface{}); !ok {
+			return err
+		}
 	default:
 		return fmt.Errorf("unknown key type: %T", key)
 	}
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2527,7 +2527,7 @@ const (
 	// IsolationGroupStateRefreshInterval
 	// KeyName: system.isolationGroupStateRefreshInterval
 	// Value type: Duration
-	// Default value: 2 minutes
+	// Default value: 30 seconds
 	IsolationGroupStateRefreshInterval
 	// IsolationGroupStateFetchTimeout is the dynamic config DB fetch timeout value
 	// KeyName: system.isolationGroupStateFetchTimeout
@@ -2613,6 +2613,12 @@ const (
 	// Default value: N/A
 	// Allowed filters: N/A
 	AllIsolationGroups
+
+	// DefaultIsolationGroupConfigStoreManagerGlobalMapping is the dynamic config value for isolation groups
+	// Note: This is not typically used for normal dynamic config (type 0), but instead
+	// it's used only for IsolationGroup config (type 1).
+	// KeyName: system.defaultIsolationGroupConfigStoreManagerGlobalMapping
+	DefaultIsolationGroupConfigStoreManagerGlobalMapping
 
 	LastListKey
 )
@@ -4480,7 +4486,7 @@ var DurationKeys = map[DurationKey]DynamicDuration{
 	IsolationGroupStateRefreshInterval: DynamicDuration{
 		KeyName:      "system.isolationGroupStateRefreshInterval",
 		Description:  "the frequency by which the IsolationGroupState handler will poll configuration",
-		DefaultValue: time.Minute * 2,
+		DefaultValue: time.Second * 30,
 	},
 	IsolationGroupStateFetchTimeout: DynamicDuration{
 		KeyName:      "system.IsolationGroupStateFetchTimeout",
@@ -4537,6 +4543,15 @@ var MapKeys = map[MapKey]DynamicMap{
 }
 
 var ListKeys = map[ListKey]DynamicList{
+	AllIsolationGroups: {
+		KeyName:     "system.allIsolationGroups",
+		Description: "A list of all the isolation groups in a system",
+	},
+	DefaultIsolationGroupConfigStoreManagerGlobalMapping: {
+		KeyName: "system.defaultIsolationGroupConfigStoreManagerGlobalMapping",
+		Description: "A configuration store for global isolation groups - used in isolation-group config only, not normal dynamic config." +
+			"Not intended for use in normal dynamic config",
+	},
 	HeaderForwardingRules: {
 		KeyName: "admin.HeaderForwardingRules", // make a new scope for global?
 		Description: "Only loaded at startup.  " +
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2606,7 +2606,6 @@ const (
 	// Value type: []rpc.HeaderRule or an []interface{} containing `map[string]interface{}{"Add":bool,"Match":string}` values.
 	// Default value: forward all headers.  (this is a problematic value, and it will be changing as we reduce to a list of known values)
 	HeaderForwardingRules
-
 	// AllIsolationGroups is the list of all possible isolation groups in a service
 	// KeyName: system.allIsolationGroups
 	// Value type: []string
@@ -2614,14 +2613,14 @@ const (
 	// Allowed filters: N/A
 	AllIsolationGroups
 
-	// DefaultIsolationGroupConfigStoreManagerGlobalMapping is the dynamic config value for isolation groups
-	// Note: This is not typically used for normal dynamic config (type 0), but instead
-	// it's used only for IsolationGroup config (type 1).
-	// KeyName: system.defaultIsolationGroupConfigStoreManagerGlobalMapping
-	DefaultIsolationGroupConfigStoreManagerGlobalMapping
-
 	LastListKey
 )
+
+// DefaultIsolationGroupConfigStoreManagerGlobalMapping is the dynamic config value for isolation groups
+// Note: This is not typically used for normal dynamic config (type 0), but instead
+// it's used only for IsolationGroup config (type 1).
+// KeyName: system.defaultIsolationGroupConfigStoreManagerGlobalMapping
+const DefaultIsolationGroupConfigStoreManagerGlobalMapping ListKey = -1 // This is a hack to put it in a different list due to it being a different config type
 
 var IntKeys = map[IntKey]DynamicInt{
 	TestGetIntPropertyKey: DynamicInt{
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1360,6 +1360,13 @@ const (
 	// KeyName: system.largeShardHistoryBlobMetricThreshold
 	// Value type: Int
 	// Default value: 262144 (1/4mb)
+
+	// IsolationGroupStateUpdateRetryAttempts
+	// KeyName: system.isolationGroupStateUpdateRetryAttempts
+	// Value type: int
+	// Default value: 2
+	IsolationGroupStateUpdateRetryAttempts
+
 	LargeShardHistoryBlobMetricThreshold
 	// LastIntKey must be the last one in this const group
 	LastIntKey
@@ -2517,12 +2524,21 @@ const (
 	// Value type: Duration
 	// Default value: 30 minutes
 	ESAnalyzerBufferWaitTime
-
 	// IsolationGroupStateRefreshInterval
 	// KeyName: system.isolationGroupStateRefreshInterval
 	// Value type: Duration
 	// Default value: 2 minutes
 	IsolationGroupStateRefreshInterval
+	// IsolationGroupStateFetchTimeout is the dynamic config DB fetch timeout value
+	// KeyName: system.isolationGroupStateFetchTimeout
+	// Value type: Duration
+	// Default value: 30 seconds
+	IsolationGroupStateFetchTimeout
+	// IsolationGroupStateUpdateTimeout is the dynamic config DB update timeout value
+	// KeyName: system.isolationGroupStateUpdateTimeout
+	// Value type: Duration
+	// Default value: 30 seconds
+	IsolationGroupStateUpdateTimeout
 
 	// LastDurationKey must be the last one in this const group
 	LastDurationKey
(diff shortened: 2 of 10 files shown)


common/isolationgroup/default-isolation-group-state.go
```diff
@@ -24,50 +24,106 @@ package isolationgroup
...
-	subscriptionMu             sync.Mutex
-	valuesMu                   sync.RWMutex
-	lastSeen                   *isolationGroups
-	updateCB                   func()
+	status                          int32
+	done                            chan struct{}
+	log                             log.Logger
+	domainCache                     cache.DomainCache
+	globalIsolationGroupDrains      dynamicconfig.Client
+	globalIsolationGroupDrainClient dynamicconfig.Client
```
Shaddoll, line 61: this field is duplicate

common/resource/resource_test.go
```diff
@@ -25,77 +25,12 @@ package resource
...
-	"github.com/uber/cadence/common/cluster"
-	"github.com/uber/cadence/common/dynamicconfig"
-	"github.com/uber/cadence/common/isolationgroup"
-	"github.com/uber/cadence/common/log/loggerimpl"
-	"github.com/uber/cadence/common/metrics"
-	"github.com/uber/cadence/common/persistence"
-	"github.com/uber/cadence/common/service"
 )
 
-func TestEnsureIsolationGroupImpl(t *testing.T) {
```
davidporter-id-au, line 41: removed this because this logic was moved to the constructor in the library

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -168,21 +252,33 @@ func (z *defaultIsolationGroupStateHandler) checkIfChanged() {
...
-		select {
-		case <-z.done:
-			return
-		case <-ticker.C:
-			z.checkIfChanged()
+// A simple explicit deny-based isolation group implementation
+func availableIG(allIsolationGroups []string, global types.IsolationGroupConfiguration, domain types.IsolationGroupConfiguration) types.IsolationGroupConfiguration {
+	out := types.IsolationGroupConfiguration{}
+	for _, isolationGroup := range allIsolationGroups {
+		globalCfg, hasGlobalConfig := global[string(isolationGroup)]
```
Shaddoll, line 259: nit: `string()` can be removed

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -52,22 +68,61 @@ type defaultIsolationGroupStateHandler struct {
 	subscriptions map[string]map[string]chan<- ChangeEvent
 }
 
+const (
+	DefaultIsolationGroupConfigStoreManagerGlobalMapping dynamicconfig.MapKey = iota
```
Shaddoll, line 72: The key should not be exported. And why do we use MapKey?

davidporter-id-au: Regarding the export, good call, let me fix, 

regarding the mapkey, my possibly flawed memory was that it is because the value encoded actually is a map (`echo eyJOYW1lIjoiem9uZS0xIiwiU3RhdGUiOjJ9 | base64 -D
{"Name":"zone-1","State":2}`) but let me verify

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -196,30 +292,65 @@ func isDrained(isolationGroup string, global types.IsolationGroupConfiguration,
...
 		}
+		allIsolationGroups = append(allIsolationGroups, v)
 	}
-	return out
+	return allIsolationGroups, nil
 }
 
-func fromCfgStore(in *persistence.FetchDynamicConfigResponse) (types.IsolationGroupConfiguration, error) {
-	panic("not implemented")
+func mapUpdateGlobalIsolationGroupsRequest(in types.IsolationGroupConfiguration) ([]*types.DynamicConfigValue, error) {
```
Shaddoll, line 336: I think it's not the right way to persist configuration. Why do we need to convert it to `types.DynamicConfigValue`?
Why not just store `types.IsolationGroupConfiguration`? If we do this, the data is more clean. We'll have something like 
```
{
  "schemaVersion": 0,
  "entries": [
    {
      "name": "",
      "values": [
        {
          "value": {
            "IsolationGroup": "zone-1",
            "State": 2
          }
        },
        {
          "value": {
            "IsolationGroup": "zone-2",
            "State": 1
          }
        }
      ]
    }
  ]
}
```

davidporter-id-au:
As per discussion: 

For comparison with the Update / Get operations on dynamic config
```
# creating a map-key
# ./cadence admin config update  --name "testGetMapPropertyKey"  --value '{"Value": {"isolation-groups": { "zone-123": {"State": 1}, "zone-321": {"State": 2}}} ,"Filters":[]}'
#
# Note that a list key doesn't appear to be implemented correctly, so not used here.
{
  "schemaVersion": 0,
  "entries": [
    {
      "name": "testGetMapPropertyKey",
      "values": [
        {
          "value": {
            "EncodingType": "JSON",
            "Data": "eyJpc29sYXRpb24tZ3JvdXBzIjp7InpvbmUtMTIzIjp7IlN0YXRlIjoxfSwiem9uZS0zMjEiOnsiU3RhdGUiOjJ9fX0="
          }
        }
      ]
    }
  ]
}


# ./cadence admin config get --name "testGetMapPropertyKey"

[
  {
    "Name": "testGetMapPropertyKey",
    "Values": [
      {
        "Value": {
          "isolation-groups": {
            "zone-123": {
              "State": 1
            },
            "zone-321": {
              "State": 2
            }
          }
        },
        "Filters": []
      }
    ]
  }
]

```

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -196,30 +293,77 @@ func isDrained(isolationGroup string, global types.IsolationGroupConfiguration,
...
+	for _, v := range dcData {
+		v1, ok := v.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("failed parse a dynamic config entry, %v, (got %v)", v1, v)
 		}
-		if hasDomainConfig {
-			if domainCfg.State == types.IsolationGroupStateDrained {
-				continue
-			}
+		n, okName := v1["Name"]
```
davidporter-id-au, line 313: I'm really unhappy having to hardcode these like this, but I'm covering possible serialization changes with a unit-test

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -677,6 +677,10 @@ func validateKeyDataBlobPair(key dc.Key, blob *types.DataBlob) error {
 		if _, ok := value.(map[string]interface{}); !ok {
 			return err
 		}
+	case dc.ListKey:
```
davidporter-id-au, line 680: this was missing, weirdly

common/dynamicconfig/constants.go
```diff
@@ -2598,6 +2614,12 @@ const (
 	// Allowed filters: N/A
 	AllIsolationGroups
 
+	// DefaultIsolationGroupConfigStoreManagerGlobalMapping is the dynamic config value for isolation groups
+	// Note: This is not typically used for normal dynamic config (type 0), but instead
+	// it's used only for IsolationGroup config (type 1).
+	// KeyName: system.defaultIsolationGroupConfigStoreManagerGlobalMapping
+	DefaultIsolationGroupConfigStoreManagerGlobalMapping
```
Shaddoll, line 2621: could you set it to a minus int so that if we iterate the key from Unknown to Last, this key won't be included.

common/resource/resourceImpl.go
```diff
@@ -568,38 +573,23 @@ func (h *Impl) GetPartitioner() partition.Partitioner {
...
 	if params.IsolationGroupState != nil {
 		return params.IsolationGroupState
 	}
-	cfg := isolationgroup.Config{
-		IsolationGroupEnabled: dc.GetBoolPropertyFilteredByDomain(dynamicconfig.EnableTasklistIsolation),
-		UpdateFrequency:       dc.GetDurationProperty(dynamicconfig.IsolationGroupStateRefreshInterval),
-		AllIsolationGroups:    allIsolationGroups,
+
+	ig, err := isolationgroup.NewDefaultIsolationGroupStateWatcher(params.Logger, dc, &params.PersistenceConfig, domainCache)
+	if err != nil {
```
Shaddoll, line 588: I think we should return the error and fail the start of the service.

common/isolationgroup/default-isolation-group-state.go
```diff
@@ -196,30 +293,77 @@ func isDrained(isolationGroup string, global types.IsolationGroupConfiguration,
...
-	out := types.IsolationGroupConfiguration{}
-	for _, isolationGroup := range allIsolationGroups {
-		globalCfg, hasGlobalConfig := global[string(isolationGroup)]
-		domainCfg, hasDomainConfig := domain[string(isolationGroup)]
-		if hasGlobalConfig {
-			if globalCfg.State == types.IsolationGroupStateDrained {
-				continue
-			}
+// ----- Mappers -----
+func mapDynamicConfigResponse(in interface{}) (out types.IsolationGroupConfiguration, err error) {
```
Shaddoll, line 297: the input type can be `[]interface{}`

common/isolationgroup/default-isolation-group-state.go
Shaddoll: an alternative way is to serialize it and deserialize it into the type you want

common/isolationgroup/default-isolation-group-state_test.go
Shaddoll, line 1: The test coverage is only 52%.

common/isolationgroup/default-isolation-group-state.go
davidporter-id-au: True, I didn't think of that.... I'm ok with this approach as-is for now, but if this were of significant overhead I'd give that some consideration.

common/resource/resourceImpl.go
davidporter-id-au: There will be scenarios for which it won't be available (SQL DBs, for example). I'm not sure we want to do make this a requirement. 

I understand that you might be relying on it being available in the matching changes. Open to discussing, but I'd suggest we want to take the availability of the library as part of the decision to enable the feature.

Shaddoll: if it can be `nil`, should `partitioner` be treated in the same way? Because it depends on it

common/isolationgroup/default-isolation-group-state_test.go
```diff
@@ -23,13 +23,110 @@
...
+			Name:  "zone-1",
+			State: types.IsolationGroupStateHealthy,
+		},
+		"zone-2": types.IsolationGroupPartition{
+			Name:  "zone-2",
+			State: types.IsolationGroupStateDrained,
+		},
+	}
+
+	validCfg, _ := mapUpdateGlobalIsolationGroupsRequest(validInput)
```
Shaddoll, line 56: nit: I'd still check these errors with `require.NoError(t, err)`.

Shaddoll (approved):

common/isolationgroup/default-isolation-group-state_test.go
davidporter-id-au: fair, let me do that

PR #5193, opened 2023-04-05
Update matching to support tasklist isolation

Shaddoll:
**What changed?**
Update matching to support tasklist isolation
- For domain with tasklist isolation enabled, tasks are only assigned to specific pollers from one isolation group which is calculated from partition config by using a partitioner library

**Why?**
To support worker isolation

**How did you test it?**
new test cases

**Potential risks**
If the feature is enabled, losing pollers from one isolation group can block the task processing of other isolation groups

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/config.go
+++ b/service/matching/config.go
@@ -69,6 +69,10 @@ type (
 		EnableTaskInfoLogByDomainID dynamicconfig.BoolPropertyFnWithDomainIDFilter
 
 		ActivityTaskSyncMatchWaitTime dynamicconfig.DurationPropertyFnWithDomainFilter
+
+		// isolation configuration
+		EnableTasklistIsolation dynamicconfig.BoolPropertyFnWithDomainFilter
+		AllIsolationGroups      []string
 	}
 
 	forwarderConfig struct {
@@ -96,6 +100,8 @@ type (
 		MaxTaskBatchSize                func() int
 		NumWritePartitions              func() int
 		NumReadPartitions               func() int
+		// isolation configuration
+		EnableTasklistIsolation bool
 	}
 )
 
@@ -130,7 +136,20 @@ func NewConfig(dc *dynamicconfig.Collection) *Config {
 		EnableDebugMode:                 dc.GetBoolProperty(dynamicconfig.EnableDebugMode)(),
 		EnableTaskInfoLogByDomainID:     dc.GetBoolPropertyFilteredByDomainID(dynamicconfig.MatchingEnableTaskInfoLogByDomainID),
 		ActivityTaskSyncMatchWaitTime:   dc.GetDurationPropertyFilteredByDomain(dynamicconfig.MatchingActivityTaskSyncMatchWaitTime),
+		EnableTasklistIsolation:         dc.GetBoolPropertyFilteredByDomain(dynamicconfig.EnableTasklistIsolation),
+		AllIsolationGroups:              mapIGs(dc.GetListProperty(dynamicconfig.AllIsolationGroups)()),
 	}
+}
+
+func mapIGs(in []interface{}) []string {
+	var allIsolationGroups []string
+	for k := range in {
+		v, ok := in[k].(string)
+		if ok {
+			allIsolationGroups = append(allIsolationGroups, v)
+		}
+	}
+	return allIsolationGroups
 }
 
 func newTaskListConfig(id *taskListID, config *Config, domainCache cache.DomainCache) (*taskListConfig, error) {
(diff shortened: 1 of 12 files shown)


service/matching/forwarder.go
```diff
@@ -200,6 +202,7 @@ func (fwdr *Forwarder) ForwardPoll(ctx context.Context) (*InternalTask, error) {
 
 	pollerID, _ := ctx.Value(pollerIDKey).(string)
 	identity, _ := ctx.Value(identityKey).(string)
+	isolationGroup, _ := ctx.Value(_isolationGroupKey).(string)
```
davidporter-id-au, line 205: This context.Value() is... probably nil safe, but mgiht be worth double-checking

service/matching/matcher.go
```diff
@@ -37,7 +38,8 @@ import (
 // that drains backlog from db. Consumers are the task list pollers
 type TaskMatcher struct {
 	// synchronous task channel to match producer/consumer
-	taskC chan *InternalTask
+	taskC         chan *InternalTask
+	isolatedTaskC map[string]chan *InternalTask
```
davidporter-id-au, line 42: Do you mind adding a small comment specifying the key to the string (zone / isolation-group I assume), but worth probably just makign that clear here

service/matching/taskListManager.go
```diff
@@ -380,7 +393,11 @@ func (c *taskListManagerImpl) getTask(ctx context.Context, maxDispatchPerSecond
 		return c.matcher.PollForQuery(childCtx)
 	}
 
-	return c.matcher.Poll(childCtx)
+	var isolationGroup string
+	if c.isIsolationEnabled() {
+		isolationGroup, _ = ctx.Value(_isolationGroupKey).(string)
```
davidporter-id-au, line 398: What is the intent of passing this through context?

service/matching/config.go
```diff
@@ -69,6 +69,10 @@ type (
 		EnableTaskInfoLogByDomainID dynamicconfig.BoolPropertyFnWithDomainIDFilter
 
 		ActivityTaskSyncMatchWaitTime dynamicconfig.DurationPropertyFnWithDomainFilter
+
+		// isolation configuration
+		EnableTasklistIsolation dynamicconfig.BoolPropertyFnWithDomainFilter
+		AllIsolationGroups      []string
```
davidporter-id-au, line 75: Nonblocking, but since we're using this on startup in two libraries and these matching changes I'd be leaning to just putting this as part of the shared service resource config and loading it at startup so there's a single source of truth.

service/matching/config.go
```diff
@@ -96,6 +100,8 @@ type (
 		MaxTaskBatchSize                func() int
 		NumWritePartitions              func() int
 		NumReadPartitions               func() int
+		// isolation configuration
+		EnableTasklistIsolation bool
```
davidporter-id-au, line 104: What is the relationship between this and the above `EnableTasklistIsolation`?

service/matching/taskReader.go
```diff
@@ -129,7 +134,15 @@ dispatchLoop:
...
 				break dispatchLoop
 			}
-			task := newInternalTask(taskInfo, tr.completeTask, types.TaskSourceDbBacklog, "", false, nil)
+			isolationGroup := ""
+			if tr.isIsolationEnabled() {
+				ctx, cancel := context.WithTimeout(tr.cancelCtx, time.Second)
+				if group, err := tr.partitioner.GetIsolationGroupByDomainID(ctx, taskInfo.DomainID, taskInfo.PartitionConfig); err == nil && group != nil {
+					isolationGroup = *group
+				}
+				cancel()
```
davidporter-id-au, line 143: If we see timeouts here we probably want a log or something

service/matching/taskListManager.go
```diff
@@ -323,6 +345,10 @@ func (c *taskListManagerImpl) GetTask(
 	ctx context.Context,
 	maxDispatchPerSecond *float64,
 ) (*InternalTask, error) {
+	if c.shouldReload() {
```
davidporter-id-au, line 348: I think this is a nice compromise with regard to feature enablement, though I would definately add a line in your implementation ERD About this choice. 

If you return with errNoTasks, I imagine the poller behaviour is to immediately re-poll, is that correct?

service/matching/matcher.go
```diff
@@ -266,28 +280,32 @@ forLoop:
 // Poll blocks until a task is found or context deadline is exceeded
 // On success, the returned task could be a query task or a regular task
 // Returns ErrNoTasks when context deadline is exceeded
-func (tm *TaskMatcher) Poll(ctx context.Context) (*InternalTask, error) {
+func (tm *TaskMatcher) Poll(ctx context.Context, isolationGroup string) (*InternalTask, error) {
+	isolatedTaskC, ok := tm.isolatedTaskC[isolationGroup]
+	if !ok && isolationGroup != "" {
+		return nil, fmt.Errorf("invalid isolation group: %s", isolationGroup)
```
davidporter-id-au, line 286: Nit: this probably should be a `types.BadRequestError{}` or some kind of error to indicate that it's a caller problem, not a server-side problem, otherwise it'll burn our SLAs

davidporter-id-au (approved):

service/matching/taskListManager.go
Shaddoll: Yes, and it has no effect on our SLA

service/matching/matcher.go
Shaddoll: good point.

PR #5194, opened 2023-04-05
Adds global Isolation-group client and admin commands

davidporter-id-au:
**What changed?**

This adds two commands for updating and getting global isolation groups: 
`admin isolation-groups update-global`
`admin isolation-groups get-global`

This also adds the same commands to the client library: 
- GetGlobalIsolationGroups
- UpdateGlobalIsolationGroups

There remains also the domain-level isolation groups yet to be implemented, though they will be similarly be added in this command. 

**Why?**

**How did you test it?**

Unit-tests
End to end testing manually with the feature fully written with the following: 
```
./cadence admin isolation-groups update-global --config '[{"Name": "zone-1", "State": 2}, {"Name": "zone-2", "State": 1}]'
./cadence admin isolation-groups get-global
```

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/client/admin/errorInjectionClient.go
+++ b/client/admin/errorInjectionClient.go
@@ -778,7 +778,24 @@ func (c *errorInjectionClient) GetGlobalIsolationGroups(
 	request *types.GetGlobalIsolationGroupsRequest,
 	opts ...yarpc.CallOption,
 ) (*types.GetGlobalIsolationGroupsResponse, error) {
-	panic("not implemented")
+	fakeErr := errors.GenerateFakeError(c.errorRate)
+	var resp *types.GetGlobalIsolationGroupsResponse
+	var clientErr error
+	var forwardCall bool
+	if forwardCall = errors.ShouldForwardCall(fakeErr); forwardCall {
+		resp, clientErr = c.client.GetGlobalIsolationGroups(ctx, request, opts...)
+	}
+
+	if fakeErr != nil {
+		c.logger.Error(msgInjectedFakeErr,
+			tag.AdminClientOperationGetGlobalIsolationGroups,
+			tag.Error(fakeErr),
+			tag.Bool(forwardCall),
+			tag.ClientError(clientErr),
+		)
+		return nil, fakeErr
+	}
+	return resp, clientErr
 }
 
 func (c *errorInjectionClient) UpdateGlobalIsolationGroups(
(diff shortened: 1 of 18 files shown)


client/admin/errorInjectionClient.go
```diff
@@ -772,3 +772,19 @@ func (c *errorInjectionClient) ListDynamicConfig(
 	}
 	return resp, clientErr
 }
+
+func (c *errorInjectionClient) GetGlobalIsolationGroups(
+	ctx context.Context,
+	request *types.GetGlobalIsolationGroupsRequest,
+	opts ...yarpc.CallOption,
+) (*types.GetGlobalIsolationGroupsResponse, error) {
+	panic("not implemented")
```
Shaddoll, line 781: could you implement this?

common/metrics/defs.go
```diff
@@ -799,6 +803,10 @@ const (
 	AdminDeleteWorkflowScope
 	// MaintainCorruptWorkflowScope is the metric scope for admin.MaintainCorruptWorkflow
 	MaintainCorruptWorkflowScope
+	// GetGlobalIsolationGroups is the scope for getting global isolation groups
+	GetGlobalIsolationGroups
```
Shaddoll, line 807: these 2 are not used

common/types/mapper/proto/admin.go
Shaddoll, line 1: You didn't add new test cases.

common/types/mapper/proto/admin.go
```diff
@@ -1173,13 +1173,42 @@ func FromGetGlobalIsolationGroupsResponse(t *types.GetGlobalIsolationGroupsRespo
...
+func ToGetGlobalIsolationGroupsResponse(t *adminv1.GetGlobalIsolationGroupsResponse) *types.GetGlobalIsolationGroupsResponse {
+	if t == nil {
+		return nil
+	}
+	ig := isolationGroupConfigFromIDL(t.IsolationGroups)
+	if ig == nil {
+		return nil
+	}
+	return &types.GetGlobalIsolationGroupsResponse{
+		IsolationGroups: *ig,
```
Shaddoll, line 1208: I'm surprised that the field is not a pointer.

tools/cli/admin.go
Shaddoll, line 1: The CLI needs more discussion. The current one, I think, it is hard to use, because user has to use a JSON string as input.

davidporter-id-au: I completely agree, but I'd like to land it and iterate if possible, because even things such as the API are not entirely finalized, so the final shape of the API I think needs to be determined, but this is enough to unblock e2t testing. 

I'm not at all happy with thing such as `2`, and `1` being represented in JSON for the various states. This will need to change

common/types/mapper/proto/admin.go
davidporter-id-au: good point

PR #5195, opened 2023-04-06
Add helper function to store/retrieve partition config from context

Shaddoll:
**What changed?**
- Add helper function to store/retrieve partition config into/from context
- Update frontend workflowHandler to retrieve partition config from context and set it to the requests to history
- Add a middleware (`ZonalPartitionConfigMiddleware`) as an example to demonstrate how to store partition config into context
- Update clusterRedirectionHandler to set a header to indicate which cluster the requests are from for all auto-forwarding requests
- Add a middleware (`ForwardPartitionConfigMiddleware`) which forwards partition config to remote cluster

`ForwardPartitionConfigMiddleware` must be used if we want to enable tasklist isolation feature. `ZonalPartitionConfigMiddleware` can be replaced with a different implementation. For example, if zonal isolation is enabled at network layer, we can create a middleware which derives the zone which the requests are from on the server side, which is more reliable than letting client set a header indicating which zone the requests are from.

**Why?**
To provide a way to set isolation groups for frontend requests

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/partition/default-partitioner.go
+++ b/common/partition/default-partitioner.go
@@ -36,8 +36,8 @@ import (
 )
 
 const (
-	defaultPartitionConfigWorkerIsolationGroup = "worker-isolation-group"
-	defaultPartitionConfigRunID                = "wf-run-id"
+	IsolationGroupKey = "isolation-group"
+	WorkflowRunIDKey  = "wf-run-id"
 )
 
 // DefaultWorkflowPartitionConfig Is the default dataset expected to be passed around in the
@@ -80,8 +80,8 @@ func (r *defaultPartitioner) GetIsolationGroupByDomainID(ctx context.Context, do
 }
 
 func mapPartitionConfigToDefaultPartitionConfig(config PartitionConfig) defaultWorkflowPartitionConfig {
-	isolationGroup, _ := config[defaultPartitionConfigWorkerIsolationGroup]
-	runID, _ := config[defaultPartitionConfigRunID]
+	isolationGroup, _ := config[IsolationGroupKey]
+	runID, _ := config[WorkflowRunIDKey]
 	return defaultWorkflowPartitionConfig{
 		WorkflowStartIsolationGroup: isolationGroup,
 		RunID:                       runID,
--- a/common/rpc/middleware.go
+++ b/common/rpc/middleware.go
@@ -22,11 +22,13 @@ package rpc
 
 import (
 	"context"
+	"encoding/json"
 	"io"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/config"
 	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/partition"
 
 	"go.uber.org/cadence/worker"
 	"go.uber.org/yarpc"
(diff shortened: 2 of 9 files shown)


common/rpc/middleware.go
```diff
@@ -177,3 +178,15 @@ func (m *HeaderForwardingMiddleware) Call(ctx context.Context, request *transpor
 
 	return out.Call(ctx, request)
 }
+
+// DefaultPartitionConfigMiddleware stores the partition config and isolation group of the request into the context
+type DefaultPartitionConfigMiddleware struct{}
+
+func (m *DefaultPartitionConfigMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, h transport.UnaryHandler) error {
+	isolationGroup, _ := req.Headers.Get("isolation-group")
```
davidporter-id-au, line 186: maybe make this a constant

common/rpc/middleware.go
```diff
@@ -177,3 +178,15 @@ func (m *HeaderForwardingMiddleware) Call(ctx context.Context, request *transpor
 
 	return out.Call(ctx, request)
 }
+
+// DefaultPartitionConfigMiddleware stores the partition config and isolation group of the request into the context
+type DefaultPartitionConfigMiddleware struct{}
+
+func (m *DefaultPartitionConfigMiddleware) Handle(ctx context.Context, req *transport.Request, resw transport.ResponseWriter, h transport.UnaryHandler) error {
```
davidporter-id-au, line 185: Should this have a guard on config to determine if the isolation-group-feature is enabled? elsewhere - unless I'm confused, it looked like you were using the presence of the poller Isolation-group to determine if the feature is enabled.

Shaddoll: this is NOT the most suitable place to do the check. 
1. We only get the request which is not deserialized and are not able to get the domain name.
2. It's hard to introduce a dependency to middleware because cadence is not an Fx application.

Shaddoll: The best place is `workflowHandler.go`. We can add a check, but I don't think it's necessary. 
It can only save a few bytes storage per workflow for domains not having this feature enabled.

davidporter-id-au: yeah, I think I wrote this message before we talked and I was just confused, what you have seems reasonable

common/partition/default-partitioner.go
```diff
@@ -36,8 +36,8 @@ import (
 )
 
 const (
-	defaultPartitionConfigWorkerIsolationGroup = "worker-isolation-group"
-	defaultPartitionConfigRunID                = "wf-run-id"
+	IsolationGroupKey = "isolation-group"
+	WorkflowRunIDKey  = "wf-run-id"
```
emrahs, line 40: Why don't we use workflow ID instead of run ID? Workflow ID would give a more consistent experience across resets, continueAsNews, etc., which would make everything easier to reason about from user's perspective.

davidporter-id-au (approved):

common/partition/default-partitioner.go
davidporter-id-au: Fair, Let me follow that up in a different diff

PR #5201, opened 2023-04-10
Add Pinot as new visibility option

neil-xie:
**What changed?**
Updated server start up to choose visibility manager based on the advanced visibility option
Added new triple visibility manager to provide options to write to both ES and Pinot for future migration
Added pinotVisibilityStore and pinotClient to support storing visibility messages in Pinot
Added util methods to flatten the search attributes into columns
Added util methods to validate pinot response and query
Added new kafka indexer message type for Pinot since Pinot doesn't need indexer to process
Added integration test to set up Pinot test cluster and test Pinot functionality
image

**Why?**
ElasticSearch is going to be deprecated with Uber, we will migrate from ElasticSearch to Pinot. This is the first step to make Cadence work with Pinot as the advanced visibility store. In the future, we will add support for more components like migration between ES and Pinot.

**How did you test it?**
Added unit test for Pinot visibility store and pinot client. 
Added unit test for Pinot response comparator
Added integration test for Pinot to test the functionality
Bench test in staging environment to test performance.

**Potential risks**
Pinot doesn't support deletion yet, the storage may continue growing without segment compaction.
We have the pinot/es response comparator to help detect data issue during migration

**Release notes**

**Documentation Changes**


cmd/server/cadence/server.go
```diff
@@ -220,7 +223,20 @@ func (s *server) startService() common.Daemon {
 			log.Fatalf("not able to find advanced visibility store in config: %v", advancedVisStoreKey)
 		}
 
-		params.ESConfig = advancedVisStore.ElasticSearch
+		if params.PersistenceConfig.AdvancedVisibilityStore == "pinot-visibility" {
```
shijiesheng, line 226: use `PinotVisibilityAppName`

bowenxia: I thought the plugins (like what mySQL/Cassandra/mongoDB/postgre have) were not for visibility stores? Because we don't have the plugin files for ElasticSearch, and we don't need to register ElasticSearch as a plugin.

neil-xie: You are right, it is better to use constants here or it can mess up since it is used for multiple places. Updated.

common/persistence/pinotVisibilityTripleManager.go
bowenxia, line 1: The purpose of this is only for the migration reason. We are migrating from ES to Pinot.

common/persistence/pinotResponseComparator.go
bowenxia, line 1: This comparator is still in progress. Will possibly redesign and reimplement it.

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -397,23 +397,31 @@ func (v *pinotVisibilityTripleManager) CountWorkflowExecutions(
...
+		//	v.logger.Warn("Enable dual read mode failed! Will use ES visibility manager. ")
+		//	visibilityMgr = v.esVisibilityManager
+		//} else if v.pinotVisibilityManager != nil {
+		//	v.logger.Warn("Enable dual read mode failed! Will use Pinot visibility manager. ")
+		//	visibilityMgr = v.pinotVisibilityManager
+		//} else {
+		//	v.logger.Warn("Enable dual read mode failed! Will use db visibility manager. ")
+		//	visibilityMgr = v.dbVisibilityManager
+		//}
+		// TODO: Turn off Comparator for now, because we need to redesign it.
```
bowenxia, line 418: Turn off Comparator for now, because we need to redesign it

cmd/server/cadence/server.go
```diff
@@ -220,7 +223,20 @@ func (s *server) startService() common.Daemon {
 			log.Fatalf("not able to find advanced visibility store in config: %v", advancedVisStoreKey)
 		}
 
-		params.ESConfig = advancedVisStore.ElasticSearch
+		if params.PersistenceConfig.AdvancedVisibilityStore == common.PinotVisibilityStoreName {
+			params.PinotConfig = advancedVisStore.Pinot
+			esVisStore := s.cfg.Persistence.DataStores[common.ESVisibilityStoreName]
```
davidporter-id-au, line 228: why would this be defined?

neil-xie: We need esVisStore to write to ES, we have components like ESAnalyzer is using ElasticSearch. The plan is to clean those after we switch to operate on Pinot.

common/persistence/Pinot/pinotVisibilityStore.go
```diff
@@ -0,0 +1,1111 @@
...
+		pinotClient:         pinotClient,
+		producer:            producer,
+		logger:              logger.WithTags(tag.ComponentPinotVisibilityManager),
+		config:              config,
+		pinotQueryValidator: pnt.NewPinotQueryValidator(config.ValidSearchAttributes()),
+	}
+}
+
+func (v *pinotVisibilityStore) Close() {
+	// TODO: need to double check what is close trace do. Does it close the client?
```
shijiesheng, line 120: from document it's used for dev. so we don't need to do anything here https://docs.pinot.apache.org/users/api/querying-pinot-using-standard-sql/response-format

common/persistence/Pinot/pinotVisibilityStore.go
```diff
@@ -0,0 +1,1111 @@
...
+
+func (v *pinotVisibilityStore) GetName() string {
+	return pinotPersistenceName
+}
+
+func (v *pinotVisibilityStore) RecordWorkflowExecutionStarted(
+	ctx context.Context,
+	request *p.InternalRecordWorkflowExecutionStartedRequest,
+) error {
+	v.checkProducer()
```
shijiesheng, line 131: move this check at creation `NewPinotVisibilityStore` to avoid runtime panic. Use log.Fatal instead.

common/pinot/pinotQueryValidator.go
```diff
@@ -0,0 +1,271 @@
...
+			"AND REGEXP_LIKE(JSON_EXTRACT_SCALAR(Attr, '$.%s', 'string'), '%s*'))", colNameStr, colNameStr, colValStr), nil
+	}
+	// case2-2: otherwise, exact match
+	// case2-2-1: if it is keyword, need to deal with a situation when value is an array
+	if indexValType == types.IndexedValueTypeKeyword {
+		return fmt.Sprintf("(JSON_MATCH(Attr, '\"$.%s\"=''%s''') or JSON_MATCH(Attr, '\"$.%s[*]\"=''%s'''))",
+			colNameStr, colValStr, colNameStr, colValStr), nil
+	}
+	// case2-2-2: other cases:
+	return fmt.Sprintf("JSON_MATCH(Attr, '\"$.%s\"=''%s''')", colNameStr, colValStr), nil
```
shijiesheng, line 270: Do we support Datetime customer key range query? From this, it doesn't see so.

common/persistence/Pinot/pinotVisibilityStore.go
```diff
@@ -0,0 +1,1111 @@
...
+	m[CloseStatus] = int(closeStatus)
+	m[HistoryLength] = historyLength
+	m[UpdateTime] = updateTimeUnixNano
+	m[ShardID] = shardID
+	m[IsDeleted] = isDeleted
+
+	SearchAttributes := make(map[string]interface{})
+	var err error
+	for key, value := range rawSearchAttributes {
+		value, err = isTimeStruct(value)
```
shijiesheng, line 597: i'm not familiar with pinot. Does it not support time.Time type?

common/pinot/pinotQueryValidator_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/uber/cadence/common/definition"
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+func TestValidateQuery(t *testing.T) {
+	tests := map[string]struct {
```
shijiesheng, line 35: Can we add test cases on `CustomDatetimeField` fields? ("=" ">" "<" and "order by" cases)

common/pinot/responseUtility.go
```diff
@@ -0,0 +1,131 @@
...
+	CloseTime     int64
+	CloseStatus   int
+	HistoryLength int64
+	Encoding      string
+	TaskList      string
+	IsCron        bool
+	NumClusters   int16
+	UpdateTime    int64
+	ShardID       int16
+	//SearchAttributes map[string]interface{}
```
shijiesheng, line 66: remove?

common/pinot/responseUtility.go
```diff
@@ -0,0 +1,131 @@
...
+	if len(hit) != len(columnNames) {
+		return nil
+	}
+
+	systemKeyMap := buildMap(hit, columnNames)
+
+	jsonSystemKeyMap, err := json.Marshal(systemKeyMap)
+	if err != nil {
+		logger.Error("unable to marshal systemKeyMap",
+			tag.Error(err), //tag.ESDocID(fmt.Sprintf(columnNameToValue["DocID"]))
```
shijiesheng, line 79: remove comment?

common/persistence/Pinot/pinotVisibilityStore.go
neil-xie: Thanks! I will update the comment here. Cannot remove the method since it implements VisibilityStore interface

common/persistence/Pinot/pinotVisibilityStore.go
```diff
@@ -0,0 +1,1111 @@
...
+type (
+	pinotVisibilityStore struct {
+		pinotClient         pnt.GenericClient
+		producer            messaging.Producer
+		logger              log.Logger
+		config              *service.Config
+		pinotQueryValidator *pnt.VisibilityQueryValidator
+	}
+
+	visibilityMessage struct {
```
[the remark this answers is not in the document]
neil-xie: Removed

common/persistence/Pinot/pinotVisibilityStore.go
neil-xie: From the [official doc](https://docs.pinot.apache.org/basics/components/table/schema#date-and-time-fields), it is not supported

> Since Pinot doesn't have a dedicated DATETIME datatype support, you need to input time in either STRING, LONG, or INT format. However, Pinot needs to convert the date into an understandable format such as epoch timestamp to do operations.

common/pinot/pinotQueryValidator.go
neil-xie: For the customized search attribtues, we don't support range queries even for integer columns. @bowenxia will test next week to see if we can use similar SQL queries to partial match query we used. [CDNC-5946](https://t3.uberinternal.com/browse/CDNC-5946)

common/persistence/Pinot/pinotVisibilityStore.go
neil-xie: Fixed

common/persistence/Pinot/pinotVisibilityMetricClients.go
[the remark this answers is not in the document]
neil-xie: Seems the way to rename is creating new directory and move the files there. I will do it at the end so I won't lose the comments here

common/pinot/pinotQueryValidator.go
neil-xie: Update: I tested range queries and it worked. Need to update the query validator and query builder to support that. Bowen will add in a following diff.

common/persistence/Pinot/pinotVisibilityMetricClients.go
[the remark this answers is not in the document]
neil-xie: Renamed

common/pinot/pinotQueryValidator_test.go
neil-xie: Same here, will add in a following diff of [CDNC-5946](https://t3.uberinternal.com/browse/CDNC-5946)

common/config/pinot.go
```diff
@@ -0,0 +1,31 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
+
+// PinotVisibilityConfig for connecting to Pinot
```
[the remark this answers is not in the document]
neil-xie: All the configs related to yaml structure is in the config folder here, so I will keep it here

common/persistence/Pinot/pinotVisibilityMetricClients.go
```diff
@@ -0,0 +1,389 @@
...
+	request *p.RecordWorkflowExecutionStartedRequest,
+) error {
+
+	scopeWithDomainTag := p.metricClient.Scope(metrics.PinotRecordWorkflowExecutionStartedScope, metrics.DomainTag(request.Domain))
+	scopeWithDomainTag.IncCounter(metrics.PinotRequestsPerDomain)
+
+	sw := scopeWithDomainTag.StartTimer(metrics.PinotLatencyPerDomain)
+
+	err := p.persistence.RecordWorkflowExecutionStarted(ctx, request)
+	sw.Stop()
```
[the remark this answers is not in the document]
neil-xie: Fixed

common/constants.go
```diff
@@ -85,7 +85,14 @@ const (
 
 const (
 	// VisibilityAppName is used to find kafka topics and ES indexName for visibility
-	VisibilityAppName = "visibility"
+	VisibilityAppName      = "visibility"
+	PinotVisibilityAppName = "pinot-visibility"
+)
+
+const (
+	// VisibilityStoreName is used to find advanced visibility store
```
shijiesheng, line 93: nit: ESVisibilityStoreName

go.mod
```diff
@@ -34,7 +34,8 @@ require (
 	github.com/otiai10/copy v1.1.1
 	github.com/pborman/uuid v0.0.0-20180906182336-adf5a7427709
 	github.com/robfig/cron v1.2.0
-	github.com/sirupsen/logrus v1.8.1
+	github.com/sirupsen/logrus v1.9.0
+	github.com/startreedata/pinot-client-go v0.0.0-20230303070132-3b84c28a9e95
```
shijiesheng, line 38: nit: maybe a comment on why this particular version is needed

shijiesheng (approved):

go.mod
neil-xie: Sure, their last release (0.1.0) was from 2022, so we used master here.

common/pinot/pinotQueryValidator.go
```diff
@@ -0,0 +1,271 @@
...
+// NewPinotQueryValidator create VisibilityQueryValidator
+func NewPinotQueryValidator(validSearchAttributes map[string]interface{}) *VisibilityQueryValidator {
+	return &VisibilityQueryValidator{
+		validSearchAttributes: validSearchAttributes,
+	}
+}
+
+// ValidateQuery validates that search attributes in the query and returns modified query.
+func (qv *VisibilityQueryValidator) ValidateQuery(whereClause string) (string, error) {
+	if len(whereClause) != 0 {
```
3vilhamster, line 53: nit: use short return.
```
if len(whereClause) == 0 {
   return whereClause, nil
}
```
that will make this function easier to read.

common/pinot/pinotQueryValidator.go
```diff
@@ -0,0 +1,271 @@
...
+
+// isValidSearchAttributes return true if key is registered
+func (qv *VisibilityQueryValidator) isValidSearchAttributes(key string) bool {
+	validAttr := qv.validSearchAttributes
+	_, isValidKey := validAttr[key]
+	return isValidKey
+}
+
+func (qv *VisibilityQueryValidator) processSystemKey(expr sqlparser.Expr) (string, error) {
+	comparisonExpr := expr.(*sqlparser.ComparisonExpr)
```
3vilhamster, line 177: nit: for all methods that expect a single type, it would be easier to use it directly instead of asserting to the interface.

3vilhamster (commented):
some nit comments from looking at this file later on. Might be worth improving

PR #5203, opened 2023-04-12
Corrects the config-store handling for not-found errors

davidporter-id-au:
**What changed?**
- Swaps the config-store to use the DB plugin checker for not-found entries rather than doing a direct error comparison. This makes it compatible with internal forks of goCQL

**Why?**

**How did you test it?**
Unit-tests, manual-tests. However, given this error it's probably not been tested extensively. 

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/configStore.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/configStore.go
@@ -56,9 +56,6 @@ func (db *cdb) SelectLatestConfig(ctx context.Context, rowType int) (*persistenc
 
 	query := db.session.Query(templateSelectLatestConfig, rowType).WithContext(ctx)
 	err := query.Scan(&rowType, &version, &timestamp, &data, &encoding)
-	if db.IsNotFoundError(err) {
-		return nil, nil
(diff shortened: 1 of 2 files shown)


common/persistence/nosql/nosqlplugin/cassandra/configStore.go
```diff
@@ -58,7 +56,7 @@ func (db *cdb) SelectLatestConfig(ctx context.Context, rowType int) (*persistenc
 
 	query := db.session.Query(templateSelectLatestConfig, rowType).WithContext(ctx)
 	err := query.Scan(&rowType, &version, &timestamp, &data, &encoding)
-	if err == gocql.ErrNotFound {
+	if db.IsNotFoundError(err) {
```
Shaddoll, line 59: I think we should move the logic to make it work for all nosqlplugins

davidporter-id-au: happy to do that, but I don't quite understand

Shaddoll (approved):

PR #5204, opened 2023-04-12
Do Not Merge: Adding TTL to the workflow Execution

agautam478:
**What changed?**
The Workflow Executions will have a time to live attached to them when they get created. This means that after the execution is over and it's retention period is over it will be removed from the Execution table.

TTL value = workflow start-to-close timeout + domain retention (or, we could just pick 30 days here, there's no reason to get the exact retention, just an upper bound is fine) + 15 days of leeway

**Why?**
These changes have been made as part of clean up process so that the partition size doesn't become bloated after the executions have served their purpose.

**How did you test it?**
Testing locally and added a test case.

**Potential risks**
In a worse case scenario, executions that are not supposed to be removed can be deleted permanently.

**Release notes**
NA

**Documentation Changes**
NA


service/history/execution/context.go
```diff
@@ -45,6 +45,7 @@ import (
 
 const (
 	defaultRemoteCallTimeout = 30 * time.Second
+	buffertimeforTTL         = 15
```
davidporter-id-au, line 48: Would you mind changing this to be a value which indicates the value it's on, I assume it's days? 

`ttlBufferDays` might be one option?

common/persistence/dataManagerInterfaces.go
```diff
@@ -872,6 +872,8 @@ type (
 		NewWorkflowSnapshot WorkflowSnapshot
 
 		DomainName string
+
+		TTL int64
```
Shaddoll, line 876: nit: we need a better name for this. `TTLInSeconds` is a bit better

common/dynamicconfig/constants.go
```diff
@@ -3871,6 +3879,11 @@ var BoolKeys = map[BoolKey]DynamicBool{
 		Description:  "Enable shardId metrics in persistence client",
 		DefaultValue: true,
 	},
+	EnableTTL: DynamicBool{
```
Shaddoll, line 3882: We need a better name. e.g. EnableExecutionTTL indicating that it's for execution table

common/dynamicconfig/constants.go
```diff
@@ -3871,6 +3879,11 @@ var BoolKeys = map[BoolKey]DynamicBool{
...
+		KeyName:      "enableTTL",
```
Shaddoll, line 3883: the key name should follow the existing naming pattern.
We can name it as `system.enableExecutionTTL`

common/dynamicconfig/constants.go
```diff
@@ -3871,6 +3879,11 @@ var BoolKeys = map[BoolKey]DynamicBool{
...
+		Description:  "EnableTTL is which domains are allowed to have workflow executions with a TTL",
+		DefaultValue: true,
```
Shaddoll, line 3885: DefaultValue should be false

agautam478:
Closing the PR as the CreateWorkflow flow is not the correct place to update the TTL. It keeps on become zero as the workflow data gets updated.
New PR: https://github.com/uber/cadence/pull/5243

PR #5209, opened 2023-04-13
Feature/implementing domain drains

davidporter-id-au:
Refactors the frontend Workflow and Admin handlers to take DomainHandler as a common parameter rather than doing the weird self-pointing thing they were doing in their constructors previously. 

Refactors the isolation-group implementation by splitting the single package into three parts: 

- The base `isolationgroup` package which just specifies the state interface and common types
- A default implementation of the state package as a subpackage so it can be split up a bit into smaller files
- A handler subpackage which is responsible for doing CRUD operations for frontend and mapping.

Refactors config store:
- Compose the dynamicConfig and Daemon into a unified interface for the config store to represent the fact that it's a stateful implementation of the dynamic config client.
- To remove the `done` channel and instead use a daemon interface
- Adds requisite Start/Stop functions

Refactors the unit-test for adminHandler_test.go 
- Mocks out the isolationgroup handler rather than mocking below


service/frontend/adminHandler.go
```diff
@@ -142,44 +145,46 @@ var (
 	}
 )
 
-// NewAdminHandler creates a thrift handler for the cadence admin service
+// NewAdminHandler creates a thrift service for the cadence admin service
 func NewAdminHandler(
-	resource resource.Resource,
+	service Service,
```
davidporter-id-au, line 150: I don't like this much, passing itself as a parameter, but I don't really want to make this a huge refactor and this isn't really significantly changing much

common/resource/resourceImpl.go
```diff
@@ -563,28 +577,71 @@ func (h *Impl) GetDispatcher() *yarpc.Dispatcher {
...
+	}
+	return cfgStoreClient
+}
+
 // Use the provided IsolationGroupStateHandler or the default one
-func ensureIsolationGroupStateHandlerOrDefault(
+// due to the config store being only available for some
+// persistence layers, *both* the configStoreClient and IsolationGroupState
+// will be optionally available
+func createIsolationGroupStateHandlerOrDefault(
```
davidporter-id-au, line 624: changing the name to `create...` to indicate that it's not guaranteed to not be nil. `ensure` has, in my experience mostly been used to do the opposite - to guarantee non-nil-ness

common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -342,6 +351,19 @@ func (csc *configStoreClient) ListValue(name dc.Key) ([]*types.DynamicConfigEntr
...
+	if !atomic.CompareAndSwapInt32(&csc.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
+		return
+	}
+	close(csc.doneCh)
+}
+
+func (csc *configStoreClient) Start() {
+	if !atomic.CompareAndSwapInt32(&csc.status, common.DaemonStatusInitialized, common.DaemonStatusStarted) {
+		return
+	}
```
Shaddoll, line 364: nit: it seems better if we move `startUpdate` call from `new` method to `Start`

common/isolationgroup/defaultisolationgroupstate/state.go
```diff
@@ -20,73 +20,41 @@
...
-	AllIsolationGroups []string
-}
-
 type defaultIsolationGroupStateHandler struct {
 	status                     int32
 	done                       chan struct{}
 	log                        log.Logger
 	domainCache                cache.DomainCache
 	globalIsolationGroupDrains dynamicconfig.Client
+	domainManager              persistence.DomainManager
```
Shaddoll, line 49: Do we need this field?

davidporter-id-au: no, this is obviated, good catch

common/resource/resourceImpl.go
```diff
@@ -563,28 +577,71 @@ func (h *Impl) GetDispatcher() *yarpc.Dispatcher {
...
+	}
+	cscConfig := &csc.ClientConfig{
+		PollInterval:        dc.GetDurationProperty(dynamicconfig.IsolationGroupStateRefreshInterval)(),
+		UpdateRetryAttempts: dc.GetIntProperty(dynamicconfig.IsolationGroupStateUpdateRetryAttempts)(),
+		FetchTimeout:        dc.GetDurationProperty(dynamicconfig.IsolationGroupStateFetchTimeout)(),
+		UpdateTimeout:       dc.GetDurationProperty(dynamicconfig.IsolationGroupStateUpdateTimeout)(),
+	}
+	cfgStoreClient, err := configstore.NewConfigStoreClient(cscConfig, &params.PersistenceConfig, params.Logger, persistence.GlobalIsolationGroupConfig)
+	if err != nil {
+		// not possible to create the client under some persistence configurations, so this is expected
```
Shaddoll, line 614: nit: add a log

service/frontend/adminHandler_test.go
```diff
@@ -109,7 +110,11 @@ func (s *adminHandlerSuite) SetupTest() {
 		EnableAdminProtection:  dynamicconfig.GetBoolPropertyFn(false),
 		EnableGracefulFailover: dynamicconfig.GetBoolPropertyFn(false),
 	}
-	s.handler = NewAdminHandler(s.mockResource, params, config).(*adminHandlerImpl)
+
+	ctrl := gomock.NewController(s.T())
+	dh := domain.NewMockHandler(ctrl)
```
Shaddoll, line 115: the parameter should be `s.controller`

service/frontend/workflowHandler_test.go
```diff
@@ -119,6 +120,19 @@ func (s *workflowHandlerSuite) SetupTest() {
...
 	s.mockVersionChecker = client.NewMockVersionChecker(s.controller)
 
+	// these tests don't mock the domain handler
+	config := s.newConfig(dc.NewInMemoryClient())
+	s.domainHandler = domain.NewHandler(
+		config.domainConfig,
+		s.mockResource.GetLogger(),
+		s.mockResource.GetDomainManager(),
+		s.mockResource.GetClusterMetadata(),
+		domain.NewDomainReplicator(nil, s.mockResource.GetLogger()),
```
Shaddoll, line 130: nil should be `s.mockProducer` and then you don't need the code change in `TestUpdateDomain_Success_FailOver`

common/isolationgroup/isolationgroupapi/handler.go
```diff
@@ -0,0 +1,43 @@
+package isolationgroupapi
+
+import (
+	"github.com/uber/cadence/common/domain"
+	"github.com/uber/cadence/common/dynamicconfig"
+	"github.com/uber/cadence/common/log"
+)
+
+// The MIT License (MIT)
```
Shaddoll, line 9: nit: license should be moved to the top

common/isolationgroup/isolationgroupapi/interface.go
```diff
@@ -0,0 +1,42 @@
...
+
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination isolation_handler_mock.go -self_package github.com/uber/cadence/common/isolationgroup/isolationgrouphandler
+
+type Handler interface {
+	// GetGlobalState returns the current configuration of isolation-groups which apply to all domains
+	GetGlobalState(ctx context.Context) (*types.GetGlobalIsolationGroupsResponse, error)
+	// UpdateGlobalState updates isolation-groups which apply to all domains
+	UpdateGlobalState(ctx context.Context, state types.UpdateGlobalIsolationGroupsRequest) error
+	// GetDomainState is the read operation for getting the current state of a domain's isolation-groups
+	GetDomainState(ctx context.Context, request types.GetDomainIsolationGroupsRequest) (*types.GetDomainIsolationGroupsResponse, error)
```
Shaddoll, line 39: It's a bit weird to pass the parameter by value not by pointer

service/frontend/adminHandler_test.go
```diff
@@ -975,10 +900,10 @@ func Test_UpdateGlobalIsolationGroups(t *testing.T) {
...
 		Resource: &resource.Test{
-			Logger:          loggerimpl.NewNopLogger(),
-			IsolationGroups: nil, // valid state, the isolation-groups feature is not available for all persistence types
-			MetricsClient:   metrics.NewNoopMetricsClient(),
+			Logger:        loggerimpl.NewNopLogger(),
+			MetricsClient: metrics.NewNoopMetricsClient(),
 		},
+		isolationGroups: nil, // valid state, the isolation-groups feature is not available for all persistence types
 	}
 	assert.NotPanics(t, func() {
```
Shaddoll, line 908: it's probably better to check the error type, which implies that it doesn't panic

common/isolationgroup/isolationgroupapi/global-api-handlers_test.go
```diff
@@ -0,0 +1,137 @@
...
+	dynamicConfigResponse := []interface{}{}
+	json.Unmarshal(validCfgData, &dynamicConfigResponse)
+
+	tests := map[string]struct {
+		in           types.GetGlobalIsolationGroupsRequest
+		dcAffordance func(client *dynamicconfig.MockClient)
+		expected     *types.GetGlobalIsolationGroupsResponse
+		expectedErr  error
+	}{
+		"updating normal value": {
```
Shaddoll, line 99: nit: the message needs to be updated

common/isolationgroup/isolationgroupapi/global-api-handlers_test.go
```diff
@@ -0,0 +1,137 @@
...
+				"zone-1": {Name: "zone-1", State: types.IsolationGroupStateHealthy},
+				"zone-2": {Name: "zone-2", State: types.IsolationGroupStateDrained},
+			}},
+			dcAffordance: func(client *dynamicconfig.MockClient) {
+				client.EXPECT().UpdateValue(
+					dynamicconfig.DefaultIsolationGroupConfigStoreManagerGlobalMapping,
+					gomock.Any(), // covering the mapping in the mapper unit-test instead
+				)
+			},
+		},
```
Shaddoll, line 57: nit: an error case

service/frontend/workflowHandler_test.go
davidporter-id-au: I do for a different reason, but good point, thank you

service/frontend/adminHandler_test.go
davidporter-id-au: 🤦

common/isolationgroup/isolationgroupapi/interface.go
davidporter-id-au: this is a [go/glue convention](https://sourcegraph.uberinternal.com/code.uber.internal/go-code/-/blob/src/code.uber.internal/crack/luna-gateway/controller/benefits/benefits.go#L27) which isn't mine: Pass in values, return pointers to denote optionality, handle mapping, error handling, authorization at the outer layer and pass in validated values for the inner business logic. I don't feel strongly about it though, and in this instance it doesn't make a lot of difference. 

Obviously it's slightly more memory usage, but for operations like this it hardly matters.

common/isolationgroup/isolationgroupapi/global-api-handlers_test.go
davidporter-id-au: not sure I understand?

common/isolationgroup/isolationgroupapi/global-api-handlers_test.go
davidporter-id-au: tricky... I am not sure I can trigger an error case given the type system and the only thing that can trigger an error is the json serialization

Shaddoll (approved):

PR #5211, opened 2023-04-14
Changes the config store to have a daemon

davidporter-id-au:
- Removes channel from constructor for closing 
- Changes config store constructor to emphasize that this is a daemon


common/dynamicconfig/configstore/config_store_client.go
```diff
@@ -342,6 +351,19 @@ func (csc *configStoreClient) ListValue(name dc.Key) ([]*types.DynamicConfigEntr
...
 }
 
+func (csc *configStoreClient) Stop() {
+	if !atomic.CompareAndSwapInt32(&csc.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
+		return
+	}
+	close(csc.doneCh)
+}
+
+func (csc *configStoreClient) Start() {
```
Shaddoll, line 361: where is it being called?

davidporter-id-au: it's not being called

Shaddoll: nobody is calling start and stop, and the goroutine of updating cache is leaked

davidporter-id-au: I should have provided better context. I'm pulling the store out into it's own field in resource and stop is being called on resource stop, similar to the isolation-group.

Shaddoll: you should make it in the same diff

davidporter-id-au: I'll leave it in the larger refactoring diff then

PR #5216, opened 2023-04-17
Spread task backlog to available pollers

Shaddoll:
**What changed?**
If isolation is enabled and a task is allocated to an isolation group with no poller, we re-allocate it to an isolation group with poller.

**Why?**
If an isolation group doesn't have any poller, async task processing is blocked. In that case, we need to allocate those tasks to pollers available. 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


service/matching/pollerHistory.go
```diff
@@ -91,13 +89,39 @@ func (pollers *pollerHistory) getPollerInfo(earliestAccessTime time.Time) []*typ
...
-				RatePerSecond:  value.ratePerSecond,
+				RatePerSecond:  rps,
 			})
 		}
 	}
 
 	return result
 }
+
+func (pollers *pollerHistory) getPollerIsolaionGroups(earliestAccessTime time.Time) []string {
```
davidporter-id-au, line 107: nit: typo: 

```
getPollerIsolationGroups
```

service/matching/taskListManager.go
```diff
@@ -573,6 +566,42 @@ func (c *taskListManagerImpl) shouldReload() bool {
...
+			for _, pollerGroup := range pollerIsolationGroups {
+				if pollerGroup == isolationGroup {
+					return isolationGroup
+				}
+			}
+			c.logger.Warn("Failed to find a poller from the given isolation group", tag.WorkflowID(taskInfo.WorkflowID), tag.WorkflowRunID(taskInfo.RunID), tag.TaskID(taskInfo.TaskID))
+		}
+		// if we hit this branch, either we failed to get isolation group from the partition library, or we don't have pollers from the given isolation group
+		// use runID to deterministically choose an isolation group with pollers
+		hashv := farm.Hash32([]byte(taskInfo.RunID))
```
davidporter-id-au, line 599: I do think we should pull this into either the partition library or otherwise ensure there's a single place where we're doing partitioning. Spreading the logic around is I think is risky when you accidentally change one but not the other.

service/matching/taskListManager.go
```diff
@@ -573,6 +566,42 @@ func (c *taskListManagerImpl) shouldReload() bool {
...
+		}
+		partitionConfig[partition.WorkflowRunIDKey] = taskInfo.RunID
+		var isolationGroup string
+		group, err := c.partitioner.GetIsolationGroupByDomainID(ctx, taskInfo.DomainID, partitionConfig)
+		if err != nil {
+			c.logger.Error("Failed to get isolation group from partition library", tag.WorkflowID(taskInfo.WorkflowID), tag.WorkflowRunID(taskInfo.RunID), tag.TaskID(taskInfo.TaskID), tag.Error(err))
+		} else {
+			isolationGroup = group
+		}
+		// check if the isolation group has pollers, if not use the fallback logic
```
davidporter-id-au, line 583: I think we want to guard this fallback behaviour with a dynamic-config guard with domain filter, I can imagine there'll be scenarios were we do and don't want this behaviour.

Shaddoll: We have to fallback, and there is no other choice, because if we don't, async task processing is totally blocked.

service/matching/taskReader.go
```diff
@@ -134,28 +131,27 @@ dispatchLoop:
...
+				cancel()
 				if err == nil {
 					break
 				}
 				if err == context.Canceled {
 					tr.logger.Info("Tasklist manager context is cancelled, shutting down")
 					break dispatchLoop
 				}
+				if err == context.DeadlineExceeded {
+					// it only happens when isolation is enabled and there is no pollers from the given isolation group
```
davidporter-id-au, line 149: Can we put a metric and log here?

service/matching/taskReader.go
```diff
@@ -134,28 +131,27 @@ dispatchLoop:
...
-				group, err := tr.partitioner.GetIsolationGroupByDomainID(ctx, taskInfo.DomainID, taskInfo.PartitionConfig)
-				if err != nil {
-					tr.logger.Error("Failed to get isolation group for async match task", tag.TaskID(taskInfo.TaskID), tag.Error(err))
-				} else if group != "" {
-					isolationGroup = group
-				}
-				cancel()
-			}
-			task := newInternalTask(taskInfo, tr.completeTask, types.TaskSourceDbBacklog, "", false, nil, isolationGroup)
 			for {
```
davidporter-id-au, line 134: Question: I don't quite understand this, what's the second for-loop for?

davidporter-id-au (approved):

service/matching/taskListManager.go
davidporter-id-au: I keep forgetting 🤦

service/matching/taskListManager.go
```diff
@@ -573,6 +566,34 @@ func (c *taskListManagerImpl) shouldReload() bool {
 	return c.config.EnableTasklistIsolation() != c.enableIsolation && c.taskListKind != types.TaskListKindSticky
 }
 
+func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, taskInfo *persistence.TaskInfo) string {
+	if c.isIsolationEnabled() && len(taskInfo.PartitionConfig) > 0 {
```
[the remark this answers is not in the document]
Shaddoll: I prefer to delete that type.

service/matching/config.go
```diff
@@ -95,13 +96,15 @@ type (
...
 		MaxTaskDeleteBatchSize        func() int
+		AsyncTaskDispatchTimeout      func() time.Duration
 		// taskWriter configuration
 		OutstandingTaskAppendsThreshold func() int
 		MaxTaskBatchSize                func() int
 		NumWritePartitions              func() int
 		NumReadPartitions               func() int
 		// isolation configuration
 		EnableTasklistIsolation func() bool
+		AllIsolationGroups      []string
```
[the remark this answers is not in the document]
Shaddoll: slice is actually a pointer to array

service/matching/taskReader.go
Shaddoll: keep trying to dispatch a single task to pollers

davidporter-id-au (approved):

service/matching/taskListManager.go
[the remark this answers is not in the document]
davidporter-id-au: ok, let's do that

PR #5218, opened 2023-04-19
Add tasklist traffic metrics for non-sticky and non-forwarded tasklists

timl3136:
**What changed?**
Add a new emit metrics to track the rate of task being added into non-sticky and non-forwarded tasklists.

**Why?**
This metric allow us to better monitor tasklist and determine whether we need to partition tasklist if the tasklist becomes too long.

**How did you test it?**
Test both on existing local unit tests and on staging env.

**Potential risks**
Got blocked due to reaching series limit on emitted data.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -28,8 +28,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"strconv"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Shaddoll (commented):
"This metric allow us to better monitor tasklist and determine whether we need to partition tasklist if the tasklist becomes too long."

If this is the goal, this metric is not necessary. `task_backlog_per_tl` and `task_lag_per_tl` is enough to tell us if a tasklist has too many tasks

service/matching/matchingEngine.go
```diff
@@ -351,6 +366,21 @@ func (e *matchingEngineImpl) AddActivityTask(
 		return false, err
 	}
 
+	// get the domainName
+	domainName, err := e.domainCache.GetDomainName(domainID)
+	if err != nil {
+		return false, err
+	}
+
+	// tell if the tasklist is partitioned or sticky
```
neil-xie, line 375: could you add one more line of comment here to say the purpose so other people can know why you emit the metrics here, just like the summary you added for the pr

service/matching/matchingEngine.go
```diff
@@ -296,6 +296,21 @@ func (e *matchingEngineImpl) AddDecisionTask(
...
+	if int32(request.GetTaskList().GetKind()) == 0 && request.ForwardedFrom == "" {
```
neil-xie, line 306: taskListKind was retrieved at the beginning of the method, you can just use it

allenchen2244:
> "This metric allow us to better monitor tasklist and determine whether we need to partition tasklist if the tasklist becomes too long."
> 
> If this is the goal, this metric is not necessary. `task_backlog_per_tl` and `task_lag_per_tl` is enough to tell us if a tasklist has too many tasks

There's a few other things that we are gonna need this for. Such as creating a client dashboard that shows tasklist traffic percentage which I don't think we have.

service/matching/matchingEngine.go
timl3136: Resolved in newest commit

service/matching/matchingEngine.go
timl3136: Resolved in newest commit

service/matching/matchingEngine.go
```diff
@@ -296,6 +296,21 @@ func (e *matchingEngineImpl) AddDecisionTask(
...
+	// get the domainName
+	domainName, err := e.domainCache.GetDomainName(domainID)
+	if err != nil {
+		return false, err
+	}
+
+	// Only emit traffic metrics if the tasklist is not sticky and is not forwarded
+	if taskListKind != nil && *taskListKind != types.TaskListKindSticky && request.ForwardedFrom == "" {
+		e.metricsClient.Scope(metrics.MatchingAddTaskScope).Tagged(metrics.DomainTag(domainName),
+			metrics.TaskListTag(taskListName), metrics.TaskListTypeTag("activity_task")).IncCounter(metrics.CadenceTasklistRequests)
```
neil-xie, line 308: Tasklist type tag was changed from "decision_task" to "activity_task" in your latest commit, is it by accident or you combined the types?

timl3136: It was an error, ty for spotting

allenchen2244 (approved):
I think we're good to go!

neil-xie (approved):

PR #5219, opened 2023-04-20
Adds schema for domain config

davidporter-id-au:
Adds domain isolation group schema

### DIFF
--- a/schema/cassandra/cadence/schema.cql
+++ b/schema/cassandra/cadence/schema.cql
@@ -294,7 +294,7 @@ CREATE TYPE domain_config (
   visibility_archival_uri text,
   bad_binaries    blob,
   bad_binaries_encoding blob,
-  isolation_groups frozen<list<isolation_group_partition>>;
+  isolation_groups frozen<list<isolation_group_partition>>,
 );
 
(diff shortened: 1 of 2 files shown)


schema/cassandra/cadence/versioned/v0.35/partition_config.cql
Shaddoll, line 1: could you also update `schema.cql`?

davidporter-id-au: Little bit confused about this, if I add it to the base schema.cql, won't this cause problems

Shaddoll (approved):

schema/cassandra/cadence/schema.cql
```diff
@@ -289,6 +294,7 @@ CREATE TYPE domain_config (
   visibility_archival_uri text,
   bad_binaries    blob,
   bad_binaries_encoding blob,
+  isolation_groups frozen<list<isolation_group_partition>>;
```
Shaddoll, line 297: nit: ";" should be ","

davidporter-id-au (commented):
fix typo

PR #5228, opened 2023-04-24
[changes planned] Setting up some lightweight automation for publishing docker images

Groxx:
We have this in an internal wiki, but it's still quite manual, and there are a lot of steps and verbose docker flags and whatnot.
Since I got tired of rerunning it by hand, I figured others might too, so here's something simple.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,6 +396,49 @@ copyright: $(BIN)/copyright | $(BUILD) ## update copyright headers
 	$(BIN)/copyright
 	$Q touch $(BUILD)/copyright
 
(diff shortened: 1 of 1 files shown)


Makefile
Shaddoll, line 1: we have a workflow doing it https://github.com/uber/cadence/blob/master/.github/workflows/docker_publish.yml

Groxx: aah, TIL.  that's better than manually logging in, yeah.
might be worth only creating the build side of things then, for manual validation before running that workflow / creating that tag?

Groxx:
just gonna close it, I can always reopen

PR #5232, opened 2023-04-26
Fix isolation for sticky tasklist

Shaddoll:
**What changed?**
- Update sticky tasklist isolation matching to not use poller information to get isolation group

**Why?**
sticky tasklist doesn't know pollers from all isolation groups

**How did you test it?**
new test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -194,7 +194,7 @@ func newTaskListManager(
 		fwdr = newForwarder(&taskListConfig.forwarderConfig, taskList, *taskListKind, e.matchingClient)
 	}
 	var isolationGroups []string
-	if tlMgr.isIsolationEnabled() {
+	if tlMgr.isIsolationMatcherEnabled() {
 		isolationGroups = config.AllIsolationGroups
 	}
 	tlMgr.matcher = newTaskMatcher(taskListConfig, fwdr, tlMgr.scope, isolationGroups)
@@ -281,7 +281,10 @@ func (c *taskListManagerImpl) AddTask(ctx context.Context, params addTaskParams)
 			return r, err
 		}
 
-		isolationGroup := c.getIsolationGroupForTask(ctx, params.taskInfo)
+		isolationGroup, err := c.getIsolationGroupForTask(ctx, params.taskInfo)
+		if err != nil {
+			return false, err
+		}
 		// active task, try sync match first
 		syncMatch, err = c.trySyncMatch(ctx, params, isolationGroup)
 		if syncMatch {
@@ -362,10 +365,7 @@ func (c *taskListManagerImpl) getTask(ctx context.Context, maxDispatchPerSecond 
 	childCtx, cancel := c.newChildContext(ctx, c.config.LongPollExpirationInterval(), returnEmptyTaskTimeBudget)
 	defer cancel()
 
-	var isolationGroup string
-	if c.isIsolationEnabled() {
-		isolationGroup, _ = ctx.Value(_isolationGroupKey).(string)
-	}
+	isolationGroup, _ := ctx.Value(_isolationGroupKey).(string)
 	pollerID, ok := ctx.Value(pollerIDKey).(string)
 	if ok && pollerID != "" {
 		// Found pollerID on context, add it to the map to allow it to be canceled in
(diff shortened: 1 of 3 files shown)


service/matching/taskReader.go
```diff
@@ -133,11 +133,24 @@ dispatchLoop:
...
 			for {
 				// find isolation group of the task
-				isolationGroup := tr.getIsolationGroupForTask(tr.cancelCtx, taskInfo)
+				isolationGroup, err := tr.getIsolationGroupForTask(tr.cancelCtx, taskInfo)
+				if err != nil {
+					// it only errors when the tasklist is a sticky tasklist and
+					// the sticky pollers are not available, in this case, we just complete the task
+					// and let the decision get timed out and rescheduled to non-sticky tasklist
+					if err == _stickyPollerUnavailableError {
+						tr.completeTask(taskInfo, nil)
```
davidporter-id-au, line 142: I might need help understanding what the consequences of completing a task like this are

davidporter-id-au: I'm also still slightly unclear why this is in a loop

service/matching/taskListManager.go
```diff
@@ -587,12 +591,30 @@ func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, task
 		}
 		group, err := c.partitioner.GetIsolationGroupByDomainID(ctx, taskInfo.DomainID, partitionConfig, pollerIsolationGroups)
 		if err != nil {
```
davidporter-id-au, line 593: This should pretty much never occur unless there's truly exceptional conditions

service/matching/taskListManager.go
```diff
@@ -587,12 +591,30 @@ func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, task
...
+			if err == partition.ErrNoIsolationGroupsAvailable && c.taskListKind == types.TaskListKindSticky {
+				return "", _stickyPollerUnavailableError
+			}
+			// if we're unable to get the isolation group, log the error and fallback to no isolation
 			c.logger.Error("Failed to get isolation group from partition library", tag.WorkflowID(taskInfo.WorkflowID), tag.WorkflowRunID(taskInfo.RunID), tag.TaskID(taskInfo.TaskID), tag.Error(err))
-		} else {
-			return group
+			return "", nil
+		}
+		// For a sticky tasklist, it is possible that when an isolation group is undrained, the tasks from one workflow is reassigned
```
davidporter-id-au, line 602: Why is this only on undrain? Can't it happen on drain as well?

service/matching/taskListManager.go
```diff
@@ -587,12 +591,30 @@ func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, task
...
+		// to the isolation group undrained. If there is no poller from the isolation group, we should return StickyUnavailableError
+		// to let the task to be re-enqueued to the non-sticky tasklist. If there is poller, just return an empty isolation group, because
+		// there is at most one isolation group for sticky tasklist and we could just use empty isolation group for matching.
+		if c.taskListKind == types.TaskListKindSticky {
+			pollerIsolationGroups = c.pollerHistory.getPollerIsolationGroups(time.Now().Add(-time.Minute))
+			for _, pollerGroup := range pollerIsolationGroups {
+				if group == pollerGroup {
```
davidporter-id-au, line 609: How do we know this sticky poller is in an ok zone, it's not drained or something?

service/matching/taskReader.go
Shaddoll: outer loop is for draining all tasks from the buffer
inner loop is for retrying on dispatching one task to a poller

service/matching/taskListManager.go
Shaddoll: it can happen on drain as well. I didn't mention it because that's not surprising.

service/matching/taskListManager.go
davidporter-id-au: I missed the earlier point about this already being filtered down to the correct isolation-group on line 592

davidporter-id-au (approved):

PR #5234, opened 2023-04-27
Add pagination and flatten customized search attributes

bowenxia:
**What changed?**
Add pagination for pinot queries
Flatten search attributes into columns in Pinot tables and kafka message
Add method to handle query parsing and build match query for Pinot

**Why?**

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/pinot/pinotClient.go
```diff
@@ -172,6 +182,9 @@ func (c *PinotClient) convertSearchResultToVisibilityRecord(hit []interface{}, c
 func (c *PinotClient) getInternalListWorkflowExecutionsResponse(
 	resp *pinot.BrokerResponse,
 	isRecordValid func(rec *p.InternalVisibilityWorkflowExecutionInfo) bool,
+	token *elasticsearch.ElasticVisibilityPageToken,
```
neil-xie, line 185: We should create a new page token for Pinot or change it to a more generic one. It will be confused if you use the es token in Pinot. And one more question, does this need to be implemented in monorepo?

bowenxia: yes we also need to implement this in the mono repo, but those will be similar

bowenxia: Pinot token added

common/pinot/pinotClient.go
```diff
@@ -60,18 +61,33 @@ func (c *PinotClient) Search(request *SearchRequest) (*SearchResponse, error) {
...
 	resp, err := c.client.ExecuteSQL(c.tableName, query)
 	if err != nil {
 		return 0, &types.InternalServiceError{
-			Message: fmt.Sprintf("ListClosedWorkflowExecutions failed, %v", err),
+			Message: fmt.Sprintf("CountWorkflowExecutions ExecuteSQL failed, %v", err),
 		}
 	}
 
-	return int64(resp.ResultTable.GetRowCount()), nil
+	count, err := resp.ResultTable.Rows[0][0].(json.Number).Int64()
```
neil-xie, line 83: Just curious why GetRowCount() is not working here

PR #5241, opened 2023-05-01
Add Opensearch2 client with bulk API shared between clients

mantas-sidlauskas:
**What changed?**
Adding OpenSearch2 client support

Additional changes in this PR:
* AWS Credentials refactoring
* Added local Bulk API request utilities. The `bulk` package will be used by next ES/OS client versions (like `v8` or `os3`) 

**Why?**
To be able to use OpenSearch2 instead of Elasticsearch for indexing needs.

**How did you test it?**
Unit tests/local test

**Potential risks**

**Release notes**

**Documentation Changes**
Documentation changes will be provided as a separate PR


neil-xie (approved):

common/elasticsearch/bulk/bulk_delete_request.go
```diff
@@ -0,0 +1,184 @@
...
+// See https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docs-bulk.html
+// for details.
+type BulkDeleteRequest struct {
+	GenericBulkableRequest
+	index         string
+	typ           string
+	id            string
+	parent        string
+	routing       string
+	version       int64  // default is MATCH_ANY
```
Groxx, line 44: `// default is MATCH_ANY`
since I can't find that anywhere in ES's docs or the opensearch-go module: where is this defined?
I'm guessing it's something like MAX_INT, but:
- I can't find it described anywhere
- it doesn't seem to be 0, the actual default value
- it doesn't seem to be set to any non-zero value in here

so it seems misleading or possibly just completely wrong

common/elasticsearch/bulk/bulk_index_request.go
```diff
@@ -0,0 +1,253 @@
...
+	r.ifSeqNo = &ifSeqNo
+	return r
+}
+
+// IfPrimaryTerm indicates to only perform the index operation if the
+// last operation that has changed the document has the specified primary term.
+func (r *BulkIndexRequest) IfPrimaryTerm(ifPrimaryTerm int64) *BulkIndexRequest {
+	r.ifPrimaryTerm = &ifPrimaryTerm
+	return r
+}
```
Groxx, line 175: missing `r.source = nil`, and another reason I would prefer to not have that cache unless there's a demonstrated need for it.

common/elasticsearch/bulk/bulk_index_request.go
```diff
@@ -0,0 +1,253 @@
...
+		default:
+			body, err := json.Marshal(r.doc)
+			if err != nil {
+				return nil, err
+			}
+			lines[1] = string(body)
+		case json.RawMessage:
+			lines[1] = string(t)
+		case *json.RawMessage:
+			lines[1] = string(*t)
```
Groxx, line 241: Wouldn't this be handled by `json.Marshal` too, for "free"?

common/elasticsearch/bulk/bulk_index_request.go
```diff
@@ -0,0 +1,253 @@
...
+		case string:
+			lines[1] = t
+		case *string:
+			lines[1] = *t
```
Groxx, line 245: seems like these could all be rawmessages instead.  or use a `json.Valid(t)` check just in case.  (maybe valid isn't necessary, if it fails on push?  that's still a noticeable failure, that's probably good enough)

common/elasticsearch/client/os2/client.go
```diff
@@ -0,0 +1,320 @@
...
+		decoder: &NumberDecoder{},
+	}, nil
+}
+
+func (c *OS2) IsNotFoundError(err error) bool {
+	switch e := err.(type) {
+	case *Error:
+		return e.Status == http.StatusNotFound
+	}
+
```
Groxx, line 136: errors.As

common/elasticsearch/client/os2/client_bulk.go
```diff
@@ -0,0 +1,205 @@
...
+	// Opensearch client doesn't provide flush method
+	return nil
+}
+
+func (c *OS2) RunBulkProcessor(_ context.Context, parameters *bulk.BulkProcessorParameters) (bulk.GenericBulkProcessor, error) {
+	processor, err := opensearchutil.NewBulkIndexer(opensearchutil.BulkIndexerConfig{
+		Client:        c.client,
+		FlushInterval: parameters.FlushInterval,
+		NumWorkers:    parameters.NumOfWorkers,
+		Decoder:       &NumberDecoder{},
```
Groxx, line 176: tbh I wonder if this would be better as a `(*NumberDecoder)(nil)`.  zero-sized object pointers are quite funky and sometimes error-prone: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/#zero_size_var

host/esutils/client_os2.go
```diff
@@ -0,0 +1,157 @@
...
+			RetryBackoff: func(i int) time.Duration { return time.Duration(i) * 100 * time.Millisecond },
+		},
+	)
+
+	return &os2Client{
+		client: esClient,
+	}, err
+}
+
+func (os2 *os2Client) PutIndexTemplate(s suite.Suite, templateConfigFile, templateName string) {
```
Groxx, line 58: bleh.  would be good to change these to just use `*testing.T`, there's no reason at all to use a `testify/suite.Suite`

host/esutils/client_os2.go
```diff
@@ -0,0 +1,157 @@
...
+	s.Require().NoError(err)
+	req := opensearchapi.IndicesPutIndexTemplateRequest{
+		Body: template,
+		Name: templateName,
+	}
+
+	resp, err := req.Do(createContext(), os2.client)
+	s.Require().NoError(err)
+	s.Require().False(resp.IsError(), fmt.Sprintf("OS2 error: %s", resp.String()))
+	if resp.Body != nil {
```
Groxx, line 71: `s.Require().NoError(err)` should make this a guarantee / the else branch is unreachable.
these are plain http requests, and: https://pkg.go.dev/net/http#Client.Do
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.

(which is also what the request code assumes, so this is already effectively a requirement)

host/esutils/client_os2.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (os2 *os2Client) PutMaxResultWindow(indexName string, maxResultWindow int) error {
+
+	req := opensearchapi.IndicesPutSettingsRequest{
+		Body:  strings.NewReader(fmt.Sprintf(`{"max_result_window" : %d}`, maxResultWindow)),
+		Index: []string{indexName},
+	}
+
+	_, err := req.Do(createContext(), os2.client)
```
Groxx, line 126: missing close

host/esutils/client_os2.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (os2 *os2Client) GetMaxResultWindow(indexName string) (string, error) {
+
+	req := opensearchapi.IndicesGetSettingsRequest{
+		Index: []string{indexName},
+	}
+
+	res, err := req.Do(createContext(), os2.client)
+
+	defer res.Body.Close()
```
Groxx, line 139: missing error check

host/integrationbase.go
```diff
@@ -141,9 +140,7 @@ func (s *IntegrationBase) setupSuite() {
 }
 
 func (s *IntegrationBase) setupLogger() {
-	zapLogger, err := zap.NewDevelopment()
-	s.Require().NoError(err)
-	s.Logger = loggerimpl.NewLogger(zapLogger)
+	s.Logger = loggerimpl.NewLoggerForTest(s.Suite)
```
Groxx, line 143: I _think_ I tried this a while ago and ran into lots of race conditions due to some background logging after test shutdown (e.g. gocql I think?).

So test it thoroughly for flakiness.  But if it's working, 👍 absolutely an improvement.

Groxx (changes requested):
gonna mark as a no for a couple gaps, but in a broad sense it seems good.  I can't really tell if it's _complete_ though, I'm not familiar enough with these apis.

common/elasticsearch/bulk/bulk_delete_request.go
mantas-sidlauskas: that comment was misleading, removing it

host/esutils/client_os2.go
mantas-sidlauskas: fixed

host/esutils/client_os2.go
mantas-sidlauskas: thanks, fixed

host/integrationbase.go
mantas-sidlauskas: I was trying to reduce the amount of logs tests produce here, but for tests we are always passing `-v` which turns on verbosity and all logs are emitted. I can leave it as it is, or revert the usage of zap test logger

common/elasticsearch/client/client.go
```diff
@@ -54,17 +54,23 @@ type Client interface {
 
 // Response is used to pass data retrieved from Elasticsearch/OpenSearch to upper layer
 type Response struct {
-	TookInMillis int64
+	TookInMillis int64 `json:"took,omitempty"`
```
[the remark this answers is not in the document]
mantas-sidlauskas: yeah, that's something I have overlooked. This struct is used for both: assigning olivere/elastic produced result AND unmarshaling data from OpenSearch http response when opensearch client is used. I think I need to add OpenSearch related Response struct with json annotations and keep this `Response{}` without any json annotations

host/esutils/client_os2.go
mantas-sidlauskas: done

host/esutils/client_os2.go
mantas-sidlauskas: done

common/elasticsearch/client/client.go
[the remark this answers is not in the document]
mantas-sidlauskas: done

common/elasticsearch/bulk/bulk_index_request.go
mantas-sidlauskas: We don't actually use those capabilities, removing it from Cadence code

common/elasticsearch/bulk/bulk_delete_request.go
```diff
@@ -0,0 +1,184 @@
...
+	var body []byte
+
+	body, err = json.Marshal(command)
+
+	if err != nil {
+		return nil, err
+	}
+
+	lines := []string{string(body)}
+	r.source = lines
```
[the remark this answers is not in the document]
mantas-sidlauskas: It depends on a "processor" implementation: for example `olivere/v6` and `v7` is calling `.Source()` multiple times:
- https://github.com/olivere/elastic/blob/29ee98974cf1984dfecf53ef772d721fb97cb0b9/bulk_processor.go#L484
- https://github.com/olivere/elastic/blob/29ee98974cf1984dfecf53ef772d721fb97cb0b9/bulk.go#L196
- https://github.com/olivere/elastic/blob/29ee98974cf1984dfecf53ef772d721fb97cb0b9/bulk.go#L217

common/elasticsearch/bulk/bulk_index_request.go
```diff
@@ -0,0 +1,190 @@
...
+	r.opType = opType
+	r.source = nil
+	return r
+}
+
+// Version indicates the version of the document as part of an optimistic
+// concurrency model.
+func (r *BulkIndexRequest) Version(version int64) *BulkIndexRequest {
+	v := version
+	r.version = &v
```
Groxx, line 98:
```suggestion
	r.version = &version
```
it's already a copy (value arg), no need to make another.

common/elasticsearch/bulk/bulk_index_request.go
```diff
@@ -0,0 +1,190 @@
...
+	// { "index" : { "_index" : "test", "_type" : "type1", "_id" : "1" } }
+	// { "field1" : "value1" }
+
+	if r.source != nil {
+		return r.source, nil
+	}
+
+	lines := make([]string, 2)
+
+	// "index" ...
```
Groxx, line 145: was there something intended here, or should we just get rid of it?
```suggestion
```

common/config/elasticsearch_test.go
```diff
@@ -0,0 +1,101 @@
...
+			err: false,
+		},
+	}
+
+	for _, tc := range tests {
+		if tc.err {
+			assert.Error(t, tc.config.Validate(), tc.msg)
+		} else {
+			assert.NoError(t, tc.config.Validate(), tc.msg)
+		}
```
Groxx, line 98: usually I'll vote to use an error string, and check if the error contains that string (to make sure you're getting the error you expect, rather than literally any error).

but no requirement at all, this is what we have in most code.  it's fine as-is.

common/elasticsearch/bulk/bulk_delete_request.go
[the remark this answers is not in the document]
Groxx: yea, that does seem to be at least 2, maybe 3+ times per thing...

tbh I still have doubts that it's worth the upkeep, but these are pretty high volume.  seems defensible.  I just wish it were less error-prone, though they look good so far and they're unlikely to need frequent changes anyway.

host/integrationbase.go
Groxx: nah, if it works that's absolutely worth doing.  I just had problems doing similar things previously, but that was several months ago at least.

common/elasticsearch/bulk/bulk_delete_request.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// -- Bulk delete request --
+
+// BulkDeleteRequest is a request to remove a document from Elasticsearch.
+//
+// See https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docs-bulk.html
+// for details.
+type BulkDeleteRequest struct {
+	GenericBulkableRequest
```
Groxx, line 38: after a bunch of reading: I think there are a few issues with these three special request types.

Special types in general: no complaint at all!  More types is more good IMO.  But for these:

- this is declaring a *struct field* which contains a generic bulkable request, but it doesn't seem to be assigned anywhere.
  - I'm guessing this is intended as a "this struct implements this interface", but that's not what this does; if anything it makes it _riskier_ since nothing is ever assigned to it.  it's likely null -> the BulkDeleteRequest still implements these methods because it's embedded, but now you have no warning if you forget a method, it'll just panic.
  - try a `var _ GenericBulkableRequest = (*BulkDeleteRequest)(nil)` as a type assertion if the goal is "must implement X interface".
- these are in `bulk`, but only `os2` uses them.
  - is the plan to migrate all implementations to them?  that seems fine (yay types rather than enum + generic!), it's just not documented for future readers prior to that being built.  it doesn't have to be in this PR, but there should probably be something ready-to-go immediately after, or comments so it's not confusing in the future if that gets pushed back.

common/elasticsearch/client/os2/client.go
```diff
@@ -0,0 +1,378 @@
...
+	osapi "github.com/opensearch-project/opensearch-go/v2/opensearchapi"
+	requestsigner "github.com/opensearch-project/opensearch-go/v2/signer/aws"
+
+	"github.com/uber/cadence/common/config"
+	"github.com/uber/cadence/common/elasticsearch/client"
+	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/types"
+)
+
+type (
```
Groxx, line 45: Aside from `OS2`, these appear to be internal-to-this-package types.
Can they be made private?  Private *types* are fine for JSON, it's only private *fields* that are a problem.

common/elasticsearch/common.go
```diff
@@ -55,18 +50,6 @@ func buildTLSHTTPClient(config config.TLS) (*http.Client, error) {
 	return tlsClient, nil
 }
 
-func buildAWSSigningClient(awsconfig config.AWSSigning) (*http.Client, error) {
```
Groxx, line 58: looks like this was just totally unused?  :+1:

common/config/elasticsearch.go
```diff
@@ -93,13 +98,44 @@ func (cfg *ElasticSearchConfig) SetUsernamePassword() {
...
+	}
+
+	if a.StaticCredential != nil && len(a.StaticCredential.Region) == 0 {
+		return errors.New("missing region in staticCredential")
+	}
+
 	return nil
 }
+
+func (a AWSSigning) GetCredentials() (*credentials.Credentials, *string, error) {
```
Groxx, line 120: yea, makes sense to move validation and creation in here I think.  :+1:

docker/buildkite/docker-compose-opensearch2.yml
```diff
@@ -36,7 +36,7 @@ services:
           - elasticsearch
     environment:
       - discovery.type=single-node
-      - plugins.security.disabled=true
+      - "DISABLE_SECURITY_PLUGIN=true"
```
Groxx, line 39: did it just change env-var formats?  I lightly prefer the dot-separated one tbh.

host/esutils/client_os2.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+)
+
+func newOS2Client(url string) (*os2Client, error) {
+
+	esClient, err := opensearch.NewClient(
+		opensearch.Config{
+			Addresses:    []string{url},
+			MaxRetries:   5,
+			RetryBackoff: func(i int) time.Duration { return time.Duration(i) * 100 * time.Millisecond },
```
Groxx, line 50: should this be exponential, like the others?

host/esutils/client_v6.go
```diff
@@ -45,34 +46,34 @@ func newV6Client(url string) (*v6Client, error) {
 	}, err
 }
 
-func (es *v6Client) PutIndexTemplate(s suite.Suite, templateConfigFile, templateName string) {
+func (es *v6Client) PutIndexTemplate(t *testing.T, templateConfigFile, templateName string) {
```
Groxx, line 49: ah, and you rewrote these too.  excellent, thank you!

host/esutils/client_os2.go
```diff
@@ -0,0 +1,162 @@
...
+	)
+
+	return &os2Client{
+		client: esClient,
+	}, err
+}
+
+func (os2 *os2Client) PutIndexTemplate(t *testing.T, templateConfigFile, templateName string) {
+	// This function is used exclusively in tests. Excluding it from security checks.
+	// #nosec
```
Groxx, line 61: I'm kinda curious if this would be done implicitly if you call `t.Helper()`... but eh, it's the same as the others, seems fine.

host/esutils/client_os2.go
```diff
@@ -0,0 +1,162 @@
...
+		Index: []string{indexName},
+	}
+	resp, err := deleteReq.Do(createContext(), os2.client)
+	require.NoError(t, err)
+	require.False(t, resp.IsError(), fmt.Sprintf("OS2 error: %s", resp.String()))
+	resp.Body.Close()
+
+}
+
+func (os2 *os2Client) PutMaxResultWindow(indexName string, maxResultWindow int) error {
```
Groxx, line 116: while you're in here: these also seem to only be used in tests.  add a `*testing.T` arg maybe, to make that explicit?

(no requirement, it's already improved and this is pretty easy to verify)

host/testdata/es_os2_index_template.json
```diff
@@ -0,0 +1,116 @@
...
+      "properties": {
+        "DomainID": {
+          "type": "keyword"
+        },
+        "WorkflowID": {
+          "type": "keyword"
+        },
+        "RunID": {
+          "type": "keyword"
+        },
```
Groxx, line 23: pass this through `jq -S .` maybe?  that'll sort all keys, makes it a bit nicer if it's changed in the future.

schema/elasticsearch/os2/visibility/index_template.json
```diff
@@ -0,0 +1,85 @@
...
+  "settings": {
+    "index": {
+      "number_of_shards": "5",
+      "number_of_replicas": "0"
+    }
+  },
+  "mappings": {
+    "dynamic": "false",
+    "properties": {
+      "DomainID": {
```
Groxx, line 15: same `jq -S .`

host/elastic_search_test.go
```diff
@@ -18,9 +18,6 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-//go:build esintegration
-// +build esintegration
```
Groxx, line 22: seems pretty likely an improvement (build flags are quite problematic imo).  :+1: assuming no surprises in CI or something.

host/esutils/client_os2.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (os2 *os2Client) CreateIndex(t *testing.T, indexName string) {
+	existsReq := opensearchapi.IndicesExistsRequest{
+		Index: []string{indexName},
+	}
+	resp, err := existsReq.Do(createContext(), os2.client)
+	require.NoError(t, err)
+
+	if resp.StatusCode == http.StatusOK {
+		deleteReq := opensearchapi.IndicesDeleteRequest{
```
Groxx, line 85: kinda odd that these delete the index implicitly... but eh.  the others do too.

Groxx (changes requested):
alrighty, last round I think.  mostly just the `Bulk*Request` notes since that seems like A Problem™, the rest you can pretty much take or leave however you think is reasonable 👍

common/elasticsearch/bulk/bulk_index_request.go
Groxx: yea, these are handled by json.Marshal:
```
func TestJSON(t *testing.T) {
	orig := `{"valid":"json"}`
	t.Run("raw value", func(t *testing.T) {
		r := json.RawMessage(orig)
		s, err := json.Marshal(r)
		require.NoError(t, err)
		require.Equal(t, string(s), orig)
	})
	t.Run("raw pointer", func(t *testing.T) {
		r := json.RawMessage(orig)
		s, err := json.Marshal(&r)
		require.NoError(t, err)
		require.Equal(t, string(s), orig)
	})
}
```
it just passes through, essentially "for free" (but safer, because it does still validate that it's valid json, rather than throwing it in regardless)

common/elasticsearch/bulk/bulk_index_request.go
Groxx: ^ gonna bump on these, as it seems like they're unused.  AFAICT this and the duplicate 10 lines below it are the only places that `Doc` is assigned: https://github.com/uber/cadence/blob/b2bc8bf0edbafc144f318bbd4e2a222d3e96c42c/service/worker/indexer/indexer.go#L238
^ that's assigning a `map[string]interface{}`.  no need for explicit, unsafe string handling.

common/elasticsearch/bulk/bulk_index_request.go
mantas-sidlauskas: done

common/config/elasticsearch_test.go
mantas-sidlauskas: adding `assert.equal()` to match to the expected error

docker/buildkite/docker-compose-opensearch2.yml
mantas-sidlauskas: you can use dot-separated in `opensearch.yaml`, but for starting docker container with security disabled, you have to use this type of variable. Related documentation: https://github.com/opensearch-project/opensearch-build/blob/main/docker/release/README.md#disable-security-plugin-security-dashboards-plugin-security-demo-configurations-and-related-configurations

host/esutils/client_os2.go
mantas-sidlauskas: done

host/testdata/es_os2_index_template.json
mantas-sidlauskas: good idea, thanks

host/esutils/client_os2.go
mantas-sidlauskas: yes, that seems to be a bit odd, given that `TearDownSuite` should delete that index anyway.

common/elasticsearch/bulk/bulk_delete_request.go
mantas-sidlauskas: re this `bulk` package:
There's no plan to migrate old versions (v6,v7) to use local version of `bulk` package as `v6` and `v7` use their own `bulk` packages. When implementing support for Elasticsearch v8 client, cadence-local `bulk` package will be used. 

re type-assert: done

common/elasticsearch/client/os2/client.go
mantas-sidlauskas: yep, good catch, making them private

host/elastic_search_test.go
mantas-sidlauskas: actually, I removed this for my local tests. Buildkite is just working fine, but it might be annoying to start those tests locally without ES running, I might revert this change

common/elasticsearch/bulk/bulk_index_request.go
mantas-sidlauskas: The actual request to ES/Opensearch looks like this: https://gist.github.com/mantas-sidlauskas/c2abb891f3c21b0c09b2b9d4a547e1be

Keeping only `json.Marshal`

common/elasticsearch/bulk/bulk_index_request.go
mantas-sidlauskas: done

host/esutils/client_os2.go
mantas-sidlauskas: keeping as it is for now

common/elasticsearch/client/os2/client.go
```diff
@@ -328,11 +328,11 @@ func (c *OS2) Search(ctx context.Context, index, body string) (*client.Response,
 	defer closeBody(resp)
 	if resp.IsError() {
 		return nil, types.InternalServiceError{
-			Message: fmt.Sprintf("OpenSearch Search Error: %v", c.parseError(resp)),
+			Message: fmt.Sprintf("OpenSearch SearchsError: %v", c.parseError(resp)),
```
Groxx, line 331:
```suggestion
			Message: fmt.Sprintf("OpenSearch Search Error: %v", c.parseError(resp)),
```

common/elasticsearch/bulk/bulk_delete_request.go
Groxx: `os2` and `v8` are sufficiently similar that that'll be easy then?  could be reasonable, yea.  and not-migrating very old stuff seems fine too, we can consider that frozen unless forced to change.

any idea if we can remove `v6` any time soon?  7 came out in 2019.  
though surprisingly it seems they still offer support all the way back to _5.5_ which is ancient: https://www.elastic.co/support/matrix

host/elastic_search_test.go
Groxx: yeah, what I'd prefer to do here is to always build everything (greatly improves IDE behavior, and ensures breaking refactors are actually complete), but skip tests that don't have the necessary env setup running.

but 👍 that's an unrelated change to this PR.

Groxx (approved):
One minor thing but yep, SGTM :)

Re commit message: maybe mention there that this is kinda intended to share with ES v8?  otherwise it's pretty good imo.

common/elasticsearch/bulk/bulk_delete_request.go
mantas-sidlauskas: IMHO, it should be good to remove dependencies on old versions in general. It would be good idea to add support for `v8` and ind the same time, remove `v6` client

PR #5243, opened 2023-05-02
Create TTL in the UpdateWorkflowExecution cycles.

agautam478:
**What changed?**
Adding TTL to the CreateWorkflowRequest is not enough because it gets overwritten by the 'UpdateWorkflowRequest' and the final value of the TTL is persisted.
So we reimplemented the code and added TTL into the UpdateworkflowRequest instead. 
Supporting document:
https://docs.google.com/document/d/14Bn3W32DdJEP9dVi2MDTyGD0joUG8K_i4wCFEfFZJns/edit?usp=sharing

**Why?**
Calculate the TTL in the Update Workflow context and update the subsequent queries down the persistence layer. Adding TTL to the CreateWorkflowRequest is not enough because it gets overwritten by the 'UpdateWorkflowRequest' and the TTL is removed. Cassandra does not allow persistence of the TTL on its tables. If you perform an update without TTL then the TTL will be lost. In other words, we had to insert the TTL during the Workflow state update.

Moved the TTL to Workflow mutation

**How did you test it?**
Tested locally and unit tests.

**Potential risks**
We might experience data loss if the TTL value is not appropriate or if the TTL accidentally cause data deletion.

**Release notes**
NA

**Documentation Changes**
NA


common/persistence/dataManagerInterfaces.go
```diff
@@ -964,6 +964,8 @@ type (
 		Encoding common.EncodingType // optional binary encoding type
 
 		DomainName string
+
+		TTLInSeconds int
```
Shaddoll, line 968: `UpdateWorkflowMutation` and `NewWorkflowSnapshot` in this struct both requires a TTL, and their TTLs should be different. What does this TTL belong to?

Shaddoll: Could you move the field to `WorkflowMutation` if it's only for update?

service/history/execution/context.go
```diff
@@ -1215,6 +1217,20 @@ func (c *contextImpl) updateWorkflowExecutionWithRetry(
 		return err
 	}
 
+	//Calculating the TTL for workflow Execution.
+	domainObj, _ := c.shard.GetDomainCache().GetDomainByID(c.domainID)
```
Shaddoll, line 1221: the code below this line is not safe if this line returns an error

service/history/execution/context.go
```diff
@@ -1215,6 +1217,20 @@ func (c *contextImpl) updateWorkflowExecutionWithRetry(
...
+	config := domainObj.GetConfig()
+	retention := time.Duration(config.Retention)
+	daysInSeconds := int((retention + ttlBufferDays) * dayToSecondMultiplier)
+
+	startTime := request.UpdateWorkflowMutation.ExecutionInfo.StartTimestamp
+	if time.Time.IsZero(startTime) {
+		request.TTLInSeconds = daysInSeconds
+	} else {
+		request.TTLInSeconds = int(request.UpdateWorkflowMutation.ExecutionInfo.DecisionStartToCloseTimeout) - int(time.Now().Sub(startTime).Seconds()) + daysInSeconds
```
Shaddoll, line 1230: I think it should not be `DecisionStartToCloseTimeout`, and it should be the workflow timeout, not the decision timeout

Shaddoll: Another thing is to make sure `int(request.UpdateWorkflowMutation.ExecutionInfo.DecisionStartToCloseTimeout) - int(time.Now().Sub(startTime).Seconds())` is >= 0. 
It is possible to be negative when there is an update after workflow timeout.

service/history/execution/context.go
```diff
@@ -1215,6 +1217,20 @@ func (c *contextImpl) updateWorkflowExecutionWithRetry(
...
+		c.logger.SampleInfo("Calculated TTL", request.TTLInSeconds)
```
Shaddoll, line 1231: Can you remove this log?

service/history/execution/context.go
```diff
@@ -1215,6 +1217,20 @@ func (c *contextImpl) updateWorkflowExecutionWithRetry(
 		return err
 	}
 
+	//Calculating the TTL for workflow Execution.
```
Shaddoll, line 1220: The following logic is better to be moved to `CloseTransactionAsMutation` function.
And it can be shared with UpdateWorkflow and ConflictResolveWorkflow.

common/persistence/nosql/nosqlplugin/mongodb/workflow.go
```diff
@@ -55,6 +55,7 @@ func (db *mdb) UpdateWorkflowExecutionWithTasks(
 	replicationTasks []*nosqlplugin.ReplicationTask,
 	timerTasks []*nosqlplugin.TimerTask,
 	shardCondition *nosqlplugin.ShardCondition,
+	ttl int,
```
shijiesheng, line 58: nit `ttlInSeconds`

common/persistence/nosql/nosqlExecutionStore.go
```diff
@@ -472,7 +472,7 @@ func (d *nosqlExecutionStore) ConflictResolveWorkflowExecution(
 		ctx, currentWorkflowWriteReq,
 		mutateExecution, insertExecution, resetExecution,
 		nosqlTransferTasks, nosqlCrossClusterTasks, nosqlReplicationTasks, nosqlTimerTasks,
-		shardCondition)
+		shardCondition, 0)
```
Shaddoll, line 475: TTL should be also set for `ConflictResolveWorkflowExecution`

shijiesheng: 0 means delete current TTL. https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useExpireExample.html

common/persistence/nosql/nosqlplugin/interfaces.go
```diff
@@ -447,6 +447,7 @@ type (
 			replicationTasks []*ReplicationTask,
 			timerTasks []*TimerTask,
 			shardCondition *ShardCondition,
+			ttlInSeconds int,
```
Shaddoll, line 450: the ttl parameter should be added as a new field to `WorkflowExecutionRequest `

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -609,6 +609,159 @@ func (s *ExecutionManagerSuite) TestUpdateWorkflowExecutionStateCloseStatus() {
 	}
 }
 
+// TestUpdateWorkflowExecutionTTL test
+func (s *ExecutionManagerSuite) TestUpdateWorkflowExecutionTTL() {
+	//TODO: Update the test when the TTLbuffer becomes configurable.
```
Shaddoll, line 614: Can we make sure the test only runs when the database is Cassandra?

Shaddoll: example https://github.com/uber/cadence/blob/master/common/persistence/persistence-tests/matchingPersistenceTest.go#L558

common/persistence/nosql/nosqlExecutionStoreUtil.go
```diff
@@ -85,9 +85,10 @@ func (d *nosqlExecutionStore) prepareResetWorkflowExecutionRequestWithMapsAndEve
 	versionHistories := resetWorkflow.VersionHistories
 	nowTimestamp := time.Now()
 
+	//TTLInSeconds in invalid in this case so passing a default value of 0.
```
shijiesheng, line 88: typo

common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
```diff
@@ -1205,8 +1203,13 @@ func (db *cdb) updateWorkflowExecution(
 ) error {
 	execution.StartTimestamp = db.convertToCassandraTimestamp(execution.StartTimestamp)
 	execution.LastUpdatedTimestamp = db.convertToCassandraTimestamp(execution.LastUpdatedTimestamp)
-
+	//default TTL Value. 0 TTL means no ttl is set, hence your records will persist forever unless explicitly deleted.
+	ttlInSeconds := 0
+	if db.dc.EnableExecutionTTL(domainID) && execution.TTLInSeconds >= 0 {
```
shijiesheng, line 1208: If we add another nosql plugin, we need this logic as well. Can you move the feature flag to where you calculate it?

common/dynamicconfig/config.go
```diff
@@ -131,7 +131,7 @@ type DurationPropertyFn func(opts ...FilterOption) time.Duration
 // DurationPropertyFnWithDomainFilter is a wrapper to get duration property from dynamic config with domain as filter
 type DurationPropertyFnWithDomainFilter func(domain string) time.Duration
 
-// DurationPropertyFnWithDomainIDFilter is a wrapper to get duration property from dynamic config with domainID as filter
+// DurationPropertyFnWithDomainIDFilter is a wrapper to get duration property git statfrom dynamic config with domainID as filter
```
Shaddoll, line 134: nit: comment

common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
Shaddoll: as shijie said, could you move the feature flag to that place? That we don't need to do check in all types of databases.

service/history/execution/mutable_state_builder.go
```diff
@@ -4807,3 +4813,23 @@ func (e *mutableStateBuilder) logDataInconsistency() {
...
+	if err != nil {
+		return 0, err
+	}
+	config := domainObj.GetConfig()
+	retention := time.Duration(config.Retention)
+	daysInSeconds := int((retention + ttlBufferDays) * dayToSecondMultiplier)
+	//Default state of TTL, means there is no TTL attached.
+	TTLInSeconds := 0
+	startTime := e.executionInfo.StartTimestamp
+	if !time.Time.IsZero(startTime) {
```
Shaddoll, line 4830: how do we handle cron workflow?

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -609,6 +609,163 @@ func (s *ExecutionManagerSuite) TestUpdateWorkflowExecutionStateCloseStatus() {
...
+		s.Nil(err)
+		info, err = s.GetWorkflowExecutionInfo(ctx, domainID, workflowExecution)
+		s.Nil(err)
+		s.Equal(p.WorkflowStateCompleted, info.ExecutionInfo.State)
+		s.Equal(closeStatus, info.ExecutionInfo.CloseStatus)
+
+		time.Sleep(65)
+
+		_, err = s.GetWorkflowExecutionInfo(ctx, domainID, workflowExecution)
+		s.Nil(err)
```
Shaddoll, line 764: I think a `not found` error should be expected.

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -609,6 +609,163 @@ func (s *ExecutionManagerSuite) TestUpdateWorkflowExecutionStateCloseStatus() {
...
+		})
+		s.Nil(err)
+		info, err = s.GetWorkflowExecutionInfo(ctx, domainID, workflowExecution)
+		s.Nil(err)
+		s.Equal(p.WorkflowStateCompleted, info.ExecutionInfo.State)
+		s.Equal(closeStatus, info.ExecutionInfo.CloseStatus)
+
+		time.Sleep(65)
+
+		_, err = s.GetWorkflowExecutionInfo(ctx, domainID, workflowExecution)
```
Shaddoll, line 763: I think a `not-found` error should be expected.

common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
```diff
@@ -1205,8 +1203,26 @@ func (db *cdb) updateWorkflowExecution(
 ) error {
 	execution.StartTimestamp = db.convertToCassandraTimestamp(execution.StartTimestamp)
 	execution.LastUpdatedTimestamp = db.convertToCassandraTimestamp(execution.LastUpdatedTimestamp)
-
-	batch.Query(templateUpdateWorkflowExecutionWithVersionHistoriesQuery,
+	//default TTL Value. 0 TTL means no ttl is set, hence your records will persist forever unless explicitly deleted.
+	ttlInSeconds := 0
+	//Only fires when the workflow is closing.
+	if execution.CloseStatus == 1 {
```
Shaddoll, line 1209: this condition doesn't cover all closed workflows https://github.com/uber/cadence/blob/714aec48796b62af163e9d324462a1cc2afb4981/common/persistence/dataManagerInterfaces.go#L129C5-L129C5

Shaddoll: I think you should check `state == WorkflowStateCompleted` instead https://github.com/uber/cadence/blob/714aec48796b62af163e9d324462a1cc2afb4981/common/persistence/dataManagerInterfaces.go#L120

Shaddoll (approved):

PR #5244, opened 2023-05-02
Fix docker

davidporter-id-au:
**What changed?**
- Fixes docker build containers which were failing to install some dependencies
- Bumps python on the build container to be python3

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Groxx (approved):
yea, python 3 is probably the easy short-term fix.
we should really replace this with a tiny custom go binary tho.  all we use it for is healthchecks and schema commands afaik.

PR #5245, opened 2023-05-02
Insert Hostname tag into metrics and other services

timl3136:
**What changed?**
Add hostname tag into cadence matching as well as other services

**Why?**
It enables tracking metrics for each host and can better track machine usage

**How did you test it?**
Local unit tests and test functionality on dev server

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/config/config.go
```diff
@@ -323,10 +323,13 @@ type Config struct {
 	LargeShardHistorySizeMetricThreshold  dynamicconfig.IntPropertyFn
 	LargeShardHistoryEventMetricThreshold dynamicconfig.IntPropertyFn
 	LargeShardHistoryBlobMetricThreshold  dynamicconfig.IntPropertyFn
+
+	// Hostname addition
+	HostName string
```
Groxx, line 328: ... addition of what?

allenchen2244 (approved):

service/history/config/config.go
timl3136: Add description of HostName in newest diff

PR #5250, opened 2023-05-04
Domain name validation

ketsiambaku:
**What changed?**
-Added domain name validation on API level
-Updated unit tests because some unit tests were failing due to bad domain naming

**Why?**
Domain names are used in URLs therefore domain names should be URL friendly. We should validate the domain names to be url friendly.

**How did you test it?**
Tested in cadence-frontend staging2

**Potential risks**
breaking domain registration API

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/domain/handler.go
+++ b/common/domain/handler.go
@@ -25,6 +25,7 @@ package domain
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/pborman/uuid"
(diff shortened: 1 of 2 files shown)


allenchen2244:
Can we add a negation test case such that a bad input actually returns the error you expect?

ketsiambaku:
> Can we add a negation test case such that a bad input actually returns the error you expect?

Done @allenchen2244

allenchen2244 (approved):

PR #5255, opened 2023-05-07
Change time procces func to admit sec,milli,micro and nano in command queries

ignaciogonzalezsomoza:
**What changed?**
The time processor is modified to allow queries in which search queries are entered with precision of seconds, milliseconds, microseconds and nanoseconds.

**Why?**
Currently, the queries made to obtain a list of flows forced to introduce the complete format of the date in nano type, for example:
`cadence --domain samples-domain wf list --query "CloseTime >1683388191918908000"
`
If queries were made with less precision, for example micro (1683388191918908), milli (1683388191918) or sec (1683388191) the query did not return information when comparing with a different type. To avoid this problem, padding is performed by completing the searched dates with zero until the types are equal and all formats can be consulted.

**How did you test it?**
Running various tests via command with different operators (><=) and different date formats

`cadence --domain samples-domain wf list --query "CloseTime >1683388191918908000"
`

**Potential risks**

**Release notes**

**Documentation Changes**


ignaciogonzalezsomoza:
Incorrect user

PR #5257, opened 2023-05-08
Update history to persist partition config

Shaddoll:
**What changed?**
Update history to persist partition config in mutable state and workflow started event and forward it to matching when pushing activity and decision tasks to matching

**Why?**

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -519,10 +519,12 @@ func CreateHistoryStartWorkflowRequest(
 	domainID string,
 	startRequest *types.StartWorkflowExecutionRequest,
 	now time.Time,
+	partitionConfig map[string]string,
 ) (*types.HistoryStartWorkflowExecutionRequest, error) {
 	histRequest := &types.HistoryStartWorkflowExecutionRequest{
-		DomainUUID:   domainID,
-		StartRequest: startRequest,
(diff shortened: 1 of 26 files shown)


common/util.go
```diff
@@ -519,10 +519,12 @@ func CreateHistoryStartWorkflowRequest(
 	domainID string,
 	startRequest *types.StartWorkflowExecutionRequest,
 	now time.Time,
+	partitionConfig map[string]string,
```
davidporter-id-au, line 522: I'm worried about changes such as this will presumably need to include subsequent changes once the rest of the feature is added in which will lead to a lot of conflicts as both commits will be editing the same line. Are you sure you want to do it this way? 

What about an alternative where we merge: 
1. Add to, or expand existing Types in a non-breaking way
2. add the Libraries and other dependencies we'll need
3. Doing all these 'breaking' changes in a single commit where the CreateHistoryStartWorkflowRequest immediately uses the data it's expecting, rather than requirign a subsequent refactor

Shaddoll: There are no conflicts from the feature branch.

davidporter-id-au (approved):

PR #5267, opened 2023-05-09
Fix panic while parsing workflows with timeouts

dkrotx:
When cadence-cli parses workflows with timeouts it panic since these for this workflows Cadence is using time.Time and tries to dig into it's private fields with reflection.
The fix is to detect this edge-case and to print time.Time as fmt would do.

+Rename util.go -> utils.go so that utils_test.go are now contribute to coverage%

**What changed?**
Instead of calling reflection on time.Time and its private fields, we're now print it just like `fmt.Sprint` would

**Why?**
An attempt to call `workflow show` for cadence.canary.cron workflows is making CLI panic
That's because of FirstScheduleTime having time.Time type and CLI was diving into it with reflection.

**How did you test it?**
Unit-test + testing with cadence-cli

**Potential risks**
I don't see any. This is CLI + it's been panicing before, and shouldn't panic now

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/cli/utils.go
+++ b/tools/cli/utils.go
@@ -39,18 +39,16 @@ import (
 	"strings"
 	"time"
 
-	"github.com/uber/cadence/common/pagination"
-
-	"github.com/uber/cadence/client/frontend"
-	"github.com/uber/cadence/common/types"
-
 	"github.com/cristalhq/jwt/v3"
 	"github.com/fatih/color"
 	"github.com/urfave/cli"
 	"github.com/valyala/fastjson"
 
+	"github.com/uber/cadence/client/frontend"
 	"github.com/uber/cadence/common"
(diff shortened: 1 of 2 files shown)


tools/cli/utils.go
```diff
@@ -39,18 +39,16 @@ import (
 	"strings"
 	"time"
 
-	"github.com/uber/cadence/common/pagination"
-
-	"github.com/uber/cadence/client/frontend"
-	"github.com/uber/cadence/common/types"
```
jakobht, line 45: Should the imports have been reordered?

jakobht (approved):

tools/cli/utils.go
dkrotx: I think it's OK.
It doesn't look correct to split them anyway.

PR #5269, opened 2023-05-09
Make ESClient fields public to use them in monorepo

neil-xie:
**What changed?**
Change the fields of ESClient to public so they can be accessed in monorepo

**Why?**

**How did you test it?**
Tested at dev2 server

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/elasticsearch/client.go
+++ b/common/elasticsearch/client.go
@@ -45,20 +45,20 @@ import (
 var _ GenericClient = (*ESClient)(nil)
 
 type ESClient struct {
-	client client.Client
-	logger log.Logger
+	Client client.Client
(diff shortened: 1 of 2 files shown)


bowenxia (approved):

Shaddoll (changes requested):
There should be a better approach

bowenxia:
Hi @Shaddoll, do you think add a new function for ESClient like `NewESClient(c client.Client, logger log.Logger)` would possibly be a better idea?

Shaddoll (approved):

Shaddoll:
> There should be a better approach

Ignore my comments.

PR #5270, opened 2023-05-09
Adds isolation groups to persistence

davidporter-id-au:
Adds the persistence layer changes for the ['Zonal-isolation' feature](https://github.com/uber/cadence-idl/commit/88be1470a45198d4546a484104d54468d6c17dbf): 

- Adds schema changes to include isolation-group information for domain configuration data
- Adds Isolation-groups to the DB for domains
- Adds Isolation-group as a different config-store dimension, so that config store may be used to persist cluster-wide configuration.


schema/cassandra/cadence/schema.cql
```diff
@@ -278,6 +278,11 @@ CREATE TYPE domain (
   owner_email text,
 );
 
+CREATE TYPE isolation_group_partition (
```
Shaddoll, line 281: it can be removed

davidporter-id-au: nice catch

Shaddoll (approved):

PR #5280, opened 2023-05-11
Update matching to support tasklist isolation

Shaddoll:
**What changed?**
Update matching to support tasklist isolation
- For domain with tasklist isolation enabled, tasks are only assigned to specific pollers from one isolation group which is calculated from partition config by using a partitioner library

**Why?**
To support worker isolation

**How did you test it?**
new test cases

**Potential risks**
If the feature is enabled, losing pollers from one isolation group can block the task processing of other isolation groups

**Release notes**

**Documentation Changes**


service/matching/pollerHistory.go
```diff
@@ -91,13 +89,39 @@ func (pollers *pollerHistory) getPollerInfo(earliestAccessTime time.Time) []*typ
...
+		entry := ite.Next()
+		value := entry.Value().(*pollerInfo)
+		lastAccessTime := entry.CreateTime()
+		if earliestAccessTime.Before(lastAccessTime) {
+			if value.isolationGroup != "" {
+				groupSet[value.isolationGroup] = struct{}{}
+			}
+		}
+	}
+	var result []string
```
davidporter-id-au, line 121: nit: Known size ahead of time `make([]string, len(groupSet))`

service/matching/pollerHistory.go
```diff
@@ -91,13 +89,39 @@ func (pollers *pollerHistory) getPollerInfo(earliestAccessTime time.Time) []*typ
 		// TODO add IP, T1396795
 		lastAccessTime := entry.CreateTime()
 		if earliestAccessTime.Before(lastAccessTime) {
+			rps := _defaultTaskDispatchRPS
+			if value.ratePerSecond != nil {
```
davidporter-id-au, line 93: I don't understand this - why did it get changed?

service/matching/taskListManager.go
```diff
@@ -534,6 +562,61 @@ func (c *taskListManagerImpl) isFowardingAllowed(taskList *taskListID, kind type
...
+	return c.taskListKind != types.TaskListKindSticky && c.enableIsolation
+}
+
+func (c *taskListManagerImpl) shouldReload() bool {
+	return c.config.EnableTasklistIsolation() != c.enableIsolation && c.taskListKind != types.TaskListKindSticky
+}
+
+func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, taskInfo *persistence.TaskInfo) (string, error) {
+	if c.enableIsolation && len(taskInfo.PartitionConfig) > 0 {
+		partitionConfig := make(map[string]string)
```
davidporter-id-au, line 575: nit: length known ahead of time: `make(map[string]string, len(taskinfo.PartitionConfig))`

davidporter-id-au (approved):

service/matching/matcher.go
```diff
@@ -217,18 +228,21 @@ func (tm *TaskMatcher) MustOffer(ctx context.Context, task *InternalTask) error
...
-	case tm.taskC <- task:
+	case taskC <- task: // poller picked up the task
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
 	default:
 	}
 
+	// TODO: we need a way to get the notification of the draining events of isolation groups,
+	// otherwise the loop will get stuck forever because the context doesn't have any deadline
```
davidporter-id-au, line 241: Is this comment still true?

service/matching/pollerHistory.go
Shaddoll: because the default value is overwritten by a user-provided value

service/matching/matcher.go
Shaddoll: no, deleted

PR #5287, opened 2023-05-12
add more logs around unsupported version for consistent query

allenchen2244:
**What changed?**
Add some more logs since this is happening and we don't have any idea why.

**Why?**
Try and get to bottom of this issue.

**How did you test it?**
Unit test pass

**Potential risks**
nothing

### DIFF
--- a/service/history/decision/handler.go
+++ b/service/history/decision/handler.go
@@ -730,6 +730,8 @@ func (handler *handlerImpl) handleBufferedQueries(
 		buffered := queryRegistry.GetBufferedIDs()
 		handler.logger.Info(
 			"failing query because worker does not support consistent query",
+			tag.Dynamic("clientImpl", clientImpl),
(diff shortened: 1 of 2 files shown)


davidporter-id-au (approved):

service/history/decision/handler.go
```diff
@@ -730,6 +730,8 @@ func (handler *handlerImpl) handleBufferedQueries(
 		buffered := queryRegistry.GetBufferedIDs()
 		handler.logger.Info(
 			"failing query because worker does not support consistent query",
+			tag.Dynamic("clientImpl", clientImpl),
+			tag.Dynamic("clientFeatureVersion", clientFeatureVersion),
```
davidporter-id-au, line 734: did you want to make this a proper tag maybe? also I think we use kebab-case with logs `client-feature-version` maybe a string tag?

PR #5289, opened 2023-05-16
[CDNC-3578] Worklfow start metric

ketsiambaku:
**What changed?**
Add counter metric and emit when a workflow starts

**Why?**
The metric was missing

**How did you test it?**
Tested on staging

**Potential risks**
breaking workflowStartedCount metric

**Release notes**

**Documentation Changes**


neil-xie:
qucik question: do we need to update transfer_standby_task_executor.go, since it also has the processRecordWorkflowStartedOrUpsertHelper method and can record workflow started

ketsiambaku:
> qucik question: do we need to update transfer_standby_task_executor.go, since it also has the processRecordWorkflowStartedOrUpsertHelper method and can record workflow started

@neil-xie Done! thanks for catching that.

allenchen2244 (approved):

PR #5291, opened 2023-05-18
Improvements on tasklist isolation

Shaddoll:
**What changed?**
- Increase the look-back window for getting poller isolation groups
- Update task reader to not set timeout when dispatching async tasks
- Add debug log for getting isolation group for tasks

**Why?**
- Reduce the probability of task leak for the root partition of scalable tasklists
- Prevent task leak for non-root partition of scalable tasklists

**How did you test it?**
tested in dev environment

**Potential risks**

**Release notes**

**Documentation Changes**


service/matching/taskListManager.go
```diff
@@ -581,9 +581,9 @@ func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, task
 		// Not all poller information are available at the time of task list manager creation,
 		// because we don't persist poller information in database, so in the first minute, we always assume
 		// pollers are available in all isolation groups to avoid the risk of leaking a task to another isolation group.
-		// Besides, for sticky tasklists, not all poller information are available, we also use all isolation group.
-		if time.Now().Sub(c.createTime) > time.Minute && c.taskListKind != types.TaskListKindSticky {
-			pollerIsolationGroups = c.pollerHistory.getPollerIsolationGroups(time.Now().Add(-time.Minute))
+		// Besides, for sticky and scalable tasklists, not all poller information are available, we also use all isolation group.
+		if time.Now().Sub(c.createTime) > time.Minute && c.taskListKind != types.TaskListKindSticky && c.taskListID.IsRoot() {
+			pollerIsolationGroups = c.pollerHistory.getPollerIsolationGroups(time.Time{}) // the lookback window must be larger than the timeout of poller requests (2 mins), otherwise we don't get all pollers
```
davidporter-id-au, line 586: ok, but why should it be... this big?  I don't follow how you've gone from a minute ago to effectively infinite?

service/matching/taskReader.go
```diff
@@ -361,7 +367,15 @@ func (tr *taskReader) completeTask(task *persistence.TaskInfo, err error) {
 
 func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context, context.CancelFunc) {
 	if isolationGroup != "" {
-		return context.WithTimeout(tr.cancelCtx, tr.config.AsyncTaskDispatchTimeout())
+		domainEntry, err := tr.domainCache.GetDomainByID(tr.taskListID.domainID)
+		if err != nil {
+			// we don't know if the domain is active in the current cluster, assume it is active and set the timeout
+			return context.WithTimeout(tr.cancelCtx, tr.config.AsyncTaskDispatchTimeout())
```
davidporter-id-au, line 373: why is the timeout only set on the active side?

service/matching/taskListManager.go
Shaddoll: the poller cache's TTL is 5 mins, so it's just the same as the cache TTL

service/matching/taskReader.go
Shaddoll: because the passive side is not actually polling

davidporter-id-au: If it isn't, why do we care about differentiating it then?

davidporter-id-au (approved):

PR #5293, opened 2023-05-18
Update ClusterNameForFailoverVersion to return error

Shaddoll:
**What changed?**
Update ClusterNameForFailoverVersion to return error instead of panicking

**Why?**
To prevent server from crashloop

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -4347,7 +4347,10 @@ func (e *mutableStateBuilder) eventsToReplicationTask(
 	lastEvent := events[len(events)-1]
 	version := firstEvent.Version
 
-	sourceCluster := e.clusterMetadata.ClusterNameForFailoverVersion(version)
+	sourceCluster, err := e.clusterMetadata.ClusterNameForFailoverVersion(version)
+	if err != nil {
+		return nil, err
+	}
 	currentCluster := e.clusterMetadata.GetCurrentClusterName()
 
 	if currentCluster != sourceCluster {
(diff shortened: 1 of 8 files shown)


service/history/execution/workflow.go
```diff
@@ -182,7 +182,10 @@ func (r *workflowImpl) SuppressBy(
 		return TransactionPolicyPassive, nil
 	}
 
-	lastWriteCluster := r.clusterMetadata.ClusterNameForFailoverVersion(lastWriteVersion)
+	lastWriteCluster, err := r.clusterMetadata.ClusterNameForFailoverVersion(lastWriteVersion)
+	if err != nil {
+		return TransactionPolicyActive, err
```
davidporter-id-au, line 187: Without knowing this implementation well it's hard to know exactly what the correct behaviour is, but I guess this is in line with the other error handling

service/history/historyEngine.go
```diff
@@ -498,12 +498,13 @@ func (e *historyEngineImpl) registerDomainFailoverCallback() {
 				domainFailoverNotificationVersion := nextDomain.GetFailoverNotificationVersion()
 				domainActiveCluster := nextDomain.GetReplicationConfig().ActiveClusterName
 				previousFailoverVersion := nextDomain.GetPreviousFailoverVersion()
+				previousClusterName, _ := e.clusterMetadata.ClusterNameForFailoverVersion(previousFailoverVersion)
```
davidporter-id-au, line 501: I would not suppress the error here and log it at the very least. If I'm reading this correctly the callback will be messed up

service/history/historyEngine.go
```diff
@@ -599,10 +600,11 @@ func (e *historyEngineImpl) newDomainNotActiveError(
 	failoverVersion int64,
 ) error {
 	clusterMetadata := e.shard.GetService().GetClusterMetadata()
+	clusterName, _ := clusterMetadata.ClusterNameForFailoverVersion(failoverVersion)
```
davidporter-id-au, line 603: you don't want to pass the error back?

service/history/historyEngine.go
Shaddoll: you're right

service/history/historyEngine.go
Shaddoll: the error is not really useful

davidporter-id-au (approved):

service/history/execution/mutable_state_builder.go
```diff
@@ -4508,8 +4511,14 @@ func (e *mutableStateBuilder) startTransactionHandleDecisionFailover(
 		)}
 	}
 
-	lastWriteSourceCluster := e.clusterMetadata.ClusterNameForFailoverVersion(lastWriteVersion)
-	currentVersionCluster := e.clusterMetadata.ClusterNameForFailoverVersion(currentVersion)
+	lastWriteSourceCluster, err := e.clusterMetadata.ClusterNameForFailoverVersion(lastWriteVersion)
+	if err != nil {
+		return false, err
```
Groxx, line 4516: since I'm guessing these will all be hard to tell apart in logs: probably worth adding some descriptive wrapping (and it's safe to wrap here regardless of existing error checks, because this is a new category of errors for this code)

PR #5295, opened 2023-05-19
Feature/isolation group library logging improvements

davidporter-id-au:
- Adds debug logs to the partitioner
- Adds metric counters around isolation group selection based on their being either healthy, missing pollers or drained
- includes some minor refactoring in the isolation-group library to make it clearer what was going on with the selection of groups. 

Example metrics query: 

```
fetch  domain:somedomain operation:getavailableisolationgroups name:isolation_group_healthy| sum poller_isolation_group | moving 1m sum ```

### DIFF
--- a/common/isolationgroup/defaultisolationgroupstate/state.go
+++ b/common/isolationgroup/defaultisolationgroupstate/state.go
@@ -87,7 +87,7 @@ func (z *defaultIsolationGroupStateHandler) AvailableIsolationGroupsByDomainID(c
 		return nil, fmt.Errorf("unable to get isolation group state: %w", err)
 	}
 	availableIsolationGroupsCfg := isolationGroupHealthyListToConfig(availableIsolationGroups)
-	scope := z.createAvailableisolationGroupMetricsScope(domainID, availableIsolationGroups)
+	scope := z.createAvailableisolationGroupMetricsScope(domainID)
 	return availableIG(z.config.AllIsolationGroups, availableIsolationGroupsCfg, state.Global, state.Domain, scope), nil
 }
 
@@ -169,12 +169,9 @@ func (z *defaultIsolationGroupStateHandler) get(ctx context.Context, domain stri
 	return ig, nil
 }
 
-func (z *defaultIsolationGroupStateHandler) createAvailableisolationGroupMetricsScope(domainID string, availableIsolationGroups []string) metrics.Scope {
-	domainName, err := z.domainCache.GetDomainName(domainID)
-	if err != nil {
-		domainName = "unknown-domain"
-	}
-	return z.metricsClient.Scope(metrics.DefaultIsolationGroupStateHandlerGetAvailableIsolationGroupsScope).Tagged(metrics.DomainTag(domainName))
+func (z *defaultIsolationGroupStateHandler) createAvailableisolationGroupMetricsScope(domainID string) metrics.Scope {
+	domainName, _ := z.domainCache.GetDomainName(domainID)
+	return z.metricsClient.Scope(metrics.GetAvailableIsolationGroupsScope).Tagged(metrics.DomainTag(domainName))
 }
 
 // A simple explicit deny-based isolation group implementation
--- a/common/isolationgroup/defaultisolationgroupstate/state.go
+++ b/common/isolationgroup/defaultisolationgroupstate/state.go
@@ -187,11 +187,6 @@ func availableIG(
 		_, hasAvailablePollers := availablePollers[isolationGroup]
 		globalCfg, hasGlobalConfig := global[isolationGroup]
 		domainCfg, hasDomainConfig := domain[isolationGroup]
-		if !hasAvailablePollers {
-			// we don't attempt to dispatch tasks to isolation groups where there are no pollers
-			scope.Tagged(metrics.PollerIsolationGroupTag(isolationGroup)).IncCounter(metrics.IsolationGroupStatePollerUnavailable)
-			continue
-		}
 		if hasGlobalConfig {
 			if globalCfg.State == types.IsolationGroupStateDrained {
 				scope.Tagged(metrics.PollerIsolationGroupTag(isolationGroup)).IncCounter(metrics.IsolationGroupStateDrained)
@@ -203,6 +198,11 @@ func availableIG(
 				scope.Tagged(metrics.PollerIsolationGroupTag(isolationGroup)).IncCounter(metrics.IsolationGroupStateDrained)
 				continue
 			}
+		}
+		if !hasAvailablePollers {
+			// we don't attempt to dispatch tasks to isolation groups where there are no pollers
+			scope.Tagged(metrics.PollerIsolationGroupTag(isolationGroup)).IncCounter(metrics.IsolationGroupStatePollerUnavailable)
+			continue
 		}
 		scope.Tagged(metrics.PollerIsolationGroupTag(isolationGroup)).IncCounter(metrics.IsolationGroupStateHealthy)
 		out[isolationGroup] = types.IsolationGroupPartition{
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: common/partition/default-partitioner.go [hunk 1]
```diff
@@ -123,7 +130,14 @@ func pickIsolationGroup(wfPartition defaultWorkflowPartitionConfig, available ty
 	sort.Slice(availableList, func(i int, j int) bool {
 		return availableList[i] > availableList[j]
 	})
-	return pickIsolationGroupFallback(availableList, wfPartition)
+	r.metrics.IncCounter(metrics.PartitionerUnPinnedTask)
+	fallback := pickIsolationGroupFallback(availableList, wfPartition)
+	r.log.Debug("isolation group falling back to an available zone:",
```
Shaddoll, line 135: remove ':'
and we should log `partitionConfig`,
and there is a tag for available isolation groups

common/metrics/defs.go
```diff
@@ -1800,6 +1802,7 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		MatchingDescribeTaskListScope:          {operation: "DescribeTaskList"},
 		MatchingListTaskListPartitionsScope:    {operation: "ListTaskListPartitions"},
 		MatchingGetTaskListsByDomainScope:      {operation: "GetTaskListsByDomain"},
+		MatchingPartitionerDefaultPartitioner:  {operation: "Partitioner"},
```
Shaddoll, line 1805: the value of operation should be something like `GetIsolationGroup`

common/partition/default-partitioner.go
davidporter-id-au: good point, thank you

common/isolationgroup/defaultisolationgroupstate/state.go
```diff
@@ -163,22 +169,45 @@ func (z *defaultIsolationGroupStateHandler) get(ctx context.Context, domain stri
 	return ig, nil
 }
 
+func (z *defaultIsolationGroupStateHandler) createAvailableisolationGroupMetricsScope(domainID string, availableIsolationGroups []string) metrics.Scope {
```
davidporter-id-au, line 172: I think adding the tasklist ID being passed in here would be simple enough and provide that additional dimension, should we need it (which think we do)

common/isolationgroup/defaultisolationgroupstate/state.go
```diff
@@ -81,8 +86,9 @@ func (z *defaultIsolationGroupStateHandler) AvailableIsolationGroupsByDomainID(c
 	if err != nil {
 		return nil, fmt.Errorf("unable to get isolation group state: %w", err)
 	}
-	isolationGroups := common.IntersectionStringSlice(z.config.AllIsolationGroups, availableIsolationGroups)
-	return availableIG(isolationGroups, state.Global, state.Domain), nil
```
davidporter-id-au, line 85: I moved this into inside the `availableIG` function for clarity about the decision steps that are occurring

common/log/tag/tags.go
```diff
@@ -927,3 +929,11 @@ func PartitionConfig(p map[string]string) Tag {
 func PollerGroups(pollers []string) Tag {
 	return newObjectTag("poller-isolation-groups", pollers)
 }
+
+func FallbackIsolationGroup(group string) Tag {
+	return newStringTag("fallback-isolation-group", group)
+}
+
+func PollerGroupsConfiguration(pollers types.IsolationGroupConfiguration) Tag {
```
Shaddoll, line 937: It's a duplicate of `PollerGroups`

davidporter-id-au: no, this is covering `types.IsolationGroupConfiguration` which is how most if the isolation groups are represented internally

common/isolationgroup/defaultisolationgroupstate/state.go
```diff
@@ -163,22 +169,45 @@ func (z *defaultIsolationGroupStateHandler) get(ctx context.Context, domain stri
 	return ig, nil
 }
 
+func (z *defaultIsolationGroupStateHandler) createAvailableisolationGroupMetricsScope(domainID string, availableIsolationGroups []string) metrics.Scope {
+	domainName, err := z.domainCache.GetDomainName(domainID)
+	if err != nil {
+		domainName = "unknown-domain"
```
Shaddoll, line 175: DomainTag handles empty string properly, so remove this logic and just use empty string

common/metrics/defs.go
```diff
@@ -1565,6 +1567,8 @@ var ScopeDefs = map[ServiceIdx]map[int]scopeDefinition{
 		BlobstoreClientDeleteScope:          {operation: "BlobstoreClientDelete", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
 		BlobstoreClientDirectoryExistsScope: {operation: "BlobstoreClientDirectoryExists", tags: map[string]string{CadenceRoleTagName: BlobstoreRoleTagValue}},
 
+		DefaultIsolationGroupStateHandlerGetAvailableIsolationGroupsScope: {operation: "DefaultIsolationGroupStateHandlerGetAvailableIsolationGroups"},
```
Shaddoll, line 1570: The name is bad. Don't put `DefaultIsolationGroupStateHandler` here. If you want to have this information, create another tag.

common/metrics/defs.go
```diff
@@ -2326,6 +2334,8 @@ const (
 	TaskListManagersGauge
 	TaskLagPerTaskListGauge
 	TaskBacklogPerTaskListGauge
+	PartitionerPinnedTask
```
Shaddoll, line 2337: not being used anywhere

common/isolationgroup/defaultisolationgroupstate/state.go
```diff
@@ -163,22 +169,45 @@ func (z *defaultIsolationGroupStateHandler) get(ctx context.Context, domain stri
...
+	global types.IsolationGroupConfiguration,
+	domain types.IsolationGroupConfiguration,
+	scope metrics.Scope,
+) types.IsolationGroupConfiguration {
 	out := types.IsolationGroupConfiguration{}
 	for _, isolationGroup := range allIsolationGroups {
+		_, hasAvailablePollers := availablePollers[isolationGroup]
 		globalCfg, hasGlobalConfig := global[isolationGroup]
 		domainCfg, hasDomainConfig := domain[isolationGroup]
+		if !hasAvailablePollers {
```
Shaddoll, line 193: check drain before poller. Because we block poller requests in frontend if it's drained

davidporter-id-au: This function isn't used in the `IsDrained`, it doesn't make sense to look at where pollers are in blocking requests in frontend because there's no pollers there.

common/metrics/defs.go
davidporter-id-au: true, thank you!

common/metrics/defs.go
davidporter-id-au: yeah, good point, I'm blunding around a bit here with naming

common/isolationgroup/defaultisolationgroupstate/state.go
Shaddoll: If an isolation group is drained, you won't see `IsolationGroupStateDrained` metric. You'll just see `IsolationGroupStatePollerUnavailable` metric.

davidporter-id-au: oh, I misunderstood your point. Yeah, that's a fair point

common/log/tag/tags.go
Shaddoll: do you mind using a different key? "poller-isolation-groups" is used by another tag

PR #5310, opened 2023-06-02
Increase number of forward tokens for isolation tasklists

Shaddoll:
**What changed?**
Increase number of forward tokens for isolation tasklists

**Why?**
Fix the latency degradation for scalable tasklists with isolation

**How did you test it?**
staging2 test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/forwarder.go
+++ b/service/matching/forwarder.go
@@ -57,6 +57,8 @@ type (
 		// todo: implement a rate limiter that automatically
 		// adjusts rate based on ServiceBusy errors from API calls
 		limiter *quotas.DynamicRateLimiter
+
+		isolationGroups []string
 	}
 	// ForwarderReqToken is the token that must be acquired before
 	// making forwarder API calls. This type contains the state
@@ -91,6 +93,7 @@ func newForwarder(
 	taskListID *taskListID,
 	kind types.TaskListKind,
 	client matching.Client,
+	isolationGroups []string,
 ) *Forwarder {
 	rpsFunc := func() float64 { return float64(cfg.ForwarderMaxRatePerSecond()) }
 	fwdr := &Forwarder{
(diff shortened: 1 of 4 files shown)


service/matching/forwarder.go
```diff
@@ -91,19 +93,21 @@ func newForwarder(
 	taskListID *taskListID,
 	kind types.TaskListKind,
 	client matching.Client,
+	isolationGroups []string,
```
davidporter-id-au, line 96: Is this intended to be that private-package type?

service/matching/forwarder.go
```diff
@@ -91,19 +93,21 @@ func newForwarder(
...
 ) *Forwarder {
 	rpsFunc := func() float64 { return float64(cfg.ForwarderMaxRatePerSecond()) }
 	fwdr := &Forwarder{
 		cfg:                   cfg,
 		client:                client,
 		taskListID:            taskListID,
 		taskListKind:          kind,
-		outstandingTasksLimit: int32(cfg.ForwarderMaxOutstandingTasks()),
-		outstandingPollsLimit: int32(cfg.ForwarderMaxOutstandingPolls()),
+		outstandingTasksLimit: int32(cfg.ForwarderMaxOutstandingTasks() * (len(isolationGroups) + 1)),
```
davidporter-id-au, line 104: Mind adding a comment here as to the intuition of this?

service/matching/forwarder.go
```diff
@@ -57,6 +57,8 @@ type (
 		// todo: implement a rate limiter that automatically
 		// adjusts rate based on ServiceBusy errors from API calls
 		limiter *quotas.DynamicRateLimiter
+
+		isolationGroups []string
```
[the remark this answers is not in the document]
Shaddoll: it is used

service/matching/forwarder.go
Shaddoll: it's allisolationgroups if isolation is enabled, it's nil if not

service/matching/forwarder.go
Shaddoll: there is an intuition in line 264

service/matching/forwarder.go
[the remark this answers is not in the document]
davidporter-id-au: sorry, I was tired... this is not a type declaration like I thought it was, it's a field on a struct 🤦

davidporter-id-au (approved):

davidporter-id-au (approved):

PR #5313, opened 2023-06-02
Fixing consistency

davidporter-id-au:
Hotfix for issue where schema checks seem to be breaking on service start due to using consistency-level ALL under some scenarios


Shaddoll (commented):
Provision tool also uses this. I don't think it's the correct change

PR #5315, opened 2023-06-04
Bugfix/isolation groups domain drains

davidporter-id-au:
**What changed?**
Fixes several bugs in the domain isolation-group handling which wasn't tested properly in replication it notably:
- Fixes the problem of upserting configuration from the inactive region, which was previously would error
- Fixes the problem of replication of configuration, which was entirely not working
- Refactors the domain controller by splitting out this functionality into its own, much simpler function rather than continuing to overload the already incomprehensible domain controller. 

**Why?**

**How did you test it?**
```
cadence --env docstore-prod11 --proxy_region dca admin isolation-groups get-domain --domain cadence-canary-global
Isolation Groups State
asdf5            Drained
asfd             Drained
-------------------------------------------------------------------------------------------------------------------------------------------------------------
~ » cadence --env an-env --proxy_region dca admin isolation-groups update-domain --domain cadence-canary-global  --remove-all-drains
-------------------------------------------------------------------------------------------------------------------------------------------------------------
~ » cadence --env an-env --proxy_region dca admin isolation-groups get-domain --domain cadence-canary-global
-- No groups found --
-------------------------------------------------------------------------------------------------------------------------------------------------------------
~ » cadence --env an-env --proxy_region phx admin isolation-groups get-domain --domain cadence-canary-global
-- No groups found --
```

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/domain/handler.go
+++ b/common/domain/handler.go
@@ -71,6 +71,10 @@ type (
 			ctx context.Context,
 			updateRequest *types.UpdateDomainRequest,
 		) (*types.UpdateDomainResponse, error)
+		UpdateIsolationGroups(
+			ctx context.Context,
+			updateRequest types.UpdateDomainIsolationGroupsRequest,
+		) error
 	}
 
 	// handlerImpl is the domain operation handler implementation
@@ -400,24 +404,6 @@ func (d *handlerImpl) UpdateDomain(
 	visibilityArchivalConfigChanged := false
 	// whether active cluster is changed
 	activeClusterChanged := false
-
-	// Whether isolation-groups are changed. There's semantic API difference between the
-	// domain isolation-group update API which is a flat upsert and this handler API which
-	// relies on modifying specified values only, and leaving the rest as-is.
-	// for callers of the Domain API wishing to update isolation groups,
-	// To distinguish between the operation to remove all drains and
-	// take no action, the UpdateDomain handler uses the presence or nil-ness of the IsolationGroup
-	// field to indicate either drain removal (where the struct is present), and to take no action
-	// where the field is nil.
-	//
-	// ie, for the purposes of this handler:
-	// h.UpdateDomain(ctx, UpdateDomainRequest{
-	//   IsolationGroups: nil,                              // take no action for isolation groups
-	// })
-	// h.UpdateDomain(ctx, UpdateDomainRequest{
-	//   IsolationGroups: IsolationGroupConfiguration{},    // remove all isolation groups for domain
-	// })
-	config, isolationGroupConfigurationChanged, err := d.getIsolationGroupStatus(config, updateRequest)
 	// whether anything other than active cluster is changed
 	configurationChanged := false
 
@@ -502,7 +488,7 @@ func (d *handlerImpl) UpdateDomain(
 		previousFailoverVersion = failoverVersion
 	}
 
-	configurationChanged = historyArchivalConfigChanged || visibilityArchivalConfigChanged || domainInfoChanged || domainConfigChanged || deleteBinaryChanged || replicationConfigChanged || isolationGroupConfigurationChanged
+	configurationChanged = historyArchivalConfigChanged || visibilityArchivalConfigChanged || domainInfoChanged || domainConfigChanged || deleteBinaryChanged || replicationConfigChanged
 
 	if err := d.domainAttrValidator.validateDomainConfig(config); err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


common/domain/replicationTaskExecutor.go
```diff
@@ -192,12 +194,14 @@ func (h *domainReplicationTaskExecutorImpl) handleDomainUpdateReplicationTask(ct
 	// task already validated
 	status, err := h.convertDomainStatusFromThrift(task.Info.Status)
 	if err != nil {
+		h.logger.Info("Error handling replication task", tag.Dynamic("task", task), tag.Error(err))
```
Shaddoll, line 197: we should avoid using Dynamic

common/domain/replicationTaskExecutor.go
```diff
@@ -262,8 +268,10 @@ func (h *domainReplicationTaskExecutorImpl) handleDomainUpdateReplicationTask(ct
 	}
 
 	if !recordUpdated {
+		h.logger.Warn("Domain update failed, record not updated", tag.Dynamic("replication-request", request))
```
Shaddoll, line 271: this is an expected path, the log should be removed

common/domain/handler.go
```diff
@@ -669,6 +658,111 @@ func (d *handlerImpl) DeprecateDomain(
...
+// UpdateIsolationGroups is used for draining and undraining of isolation-groups for a domain.
+// Like the isolation-group API, this controller expects Upsert semantics for
+// isolation-groups and does not modify any other domain information.
+//
+// Isolation-groups are regional in their configuration scope, so it's expected that this upsert
+// includes configuration for both clusters every time.
+//
+// The update is handled like other domain updates in that they expected to be replicated. So
+// unlike the global isolation-group API it shouldn't be necessary to call
+func (d *handlerImpl) UpdateIsolationGroups(
```
Shaddoll, line 670: The refactoring is not a good idea. It introduces duplicate logic and the most important thing is that replication is using `UpdateDomain`

common/domain/replicationTaskExecutor.go
davidporter-id-au: Is.. it? I'd actually be surprised to see conditional failures like this

davidporter-id-au: Yes, I understand that, but what I was meaning to say was that conditional update failures are surprising, probably shouldn't happen unless something's not getting processed, unless I'm confused? 

You cannot trigger a domain update more than 30 seconds apart, so actually hitting this codepath seems... surprising

common/domain/handler.go
```diff
@@ -669,6 +658,115 @@ func (d *handlerImpl) DeprecateDomain(
...
+	if lastUpdatedTime.Add(d.config.FailoverCoolDown(currentDomainConfig.Info.Name)).After(d.timeSource.Now()) {
+		return errDomainUpdateTooFrequent
+	}
+
+	// Check the failover cool down time
+	if lastUpdatedTime.Add(d.config.FailoverCoolDown(currentDomainConfig.Info.Name)).After(d.timeSource.Now()) {
+		return errDomainUpdateTooFrequent
+	}
+
+	if !d.clusterMetadata.IsPrimaryCluster() {
```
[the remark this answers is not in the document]
Shaddoll: only check for global domain

common/domain/replicationTaskExecutor.go
```diff
@@ -262,6 +267,7 @@ func (h *domainReplicationTaskExecutorImpl) handleDomainUpdateReplicationTask(ct
 	}
 
 	if !recordUpdated {
+		h.logger.Warn("Domain update failed, record not updated")
```
Shaddoll, line 270: i still don't think we should add this log

davidporter-id-au: What is your concern? I don't see how this is a noisy log

Shaddoll: The message is misleading and it's asking people to investigate a non-issue.

davidporter-id-au: Maybe my mental model is wrong, under what conditions should a conditional update normally occur in this codepath? I get domain update queue replays, but that seems fairly rare

Shaddoll: I don't know the details, but it probably could occur when merging dlq.

common/domain/handler.go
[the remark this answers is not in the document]
davidporter-id-au: good call

common/domain/replicationTaskExecutor.go
```diff
@@ -198,6 +200,7 @@ func (h *domainReplicationTaskExecutorImpl) handleDomainUpdateReplicationTask(ct
 	// first we need to get the current notification version since we need to it for conditional update
 	metadata, err := h.domainManager.GetMetadata(ctx)
 	if err != nil {
+		h.logger.Info("Error getting metadata while handling replication task", tag.Error(err))
```
Shaddoll, line 203: nit: Error

Shaddoll (approved):

common/domain/replicationTaskExecutor.go
davidporter-id-au: thank you

PR #5322, opened 2023-06-14
changed bench workers to use grpc port instead of tchannel

MrBoss276:
Made bench workers to use grpc instead of tchannel transport, and changed development.yaml in config/bench to use 7833 port.

Since the thrift port is getting deprecated, it reduces the overhead of exposing thrift port of cadence service while bench testing.

I tested this locally and workers started, All test cases passed. Since its a small change I didn't add test case.

There is no risks for main code as it only changes bench client connection.

No it doesn't require anything.

No changes required


Shaddoll (approved):

taylanisikdemir:
Closing the PR as part of old PR cleanup effort. If still applicable/needed please open a new PR.

PR #5323, opened 2023-06-14
Create hardware usage tracking for matching service

timl3136:
**What changed?**
Add hardware usage tracking for matching hosts

**Why?**
We need a way to monitor customer utilization on the matching service in order to improve efficiency

**How did you test it?**
Tested on dev environment

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au:
The changes broadly look like they work, and I do think we need to do some more of this regarding per-domain/host based limits like this. But I'm a bit confused by the intent of these specifically? 

What do we intend to understand by emitting per-domain metrics around matching's CPU and memory? This is a shared resource so it doesn't exactly make sense to me to domain as a dimension? Like the memory might be high for all sorts of reasons, and the fact that you're looking at a domain's view is kind of... not related? Am I missing something?

davidporter-id-au (approved):

taylanisikdemir:
Closing as part of old PR cleanup effort. Feel free to create a new PR from your fork branch when/if this gets picked up again.

PR #5339, opened 2023-07-07
upgrade mysql version

bowenxia:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/docker/dev/mysql.yml
+++ b/docker/dev/mysql.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   mysql:
-    image: mysql:5.7
+    image: mysql:latest
     ports:
       - "3306:3306"
     environment:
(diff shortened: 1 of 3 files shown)


docker/dev/mysql.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   mysql:
-    image: mysql:5.7
+    image: mysql:latest
```
3vilhamster, line 4: I would suggest using a specific version like 8.0.3.
Using latest can cause issues if mysql team bumps latest tag to a new version.

bowenxia: Yes you are right. I've merged a new PR with the 8.0 version. This PR should be abandoned.

PR #5340, opened 2023-07-11
Validate search-attribute-key so keys are fine in advanced search

dkrotx:
ElasticSearch engine fails to parse keys with dashes, making those
search-attributes broken for CLI and web.

The reason it is not in the server, is that currently even at Uber we
have keys with dashes, and it will be a problem if we gonna migrate
domains between clusters.

Validate search-attribute before creating on server so it looks like a
valid variable name in most of languages, should be more than expected for
programmers.

Stakeholders faced with the issue in the UI.
We spent time investingating why it is not working.

Unit tests.

It is change to CLI, no risks of breaking production.

Nope

optionally - update description of KeyName here: https://cadenceworkflow.io/docs/operation-guide/setup/#dynamic-configuration

### DIFF
--- a/tools/cli/adminClusterCommands.go
+++ b/tools/cli/adminClusterCommands.go
@@ -23,6 +23,7 @@ package cli
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 
 	"github.com/fatih/color"
 	"github.com/pborman/uuid"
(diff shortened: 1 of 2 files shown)


tools/cli/adminClusterCommands.go
```diff
@@ -36,10 +37,15 @@ import (
 
 // An indirection for the prompt function so that it can be mocked in the unit tests
 var promptFn = prompt
+var validSearchAttributeKey = regexp.MustCompile(`^[a-zA-Z][a-zA-Z_0-9]*$`)
 
 // AdminAddSearchAttribute to whitelist search attribute
 func AdminAddSearchAttribute(c *cli.Context) {
 	key := getRequiredOption(c, FlagSearchAttributesKey)
+	if !isValidSearchAttributeKey(key) {
+		ErrorAndExit("Invalid search-attribute key.", nil)
```
3vilhamster, line 46: Could you provide more context here with what is wrong with the key? It is not obvious what it allowed and what failed the validation.

3vilhamster (approved):

PR #5343, opened 2023-07-13
Ignore .gen/ directory for goimports

dkrotx:
Otherwise there are 45 changed files after any invocation of `make`.
While goimports can change at any point, the only change affecting
cadence source from Go v1.17 is that goimports re-tabs commented
Go-code.

Make won't run goimports for .gen/ directory

Otherwise there are 45 changed files after any invocation of `make`.

**How did you test it?**

No any. I also highly in doubt we need to reformat generated code.

**Release notes**

**Documentation Changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,6 +134,7 @@ FRESH_ALL_SRC = $(shell \
 #
 # if you require a fully up-to-date list, e.g. for shell commands, use FRESH_ALL_SRC instead.
 ALL_SRC := $(FRESH_ALL_SRC)
+ALL_BUT_GEN_SRC := $(filter-out ./.gen/%, $(ALL_SRC))
 # as lint ignores generated code, it can use the cached copy in all cases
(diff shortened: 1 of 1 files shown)


Groxx:
if there are changes, commit them?  and double check that you're using the right Go version, as that can affect formatting, and not just in `.gen`.

goimports _doesn't_ change at any point.  it's a specific version, which the makefile controls: https://github.com/uber-go/cadence-client/blob/master/Makefile#L146

dkrotx:
@Groxx changed a bit. I think depends on go version, isn't it? Otherwise why it's one format on github and another on local machine with go 1.20?
postgresql-related tests are failing.
If you OK with it, can you merge?

Groxx:
It does, but we have a stated "dev" version, and that's checked in the makefile (printed every time), because it needs to match CI for stuff like this. Which affects everything, not just .gen.

Our options are basically "use the correct Go version" or "ignore formatting, and ignore automated reformatting in PRs, because it's gonna happen a lot".

taylanisikdemir:

PR #5346, opened 2023-07-13
[dynamic config] add Filters method to dynamic config Key

shijiesheng:
**What changed?**

* Add Filters method to `Key` interface
* Add implementations on most keys by parsing the comments on keys (assuming they are correct)

**Why?**

This is needed to know what dynamic config is domain specific. And this could possible simplify the collection struct by consolidating all GetPropertyFilterBy** methods.

**How did you test it?**

**Potential risks**

no risk since the method will be read only in non-critical path

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -92,6 +92,8 @@ type (
 		String() string
 		Description() string
 		DefaultValue() interface{}
+		// Filters is used to identify what filters a DynamicConfig key may have.
+		// For example, CLI tool uses this to figure out all domain specific configurations for migration validation.
 		Filters() []Filter
 	}
 )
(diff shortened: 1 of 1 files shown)


Groxx:
Broadly makes sense and I believe the comments are like 95%+ correct...

But since this seems like half of an idea: what's the goal? It seems fine to include it, but it's not doing anything in this PR. Seems like it needs a use somewhere, or at least a comment in code about what the intended _external_ use would be, so it's more likely to be maintained in the long run.

Groxx (approved):
Yea, with the comment I think it's a bit more clear that it's not (currently anyway) intended for validation, mostly just informational.

Which seems useful and fine.  👍 

Validation might be nice some day, but that'll need a lot more scrutiny around correctness / how people extend with custom config / careful rollout.

PR #5352, opened 2023-07-18
Enforce TLS for all inbounds if configured

mantas-sidlauskas:
**What changed?**
Set up TLS for all inbounds

**Why?**
Previously, only gRPC inbound supported TLS, but with newer yarpc version, support for all transports is available. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/rpc/factory.go
```diff
@@ -47,19 +48,26 @@ type Factory struct {
...
+		tchannel.ServiceName(p.ServiceName),
+		tchannel.ListenAddr(p.TChannelAddress),
+	}
+
+	if p.InboundTLS != nil {
+		transportOptions = append(transportOptions,
+			tchannel.InboundTLSConfiguration(p.InboundTLS),
+			tchannel.InboundTLSMode(yarpctls.Enforced),
+		)
+	}
```
Groxx, line 62: do you know if TLS is (probably silently) ignored/downgraded if it's not enabled?  kinda seems like it shouldn't be anyway...

more generally though, we probably can't enable and enforce in the same step.
validation's useful, and e.g. if this is used inter-Cadence-service, it'll mean we can't deploy because non-TLS won't be able to communicate with the new TLS instances, effectively making a split brain.

mantas-sidlauskas: yep, good points. Will update this PR to include option for setting TLS mode. https://pkg.go.dev/go.uber.org/yarpc@v1.70.3/api/transport/tls#Mode

mantas-sidlauskas:
Closing this. Will bring up another PR for HTTP TLS only. TChannel/Ringpop/YARPC combination makes TLS for TChannel quite complicated.

PR #5354, opened 2023-07-19
Made the TTLBufferDays configurable

agautam478:
**What changed?**
Added a new param called `TTLBufferDays` in Dynamic config to make the TTL Buffer Day value configurable.

**Why?**
We have a very high number of 15 days hard coded as a buffer value for the TTL calculation for workflow executions. Once the feature is rolled out we may not need such a high buffer to ensure safety.

**How did you test it?**
Tested locally.

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**


timl3136 (approved):

service/history/execution/mutable_state_builder.go
```diff
@@ -4823,7 +4823,7 @@ func (e *mutableStateBuilder) calculateTTL() (int, error) {
 	}
 	config := domainObj.GetConfig()
 	retention := time.Duration(config.Retention)
-	daysInSeconds := int((retention + ttlBufferDays) * dayToSecondMultiplier)
+	daysInSeconds := int(retention) + e.config.TTLBufferDays()*dayToSecondMultiplier
```
Shaddoll, line 4826: this is incorrect

PR #5355, opened 2023-07-19
Fix garbage collection logic for matching tasks

Shaddoll:
**What changed?**
Fix garbage collection logic for matching tasks

**Why?**
We should delete all acked tasks

**How did you test it?**
eyeball check

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):
this wasn't caught by a test?

PR #5359, opened 2023-07-25
StartWorkflowExecution: validate RequestID before calling history

mantas-sidlauskas:
**What changed?**
Validating RequestID (uuid) at handler level.

**Why?**
When calling StartWorkflowExecution and passing wrongly formatted UUID, a generic error will be returned from persistence layer (cassandra for example). This error is not treated as non-retryable, so Cadence will try to insert wrong data multiple times. On the client side, only request-timeout will be returned which reveals no details about the nature for this failure.

This change will validate UUID on handler side and no calls to history/persistence will be made. Additionally, user will get information on what data is missing or malformed.

**How did you test it?**
Unit test updated to include malformed UUID check

**Potential risks**

**Release notes**

**Documentation Changes**


Shaddoll (approved):

service/frontend/workflowHandler.go
```diff
@@ -28,7 +28,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/pborman/uuid"
+	"github.com/google/uuid"
```
davidporter-id-au, line 31: we prorobably need to do this everywhere

davidporter-id-au (approved):

PR #5360, opened 2023-07-25
Adding TTL on other Execution types and bug fixes.

agautam478:
**What changed?**
- Fixed a minor calculation mistake while calculating TTL value.
- Included the cassandra upper bound consideration into the calculation. The maximum expiration timestamp that can be represented by the storage engine is
2038-01-19T03:14:06+00:00, which means that INSERTS using TTL that would expire
after this date are not currently supported.
- Updated the db updation query calls on execution close for all types of excutions. context: In certain execution flows some columns get a new value updated which doesn;t get TTL because of a separate update query running in the bg.:

`err = db.resetActivityInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.ActivityInfos)
	if err != nil {
		return err
	}
	err = db.resetTimerInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.TimerInfos)
	if err != nil {
		return err
	}
	err = db.resetChildExecutionInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.ChildWorkflowInfos)
	if err != nil {
		return err
	}
	err = db.resetRequestCancelInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.RequestCancelInfos)
	if err != nil {
		return err
	}
	err = db.resetSignalInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.SignalInfos)
	if err != nil {
		return err
	}
	return db.resetSignalsRequested(batch, shardID, domainID, workflowID, execution.RunID, execution.SignalRequestedIDs)`

**Why?**
- The calculation wasn't taking retention days into consideration.
- The upper bound needs to be added to provide coverage for workflows with unlimited timeouts which were getting an extremely high TTL value. Currently we are passing a value of 0 for such cases but may or may not decide to keep the upper bound value as TTL.
- To ensure that the complete row gets deleted when a TTL expires to avoid corruptions.

**How did you test it?**
Tested locally, will check in the staging env.

**Potential risks**
Stray or half deleted execution entires. Premature deletions.

**Release notes**
NA

**Documentation Changes**


common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
```diff
@@ -1171,27 +1200,27 @@ func (db *cdb) updateWorkflowExecutionAndEventBufferWithMergeAndDeleteMaps(
 		return fmt.Errorf("should only support WorkflowExecutionMapsWriteModeUpdate")
 	}
```
Shaddoll, line 1202: buffered event ttl is missing

service/history/execution/mutable_state_builder.go
```diff
@@ -54,6 +54,7 @@ const (
 	mutableStateInvalidHistoryActionMsgTemplate = mutableStateInvalidHistoryActionMsg + ": %v"
 
 	timerCancellationMsgTimerIDUnknown = "TIMER_ID_UNKNOWN"
+	cassandraUpperBound                = "2038-01-19T03:14:06+00:00"
```
Shaddoll, line 57: the related logic should be moved into cassandra plugin

common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
agautam478: ahh! do we need ttl in appendBufferedEvents too?

common/persistence/nosql/nosqlplugin/cassandra/workflowUtils.go
```diff
@@ -1032,27 +1056,27 @@ func (db *cdb) createWorkflowExecutionWithMergeMaps(
 		return fmt.Errorf("should only support WorkflowExecutionMapsWriteModeCreate")
 	}
 
-	err = db.updateActivityInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.ActivityInfos, nil)
+	err = db.updateActivityInfos(batch, shardID, domainID, workflowID, execution.RunID, execution.ActivityInfos, nil, 0)
```
Shaddoll, line 1059: If activities are created in workflow creation without ttl and never updated, these activities won't have ttls attached.

agautam478: but this is the create workflow execution flow. the execution in this flow has not ttl to begin with. :'( Also in case of `delay start` etc we can not have ttls attached to workflows that have not been started which translates to activities as well.

Should I just add an exception note here?

PR #5361, opened 2023-07-25
Add a "stale workflow" scanner/fixer impl, to remove data beyond reasonable retention

Groxx:
For a variety of reasons, we have workflows sitting in some of our databases far beyond any reasonable time range.
E.g. a workflow completed months ago in a domain with 7 day retention -> unfortunately still retained.

We've been gradually finding and fixing problems that cause this, but the old data needs to be cleaned up
at some point.  They can sometimes cause problems, particularly if they appear to be "running".

So this is a *conservative* scanner + fixer for workflows that meet these criteria.
It does not handle all states, nor does it support domains configured with >200 day retention (hardcoded but
trivially changeable, as a sanity check), but it's catching a large amount internally and seems safe to share
and run in more locations.

---

Since building and verifying this required figuring out the scanner/fixer code in detail,
and that proved to be a significant effort, this also includes three major additions on top
of the stale scanner:
- there have been a variety of renames and public -> private changes to make naming consistent
  - this helped discover a significant issue with the current execution scanner, so I'm keeping it
- concrete execution fixer invariants can now be individually enabled and disabled via dynamic config
- a new readme.md file that attempts to summarize the learnings, and how to config and verify scanner/fixer.


tools/cli/flags.go
```diff
@@ -24,6 +24,7 @@ import "github.com/urfave/cli"
 
 // Flags used to specify cli command line arguments
 const (
+	FlagVerbose                           = "verbose"
```
Groxx, line 27: I am honestly shocked that this is the first verbose flag

service/worker/scanner/executions/types.go
```diff
@@ -102,6 +104,10 @@ func (st ScanType) ToInvariants(collections []invariant.Collection) []InvariantF
 				fns = append(fns, invariant.NewInactiveDomainExists)
 			case invariant.CollectionHistory:
 				fns = append(fns, invariant.NewHistoryExists)
+			case invariant.CollectionStale:
```
davidporter-id-au, line 107: Digging around and looking at what's deployed with flipr keys like `worker.concreteExecutionFixerEnabled`, I think this will still execute as part of the `ConcreteExecutionType` category, making it still a bit dangerous. I don't there's currently a more granular flipr configuration. 

If you want to roll this out separately, I think you need to find a way to introduce a new category similar to `worker.concreteExecutionFixerEnabled`, unless I'm missing something (could easily be wrong, this code always confuses me)

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,624 @@
...
+	ts := time.Unix(0, *last.Timestamp)
+	if ok, result := c.checkTimeInSaneRange(ts, "last event timestamp"); !ok {
+		return when, false, result
+	}
+	return ts, true, result
+}
+
+func anyPresent(items ...interface{}) bool {
+	for _, i := range items {
+		// cannot use `i == nil` because typed nil pointers are boxed into non-nil interface{} due to having a type
```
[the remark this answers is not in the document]
Groxx: yea.  it really is.

service/worker/scanner/executions/types.go
Groxx: yea, I'll explore that and update somehow.  I'm not sure either.

Groxx: Got a concrete answer for ya!  https://github.com/uber/cadence/pull/5361/files#diff-5bdda90b33fbf138bc464971e79ca94b938e3ef2461b9f4dcceddc3088fff7f5R73-R117

service/worker/scanner/README.md
```diff
@@ -0,0 +1,247 @@
...
+Sadly, links to specific lines and repo-relative `/path/to/file` support is widely varying, so it should be avoided.
+
+# Overall behavior
+
+At the highest level, keep in mind:
+- Invariants define things to check and fix, and are defined 1:1 with their Collections but do not share names directly
+  - These are what actually check if X is bad, and do the things necessary to fix it
+  - Each handles only one source/type for their primary data, e.g. concrete vs current vs timer
+  - Invariant Managers are just N invariants that are run in a list.  Note that they are _not_ related to invariant collections.
+- Scanner and Fixer can be enabled for per data type (separate workflows) and per Collection (via config only)
```
davidporter-id-au, line 20: not sure sure i understand your point with this comment? Perhaps you could use an example? I vaguely recall there being a scanner workflow for concrete and current records

service/worker/scanner/executions/types.go
davidporter-id-au: oh god, I am still not sure I undrestand

service/worker/scanner/README.md
```diff
@@ -0,0 +1,247 @@
+# Scanner / fixer / etc how-to
```
davidporter-id-au, line 1: this is ... kind of is a how-to, but then descends into a manifesto. I agree 100% with all your points, but I'd maybe separate out the descriptive part ("tihs is how it works") from the normative part ("this is how it should be") and dump the normative part into a design-proposal. 

Having the two together, enmeshed would be an utterly baffling experience for the casual reader.

service/worker/scanner/README.md
```diff
@@ -0,0 +1,247 @@
...
+- [/common/reconciliation/invariant/types.go](../../../common/reconciliation/invariant/types.go)
+  - `Name` entries name and refer to individual `Invariant` implementations.
+  - `Collection` entries define... categories of sorts, which are used elsewhere to configure which `Invariant`s run.
+- [/service/worker/scanner/executions/types.go](executions/types.go)
+  - `Collection` selections elsewhere are converted to `Invariant`s, based on the scanner type
+    - Timer scanners are handled in [scanner/timers/timers.go](timers/timers.go) rather than in here
+- [/service/worker/scanner/executions/concrete_execution.go](executions/concrete_execution.go)
+  - this and [current_execution.go](executions/current_execution.go) define a `[ExecutionType]ExecutionConfig` func that can override which `Collection`s are run in a scanner.
+
+# Wishlist
```
davidporter-id-au, line 223: I would just cut this into a design proposal, it's inclusion here as-is makes it look to the casual reader as if we have no control over our own destiny (which may or may not be true) but doesn't make it clear that you're making a time-tradeoff with the PR.

service/worker/scanner/README.md
```diff
@@ -0,0 +1,247 @@
...
+      has a `map[string]string` pulling double duty as a set (it is at least uniquely typed!), doesn't complain about passing incorrect keys (a wrong-type invariant is ignored),
+      uses multiple different and disconnected kinds of strings and keys and values to control if an invariant is included or not, and is spread across several files.
+- "Don't run the workers if they are not configured" seems... maybe not ideal?
+  - Some benefits, some downsides, but definitely more complicated.
+  - Maybe just run these workers all the time?  Workflows/etc can be canceled if you want to stop them, the current setup only _pauses_ them (with essentially no timeout).
+- "Scan everything then fix everything" has pretty clear downsides, not least of which is potentially-massive latency.
+  - Being _able_ to separate them is useful for verification, that should be kept.
+- Invariants (some anyway) are used elsewhere, which is part of why they're highly generic.
+  - I have not investigated this use of invariants.
+  - Might only be related to NDC replication?
```
davidporter-id-au, line 242: the idea of invariants more generally yeah, I think could be super useful in all sorts of areas, including history building too. But the idea of an invariant as it *might* Be expressed (a pure fucntion which takes a series of states and returns true or false) and its implementation here in the scanner where it's coupled with DB iterators is probably a detail for an entire deep-dive.

service/worker/scanner/README.md
```diff
@@ -0,0 +1,247 @@
...
+<details><summary>Enabling scanners for local dev</summary>
+
+For local purposes, you can pretty easily enable these via the dynamicconfig
+file store [config/dynamicconfig/development.yaml](../../../config/dynamicconfig/development.yaml):
+```yaml
+# inconsistent key names, unfortunately.  don't guess, copy/paste
+worker.concreteExecutionFixerEnabled:
+- value: true
+  constraints: {}
+worker.executionsScannerInvariantCollectionStale:
```
davidporter-id-au, line 190: Is the 'fixing' behaviour enabled by default?

davidporter-id-au (approved):

common/dynamicconfig/constants.go
```diff
@@ -1709,6 +1709,12 @@ const (
 	// Default value: true
 	// Allowed filters: N/A
 	ConcreteExecutionsScannerInvariantCollectionHistory
+	// ConcreteExecutionsScannerInvariantCollectionStale indicates if the stale workflow invariant should be run
+	// KeyName: worker.executionsScannerInvariantCollectionStale
+	// Value type: Bool
+	// Default value: false
+	// Allowed filters: N/A
+	ConcreteExecutionsScannerInvariantCollectionStale
```
davidporter-id-au, line 1717: Question, since I'm not fully understanding: Does this cover the fixer operation too?

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+}
+
+func (c *staleWorkflowCheck) Name() Name {
+	return StaleWorkflow
+}
+
+// sane time ranges for workflows.
+// this is largely to protect against accidental conversions from zero values, or incorrect second/nanosecond/etc scales.
+var (
```
davidporter-id-au, line 190: nit: I'd just move this up to the top

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,628 @@
...
+
+	if info.State == persistence.WorkflowStateRunning {
+		stale, expected, result := c.checkRunningAge(workflow, maxLifespan, domainName)
+		if stale {
+			// we know these exist, but we don't expect them to be numerous "recently".
+			// log for verification.
+			logStaleWorkflow("running", expected)
+		}
+		return stale, result
+	} else if info.State == persistence.WorkflowStateZombie {
```
davidporter-id-au, line 255: nit, no else required I think?

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,628 @@
...
+		if attr.Attempt > 0 && !executionInfo.ExpirationTime.IsZero() && workflowTimeoutTimestamp.After(executionInfo.ExpirationTime) {
+			workflowTimeoutTimestamp = executionInfo.ExpirationTime
+		}
+	*/
+	first, ok, result := c.firstEvent(workflow, domainName)
+	if !ok {
+		// completely missing history should be caught by the history-exists check, no need to handle here
+		return false, expected, result
+	}
+	startedAt := workflow.State.ExecutionInfo.StartTimestamp                             // nonzero
```
davidporter-id-au, line 297: I didn't catch this earlier, but perhaps add your impossibly-early start guard here in-case there was some accidental nil start values?

davidporter-id-au: if startedAt.Before(impossiblyOld) {
    return false, ....
}

davidporter-id-au: oh, you're doing it below, nevermind

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,628 @@
...
+		batchTimeout  = 5 * time.Second
+	)
+	iter := 0
+	var nextPageToken []byte
+	var lastEvent *types.HistoryEvent
+	for ; iter < (maxHistoryLen / pageSize); iter++ { // loose sanity check
+		ctx, cancel := context.WithTimeout(context.Background(), batchTimeout)
+		defer cancel() // revive:disable-line:defer clean up on panics, dups are just noops
+
+		shard := c.pr.GetShardID()
```
davidporter-id-au, line 599: do you need to get the shard every iteration?

common/reconciliation/invariant/staleWorkflow.go
```diff
@@ -0,0 +1,628 @@
...
+			BranchToken: branchToken,
+			// only interested in the last event, but we have to read in order to get there.
+			MinEventID:    common.FirstEventID,
+			MaxEventID:    common.EndEventID,
+			ShardID:       &shard,
+			PageSize:      pageSize,
+			DomainName:    domainName,
+			NextPageToken: nextPageToken,
+		})
+		cancel() // handle non-panics
```
davidporter-id-au, line 610: you're cancelling twice?

Groxx: the deferred one would run at function exit, not on the next loop.

there are other ways to achieve this (mutating a var, deferring a late-bound call at the beginning), it just seemed simpler to call it twice.

service/worker/scanner/README.md
Groxx: there are three scanner workflows and three fixer workflows.
all three of each are nearly the same, but they operate on different types and have different workflow names/types.  e.g.:
```go
ConcreteExecutionsScannerWFTypeName   = "cadence-sys-executions-scanner-workflow
```

within each of those, the invariants that are run are individually enable/disable-able.

but yea, gonna reword pretty much all of this 👍

service/worker/scanner/README.md
Groxx: after a bit of digging: it appears both "yes" and "oh no".  I'm debating changing it...

the long and the short of it is that fixer does not have any per-invariant config.  what it does is:
- reads executions that scanner identified as corrupted
- ignores what it decided / what scan identified the problem
- runs all invariants on that execution (after checking that the domain is enabled)

So there is both no way to not-fix, and by introducing a new fix you start *immediately* using it on all scan results.

bleh.

I think I can get a scanner-config-like addition in fixer without much trouble, so we can toggle invariants for both scanner and fixer independently.  doing anything fancier might take a lot more work :\

service/worker/scanner/executions/types.go
Groxx: np, I'm gonna reword to put config as a separate and clearer thing.

basically:
- config controls what invariants get into ^ the list of invariants passed in here, because the enabled config names are converted to an enum lookup which produces this list
- these are filtered by type (which should always match what are passed in, but this is not checked)
- so if you don't have a scanner enabled, its config-key doesn't exist, so the enum isn't found, so it isn't in this list, so the invariant doesn't run in scanner.

which is, like, bleh, but it does work.

Groxx: for "fixer enabled" config: yeah, I think you may be right.  since right now it runs all invariants in fixer :\  (see other comment)

service/worker/scanner/shardscanner/activities.go
```diff
@@ -296,8 +299,38 @@ func FixerCorruptedKeysActivity(
 	}, nil
 }
 
-// FixShardActivity will fix a collection of shards.
-func FixShardActivity(
+type (
```
davidporter-id-au, line 302: nit: type declarations maybe ought to be at the top?

service/worker/scanner/shardscanner/fixer_workflow.go
```diff
@@ -121,10 +124,28 @@ func NewFixerWorkflow(
 	return &wf, nil
 }
 
+func supportsFixerConfig(ctx workflow.Context) bool {
```
davidporter-id-au, line 127: Happy to :stamp: but I would suggest that a terminate and restart is probably ok too. It's not clear to me that we should be too worried about backwards compat with this

service/worker/scanner/README.md
```diff
@@ -0,0 +1,598 @@
+# A brief* overview of Scanner and Fixer and how to control them
+
+There are a variety of problems with the Scanner and Fixer related code, so
```
davidporter-id-au, line 3: Maybe put a todo for the assumed audience knowledge and the why: `what is Scanner and Fixer?` and `what problmes are they intended to solve`. The rest of this doc is concerned the how.

davidporter-id-au: I wouldn't block on this though. writing the purpose and intent of scanner and fixer are out of the scope of this PR

service/worker/scanner/README.md
```diff
@@ -0,0 +1,598 @@
...
+        "InfoDetails": ""
+      }
+    ]
+  }
+}
+```
+You can see scanner's data as the input-result, and _completely different invariants_
+running and failing as the fixer result.
+
+## Thoughts and prayers
```
davidporter-id-au, line 553: Patient gains 0 life

service/worker/scanner/README.md
davidporter-id-au: I would call out maybe in words of fire though that these features are all essentially in pre-release/alpha

service/worker/scanner/README.md
davidporter-id-au: (dumb joke from Xwitter)

davidporter-id-au (commented):
LGTM. 

Overall the testing you did certainly makes me a lot more confident in the overall approach, though looked extremely annoying. 

If you'd done the work already I'd suggest not doing the backwards-compat, but no strong feelings that it's there, though I would be awary introducing further complexity or working on the assumption or just set to terminate-if-running

davidporter-id-au (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 You can find a list of previous releases on the [github releases](https://github.com/uber/cadence/releases) page.
 
+## [Unreleased]
```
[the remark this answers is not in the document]
Groxx: It's the main one I've seen - I like it. It lets you build the changelog literally as you build it, so there's no need to scramble to gather the list at the end.  And you can review the additions.

service/worker/scanner/executions/concrete_execution.go
```diff
@@ -125,26 +125,41 @@ func ScannerIterator(
...
 
-// FixerManager provides invariant manager for concrete execution fixer.
-func FixerManager(_ context.Context, pr persistence.Retryer, _ shardscanner.FixShardActivityParams, domainCache cache.DomainCache) invariant.Manager {
+// concreteExecutionFixerManager provides invariant manager for concrete execution fixer.
+func concreteExecutionFixerManager(_ context.Context, pr persistence.Retryer, params shardscanner.FixShardActivityParams, domainCache cache.DomainCache) invariant.Manager {
 	var ivs []invariant.Invariant
 	var collections []invariant.Collection
 
-	collections = append(collections, invariant.CollectionHistory, invariant.CollectionMutableState)
+	if len(params.EnabledInvariants) == 0 {
```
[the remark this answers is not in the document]
Groxx: Not overly.

The main consequence is that you could switch behaviors part way through the run.  You'll still have the output showing the change though.

I don't feel too strongly on it, I just default to consistency where able and this wasn't too hard.  I'd be game to drop it if you think the cleanup is worth more (I might even agree).

service/worker/scanner/executions/current_execution.go
```diff
@@ -55,6 +56,20 @@ const (
 	CurrentExecutionsFixerTaskListName = "cadence-sys-current-executions-fixer-tasklist-0"
 )
 
+/*
+
+!!!!!!!!!!!!!!
+NOTE: Current execution fixers have never been run.
```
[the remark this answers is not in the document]
Groxx: I'm pretty tempted to, yeah. This PR was just getting too large already.

service/worker/scanner/README.md
Groxx: like scanner/fixer as a whole?

service/worker/scanner/executions/concrete_execution.go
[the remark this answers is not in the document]
Groxx: I'll drop this I think, and keep the workflow versioning.
- the workflow code is simpler and it'll prevent a non-determinism alert
- this code's behavior will just mean fixer doesn't fix anything in the current run after a deploy.  silently noops and the next pass should catch the same things anyway.

service/worker/scanner/shardscanner/fixer_workflow.go
Groxx: I think it may be worth it on this one.  These workflows are essentially eternal (years of timeout) so if they go unnoticed they might get into a non-stop loop of failure :\

I believe they're disabled by default, but I'd rather not make it possible to miss that upgrade step.

PR #5377, opened 2023-08-15
Fix context leak in tests

munahaf:
**What changed?**
A marker has been added to the code. This is not an actual fix, but a marker denoting that there may be a bug.

**Why?**
In file: taskpoller.go, there is a function ```createContext``` that contains a cancellation function returned by ```context.WithTimeout```. However, the cancellation function is ignored using ```_```. I suggested that the cancellation function must be called or the new context will remain live until the parent context is cancelled. I did not propose a fix since I am not familiar with the business logic. However, the codebase appears to be following that prescribed convention in other parts.

Sponsorship and Support:
This work is done by the security researchers from OpenRefactory and is supported by the Open Source Security Foundation (OpenSSF)(https://openssf.org/): Project Alpha-Omega(https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

**How did you test it?**
Not reproduced.

**Potential risks**
Cannot ascertain.

**Release notes**
Appears not a release candidate.

**Documentation Changes**
Appears no change is needed.


host/taskpoller.go
```diff
@@ -454,7 +454,8 @@ func (p *TaskPoller) PollAndProcessActivityTaskWithID(dropTask bool) error {
 }
 
 func createContext() context.Context {
-	ctx, _ := context.WithTimeout(context.Background(), 90*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)
+        defer cancel()
```
Shaddoll, line 458: the context will be cancelled immediately after this function returns. We should also return cancel and let caller to call cancel.
BTW, this code is only used in our integration test. It's not good but it's ok.

adonovan: If you run "go vet" over your code, it should point out mistakes of this kind. For example:
https://go.dev/play/p/1zlF7qFbzhY

taylanisikdemir: hey @adonovan, the proper fix would be to return cancel callback and change all the callers with `defer cancel()`. Do you want to proceed with that? Otherwise let's close this PR and we will find someone else to address it

openrefactorymunawar:
I have prepared an updated patch. However, I am traveling now and cannot access the GitHub account used to create the pull request. I will address this in about a week.

host/taskpoller.go
openrefactorymunawar: I have prepared an updated patch. However, I am traveling now and cannot access the GitHub account used to create the pull request. I will address this in about a week.

Groxx:
While I agree with this in general, I don't think it's true for `context.Background()`.

The leak would be coming from the parent's list of children to propagate cancels, but a background context both does not have that list, and it has a nil `done` chan so this branch is taken when making a cancel-context derived from it: https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/context/context.go;l=461

So these should be cleaned up as soon as they're garbage-collectable.  Which is fine.

That said, these kinds of funcs mean we can't easily add context data (no parent-ctx arg), and this is a potential landmine if we did (because at that point it'd likely leak).  So as long as we've got a clear path to cancel correctly, it seems like an improvement.

taylanisikdemir (approved):
thanks for addressing this!

PR #5389, opened 2023-08-29
Bump Sarama to v1.41.0 and handle consumer state

dnwe:
**What changed?**

Bump Sarama to v1.41.0 and handle consumer state. When either the messages channel is closed or the session context is done, the ConsumeClaim func should break out of the for loop and return.

Note: the module path has changed to github.com/IBM/sarama since ownership transitioned away from Shopify

**Why?**

Pulling in latest updates and new module path. Also correcting consumer group implementation to handle rebalance states.

**How did you test it?**

`go test --failfast ./...`

**Potential risks**

If user has not previously configured a kafkaVersion setting that is equal to or older than their target server then they might see incompatible protocol attempted, as Sarama now supports a more complete set of versions

**Release notes**

N/A

**Documentation Changes**

N/A


taylanisikdemir:
Closing this one as part of old PR cleanup effort. @neil-xie will evaluate the risks of this upgrade and apply in a separate PR with proper tests if we see fit.

PR #5390, opened 2023-08-29
Enable json index

bowenxia:
**What changed?**
1. removed previously flattened custom search attributes. Need to change code of both Pinot request and response. 
2. update pinot schema: delete the custom search attributes in schema & add an Attr column
3. update pinot config: enable json index for Attr column
4. re-design & re-implement PinotQueryValidator
5. update unit tests accordingly

**Why?**
We flatten the custom search attributes previously for Pinot, and this caused a problem that the response will return every attribute even though those were null, and the impact is, the response is 7x bigger in dev2. We believe this will be bigger in production envs since there are more custom search attributes. And it is not solvable on Pinot side because of its response format. 
Using json index: 
1. Pinot can treat the custom search attribute field as a json format, which means, all the custom attributes will be combined in 1 field within Pinot, and when there're null values, it won't store in Pinot, thus also won't return in the response.
2. If we use flattened custom search attributes, every time we need to onboard new custom attributes, we need to update the kafka schema, which will take 24h of waiting to take effect on Pinot side, which is unacceptable.  Using json index, everything will be stored in custom search attribute field as a Json format.

Why we need to redesign and re-implement PinotQueryValidator:
At the very first beginning, when we translate the exact match & partial match query from ES format to Pinot format, our code can't deal with a string with and/or, so we reused the validator code from frontend that can split the query using an object called *sql expression*.
The validator code has a string input, and a sql expression output. Given that right now, we are exploring Json index, we need to translate exact match & partial match into Json index format. But the problem is, sql expression doesn't have the related component. So we need to redesign and reimplement the validator to only use the parser to split the left & right side of a comparison expression, and then output a string instead of a sql expression.

**How did you test it?**
Unit test. Will run integration test after this commit. 

**Potential risks**

**Release notes**

**Documentation Changes**


common/pinot/pinotClient.go
```diff
@@ -153,7 +151,7 @@ func (c *PinotClient) getInternalGetClosedWorkflowExecutionResponse(resp *pinot.
 	schema := resp.ResultTable.DataSchema // get the schema to map results
 	columnNames := schema.ColumnNames
 	actualHits := resp.ResultTable.Rows
-	response.Execution = ConvertSearchResultToVisibilityRecord(actualHits[0], columnNames, c.logger, isSystemKey)
+	response.Execution = ConvertSearchResultToVisibilityRecord(actualHits[0], columnNames, c.logger)
```
neil-xie, line 154: The isSystemKey() method below can also be removed

neil-xie: Do we no longer need any kind of system key check or it is required when build the query?

bowenxia: Yes we don't need it anymore since it is already be checked in frontend. We can be 100% sure that those are valid.

common/types/mapper/thrift/shared.go
bowenxia, line 1: I commented these 2 things because of some dependency/version issue

docker/dev/cassandra-pinot-kafka.yml
bowenxia, line 1: Pinot has a way better performance, and more stable when I updated the version.

neil-xie (approved):

PR #5392, opened 2023-08-31
Expose workflow history size and count to client

timl3136:
**What changed?**
Added fields regarding workflow history size and count in PollForDecisionResponse to be exposed to cadence-client.
History size will be exposed in unit of bytes and history count will be the number of completed events.
Added fields sending history data from cadence-history to cadence-matching when matching service calls the persistence layer.
Updated dependency for cadence-idl to include new fields in idl in sync with cadence-client.

**Why?**
There have been requests from customers wanting history information regarding workflow to be exposed in client so that they can optimize their workflow through methods like continueAsNew when approaching the history size limit.
Resolved cadence-client issue [#917 ](https://github.com/uber-go/cadence-client/issues/917)

**How did you test it?**
Standard unit test as well as tested in development environment.

**Potential risks**
Since we have added new fields to thrift and proto layer, there should not be any breaking change since new fields will be defaulted to 0. However, there remains the possibility that the change will result in thrift/proto field mismatch between serve and client hence result in information loss.

**Release notes**

**Documentation Changes**


client/history/client.go
```diff
@@ -127,6 +128,7 @@ func (c *clientImpl) GetMutableState(
 	if err != nil {
 		return nil, err
 	}
+	c.logger.Info(fmt.Sprintf("History ClientImpl History size %d", response.HistorySize))
```
Shaddoll, line 131: can you remove this log?

common/types/matching.go
```diff
@@ -455,6 +456,14 @@ func (v *MatchingPollForDecisionTaskResponse) GetBranchToken() (o []byte) {
 	return
 }
 
+// GetHistorySize is an internal getter of returning the history size in bytes
+func (v *MatchingPollForDecisionTaskResponse) GetHistorySize() (o int64) {
```
Shaddoll, line 460: The mismatch between field name and getter name is confusing.

service/matching/matchingEngine.go
```diff
@@ -488,12 +495,14 @@ pollLoop:
 				StickyExecutionEnabled:    isStickyEnabled,
 				WorkflowExecutionTaskList: mutableStateResp.TaskList,
 				BranchToken:               mutableStateResp.CurrentBranchToken,
+				HistorySize:               mutableStateResp.HistorySize,
```
Shaddoll, line 498: are you exposing history size only for query tasks? what if users don't do queries?

common/types/history.go
Shaddoll, line 1: Could you update the unit tests of this package to make sure the new field is included by mapper?

common/util.go
```diff
@@ -413,6 +413,7 @@ func CreateMatchingPollForDecisionTaskResponse(historyResponse *types.RecordDeci
 		ScheduledTimestamp:        historyResponse.ScheduledTimestamp,
 		StartedTimestamp:          historyResponse.StartedTimestamp,
 		Queries:                   historyResponse.Queries,
+		TotalHistoryBytes:         historyResponse.HistorySize,
```
Shaddoll, line 416: Do we expose historyCount?

service/matching/matchingEngine.go
timl3136: For queue task, history size is exposed through the existing GetMutableState function calling (here). For non-query task, history size is exposed through the RecordDecisionTaskStarted function.

client/history/client.go
timl3136: Resolved

common/util.go
timl3136: Yes, the historyCount is equivalent to NextEventID - 1

common/types/matching.go
timl3136: Resolved

common/types/history.go
timl3136: Resolved

Shaddoll (approved):

PR #5393, opened 2023-09-01
Update default values of a few dynamic configs

Shaddoll:
**What changed?**
Update default values of a few dynamic configs

**Why?**
To adopt the default settings we're using at Uber

**How did you test it?**

**Potential risks**

**Release notes**
If you're using the old default values, you need to manually create those dynamic configs with the old default values so that you're not affected by the change. And then you can update the dynamic configs to the new default value gradually.

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -2694,8 +2694,9 @@ var IntKeys = map[IntKey]DynamicInt{
 	},
 	BlobSizeLimitError: DynamicInt{
 		KeyName:      "limit.blobSize.error",
+		Filters:      []Filter{DomainName},
 		Description:  "BlobSizeLimitError is the per event blob size limit",
-		DefaultValue: 2 * 1024 * 1024,
+		DefaultValue: 512 * 1024,
 	},
 	BlobSizeLimitWarn: DynamicInt{
 		KeyName:      "limit.blobSize.warn",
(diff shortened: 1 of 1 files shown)


common/dynamicconfig/constants.go
```diff
@@ -2707,25 +2708,25 @@ var IntKeys = map[IntKey]DynamicInt{
...
 		Description:  "HistorySizeLimitWarn is the per workflow execution history size limit for warning",
-		DefaultValue: 50 * 1024 * 1024,
+		DefaultValue: 25 * 1024 * 1024,
 	},
 	HistoryCountLimitError: DynamicInt{
 		KeyName:      "limit.historyCount.error",
 		Filters:      []Filter{DomainName},
 		Description:  "HistoryCountLimitError is the per workflow execution history event count limit",
-		DefaultValue: 200 * 1024,
+		DefaultValue: 50000,
```
davidporter-id-au, line 2723: I would worry that this would be a bit disruptive to get more restrictivre config during an upgrade - its' a bit of a breaking change which would likely affect customers. though I agree it's probably too high to be safe for a lot of high throughput prod envs. 

For our own internal use I have some ideas (will discuss offline) about handling our config defaults internally. I am not sure open-source customers would welcome having their workflows hit errors on upgrade however.

taylanisikdemir:
Closing as part of cleanup effort.

PR #5399, opened 2023-09-13
Improve poller detection for isolation

Shaddoll:
**What changed?**
Improve poller isolation group detection

**Why?**
If pollers from one isolation group is completely down, we want to redistribute the tasks from that isolation to other pollers ASAP. Previously, it takes 5 mins to detect that. With this change we can reduce it to 10s.

**How did you test it?**
unit test

**Potential risks**
If not implemented correctly, there could be more leaks (tasks not picked by pollers from the same zone)

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager.go
+++ b/service/matching/taskListManager.go
@@ -27,6 +27,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -92,6 +93,11 @@ type (
 		TaskListID() *taskListID
 	}
 
+	outstandingPollerInfo struct {
+		isolationGroup string
+		cancel         context.CancelFunc
+	}
+
 	// Single task list in memory state
 	taskListManagerImpl struct {
 		createTime      time.Time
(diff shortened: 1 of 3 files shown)


service/matching/taskListManager.go
```diff
@@ -120,7 +126,7 @@ type (
 		// outstanding poller when the frontend detects client connection is closed to
 		// prevent tasks being dispatched to zombie pollers.
 		outstandingPollsLock sync.Mutex
-		outstandingPollsMap  map[string]context.CancelFunc
+		outstandingPollsMap  map[string]outstandingPollerInfo
```
davidporter-id-au, line 129: Mind adding a comment as to what the key is? I think it's pollerid?

service/matching/taskListManager.go
```diff
@@ -618,6 +624,21 @@ func (c *taskListManagerImpl) getIsolationGroupForTask(ctx context.Context, task
 	return defaultTaskBufferIsolationGroup, nil
 }
 
+func (c *taskListManagerImpl) getPollerIsolationGroups() []string {
+	groupSet := c.pollerHistory.getPollerIsolationGroups(time.Now().Add(-10 * time.Second))
```
davidporter-id-au, line 628: Mind adding a comment about what 10 seconds comes from?

davidporter-id-au: Ah, I hadn't read the code in func `(c *taskListManagerImpl) HasPollerAfter` closely enough, thank you for the correction. I had the wrong picture in my head regarding the risk of missing the poller. this historical lookback is used in addition to a map of outstanding polls.

davidporter-id-au (approved):

PR #5401, opened 2023-09-18
Remove database check for config store tests

Shaddoll:
**What changed?**
Remove database check for config store tests

**Why?**
Because we want to test config store with other databases

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


davidporter-id-au (approved):

Groxx:
Seems like this will still be testing against databases?  We do have DB-stored config managers: https://github.com/Shaddoll/cadence/blob/master/common/persistence/nosql/nosqlConfigStore.go#L43

Shaddoll:
Synced offline with @Groxx , everything is OK.

PR #5403, opened 2023-09-19
Implement config store for MySQL

Shaddoll:
**What changed?**
- Implement config store my MySQL
- Update config store test to use different row types for different tests so that we don't need to truncate the table in every test

**Why?**
- To support config store for MySQL

**How did you test it?**
persistence tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/sql/sqlplugin/mysql/configstore.go
+++ b/common/persistence/sql/sqlplugin/mysql/configstore.go
@@ -24,15 +24,37 @@ package mysql
 
 import (
 	"context"
-	"fmt"
 
+	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/common/persistence/sql/sqlplugin"
+)
+
+const (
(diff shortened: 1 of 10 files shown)


common/persistence/sql/sqlplugin/mysql/configstore.go
```diff
@@ -24,15 +24,37 @@ package mysql
...
 }
 
 func (mdb *db) SelectLatestConfig(ctx context.Context, rowType int) (*persistence.InternalConfigStoreEntry, error) {
-	return nil, fmt.Errorf("not implemented")
+	var row sqlplugin.ClusterConfigRow
+	err := mdb.driver.GetContext(ctx, sqlplugin.DbDefaultShard, &row, _selectLatestConfigQuery, rowType)
+	if err != nil {
+		return nil, err
+	}
+	row.Version *= -1
```
davidporter-id-au, line 50: Are versions being stored in descending order in the CAAS implementation?

davidporter-id-au (approved):

common/persistence/sql/sqlplugin/mysql/configstore.go
Shaddoll: yeah, it's descending order in CAAS

PR #5404, opened 2023-09-19
Pinot deletion

bowenxia:
**What changed?**
Added a deletion column in Pinot

**Why?**
Need to implement deletion

**How did you test it?**
unit test, integration test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/Pinot/pinotVisibilityStore.go
+++ b/common/persistence/Pinot/pinotVisibilityStore.go
@@ -550,7 +550,7 @@ func createDeleteVisibilityMessage(domainID string,
 }
 
 func createVisibilityMessage(
-// common parameters
+	// common parameters
 	domainID string,
 	wid,
 	rid string,
@@ -563,11 +563,11 @@ func createVisibilityMessage(
 	encoding common.EncodingType,
 	isCron bool,
 	numClusters int16,
-// specific to certain status
-	closeTimeUnixNano int64,                           // close execution
+	// specific to certain status
+	closeTimeUnixNano int64, // close execution
 	closeStatus workflow.WorkflowExecutionCloseStatus, // close execution
-	historyLength int64,                               // close execution
-	updateTimeUnixNano int64,                          // update execution,
+	historyLength int64, // close execution
+	updateTimeUnixNano int64, // update execution,
 	shardID int64,
 	rawSearchAttributes map[string][]byte,
 	isDeleted bool,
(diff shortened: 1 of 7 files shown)


common/persistence/Pinot/pinotVisibilityStore.go
```diff
@@ -471,28 +517,36 @@ func (v *pinotVisibilityStore) CountWorkflowExecutions(ctx context.Context, requ
...
-}
-
 func (v *pinotVisibilityStore) checkProducer() {
 	if v.producer == nil {
 		// must be bug, check history setup
 		panic("message producer is nil")
 	}
 }
 
+func createDeleteVisibilityMessage(domainID string,
```
neil-xie, line 527: I'm think if we can combine with createVisibilityMessage using if isDeleted to return different message. But for now, at least we need a comment here to say why we have new method here and just have these fields since it is too similar to the createVisibilityMessage method so other people can understand

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -397,23 +397,32 @@ func (v *pinotVisibilityTripleManager) CountWorkflowExecutions(
 
 func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForRead(domain string) (VisibilityManager, VisibilityManager) {
 	var visibilityMgr VisibilityManager
-	var visibilityMgr2 VisibilityManager
```
neil-xie, line 400: I think this is just for debugging purpose to read from Pinot, we can change this part after we redesign the comparator

config/dynamicconfig/development.yaml
```diff
@@ -54,4 +54,5 @@ frontend.validSearchAttributes:
     project: 1
     service: 1
     user: 1
+    IsDeleted: 1
```
neil-xie, line 57: this should be 4 boolean, though it actually doesn't matter here

config/dynamicconfig/development_es.yaml
```diff
@@ -38,6 +38,7 @@ frontend.validSearchAttributes:
       BinaryChecksums: 1
       Passed: 4
       ShardID: 2
+      IsDeleted: 1
```
neil-xie, line 41: same

config/dynamicconfig/development_pinot.yaml
```diff
@@ -40,6 +40,7 @@ frontend.validSearchAttributes:
       BinaryChecksums: 1
       Passed: 4
       ShardID: 2
+      IsDeleted: 1
```
neil-xie, line 43: same

neil-xie:
Looks good to me overall, just some minor fix. Did you change anything in idls? If not, we can discard that one. It can cause conflicts when rebase to master.

neil-xie (approved):

bowenxia:
I didn't change idls.

common/persistence/Pinot/pinotVisibilityStore.go
bowenxia: comment added

PR #5410, opened 2023-10-03
Trigger data inconsistency scan during replication task processing

agautam478:
- Uncommented the code that triggered scanner on a seemingly corrupted workflow during single replication task processing.
- Added an additional flipr check to control the scan per domainID. Chose DomainID because the Domain Name would have taken an additional call. Domain ID serves the purpose well enough.
- The goal here is to try and push this into production, enable this for a bunch of known problematic domains and then consider adding these for other flows.

These changes were made to ensure that any workflows that looks bad get handled by the Scanner instead of lounging in the DLQ to be looked at later. This is an ongoing experiment that has to be taken slow hence the changes have been made at a single place for now.
The future plan is to enable this for every other processing type if verified.

**How did you test it?**
Testing it in the staging env. There's a unit test for the Scanner function as well.

**Potential risks**
If deployed to production without verification it may end up triggering scanner on each task processing. I don't foresee any breaking changes associated with it though.

**Release notes**

**Documentation Changes**


service/history/replication/task_processor.go
```diff
@@ -430,10 +430,12 @@ func (p *taskProcessorImpl) processSingleTask(replicationTask *types.Replication
 			tag.Error(err),
 		)
 		//TODO: uncomment this when the execution fixer workflow is ready
-		//if err = p.triggerDataInconsistencyScan(replicationTask); err != nil {
-		//	p.logger.Warn("Failed to trigger data scan", tag.Error(err))
-		//	p.metricsClient.IncCounter(metrics.ReplicationDLQStatsScope, metrics.ReplicationDLQValidationFailed)
-		//}
+		// Uncommenting this to verify that the scanner indeed gets triggered.
+		if err = p.triggerDataInconsistencyScan(replicationTask); err != nil {
```
Shaddoll, line 434: I think we should add a flipr control over it, because it might add lots of extra load.

agautam478: Good idea! Will add a flipr for this. Thanks for the suggestion.

service/history/replication/task_processor.go
```diff
@@ -430,10 +430,12 @@ func (p *taskProcessorImpl) processSingleTask(replicationTask *types.Replication
...
+			p.logger.Info("Triggering a data scan")
```
Shaddoll, line 435: the info logs should be removed, because they don't help with debugging and also uses lots of quota

agautam478: Added these to test the stuff on staging. Will remove afterwards.

service/history/replication/task_processor.go
```diff
@@ -429,11 +431,16 @@ func (p *taskProcessorImpl) processSingleTask(replicationTask *types.Replication
...
 		)
-		//TODO: uncomment this when the execution fixer workflow is ready
-		//if err = p.triggerDataInconsistencyScan(replicationTask); err != nil {
-		//	p.logger.Warn("Failed to trigger data scan", tag.Error(err))
-		//	p.metricsClient.IncCounter(metrics.ReplicationDLQStatsScope, metrics.ReplicationDLQValidationFailed)
-		//}
+
+		// Uncommenting this to verify that the scanner indeed gets triggered.
+		//Adding extra flipr protection since this is an ongoing test.
+		if p.config.EnableDataScan(request.TaskInfo.GetDomainID()) {
```
3vilhamster, line 437: There is a shard cache with domain names in task processor: p.shard.GetDomainCache().GetDomainName(domainID)

agautam478: thanks for the pointer. Will update the flipr to domain Name.

agautam478:
Dropping the plan to utilise Scanner for detection and removal.

PR #5412, opened 2023-10-03
Fix workflow replication for reset workflow

Shaddoll:
**What changed?**
Fix workflow replication for reset workflow

**Why?**
Currently if we failover a domain with failover_version equal to 0 and reset a workflow, the replication of that reset workflow can potentially fail.

**How did you test it?**
new unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/ndc/replication_task.go
+++ b/service/history/ndc/replication_task.go
@@ -49,7 +49,7 @@ type (
 		getLogger() log.Logger
 		getVersionHistory() *persistence.VersionHistory
 		isWorkflowReset() bool
-		getWorkflowResetMetadata() (string, string, int64)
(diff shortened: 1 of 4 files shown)


service/history/ndc/replication_task.go
```diff
@@ -214,22 +214,26 @@ func (t *replicationTaskImpl) getVersionHistory() *persistence.VersionHistory {
 
 func (t *replicationTaskImpl) isWorkflowReset() bool {
```
shijiesheng, line 215: nit: can we add some unit test around it?

shijiesheng (approved):

PR #5413, opened 2023-10-03
Fix CreateWorkflowModeContinueAsNew for SQL

Shaddoll:
**What changed?**
Update CreateWorkflowModeContinueAsNew of SQL to check if the run ID from the request matches the current run ID.
And also makes the behavior the same as NoSQL implementation, see https://github.com/uber/cadence/blob/v1.2.4/common/persistence/nosql/nosqlExecutionStoreUtil.go#L715

**Why?**
This is to make sure the behavior is the same as NoSQL implementation. And without this check, concurrent workflow resets could result in multiple open workflows with the same workflow id, and one of them will get stuck forever.

**How did you test it?**
eyeball check, 

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/sql/sqlExecutionStore.go
+++ b/common/persistence/sql/sqlExecutionStore.go
@@ -185,9 +185,13 @@ func (m *sqlExecutionStore) createWorkflowExecutionTx(
 			}
 
 		case p.CreateWorkflowModeContinueAsNew:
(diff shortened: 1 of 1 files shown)


davidporter-id-au (approved):
took me quite a long time for me to undertand this change. Anyway, lgtm

PR #5414, opened 2023-10-07
Adding a metric for hosts not being found in resolver

davidporter-id-au:
**What changed?**
Adds a metric around hosts not being visible in the resolver. This is for debugging and early alerting, though it's normal for this metric to appear during rebalancing or deployments temporarily. Long periods of this metric firing may indicate a problem with service discovery.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


common/membership/resolver.go
```diff
@@ -193,6 +198,7 @@ func (rpo *MultiringResolver) Unsubscribe(service string, name string) error {
 func (rpo *MultiringResolver) Members(service string) ([]HostInfo, error) {
 	ring, err := rpo.getRing(service)
 	if err != nil {
+		rpo.metrics.Scope(metrics.ResolverHostNotFoundScope).IncCounter(1)
```
dkrotx, line 201: Isn't it better to move to LookupByAddress() which calls `Members`?
Then it would be clear - every time `LookupByAddress` returns empty `HostInfo` we emit the metric.

dkrotx (approved):

common/membership/resolver.go
davidporter-id-au: yes, you're right, I was being a bit hurried and this doesn't make sense

common/membership/resolver.go
```diff
@@ -26,6 +26,7 @@ package membership
 import (
 	"errors"
 	"fmt"
+	"github.com/uber/cadence/common/metrics"
```
3vilhamster, line 29: imports are messed up.

common/membership/resolver_test.go
```diff
@@ -23,6 +23,7 @@
 package membership
 
 import (
+	"github.com/uber/cadence/common/metrics"
```
3vilhamster, line 26: imports are messed up

3vilhamster (commented):
Something is off with you goimports setup. Otherwise LGTM

PR #5423, opened 2023-10-16
Hotfix for async matching for isolation-group redirection

davidporter-id-au:
**What changed?**

We observed some problems in production where a heavily backlogged backfill event got stuck and tasks were not dispatched. The issue was mitigated by rebooting matching, but more so by turning off zonal isolation for this one service. 

### Context 

When a task is attempting to be dispatched through async dispatch, but it times out and there doesn't appear to be any pollers in the new zone, the feature will attempt to redirect the traffic to an isolation group where there are pollers - aka 'task redispatch'. This is intended to be a feature to avoid traffic from getting blackholed, but also to prevent it from blocking progression for other isolation groups in the tasklist, since it's a FIFO queue.

### Suspected problem:

The suspicion is that the task processing is failing due to getting to a deadlock around task redispatch. 

This pull-request attempts to fix the suspected problem of isolation-group task redirection causing a deadlock between the task pump and task dispatch. The basic idea being there's a few scenarios where it's conceivable that redispatching might block on putting tasks into that channel, and so it needs to be made nonblocking (all credit to Zijian/@shaddoll for thinking of this / Steven for suggesting that also invalid channels should be checked)

### Changes 
- Refactors async dispatch code into two sub-functions to make it easier to read
- Adds some tests around blocking behaviour
- **Adds default** on select to prevent this from blocking (the actual bugfix)
- Adds check on redirection to ensure there's no invalid channel pushing, since that will also block, I'm told (til)

**How did you test it?**


service/matching/taskReader.go
```diff
@@ -421,3 +376,90 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+				case tr.taskBuffers[group] <- taskInfo:
+					// successful redirect
+					tr.scope.IncCounter(metrics.BufferIsolationGroupRedirectCounter)
+					tr.logger.Warn("some tasks were redirected to another isolation group", tag.IsolationGroup(group),
+						tag.WorkflowRunID(taskInfo.RunID),
+						tag.WorkflowID(taskInfo.WorkflowID),
+						tag.TaskID(taskInfo.TaskID),
+						tag.WorkflowDomainID(taskInfo.DomainID),
+					)
+				default:
```
davidporter-id-au, line 442: This, here, is the single line hotfix basically. The default should prevent any attempts to put stuff on the channel from deadlocking if it's full.

service/matching/taskReader.go
```diff
@@ -150,59 +155,10 @@ dispatchLoop:
 			if !ok { // Task list getTasks pump is shutdown
 				break dispatchLoop
 			}
-			for {
```
davidporter-id-au, line 153: pulled this into a function because it was annoying to test

service/matching/taskReader.go
```diff
@@ -421,3 +376,88 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+		err := tr.dispatchTask(dispatchCtx, task)
+		timerScope.Stop()
+		cancel()
+		if err == nil {
+			return
+		}
+		if err == context.Canceled {
+			tr.logger.Info("Tasklist manager context is cancelled, shutting down")
+			breakDispatchLoop = true
+			return
```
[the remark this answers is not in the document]
davidporter-id-au: yeah, good feedback, I was just doing early iteration. I've updated it to subdivide it further for easier testing.

service/matching/taskReader.go
```diff
@@ -421,3 +376,88 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+				} else {
+					// it should never happen, unless there is a bug in 'getIsolationGroupForTask' method
+					tr.logger.Error("taskReader: unexpected error getting isolation group", tag.Error(err))
+					tr.completeTask(taskInfo, err)
+				}
+				break
+			}
+			if group == isolationGroup {
+				continue
+			} else {
```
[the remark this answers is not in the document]
davidporter-id-au: I would normally always go for this, but also am trying to limit the amount of changes I'm making, but maybe in here it's worth the additional improvement. Let me update.

service/matching/taskReader.go
```diff
@@ -421,3 +376,120 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+			return false, false
+		}
+
+		// ensure the isolation group is configured and available
+		_, taskGroupReaderIsPresent := tr.taskBuffers[group]
+		if !taskGroupReaderIsPresent {
+			// there's a programmatic error. Something has gone wrong with tasklist instantiation
+			// don't block and redirect to the default group
+			tr.scope.IncCounter(metrics.BufferIsolationGroupRedirectFailureCounter)
+			tr.logger.Error("An isolation group buffer was misconfigured and couln't be found. Redirecting to default",
```
Groxx, line 441:
```suggestion
			tr.logger.Error("An isolation group buffer was misconfigured and couldn't be found. Redirecting to default",
```

service/matching/taskReader.go
```diff
@@ -421,3 +376,120 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+	timerScope.Stop()
+	cancel()
+	if err == nil {
+		return false, true
+	}
+	if err == context.Canceled {
+		tr.logger.Info("Tasklist manager context is cancelled, shutting down")
+		return true, true
+	}
+	if err == context.DeadlineExceeded {
```
taylanisikdemir, line 404: we can also early terminate here 
```
if err != deadline exceeded {
   // line 491-494 goes here
}

// handle deadline exceeded

File: service/matching/taskReader.go [hunk 7]
```diff
@@ -421,3 +376,120 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+			// no change, retry to dispatch the task again
+			return false, false
+		}
+
+		// ensure the isolation group is configured and available
+		_, taskGroupReaderIsPresent := tr.taskBuffers[group]
+		if !taskGroupReaderIsPresent {
+			// there's a programmatic error. Something has gone wrong with tasklist instantiation
+			// don't block and redirect to the default group
+			tr.scope.IncCounter(metrics.BufferIsolationGroupRedirectFailureCounter)
```
taylanisikdemir, line 440: do you think it would be useful to also include `isolationGroup` in the error message? same applies to other log messages below

taylanisikdemir (approved):

service/matching/taskReader.go
```diff
@@ -421,3 +376,120 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
 	}
 	return tr.cancelCtx, func() {}
 }
+
+func (tr *taskReader) dispatchSingleTaskFromBufferWithRetries(isolationGroup string, taskInfo *persistence.TaskInfo) (breakDispatchLoop bool) {
+	// retry loop for dispatching a single task
+	for {
+		breakDispatchLoop, breakRetryLoop := tr.dispatchSingleTaskFromBuffer(isolationGroup, taskInfo)
```
Groxx, line 383: hmmmmm.  double boolean :|
usage looks correct, just wondering if this should be something else to reduce risk...

a brief stab at enums feels even worse, but what if it was a result struct?
```go
// return
return dispatchResult{breakDispatchLoop: true, breakRetryLoop: true}

// check
for {
  result := tr.dispatchSingleTaskFromBuffer(isolationGroup, taskInfo)
  if result.breakRetryLoop {
    result.breakDispatchLoop
  }
}
```
^ more verbose obviously, but also pretty hard to screw up vs `return true, false` and relying on correct (but not required) var naming patterns.

service/matching/taskReader.go
```diff
@@ -421,3 +376,120 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+			)
+			// the task async buffers on the other isolation-group are already full, wait and retry
+		}
+	}
+	// this should never happen unless there is a bug - don't drop the task
+	tr.scope.IncCounter(metrics.BufferThrottlePerTaskListCounter)
+	tr.logger.Error("taskReader: unexpected error dispatching task", tag.Error(err))
+	runtime.Gosched()
+	return false, false
+}
```
Groxx, line 495: probably deserves a comment update at the very least, but I'm gonna pause on this and read a bit deeper in general:

since this changed from blocking to non-blocking, this is _essentially_ now no longer true - we'd expect it to occur when the destination buffer is full.

I'm not entirely sure what the consequences are of this though.  e.g. it'll try to re-dispatch to the original (now known to be old) group - is that fine?
it seems like probably yes, it'll just time out again and reach these lines again?  that probably serves as a throttler on the whole process to make sure it doesn't hot-loop, which would be good, as long as it doesn't incorrectly perturb the order or something.

Groxx: yeah, that does look safe to me.  waits on a ratelimit, tries to dispatch locally or forward, then fails.  timeout or "successfully dispatched" both seem like reasonable outcomes.

I'm not familiar with exactly what a forwarded task implies though.  I'm guessing no, but is there a chance this bounces around between hosts, wasting more resources than if it stayed here?

common/metrics/defs.go
```diff
@@ -2912,6 +2915,9 @@ var MetricDefs = map[ServiceIdx]map[int]metricDefinition{
 		RespondQueryTaskFailedPerTaskListCounter:    {metricName: "respond_query_failed_per_tl", metricRollupName: "respond_query_failed"},
 		SyncThrottlePerTaskListCounter:              {metricName: "sync_throttle_count_per_tl", metricRollupName: "sync_throttle_count"},
 		BufferThrottlePerTaskListCounter:            {metricName: "buffer_throttle_count_per_tl", metricRollupName: "buffer_throttle_count"},
+		BufferIsolationGroupRedirectCounter:         {metricName: "buffer_isolation_group_redirected", metricRollupName: "buffer_isolation_group_redirected"},
+		BufferIsolationGroupRedirectFailureCounter:  {metricName: "buffer_isolation_group_redirect_failure", metricRollupName: "buffer_isolation_group_redirect_failure"},
+		BufferIsolationGroupMisconfiguredCounter:    {metricName: "buffer_isolation_group_misconfigured_failure", metricRollupName: "buffer_isolation_group_misconfigured_failure"},
```
Groxx, line 2920: likely needs to follow the `_tl` suffix pattern, as I would assume these scopes have tasklist tags already included (so the rollup is desired + needs a different name)

service/matching/taskReader.go
davidporter-id-au: Yeah, I think it's safe to retry / won't hotloop, in isolated cases it'll time out, if non-isolated case it'll block, in both I don't think it'll spin. 

The forwarding is to forward on to the parent tasklist in tasklist partitioning trees. I don't think it affects this (As far as I can tell)

service/matching/taskReader.go
davidporter-id-au: I thought about it, but decided against it because we don't have a tag specifically for this kind of key (we're pretty disciplined about which log tags get used) and it can be derived by looking up the task with the taskID or the workflow + runID

service/matching/taskReader.go
davidporter-id-au: I don't feel super strongly, but I'm not super sure it's much safer tbh. The defaulting behaviour is not particularly different and Imho unless there's a way to differentiate undefined values from false, it ends up with the same net effect

service/matching/taskReader.go
davidporter-id-au: uh... not sure I understood sorry?

taylanisikdemir: [This block](https://github.com/uber/cadence/blob/b2667129d5cfe630ea6f07d908c82b3e8487314c/service/matching/taskReader.go#L404-L489) is unnecessary level of indentation which can be avoided by handling DeadlineExceeded first. Updated my example above to make the early termination obvious.

service/matching/taskReader.go
taylanisikdemir: Got it if we have a workaround and we think this will be super rare troubleshooting scenario no need to have that extra info as a log tag. Still could be baked into the message if you think it would be useful. 
`tr.logger.Error("Isolation group %s is missing task buffer. Task originally was assigned to group %s. Redirecting to default", tags...)`

service/matching/taskReader.go
Groxx: yea, I mean more that relying on IDE variable name autocomplete (particularly in the age of LLMs) for review-correctness seems risky :|

not a big thing, obviously.  just pondering.

service/matching/taskReader.go
```diff
@@ -421,3 +376,118 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+			tr.logger.Warn("some tasks were redirected to another isolation group", tag.IsolationGroup(group),
+				tag.WorkflowRunID(taskInfo.RunID),
+				tag.WorkflowID(taskInfo.WorkflowID),
+				tag.TaskID(taskInfo.TaskID),
+				tag.WorkflowDomainID(taskInfo.DomainID),
+			)
+			return false, true
+		default:
+			tr.scope.IncCounter(metrics.BufferIsolationGroupRedirectFailureCounter)
+			tr.logger.Error("some tasks could not be redirected to another isolation group as the buffer's already full", tag.IsolationGroup(group),
```
Groxx, line 481:
```suggestion
			tr.logger.Error("some tasks could not be redirected to another isolation group as the buffer's already full",
				tag.IsolationGroup(group),
```

service/matching/taskReader.go
```diff
@@ -421,3 +376,118 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+		// we want to redistribute the tasks to other isolation groups in this case to drain
+		// the backlog.
+		select {
+		case <-tr.cancelCtx.Done():
+			// the task reader is shutting down
+			return true, true
+		case tr.taskBuffers[group] <- taskInfo:
+			// successful redirect
+			tr.scope.IncCounter(metrics.BufferIsolationGroupRedirectCounter)
+			tr.logger.Warn("some tasks were redirected to another isolation group", tag.IsolationGroup(group),
```
Groxx, line 472:
```suggestion
			tr.logger.Warn("some tasks were redirected to another isolation group",
				tag.IsolationGroup(group),
```

service/matching/taskListManager_test.go
```diff
@@ -388,3 +388,85 @@ func TestGetPollerIsolationGroup(t *testing.T) {
...
+	// We should see them all being redirected
+	assert.Len(t, tlm.taskReader.taskBuffers["datacenterB"], maxBufferSize)
+	// we should see none in DC A (they're not being redirected here)
+	assert.Len(t, tlm.taskReader.taskBuffers["datacenterA"], 0)
+
+	// ok, and here we try and ensure that this *does not block
+	// and instead complains and live-retries
+	breakDispatcher, breakRetryLoop := tlm.taskReader.dispatchSingleTaskFromBuffer("datacenterA", &persistence.TaskInfo{})
+	assert.False(t, breakDispatcher, "dispatch isn't shutting down")
+	assert.False(t, breakRetryLoop, "in the event of a task being unable to be cross-dispatched, the expectation is that it'll keep retrying")
```
Groxx, line 430:
```suggestion
	// ok, and here we try and ensure that this *does not block
	// and instead complains and live-retries
	breakDispatcher, breakRetryLoop := tlm.taskReader.dispatchSingleTaskFromBuffer("datacenterA", &persistence.TaskInfo{})
	assert.False(t, breakDispatcher, "dispatch isn't shutting down")
	assert.False(t, breakRetryLoop, "in the event of a task being unable to be cross-dispatched, the expectation is that it'll keep retrying")
	
	// and test again, to cover max-buffer +1, to make sure "keep retrying"
	// is a safe and stable state to be in.
	breakDispatcher, breakRetryLoop := tlm.taskReader.dispatchSingleTaskFromBuffer("datacenterA", &persistence.TaskInfo{})
	assert.False(t, breakDispatcher, "dispatch isn't shutting down")
	assert.False(t, breakRetryLoop, "in the event of a task being unable to be cross-dispatched, the expectation is that it'll keep retrying")
```

Groxx: hmm.  or maybe I misinterpreted.  just seems odd that max-minus-1 is actually "full", and it doesn't occur when _exceeding_ max.

service/matching/taskListManager_test.go
```diff
@@ -388,3 +388,85 @@ func TestGetPollerIsolationGroup(t *testing.T) {
...
+}
+
+// This is a bit of a strange unit-test as it's
+// ensuring that invalid behaviour is handled defensively.
+// It *should never be the case* that the isolation group tries to
+// dispatch to a buffer that isn't there, however, if it does, we want to just
+// log this, emit a metric and fallback to the default isolation group.
+func TestMisconfiguredZoneDoesNotBlock(t *testing.T) {
```
Groxx, line 438: since these are kinda suspected to be a regression test:
did they fail (or get stuck) with the old logic?

they seem useful either way, just wondering if that was checked.

Groxx (approved):
yea, looks good to me.  a minor Q and nits but afaict it's correct and also an improvement.  merge whenever.

service/matching/taskListManager_test.go
davidporter-id-au: yes, I made sure they were failing before making the change

service/matching/taskReader.go
davidporter-id-au: after debugging I completely agree, let me address

service/matching/taskReader.go
davidporter-id-au: Yeah, this is messy. Let me see if I can make it better.

service/matching/taskReader.go
davidporter-id-au: added with some tag.Dynamic tags. I was not wanting to template the log message just because it can make searching tricky

service/matching/taskListManager_test.go
davidporter-id-au: good call

service/matching/taskReader.go
```diff
@@ -421,3 +377,149 @@ func (tr *taskReader) newDispatchContext(isolationGroup string) (context.Context
...
+				tag.WorkflowID(taskInfo.WorkflowID),
+				tag.TaskID(taskInfo.TaskID),
+				tag.WorkflowDomainID(taskInfo.DomainID),
+			)
+			// the task async buffers on the other isolation-group are already full, wait and retry
+			return false, false
+		}
+	}
+
+	if errors.Is(err, ErrTasklistThrottled) {
```
davidporter-id-au, line 510: callout, this is a behaviour change, this previously would just fall into this section for all non-recognised errors and emit the metric for all. I've narrowed the scope that this can happen and put an explicit log around the unrecogised error case which I think should never be valid.

service/matching/taskReader.go
davidporter-id-au: I realised we were handling unknown errors in a way which I think was wrong (they were just assumed to be rate-limit errors, but that seems... naieve. So ended up creating a default catchall for unknown errors as I feel this is more readable

service/matching/taskListManager_test.go
Groxx: ah, yeah, in retrospect I guess they were intended to block there, aren't they.
still: thanks!

service/matching/matcher.go
```diff
@@ -61,7 +61,8 @@ const (
 	_defaultTaskDispatchRPSTTL = 60 * time.Second
 )
 
-var errTasklistThrottled = errors.New("cannot add to tasklist, limit exceeded")
+// ErrTasklistThrottled implies a tasklist was throttled from the client-side
```
Groxx, line 64: hmm.  throttled how?
is this the "10/s global tasklist response-limit" throttle that client-library-users can set in their worker config, or something else?

service/matching/matchingEngine.go
```diff
@@ -896,11 +896,6 @@ func (e *matchingEngineImpl) getTask(
 	return tlMgr.GetTask(ctx, maxDispatchPerSecond)
 }
 
-//
-//func (e *matchingEngineImpl) getHistoryInfo() (HistoryInfo, error) {
-//
-//}
-
```
Groxx, line 903: huh.  wonder why this (and HistoryInfo's definition) were introduced as commented-out.

service/matching/taskReader.go
Groxx: useful to separate I think, but we almost definitely want a metric on the unknown-error case.  harder to notice if it's not a metric.

davidporter-id-au: true, let me add

service/matching/matcher.go
davidporter-id-au: I was under the impression this was tasklist based throttling set in client config?

service/matching/matchingEngine.go
davidporter-id-au: zero clue

service/matching/matcher.go
Groxx: pretty sure it's not:
```go
	dPtr := _defaultTaskDispatchRPS
	limiter := quotas.NewRateLimiter(&dPtr, _defaultTaskDispatchRPSTTL, config.MinTaskThrottlingBurstSize())
```
^ the only non-constant there is the burst-size argument, and that's coming from dynamic config, not the client AFAICT:
```go
		MinTaskThrottlingBurstSize: func() int {
			return config.MinTaskThrottlingBurstSize(domainName, taskListName, taskType)
		},
```

Groxx: yea, looks like this is just a constant ratelimiter (100,000/s) with configurable burst size (.... that seems a bit nuts with 100k/s, but alright) that can be updated from dynamic config every minute.

feels more like hot-loop protection or something to me.  exceeding 100k/s would probably require both large amounts of data and running fully in memory, and I'm not sure that's feasible here without bugs.

service/matching/taskReader.go
```diff
@@ -394,19 +395,28 @@ func (tr *taskReader) dispatchSingleTaskFromBuffer(isolationGroup string, taskIn
...
 	}
-	if err == context.Canceled {
+
+	if errors.Is(err, context.Canceled) {
 		tr.logger.Info("Tasklist manager context is cancelled, shutting down")
 		return true, true
 	}
-	if err == context.DeadlineExceeded {
+
+	if errors.Is(err, context.DeadlineExceeded) {
```
Groxx, line 408: 👍 seems very likely safe and correct.  code here doesn't seem to wrap anything ever, but this might allow us to.

Groxx (approved):
re-stamping: looks even gooder to me than before, just the comment to fix.
(thanks for adding the comment!  I wasn't sure what that one meant either, now we know)

service/matching/matcher.go
davidporter-id-au: my mistake, I do not know precisely. I'll update

PR #5426, opened 2023-10-19
Range query for custom attribute

bowenxia:
**What changed?**
Added "between and" for custom attributes
Added <, >, <=, >= for custom attributes

**Why?**
Previously custom attributes only support equal operation

**How did you test it?**
unit test
Integration test

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/pinot/pinotVisibilityStore_test.go
```diff
@@ -185,7 +185,7 @@ LIMIT 0, 10
 FROM %s
 WHERE DomainID = 'bfd5c907-f899-4baf-a7b2-2ab85e623ebd'
 AND IsDeleted = false
-AND WorkflowID = 'wid' and ((JSON_MATCH(Attr, '"$.CustomStringField" is not null') AND REGEXP_LIKE(JSON_EXTRACT_SCALAR(Attr, '$.CustomStringField', 'string'), 'custom and custom2 or custom3 order by*')) or CustomIntField between 1 and 10)
```
bowenxia, line 188: previously we didn't find the range query was not working since we used the old pinot table which we had flatten attribute. So the range query "between and" worked.

common/pinot/pinotQueryValidator_test.go
neil-xie, line 1: one qq: do we have range query test case in the integration test?

common/pinot/pinotQueryValidator.go
```diff
@@ -91,24 +91,56 @@ func (qv *VisibilityQueryValidator) validateWhereExpr(expr sqlparser.Expr) (stri
...
-		//return qv.validateRangeExpr(expr)
+		return qv.validateRangeExpr(expr)
 	case *sqlparser.ParenExpr:
 		return qv.validateWhereExpr(expr.Expr)
 	default:
 		return "", errors.New("invalid where clause")
 	}
 }
 
+func (qv *VisibilityQueryValidator) validateRangeExpr(expr sqlparser.Expr) (string, error) {
```
neil-xie, line 108: Does this range query validator only handle (BETWEEN * AND *) query?

bowenxia: Yes. <, > things are also handled validateComparisonExpr()

neil-xie: Could you add comment for those 2 methods? Usually range query includes >,<,>=,<= these conditions, not just the "between" query.

common/pinot/pinotQueryValidator_test.go
```diff
@@ -113,6 +113,43 @@ func TestValidateQuery(t *testing.T) {
...
+		"Case15-1: custom date attribute less than": {
+			query:     "CustomDatetimeField < 1697754674",
+			validated: "(JSON_MATCH(Attr, '\"$.CustomDatetimeField\" is not null') AND CAST(JSON_EXTRACT_SCALAR(Attr, '$.CustomDatetimeField') AS BIGINT) < 1697754674)",
+		},
+		"Case15-2: custom date attribute greater than or equal to": {
+			query:     "CustomDatetimeField >= 1697754674",
+			validated: "(JSON_MATCH(Attr, '\"$.CustomDatetimeField\" is not null') AND CAST(JSON_EXTRACT_SCALAR(Attr, '$.CustomDatetimeField') AS BIGINT) >= 1697754674)",
+		},
+		"Case15-3: system date attribute greater than or equal to": {
+			query:     "StartTime >= 1697754674",
```
neil-xie, line 133: When user input a date string like StartTime > "2023-11-02T15:36:47", is this handled in the query validator?

neil-xie (approved):
LGTM

common/pinot/pinotQueryValidator_test.go
bowenxia: This is not handled here. But that will be converted to `UnixMilli()` in PiontVisibilityStore.

PR #5429, opened 2023-10-24
Introduce User entity

mantas-sidlauskas:
**What changed?**
Adding `user.Info` entity

**Why?**
`user.Info` will be used to authenticate user and generate JWT token for authorization 

**How did you test it?**

**Potential risks**
None

**Release notes**

**Documentation Changes**


common/user/info.go
```diff
@@ -0,0 +1,32 @@
...
+// SOFTWARE.
+
+package user
+
+// Info holds information about user
+// This information is used to generate corresponding JWT claim
+type Info struct {
+	Name     string
+	Password string
+	Admin    bool
```
davidporter-id-au, line 30: Same as before, role, probably a string or enum is fine. Enum requires making IDL changes every time there's a role change, which may be a annoying, but is clearer. String allows for finer and faster iteration on what roles may actually do.

common/user/info.go
```diff
@@ -0,0 +1,32 @@
...
+	Groups   []string
```
davidporter-id-au, line 31: What is group?

davidporter-id-au (approved):

davidporter-id-au (changes requested):
sorry, I meant to comment before but accidentally hit approve, I do think the role/group thing needs clarification.

common/user/info.go
mantas-sidlauskas: In Cadence we have OAuth authorization already implemented. This struct mainly reflects model for authorization already in use.

On server side, you can update domain metadata with keys: READ_GROUPS and WRITE_GROUPS to define what "group" has what permissions. Then groups and admin flag are checked in JWT claim verification part when "access controlled handler" is used.

References to get more context: 
* [JWT Claim struct](https://github.com/uber/cadence/blob/5f9eb08abfdc1f5ec68b097a5e5e93a6fcdf283c/common/authorization/oauthAuthorizer.go#L48)
* [JWT claim verification](https://github.com/uber/cadence/blob/5f9eb08abfdc1f5ec68b097a5e5e93a6fcdf283c/common/authorization/oauthAuthorizer.go#L148)
* [Access controlled handler authorize flow example](https://github.com/uber/cadence/blob/5f9eb08abfdc1f5ec68b097a5e5e93a6fcdf283c/service/frontend/accessControlledHandler.go#L747)

PR #5432, opened 2023-10-27
Added nit comments describing the Update workflow cycle

agautam478:
- Added nit comments in the execution update cycle flow for educational purposes.
- Explained how the StartTime Stamp gets populated in case of different type of workflows. 

These changes were made to ensure that the future developers are aware of these patterns and make changes accordingly.

tested locally.

NA

NA
<!-- Is there any documentation updates should be made for config, https://cadenceworkflow.io/docs/operation-guide/setup/ NA


service/history/execution/mutable_state_builder.go
```diff
@@ -114,6 +114,12 @@ type (
 		updateBufferedEvents []*types.HistoryEvent // buffered history events that needs to be persisted
 		clearBufferedEvents  bool                  // delete buffered events from persistence
 
+		// This section includes Workflow Execution Info parameters like StartTimestamp,
+		//which are only visible after the workflow has begun.
```
taylanisikdemir, line 118: nit: let's add space between `//` and the first word to be consistent.

taylanisikdemir (approved):

PR #5437, opened 2023-11-01
Add log user query param

bowenxia:
**What changed?**
Add log user query param and filtered out StartTime range queries

**Why?**
for Pinot/ES comparator usage

**How did you test it?**
unit test, integration test

**Potential risks**

**Release notes**

**Documentation Changes**


neil-xie:
How do you think adding a flipr flag here to turn on user query log? We don't to need keep this feature running for all the environments all the time.

common/persistence/pinotUserQueryTimeFilter.go
```diff
@@ -0,0 +1,170 @@
...
+func (qv *VisibilityQueryFilter) filterRangeExpr(expr sqlparser.Expr) (string, error) {
+	buf := sqlparser.NewTrackedBuffer(nil)
+	rangeCond := expr.(*sqlparser.RangeCond)
+	colName, ok := rangeCond.Left.(*sqlparser.ColName)
+	if !ok {
+		return "", errors.New("invalid range expression: fail to get colname")
+	}
+	colNameStr := colName.Name.String()
+
+	if colNameStr == definition.StartTime {
```
neil-xie, line 111: You filtered the startTime query here, is it because startTIme will be defined in the comparator workflow?

service/history/resource/resource.go
```diff
@@ -93,9 +93,10 @@ func New(
 			PersistenceGlobalMaxQPS: config.PersistenceGlobalMaxQPS,
 			ThrottledLoggerMaxRPS:   config.ThrottledLogRPS,
 
-			EnableReadVisibilityFromES:    nil, // history service never read,
-			AdvancedVisibilityWritingMode: config.AdvancedVisibilityWritingMode,
-			EnableReadVisibilityFromPinot: nil, // history service never read,
+			EnableReadVisibilityFromES:      nil, // history service never read,
+			AdvancedVisibilityWritingMode:   config.AdvancedVisibilityWritingMode,
+			EnableReadVisibilityFromPinot:   nil, // history service never read,
+			EnableLogCustomerQueryParameter: nil,
```
neil-xie, line 99: Let's add a comment here to explain why we pass nil for this parameter

service/frontend/service.go
```diff
@@ -45,9 +45,10 @@ type Config struct {
 	EnableVisibilitySampling        dynamicconfig.BoolPropertyFn
 	EnableReadFromClosedExecutionV2 dynamicconfig.BoolPropertyFn
 	// deprecated: never used for ratelimiting, only sampling-based failure injection, and only on database-based visibility
-	VisibilityListMaxQPS          dynamicconfig.IntPropertyFnWithDomainFilter
-	EnableReadVisibilityFromES    dynamicconfig.BoolPropertyFnWithDomainFilter
-	EnableReadVisibilityFromPinot dynamicconfig.BoolPropertyFnWithDomainFilter
+	VisibilityListMaxQPS            dynamicconfig.IntPropertyFnWithDomainFilter
+	EnableReadVisibilityFromES      dynamicconfig.BoolPropertyFnWithDomainFilter
+	EnableReadVisibilityFromPinot   dynamicconfig.BoolPropertyFnWithDomainFilter
+	EnableLogCustomerQueryParameter dynamicconfig.BoolPropertyFn
```
neil-xie, line 51: It is better to use BoolPropertyFnWithDomainFilter, so we can turn on/off this for some specific domains, for example we can turn this off for system domains

config/dynamicconfig/development_pinot.yaml
```diff
@@ -6,6 +6,8 @@ system.enableReadVisibilityFromES:
   - value: false
 system.enableReadVisibilityFromPinot:
   - value: true
+system.enableLogCustomerQueryParameter:
```
neil-xie, line 9: default should be false, since OSS doesn't need this functionality

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -288,82 +295,248 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
+		isOpen:       true,
+		domainName:   request.Domain,
+		workflowType: "",
+		workflowID:   "",
+		closeStatus:  math.MinInt, // set the default close status to be Integer.MIN_VALUE
+		customQuery:  "",
+	})
+
 	manager := v.chooseVisibilityManagerForRead(request.Domain)
+	override := ctx.Value("visibility-override")
```
neil-xie, line 352: We can write in one statement 
```
if override := ctx.Value("visibility-override"); override == "ES" {
		manager = v.esVisibilityManager
}
```

neil-xie: Currently the default reader is ES, the logic here is the reverse way.

neil-xie: "ES" and "visibility-override" could put in constant string so in the future you can change them easily

neil-xie: Maybe change like this, so it can work no matter what default option is, and even if it is not set override, it still works
```
if override := ctx.Value("visibility-override"); override == "ES" {
        manager = v.esVisibilityManager
} else if override == "Pinot" {
        manager = v.pinotVisibilityManager
}
```

common/persistence/pinotUserQueryTimeFilter.go
bowenxia: yes

common/persistence/pinotVisibilityTripleManager.go
bowenxia: do we really need to set override value for both cases? I was thinking to only have the attribute `override` when we really need to "override" something.

neil-xie: Before we fully switch to Pinot, it will read from ES by fault, so we need if override == "Pinot" case. I think the question will if we need to keep this comparator running after we switch to Pinot. If so, we gonna need  if override == "es" case.

common/dynamicconfig/constants.go
```diff
@@ -3683,9 +3688,16 @@ var BoolKeys = map[BoolKey]DynamicBool{
 	},
 	EnableReadVisibilityFromPinot: DynamicBool{
 		KeyName:      "system.enableReadVisibilityFromPinot",
+		Filters:      []Filter{DomainName},
```
bowenxia, line 3691: We previously forgot to add this here.

common/persistence/pinotUserQueryTimeFilter.go
```diff
@@ -0,0 +1,172 @@
...
+type VisibilityQueryFilter struct {
+}
+
+// NewVisibilityQueryFilter create VisibilityQueryValidator
+func NewVisibilityQueryFilter() *VisibilityQueryFilter {
+	return &VisibilityQueryFilter{}
+}
+
+// FilterQuery validates that search attributes in the query and returns modified query.
+func (qv *VisibilityQueryFilter) FilterQuery(whereClause string) (string, error) {
```
neil-xie, line 47: Can we reuse queryValidator here, it is the same code

bowenxia: not quite. This one is a much easier version that we only need to replace everything that contains "StartTime" to be "". but, if we want to re-use the queryValidator, there'll be more parameters that needed to be passed in, and there'll be more branches, which can make the code more complicated. 

We can make the queryValidator to be an interface so that this filter can implement the interface, but given that this is only a temporary component, I think that this might be an overkill?

common/persistence/pinotUserQueryTimeFilter.go
neil-xie: What if user passed CloseTime=missing, since we have list open conditions in comparator workflow.

common/persistence/pinotUserQueryTimeFilter.go
neil-xie: Thanks for the explanation! Yeah, it might be too much to change the existing validator. We need to deprecate this after we remove ES anyway. You can add what you said above in the comment and add one more line says "Will be removed after fully operate on Pinot".

neil-xie:
Is there any reason we parse the query here instead of doing it in the workflow? I am asking because if we moved to the workflow code, it will be much easier here to log and export the queries, we don't need to remove the code after migration, etc.

bowenxia:
> Is there any reason we parse the query here instead of doing it in the workflow? I am asking because if we moved to the workflow code, it will be much easier here to log and export the queries, we don't need to remove the code after migration, etc.

You are right. We can put it in the workflow.

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -290,91 +300,279 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
 	ctx context.Context,
 	request *ListWorkflowExecutionsRequest,
 ) (*ListWorkflowExecutionsResponse, error) {
+	v.logUserQueryParameters(userParameters{
+		operation:    "list",
+		isOpen:       true,
+		domainName:   request.Domain,
+		workflowType: "",
+		workflowID:   "",
+		closeStatus:  math.MinInt, // set the default close status to be Integer.MIN_VALUE
```
neil-xie, line 346: Let's use -1 here. When we record workflow started, we use -1 as close status

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -290,91 +300,279 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
+
 func (v *pinotVisibilityTripleManager) ListOpenWorkflowExecutions(
 	ctx context.Context,
 	request *ListWorkflowExecutionsRequest,
 ) (*ListWorkflowExecutionsResponse, error) {
+	v.logUserQueryParameters(userParameters{
+		operation:    "list",
+		isOpen:       true,
+		domainName:   request.Domain,
+		workflowType: "",
```
neil-xie, line 344: We don't need to pass these empty strings, it will have default value of "". You could just remove these fields in this passed in struct. Like:
`v.logUserQueryParameters(userParameters{
		operation:    "list",
		isOpen:       true,
		domainName:   request.Domain,
		closeStatus:  math.MinInt, // set the default close status to be Integer.MIN_VALUE
	}, request.Domain)`

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -290,91 +300,279 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
+	}
 	return manager.ListClosedWorkflowExecutions(ctx, request)
 }
 
 func (v *pinotVisibilityTripleManager) ListOpenWorkflowExecutionsByType(
 	ctx context.Context,
 	request *ListWorkflowExecutionsByTypeRequest,
 ) (*ListWorkflowExecutionsResponse, error) {
+	v.logUserQueryParameters(userParameters{
+		operation:    "list",
```
neil-xie, line 386: It is better to put the operations as enum and cast to string when you log it. At least you could use constants here.

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -290,91 +309,253 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
+
 func (v *pinotVisibilityTripleManager) ListOpenWorkflowExecutions(
 	ctx context.Context,
 	request *ListWorkflowExecutionsRequest,
 ) (*ListWorkflowExecutionsResponse, error) {
+	v.logUserQueryParameters(userParameters{
+		operation:   string(Operation.LIST),
+		isOpen:      true,
+		domainName:  request.Domain,
+		closeStatus: -1, // set the default close status to be Integer.MIN_VALUE
```
neil-xie, line 353: Comment need update

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -33,16 +34,34 @@ import (
...
+	Primary   = "Primary"
+	Secondary = "Secondary"
+)
+
+type OperationType string
+
+var Operation = struct {
+	LIST                   OperationType
+	COUNT                  OperationType
+	LISTWORKFLOWEXECUTIONS OperationType
```
neil-xie, line 58: This is not used by any methods, I think you forgot to update

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -290,91 +309,240 @@ func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForWrite(ctx conte
...
 
 func (v *pinotVisibilityTripleManager) ListClosedWorkflowExecutionsByType(
 	ctx context.Context,
 	request *ListWorkflowExecutionsByTypeRequest,
 ) (*ListWorkflowExecutionsResponse, error) {
+	v.logUserQueryParameters(userParameters{
+		operation:    string(Operation.LIST),
+		domainName:   request.Domain,
+		workflowType: request.WorkflowTypeName,
+		closeStatus:  0,
```
neil-xie, line 407: When we list closed workflows, you set the closeStatus to 0. Do you know if it should include the terminated or cancelled workflows whose closeStatus is not 0.

bowenxia: What I was thinking is: as we are converting to a CLI cmd in the comparator workflow, we only need the status to be either open or closed (Either need the `--open` flag or not). So, as long as it is one of the closed status, it should be okay. The only API that cares about closeStatus is `ListXXXByStatus`.

neil-xie: OK, so you mean when clostStatus != 0 or 1, then we know it is ListXXXByStatus?

neil-xie (approved):
LGTM

PR #5438, opened 2023-11-01
Improves history handler error metrics and logs

davidporter-id-au:
**What changed?**

Some bad data caused a tasklist to get stuck while trying to update a workflow which was in a bad state. This was more difficult to identify than it should be because the RunID was missing information from the error logs. This change includes this additional information and also protects some of the metrics emission from wrapped errors being misclassified. 

This is not an attempt to do all error handling / wrapping correctly, it's just an incremental change, of which many more are needed. 

**Why?**

**How did you test it?**

Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -2141,7 +2141,7 @@ func (h *handlerImpl) updateErrorMetric(
 
 	var yarpcE *yarpcerrors.Status
 
-	if err == context.DeadlineExceeded || err == context.Canceled {
+	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
 		scope.IncCounter(metrics.CadenceErrContextTimeoutCounter)
 		return
 	}
--- a/service/history/handler.go
+++ b/service/history/handler.go
@@ -308,34 +308,34 @@ func (h *handlerImpl) RecordActivityTaskHeartbeat(
 
 	domainID := wrappedRequest.GetDomainUUID()
 	if domainID == "" {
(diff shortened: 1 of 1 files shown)


allenchen2244 (approved):

service/history/handler.go
```diff
@@ -2110,51 +2135,74 @@ func (h *handlerImpl) updateErrorMetric(
...
 		scope.IncCounter(metrics.CadenceErrServiceBusyCounter)
-	case *yarpcerrors.Status:
-		if err.Code() == yarpcerrors.CodeDeadlineExceeded {
+
+	} else if errors.As(err, &yarpcE) {
+
+		if yarpcE.Code() == yarpcerrors.CodeDeadlineExceeded {
 			scope.IncCounter(metrics.CadenceErrContextTimeoutCounter)
 		}
 		scope.IncCounter(metrics.CadenceFailures)
```
taylanisikdemir, line 2187: not related to your change: why do we increment both `CadenceErrContextTimeoutCounter` and `CadenceFailures` in this case? if there are dashboards looking at sum of these failures we will see extra numbers.

taylanisikdemir (approved):

service/history/handler.go
davidporter-id-au: I don't have context there to be honest

taylanisikdemir: @Groxx do we have dashboards summing these up and showing as total error count?

PR #5445, opened 2023-11-09
Get rid of noisy task adding failure log in matching service

taylanisikdemir:
**What changed?**
[Failed to add task error log](https://github.com/uber/cadence/blob/v1.2.6-prerelease1/service/matching/taskListManager.go#L313) for `remote sync match failed` scenario seems to be too noisy (96% of matching logs). We log it for multiple cases such as
- task forwarded from child partition but domain is not active in current cluster ([ref](https://github.com/taylanisikdemir/cadence/blob/77615779570d7d41c0642c38e237843d80099794/service/matching/taskListManager.go#L283))
- task forwarded from child partition but couldn't be sync matched ([ref](https://github.com/taylanisikdemir/cadence/blob/77615779570d7d41c0642c38e237843d80099794/service/matching/taskListManager.go#L306)) 

There's already a metric `cadence_errors_remote_syncmatch_failed_per_tl` counting this specific case and it has the same breakdown as the log does
- domain
- tasklist name
- operation (adddecision vs addactivity task)

So the change is to remove [this](https://github.com/uber/cadence/blob/v1.2.6-prerelease1/service/matching/taskListManager.go#L313) log line. Other sub categories of `Failed to add task` error log will still be logged by the matching context's handleErr as an `Uncategorized error`. This seems fine to me but open for feedback here. We can keep those other sub types around and just get rid of `remote sync match failed` because it's a known type w/ its own counter. 

Side Change: Only signal taskReader if there's no error and no sync match found. Previously it was signaling even for sync matches but that looked unnecessary as nothing is written to persistence for sync match AFAICT.

**Why?**
Since this log entry adds no extra visibility and floods the logs, there's no point of keeping it. 

**How did you test it?**
`make test`

**Potential risks**
Not signaling task reader for sync matches can have side effects I am not able to see. Looking forward to review feedback on this.


Shaddoll (approved):

service/matching/context.go
```diff
@@ -87,45 +87,44 @@ func (reqCtx *handlerContext) handleErr(err error) error {
 		return nil
 	}
 
-	scope := reqCtx.scope
-
+	// TODO: remove this comment. Added to trigger a buildkite build
```
taylanisikdemir, line 90: reminder for myself to remove

service/matching/taskListManager.go
```diff
@@ -309,13 +309,7 @@ func (c *taskListManagerImpl) AddTask(ctx context.Context, params addTaskParams)
...
 	})
 
-	if err != nil {
-		c.logger.Error("Failed to add task",
-			tag.Error(err),
-			tag.WorkflowTaskListName(c.taskListID.name),
-			tag.WorkflowTaskListType(c.taskListID.taskType),
-		)
-	} else {
+	if err == nil && !syncMatch {
```
taylanisikdemir, line 312: @davidporter-id-au the `!syncMatch` is the change I mentioned offline. With this change `taskReader` is not going to be signaled for successful sync matches because nothing is written to `taskWriter` for those cases

davidporter-id-au: Yeah, I think this is ok. I'd normally suggest pulling this into a separate PR, but it's small enough for it to not be a problem I think.

davidporter-id-au (approved):
For higher-risk changes I'd suggest doing some integration-deployments, will discuss offline. otherwise LGTM

service/matching/taskListManager.go
taylanisikdemir: This seemed like a missed case that can be fixed inline here but if it required additional testing etc. I'd track it in separate PR.

PR #5446, opened 2023-11-09
Adding a framework for the Task validator service.

agautam478:
**What changed?**
The primary goal of this framework is to establish a reliable mechanism for detecting corruption during task processing within the Cadence system. It aims to create a versatile task validation framework applicable throughout the task processing environment. Various validation checks, such as StaleWorkflowCheck and DeprecateDomainCheck, will be introduced to enhance the framework.

The current pull request comprises the fundamental structure of the service, accompanied by its corresponding unit test.

In scenarios where we identify a synchronous API call error from the worker, indicating the completion of a decision task, we occasionally encounter the error "maximum attempts exceeded to update history." This situation proves to be detrimental, as the decision task enters an endless retry loop, causing the entire task list to become stuck.

`err := h.GetTaskValidator().WorkflowCheckforValidation(workflowID, domainID, "")
		if err != nil {
			return err
		}`

To address this, we are testing the invocation of the function call on this reported decision task to observe whether the logs are generated as expected.

**Why?**
These changes were made to ensure a slow incremental introduction of the taskvalidator in the system. We will proceed forward once we have verified that the logs generate as expected.

**How did you test it?**
Added a unit test.
Tested locally.

**Potential risks**
No potential risks.

**Release notes**
NA

**Documentation Changes**
NA


common/pinot/GenericClient_mock.go
```diff
@@ -0,0 +1,100 @@
+// The MIT License (MIT)
```
neil-xie, line 1: this file should be removed

agautam478: will delete it.

service/history/handler.go
```diff
@@ -2214,7 +2216,18 @@ func (h *handlerImpl) error(
...
 ) error {
 	err = h.convertError(err)
+
 	h.updateErrorMetric(scope, domainID, workflowID, runID, err)
+	if errors.Is(err, workflow.ErrMaxAttemptsExceeded) {
+		// Calling the dummy Workflow Check from task Validator. This is an ongoing project where we plan to do some validations on
+		// the following workflow. Based on the validations (is the workflow stale? does the workflow come from a deprecated domain?)
+		// We will delete the workflow or mark the workflow as corrupted.
+		// Placing a dummy call to the function to check the coherency of the design.
+		err := h.GetTaskValidator().WorkflowCheckforValidation(workflowID, domainID, "")
```
allenchen2244, line 2226: we should add some metrics here if this errors.

common/taskvalidator/validateworkflow.go
```diff
@@ -0,0 +1,56 @@
...
+// NewWfChecker creates a new instance of Checker.
+func NewWfChecker(logger log.Logger) Checker {
+	return &checkerImpl{logger: logger}
+}
+
+// WorkflowCheckforValidation is a dummy implementation of workflow validation.
+func (w *checkerImpl) WorkflowCheckforValidation(workflowID string, domainID string, runID string) error {
+	// Emitting just the log to ensure that the workflow is called for now.
+	// TODO: add some validations to check the wf for corruptions.
+	w.logger.Info(fmt.Sprintf("WorkflowCheckforValidation. DomainID: %v, WorkflowID: %v, RunID: %v",
```
allenchen2244, line 51: This is only for errors not every task right? Otherwise we might want to sample this loggiing or else it'll flood

PR #5454, opened 2023-11-20
Create and use a limiter struct instead of just passing a function

jakobht:
Creates a limiter struct and interface instead of just passing a function - will create dedicated factories later

**What changed?**
We now parse a LimiterFactory to the quotas.Collection instead of just a function, this gives the code more structure

**Why?**
The code is very hard to read when all state are kept in closures 

**How did you test it?**
Unit tests still pass

**Potential risks**

**Release notes**

**Documentation Changes**


common/quotas/global.go
```diff
@@ -41,8 +41,3 @@ func PerMember(service string, globalRPS, instanceRPS float64, resolver membersh
 	avgQuota := math.Max(globalRPS/float64(memberCount), 1)
 	return math.Min(avgQuota, instanceRPS)
 }
-
-// PerMemberDynamic is a dynamic variant (using RPSFunc) of PerMember
-func PerMemberDynamic(service string, globalRPS, instanceRPS RPSFunc, resolver membership.Resolver) RPSFunc {
```
3vilhamster, line 46: I'm not sure I understand this change. PerMemberDynamic is removed, though you've introduced much more closures down the code. It looks like it is now more code/harder to read, though easier to understand.

service/frontend/workflowHandler.go
```diff
@@ -174,36 +174,39 @@ func NewWorkflowHandler(
...
 			quotas.NewDynamicRateLimiter(config.UserRPS.AsFloat64()),
-			quotas.NewCollection(func(domain string) quotas.Limiter {
-				return quotas.NewDynamicRateLimiter(quotas.PerMemberDynamic(
-					service.Frontend,
-					config.GlobalDomainUserRPS.AsFloat64(domain),
-					config.MaxDomainUserRPSPerInstance.AsFloat64(domain),
-					resource.GetMembershipResolver(),
-				))
-			}),
+			quotas.NewCollection(quotas.NewDynamicRateLimiterFactory(
```
3vilhamster, line 177: This is definitely better. But I suggest to leave this and put back PerMemberDynamic

common/quotas/global.go
jakobht: This function takes two functions, then returns a function which calls a function with these two functions as input. Additionally, all the callers need to "convert" their functions to match the signatures of this function. For this they use the .AsFloat64 method, which is defined on the function type. 
The function is (IMO) extremely convoluted, which is the reason I removed it. 

Now the callers just pass a simple closure, and thats it.

This is not the important change so we can take it out for now, the other change (your other comment) is the important one. But I vote to keep this change as well 🙂

PR #5456, opened 2023-11-22
added workflow query-types command

arzonus:
A new command `workflow query-types` has been added.

**Why?**
The command will allows users to retrieve the list of available query types. https://github.com/uber/cadence/issues/382

**How did you test it?**
It has been tested locally. 

**Potential risks**
No potential risks, the existing functionality hasn't been touched.

**Release notes**

**Documentation Changes**
https://github.com/uber/Cadence-Docs/pull/168

### DIFF
--- a/tools/cli/workflowCommands.go
+++ b/tools/cli/workflowCommands.go
@@ -629,6 +629,11 @@ func QueryWorkflowUsingStackTrace(c *cli.Context) {
 	queryWorkflowHelper(c, "__stack_trace")
 }
 
+// QueryWorkflowUsingQueryTypes list all query types of the workflow using __query_types as query type
+func QueryWorkflowUsingQueryTypes(c *cli.Context) {
+	queryWorkflowHelper(c, "__query_types")
+}
+
 func queryWorkflowHelper(c *cli.Context, queryType string) {
 	serviceClient := cFactory.ServerFrontendClient(c)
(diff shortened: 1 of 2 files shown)


tools/cli/workflowCommands.go
```diff
@@ -629,6 +629,11 @@ func QueryWorkflowUsingStackTrace(c *cli.Context) {
 	queryWorkflowHelper(c, "__stack_trace")
 }
 
+// QueryWorkflowUsingQueryTypes list all query types of the workflow using __query_types as query type
+func QueryWorkflowUsingQueryTypes(c *cli.Context) {
+	queryWorkflowHelper(c, "__query_types")
```
3vilhamster, line 634: This feature on the client side was not yet released and is not yet supported by java client.
I'm not sure what is the best way to communicate this.

dkrotx: I think that's fine to have CLI doing more than you can do programmatically.

dkrotx (approved):

tools/cli/workflowCommands.go
Groxx: it is, but it's also implying [something] will work.

what's the failure message when this reaches a worker that doesn't support the query?  it should probably be something more descriptive than a default one.

PR #5458, opened 2023-11-22
Update github action for image publishing to released

3vilhamster:
**What changed?**
I updated event that is trigering image publication to subscribe to released event instead of published.
Published is triggered for both prerelease and release and we want to publish images only for releases.
See the docs https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release

**Why**
When we've introduced new process for release process with prereleases we've broke image publication and were publishing images manually.

Should resolve https://github.com/uber/cadence/issues/5416 and https://github.com/uber/cadence/issues/5434

**Release notes**
Fixed image publication on release.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -6,7 +6,7 @@ name: Publish Docker image
 # events but only for the master branch
 on:
   release:
-    types: [ published ]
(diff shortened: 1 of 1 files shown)


3vilhamster:
Will redo from fork

PR #5459, opened 2023-11-22
Simplified the make file to use the standard go tooling

jakobht:
**What changed?**
Simplified the make file to use the standard go tooling

**Why?**
The standard tooling runs the tests in parallel, meaning the time to run the tests goes from 7 minutes to 1.5 minutes. 

**How did you test it?**
Ran the tests, also compared the final coverprofile and they look more or less identical. 

- The counts differ slightly for some lines, compared some runs in different PRs and it looks like they just vary 
- The files from the `host` package are now present in the coverfile with count 0 (they were not in the file before), I don't think it's an issue?

**Potential risks**

**Release notes**

**Documentation Changes**


Makefile
```diff
@@ -529,9 +529,7 @@ test: bins ## Build and run all tests. This target is for local development. The
 	$Q rm -f test.log
 	$Q echo Running special test cases without race detector:
 	$Q go test -v ./cmd/server/cadence/
-	$Q for dir in $(PKG_TEST_DIRS); do \
-		go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log; \
-	done;
+	$Q go test $(TEST_ARG) -coverprofile=$@ ./... $(TEST_TAG) | tee -a test.log
```
Groxx, line 532: `-coverprofile=` needs to be changed, `test` isn't correct here

Makefile
```diff
@@ -551,16 +549,11 @@ test_e2e_xdc: bins
 cover_profile: bins
 	$Q mkdir -p $(BUILD)
 	$Q mkdir -p $(COVER_ROOT)
-	$Q echo "mode: atomic" > $(UNIT_COVER_FILE)
```
Groxx, line 554: is atomic default?  I think yes?
if not, I think it can be set via an arg for `go test`

Makefile
jakobht: I'm not sure I understand - it's almost the same command as before, just for all test instead of single tests?

Makefile
jakobht: Yeah, i checked that the two generated cover files are the same, they are. So the `mode: atomic` is output by go test

Makefile
jakobht: Also i made sure the generated cover profile is the same as before

jakobht:
> since `$(PKG_TEST_DIRS)` isn't all dirs: do all these run locally? skipping is fine (e.g. if postgres isn't running), but "build and pass normally" is definitely worth maintaining.
> 
> I think the answer is "yes" (possibly except `cmd/server/cadence` which may now run 2x) and we definitely _want_ it to be "yes" and I definitely vote "yes" to this change (I've wanted to do it for a while now). Last I poked at it I think there was still a (probably small) issue or two though.

Yes `make test` and `make cover_profile` runs locally and much faster as go now has a change to cache, and they can run in parallel (15min - 2min)

Makefile
```diff
@@ -551,16 +549,11 @@ test_e2e_xdc: bins
...
 
 	$Q echo Running special test cases without race detector:
 	$Q go test -v ./cmd/server/cadence/
 	$Q echo Running package tests:
-	$Q for dir in $(PKG_TEST_DIRS); do \
-		mkdir -p $(BUILD)/"$$dir"; \
-		go test "$$dir" $(TEST_ARG) -coverprofile=$(BUILD)/"$$dir"/coverage.out || exit 1; \
-		cat $(BUILD)/"$$dir"/coverage.out | grep -v "^mode: \w\+" >> $(UNIT_COVER_FILE); \
-	done;
+	$Q go test -short ./... $(TEST_ARG) -coverprofile=$(UNIT_COVER_FILE);
```
taylanisikdemir, line 556: `PKG_TEST_DIRS` is not used anymore that means all go tests including integration tests will be executed, right? We should keep `cover_profile` lightweight and not depend on C* or other db containers being available.
Looks like unit test in CI is failing. Once you make it pass please share coveralls link so we can compare before and after

Makefile
Groxx: aah, sorry, I was misreading.  yea this was just kinda weird wasn't it - we should probably change that to `$(UNIT_COVER_FILE)` or something more coverage-looking than a file named `test`...

anyway, unchanged is also good, so 👍

Makefile
Groxx: yeah, that's what was preventing me from doing this earlier.

_build_ all packages for `go test ./...`: yes absolutely we should.  integration probably need `t.Skip("explanation")` stuff tho, like we have for cassandra now.  It's just not universally applied iirc.

jakobht:
So status is that CI failed with three distinct reasons, and then passed, see the unit test build kite.

I have raised times for two tests already, so timing depended behavior stopped failing.

The third failure was caused by resently introduced fuzzing, I expect the last two are timing depended as well.

PR #5462, opened 2023-11-27
Added a new dynamic config option MaxWorkflowCacheSize

jakobht:
**What changed?**
Added new dynamic config option for controlling the size of the workflow cache.

**Why?**
It's important we can change the size of the workflow cache in the running system. So if it is using too much memory we can shrink it, and if its evicting too much we can grow it. 

**How did you test it?**
Tested locally with the file system client 

**Potential risks**
N/A

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1178,7 +1178,7 @@ const (
 	// Allowed filters: ShardID
 	ReplicationTaskProcessorErrorRetryMaxAttempts
 
-	// MaxBufferedReplicationTasks is the max number of workflows for which we store in memory information
+	// MaxWorkflowCacheSize is the max number of workflows for which we store in memory information
 	// Keyname: history.MaxWorkflowCacheSize
 	// Value type: Int
 	// Default value: 1000
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1178,6 +1178,13 @@ const (
 	// Allowed filters: ShardID
 	ReplicationTaskProcessorErrorRetryMaxAttempts
 
+	// MaxBufferedReplicationTasks is the max number of workflows for which we store in memory information
+	// Keyname: history.MaxWorkflowCacheSize
+	// Value type: Int
+	// Default value: 1000
+	// Allowed filters: N/A
+	MaxWorkflowCacheSize
+
 	// key for worker
 
 	// WorkerPersistenceMaxQPS is the max qps worker host can query DB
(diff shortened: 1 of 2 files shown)


common/dynamicconfig/constants.go
```diff
@@ -3456,6 +3463,11 @@ var IntKeys = map[IntKey]DynamicInt{
 		Description:  "ReplicationTaskProcessorErrorRetryMaxAttempts is the max retry attempts for applying replication tasks",
 		DefaultValue: 10,
 	},
+	MaxWorkflowCacheSize: DynamicInt{
+		KeyName:      "history.MaxWorkflowCacheSize",
+		Description:  "is the max number of workflows for which we store in memory information",
+		DefaultValue: 1000,
```
taylanisikdemir, line 3469: A few clarification questions:
1. where is this going to be used exactly and is it unlimited right now? 
2. Is this going to be per-domain or per-shard limit? 
3. what is the potential side effects of limiting to 1k?

common/dynamicconfig/constants.go
```diff
@@ -1178,6 +1178,13 @@ const (
 	// Allowed filters: ShardID
 	ReplicationTaskProcessorErrorRetryMaxAttempts
 
+	// MaxBufferedReplicationTasks is the max number of workflows for which we store in memory information
+	// Keyname: history.MaxWorkflowCacheSize
```
davidporter-id-au, line 1182: Is this cache used for Replication tasks only or is it used for other things? If it's replication only, may I suggest we add Replication into the key-name?

common/dynamicconfig/constants.go
```diff
@@ -1178,6 +1178,13 @@ const (
 	// Allowed filters: ShardID
 	ReplicationTaskProcessorErrorRetryMaxAttempts
 
+	// MaxBufferedReplicationTasks is the max number of workflows for which we store in memory information
```
davidporter-id-au, line 1181: Question: 
- Is this the hydration cache size? guess I'm not super sure which cache this is and it's relation to Vytautas' replication cache project.
- There was a (possibly separate/different) task to try and make the caches bound by Byte size, not workflow size (since some workflow executions are quite large). Is that something considered here?

common/dynamicconfig/constants.go
jakobht: sorry thats a copy paste error

This is the workflow rate limit cache size, fixed the name - sorry about the confusion

dkrotx:
I don't really understand why we should have this limit: the amount of workflows on specific shard in short period of time (we talk about a second) can't be huge memory-pressure. I bet it makes much more memory pressure to just handle that many workflows in a second cause handlers should have a larger footprint. Isn't it?
What if you'll just drop stats for some workflow since it is not a subject to ratelimit? (happened seconds ago). Is that an option?

jakobht:
Synced with Jan on zoom. We decided to see if we can avoid introducing this additional handle. As Jan writes the number of workflows in a very shot amount of time should never be able to cause too much memory pressure. We "just" need to make sure the state is cleaned up. 
Abandoning this PR.

PR #5463, opened 2023-11-27
change variable name from SecondsSinceEpoch into EventTimeMs

bowenxia:
**What changed?**
change variable name from SecondsSinceEpoch into EventTimeMs

**Why?**
Pinot team made a mistake that they wanted us to create this variable but right now they realized that this variable has a name conflict with one of their system key. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


taylanisikdemir (approved):
LGTM since Pinot is still an in-progress visibility store. Once declared GA we should be careful about such schema changes.

neil-xie (approved):

PR #5464, opened 2023-11-28
Added a function to evict all elements older than the cache TTL

jakobht:
**What changed?**
Added a function to evict all elements in the LRU cache older than the TTL

**Why?**
This will enable us to clean out the cache periodically. This is useful for the workflow specific rate limits, as we will only need to keep items which are accessed very frequently. 

We will set the TTL to a low value and periodically clear out the elements that are above this limit to save memory. 

This way there is no need to set and manage a specific cache size 

**How did you test it?**
Unit tests

**Potential risks**
No risk, just adding a function 

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -53,10 +53,6 @@ type Cache interface {
 
 	// Size returns the number of entries currently stored in the Cache
 	Size() int
-
-	// EvictItemsPastTimeToLive evicts all items in the cache which are expired
-	// This is not called automatically, but can be called periodically to evict expired items
-	EvictItemsPastTimeToLive()
 }
 
 // Options control the behavior of the cache
@@ -94,7 +90,7 @@ type Options struct {
 	// Should be used when it's important for memory that the expired items are evicted as soon as possible
 	// If not set expired items will be evicted when one of these happens
 	// - when the cache is full
-	// - when EvictItemsPastTimeToLive is called
+	// - when evictItemsPastTimeToLive is called
 	// - when the item is accessed
 	ActivelyEvict bool
 }
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -88,6 +88,15 @@ type Options struct {
 	// to control the max size in bytes of the cache
 	// It is required option if MaxCount is not provided
 	MaxSize uint64
+
+	// ActivelyEvict will evict items that has expired TTL at every operation in the cache
+	// This can be expensive if a lot of items expire at the same time
+	// Should be used when it's important for memory that the expired items are evicted as soon as possible
+	// If not set expired items will be evicted when one of these happens
+	// - when the cache is full
+	// - when EvictItemsPastTimeToLive is called
+	// - when the item is accessed
+	ActivelyEvict bool
 }
 
 // SimpleOptions provides options that can be used to configure SimpleCache
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -53,6 +53,10 @@ type Cache interface {
 
 	// Size returns the number of entries currently stored in the Cache
 	Size() int
+
+	// EvictItemsPastTimeToLive evicts all items in the cache which are expired
+	// This is not called automatically, but can be called periodically to evict expired items
+	EvictItemsPastTimeToLive()
 }
 
 // Options control the behavior of the cache
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -90,7 +90,6 @@ type Options struct {
 	// Should be used when it's important for memory that the expired items are evicted as soon as possible
 	// If not set expired items will be evicted when one of these happens
 	// - when the cache is full
-	// - when evictItemsPastTimeToLive is called
 	// - when the item is accessed
 	ActivelyEvict bool
 }
--- a/common/cache/lru.go
+++ b/common/cache/lru.go
@@ -174,9 +174,7 @@ func (c *lru) Get(key interface{}) interface{} {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
-	if c.activelyEvict {
-		c.evictExpiredItems()
-	}
+	c.evictExpiredItems()
 
 	element := c.byKey[key]
 	if element == nil {
@@ -227,9 +225,7 @@ func (c *lru) Delete(key interface{}) {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
-	if c.activelyEvict {
-		c.evictExpiredItems()
-	}
+	c.evictExpiredItems()
 
 	element := c.byKey[key]
 	if element != nil {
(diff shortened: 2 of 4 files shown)


common/cache/lru.go
```diff
@@ -239,15 +249,34 @@ func (c *lru) Release(key interface{}) {
...
 	defer c.mut.Unlock()
 
 	return len(c.byKey)
 }
 
+// EvictItemsPastTimeToLive evicts all items in the cache which are expired
+// This is not called automatically, but can be called periodically to evict expired items
+func (c *lru) EvictItemsPastTimeToLive() {
+	c.mut.Lock()
+	defer c.mut.Unlock()
```
taylanisikdemir, line 266: we can avoid multiple lock/unlocks that happens here and on the caller side after this func returns. you can omit locks here and make this a private func. 
nit: also renaming to `evictExpiredItems` sounds simpler.

common/cache/lru_test.go
```diff
@@ -384,3 +384,74 @@ func TestPanicOptionsIsNil(t *testing.T) {
 
 	New(nil)
 }
+
+func TestEvictItemsPastTimeToLive_AllExpired(t *testing.T) {
```
taylanisikdemir, line 388: we are going to slowly switch to table driven tests. please rewrite these new tests using that pattern https://golang.cafe/blog/golang-table-test-example

common/cache/cache.go
```diff
@@ -53,6 +53,10 @@ type Cache interface {
 
 	// Size returns the number of entries currently stored in the Cache
 	Size() int
+
+	// EvictItemsPastTimeToLive evicts all items in the cache which are expired
+	// This is not called automatically, but can be called periodically to evict expired items
+	EvictItemsPastTimeToLive()
```
taylanisikdemir, line 59: doesn't need to be exposed

common/cache/lru.go
jakobht: made evictExpiredItems private, then it makes sense to just move the locking for calls to this method to the callers (also done)

common/cache/cache.go
jakobht: Agree, this makes a lot of things simpler, thanks!

common/cache/lru.go
```diff
@@ -169,6 +171,10 @@ func (c *lru) Get(key interface{}) interface{} {
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
+	if c.activelyEvict {
```
3vilhamster, line 174: Instead of spreading the condition around, I suggest to call a function that checks this in one place.
```suggestion
		c.processEvicted()
```

And inside 
```
func (c *lru) processEvicted() {
   if c.activelyEvict {
      return
   }
```

common/cache/lru_test.go
```diff
@@ -384,3 +384,49 @@ func TestPanicOptionsIsNil(t *testing.T) {
...
+
+func TestEvictItemsPastTimeToLive_AllExpired_ActivelyEvict(t *testing.T) {
+	cache := New(&Options{
+		MaxCount:      5,
+		TTL:           time.Millisecond * 75,
+		ActivelyEvict: true,
+	})
+
+	_, err := cache.PutIfNotExist("A", t)
+	assert.NoError(t, err)
```
3vilhamster, line 396: this should be required.
My rationale for assert/require - assert only for the path that you test. Everything that is a test setup/error checks should be require, since there is no point of proceeding in this test if PutIfNotExist is broken.

common/cache/lru.go
```diff
@@ -242,16 +252,31 @@ func (c *lru) Size() int {
...
+		c.evictExpiredItems()
+	}
+
 	return len(c.byKey)
 }
 
+// evictExpiredItems evicts all items in the cache which are expired
+func (c *lru) evictExpiredItems() {
+	for elt := c.byAccess.Back(); len(c.byKey) > 0 && c.isEntryExpired(elt.Value.(*entryImpl), time.Now()); elt = c.byAccess.Back() {
```
3vilhamster, line 264: Maybe it is a bigger change, but I suggest switching from direct time.Now to either stub function now() that returns time.Now that you can mimic, or switching to use of github.com/benbjohnson/clock. That will allow us to not write time related tests and avoid using time.Sleep in tests.

common/cache/lru_test.go
taylanisikdemir: I would say table tests work very well compared to the other approaches. Some benefits are
- Maintainability: For the same unit of work you write the body once and you can easily add more cases without making redundant copy of the validation logic. If you need to change something in your test body you have to change it in N different places without table driven tests.
- Readability: Creating structs for test cases makes it very clear to understand what are the inputs and expected outputs. Much better readability.

Scope of your test function and what you put in the test body can of course increase complexity but that can and does happen in other approaches. [Style guide](https://github.com/uber-go/guide/blob/master/style.md#avoid-unnecessary-complexity-in-table-tests) has a good point about not combining irrelevant cases together.

3vilhamster: From my experience:
Table test works great for a set of similar cases that can be easily defined.
But every complication requires modification of initial struct and very fast it becomes hard to manage.
Lru is pretty complicated and codiying steps will be harder than just expressing what you are trying. At least the way I see how it could be codified.

taylanisikdemir: To be clear there's no suggestion to fit all of the lru_test.go into single tabular function to simulate all the scenarios.
Different scenarios deserve their own functions which simulate something we are interested to validate.

Regarding the new functions in this PR, all lines are duplicate except ~1-2 line difference (either an extra sleep or calling EvictItemsPastTimeToLive vs Size). This can be turned into table format easily.

common/cache/lru.go
```diff
@@ -242,28 +251,43 @@ func (c *lru) Size() int {
...
 	return len(c.byKey)
 }
 
+// evictExpiredItems evicts all items in the cache which are expired
+func (c *lru) evictExpiredItems() {
+	if !c.activelyEvict {
+		return // do nothing if activelyEvict is not set
+	}
+
+	for elt := c.byAccess.Back(); len(c.byKey) > 0 && c.isEntryExpired(elt.Value.(*entryImpl), c.now()); elt = c.byAccess.Back() {
```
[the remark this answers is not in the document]
jakobht: It works as expected, the test passes :) 

I have made the check a bit more explicit:
ByAccess list is sorted by access time, so we can evict expired items from the back `.Back` returns nil if the list is empty, so this loop will terminate when the list is empty or the first non-expired item is encountered

common/cache/lru.go
```diff
@@ -242,28 +251,43 @@ func (c *lru) Size() int {
...
 	return len(c.byKey)
 }
 
+// evictExpiredItems evicts all items in the cache which are expired
+func (c *lru) evictExpiredItems() {
+	if !c.activelyEvict {
+		return // do nothing if activelyEvict is not set
+	}
+
+	for elt := c.byAccess.Back(); elt != nil && c.isEntryExpired(elt.Value.(*entryImpl), c.now()); elt = c.byAccess.Back() {
```
3vilhamster, line 265: I still propose to simplify this for a reader:
```suggestion
	for elt := c.byAccess.Back(); elt != nil; elt = c.byAccess.Back() {
	if !c.isEntryExpired(elt.Value.(*entryImpl), c.now()) {
	    // List is sorted by item age, so we can stop as soon as we found first non expired item.
	    break
	}
```

jakobht: Sure, moved it around

common/cache/lru.go
```diff
@@ -242,28 +251,47 @@ func (c *lru) Size() int {
...
 }
 
+// evictExpiredItems evicts all items in the cache which are expired
+func (c *lru) evictExpiredItems() {
+	if !c.activelyEvict {
+		return // do nothing if activelyEvict is not set
+	}
+
+	for elt := c.byAccess.Back(); elt != nil; elt = c.byAccess.Back() {
+		if !c.isEntryExpired(elt.Value.(*entryImpl), c.now()) {
```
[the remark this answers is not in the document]
3vilhamster: time.Now required a syscall, not sure about now. But it is not free. https://github.com/golang/go/issues/57749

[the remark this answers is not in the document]
jakobht: I don't think it's expensive enough to worry about, but why not? Changed it

PR #5465, opened 2023-12-05
Extract JWT token group check for access

mantas-sidlauskas:
**What changed?**
Refactoring `oauthAuthorizer` and error messages

**Why?**
To reuse validation logic in other OAuth/OIDC implementations

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/authorization/authorizer.go
```diff
@@ -103,3 +106,28 @@ func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvide
...
+	switch attributes.Permission {
+	case PermissionRead:
+		groups = data[common.DomainDataKeyForReadGroups] + groupSeparator + data[common.DomainDataKeyForWriteGroups]
+	case PermissionWrite:
+		groups = data[common.DomainDataKeyForWriteGroups]
+	default:
+		return fmt.Errorf("token doesn't have permission for %v API", attributes.Permission)
+	}
+	// groups are separated by space
+	allowedGroups := strings.Split(groups, groupSeparator)    // groups that allowed by domain configuration(in domainData)
```
taylanisikdemir, line 121: if we are going to split the `groups` string here then I don't get why we use string concatenation to combine the groups. Better to keep a `var allowedGroups map[string]bool`

common/authorization/authorizer.go
```diff
@@ -103,3 +106,28 @@ func GetAuthProviderClient(privateKey string) (clientworker.AuthorizationProvide
 type FilteredRequestBody interface {
 	SerializeForLogging() (string, error)
 }
+
+func validatePermission(claims *JWTClaims, attributes *Attributes, data map[string]string) error {
+	groups := ""
+	switch attributes.Permission {
```
taylanisikdemir, line 112: does existing unit tests cover all cases?

common/authorization/authorizer.go
mantas-sidlauskas: changed code here. Added utility functions to deal with intersections a bit easier.

common/authorization/authorizer.go
mantas-sidlauskas: this part was not covered directly, added dedicated unit tests to this function

common/authorization/oauthAuthorizer.go
```diff
@@ -137,34 +140,10 @@ func (a *oauthAuthority) parseToken(tokenStr string, verifier jwt.Verifier) (*JW
 
 func (a *oauthAuthority) validateTTL(claims *JWTClaims) error {
 	if claims.TTL > a.authorizationCfg.MaxJwtTTL {
-		return fmt.Errorf("TTL in token is larger than MaxTTL allowed")
+		return fmt.Errorf("token TTL: %d is larger than MaxTTL allowed: %d", claims.TTL, a.authorizationCfg.MaxJwtTTL)
 	}
 	if claims.Iat+claims.TTL < time.Now().Unix() {
-		return fmt.Errorf("JWT has expired")
+		return errors.New("JWT has expired")
```
davidporter-id-au, line 146: Remark: No need to do anything

We probably need to pull these out as distinct error types such that folks can match against them rather than string matching. Anyway, a change for another PR

davidporter-id-au (approved):

davidporter-id-au (approved):

common/authorization/authorizer.go
```diff
@@ -86,13 +89,21 @@ func NewPermission(permission string) Permission {
 	}
 }
 
+func (d domainData) Groups(groupType string) []string {
```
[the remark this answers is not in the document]
mantas-sidlauskas: I'm not sure if moving this method to elsewhere would add any benefit. "Groups" are not used elsewhere except in this, authorization related package. Maybe "domainData" name is a bit misleading, this can be named as "groupData".

On the other hand, if there's a need to add first-class "domain groups" support, then yeah, this should be moved to a dedicated, group management place. But for now, you have to use domain free-form metadata functionality to describe read/write groups and moving helper method out of here doesn't add any benefit IMHO.

common/authorization/oauthAuthorizer.go
mantas-sidlauskas: I was thinking about that, but the default behaviour now is to return `DecisionDeny` without any explanation. You can see this error in server logs with logging level set to `debug`

common/authorization/authorizer.go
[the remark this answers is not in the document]
davidporter-id-au: yeah, sorry, that was a discussion I was having internally. No concerns.

PR #5466, opened 2023-12-05
Add metrics to monitor task validation.

agautam478:
**What changed?**
- Introduced the metric monitoring to the taskvalidator service.
- We are counting the number of workflows that enter the service for validation. This will give us a perspective of how many workflows are suspected to be corrupted when going through it's respective processing.
- Currently we are only monitoring the workflows from history update where the `maximum attempts exceeded to update history`.
- The plan is to roll it out for other task processing flows as well.

**Why?**
Made these changes to get a sense of how many workflows are assumed to be corrupted.

**How did you test it?**
tested locally.
Wrote a unit test.

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


ketsiambaku (approved):
LGTM

common/taskvalidator/validateworkflow.go
```diff
@@ -36,21 +37,25 @@ type Checker interface {
...
-	w.logger.Info(fmt.Sprintf("WorkflowCheckforValidation. DomainID: %v, WorkflowID: %v, RunID: %v",
-		domainID,
-		workflowID,
-		runID))
+	w.logger.Info("WorkflowCheckforValidation",
+		tag.WorkflowID(workflowID),
+		tag.WorkflowRunID(runID),
+		tag.WorkflowDomainID(domainID))
+	// Emit the number of workflows that have come in for the validation.
+	w.metricsClient.Scope(metrics.TaskValidatorScope).IncCounter(metrics.ValidatedWorkflowCount)
```
taylanisikdemir, line 59: which tags will this metric have? can you share example from local run?

PR #5468, opened 2023-12-06
Minor change to include domainTag and pass domainName.

agautam478:
**What changed?**
Added domainTag to the validationwf metric and passed domainName to the method for future development ease.

**Why?**
Made these changes to increase observability and prepare for future code changes.

**How did you test it?**
locally.
make test

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


service/history/handler.go
```diff
@@ -26,13 +26,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/uber/cadence/service/history/workflow"
```
taylanisikdemir, line 29: move below. builtin imports should be separated

service/history/handler.go
```diff
@@ -2175,7 +2174,11 @@ func (h *handlerImpl) error(
 		// We will delete the workflow or mark the workflow as corrupted.
 		// Placing a dummy call to the function to check the coherency of the design.
 		// The function returns nil error so removing error handling for now.
-		h.GetTaskValidator().WorkflowCheckforValidation(workflowID, domainID, "")
+		domainName, err := h.GetDomainCache().GetDomainName(domainID)
+		if err != nil {
+			domainName = ""
```
taylanisikdemir, line 2179: It should be rare to have a domain retrieval from cache fail. Ignoring the error and moving on with empty domain name is going to introduce yet another caveat to the system. 
I'd return the `err` here instead of moving on with undefined domain name. In general, "fail fast" approach helps eliminate bugs by eliminating bunch of undefined behavior

taylanisikdemir (approved):

3vilhamster (approved):

PR #5470, opened 2023-12-06
Add an "all results" query to scanner/fixer workflows

Groxx:
Currently, getting all output filenames from these workflows is an exercise in frustration.

You can:
- query `shard_corrupt_keys` to get all shards with corruptions (no data on fails, etc)
- query `shard_report` to get a *single* shard's corruptions, errors, skips, control-flow failures
- browse activity results by hand to discover ^ this in bulk

But unfortunately:
- metrics do not contain per-shard info so finding the relevant activity or shard is hard
- there are essentially no logs in this entire system (!?!)
- there is currently no query to get both failures and corruptions/fixes in bulk
- if one invariant reports "fixed" and then the next returns "fail" because the fix removed data,
  the end result goes into "failures".  this is true for scans too, corrupt + fail == fail.

Many small bits of friction make trying to bulk-analyze this system *incredibly* painful.

While we do need to just rewrite the whole thing to be less... like it is... we can at least expose this bulk info quite easily in a new query.

### DIFF
--- a/service/worker/scanner/shardscanner/aggregators.go
+++ b/service/worker/scanner/shardscanner/aggregators.go
@@ -557,6 +557,14 @@ func (a *ShardScanResultAggregator) adjustAggregation(stats ScanStats, fn func(a
 	}
 }
 
+func (a *ShardScanResultAggregator) GetAllScanResults() (map[int]ScanResult, error) {
+	result := make(map[int]ScanResult, len(a.reports))
+	for k, v := range a.reports {
+		result[k] = v.Result
+	}
+	return result, nil
+}
+
 func getStatusResult(
 	minShardID int,
 	maxShardID int,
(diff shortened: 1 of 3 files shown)


neil-xie (approved):

service/worker/scanner/shardscanner/aggregators.go
```diff
@@ -330,6 +330,17 @@ func (a *ShardFixResultAggregator) GetReport(shardID int) (*FixReport, error) {
 	return nil, fmt.Errorf("shard %v has not finished yet, check back later for report", shardID)
 }
 
+func (a *ShardFixResultAggregator) GetAllFixResults() (map[int]FixResult, error) {
+	result := make(map[int]FixResult, len(a.reports))
+	for k, v := range a.reports {
+		if v.Result.Empty() {
+			continue
+		}
```
Groxx, line 338: all shards have reports, so in the normal / healthy case this was producing a pile of 
```
"1": {
  "ShardFixKeys": {
    "Fixed": null,
    "Skipped": null,
    "Failed": null,
  }
  "ControlFlowFailure": null,
}
```
which is a lot of unnecessary output.

a possibly-nicer way to do this would be to use json tags and mark these as `omitempty`, but I'm not sure what things might be depending on the exact serialized structure at the moment, so I'm avoiding that.

PR #5471, opened 2023-12-07
Add retries into Scanner BlobWriter

agautam478:
**What changed?**
The scanner environment consistently encounters issues with broken shards, which remain unresolved and likely contain erroneous data. This issue is highlighted by the `cadence_shard_failure metric`, which not only signals these specific shard failures but also other scan-related issues.

A primary contributor to these failures, among various factors, is the malfunctioning of the Blobwriter during the process of writing scan outputs into the database. The root causes for this malfunction could be varied, but a notable and addressable concern is the writer's inability to handle transient errors or network issues effectively.

To mitigate this, we are implementing a retry mechanism in the writer. 

- Inside the getBlobstoreWriteFn, we added a loop that attempts the write operation up to MaxRetries times.
- After each failed attempt (if an error occurs), we wait for RetryDelay before retrying.
- If all attempts fail, we return the last error encountered.

**Why?**
This enhancement aims to offer the writer additional opportunities to successfully upload data, especially in scenarios where failures are due to temporary network disruptions or similar transient issues. By allowing retries, we can potentially reduce the frequency of these scan failures and improve the overall stability of the system.

**How did you test it?**
Tested locally.

**Potential risks**
might cause delay in the blobstore uploads.

**Release notes**
NA

**Documentation Changes**
NA


common/reconciliation/store/blobstoreWriter.go
```diff
@@ -106,12 +114,33 @@ func getBlobstoreWriteFn(
...
 
-		ctx, cancel := context.WithTimeout(context.Background(), Timeout)
-		defer cancel()
-		if _, err := client.Put(ctx, req); err != nil {
-			return nil, err
+		var lastErr error
+		retryDelay := InitialRetryDelay
+		// The idea is to implement a loop that retries the write operation a certain number of times before finally failing.
+		// We'll also include a delay between retries to give transient issues (like temporary network glitches) a chance to resolve.
+		for attempt := 0; attempt < MaxRetries; attempt++ {
```
taylanisikdemir, line 121: don't we have a retry library for this?

taylanisikdemir: if not we can try [this one](https://github.com/cenkalti/backoff). I used it before without any issues

agautam478: Not that I know of. Will try it. Thanks for the suggestion.

Groxx: we have one or two backoff-things through the code, it's used very heavily internally.

service/worker/scanner/shardscanner/scanner.go
```diff
@@ -162,6 +162,9 @@ func (s *ShardScanner) Scan(ctx context.Context) ScanReport {
 				Execution: execution,
 				Result:    checkResult,
 			}); err != nil {
+				// This looks like one of the contributors of the high number of skipped_scans in the system.
+				// Upon closer investigation we discovered that there is no retry mechanism in the Blobstore write function, which might
+				// result in unnecessary skip of some scans. Especially, the ones that might have happened due to network blip.
```
Groxx, line 167: probably worth removing this, as it's (at least partially) no longer true after this merges.

useful as a commit message / review context, but not really for long-term purposes

common/reconciliation/store/blobstoreWriter.go
```diff
@@ -106,9 +114,35 @@ func getBlobstoreWriteFn(
...
+		throttlePolicy.SetMaximumInterval(MaxRetryDelay)
+		throttlePolicy.SetExpirationInterval(Timeout)
+
+		throttleRetry := backoff.NewThrottleRetry(
+			backoff.WithRetryPolicy(retryPolicy),
+			backoff.WithThrottlePolicy(throttlePolicy),
+			backoff.WithRetryableError(func(err error) bool {
+				return true // assuming all errors are retryable
+			}),
+		)
```
Groxx, line 141: general pattern elsewhere worth following: inject the policies.  they're not mutated, so you can reuse them.
this also lets you modify them for tests, to fake time and whatnot.

otherwise, while ThrottleRetry does seem like a better option, I'm not sure it's relevant here.  and if it is, it probably deserves a different policy: 
- it only changes behavior on `*types.ServiceBusyError`, which is probably not being returned by the blobstore.  or is it?
- the throttle-policy only advances when a throttle response is returned, using the same for both will mean it just always uses the non-throttle policy because it's always going to be the same delay or greater.

common/reconciliation/store/blobstoreWriter.go
```diff
@@ -27,11 +27,19 @@ import (
...
 	"github.com/uber/cadence/common/blobstore"
 	"github.com/uber/cadence/common/pagination"
 )
 
+const (
+	MaxRetries        = 3
+	InitialRetryDelay = 2 * time.Second  // Initial delay between retries
+	MaxRetryDelay     = 30 * time.Second // Maximum delay between retries
+)
+
```
Groxx, line 42: const: 👍 
public: nothing else needs these, just leave them private.

common/reconciliation/store/blobstoreWriter.go
agautam478: I thought using throttling will be future proof in case we discover some error types that do not require retries so left it there. Should I remove it?

common/reconciliation/store/blobstoreWriter.go
```diff
@@ -106,9 +114,35 @@ func getBlobstoreWriteFn(
 			},
 		}
 
-		ctx, cancel := context.WithTimeout(context.Background(), Timeout)
-		defer cancel()
-		if _, err := client.Put(ctx, req); err != nil {
+		operation := func() error {
+			ctx, cancel := context.WithTimeout(context.Background(), Timeout)
+			defer cancel()
+			_, err := client.Put(ctx, req)
```
davidporter-id-au, line 120: do we have unit test on this?

agautam478: yes! added one today.

common/reconciliation/store/blobstoreWriter.go
Groxx: non-retryable errors aren't throttle-retryable either, so in that sense: no.
throttle-retry in general: yeah, it's a bit more future-proof I think.  basically fine to leave it.

it's a bit more future-proof in that returning a ServiceBusy error is more likely to be handled reasonably / more easily adjusted / etc.  with things at the same value it doesn't make a difference, but it can at least be centrally adjusted, and e.g. one of the things we'd like to change is to allow busy-errors to define a minimum duration to wait before retrying -> that'd more likely get inherited here, which is good.

common/reconciliation/store/blobstoreWriter.go
```diff
@@ -47,9 +55,19 @@ func NewBlobstoreWriter(
 	client blobstore.Client,
 	flushThreshold int,
 ) ExecutionWriter {
+	retryPolicy := backoff.NewExponentialRetryPolicy(initialRetryDelay)
+	retryPolicy.SetMaximumInterval(maxRetryDelay)
+	retryPolicy.SetExpirationInterval(Timeout)
```
Groxx, line 60: just noticed: do we actually want `Timeout` here?  the way it's set up now, a single request-timeout will fail the whole thing rather than retry, because it'll have exceeded the expiration-interval already.

agautam478: ahh yeah! probably need to have a new total retry duration that's longer that timeout.
kinda like

`totalRetryDuration := Timeout * 2` ?

wdyt

neil-xie (approved):

service/history/handler_test.go
```diff
@@ -23,13 +23,15 @@ package history
...
-	"go.uber.org/yarpc/yarpcerrors"
 	"math/rand"
 	"sync/atomic"
 	"testing"
 
+	"github.com/stretchr/testify/mock"
+	"go.uber.org/yarpc/yarpcerrors"
+
+	"github.com/uber/cadence/common/metrics/mocks"
+
```
taylanisikdemir, line 34: inconsistent new lines in the import section. One common practice that I followed in the past is to group imports like this:
- builtin imports (e.g. fmt)
- local imports (stuff from current repo)
- external imports (vendor libraries)

There would be new line between each group.
I know we don't follow everywhere this but it can help organize imports when unsure.

taylanisikdemir (approved):

Groxx (approved):

PR #5472, opened 2023-12-07
Added a unit test for the BlobStoreWriter.

agautam478:
**What changed?**
Added a test for the BlobstoreWriter since it didn't have any.

**Why?**
It's an attempt at improving our code coverage and introduce better testing practices. Tried the table test for the first time here. This is something that will simplify tedious and repetitive unit tests in Go (Golang) when testing functions with multiple scenarios and data. 

**How did you test it?**
Yes ran locally.

**Potential risks**
NA 

**Release notes**
NA

**Documentation Changes**
NA


common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,95 @@
...
+	*require.Assertions
+	blobstoreWriter *blobstoreWriter
+	uuid            string
+	extension       Extension
+	outputDir       string
+	blobstoreClient blobstore.Client
+}
+
+func TestBlobstoreWriterSuite(t *testing.T) {
+	suite.Run(t, new(BlobstoreWriterSuite))
```
taylanisikdemir, line 49: we are trying to avoid suites where possible and I believe we don't need it here.

agautam478: should I mock the client and write a standard unit test?

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,99 @@
...
+	outputDir := t.TempDir()
+
+	cfg := &config.FileBlobstore{
+		OutputDirectory: outputDir,
+	}
+	//Reusing the Filestoreclient from the other sister test in the same package.
+	blobstoreClient, err := filestore.NewFilestoreClient(cfg)
+	require.NoError(t, err)
+
+	blobstoreWriter := NewBlobstoreWriter(uuid, extension, blobstoreClient, 10).(*blobstoreWriter)
```
taylanisikdemir, line 50: move these inside the test body at line 69 so each test case runs isolated from each other

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Add data to the writer
+			err := blobstoreWriter.Add(tc.input)
+			if tc.expectedErr {
+				assert.Error(t, err)
+			} else {
```
taylanisikdemir, line 73: avoid extra indentation by handling special case and returning.
```
if tc.expectedErr {
  assert.Error(t, err)
  return
}

// rest of the code goes here without an else { } wrapper
```

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,99 @@
...
+				assert.NoError(t, err)
+
+				// Flush the writer to write data to the blobstore
+				err = blobstoreWriter.Flush()
+				assert.NoError(t, err)
+
+				// Retrieve the keys of flushed data
+				flushedKeys := blobstoreWriter.FlushedKeys()
+				assert.NotNil(t, flushedKeys)
```
taylanisikdemir, line 82: do these assert calls include a message to pinpoint where it failed? more simple way to write assertions in native go is
```
if flushedKeys == nil {
  t.Error("Expected flushedKeys to be not nil")
}
```

what does assert.NotNil show in test output when `flushedKeys` is nil?

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func TestBlobstoreWriter(t *testing.T) {
+	type testCase struct {
+		name        string
+		input       string
+		expectedErr bool
+	}
+
+	//Trying a table test approach this time.
```
taylanisikdemir, line 45: nit: can remove this comment

taylanisikdemir (approved):

davidporter-id-au (approved):

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,99 @@
...
+			req := &blobstore.GetRequest{Key: key}
+			ctx := context.Background()
+			resp, err := blobstoreClient.Get(ctx, req)
+			assert.NoError(t, err)
+
+			// Verify the contents
+			var result string
+			err = json.Unmarshal(resp.Blob.Body, &result)
+			require.NoError(t, err)
+			assert.Equal(t, tc.input, result)
```
Groxx, line 96: hmmm.  tbh this feels kinda misleading to me - this same Unmarshal won't work if you .Add twice because of how it's serializing in `getBlobstoreWriteFn(...)`

to be more accurate it'd probably need to use blobstoreiterator to deserialize.  though that'd require using a `reconciliation/entity.Entity` which does a fair amount of irrelevant things here...

---

since using an `Entity` seems mostly unrelated and overkill, maybe something like this would make it clear that the format is decided by the writer, and it's not intended to be valid JSON on its own?
```go
s // add 2 things since that's common in practice + makes it no longer a single valid json thing
blobstoreWriter.Add("one")
blobstoreWriter.Add("two")

// verify the bytes are as expected
assert.Equal(t, resp.Blob.Body, []byte(`"one"\r\n"two"`)
```

Groxx (changes requested):
Minor assert-vs-request-vs-t.Error fuzziness to clean up and I do think it's worth not implying "produces valid json" (because this writer does not).  but otherwise looks good, should be an easy 👍

common/reconciliation/store/blobstorewriter_test.go
Groxx: `assert` and `t.*` calls do print the line and a partial trace that failed, yeah.

common/reconciliation/store/blobstorewriter_test.go
agautam478: this one was failing with some weird unmarshalling error so I did this.

common/reconciliation/store/blobstorewriter_test.go
Groxx: this would be an ideal case for `require.NoError(t, err)` fwiw.  that'll stop the test immediately.

(this is also one of the reasons why embedding Assertions is problematic.  it hides whether things are asserts or requires, and that's a _VERY_ important detail for this reason and for concurrency-safety reasons (you cannot safely `require` on a background thread))

common/reconciliation/store/blobstorewriter_test.go
Groxx: that'd depend on what you're comparing / where you got those bytes.  the `0x22` one looks like what I expect in body though

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+			// Flush the writer to write data to the blobstore
+			err = blobstoreWriter.Flush()
+			assert.NoError(t, err)
+
+			// Retrieve the keys of flushed data
+			flushedKeys := blobstoreWriter.FlushedKeys()
+			if flushedKeys == nil {
+				t.Error("Expected flushedKeys to be not nil")
+			}
```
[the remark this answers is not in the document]
taylanisikdemir: Not only this line but all the validations could just be written using `t.Error/t.Fatal` which forces us to write explicit error message.  On the other hand the message is optional in assert/require packages. e.g. 
```
// require.NotNil signature
NotNil(t TestingT, object interface{}, msgAndArgs ...interface{})
```

Regarding the termination when `flushedKeys` is nil, yes let's fix that by changing this to `t.Fatal` instead of `t.Error`

common/reconciliation/store/blobstorewriter_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			// Read back the data from the blobstore
+			key := pageNumberToKey(uuid, extension, flushedKeys.MinPage)
+			req := &blobstore.GetRequest{Key: key}
+			ctx := context.Background()
+			resp, err := blobstoreClient.Get(ctx, req)
+			assert.NoError(t, err)
+
+			// Verify the contents
+			assert.Equal(t, string(resp.Blob.Body), "\"one\"\r\n\"two\"\r\n")
```
Groxx, line 95: for a table-test you'd probably put this in the table (as it's likely to change with each test), but eh.  the test is reasonably simple, will be easy to adjust when we change it for future stuff.

Groxx (approved):

PR #5473, opened 2023-12-08
Update kafka config to have isSecure option

neil-xie:
**What changed?**
Update kafka config to have new isSecure field

**Why?**
serves for the secure kafka topic onboarding in monorepo

**How did you test it?**
Integration test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/config/kafkaConfig.go
+++ b/common/config/kafkaConfig.go
@@ -43,7 +43,8 @@ type (
 
 	// TopicConfig describes the mapping from topic to Kafka cluster
 	TopicConfig struct {
-		Cluster string `yaml:"cluster"`
+		Cluster  string `yaml:"cluster"`
+		IsSecure *bool  `yaml:"isSecure,omitempty"`
 	}
 
 	// TopicList describes the topic names for each cluster
@@ -98,4 +99,12 @@ func (k *KafkaConfig) GetBrokersForKafkaCluster(kafkaCluster string) []string {
 // GetTopicsForApplication gets topic from application
 func (k *KafkaConfig) GetTopicsForApplication(app string) TopicList {
 	return k.Applications[app]
+}
+
+// GetKafkaClusterSecureConfigForTopic gets isSecure status from topic
+func (k *KafkaConfig) GetKafkaSecureConfigForTopic(topic string) bool {
+	if k.Topics[topic].IsSecure != nil {
+		return *k.Topics[topic].IsSecure
+	}
+	return false
 }


common/config/kafkaConfig.go
```diff
@@ -43,7 +43,8 @@ type (
 
 	// TopicConfig describes the mapping from topic to Kafka cluster
 	TopicConfig struct {
-		Cluster string `yaml:"cluster"`
+		Cluster  string `yaml:"cluster"`
+		IsSecure *bool  `yaml:"isSecure,omitempty"`
```
Shaddoll, line 47: We don't need to make it a pointer, since it's false by default. Personally, I think it's annoying to deal with pointers.

taylanisikdemir: bool pointers are useful if you need to differentiate between unset or not. Otherwise just two regular bools with 2 state is preferred.

common/config/kafkaConfig.go
```diff
@@ -43,7 +43,8 @@ type (
 
 	// TopicConfig describes the mapping from topic to Kafka cluster
 	TopicConfig struct {
-		Cluster string `yaml:"cluster"`
+		Cluster  string `yaml:"cluster"`
+		IsSecure bool   `yaml:"isSecure,omitempty"`
```
taylanisikdemir, line 47: can you add comment on this field to explain what being secure means and what should user expect when it's set to true

neil-xie: Sure, added comment. This will only affect our internal usage, OSS is using sarama producer which has been deprecated internally

common/config/kafkaConfig.go
```diff
@@ -99,3 +102,8 @@ func (k *KafkaConfig) GetBrokersForKafkaCluster(kafkaCluster string) []string {
 func (k *KafkaConfig) GetTopicsForApplication(app string) TopicList {
 	return k.Applications[app]
 }
+
+// GetKafkaClusterSecureConfigForTopic gets isSecure status from topic
+func (k *KafkaConfig) GetKafkaSecureConfigForTopic(topic string) bool {
+	return k.Topics[topic].IsSecure
```
Groxx, line 108: huh.  is this a general kafka-topic flag, or is it something entirely internal?  may be worth making a more general config-map or something if it's internal.

neil-xie: It is entirely internal change. New topics are enforced to be secure. Config maps means converts it into a map[string]string?

taylanisikdemir: +1 for not exposing such internal only fields to avoid bloating the public contract. 
`map[string]any Properties` is a good idea

neil-xie: Updated it to use map in https://github.com/uber/cadence/pull/5474

PR #5474, opened 2023-12-11
Wrap isSecure config in config map for kafka topic

neil-xie:
**What changed?**
Unexpose isSecure config, replace with config map

**Why?**
IsSecure is internal config for kafka topic, we don't want it to be exposed for public users

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/config/kafkaConfig.go
+++ b/common/config/kafkaConfig.go
@@ -45,7 +45,7 @@ type (
 	TopicConfig struct {
 		Cluster string `yaml:"cluster"`
 		// Properties map describes whether the topic properties, such as whether it is secure
-		Properties map[string]interface{} `yaml:"properties,omitempty"`
+		Properties map[string]any `yaml:"properties,omitempty"`
 	}
 
 	// TopicList describes the topic names for each cluster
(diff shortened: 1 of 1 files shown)


common/config/kafkaConfig.go
```diff
@@ -44,9 +44,8 @@ type (
 	// TopicConfig describes the mapping from topic to Kafka cluster
 	TopicConfig struct {
 		Cluster string `yaml:"cluster"`
-		// IsSecure describes whether the topic is secure, by default it is false
-		// If it is set to true, it allows to pass in the token to initialize secure producer for this topic
-		IsSecure bool `yaml:"isSecure,omitempty"`
+		// Properties map describes whether the topic properties, such as whether it is secure
+		Properties map[string]interface{} `yaml:"properties,omitempty"`
```
taylanisikdemir, line 48: we can use `any` instead of `interface{}`

common/config/kafkaConfig.go
```diff
@@ -103,7 +102,12 @@ func (k *KafkaConfig) GetTopicsForApplication(app string) TopicList {
...
 
-// GetKafkaClusterSecureConfigForTopic gets isSecure status from topic
-func (k *KafkaConfig) GetKafkaSecureConfigForTopic(topic string) bool {
-	return k.Topics[topic].IsSecure
+// GetKafkaPropertiesForTopic gets properties from topic
+func (k *KafkaConfig) GetKafkaPropertiesForTopic(topic string) map[string]interface{} {
+	topicConfig, ok := k.Topics[topic]
+	if !ok || topicConfig.Properties == nil {
+		// No properties for the specified topic in the config
+		return nil
```
taylanisikdemir, line 110: nit: it might be better to return an empty map to avoid nil checks on the caller side. or instead of `GetKafkaPropertiesForTopic(topic string)` we can expose `GetKafkaPropertyForTopic(topic, propName string)`

taylanisikdemir (approved):

common/config/kafkaConfig.go
neil-xie: Updated, good to know it!

common/config/kafkaConfig.go
neil-xie: Yeah, updated to return the value for a specific property. It is probably not necessary to return the whole map.

davidporter-id-au (approved):

Groxx (approved):
yea, probably worth the `get(topic, prop)` API for unpacking that structure, which seems reasonable.

maybe one day we'll have a more standardized "externally-extendable config" thing.  but this seems fine for now, and it's constrained enough that changing it later should be fairly safe / straight-forward for anyone using it.

PR #5477, opened 2023-12-13
Address map access data race in matching engine

taylanisikdemir:
**What changed?**
A PR check test was [failing](https://buildkite.com/uberopensource/cadence-server/builds/16701#018c60e8-4179-4169-84a4-055966f99e4c) due to unprotected access to `taskLists` by multiple goroutines. Handling it by deferring unlock which is generally safer way to use mutex.

Misc changes:
- Get rid of inconsistent/unnecessary usage of multilines. Hurts vertical readability speed.

**Why?**
Avoid data race. This will also decrease flakiness of the tests.

### DIFF
--- a/service/matching/matchingEngine.go
+++ b/service/matching/matchingEngine.go
@@ -164,10 +164,10 @@ func (e *matchingEngineImpl) Stop() {
 	}
 }
 
-func (e *matchingEngineImpl) getTaskLists(maxCount int) (lists []taskListManager) {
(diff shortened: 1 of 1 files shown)


Shaddoll (approved):

taylanisikdemir:
> Good catch. I wonder how we did not see it in prod.

This happens during tasklist stop flow so not likely to hit or detect in prod.

PR #5479, opened 2023-12-15
fix docker compose tests

3vilhamster:
**What changed?**
Added health checks for CI runners with cassandra using healthcheck and depends_on: service_healthy.

Also, removed some unnecessary rebuilds/regenerations that were performed in the tests.

Also, removed extra steps from unit test executions and made it possible to run docker-compose like on CI locally.

**Why?**
Fixing test failures on CI

**How did you test it?**
Run locally

**Potential risks**
Failing builds on CI

**Release notes**

**Documentation Changes**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -548,7 +548,7 @@ test_e2e_xdc: bins
 		go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log; \
 	done;
 
-cover_profile: bins
+cover_profile:
 	$Q mkdir -p $(BUILD)
 	$Q mkdir -p $(COVER_ROOT)
 	$Q echo "mode: atomic" > $(UNIT_COVER_FILE)
@@ -599,12 +599,14 @@ cover_ci: $(COVER_ROOT)/cover.out $(BIN)/goveralls
 	$(BIN)/goveralls -coverprofile=$(COVER_ROOT)/cover.out -service=buildkite || echo Coveralls failed;
 
 install-schema: cadence-cassandra-tool
+	$Q echo installing schema
 	./cadence-cassandra-tool create -k cadence --rf 1
 	./cadence-cassandra-tool -k cadence setup-schema -v 0.0
 	./cadence-cassandra-tool -k cadence update-schema -d ./schema/cassandra/cadence/versioned
 	./cadence-cassandra-tool create -k cadence_visibility --rf 1
 	./cadence-cassandra-tool -k cadence_visibility setup-schema -v 0.0
 	./cadence-cassandra-tool -k cadence_visibility update-schema -d ./schema/cassandra/visibility/versioned
+	$Q echo installed schema
 
 install-schema-mysql: cadence-sql-tool
 	./cadence-sql-tool --user root --pw cadence create --db cadence
--- a/docker/buildkite/docker-compose-local.yml
+++ b/docker/buildkite/docker-compose-local.yml
@@ -213,9 +213,12 @@ services:
       - "KAFKA_SEEDS=kafka"
       - "TEST_TAG=esintegration"
     depends_on:
-      - cassandra
-      - elasticsearch
-      - kafka
+      cassandra:
+        condition: service_healthy
+      elasticsearch:
+        condition: service_started
+      kafka:
+        condition: service_started
     volumes:
       - ../../:/cadence
       - /cadence/.build/ # ensure we don't mount the build directory
@@ -238,9 +241,12 @@ services:
       - "KAFKA_SEEDS=kafka"
       - "TEST_TAG=esintegration"
     depends_on:
-      - cassandra
-      - elasticsearch
-      - kafka
+      cassandra:
+        condition: service_healthy
+      elasticsearch:
+        condition: service_started
+      kafka:
+        condition: service_started
     volumes:
       - ../../:/cadence
       - /cadence/.build/ # ensure we don't mount the build directory
--- a/docker/buildkite/docker-compose-local.yml
+++ b/docker/buildkite/docker-compose-local.yml
@@ -9,6 +9,11 @@ services:
       services-network:
         aliases:
           - cassandra
+    healthcheck:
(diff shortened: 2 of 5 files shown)


docker/buildkite/docker-compose.yml
```diff
@@ -7,6 +7,11 @@ services:
       services-network:
         aliases:
           - cassandra
+    healthcheck:
+      test: ["CMD", "cqlsh", "-u cassandra", "-p cassandra" ,"-e describe keyspaces"]
+      interval: 15s
+      timeout: 10s
```
taylanisikdemir, line 13: let's increase timeout to 30s to avoid flakiness in case cassandra bootstrap doesn't settle quickly when there's not enough cpu/memory

taylanisikdemir (approved):

docker/buildkite/docker-compose.yml
```diff
@@ -73,6 +78,7 @@ services:
       context: ../../
       dockerfile: ./docker/buildkite/Dockerfile
     environment:
+      - "CASSANDRA_HOST=cassandra"
```
Groxx, line 81: huh.  was this needed?  maybe due to install-schema?
most things should be using CASSANDRA_SEEDS (which is an awful name, _HOST is much better)

docker/buildkite/docker-compose-local.yml
```diff
@@ -115,17 +127,23 @@ services:
...
+      cassandra:
+        condition: service_healthy
+      elasticsearch:
+        condition: service_started
+      kafka:
+        condition: service_started
     volumes:
       - ../../:/cadence
+      - /cadence/.build/ # ensure we don't mount the build directory
+      - /cadence/.bin/ # ensure we don't mount the bin directory
```
Groxx, line 146: does this.... hide the out-of-docker `.build` and `.bin` folders?  any other consequences, e.g. reusing them between runs since volumes are generally intended to be persistent?

Groxx: ah, TIL "anonymous volumes".  so it sounds like this does reuse them between runs.

might be useful?  it'll save re-downloading stuff, which is likely good.  assuming the .build / .bin dirty-detection is good enough anyway, otherwise people will also have to know to delete these volumes...

if we really want this all to run quickly on local, we can also mount the GOPATH/pkg/mod folder, since that's platform-agnostic and accounts for the vast majority of the downloads.

docker/buildkite/docker-compose-local.yml
```diff
@@ -86,8 +91,9 @@ services:
     build:
       context: ../../
       dockerfile: ./docker/buildkite/Dockerfile
-    command: make cover_profile
+    command: sh -c "make .just-build && make install-schema && make cover_profile"
```
Groxx, line 94: poking at this locally, I kinda think we don't want `make install-schema` here.  it fails if the schema already exists, which means you can't re-run unit tests without tossing and restarting the cassandra container :\

Makefile
```diff
@@ -548,7 +548,7 @@ test_e2e_xdc: bins
 		go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log; \
 	done;
 
-cover_profile: bins
```
Groxx, line 551: bins may be excessive, but this is broadly useful to ensure it builds before running all tests.  otherwise it can get like 90% of the way through before failing on a syntax error or partial-refactor or something.

docker/buildkite/docker-compose.yml
3vilhamster: yes

docker/buildkite/docker-compose-local.yml
3vilhamster: It not only rebuild stuff, it reruns generate, goimports, copyright. So no point on that.
And it is broken on local machine, unless you are running on linux x86.
I want to introduce GOPATH/pkg/mod cache in a separate pr. Maybe as a separate problem we can introduce volumes. Though, all our steps are running independently, so I doubt they are reusing volumes.

docker/buildkite/docker-compose-local.yml
3vilhamster: This is not the tool that you use to run unit tests. This is a "close to CI" way to simulate what is running on buildkite, so I wanted it to behave as CI run. If it will be needed, we can separate it to an extra step, but it was useful to debug the problem

Makefile
3vilhamster: We have a separate job that generates bins and verifies copyright and etc. If something is broken, I expect tests to fail. In this case bins generation/verification takes a lot of time.

docker/buildkite/docker-compose.yml
3vilhamster: Increased

Makefile
Groxx: yea, fair.  I suppose this is really only intended for CI, most people should be using `make cover` instead?

Makefile
```diff
@@ -599,12 +599,14 @@ cover_ci: $(COVER_ROOT)/cover.out $(BIN)/goveralls
 	$(BIN)/goveralls -coverprofile=$(COVER_ROOT)/cover.out -service=buildkite || echo Coveralls failed;
 
 install-schema: cadence-cassandra-tool
+	$Q echo installing schema
```
Groxx, line 602: not complaining or anything, nor requesting it be removed, it just seems a bit unnecessary - `./cadence-cassandra-tool -k cadence setup-schema -v 0.0` will be printed a second later

docker/buildkite/docker-compose-local.yml
Groxx: I suppose in that scenario, people would have to drop cassandra each time?  and that's reasonable because it matches CI?

which may be reasonable, this is in a buildkite folder after all.  it's not a totally general thing.  just checking / understanding.

Groxx (commented):
I'm too late for an official review, but yea this all looks reasonable + an improvement, and it seems to work well for me locally.  Thank you!

Makefile
3vilhamster: good point :) I'll clean it up in a following PR. I was just confused when was debugging failures.

PR #5481, opened 2023-12-16
Fix copying suite.Suite in integration tests

3vilhamster:
- Fix integration suite copying suite while passing it around

**What changed?**
Changed TestBase and IntegrationBase to be passed as a pointer to avoid copying suite.Suite internals. That can cause side effects like missing links to test initiated and internal mutex is copied together with the struct.

**Why?**
Lint warning, though I see a lot of logs in integration tests that likely are leaking due to this copying.

**How did you test it?**
run integration tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/domain/replicationTaskExecutor_test.go
+++ b/common/domain/replicationTaskExecutor_test.go
@@ -40,8 +40,7 @@ import (
 
 type (
 	domainReplicationTaskExecutorSuite struct {
-		suite.Suite
-		persistencetests.TestBase
+		*persistencetests.TestBase
 		domainReplicator *domainReplicationTaskExecutorImpl
 	}
 )
@@ -61,7 +60,7 @@ func (s *domainReplicationTaskExecutorSuite) TearDownSuite() {
 
 func (s *domainReplicationTaskExecutorSuite) SetupTest() {
 	s.TestBase = public.NewTestBaseWithPublicCassandra(&persistencetests.TestBaseOptions{})
-	s.TestBase.Setup()
+	s.Setup()
 	zapLogger, err := zap.NewDevelopment()
 	s.Require().NoError(err)
 	logger := loggerimpl.NewLogger(zapLogger)
--- a/common/domain/replicationTaskExecutor_test.go
+++ b/common/domain/replicationTaskExecutor_test.go
@@ -57,12 +57,13 @@ func TestDomainReplicationTaskExecutorSuite(t *testing.T) {
 	s := new(domainReplicationTaskExecutorSuite)
 
 	s.TestBase = public.NewTestBaseWithPublicCassandra(&persistencetests.TestBaseOptions{})
-	s.Setup()
 
 	suite.Run(t, s)
 }
 
 func (s *domainReplicationTaskExecutorSuite) SetupTest() {
+	s.Setup()
+
 	zapLogger, err := zap.NewDevelopment()
 	s.Require().NoError(err)
 	logger := loggerimpl.NewLogger(zapLogger)
--- a/common/domain/replicationTaskExecutor_test.go
+++ b/common/domain/replicationTaskExecutor_test.go
@@ -22,6 +22,8 @@ package domain
 
 import (
 	"context"
+	"log"
+	"os"
 	"testing"
 
 	"github.com/pborman/uuid"
@@ -47,20 +49,20 @@ type (
 
 func TestDomainReplicationTaskExecutorSuite(t *testing.T) {
 	testflags.RequireCassandra(t)
+
+	if testing.Verbose() {
+		log.SetOutput(os.Stdout)
+	}
+
 	s := new(domainReplicationTaskExecutorSuite)
+
+	s.TestBase = public.NewTestBaseWithPublicCassandra(&persistencetests.TestBaseOptions{})
+	s.Setup()
+
 	suite.Run(t, s)
 }
 
-func (s *domainReplicationTaskExecutorSuite) SetupSuite() {
-}
-
-func (s *domainReplicationTaskExecutorSuite) TearDownSuite() {
-
-}
-
 func (s *domainReplicationTaskExecutorSuite) SetupTest() {
-	s.TestBase = public.NewTestBaseWithPublicCassandra(&persistencetests.TestBaseOptions{})
-	s.Setup()
 	zapLogger, err := zap.NewDevelopment()
 	s.Require().NoError(err)
 	logger := loggerimpl.NewLogger(zapLogger)
--- a/common/persistence/nosql/nosqlplugin/common.go
+++ b/common/persistence/nosql/nosqlplugin/common.go
@@ -22,18 +22,18 @@
 package nosqlplugin
 
 import (
-	"os"
+	"fmt"
+	"os/exec"
 	"strings"
 )
 
 func getCadencePackageDir() (string, error) {
-	cadencePackageDir, err := os.Getwd()
+	cmdOut, err := exec.Command("git", "rev-parse", "--show-toplevel").Output()
 	if err != nil {
 		panic(err)
 	}
-	cadenceIndex := strings.LastIndex(cadencePackageDir, "/cadence/")
-	cadencePackageDir = cadencePackageDir[:cadenceIndex+len("/cadence/")]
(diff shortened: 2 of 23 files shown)


host/testcluster.go
```diff
@@ -27,10 +27,8 @@ import (
 	"testing"
 	"time"
 
-	"github.com/uber/cadence/common/log/loggerimpl"
-
+	"github.com/startreedata/pinot-client-go/pinot"
```
taylanisikdemir, line 30: nit: import groups should ideally go in following order
- builtin packages
- repo specific packages
- external packages
- named packages

common/persistence/persistence-tests/persistenceTestBase.go
```diff
@@ -111,13 +111,13 @@ const (
...
 // NewTestBaseFromParams returns a customized test base from given input
-func NewTestBaseFromParams(params TestBaseParams) TestBase {
+func NewTestBaseFromParams(params TestBaseParams) *TestBase {
 	logger, err := loggerimpl.NewDevelopment()
 	if err != nil {
 		panic(err)
 	}
 
-	return TestBase{
+	return &TestBase{
```
taylanisikdemir, line 120: looks like some other tests depend on this too. unit tests are failing.

taylanisikdemir (approved):

host/testcluster.go
3vilhamster: I'm using default goimports behaviour. 
As far as I've seen discussions about grouping, the outcome is that there should be either 2 or 3 groups. But in Uber style guide, we've settled in 2 groups. https://github.com/uber-go/guide/blob/master/style.md#import-group-ordering

common/persistence/persistence-tests/persistenceTestBase.go
3vilhamster: Yes, it turns out that most of persistence tests use this and copy suite all the time. Fixed

host/testcluster.go
Groxx: broadly I agree, but tbh I don't think it matters much, and it doesn't affect behavior as of [some semi-recent Go version].  automate it or ignore it, imo, it's hidden in almost all editors anyway and it doesn't make code review any more difficult.

Groxx (commented):
huh.  TIL that `go vet`'s at-test-time checks don't include copy-lock checks... that's quite surprising, and yeah there are _loads_ of copy-lock problems in these tests D:

we should deeeeefinitely add `go vet -copylocks ./...` to our PR-blocking lint checks.  too many failures at the moment though.  
thankfully they're all in tests except for `common/persistence/nosql`, and those might be safe in practice (but still absolutely deserve fixing, there's no reason at all to allow that to exist)

common/domain/replicationTaskExecutor_test.go
```diff
@@ -40,28 +42,28 @@ import (
...
 	testflags.RequireCassandra(t)
-	s := new(domainReplicationTaskExecutorSuite)
-	suite.Run(t, s)
-}
 
-func (s *domainReplicationTaskExecutorSuite) SetupSuite() {
-}
+	if testing.Verbose() {
+		log.SetOutput(os.Stdout)
+	}
```
Groxx, line 55: does using `t.Log`-proxied stuff work, or do these run into races if you do that?  I know it's not directly `log` compatible, but I doubt that matters in practice for our tests.

---

otherwise:
- default stderr -> verbose stdout: essentially matches `go test -v` (stdout) so 👍 that seems reasonable.
- to match `-v` more closely, do we want to maybe `io.Discard` when not verbose?  none of our existing stuff treats stdout vs stderr differently, and CI is verbose, so I don't know if there's much benefit to specializing it like this unless it's going to lead to a pragmatic or technical improvement somewhere.
- this seems broadly useful, given all the bare log use, and this one test doesn't seem special at all at a glance (is it in practice?).  seems worth adding to a test-helper package and using everywhere?

common/persistence/nosql/nosqlplugin/common.go
```diff
@@ -31,8 +31,8 @@ func getCadencePackageDir() (string, error) {
 	if err != nil {
 		panic(err)
 	}
-	cadenceIndex := strings.LastIndex(cadencePackageDir, "/cadence/")
-	cadencePackageDir = cadencePackageDir[:cadenceIndex+len("/cadence/")]
+	cadenceIndex := strings.LastIndex(cadencePackageDir, "cadence/")
+	cadencePackageDir = cadencePackageDir[:cadenceIndex+len("cadence/")]
```
Groxx, line 35: was this failing somewhere, or just felt like removing `/` prefixes?  seems fine either way since it doesn't change the behavior, just odd to see it change.

Groxx (approved):
I _suspect_ that not all of these are necessary on a technical level... but using a pointer for all these seems entirely reasonable too, and also more copy-bug-resistant (it's very clear that copies are not being used for intentional behavior here), so 👍 

---

I'm game to approve now and let you merge with whatever since it seems all safe + a clear improvement, and/or feel free to re-request review if ya want more input after changes.

host/testcluster.go
Groxx: It's still a bit surprising to me that goimports doesn't have an option to forcibly reorder stuff... but this looks like it might be reasonable: https://github.com/incu6us/goimports-reviser

I might see if that does what we want.  The less we have to touch here, the better.

Groxx: it's like 99% good... but it strips comments on anything it moves :|
if that's fixed it seems quite promising tho.  probably not too hard, astutil.CommentMap addresses exactly this kind of thing.

davidporter-id-au: I'd vote to put whatever behaviour we settle on in a linter and just enforce it at the lint level

common/persistence/nosql/nosqlplugin/common.go
3vilhamster: My local folder is called my-cadence :D and it failed. I've tried switching to finding git root, but apparently our CI is running on windows and has some issues with safe-folder related stuff. So I just fixed the `/` part.

common/domain/replicationTaskExecutor_test.go
3vilhamster: In the following PR I want to move all logging into the root suite TestBase/IntegrationBase part and use -v to either use zaptest or use zap.NewDevelopment depending on the `-v`. At least that is my plan.

host/testcluster.go
3vilhamster: There is also this https://github.com/NonLogicalDev/gofancyimports which is pretty good. It is metioned in a issue for goimports https://github.com/golang/go/issues/20818, though it is quite fresh.

Unfortunately that even in monorepo there is no tooling to enforce proper imports.

Groxx: gofancyimports looks great at a glance:
- keeps comments
- enforces grouping
- drops `import ( "single/thing" )` parenthesis
- comments above an import keeps that group of imports organized together (manual grouping when desirable)
- comments on the same line as an import get reordered with other imports

I can get a PR for that up today, it's pretty simple.  just gotta verify the changes/license/etc first.

common/domain/replicationTaskExecutor_test.go
Groxx: if zaptest doesn't cause data races (due to unclean shutdown), _always_ prefer zaptest.  there's basically no reason to use zap.NewDevelopment or log.* or fmt.Println in tests, t.Log-based stuff is massively better.

common/persistence/nosql/nosqlplugin/common.go
Groxx: our ci is definitely on linux, so it's not that at least.  either way this seems totally fine tho.

we could eliminate this entirely if we used `//go:embed` for the indexes, but that's a separate thing.

3vilhamster: Well, from the errors like from 
```git rev-parse --show-toplevel```
that I've got when using https://github.com/Integralist/go-findroot/blob/master/find/find.go#L36
```fatal: detected dubious ownership of /cadence```
means that we are running on windows with WSL.

PR #5482, opened 2023-12-16
Unify cassandra setup in docker-compose

3vilhamster:
**What changed?**
I've set limits for java heap size to shrink cassandra memmory allocation.
Also, unified all cassandra images in docker-compose files with the same values for heap, healthchecks and depends_on

**Why?**
To have a sane and unified way to run minimal local/ci cassandra setup.

**How did you test it?**
Local docker-compose up and local docker/buildkite run

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/docker/buildkite/docker-compose-es7.yml
+++ b/docker/buildkite/docker-compose-es7.yml
@@ -3,10 +3,18 @@ version: "3.5"
 services:
   cassandra:
     image: cassandra:4.1.3
+    environment:
+      - "MAX_HEAP_SIZE=256M"
+      - "HEAP_NEWSIZE=128M"
     networks:
       services-network:
         aliases:
           - cassandra
+    healthcheck:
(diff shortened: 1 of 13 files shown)


Groxx (approved):
C* running memory drops from 2GB+ to ~1GB, so I'm kinda curious why it's so much higher than the listed values... but yeah!  Seems to work great, and I can even launch `docker-compose.yml` with only 2GB allocated to Docker.  Big improvement.

docker/buildkite/docker-compose-es7.yml
```diff
@@ -3,10 +3,18 @@ version: "3.5"
 services:
   cassandra:
     image: cassandra:4.1.3
+    environment:
+      - "MAX_HEAP_SIZE=256M"
```
taylanisikdemir, line 7: thanks!

taylanisikdemir (approved):

PR #5483, opened 2023-12-16
deprecate logrus

3vilhamster:
**What changed?**
removed legacy usages of logrus

**Why?**
We have a unified approach of logging with project logger wrapper and zap as a default implementation. It is better to have only one logger implementation to avoid confusion.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/admin.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/admin.go
@@ -27,9 +27,8 @@ import (
 	"os"
 	"strings"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/uber/cadence/common/config"
+	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin"
 	"github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra/gocql"
 	"github.com/uber/cadence/tools/cassandra"
@@ -43,7 +42,7 @@ const (
 var _ nosqlplugin.AdminDB = (*cdb)(nil)
 
 func (db *cdb) SetupTestDatabase(schemaBaseDir string) error {
-	err := createCassandraKeyspace(db.session, db.cfg.Keyspace, 1, true)
+	err := db.createCassandraKeyspace(db.session, db.cfg.Keyspace, 1, true)
 	if err != nil {
 		return err
 	}
@@ -89,7 +88,7 @@ func (db *cdb) loadVisibilitySchema(fileNames []string, schemaBaseDir string) er
 }
 
 func (db *cdb) TeardownTestDatabase() error {
-	err := dropCassandraKeyspace(db.session, db.cfg.Keyspace)
+	err := db.dropCassandraKeyspace(db.session, db.cfg.Keyspace)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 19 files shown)


common/persistence/nosql/nosqlplugin/cassandra/admin.go
```diff
@@ -43,7 +42,7 @@ const (
 var _ nosqlplugin.AdminDB = (*cdb)(nil)
 
 func (db *cdb) SetupTestDatabase(schemaBaseDir string) error {
-	err := createCassandraKeyspace(db.session, db.cfg.Keyspace, 1, true)
+	err := db.createCassandraKeyspace(db.session, db.cfg.Keyspace, 1, true)
```
Groxx, line 45: probably worth changing this and `drop*` to use the fields rather than args - the args don't seem to be customized anywhere and there are only like two uses total.

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -1171,7 +1171,7 @@ func (s *ExecutionManagerSuite) TestPersistenceStartWorkflow() {
 
 	task1, err1 := s.CreateWorkflowExecution(ctx, domainID, workflowExecution, "queue1", "wType1", 20, 14, nil, 3, 0, 2, nil, nil)
 	s.Error(err1, "Expected workflow creation to fail.")
-	log.Infof("Unable to start workflow execution: %v", err1)
+	log.Printf("Unable to start workflow execution: %v\n", err1)
```
Groxx, line 1174: printf's fine / no need to change, but this is equivalent to the somewhat-simpler:
```
log.Println("Unable to start workflow execution:", err1)
```

go.mod
```diff
@@ -66,6 +66,8 @@ require (
 	gopkg.in/yaml.v2 v2.3.0
 )
 
+require github.com/google/go-cmp v0.5.8
```
Groxx, line 69: seems somewhat unlikely, I don't think you added a go-cmp import anywhere.  try tidying again?

Groxx: huh.  yeah that is in use.  no idea why it was indirect before 🤔

LGTM

Groxx:
I'll vote to update the create/drop methods since that seems simple, but either way seems reasonable + LGTM, merge when you're happy 👍

Groxx (approved):

common/persistence/nosql/nosqlplugin/cassandra/admin.go
```diff
@@ -98,34 +97,34 @@ func (db *cdb) TeardownTestDatabase() error {
 }
 
 // createCassandraKeyspace creates the keyspace using this session for given replica count
-func createCassandraKeyspace(s gocql.Session, keyspace string, replicas int, overwrite bool) (err error) {
+func (db *cdb) createCassandraKeyspace(s gocql.Session, keyspace string, replicas int, overwrite bool) (err error) {
 	// if overwrite flag is set, drop the keyspace and create a new one
 	if overwrite {
```
davidporter-id-au, line 102: no concerns here, but just FYI for all those who read this code, this is used in DB / infra provisioning and we need to be *very* careful about changes in `cadence/tools/provision` and their interaction with this.

 No concerns here though.

davidporter-id-au (approved):

common/persistence/persistence-tests/executionManagerTest.go
taylanisikdemir: we can just use `s.T().Logf` in tests

taylanisikdemir (approved):

PR #5486, opened 2023-12-18
Getting rid of gomock.Finish

dkrotx:
Since go 1.14 it is not required - gomock registers its finilazer with
t.Cleanup().
At the same time, some of our tests don't call .Finish and some do.
Aligning the experience.

**What changed?**

Unifying testing experience.

make test - unit tests

Only if someone has go <1.14 (unmaintained since Feb 2021)

### DIFF
--- a/service/worker/failovermanager/workflow_test.go
+++ b/service/worker/failovermanager/workflow_test.go
@@ -327,9 +327,7 @@ func (s *failoverWorkflowTestSuite) TestShouldFailover() {
 }
 
 func (s *failoverWorkflowTestSuite) TestGetDomainsActivity() {
-	env, mockResource, controller := s.prepareTestActivityEnv()
-	defer controller.Finish()
-	defer mockResource.Finish(s.T())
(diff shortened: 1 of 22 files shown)


3vilhamster (approved):
LGTM

service/worker/failovermanager/workflow_test.go
```diff
@@ -635,5 +623,10 @@ func (s *failoverWorkflowTestSuite) prepareTestActivityEnv() (*testsuite.TestAct
 	s.activityEnv.SetWorkerOptions(worker.Options{
 		BackgroundActivityContext: context.WithValue(context.Background(), failoverManagerContextKey, ctx),
 	})
-	return s.activityEnv, mockResource, controller
+
+	s.T().Cleanup(func() {
+		mockResource.Finish(s.T())
+	})
```
Groxx, line 629: can we do this in the `resource.NewTest` func, so it's picked up everywhere like gomock, or does that cause too many issues at the moment?

PR #5488, opened 2023-12-18
Updated the unit test for BlobstoreIterator into a table format

agautam478:
**What changed?**

Converted the already present test that was using Testing suite into and independent table test.
- The `pItr` variable is an iterator created by fetcher.ConcreteExecutionIterator, which iterates over concrete executions.
- The `NewBlobstoreWriter` is used to create a writer for the blobstore, and executions are added to this writer.
- After adding all executions, the writer is flushed, and assertions are made to check the correctness of the operation.
- The `NewBlobstoreIterator` is then used to iterate over the written blobs, and assertions are made to ensure the data integrity.

**Why?**
These changes were made as a part of our initiative to increase the code coverage and also to maintain consistency with it's sister `BlobstoreWriter` test.

**How did you test it?**
yes. Tested locally.

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


common/reconciliation/store/writerIterator_test.go
```diff
@@ -48,60 +46,60 @@ var (
...
-func (s *WriterIteratorSuite) SetupTest() {
-	s.Assertions = require.New(s.T())
-}
+func TestWriterIterator(t *testing.T) {
+	testCases := []struct {
+		name         string
+		pages        int
+		countPerPage int
+	}{
+		{"StandardCase", 10, 10},
```
taylanisikdemir, line 55: are there different error cases to cover with new test cases?

service/history/handler.go
```diff
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/uber/cadence/service/history/workflow"
```
taylanisikdemir, line 29: is this added by mistake?

agautam478: I am not sure where that came from.

common/reconciliation/store/writerIterator_test.go
taylanisikdemir: yeah let's add the ones that validates pagination works

taylanisikdemir (approved):

PR #5490, opened 2023-12-18
fix scavenger test suite

3vilhamster:
**What changed?**
Fixed ScavengerTestSuite not waiting for the stop of the running scavenger.

**Why?**
The tests were leaking goroutine and logging outside of tests that can lead to panic: Log in goroutine after '' has completed

**How did you test it?**
unit tests with count=100

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/scanner/tasklist/scavenger_test.go
+++ b/service/worker/scanner/tasklist/scavenger_test.go
@@ -193,8 +193,16 @@ func (s *ScavengerTestSuite) TestAllExpiredTasksWithErrors() {
 func (s *ScavengerTestSuite) runScavenger() {
 	s.scvgr.Start()
 	timer := time.NewTimer(scavengerTestTimeout)
+
+	waiter := make(chan struct{})
+
+	go func() {
+		s.scvgr.stopWG.Wait()
+		close(waiter)
+	}()
+
 	select {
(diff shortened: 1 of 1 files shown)


taylanisikdemir (approved):

service/worker/scanner/tasklist/scavenger_test.go
```diff
@@ -193,8 +193,16 @@ func (s *ScavengerTestSuite) TestAllExpiredTasksWithErrors() {
 func (s *ScavengerTestSuite) runScavenger() {
 	s.scvgr.Start()
 	timer := time.NewTimer(scavengerTestTimeout)
+
+	waiter := make(chan struct{})
+
+	go func() {
+		s.scvgr.stopWG.Wait()
```
taylanisikdemir, line 200: I missed this during review. `s.scvgr.Stop()` is never called in tests and instead the passed ctx is expected to cancel after hardcoded 10s. Seems like that suite could be improved to manage lifecycle better and be self contained per test case.

3vilhamster: Kind of a fix https://github.com/uber/cadence/pull/5510

PR #5491, opened 2023-12-18
Add unit test for sqlConfigStore

Shaddoll:
**What changed?**
Add unit test for sqlConfigStore

**Why?**
To improve coverage

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/sql/sqlConfigStore_test.go
```diff
@@ -0,0 +1,177 @@
+// Modifications Copyright (c) 2020 Uber Technologies Inc.
```
taylanisikdemir, line 1: let's avoid camelCase file names. rename to `sql_config_store_test.go`. Also recommend renaming all files in this folder to keep consistent.

common/persistence/sql/sqlConfigStore_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockDB := sqlplugin.NewMockDB(ctrl)
+			store, err := NewSQLConfigStore(mockDB, nil, nil)
+			require.NoError(t, err)
```
taylanisikdemir, line 90: can you provide the optional 3rd parameter to these require/assert calls so test output contains more context?
I personally don't prefer these helpers and just do validation natively `if condition {t.Errorf("Got %v, want: %v")}`

dkrotx: +1 for the context. I think it is well-expected to have a comment (a reason) for require/assert as the latest argument though.

taylanisikdemir (approved):

PR #5487, opened 2023-12-18
Unify logging in tests

3vilhamster:
**What changed?**
I've separate test logger into a dedicated package and centralized verbose testing approach.
Switch to zaptest.New that will print out logs only if test are failing to avoid logging on success.
When required, you can always fallback to Development using -v flag.

**Why?**
It is better to have common approach to logging in tests.

**How did you test it?**
unit tests/integration tests

**Potential risks**
None

**Release notes**

**Documentation Changes**

PR #5493, opened 2023-12-18
Switch to gci, so we can actually auto-group imports

Groxx:
# If you are conflicting with this PR

I would recommend simply _recreating_ its changes, it is likely to be MUCH simpler.
Before you merge or rebase, do this:
```bash
# get these makefile + tool changes
git checkout master -- internal/tools Makefile

# force a re-format
rm .build/fmt
make fmt

# get rid of the from-main changes
git checkout HEAD -- internal/tools Makefile

# now rebase or merge
```

---

We've finally found a tool that enforces groups, not just sorts them!
Awesome.
Unfortunately it doesn't (currently?) remove unused imports, so we also
still want/need `goimports`.
Oh well.


Makefile
```diff
@@ -180,8 +180,15 @@ $(BIN)/mockery: internal/tools/go.mod
 $(BIN)/enumer: internal/tools/go.mod
 	$(call go_build_tool,github.com/dmarkham/enumer)
 
-$(BIN)/goimports: internal/tools/go.mod
-	$(call go_build_tool,golang.org/x/tools/cmd/goimports)
+# it's important that the formatter is NOT whatever version they have installed,
+# so build it for this project.
+# prefer go_build_tool to support newer go versions, but either is reasonable.
+$(BIN)/gofmt: internal/tools/go.mod
+	$(call go_build_tool,cmd/gofmt)
```
Groxx, line 187: gofancyimports doesn't format, and it conflicts with goimports (which does format).

so now we have "normal" gofmt use.

perhaps we should add `-s`?

cmd/server/cadence/server_test.go
```diff
@@ -29,18 +29,16 @@ import (
...
-
-	"github.com/uber/cadence/testflags"
-
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/config"
-	_ "github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra"              // needed to load cassandra plugin
+	_ "github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra" // needed to load cassandra plugin
```
Groxx, line 37: there is an option to group side-effect imports like this, which might be worth doing...

Makefile
3vilhamster: I guess this is the issues with fancyimports. I wonder how we can use the same logic in IDEs.
Maybe we can create a wrapper that does goimports + fancyimports afterwards?

host/testcluster.go
```diff
@@ -50,6 +51,7 @@ import (
 	"github.com/uber/cadence/common/persistence/nosql"
 	"github.com/uber/cadence/common/persistence/persistence-tests/testcluster"
 	"github.com/uber/cadence/testflags"
+
 	// the import is a test dependency
 	_ "github.com/uber/cadence/common/persistence/nosql/nosqlplugin/cassandra/gocql/public"
 	persistencetests "github.com/uber/cadence/common/persistence/persistence-tests"
```
Groxx, line 57: possibly sub-optimal due to leading comment

internal/tools/go.mod
```diff
@@ -245,7 +245,7 @@ require (
 	go.opencensus.io v0.24.0 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
 	go.uber.org/fx v1.13.1 // indirect
-	go.uber.org/multierr v1.8.0 // indirect
+	go.uber.org/multierr v1.11.0 // indirect
```
Groxx, line 248: I should check what exactly this changes.  I doubt it matters to our code now, but it could in the future.

internal/tools/tools.go
```diff
@@ -26,25 +26,32 @@ package tools
...
 
 	// protobuf stuff
 	_ "github.com/gogo/protobuf/protoc-gen-gofast"
 	_ "go.uber.org/yarpc/encoding/protobuf/protoc-gen-yarpc-go"
 
-	// goimports
-	_ "golang.org/x/tools/cmd/goimports"
+	// forces one import-order pattern
+	_ "github.com/NonLogicalDev/gofancyimports/cmd/gofancyimports"
+
```
Groxx, line 39: I'm slightly inclined to make these trailing so they sort together...

... but also meh.  happy to do so if someone feels more strongly.

schema/elasticsearch/embed.go
```diff
@@ -20,7 +20,7 @@
 
 package elasticsearch
 
-import _ "embed"
+import _ "embed" // enable embedding
```
Groxx, line 23: interesting decision to _add_ a comment... but tbh I agree with this, I think the blank embed-import is a very weird requirement.

tools/cli/factory.go
```diff
@@ -22,29 +22,26 @@ package cli
...
 
-	"go.uber.org/yarpc/peer"
-	"go.uber.org/yarpc/peer/hostport"
-	"go.uber.org/yarpc/transport/grpc"
-	"go.uber.org/yarpc/transport/tchannel"
-	"google.golang.org/grpc/credentials"
-
 	"github.com/olivere/elastic"
+	adminv1 "github.com/uber/cadence-idl/go/proto/admin/v1"
+	apiv1 "github.com/uber/cadence-idl/go/proto/api/v1"
```
Groxx, line 32: this is kinda the worst-case scenario for not grouping -idl (and arguably go.uber.org/cadence) into the "project" imports.  or some "us"-group like goimports-reviser supports.

but on balance I think I prefer to cut a strict line between _literally this project_ and everything else, vs not.

Makefile
Groxx: to be clear, removing goimports and using gofmt is not a _problem_.  this is just the motivation.  gofmt is faster, and gofancyimports takes care of the imports that gofmt doesn't touch.

internal/tools/go.mod
Groxx: yea, looks fine.  cleanups / new top-level funcs, all seem reasonable.

Makefile
Groxx: ah, for IDEs: should definitely be possible, though unlike goimports it needs to be _in addition to_ gofmt, it's not a replacement.  combining those might require a small wrapper-binary 🤔 

though since it's an Analyzer, it should integrate into gopls just fine.

---

a combined-binary might be reasonable to add to gofancyimports, for IDE purposes.  GoLand has a "goimports-like" formatting option which acts a lot like this too.

service/worker/scanner/tasklist/scavenger_test.go
```diff
@@ -66,7 +66,9 @@ func (s *ScavengerTestSuite) SetupTest() {
 	s.taskMgr = &mocks.TaskManager{}
 	s.taskListTable = &mockTaskListTable{}
 	s.taskTables = make(map[string]*mockTaskTable)
-	logger := testlogger.New(s.T())
+	zapLogger, err := zap.NewDevelopment()
+	s.Require().NoError(err)
+	logger := loggerimpl.NewLogger(zapLogger)
```
Groxx, line 71: [see #5487 unfortunately](https://github.com/uber/cadence/pull/5487#pullrequestreview-1787811461)

Makefile
taylanisikdemir: After this lands, what should we all use as Format Tool in IDE?

taylanisikdemir (approved):
thanks for trying this import organizer and making bulk changes!

Makefile
Groxx: Just goimports with -local afaict.  I honestly disable this though, IDE formatting does funky and incorrect things kinda frequently

host/integrationbase.go
```diff
@@ -182,7 +180,7 @@ func (s *IntegrationBase) setupSuiteForPinotTest() {
 }
 
 func (s *IntegrationBase) setupLogger() {
-	s.Logger = testlogger.New(s.T())
+	s.Logger = loggerimpl.NewLoggerForTest(s.Suite)
```
Groxx, line 183: sadly another:
```
panic: Log in goroutine after TestIntegrationSuite has completed: 2023-12-19T00:11:34.305Z	WARN	Async task dispatch timed out	{"component": "matching-engine", "wf-domain-name": "integration-test-domain-66697d18-7114-4288-806e-80331e8c9c86", "wf-task-list-name": "integration-signal-with-start-workflow-id-reuse-test-tasklist", "wf-task-list-type": 0, "isolation-group": "", "wf-run-id": "c29381b5-fdf8-4357-b9da-a022f109dee7", "wf-id": "integration-signal-with-start-workflow-id-reuse-test", "queue-task-id": 1, "wf-domain-id": "6a5ef3dd-228b-4424-96f3-34517d523af9", "logging-call-at": "taskReader.go:418"}
goroutine 90361 [running]:
testing.(*common).logDepth(0xc000fe7ba0, {0xc000c886c0, 0x21d}, 0x3)
	/usr/local/go/src/testing/testing.go:1003 +0x68c
testing.(*common).log(...)
	/usr/local/go/src/testing/testing.go:985
testing.(*common).Logf(0xc000fe7ba0, {0x6202a90, 0x2}, {0xc0017e8260, 0x1, 0x1})
	/usr/local/go/src/testing/testing.go:1036 +0xa5
go.uber.org/zap/zaptest.testingWriter.Write({{0x62411f0?, 0xc000fe7ba0?}, 0xe0?}, {0xc0046f5c00, 0x21e, 0x400})
	/go/pkg/mod/go.uber.org/zap@v1.13.0/zaptest/logger.go:130 +0x12d
go.uber.org/zap/zapcore.(*ioCore).Write(0xc002835da0, {0x1, {0xc15855cd92375e0e, 0x81de9b65e1, 0x7ddfec0}, {0x0, 0x0}, {0x4b0ecef, 0x1d}, {0x0, ...}, ...}, ...)
	/go/pkg/mod/go.uber.org/zap@v1.13.0/zapcore/core.go:90 +0x19a
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc00180ca50, {0xc0003f5180, 0x6, 0xa})
	/go/pkg/mod/go.uber.org/zap@v1.13.0/zapcore/entry.go:216 +0x2ed
go.uber.org/zap.(*Logger).Warn(0xc003d4b3e0?, {0x4b0ecef, 0x1d}, {0xc0003f5180, 0x6, 0xa})
	/go/pkg/mod/go.uber.org/zap@v1.13.0/logger.go:195 +0x6e
github.com/uber/cadence/common/log/loggerimpl.(*loggerImpl).Warn(0xc003d4b3e0, {0x4b0ecef, 0x1d}, {0xc00016f040, 0x5, 0x5})
	/cadence/common/log/loggerimpl/logger.go:138 +0x18f
github.com/uber/cadence/service/matching.(*taskReader).dispatchSingleTaskFromBuffer(0xc002a8bcc0, {0x0, 0x0}, 0xc0052fa900)
	/cadence/service/matching/taskReader.go:418 +0x123d
github.com/uber/cadence/service/matching.(*taskReader).dispatchSingleTaskFromBufferWithRetries(...)
	/cadence/service/matching/taskReader.go:389
github.com/uber/cadence/service/matching.(*taskReader).dispatchBufferedTasks(0xc002a8bcc0, {0x0, 0x0})
	/cadence/service/matching/taskReader.go:159 +0x2c5
created by github.com/uber/cadence/service/matching.(*taskReader).Start
	/cadence/service/matching/taskReader.go:126 +0x171
FAIL	github.com/uber/cadence/host	557.826s
```
since these are so similar I just reverted all of `host`.

taylanisikdemir: taskReader's `Stop()` doesn't wait for these goroutines to complete. We can add a waitgroup to avoid this leak

Makefile
Groxx: from some side chats: we'll probably bundle a formatter-wrapper to simplify vscode and maybe other tools.

that's a bit easier to discover and fix after it's merged though, so I'm just going to skip it for now.

Makefile
```diff
@@ -463,6 +470,14 @@ release: ## Re-generate generated code and run tests
 	$(MAKE) --no-print-directory go-generate
 	$(MAKE) --no-print-directory test
 
+build: ## Build all packages and all tests (ensures everything compiles)
+	$Q echo 'Building all packages...'
+	$Q go build ./...
+	$Q # "tests" by building and then running `true`, and hides test-success output
+	$Q echo 'Building all tests (~5x slower)...'
+	$Q # intentionally not -race due to !race build tags
+	$Q go test -exec /usr/bin/true ./... >/dev/null
```
taylanisikdemir, line 479: I like this option `-exec /usr/bin/true` to just verify test build

Makefile
```diff
@@ -352,10 +357,12 @@ $(BUILD)/lint: $(LINT_SRC) $(BIN)/revive | $(BUILD)
...
-$(BUILD)/fmt: $(ALL_SRC) $(BIN)/goimports | $(BUILD)
-	$Q echo "goimports..."
-	$Q # use FRESH_ALL_SRC so it won't miss any generated files produced earlier
-	$Q $(BIN)/goimports -local "github.com/uber/cadence" -w $(FRESH_ALL_SRC)
+# use FRESH_ALL_SRC so it won't miss any generated files produced earlier.
+$(BUILD)/fmt: $(ALL_SRC) $(BIN)/goimports $(BIN)/gofancyimports | $(BUILD)
+	$Q echo "formatting and removing unused imports..."
+	$Q $(BIN)/goimports -w $(FRESH_ALL_SRC)
+	$Q echo "grouping imports..."
+	$Q $(BIN)/gofancyimports fix --local github.com/uber/cadence/ --write $(FRESH_ALL_SRC)
```
taylanisikdemir, line 365: no strong opinions but I'd also pass `--group-effect`

taylanisikdemir: also couldn't see that option but it would be better if ordering of groups were local packages first and then everything else.

taylanisikdemir (approved):

Makefile
Groxx: I've found it on a few go-official channels (groups and github), it's... very effective but I do wish it were just `go build -tests ./...` or something.  Kinda hard to remember.

Makefile
Groxx: It's not very configurable, so no - just the built-in groups and not even e.g. "company", just "project and everything else".

It's quite new though, so it seems kinda likely it'll gain more as time goes on.  The Go world has been stuck on goimports for an unbelievably long time, very glad to see a new contender.

Groxx: Re side effects: yeah I kinda lean that way too. They're legitimately special.

Might give it a try tomorrow and see what changes.

service/worker/scanner/tasklist/scavenger_test.go
3vilhamster: I would say that introducing a channel that is closed after the log and waiting for the full top should be a better fix.

host/integrationbase.go
3vilhamster: That will generate around 40Mb of logs, so I would prefer introducing a fix instead of hacking around. Let's land the revert and I'll work on the fix.

service/worker/scanner/tasklist/scavenger_test.go
3vilhamster: If I land first https://buildkite.com/uberopensource/cadence-server/builds/16817 then please, revert changes.

host/integrationbase.go
3vilhamster: Addressed this leak in https://github.com/uber/cadence/pull/5499

3vilhamster: Should be resolved. Please revert and let me know if there are more leaking parts.

Makefile
Groxx: similar stuff with gci due to unused-import cleanup, but this does also `gofmt` so a custom wrapper to run both is probably _less_ needed - IDEs often detect and warn and offer to clean those up separately.

cmd/server/cadence/server_test.go
Groxx: yea, I think this is nicer.  especially with gci, which collapses slightly more.

host/testcluster.go
Groxx: collapsed in gci, seems reasonable.

internal/tools/tools.go
Groxx: no need in gci, though trailing might be a bit nicer to read

schema/elasticsearch/embed.go
Groxx: ^ introduced by gofancyimports, and gci does not do this.  seems fine to keep / worth keeping though.

Makefile
Groxx: added that group!

host/integrationbase.go
Groxx: undid (just copied master) and tests pass on the first try - thanks!

service/worker/scanner/tasklist/scavenger_test.go
Groxx: reverting 👍 
possibly unrelated (I don't see any clear signs that it is related) but the first pass had a failure: https://buildkite.com/uberopensource/cadence-server/builds/16871#018c8557-8aae-4026-b49a-34476c703ad7

I've got other changes and retries coming though, could just be other flakiness.  will keep checking.

Groxx: has failed a couple times but did succeed, and overall still seems reasonably stable.  leaving the reverts un-reverted 👍

.gen/proto/history/v1/service.pb.go
```diff
@@ -33,9 +33,9 @@ import (
 
 	proto "github.com/gogo/protobuf/proto"
 	types "github.com/gogo/protobuf/types"
-
```
3vilhamster, line 36: I think we should ignore generated files. Maybe we can skip .gen?

Groxx: we haven't historically, and there isn't really any downside (it takes about 1 second to reformat everything).

tbh in most ways it's easier to format everything than to exclude stuff.

3vilhamster: Makes sense

Groxx: ah, one exception to all this is when the generate-and-format makefile isn't being careful about dependencies.  then having two things touch the same files can be a huuuge pain, and tbh I suspect that's why a lot of things avoid formatting gen.

but that's fixed!  our makefile should be actually-completely reliable about this.

PR #5494, opened 2023-12-18
Fixing matching:TestCheckIdleTaskList test flackiness

dkrotx:
For some reason it almost always fails in IDE (4/5 cases, goland),
but passes in console. At the same time, the test clearly had issues:
 - it's been 3 subtests in one
 - only the first subtest checked task-list became stopped when unused
 - the third subtest (the one which has been failing) never succeeded
   with `tlm.AddTask` since writer never started - instead of tlm.Start()
   we used special version which supposed to mimic it, but never started
   the writer.

The test is still problematic since it continues using time.Sleep all
around, but so do all the other tests for matching :-(

More comments and error-messages also bring more clarity to what the
test is actually for.

fix test flackiness in IDE

To be able to run tests in GoLand

make test + running all tests in GoLand

No risks, it is only the test been improved

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager_test.go
+++ b/service/matching/taskListManager_test.go
@@ -219,68 +219,77 @@ func TestDescribeTaskList(t *testing.T) {
 	require.Zero(t, taskListStatus.GetBacklogCountHint())
 }
 
-func tlMgrStartWithoutNotifyEvent(tlm *taskListManagerImpl) {
-	// mimic tlm.Start() but avoid calling notifyEvent
-	tlm.liveness.Start()
-	tlm.startWG.Done()
-	go tlm.taskReader.dispatchBufferedTasks(defaultTaskBufferIsolationGroup)
-	go tlm.taskReader.getTasksPump()
-}
-
 func TestCheckIdleTaskList(t *testing.T) {
-	controller := gomock.NewController(t)
-	logger := testlogger.New(t)
-
 	cfg := NewConfig(dynamicconfig.NewNopCollection(), "some random hostname")
 	cfg.IdleTasklistCheckInterval = dynamicconfig.GetDurationPropertyFnFilteredByTaskListInfo(10 * time.Millisecond)
 
-	// Idle
-	tlm := createTestTaskListManagerWithConfig(logger, controller, cfg)
-	tlMgrStartWithoutNotifyEvent(tlm)
-	time.Sleep(20 * time.Millisecond)
-	require.False(t, atomic.CompareAndSwapInt32(&tlm.stopped, 0, 1))
+	t.Run("Idle task-list", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		tlm := createTestTaskListManagerWithConfig(testlogger.New(t), ctrl, cfg)
+		require.NoError(t, tlm.Start())
 
-	// Active poll-er
-	tlm = createTestTaskListManagerWithConfig(logger, controller, cfg)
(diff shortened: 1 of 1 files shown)


service/matching/taskListManager_test.go
```diff
@@ -219,68 +218,78 @@ func TestDescribeTaskList(t *testing.T) {
...
+
+		time.Sleep(8 * time.Millisecond)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		_, _ = tlm.GetTask(ctx, nil)
+		cancel()
+
+		// task list manager should have been stopped,
+		// but GetTask extends auto-stop until the next check-idle-task-list-interval
+		time.Sleep(6 * time.Millisecond)
+		require.Equal(t, int32(0), tlm.stopped)
```
taylanisikdemir, line 249: as discussed in another PRs, let's try to provide more context in the last parameter of these require/assert calls wherever applicable.

taylanisikdemir (approved):

service/matching/taskListManager_test.go
```diff
@@ -219,68 +218,78 @@ func TestDescribeTaskList(t *testing.T) {
 	require.Zero(t, taskListStatus.GetBacklogCountHint())
 }
 
-func tlMgrStartWithoutNotifyEvent(tlm *taskListManagerImpl) {
-	// mimic tlm.Start() but avoid calling notifyEvent
-	tlm.liveness.Start()
-	tlm.startWG.Done()
-	go tlm.taskReader.dispatchBufferedTasks(defaultTaskBufferIsolationGroup)
```
davidporter-id-au, line 226: sorry, small brain following along here, are we ensuring these daemons are started in the new test?

dkrotx: I looked at the code - these daemons are starting among with tlm.taskReader.Start():
```
for g := range tr.taskBuffers {
	go tr.dispatchBufferedTasks(g)
}
go tr.getTasksPump()
```

I think the previous approach where we started *selected* things bypassing the tlm methods actually a bad approach. We were trying to avoid notifyEvent (which is, I believe, no longer happening in .Start), but also missed some important changes like starting the writer.

I think it is really important for us to cover the happy-paths as well. Otherwise tests like `TestAddTaskStandby` pass unintentionally: it assumed we had to stop the writer though it's never been started!

service/matching/taskListManager_test.go
```diff
@@ -219,68 +218,78 @@ func TestDescribeTaskList(t *testing.T) {
...
-	go tlm.taskReader.getTasksPump()
-}
-
 func TestCheckIdleTaskList(t *testing.T) {
-	controller := gomock.NewController(t)
-	logger := testlogger.New(t)
-
 	cfg := NewConfig(dynamicconfig.NewNopCollection(), "some random hostname")
 	cfg.IdleTasklistCheckInterval = dynamicconfig.GetDurationPropertyFnFilteredByTaskListInfo(10 * time.Millisecond)
+	logger := testlogger.New(t)
```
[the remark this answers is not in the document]
dkrotx: Good point! Fixed!

service/matching/taskListManager_test.go
dkrotx: I this particular case it is explained in the comment above, I think it's enough.
In other cases I added a message argument since there is nothing to say in a comment.

PR #5497, opened 2023-12-18
Improve logs/metrics of HandleDecisionTaskCompleted

taylanisikdemir:
**What changed?**
history/decision/handler's [HandleDecisionTaskCompleted](https://github.com/uber/cadence/blob/9224e5de1ffdbde1a8aea0e059027d4c7d8d7d44/service/history/decision/handler.go#L285) is doing a lot of things but with limited visibility so improving it with following changes:

- Added domain name and workflow type tag to all the metrics emitted from this function
- Added domain name and other available workflow related tags to all the logs
- Added more debug logs

**Why?**
Improve debugging

**How did you test it?**
`make test`

### DIFF
--- a/service/history/decision/handler.go
+++ b/service/history/decision/handler.go
@@ -356,7 +356,7 @@ Update_History_Loop:
 
 		if !msBuilder.IsWorkflowExecutionRunning() || !isRunning || currentDecision.Attempt != token.ScheduleAttempt || currentDecision.StartedID == common.EmptyEventID {
 			logger.Debugf("Decision task not found. IsWorkflowExecutionRunning: %v, isRunning: %v, currentDecision.Attempt: %v, token.ScheduleAttempt: %v, currentDecision.StartID: %v",
-				msBuilder.IsWorkflowExecutionRunning(), isRunning, currentDecision.Attempt, token.ScheduleAttempt, currentDecision.StartedID)
+				msBuilder.IsWorkflowExecutionRunning(), isRunning, getDecisionInfoAttempt(currentDecision), token.ScheduleAttempt, getDecisionInfoStartedID(currentDecision))
 			return nil, &types.EntityNotExistsError{Message: "Decision task not found."}
 		}
 
@@ -893,4 +893,18 @@ func (handler *handlerImpl) failDecisionHelper(
 
 func (handler *handlerImpl) getActiveDomainByID(id string) (*cache.DomainCacheEntry, error) {
 	return cache.GetActiveDomainByID(handler.shard.GetDomainCache(), handler.shard.GetClusterMetadata().GetCurrentClusterName(), id)
+}
+
+func getDecisionInfoAttempt(di *execution.DecisionInfo) int64 {
+	if di == nil {
+		return 0
+	}
+	return di.Attempt
+}
+
+func getDecisionInfoStartedID(di *execution.DecisionInfo) int64 {
+	if di == nil {
+		return 0
+	}
+	return di.StartedID
 }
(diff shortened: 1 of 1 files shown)


service/history/decision/handler.go
```diff
@@ -304,6 +303,18 @@ func (handler *handlerImpl) HandleDecisionTaskCompleted(
 		RunID:      token.RunID,
 	}
 
+	domainName := domainEntry.GetInfo().Name
+	logger := handler.logger.WithTags(
```
davidporter-id-au, line 307: excellent!

service/history/decision/handler.go
```diff
@@ -330,60 +342,53 @@ Update_History_Loop:
...
+			scope.IncCounter(metrics.StaleMutableStateCounter)
+			logger.Error("Encounter stale mutable state in RespondDecisionTaskCompleted", tag.WorkflowNextEventID(msBuilder.GetNextEventID()))
 			// Reload workflow execution history
 			wfContext.Clear()
 			continue Update_History_Loop
 		}
 
-		if !msBuilder.IsWorkflowExecutionRunning() || !isRunning || currentDecision.Attempt != token.ScheduleAttempt ||
-			currentDecision.StartedID == common.EmptyEventID {
+		if !msBuilder.IsWorkflowExecutionRunning() || !isRunning || currentDecision.Attempt != token.ScheduleAttempt || currentDecision.StartedID == common.EmptyEventID {
```
davidporter-id-au, line 357: Nit/formatting: What do you think about making this multiline?

service/history/decision/handler.go
```diff
@@ -330,60 +342,53 @@ Update_History_Loop:
...
+			logger.Debugf("Decision task not found. IsWorkflowExecutionRunning: %v, isRunning: %v, currentDecision.Attempt: %v, token.ScheduleAttempt: %v, currentDecision.StartID: %v",
```
davidporter-id-au, line 358: This will make log matching tricky if you use `debugf`. I would suggest making these tags.

service/history/decision/handler.go
```diff
@@ -330,60 +342,53 @@ Update_History_Loop:
...
+				msBuilder.IsWorkflowExecutionRunning(), isRunning, getDecisionInfoAttempt(currentDecision), token.ScheduleAttempt, getDecisionInfoStartedID(currentDecision))
 			return nil, &types.EntityNotExistsError{Message: "Decision task not found."}
 		}
 
 		startedID := currentDecision.StartedID
 		maxResetPoints := handler.config.MaxAutoResetPoints(domainEntry.GetInfo().Name)
 		if msBuilder.GetExecutionInfo().AutoResetPoints != nil && maxResetPoints == len(msBuilder.GetExecutionInfo().AutoResetPoints.Points) {
-			handler.metricsClient.IncCounter(metrics.HistoryRespondDecisionTaskCompletedScope, metrics.AutoResetPointsLimitExceededCounter)
+			logger.Debugf("Max reset points %d is exceeded", maxResetPoints)
```
davidporter-id-au, line 366: As before

service/history/decision/handler.go
```diff
@@ -891,3 +894,17 @@ func (handler *handlerImpl) failDecisionHelper(
 func (handler *handlerImpl) getActiveDomainByID(id string) (*cache.DomainCacheEntry, error) {
 	return cache.GetActiveDomainByID(handler.shard.GetDomainCache(), handler.shard.GetClusterMetadata().GetCurrentClusterName(), id)
 }
+
+func getDecisionInfoAttempt(di *execution.DecisionInfo) int64 {
```
davidporter-id-au, line 898: should this be a method on DecisionInfo?

davidporter-id-au (approved):

service/history/decision/handler.go
taylanisikdemir: Yeah in some cases it's nice to move variables to tags so message can be used to group logs but certain debug logs don't fit in that category and better to include more context inline. We can always get count of such logs via regex

PR #5499, opened 2023-12-19
fix leaky goroutines in matching

3vilhamster:
**What changed?**
Fixed matching taskReader stop not waiting for stopping underlying dispatchers.

**Why?**
With recent logging changes this leaks produces:
```
Log in goroutine after TestIntegrationSuite has completed: 2023-12-19T00:11:34.305Z
```

**How did you test it?**
Run integration tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/matching/taskListManager_test.go
+++ b/service/matching/taskListManager_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	"github.com/uber/cadence/common/cache"
 	"github.com/uber/cadence/common/cluster"
 	"github.com/uber/cadence/common/dynamicconfig"
--- a/service/matching/taskReader.go
+++ b/service/matching/taskReader.go
@@ -26,6 +26,7 @@ import (
 	"context"
 	"errors"
 	"runtime"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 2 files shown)


service/matching/taskListManager_test.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
dkrotx, line 34: this is most probably because of Go 1.21 != CI (1.20)

service/matching/taskReader.go
```diff
@@ -149,6 +154,8 @@ func (tr *taskReader) Signal() {
 }
 
 func (tr *taskReader) dispatchBufferedTasks(isolationGroup string) {
+	tr.stopWg.Add(1)
+	defer tr.stopWg.Done()
```
dkrotx, line 158: Looks like a generic approach for stopping. I wonder if it's possible to have some OOP around this.

service/matching/taskListManager_test.go
3vilhamster: Lets address this after @Groxx PR. Then everything will be unified.

service/matching/taskReader.go
3vilhamster: OOP in Go is hard, I'm not aware of generic solutions on this.

dkrotx (approved):

PR #5500, opened 2023-12-19
Adding unit tests to workflowHandler_test.go

sankari165:
**What changed?**
addition of unit tests 

**Why?**
ListDomains and Health methods did not have unit tests

**How did you test it?**
unit tests passed

**Potential risks**

**Release notes**

**Documentation Changes**


Shaddoll:
We're switching from suite tests to table tests.

dkrotx:
> We're switching from suite tests to table tests.

Hi @Shaddoll !
I'm for the coherent experience. If the tests are already written using suite, let's continue making them for the sake of improving coverage %.
We will have a migration effort anyway. I also wish we could have it before improving the code-coverage.

@sankari165 at the same time, when a new test-file is introduced, let's make it a standard format (using testing.T).

sankari165:
agreed. I did consider other options and decided to stick to suite tests only because I was adding tests to the existing file

dkrotx (approved):

PR #5505, opened 2023-12-19
Added a unit test for the timer.go file in reconciliation folder.

agautam478:
**What changed?**
Added a unit test for the `timer.go` file in reconciliation folder.

Test Cases:
- success
- Invalid Timer Causes Error
- Non-nil Pagination Token Passed

**Why?**
To improve the unit test coverage.

**How did you test it?**
tested locally

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			timerTasks := tc.setupMock()
+			fetchFn := getUserTimers(mockRetryer, minTimestamp, maxTimestamp, pageSize)
+			page, err := fetchFn(ctx, nil)
+
+			// Update the expected timestamp to match the mock data
+			if len(timerTasks) > 0 {
+				tc.expectedPage.Entities[0].(*entity.Timer).VisibilityTimestamp = timerTasks[0].VisibilityTimestamp
+			}
```
taylanisikdemir, line 107: you can avoid this by passing a fixed timestamp in test cases. see [this example](https://github.com/uber/cadence/blob/2e4e4604614ac298f7ad3627d0d3c468bf091209/common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go#L538)

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,117 @@
...
+	maxTimestamp := time.Now()
+
+	testCases := []struct {
+		name          string
+		setupMock     func() []*persistence.TimerTaskInfo
+		expectedPage  pagination.Page
+		expectedError bool
+	}{
+		{
+			name: "Success",
```
taylanisikdemir, line 54: great to have at least happy path covered but the failure cases are more interesting. I'd add following cases
1. pass non-nil token and validate it's passed down to `pr.GetTimerIndexTasks`
2. response has other timer types that are skipped ([this condition](https://github.com/taylanisikdemir/cadence/blob/e13da5812c24f55b7156c9d9b02bb94f7d5512b6/common/reconciliation/fetcher/timer.go#L69))
3. invalid timer causes error to be returned ([this condition](https://github.com/taylanisikdemir/cadence/blob/e13da5812c24f55b7156c9d9b02bb94f7d5512b6/common/reconciliation/fetcher/timer.go#L83))

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+			tc.setupMock()
+
+			var token pagination.PageToken
+			if tc.name == "Non-nil Pagination Token Passed" {
+				token = nonNilToken
+			}
+			fetchFn := getUserTimers(mockRetryer, minTimestamp, maxTimestamp, pageSize)
+			page, err := fetchFn(ctx, token)
+
+			require.Equal(t, tc.expectedError, err != nil)
```
dkrotx, line 171: it is a bit hard to read - I personally have to wrap my head around every time I see this 🤯 

I know, you have to do this to avoid if-s in table-tests, but maybe you can check for specific error type instead of expected or not? Then it will be natural to just check error against `expectedError error` field of the table-test.

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+				return []*persistence.TimerTaskInfo{invalidTimer}
+			},
+			expectedPage:  pagination.Page{}, // No page is expected due to error
+			expectedError: true,              // An error is expected due to the invalid timer
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.setupMock()
```
dkrotx, line 162: setupMock returns value, but it is not used.

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+						VisibilityTimestamp: fixedTimestamp,
+					},
+				},
+			},
+			expectedError: false,
+		},
+		{
+			name: "Non-nil Pagination Token Passed",
+			setupMock: func() []*persistence.TimerTaskInfo {
+				nonNilToken := []byte("non-nil-token")
```
dkrotx, line 104: it's the same as global... is it required?

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+						RunID:               "testRunID",
+						TaskType:            persistence.TaskTypeUserTimer,
+						VisibilityTimestamp: fixedTimestamp,
+					},
+				},
+			},
+			expectedError: false,
+		},
+		{
+			name: "Non-nil Pagination Token Passed",
```
dkrotx, line 102: nit: a bit confusing to see Passed in a test name. Will be especially confusing to see "Something ... Passed Failed" in IDE :-) Consider renaming to "Provided".

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+		t.Fatalf("Failed to parse timestamp: %v", err)
+	}
+
+	ctrl := gomock.NewController(t)
+
+	mockRetryer := persistence.NewMockRetryer(ctrl)
+	ctx := context.Background()
+	pageSize := 10
+	minTimestamp := fixedTimestamp.Add(-time.Hour)
+	maxTimestamp := fixedTimestamp
```
dkrotx, line 50: Add explanation of what is this. It's not clear what is the page and other variables are stand for. Are they all  (incl.`nonNilToken`) for pagination? If so, do block surrounded with newlines and having a comment - this would really make clear what are they used for.

dkrotx (changes requested):

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+			expectedError: true,              // An error is expected due to the invalid timer
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.setupMock()
+
+			var token pagination.PageToken
+			if tc.name == "Non-nil Pagination Token Passed" {
```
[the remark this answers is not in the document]
3vilhamster: This is a clear example of extracting this case to a separate test.
My take on this that table tests should be very straightforward. If you need to write if inside test body - probably you are mixing different cases in one test.
It will be better/easy readable to group scenarios that require the same/very similar logic, rather than trying to fit all scenarios into a single test.

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+func TestGetUserTimers(t *testing.T) {
+	fixedTimestamp, err := time.Parse(time.RFC3339, "2023-12-12T22:08:41Z")
+	if err != nil {
+		t.Fatalf("Failed to parse timestamp: %v", err)
+	}
+
+	ctrl := gomock.NewController(t)
+
+	mockRetryer := persistence.NewMockRetryer(ctrl)
```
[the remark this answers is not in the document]
agautam478: The gomock.Controller and the mock persistence.Retryer are created within each subtest to ensure isolated mock setups and verifications.

common/reconciliation/fetcher/timer_test.go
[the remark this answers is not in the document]
agautam478: It was cleaner when there were no extra cases but the code is so tightly coupled that if I have to include the tests for 
`Invalid Timer Causes Error`
`Non-nil Pagination Token Passed`

And still maintain the table tests format, I will have to do the validations this way.

common/reconciliation/fetcher/timer_test.go
agautam478: The code doesn't seem to have any custom error struct but I will give this a shot.

common/reconciliation/fetcher/timer_test.go
agautam478: removing it.

common/reconciliation/fetcher/timer_test.go
agautam478: okay! I tried whatever I could.

Thanks for the suggestion.

common/reconciliation/fetcher/timer_test.go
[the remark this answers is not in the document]
agautam478: @3vilhamster Resolving all the other comments. Will take this up at last. I think I will have to think a little bit more to be able to include all scenarios. It's simpler to do with just one case.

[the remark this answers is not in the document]
3vilhamster: Though, I still suggest just creating a separate test for special cases with a comment on why they are special.
It is easier to understand, than if by name or some weird flag in table definition.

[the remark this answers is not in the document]
agautam478: This is a valuable feedback. Thanks. Let's seee! 
I am gonna try writing test for another file now. Apply your suggestion there. if it works then I will come back and update this one too.

common/reconciliation/fetcher/timer_test.go
```diff
@@ -0,0 +1,186 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockRetryer := tc.setupMock(ctrl)
+
+			var token pagination.PageToken
+			if tc.name == "Non-nil Pagination Token Provided" {
+				token = nonNilToken
+			}
```
taylanisikdemir, line 170: why not define a `token` field on test struct and use it from there? we shouldn't have ifs based on test name. `tc` should provide what we need in the test body

taylanisikdemir (approved):

davidporter-id-au (approved):

PR #5480, opened 2023-12-15
Add unit test for cassandra workflow utils part-2

taylanisikdemir:
**What changed?**
Continuing from where I left in #5476.

PR #5506, opened 2023-12-19
Add unit test for cassandra workflow utils part-3

taylanisikdemir:
**What changed?**
Continuing from where I left in #5480

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go
@@ -77,16 +77,23 @@ type fakeBatch struct {
 
 // Query is fake implementation of gocql.Batch.Query
 func (b *fakeBatch) Query(queryTmpl string, args ...interface{}) {
-	argsDereferenced := make([]interface{}, len(args))
+	argsSanitized := make([]interface{}, len(args))
 	for i, arg := range args {
+		// use values instead of pointer so that we can compare them in tests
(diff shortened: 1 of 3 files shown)


common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go
```diff
@@ -1349,6 +1366,1125 @@ func TestUpdateChildExecutionInfos(t *testing.T) {
...
+					TimerID:    "timer1",
+					Version:    1,
+					StartedID:  2,
+					ExpiryTime: ts.UTC(),
+					TaskStatus: 1,
+				},
+			},
+			deleteInfos: []string{"timer2"},
+			wantQueries: []string{
+				`UPDATE executions SET timer_map[ timer1 ] = {` +
```
3vilhamster, line 1459: I'm a bit unsure, but for me it looks like these tests are testing both updateTimerInfos and custom implementation of FakeBatcher.
I'm not sure we can guarantee that this queries are what will be executed by Cassandra, since this is a simplified stub.

taylanisikdemir: That's right and that's actually the scope of unit tests. Basically the function under test takes a `gocql.Batch` interface and calls it with certain parameters. Test is validating this part. What is out of scope for these unit tests is the actual queries generated by the concrete implementation of 3rd party cql package which should be a concern of integration test

3vilhamster (approved):
Overall this is an improval. Though, keeping fakeBatch up to date and mimic Cassandra could be messy.

PR #5510, opened 2023-12-20
Ensure scanner scavenger stops in tests

3vilhamster:
**What changed?**
Ensure that scvngr is stopped if the test scenario times out.

**Why?**
If the test scenario will somehow take more than 10 seconds the test will fail, but can leak goroutine. So I ensure that the suite does the cleanup.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/scanner/tasklist/scavenger_test.go
+++ b/service/worker/scanner/tasklist/scavenger_test.go
@@ -192,6 +192,7 @@ func (s *ScavengerTestSuite) TestAllExpiredTasksWithErrors() {
 
 func (s *ScavengerTestSuite) runScavenger() {
 	s.scvgr.Start()
+	defer s.scvgr.Stop()
 	timer := time.NewTimer(scavengerTestTimeout)
 	select {
 	case <-s.scvgr.stopped:


service/worker/scanner/tasklist/scavenger_test.go
```diff
@@ -192,6 +192,7 @@ func (s *ScavengerTestSuite) TestAllExpiredTasksWithErrors() {
 
 func (s *ScavengerTestSuite) runScavenger() {
 	s.scvgr.Start()
+	defer s.scvgr.Stop()
 	timer := time.NewTimer(scavengerTestTimeout)
 	select {
 	case <-s.scvgr.stopped:
```
Groxx, line 198: main caveat here is probably that, if it's in a live/dead-lock of some kind, it'll sit idle for N minutes until the whole test suite times out...

maybe that's better?  does a timeout give you running goroutines?  and do we have a reasonably-short `go test`-level timeout set, e.g. ideally 1 minute?

Groxx: looks like:
- yep, timeouts print stacks.  should help find the issue when it occurs 👍 
- outside the `host/ndc` folder (which has 20m), we have no specified timeout, which means currently "the whole package can test for 30 minutes".  and a small forever-blocking local test agrees with that.

so:
- 👍 yep I like this
- mind adding a default 1m timeout to all `go test`s in the makefile?

3vilhamster: I'm not sure that we can broadly state that our test will timeout in 1m. I'd say we can try setting to 5m. I'll create a separate PR for this experiment.

Groxx: my manual test took 10m to finish, so that's apparently the default (and all the "the default is 30m" and "the default is 1m" are likely coming from discussion in https://github.com/golang/go/issues/48157, and are not actually the defaults)

Groxx: broadly state: yea.  might not be good enough / a separate PR is fine IMO.
we can also selectively reduce the timeout (but I think not add?) with `t.SetTimeout` which might be good for these

Groxx (approved):

PR #5520, opened 2023-12-21
add unit tests for ValidIDLength

arzonus:
**What changed?**
unit tests have been added

**Why?**
ValidIDLength is not covered

**How did you test it?**
Unit tests are passed

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -238,7 +238,9 @@ func ValidIDLength(
 	idLength := len(id)
 	valid := idLength <= errorLimit
 	if idLength > warnLimit {
-		scope.IncCounter(metricsCounter)
+		if scope != nil {
+			scope.IncCounter(metricsCounter)
+		}
 		if logger != nil {
 			logger.Warn("ID length exceeds limit.",
 				tag.WorkflowDomainName(domainName),
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -26,7 +26,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/uber/cadence/common/log/tag"
 	"math/rand"
 	"strconv"
 	"sync"
@@ -38,6 +37,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/yarpc/yarpcerrors"
 
+	"github.com/uber/cadence/common/log/tag"
 	"github.com/uber/cadence/common/types"
 )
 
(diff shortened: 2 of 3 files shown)


common/util.go
```diff
@@ -238,7 +238,9 @@ func ValidIDLength(
 	idLength := len(id)
 	valid := idLength <= errorLimit
 	if idLength > warnLimit {
-		scope.IncCounter(metricsCounter)
+		if scope != nil {
+			scope.IncCounter(metricsCounter)
```
dkrotx, line 242: use `tally.Noop` instead. We typically assume scope and logger are not nil.

common/util_test.go
```diff
@@ -422,3 +423,14 @@ func TestAwaitWaitGroup(t *testing.T) {
...
 	})
 }
+
+func TestValidIDLength(t *testing.T) {
+	t.Run("valid id length, no warnings", func(t *testing.T) {
+		got := ValidIDLength("12345", nil, 7, 10, 0, "", nil, tag.Tag{})
+		require.True(t, got)
+	})
+	t.Run("non valid id length", func(t *testing.T) {
+		got := ValidIDLength("12345", nil, 1, 4, 0, "", nil, tag.Tag{})
```
dkrotx, line 433: It's better to have a comment (inline is fine as well) on why it is not valid? Because when I look at the ID itself, they are identical in both cases.

arzonus: Fixed it 👍

common/util.go
arzonus: Hmm... But we had the check for `logger`, so I assumed that we should have the same for `scope`. 
It's not obvious how to call `tally.Noop`, because [`NoopScope`](https://github.com/uber/cadence/blob/master/common/metrics/scope.go#L51) requires some `ServiceIdx`.

dkrotx: True, in this case it's more coherent if `scope != nil` check is present.
But it's super unusual - we almost never do that neither for logs nor for metrics.
I meant  just `tally.NoopScope`.

dkrotx (approved):

common/util.go
3vilhamster: Maybe it is worth doing in a separate PR, but I think it could be done here.

arzonus: @3vilhamster I have created another [PR](https://github.com/uber/cadence/pull/5526) with your suggestions 👍

PR #5521, opened 2023-12-21
Adding unit tests to RespondActivityTaskCompleted and RecordActivityT…

sankari165:
**What changed?**
addition of unit tests

**Why?**
RespondActivityTaskCompleted,RespondActivityTaskCompletedByID and RecordActivityTaskHeartbeatByID was not tested

**How did you test it?**
unit tests passed

**Potential risks**

**Release notes**

**Documentation Changes**


service/frontend/workflowHandler_test.go
```diff
@@ -636,6 +636,93 @@ func (s *workflowHandlerSuite) TestRecordActivityTaskHeartbeat_TaskTokenNotSet()
...
+	}
+
+	s.mockDomainCache.EXPECT().GetDomainName(s.testDomainID).Return(s.testDomain, nil)
+	s.mockHistoryClient.EXPECT().RespondActivityTaskCompleted(gomock.Any(),
+		&types.HistoryRespondActivityTaskCompletedRequest{
+			DomainUUID:      taskToken.DomainID,
+			CompleteRequest: req,
+		}).Return(nil)
+
+	err = wh.RespondActivityTaskCompleted(context.Background(), req)
```
dkrotx, line 704: I'm surprised there is no return value or anything? Is there a way to check data is written?
nit: If it's not the point of test, I suggest to add a comment like "we're only checking for successful write here".

service/frontend/workflowHandler_test.go
```diff
@@ -636,6 +636,93 @@ func (s *workflowHandlerSuite) TestRecordActivityTaskHeartbeat_TaskTokenNotSet()
...
 
+func (s *workflowHandlerSuite) TestRecordActivityTaskHeartbeatByID_Success() {
+	wh := s.getWorkflowHandler(s.newConfig(dc.NewInMemoryClient()))
+	req := &types.RecordActivityTaskHeartbeatByIDRequest{
+		Domain:     s.testDomain,
+		WorkflowID: testWorkflowID,
+		RunID:      testRunID,
+		ActivityID: "1",
+		Details:    nil,
+		Identity:   "",
```
dkrotx, line 647: I suggest to only initialise fields with non-default values unless we're already doing the same in other tests of workflowHandler.
It is expected to have Details and Identity not initialised, and you don't usually want to emphasise on this.

3vilhamster (approved):

PR #5522, opened 2023-12-21
Fix race between startup and shutdown in task reader

Groxx:
Nothing ensures all goroutines are started before shutdown, and in tests that's somewhat likely to occur.
Risk to production is probably very low, but definitely best to fix there too.

As a general statement for any readers:
```go
func (s *self) thing() {
  s.wg.Add(1)
  defer s.wg.Done()
  ...
}

// elsewhere
go s.thing()
```
^ this pattern is *almost always* unsafe, at least on a technical level.

All adds need to occur before any Wait() can possibly finish, so they should **always** be added in the "parent" thread, never in the goroutine that things will wait on.

### DIFF
--- a/service/matching/taskReader.go
+++ b/service/matching/taskReader.go
@@ -127,9 +127,17 @@ func newTaskReader(tlMgr *taskListManagerImpl, isolationGroups []string) *taskRe
 func (tr *taskReader) Start() {
 	tr.Signal()
 	for g := range tr.taskBuffers {
-		go tr.dispatchBufferedTasks(g)
+		tr.stopWg.Add(1)
+		go func() {
+			defer tr.stopWg.Done()
+			tr.dispatchBufferedTasks(g)
+		}()
 	}
-	go tr.getTasksPump()
+	tr.stopWg.Add(1)
+	go func() {
+		defer tr.stopWg.Done()
+		tr.getTasksPump()
+	}()
 }
 
 func (tr *taskReader) Stop() {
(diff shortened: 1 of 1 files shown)


service/matching/taskReader.go
```diff
@@ -127,9 +127,18 @@ func newTaskReader(tlMgr *taskListManagerImpl, isolationGroups []string) *taskRe
 func (tr *taskReader) Start() {
 	tr.Signal()
 	for g := range tr.taskBuffers {
-		go tr.dispatchBufferedTasks(g)
+		g := g
+		tr.stopWg.Add(1)
+		go func() {
+			defer tr.stopWg.Done()
+			tr.dispatchBufferedTasks(g)
+		}()
```
Groxx, line 135: there are other ways to achieve this, but tbh I'm fond of "always make concurrency controls visible where it is made concurrent".

alternatively, a `backgroundDispatch(g)` that does this internally (and you do not `go backgroundDispatch(g)`) works pretty well.  that also makes you relatively immune to mutating range vars.

3vilhamster (approved):
Looks weird, but should be good.

Groxx:
> Looks weird, ...

yea.  tbh this is my biggest complaint with Go concurrency: it's "intrusive".  you must always put concurrency in by hand, because the whole system prevents you from building robust abstractions around it like Java has had for many years.

Generics can _largely_ address that now, but there's a massive pile of history that will take a long time to upgrade / will influence habits for probably the next decade :\

PR #5523, opened 2023-12-21
Fix taskReader flaky go routine leak

3vilhamster:
**What changed?**
I've moved waitGroup.Wait outside of guarded path, since even if there is Stop was already called and process was started we need to make sure we wait for all goroutines to complete.

**Why?**
We observer goroutine leaks in tests that cause panics in log messages.

**How did you test it?**
Added a simple unit tests and run integration tests.

**Potential risks**

**Release notes**

**Documentation Changes**


service/matching/task_reader.go
```diff
@@ -150,8 +150,9 @@ func (tr *taskReader) Stop() {
 				tag.Error(err))
 		}
 		tr.taskGC.RunNow(tr.taskAckManager.GetAckLevel())
-		tr.stopWg.Wait()
 	}
+	// There could be multiple calls to Stop() but all of them should block before task reader is completely stopped.
+	tr.stopWg.Wait()
```
Groxx, line 155: I _completely_ agree with this in principle, but this is a rather significant semantic change - as a heads-up for reviewers, it deserves a careful read of all starting and stopping, to make sure it won't misbehave somewhere that stop could be called twice.

service/matching/task_reader_test.go
```diff
@@ -0,0 +1,43 @@
...
+		tl.Stop()
+		close(stoppedCh)
+	}()
+	select {
+	case <-stoppedCh:
+		t.Error("task reader should not be stopped")
+		t.Fail()
+	case waitCh <- struct{}{}:
+		close(waitCh)
+	}
```
Groxx, line 42: typically this would be
```go
tl.cancelFunc = func() {
  close(waitCh)
}

select{
case <-waitCh:
```
which is a bit simpler.  and it strictly avoids "write and close in different locations" which is a BIG land-mine with channels.

afaict this will behave identically though.

3vilhamster: This is not the same. I want to stop inside cancelFunc on first Stop and make sure that the second Stop awaits for the waitGroup to unlock.

service/matching/task_reader.go
3vilhamster: Oh well, but looking into it further, this might be a problem, but it does not explain the leak. This one is called only once. I'll close this and dig more.

PR #5525, opened 2023-12-22
Improve history queue components shutdowns

taylanisikdemir:
**What changed?**
History's timer and transfer queues are not handling shutdowns properly. This PR will make it slightly better but a proper handling of shutdowns will require bigger refactoring to ensure honoring process level shutdown deadline and ordering of Stop calls. We will most likely achieve that by integration with [Fx](https://github.com/uber-go/fx) at some point.

Changes:
- Introduce `QueueProcessorEnableGracefulSyncShutdown` dynamic config. Default is false. When set to true 
   - Synchronously shut down components (get rid of `go Stop()` calls) 
   - Fix ordering of Stop calls so last attempt to complete tasks may actually work (it's always failing today due to misordered calls)

Misc readability changes:
- Get rid of some named loop breaks. Fix buggy breaks associated with this.
- Get rid of some multiline functions

**Why?**
A recent bug shows that there are timer tasks left behind during failover. Handling `Stop()`s properly may prevent that from happening but will dig deeper on potential ack level consistency problems.

**How did you test it?**
Set `QueueProcessorEnableGracefulSyncShutdown: true` in unit tests.

**Potential risks**
Should be minimal but just to be on safe side put the changes behind dynamic config so we can slowly roll this out and rollback if we see issues.

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -176,13 +176,27 @@ func (t *timerQueueProcessor) Stop() {
 		return
 	}
 
+	if !t.shard.GetConfig().QueueProcessorEnableGracefulSyncShutdown() {
+		t.activeQueueProcessor.Stop()
+		for _, standbyQueueProcessor := range t.standbyQueueProcessors {
+			standbyQueueProcessor.Stop()
+		}
+
+		close(t.shutdownChan)
+		common.AwaitWaitGroup(&t.shutdownWG, time.Minute)
+		return
+	}
+
+	// close the shutdown channel first so processor pumps drains tasks
+	// and then stop the processors
+	close(t.shutdownChan)
+	if !common.AwaitWaitGroup(&t.shutdownWG, gracefulShutdownTimeout) {
+		t.logger.Warn("transferQueueProcessor timed out on shut down", tag.LifeCycleStopTimedout)
+	}
 	t.activeQueueProcessor.Stop()
 	for _, standbyQueueProcessor := range t.standbyQueueProcessors {
 		standbyQueueProcessor.Stop()
 	}
-
-	close(t.shutdownChan)
-	common.AwaitWaitGroup(&t.shutdownWG, time.Minute)
 }
 
 func (t *timerQueueProcessor) NotifyNewTask(clusterName string, info *hcommon.NotifyTaskInfo) {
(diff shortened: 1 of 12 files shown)


service/history/task/redispatcher.go
```diff
@@ -160,22 +157,21 @@ func (r *redispatcherImpl) AddTask(
...
+	ntf := redispatchNotification{
 		targetSize: targetSize,
 		doneCh:     doneCh,
-	}:
+	}
+
+	select {
+	case r.redispatchCh <- ntf:
+		// block until the redispatch is done
+		<-doneCh
```
davidporter-id-au, line 170: was the bug that you found that the done channel would be called during shutdown?

service/history/queue/timer_queue_processor.go
```diff
@@ -225,23 +239,37 @@ func (t *timerQueueProcessor) FailoverDomain(domainIDs map[string]struct{}) {
...
 		// other errors should never be returned for GetStateAction
 		panic(fmt.Sprintf("unknown error for GetStateAction: %v", err))
 	}
+
+	var maxReadLevelQueueLevel int
 	for _, queueState := range actionResult.GetStateActionResult.States {
 		queueReadLevel := queueState.ReadLevel().(timerTaskKey).visibilityTimestamp
 		if maxReadLevel.Before(queueReadLevel) {
 			maxReadLevel = queueReadLevel
+			maxReadLevelQueueLevel = queueState.Level()
```
davidporter-id-au, line 265: this seems reasonable, but also is it part of this conceptual change? I wonder if it should be a separate PR?

service/history/queue/timer_queue_processor.go
```diff
@@ -327,9 +361,7 @@ func (t *timerQueueProcessor) completeTimerLoop() {
 	for {
 		select {
 		case <-t.shutdownChan:
-			if err := t.completeTimer(); err != nil {
-				t.logger.Error("Error complete timer task", tag.Error(err))
-			}
+			t.drain()
```
davidporter-id-au, line 364: This needs a flipr guard as well as it's a behaviour change for shutdown

service/history/queue/timer_queue_processor.go
```diff
@@ -338,19 +370,23 @@ func (t *timerQueueProcessor) completeTimerLoop() {
...
 				}
-				backoff := time.Duration(attempt * 100)
-				time.Sleep(backoff * time.Millisecond)
 
 				select {
 				case <-t.shutdownChan:
-					// break the retry loop if shutdown chan is closed
-					break
-				default:
+					t.drain()
```
davidporter-id-au, line 386: do you think this might need a flipr guard too?

davidporter-id-au (approved):

service/history/queue/timer_queue_processor.go
taylanisikdemir: This is a side change to improve debugging of this component. I think it's small enough change to bundle here.

service/history/task/redispatcher.go
taylanisikdemir: not sure if this caused a bug but in previous version of this file line 178 would be called on closed channel during shutdown

service/history/queue/timer_queue_processor.go
taylanisikdemir: I don't think this needs dynamicconfig guard. previous version also attempts to make that drain operation after breaking here. I'm just making it explicit and returning here so the execution doesn't reevaluate the top level select. also getting rid of time.sleep in favor of time.after so it would be more responsive to shutdown signal

PR #5534, opened 2023-12-22
Switch timesource to jonboulle/clockwork

3vilhamster:
**What changed?**
Switched from a mix of time.Source and real time.Sleep usage to jonboulle/clockwork which provides a better way to wrap time related operations with a testable functionality.

As a sidecar:
- Switch tocket bucket implementation from int64 to time.Time/time.Duration. It does not create complexity, but provides a way to use debugger in tests.
- Switch tocket bucket tests from suite to normal Go tests.
- Add a test example that relies on clock.Sleep functionality.

**Why?**
To avoid time related tests in dependant packages and provide an example for time related test solution.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/clock/time_source.go
+++ b/common/clock/time_source.go
@@ -21,51 +21,118 @@
 package clock
 
 import (
-	"sync/atomic"
 	"time"
+
+	"github.com/jonboulle/clockwork"
 )
 
 type (
-	// TimeSource is an interface for any
-	// entity that provides the current
-	// time. Its primarily used to mock
-	// out timesources in unit test
+	// TimeSource provides an interface that packages can use instead of directly using
+	// the [time] module, so that chronology-related behavior can be tested.
 	TimeSource interface {
+		After(d time.Duration) <-chan time.Time
+		Sleep(d time.Duration)
 		Now() time.Time
+		Since(t time.Time) time.Duration
+		NewTicker(d time.Duration) Ticker
+		NewTimer(d time.Duration) Timer
+		AfterFunc(d time.Duration, f func()) Timer
 	}
-	// RealTimeSource serves real wall-clock time
-	RealTimeSource struct{}
 
-	// EventTimeSource serves fake controlled time
-	EventTimeSource struct {
-		now int64
+	// Ticker provides an interface which can be used instead of directly using
+	// [time.Ticker]. The real-time ticker t provides ticks through t.C which
+	// becomes t.Chan() to make this channel requirement definable in this
+	// interface.
+	Ticker interface {
+		Chan() <-chan time.Time
+		Reset(d time.Duration)
+		Stop()
+	}
+
+	// Timer provides an interface which can be used instead of directly using
+	// [time.Timer]. The real-time timer t provides events through t.C which becomes
+	// t.Chan() to make this channel requirement definable in this interface.
+	Timer interface {
+		Chan() <-chan time.Time
+		Reset(d time.Duration) bool
+		Stop() bool
+	}
+
+	// clock serves real wall-clock time
+	clock struct {
+		clockwork.Clock
+	}
+
+	// fakeClock serves fake controlled time
+	fakeClock struct {
+		clockwork.FakeClock
+	}
+
+	// MockedTimeSource provides an interface for a clock which can be manually advanced
+	// through time.
+	//
+	// MockedTimeSource maintains a list of "waiters," which consists of all callers
+	// waiting on the underlying clock (i.e. Tickers and Timers including callers of
+	// Sleep or After). Users can call BlockUntil to block until the clock has an
+	// expected number of waiters.
+	MockedTimeSource interface {
+		TimeSource
+		// Advance advances the FakeClock to a new point in time, ensuring any existing
+		// waiters are notified appropriately before returning.
+		Advance(d time.Duration)
+		// BlockUntil blocks until the FakeClock has the given number of waiters.
+		BlockUntil(waiters int)
 	}
 )
 
+var _ TimeSource = (*clock)(nil)
+var _ TimeSource = (*fakeClock)(nil)
+var _ MockedTimeSource = (*fakeClock)(nil)
+
 // NewRealTimeSource returns a time source that servers
 // real wall clock time
-func NewRealTimeSource() *RealTimeSource {
-	return &RealTimeSource{}
+func NewRealTimeSource() TimeSource {
+	return &clock{
+		Clock: clockwork.NewRealClock(),
+	}
 }
 
-// Now return the real current time
-func (ts *RealTimeSource) Now() time.Time {
-	return time.Now()
-}
-
-// NewEventTimeSource returns a time source that servers
+// NewMockedTimeSource returns a time source that servers
 // fake controlled time
-func NewEventTimeSource() *EventTimeSource {
-	return &EventTimeSource{}
+func NewMockedTimeSource() MockedTimeSource {
+	return &fakeClock{
+		FakeClock: clockwork.NewFakeClock(),
+	}
 }
 
-// Now return the fake current time
-func (ts *EventTimeSource) Now() time.Time {
-	return time.Unix(0, atomic.LoadInt64(&ts.now)).UTC()
+// NewMockedTimeSourceAt returns a time source that servers
+// fake controlled time. The initial time of the MockedTimeSource will be the given time.
+func NewMockedTimeSourceAt(t time.Time) MockedTimeSource {
+	return &fakeClock{
+		FakeClock: clockwork.NewFakeClockAt(t),
+	}
 }
 
-// Update update the fake current time
-func (ts *EventTimeSource) Update(now time.Time) *EventTimeSource {
-	atomic.StoreInt64(&ts.now, now.UnixNano())
-	return ts
+func (r *clock) NewTicker(d time.Duration) Ticker {
+	return r.Clock.NewTicker(d)
+}
+
+func (r *clock) NewTimer(d time.Duration) Timer {
+	return r.Clock.NewTimer(d)
+}
+
+func (r *clock) AfterFunc(d time.Duration, f func()) Timer {
+	return r.Clock.AfterFunc(d, f)
+}
+
+func (c *fakeClock) NewTicker(d time.Duration) Ticker {
+	return c.FakeClock.NewTicker(d)
+}
+
+func (c *fakeClock) NewTimer(d time.Duration) Timer {
+	return c.FakeClock.NewTimer(d)
+}
+
+func (c *fakeClock) AfterFunc(d time.Duration, f func()) Timer {
+	return c.FakeClock.AfterFunc(d, f)
 }
--- a/common/clock/time_source.go
+++ b/common/clock/time_source.go
@@ -81,6 +81,7 @@ type (
 		// waiters are notified appropriately before returning.
 		Advance(d time.Duration)
 		// BlockUntil blocks until the FakeClock has the given number of waiters.
+		// waiters are either time.Sleep, time.After, time.Ticker, or time.Timer.
 		BlockUntil(waiters int)
 	}
 )
--- a/common/clock/time_source.go
+++ b/common/clock/time_source.go
@@ -80,8 +80,11 @@ type (
 		// Advance advances the FakeClock to a new point in time, ensuring any existing
 		// waiters are notified appropriately before returning.
 		Advance(d time.Duration)
-		// BlockUntil blocks until the FakeClock has the given number of waiters.
-		// waiters are either time.Sleep, time.After, time.Ticker, or time.Timer.
+		// BlockUntil blocks until the FakeClock has the given number of waiters
+		// running at the same time.
+		//
+		// Waiters are either time.Sleep, time.After[Func], time.Ticker, or time.Timer,
+		// and they decrement the counter when they complete or are stopped.
 		BlockUntil(waiters int)
 	}
 )
--- a/common/persistence/visibilitySamplingClient.go
+++ b/common/persistence/visibilitySamplingClient.go
@@ -25,7 +25,7 @@ import (
 	"runtime"
 	"sync"
 
-	"github.com/uber/cadence/common/clock"
+	"github.com/benbjohnson/clock"
 	"github.com/uber/cadence/common/dynamicconfig"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
@@ -108,7 +108,7 @@ func (m *domainToBucketMap) getRateLimiter(domain string, numOfPriority, qps int
 		m.Unlock()
 		return rateLimiter
 	}
-	rateLimiter = tokenbucket.NewFullPriorityTokenBucket(numOfPriority, qps, clock.NewRealTimeSource())
+	rateLimiter = tokenbucket.NewFullPriorityTokenBucket(numOfPriority, qps, clock.New())
 	m.mappings[domain] = rateLimiter
 	m.Unlock()
 	return rateLimiter
--- a/common/persistence/visibilitySamplingClient.go
+++ b/common/persistence/visibilitySamplingClient.go
@@ -25,8 +25,7 @@ import (
 	"runtime"
 	"sync"
 
-	"github.com/benbjohnson/clock"
-
+	"github.com/uber/cadence/common/clock"
 	"github.com/uber/cadence/common/dynamicconfig"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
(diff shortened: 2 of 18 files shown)


common/persistence/visibilitySamplingClient.go
```diff
@@ -25,7 +25,8 @@ import (
 	"runtime"
 	"sync"
 
-	"github.com/uber/cadence/common/clock"
```
taylanisikdemir, line 28: Looks like this internal `cadence/common/clock` package is used in many places. I think it's good to have such internal wrapper. If `github.com/benbjohnson/clock` provides better functionality/testing we should switch the underlying implementation of `cadence/common/clock` package so all callers benefit from it. Using a custom clock vendor in one place and leaving others as is can be more confusing than it helps. 
WDYT about sticking with our wrapper and improving it?

Groxx:
I really like this API in most ways (and [facebookgo/clock] it forked from).  Advancing time also auto-firing related timers and channels and whatnot is a _wonderful_ simplification for many conceptual time-uses, and it prevents firing things in an unrealistic order or forgetting to do related things by accident, and the mock-clock-advancing is very like our client library's time simulation and it speeds things up a TON.

The main complication might be that both [facebookgo/clock] and [benbjohnson/clock] are read-only archived repos :|  It does seem simple enough that we could re-implement or fix if necessary, I'm just not sure if there's a "healthy" fork anywhere and it's... somewhat risky in principle to depend on a declared-dead project.  Though the go module proxy should mean we always have access to the contents at least.


[facebookgo/clock]: https://github.com/facebookarchive/clock
[benbjohnson/clock]: https://github.com/benbjohnson/clock

common/persistence/visibilitySamplingClient.go
3vilhamster: Makes sense. Let me try it out.

3vilhamster:
oh, I've seen that [benbjohnson/clock](https://github.com/benbjohnson/clock) was used a lot in the monorepo, but I've tried writing a test around consume, and it is hard (catching that we've started waiting is nearly impossible). So it is hard to use for cases with Timers and Sleep.


Ok. I've tried https://github.com/jonboulle/clockwork which looks much easier to use. See new tests for consume.
For future migration possibilities, I've introduced a wrapper layer, though, it complicates implementation traversal a bit.

taylanisikdemir (approved):

common/clock/time_source.go
```diff
@@ -21,51 +21,118 @@
...
+	// waiting on the underlying clock (i.e. Tickers and Timers including callers of
+	// Sleep or After). Users can call BlockUntil to block until the clock has an
+	// expected number of waiters.
+	MockedTimeSource interface {
+		TimeSource
+		// Advance advances the FakeClock to a new point in time, ensuring any existing
+		// waiters are notified appropriately before returning.
+		Advance(d time.Duration)
+		// BlockUntil blocks until the FakeClock has the given number of waiters.
+		BlockUntil(waiters int)
```
Groxx, line 84: Interesting addition.  Seems useful though it does feel sorta misleading too 🤔 

This is basically the number of _pending clocks_, e.g. things created but not yet completed, which doesn't actually relate to the number of _things waiting_ (except as an upper bound).  Though I don't think there's a way to achieve that with `<-ticker.C` reads.

3vilhamster: Actually, it block on either Sleep/timer or ticker. Everything is registered inside as a waiter in setExpirer function.

3vilhamster: Weird thing, that internally, they updated this function to BlockUntilWithContext, but it is not exposed yet via inteface. Once exposed, we should adopt it.

service/matching/liveness_test.go
```diff
@@ -70,14 +70,14 @@ func (s *livenessSuite) TearDownTest() {
...
 	liveness := newLiveness(s.timeSource, s.ttl, func() { atomic.CompareAndSwapInt32(&s.shutdownFlag, 0, 1) })
-	s.timeSource.Update(time.Now().Add(s.ttl * 2))
+	s.timeSource.Advance(s.ttl * 2)
 	alive := liveness.isAlive()
 	s.False(alive)
 }
 
 func (s *livenessSuite) TestIsAlive_Yes() {
 	liveness := newLiveness(s.timeSource, s.ttl, func() { atomic.CompareAndSwapInt32(&s.shutdownFlag, 0, 1) })
-	s.timeSource.Update(time.Now().Add(s.ttl / 2))
```
davidporter-id-au, line 80: these kinds of changes are going to make things a lot easier to test

common/tokenbucket/tb_test.go
```diff
@@ -167,62 +140,132 @@ func (s *TokenBucketSuite) TestPriorityLowRpsEnforced() {
...
+		consumeFinished := make(chan struct{})
+		go func() {
+			success := tb.Consume(1, 2*refillRate)
+			assert.True(t, success, "Consume returned wrong result")
+			close(consumeFinished)
+		}()
+		ts.BlockUntil(1)
+		select {
+		case <-consumeFinished:
+			assert.Fail(t, "Consume returned before refill")
```
davidporter-id-au, line 221: Nit: It reqiures quite close reading to follow what you're asserting here, or maybe I'm just being dumb. 

You're consuming on line 210 one entry. A second in the goroutine in 214. Why would a rate-limiter with 10 RPS block in a single time interval? I'm sorry, just not entirely following.

davidporter-id-au: ah, ok, the blockuntil(1) waits until the timeout (2 * the refil rate) finishes, so you're checking the timeout occurs ?

go.mod
```diff
@@ -74,7 +77,7 @@ require (
 	github.com/BurntSushi/toml v0.4.1 // indirect
 	github.com/anmitsu/go-shlex v0.0.0-20161002113705-648efa622239 // indirect
 	github.com/apache/thrift v0.13.0 // indirect
-	github.com/benbjohnson/clock v0.0.0-20161215174838-7dc76406b6d3 // indirect
+	github.com/benbjohnson/clock v1.3.5 // indirect
```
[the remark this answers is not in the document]
3vilhamster: I still exists as an indirect dependency. So it is used by some of our depencies. hmm will revert it to the original version.

common/tokenbucket/tb_test.go
3vilhamster: BlockUntil block until at least 1 waiter will be registered. In this case, time.Sleep inside Consume will create this waiter. After it is stopped, I can Advance time, which should unblock time.Sleep and it will run a single iteration of trying to acquire token. So therefore I'm adding backofInterval first to trigger a single loop. And then advancing time further for rifillRate, which will allow to consume 1 token.

I'll add comments around, but that is how tb logic is done. It is weird and is weird to test.

service/matching/liveness_test.go
3vilhamster: Do you mean time.Update is easier? Unfortunately, clockwork does not provide a way to set time. Though, normally, you don't need to set time, you need to advance to a certain point that is required for some logic to happen.

Groxx: no, this is a " :+1: good API" - advancing by more obvious deltas is a lot more readable than mathing out multiple "set time to X" calls

common/tokenbucket/tb_test.go
Groxx: yea, it took me a while to understand what `BlockUntil` did. had to read the source to really figure it out :\

seems useful, just... misleading.

common/clock/time_source.go
```diff
@@ -21,51 +21,119 @@
...
+	// waiting on the underlying clock (i.e. Tickers and Timers including callers of
+	// Sleep or After). Users can call BlockUntil to block until the clock has an
+	// expected number of waiters.
+	MockedTimeSource interface {
+		TimeSource
+		// Advance advances the FakeClock to a new point in time, ensuring any existing
+		// waiters are notified appropriately before returning.
+		Advance(d time.Duration)
+		// BlockUntil blocks until the FakeClock has the given number of waiters.
+		// waiters are either time.Sleep, time.After, time.Ticker, or time.Timer.
```
Groxx, line 84:
```suggestion
		// BlockUntil blocks until the FakeClock has the given number of waiters
		// running at the same time.
		// 
		// Waiters are either time.Sleep, time.After, time.Ticker, or time.Timer,
		// and they decrement the counter when they complete or are stopped.
```

Maybe something like this is clearer?
@davidporter-id-au any opinion?

common/clock/time_source.go
```diff
@@ -21,51 +21,119 @@
...
+		Stop()
+	}
+
+	// Timer provides an interface which can be used instead of directly using
+	// [time.Timer]. The real-time timer t provides events through t.C which becomes
+	// t.Chan() to make this channel requirement definable in this interface.
+	Timer interface {
+		Chan() <-chan time.Time
+		Reset(d time.Duration) bool
+		Stop() bool
```
Groxx, line 58: purely for discussion: the bool return on timers is IMO _mostly_ error-prone, though it has some rare uses.

maybe we should consider wrapping it so there's a `StopAndDrain()` and a `StopWithoutDrain() bool` so it's clear what's happening?

common/clock/time_source.go
```diff
@@ -21,51 +21,119 @@
...
+	}
+
+	// clock serves real wall-clock time
+	clock struct {
+		clockwork.Clock
```
Groxx, line 63: huh.  I never really thought about it, but I suppose `&struct{value}` behaves the same (for `value`) as `struct{*value}` since the wrapping-struct essentially compiles to nothing...

common/tokenbucket/tb.go
```diff
@@ -159,57 +160,57 @@ func (tb *tokenBucketImpl) Consume(count int, timeout time.Duration) bool {
...
 
-		remTime = expiryTime - now
+		remTime = expiryTime.Sub(now)
 	}
 }
 
 func (tb *tokenBucketImpl) reset(rps int) {
 	tb.Lock()
-	tb.fillInterval = int64(time.Millisecond * 100)
+	tb.fillInterval = refillRate
```
Groxx, line 179: huh.  looks like this is never set to any other value - what do you think about simply removing it?

common/tokenbucket/tb.go
```diff
@@ -268,10 +269,10 @@ func NewPriorityTokenBucket(numOfPriority, rps int, timeSource clock.TimeSource)
 	tb := new(priorityTokenBucketImpl)
 	tb.tokens = make([]int, numOfPriority)
 	tb.timeSource = timeSource
-	tb.fillInterval = int64(time.Millisecond * 100)
+	tb.fillInterval = time.Millisecond * 100
```
Groxx, line 272: actually, yea - probably remove or use the same const.

common/tokenbucket/tb_test.go
```diff
@@ -167,62 +140,159 @@ func (s *TokenBucketSuite) TestPriorityLowRpsEnforced() {
...
+		default:
+		}
+		// Advance time to the next refill time.
+		ts.Advance(refillRate - backoffInterval + 1)
+		// Advance should unblock Consume and it should return afterwards.
+		// I don't provide a timeout, since I have a goroutine to panic on timeout.
+		select {
+		case <-consumeFinished:
+		}
```
Groxx, line 251:
```suggestion
		<-consumeFinished
```
may also be worth mentioning that the goroutine-panic is there to dump the goroutines, which includes the test (waiting here) and the goroutine trying to consume with a timeout.  (because a timeout would have to manually dump goroutines, and that seems like more work than it's worth)

common/tokenbucket/tb_test.go
```diff
@@ -167,62 +140,159 @@ func (s *TokenBucketSuite) TestPriorityLowRpsEnforced() {
...
+			assert.Fail(t, "Consume returned before refill")
+		default:
+		}
+
+		// advance time to wake Consume up and make it return before refill.
+		ts.Advance(backoffInterval/2 + 1)
+		select {
+		case <-consumeFinished:
+		default:
+		}
```
Groxx, line 289: this doesn't assert anything, and races with the goroutine's close.

probably want the same thing as the other test?
```go
<-consumeFinished
```

common/tokenbucket/tb_test.go
```diff
@@ -167,62 +140,159 @@ func (s *TokenBucketSuite) TestPriorityLowRpsEnforced() {
...
+	})
+}
+
+func panicOnTimeout(d time.Duration) {
+	select {
+	case <-time.After(d):
+		panic("timeout")
+	}
```
Groxx, line 297: I doubt it'll be a problem any time soon that these are simply abandoned forever, but personally I'll vote:
```go
func panicOnTimeout(d time.Duration) (cancel func()) {
  stop := make(chan struct{})
  go func() {
    select {
    case <-time.After(d): panic("timeout")
    case <-stop: // noop
    }
  }()
  return func() { close(stop) }
}

// use
defer panicOnTimeout(time.Minute)()
```

service/history/task/timer_active_task_executor_test.go
```diff
@@ -530,7 +530,7 @@ func (s *timerActiveTaskExecutorSuite) TestProcessActivityTimeout_RetryPolicy_Re
 	s.mockExecutionMgr.On("GetWorkflowExecution", mock.Anything, mock.Anything).Return(&persistence.GetWorkflowExecutionResponse{State: persistenceMutableState}, nil)
 	s.mockExecutionMgr.On("UpdateWorkflowExecution", mock.Anything, mock.Anything).Return(&persistence.UpdateWorkflowExecutionResponse{MutableStateUpdateSessionStats: &persistence.MutableStateUpdateSessionStats{}}, nil).Once()
 
-	s.timeSource.Update(s.now.Add(2 * timerTimeout))
```
Groxx, line 533: well that's an unpleasant bug waiting to happen - `now` isn't updated, so calling this 2x wouldn't act like it looked.

🎉 clockwork 🎉

service/matching/liveness_test.go
```diff
@@ -118,7 +117,7 @@ func (s *livenessSuite) TestEventLoop_Shutdown() {
 	liveness := newLiveness(s.timeSource, s.ttl, func() { atomic.CompareAndSwapInt32(&s.shutdownFlag, 0, 1) })
 	liveness.Start()
 
-	s.timeSource.Update(time.Now().Add(s.ttl * 4))
+	s.timeSource.Advance(s.ttl * 4)
 	<-liveness.shutdownChan
 	s.Equal(int32(1), atomic.LoadInt32(&s.shutdownFlag))
```
Groxx, line 122: hm :\  this is a logical race, since the chan is closed before the shutdown func is called.

not related to this PR / no need to fix here tho.

Groxx (approved):
Exclusively minor notes, LGTM and thank you!

common/tokenbucket/tb_test.go
3vilhamster: Sounds good. Updating a function, might be later on we can  put it to shared folder for other race like tests.

common/tokenbucket/tb.go
3vilhamster: makes sense, removed property.

common/tokenbucket/tb_test.go
3vilhamster: Good catch, lost this while refactoring.

common/clock/time_source.go
3vilhamster: Not sure what do you mean, this is a pure Go details, which should be common. https://pkg.go.dev/time#Timer.Stop

common/tokenbucket/tb_test.go
```diff
@@ -167,62 +140,159 @@ func (s *TokenBucketSuite) TestPriorityLowRpsEnforced() {
...
+		}
+		// Advance time to the next refill time.
+		ts.Advance(refillRate - backoffInterval + 1)
+		// Advance should unblock Consume and it should return afterwards.
+		// I don't provide a timeout, since I have a goroutine to panic on timeout.
+		<-consumeFinished
+	})
+	t.Run("timeout", func(t *testing.T) {
+		// make sure we don't deadlock inside the test.
+		go panicOnTimeout(time.Minute)()
```
Groxx, line 253: or more simply `defer panicOnTimeout(time.Minute)()`.
though this will also cancel the pending-panic so I guess it works too

common/tokenbucket/tb.go
Groxx: looks like this got lost btw.  (not really a complaint, just a FYI)

3vilhamster: hmm I've removed the property, so now only const refillRate is used.

PR #5535, opened 2023-12-22
Adding logging to scanner.go

agautam478:
**What changed?**
No functionality changes but added logs to the scanner.go file for better observability.

**Why?**
Due to lack of logging it's extremely hard to figure out where, why or when something broke in the scanner. Since, it's already such a complicated system, more logging will hep us debug issues faster

**How did you test it?**
Yes, tested locally. 

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/worker/scanner/scanner.go
+++ b/service/worker/scanner/scanner.go
@@ -111,6 +111,7 @@ func New(
 	if err != nil {
 		resource.GetLogger().Fatal("failed to initialize zap logger", tag.Error(err))
 	}
+	zapLogger.Info("Initializing new scanner")
 	return &Scanner{
 		context: scannerContext{
 			resource: resource,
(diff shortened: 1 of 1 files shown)


service/worker/scanner/scanner.go
```diff
@@ -111,6 +111,7 @@ func New(
 	if err != nil {
 		resource.GetLogger().Fatal("failed to initialize zap logger", tag.Error(err))
 	}
+	zapLogger.Info("Initializing new scanner")
```
taylanisikdemir, line 114: nit: most other places in the repo uses the internal `log.Logger` which wraps zap.

service/worker/scanner/scanner.go
```diff
@@ -243,8 +247,14 @@ func (s *Scanner) startWorkflow(
 	cancel()
 
 	if cadence.IsWorkflowExecutionAlreadyStartedError(err) {
+		s.zapLogger.Error("Workflow had already started", zap.String("workflowType", workflowType), zap.Error(err))
```
taylanisikdemir, line 250: better to use predefined logs tags instead of repeating same tag key 
- `tag.WorkflowType`
- `tag.WorkflowTaskListName`
- we don't have predefined one for "workerTaskListNames" so it's probably fine to define it inline in this file

service/worker/scanner/scanner.go
```diff
@@ -243,8 +247,14 @@ func (s *Scanner) startWorkflow(
...
 		return nil
 	}
+	if err != nil {
+		s.zapLogger.Error("Failed to start workflow", zap.String("workflowType", workflowType), zap.Error(err))
+	} else {
+		s.zapLogger.Info("Workflow started", zap.String("workflowType", workflowType))
+	}
```
taylanisikdemir, line 257: nit formatting:
```
if err != nil {
   log error
   return err 
}
log info
return nil
```

taylanisikdemir (approved):

service/worker/scanner/scanner.go
```diff
@@ -111,6 +111,7 @@ func New(
 	if err != nil {
```
3vilhamster, line 111: nit: we need to propagate and use log wrapper here.

PR #5544, opened 2023-12-27
Synchronous shutdown support for replication task fetcher

taylanisikdemir:
**What changed?**
Replication task fetcher doesn't wait for full shutdown before returning from `Stop() and this causes integration test flakyness: 
```
panic: Log in goroutine after TestIntegrationSuite
...
cadence/service/history/replication/task_fetcher.go:250
...
```

To address this introduced a dynamic config `ReplicationTaskFetcherEnableGracefulSyncShutdown` and changed the behavior of replication task fetcher if the config is turned on. 
I will roll this out per cluster, similarly with the recently introduced `QueueProcessorEnableGracefulSyncShutdown`.

**Why?**
Better lifecycle management -> less test flakiness and less unknowns during restarts.

**How did you test it?**
unit and integration tests

### DIFF
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1602,8 +1602,14 @@ const (
 	// KeyName: history.queueProcessorEnableGracefulSyncShutdown
 	// Value type: Bool
 	// Default value: false
-	// Allowed filters: ShardID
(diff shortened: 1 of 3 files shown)


3vilhamster (approved):
Looking forward to have this enabled by default

PR #5555, opened 2023-12-27
Added logs into the current execution.go and a unit test

agautam478:
**What changed?**
Added some minor logs into the current_execution.go and also added a unit test.

**Why?**
as a part of the FIXIT week to improve coverage.

**How did you test it?**
locally

**Potential risks**
NA

**Release notes**

**Documentation Changes**


service/worker/scanner/executions/current_execution.go
```diff
@@ -99,6 +107,8 @@ func currentExecutionScannerManager(
 	params shardscanner.ScanShardActivityParams,
 	domainCache cache.DomainCache,
 ) invariant.Manager {
+	logger := zap.L()
```
taylanisikdemir, line 110: let's avoid using global logger. we don't do that anywhere else and it is better to keep it that way. instead you can pass logger as parameter

taylanisikdemir (approved):

PR #5560, opened 2023-12-28
Rework persistenceMetricClients to generated approach

3vilhamster:
**What changed?**
I've reworked persistenceMetricClients into files that are generated automatically. This ensures unified approach to all persistence metered method and eliminates manual job to keep them in sync. Also, all metrics/logs are now follow the same logic.
All managers except for ExecutionManager follow same logic with some minor logical branches. ExecutionManager is very different, so I've extracted it in a separate template.

In the next PR I'll remove old clients and cleanup the logic.

**Why?**
To remove manual work to keep wrappers in sync with persistence managers implementation and unify approach to metering and logging.

**How did you test it?**
I've tested that I emit the same metrics/logs between old and new logic.

**Potential risks**

**Release notes**

**Documentation Changes**


common/metrics/defs.go
```diff
@@ -201,8 +201,8 @@ const (
 	PersistenceCompleteTimerTaskScope
 	// PersistenceRangeCompleteTimerTaskScope tracks CompleteTimerTasks calls made by service to persistence layer
 	PersistenceRangeCompleteTimerTaskScope
-	// PersistenceCreateTaskScope tracks CreateTask calls made by service to persistence layer
-	PersistenceCreateTaskScope
+	// PersistenceCreateTasksScope tracks CreateTask calls made by service to persistence layer
```
davidporter-id-au, line 204: Not necessarily a problem, but what was the thinking behind this name change?

common/persistence/dataManagerInterfaces.go
```diff
@@ -1845,6 +1856,8 @@ type (
 		Payload   []byte    `json:"message_payload"`
 	}
 
+	QueueMessageList []*QueueMessage
```
davidporter-id-au, line 1859: Is this change metrics related? I would consider pulling it out separately?

go.mod
```diff
@@ -53,11 +53,11 @@ require (
 	go.uber.org/thriftrw v1.29.2
 	go.uber.org/yarpc v1.70.3
 	go.uber.org/zap v1.13.0
-	golang.org/x/net v0.17.0
+	golang.org/x/net v0.19.0
```
davidporter-id-au, line 56: should this be in a separate PR?

common/persistence/wrappers/templates/metered_execution.tmpl
```diff
@@ -0,0 +1,75 @@
...
+                    logTags := append([]tag.Tag{tag.WorkflowDomainName(domainName)}, getCustomLogTags({{$reqName}})...)
+                    c.logger.SampleInfo("Persistence CreateWorkflowExecution called", c.sampleLoggingRate(), logTags...)
+                	if c.enableShardIDMetrics() {
+                	    err = c.callWithDomainAndShardScope({{$scopeName}}, op, metrics.DomainTag(domainName),
+                	    metrics.ShardIDTag(strconv.Itoa(c.GetShardID())))
+                	} else {
+                		err = c.call({{$scopeName}}, op, metrics.DomainTag(domainName))
+                	}
+                }
+            {{ else -}}
```
davidporter-id-au, line 65: is an early return possible rather than having the else branch?

common/persistence/wrappers/templates/metered.tmpl
```diff
@@ -0,0 +1,65 @@
+import (
```
davidporter-id-au, line 1: Please include a header specifying they're generated else we'll definately have folks editing the generated code (including me probably)

common/metrics/defs.go
3vilhamster: Unifying the approach of variable naming/variable to method mappings. To be able to generate the code, there should be a simple rule to determine the scope. I've changed some names and added a couple of scopes, keeping them mapped to the same operation in the metric results. But still ended up with one ugly.
```
{{ if and (eq $interfaceName "VisibilityManager") (eq $methodName "DeleteWorkflowExecution") -}}
                {{ $scopeName = "metrics.PersistenceVisibilityDeleteWorkflowExecutionScope" }}
            {{ end -}}
```

common/persistence/dataManagerInterfaces.go
3vilhamster: It is metric related. I need a method defined on the response, and that could only be done on the type.

common/persistence/wrappers/templates/metered_execution.tmpl
3vilhamster: Unfortunately, templates are really hard to read/format with continue/break statements.

common/persistence/dataManagerInterfaces.go
davidporter-id-au: ah, yeah, sorry, need more coffee, totally makes sense

common/persistence/wrappers/templates/metered.tmpl
3vilhamster: Gowrap adds a header:
```
// Code generated by gowrap. DO NOT EDIT.
// template: ../templates/metered.tmpl
// gowrap: http://github.com/hexdigest/gowrap
```

go.mod
3vilhamster: That comes from
```
go get golang.org/x/exp
```
Not sure how to separate them apart. All x/ packages are linked to one another.

davidporter-id-au: ah, ok I didn't see `exp` being added

common/metrics/defs.go
```diff
@@ -267,8 +267,8 @@ const (
 	PersistenceGetClosedWorkflowExecutionScope
 	// PersistenceVisibilityDeleteWorkflowExecutionScope is the metrics scope for persistence.VisibilityManager.DeleteWorkflowExecution
 	PersistenceVisibilityDeleteWorkflowExecutionScope
-	// PersistenceVisibilityDeleteUninitializedWorkflowExecutionScope tracks DeleteUninitializedWorkflowExecution calls made by service to persistence layer
-	PersistenceVisibilityDeleteUninitializedWorkflowExecutionScope
+	// PersistenceDeleteUninitializedWorkflowExecutionScope tracks DeleteUninitializedWorkflowExecution calls made by service to persistence layer
+	PersistenceDeleteUninitializedWorkflowExecutionScope
```
davidporter-id-au, line 271: I think you'll need to keep the word `visibility` on this one? This is applying to different table then the normal executions table. It's for when CAAS us being used to list workflows instead of elasticsearch (we have visibility table for this)

common/persistence/wrappers/metered/base.go
```diff
@@ -0,0 +1,262 @@
...
+type base struct {
+	metricClient                  metrics.Client
+	logger                        log.Logger
+	enableLatencyHistogramMetrics bool
+	sampleLoggingRate             dynamicconfig.IntPropertyFn
+	enableShardIDMetrics          dynamicconfig.BoolPropertyFn
+}
+
+func (p *base) updateErrorMetricPerDomain(scope int, err error, scopeWithDomainTag metrics.Scope) {
+	switch err.(type) {
```
davidporter-id-au, line 45: Not needed in this diff, but we'll need to change this to use `errors.Is/As` at some point. Your change will make this a lot easier though.

common/persistence/wrappers/templates/metered_execution.tmpl
```diff
@@ -0,0 +1,75 @@
...
+        		return err
+        	}
+
+        	{{$scopeName := printf "metrics.Persistence%sScope" $methodName}}
+
+            {{ if gt (len $method.Params) 1 -}}
+                {{ $reqName := (index $method.Params 1).Name }}
+                if domainName, hasDomainName := getDomainNameFromRequest({{$reqName}}); hasDomainName {
+                    logTags := append([]tag.Tag{tag.WorkflowDomainName(domainName)}, getCustomLogTags({{$reqName}})...)
+                    c.logger.SampleInfo("Persistence CreateWorkflowExecution called", c.sampleLoggingRate(), logTags...)
```
davidporter-id-au, line 57: I am not sure the log sample is correct?

common/metrics/defs.go
davidporter-id-au: totally makes sense. It took me a little while to understand, but I can see how the concistency makes it better for templating

go.mod
```diff
@@ -68,6 +68,7 @@ require (
 require (
 	github.com/google/go-cmp v0.5.9
 	github.com/jonboulle/clockwork v0.4.0
+	golang.org/x/exp v0.0.0-20231226003508-02704c960a9b
```
davidporter-id-au, line 71: just because I'm being dumb, what was this required for?

common/metrics/defs.go
3vilhamster: It is used only inside VisibiltyManager, so Visibility prefix is not required. I'm thinking of separating this metrics by manager so it would be easier to track which component emits this, but for now it should be fine.

common/persistence/wrappers/templates/metered_execution.tmpl
3vilhamster: Good catch! Will fix.

3vilhamster: hmm I guess I need to adjust the test and make sure that logs are covered.

go.mod
3vilhamster: yes. It has a nice helper of maps.Keys

common/persistence/wrappers/templates/metered_execution.tmpl
3vilhamster: Fixed and added tests

common/persistence/wrappers/metered/base.go
```diff
@@ -0,0 +1,262 @@
...
+		metricsScope.IncCounter(metrics.PersistenceFailures)
+	}
+}
+
+func (p *base) call(scope int, op func() error, tags ...metrics.Tag) error {
+	metricsScope := p.metricClient.Scope(scope, tags...)
+	if len(tags) > 0 {
+		metricsScope.IncCounter(metrics.PersistenceRequestsPerDomain)
+	} else {
+		metricsScope.IncCounter(metrics.PersistenceRequests)
```
taylanisikdemir, line 117: Not related to this change but I don't like this pattern of defining multiple counters for same thing. For example I want to introduce shard-id tag and it doesn't make sense to create a new counter `PersistenceRequestsPerDomainPerShard` :) Ideally we could have a single counter for persistence latency with all the dimensions in it and when a dimension is not applicable it can just be empty.

common/persistence/dataManagerInterfaces.go
```diff
@@ -38,6 +38,17 @@
 //go:generate gowrap gen -g -p . -i DomainManager -t ./wrappers/templates/errorinjector.tmpl -o wrappers/errorinjectors/domain.go
 //go:generate gowrap gen -g -p . -i QueueManager -t ./wrappers/templates/errorinjector.tmpl -o wrappers/errorinjectors/queue.go
 
+// Generate metered wrappers.
```
taylanisikdemir, line 41: thanks for tackling this

taylanisikdemir (approved):

PR #5562, opened 2023-12-28
Ensure testlogger samples all logs

3vilhamster:
**What changed?**
Updated testlogger wrapper to ensure that all log messages are sampled.
To achieve this I've exposed a new consturctor of loggerimpl that passes sampleFn and ensured that default implementation stayed the same.

**Why?**
This is required in wrappers changes that use logging.

**How did you test it?**
Unit tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/log/loggerimpl/logger.go
+++ b/common/log/loggerimpl/logger.go
@@ -45,6 +45,8 @@ const (
 	defaultMsgForEmpty = "none"
 )
 
+var defaultSampleFn = func(i int) bool { return rand.Intn(i) == 0 }
+
 // NewNopLogger returns a no-op logger
 func NewNopLogger() log.Logger {
 	return NewLogger(zap.NewNop())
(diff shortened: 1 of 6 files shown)


davidporter-id-au:
No concerns, but what's the intent?

3vilhamster:
Intent to have an ability to test/verify sampled logs

common/log/loggerimpl/logger.go
```diff
@@ -62,9 +61,17 @@ func NewDevelopment() (log.Logger, error) {
 
 // NewLogger returns a new logger
 func NewLogger(zapLogger *zap.Logger) log.Logger {
+	return NewLoggerWithSampleFunc(zapLogger, func(i int) bool {
+		return rand.Intn(i) == 0
+	})
+}
+
+// NewLoggerWithSampleFunc returns a new logger with sample function.
+func NewLoggerWithSampleFunc(zapLogger *zap.Logger, sampleLocalFn func(int) bool) log.Logger {
```
taylanisikdemir, line 70: new func means new code to cover. instead of creating separate constructor we could pass optional params to existing one. that would require making `loggerImpl` public though. up to you
```
type LoggerOption func(*LoggerImpl)

func WithSamplingFn(fn func(int) bool) LoggerOption {
  return func(l *LoggerImpl) {
    l.sampleLocalFn = fn
 }
}

func NewLogger(zapLogger *zap.Logger, opts ...LoggerOption) log.Logger {
  l := &LoggerImpl{...}
  for _, o := range opts {
     o(l)
  }
  return l
}
```

taylanisikdemir (approved):

common/log/loggerimpl/logger.go
3vilhamster: Good call. Though it is not required to make LoggerImpl public. It is usually the case when you want to limit which options are provided. I've switched to options and, since we are talking about tests, reworked tests and improved coverage.

PR #5563, opened 2023-12-28
Toggling off cross-cluster polling by default

davidporter-id-au:
**What changed?**

This toggles off the cross-cluster background daemons which do a fair amount of IO to the database on startup and sometimes interfere badly with persistence layer rate-limiting during high load and outages. 

**How did you test it?**

Tested locally with docker

**Potential risks**

**Release notes**

For any users who're using the cross-cluster feature (I'm not aware of any), they will also need to include the following dynamic config to enable this feature now. 

`history.enableCrossClusterEngine` to be true

**Documentation Changes**

### DIFF
--- a/service/history/task/cross_cluster_task_processor.go
+++ b/service/history/task/cross_cluster_task_processor.go
@@ -45,6 +45,7 @@ const (
 type (
 	// CrossClusterTaskProcessorOptions configures crossClusterTaskProcessor
 	CrossClusterTaskProcessorOptions struct {
+		Enabled                    dynamicconfig.BoolPropertyFn
 		MaxPendingTasks            dynamicconfig.IntPropertyFn
 		TaskMaxRetryCount          dynamicconfig.IntPropertyFn
 		TaskRedispatchInterval     dynamicconfig.DurationPropertyFn
@@ -89,15 +90,19 @@ func NewCrossClusterTaskProcessors(
 	options *CrossClusterTaskProcessorOptions,
 ) common.Daemon {
 	processors := make(crossClusterTaskProcessors, 0, len(taskFetchers))
-	for _, fetcher := range taskFetchers {
-		processor := newCrossClusterTaskProcessor(
-			shard,
-			taskProcessor,
-			fetcher,
-			options,
-		)
-		processors = append(processors, processor)
+
+	if options.Enabled() {
+		for _, fetcher := range taskFetchers {
+			processor := newCrossClusterTaskProcessor(
+				shard,
+				taskProcessor,
+				fetcher,
+				options,
+			)
+			processors = append(processors, processor)
+		}
 	}
+
 	return processors
 }
 
--- a/common/dynamicconfig/constants.go
+++ b/common/dynamicconfig/constants.go
@@ -1646,12 +1646,19 @@ const (
 	// Default value: false
 	// Allowed filters: DomainName
 	EnableConsistentQueryByDomain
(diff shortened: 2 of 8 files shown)


service/history/task/cross_cluster_task_processor.go
```diff
@@ -45,6 +45,7 @@ const (
 type (
 	// CrossClusterTaskProcessorOptions configures crossClusterTaskProcessor
 	CrossClusterTaskProcessorOptions struct {
+		Enabled                    dynamicconfig.BoolPropertyFn
```
3vilhamster, line 48: I think this property should not be part of dynamic config. If you set this property to true, the service won't immediately start processing cross cluster tasks.
Application of this config will required restart/redeploy.

Also, restart of the service can crash/fail if the config was enabled outside of the service deployment.

service/history/task/cross_cluster_task_processor.go
```diff
@@ -89,15 +90,19 @@ func NewCrossClusterTaskProcessors(
...
 	processors := make(crossClusterTaskProcessors, 0, len(taskFetchers))
-	for _, fetcher := range taskFetchers {
-		processor := newCrossClusterTaskProcessor(
-			shard,
-			taskProcessor,
-			fetcher,
-			options,
-		)
-		processors = append(processors, processor)
+
```
3vilhamster, line 93: Maybe this check could be pushed higher by the stack. There is no point of creating CrossClusterTaskProcessors. It is misleading that you call a function that does nothing.

service/history/task/cross_cluster_task_processor.go
davidporter-id-au: Open to changing it to be a plain static config value, but this is not significantly different to `history.crossClusterFetcherParallelism` which is currently being used to reduce the overhead of this feature in prod. Both of these values require restart.

service/history/task/cross_cluster_task_processor.go
davidporter-id-au: I considered that, open to changing it, it was just following the established pattern in the history setup where all the functions that are being created on startup are quite simple and don't have if/else toggles in the megafunction `NewEngineWithShardContext`. I don't think it significantly matters either way

3vilhamster: I guess it is ok. Though, I've looked into NewEngineWithShardContext and that is a mess of a function :(

service/history/task/cross_cluster_task_processor.go
3vilhamster: I guess it is worth discussing on what we want to settle on. I'm not sure that this approach is good, maybe it is worth updating before cross cluster has mass adoption?

Also, I would propose to unify crosscluster configuration at least logically in the code. I see that we have crossClusterProcessor and crossClusterTaskProcessors. One is configured and optional, another is not configured and not optional. Can they be unified somehow and started only if they are required?

davidporter-id-au: At this point I'm not super sure we won't just deprecate this feature as it's not really had wide utility and the load cost is pretty high. I would be happy to do such refactoring, but would want to also put a lot more energy into looking into a more lightweight replication mechanism rather than shard polling (pull-based), as this is n^2 number of polls required

service/history/task/cross_cluster_task_processor.go
davidporter-id-au: haha yeah :/ there's worse sadly too (frontend handler :/ )

3vilhamster (approved):
The change makes sense.

PR #5566, opened 2023-12-28
Small refactoring in history task processors

taylanisikdemir:
**What changed?**
- Task scheduler option structs are moved into separate files under common/task folder
- Unit tests are added to cover concrete scheduler option creation 
- Debug logs are added to include task scheduling options details
- Some constants are moved to local package
- Some files renamed to snake_case

**Why?**
To improve debugging and better code organization

**How did you test it?**
`make test`


3vilhamster (approved):
LGTM

service/history/task/priority_assigner.go
```diff
@@ -39,20 +39,16 @@ var (
...
-		currentClusterName string
-		domainCache        cache.DomainCache
-		config             *config.Config
-		logger             log.Logger
-		scope              metrics.Scope
-		rateLimiters       *quotas.Collection
-	}
-)
-
-var _ PriorityAssigner = (*priorityAssignerImpl)(nil)
```
davidporter-id-au, line 55: Any reason to remove? I admit they don't do a lot, but compile-time checks are nice?

common/task/weighted_round_robin_task_scheduler.go
```diff
@@ -224,7 +213,13 @@ func (w *weightedRoundRobinTaskSchedulerImpl) dispatcher() {
 		w.updateTaskChs(taskChs)
 		weights := w.getWeights()
 		for priority, taskCh := range taskChs {
-			for i := 0; i < weights[priority]; i++ {
+			count, ok := weights[priority]
+			if !ok {
+				w.logger.Error("weights not found for task priority", tag.Dynamic("priority", priority), tag.Dynamic("weigths", weights))
```
davidporter-id-au, line 218: nit: `weights`

common/task/weighted_round_robin_task_scheduler.go
```diff
@@ -238,16 +233,14 @@ func (w *weightedRoundRobinTaskSchedulerImpl) dispatcher() {
 					return
 				default:
 					// if no task, don't block. Skip to next priority
-					break
+					break Submit_Loop
```
davidporter-id-au, line 236: I don't quite follow the reason for this addition?

service/history/task/priority_assigner.go
3vilhamster: There is no need in them, since NewPriorityAssigner returns an interface and compile time check happens there. This is only useful if you have completely separate struct, but you want to make sure that it is in sync with interface.

common/task/weighted_round_robin_task_scheduler.go
taylanisikdemir: `break` here was breaking the `select` block but the intend was to break the loop.
Before: https://go.dev/play/p/kZoHQxnesgs
After:  https://go.dev/play/p/VBJWxNSCj7D

service/history/task/priority_assigner.go
taylanisikdemir: yes `func NewPriorityAssigner(..) PriorityAssigner` already ensures this at compile time. 

Also there's a whole another debate to have about client side interfaces in Go. Instead of declaring interface and implementing it in the same package, the client (caller side) defines the interface it expects. https://medium.com/@mbinjamil/using-interfaces-in-go-the-right-way-99384bc69d39. We can discuss pros/cons of that another time. This pattern is not used in our repo at all.

common/task/weighted_round_robin_task_scheduler.go
taylanisikdemir: thanks fixed.

PR #5568, opened 2023-12-29
Switch all retryable client to gowrap

3vilhamster:
**What changed?**
Switched all retryable clients into a single folder and using a single template.

There are no logical changes, these files were generated, but for unknown reason were not marked as generated.

Later on I want to follow the same pattern for all clients wrappers to improve code coverage stat on clients and eliminate manual changes that could be caused by adding new methods to clients.

**Why?**
All generated code does not require code coverage and could be simplified.

**How did you test it?**
Added unit test for a single method of a single client. That should cover all.

**Potential risks**

**Release notes**

**Documentation Changes**


3vilhamster:
wrong branch

PR #5569, opened 2023-12-29
Switch all retryable client to gowrap

3vilhamster:
**What changed?**
All client retryable wrappers are now generated with gowrap and covered with unit test.

**Why?**
Improve code coverage and eliminate manual changes if new methods are added to clients.

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


client/wrappers/retryable/wrappers_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/uber/cadence/client/frontend"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/types"
+)
+
```
taylanisikdemir, line 36: would be nice to cover the new generated clients here as you did in common/persistence/wrappers/ratelimited/wrappers_test.go

3vilhamster: My idea is that there is no special logic in methods, so there is no point of testing every single method. Ive tested one, the rest are the same.
Persistent clients are harder, since it is more complicated and has extra logs/metrics. Though, tests there also could be cleaned up.

taylanisikdemir: Yeah agree there's not too much value in testing all methods. I wonder if we should put `gen/` somewhere in the path for generated files so we an easily exclude these from coverage reports for good.

3vilhamster: Usually, all go tooling respects generated headers (which gowrap writes)
https://github.com/golang/lint/issues/239
Coveralls respects it: https://coveralls.io/github/uber/cadence (.gen folder is not counted in lines to cover).
My assumption is that sonarqube also does, but might be we need to manually set it up.

3vilhamster: We manually set an exceptions:
https://github.com/uber/cadence/blob/master/Makefile#L601
So we can define a simple rule to separate generated files from the rest.

taylanisikdemir: +1 for `_generated.go` suffix unless the files are under .gen folder already. Ping me if you decide to go with that and I will update the internal coverage exclusions accordingly

Groxx: fwiw I'm kinda fond of this sort of structure:
```
dir/
  - thing_gen.go  // only trivially-correct codegen
  - thing.go      // anything that "completes" ^_gen, e.g. helper funcs
  - thing_test.go // tests of thing.go
```
_gen vs _generated: no preference.  I see _gen more but both are clear and that's all I really care about.

for code coverage and IDE highlighting purposes, many things look for a `// @generated` comment, which these don't seem to have and might be worth adding in general.  I'm not sure if we have anything that expects that or if it'd be easier to just filter them out in the makefile by filename suffix tho.

client/wrappers/retryable/admin.go
```diff
@@ -0,0 +1,366 @@
...
+
+func (c *adminClient) CountDLQMessages(ctx context.Context, cp1 *types.CountDLQMessagesRequest, p1 ...yarpc.CallOption) (cp2 *types.CountDLQMessagesResponse, err error) {
+	var resp *types.CountDLQMessagesResponse
+	op := func() error {
+		var err error
+		resp, err = c.client.CountDLQMessages(ctx, cp1, p1...)
+		return err
+	}
+	err = c.throttleRetry.Do(ctx, op)
+	return resp, err
```
Groxx, line 78: entirely unrelated to this PR and not worth changing here, but we do have generics now.... this could be pretty simply reduced to
```go
return c.throttleRetry.Do1(ctx, c.client.CountDLQMessages, cp1, p1...)
// and the void-return ones would be
return c.ThrottleRetry.Do(ctx, c.client.CloseShard, cp1, p1...)
```

client/templates/retry.tmpl
```diff
@@ -7,10 +7,8 @@ import (
 	"github.com/uber/cadence/common/types"
 )
 
-{{ $decorator := (printf "retryable%s" .Interface.Name) }}
-{{ $Decorator := (printf "Retryable%s" .Interface.Name) }}
-
-var _ {{.Interface.Name}} = (*{{$decorator}})(nil)
```
Groxx, line 13: possibly worth bringing back.
it's just a duplicated check done elsewhere, but it does help point out the specific issue, since it can sometimes be opaque which inferred type is referring to what and why it doesn't match at location X.

Groxx (approved):
sgtm and seems pretty trivially a no-op / everything should work.  I like the new `retryable.NewThingClient` too, nice and descriptive.

re gen vs generated vs etc: a suffix seems best to me, but that's about all.  and generally I prefer specially-named _files_, not _folders_, because that lets you add non-generated supporting files in the same package and they won't be accidentally excluded from stuff.

client/wrappers/retryable/wrappers_test.go
3vilhamster: hmm I'm not sure about @generated comment. 
I think, IDE/Linters looks for standard Go comment 
```
^// Code generated .* DO NOT EDIT\.$
```
https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source

client/templates/retry.tmpl
3vilhamster: There is a constructor that returns an interface that will point to the struct not matching the interface. So this is duplicate check, imo.

PR #5570, opened 2023-12-29
Add unit test for new Kafka client

neil-xie:
**What changed?**
Add unit test for new Kafka client
Rename clientImpl.go in snake case

**Why?**

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/messaging/kafka/client_impl_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+func TestNewKafkaClient(t *testing.T) {
+	metricsClient := metrics.NewClient(tally.NoopScope, metrics.History)
+	logger := testlogger.New(t)
+	testCases := []struct {
+		name        string
+		config      *config.KafkaConfig
+		checkApp    bool
+		hasErr      bool
+		expectedErr string
```
taylanisikdemir, line 44: nit: you don't need `hasErr` since you set `expectedErr` for those cases anyway

common/messaging/kafka/client_impl_test.go
```diff
@@ -0,0 +1,159 @@
...
+					},
+				},
+			},
+			checkApp: true,
+			hasErr:   false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if !tc.hasErr {
```
taylanisikdemir, line 141: you can get rid of this if/else and always apply the defer func. within defer func you can check if an err is expected

taylanisikdemir (approved):

common/messaging/kafka/client_impl_test.go
neil-xie: Thanks for catching this, updated!

PR #5572, opened 2024-01-01
Straightforwardly fixes a few minor copy bugs and adds a small fuzz util

davidporter-id-au:
#### What changed?

- Adds the library `github.com/google/gofuzz`. 
- Adds a small wrapper for some repo-level conventions we may wish to follow by default - notably determinitism and logging the seed value
- Adds some probably fairly low value but still easy to test copy functions as a demonstration of value. 

#### Justification & intent

The problem I am trying to solve here is: How can we ensure that testdata is kept up-to-date and in-sync with the changing functions. 

A canonical example is idl mapping, where a mapper for Protobuf generated type and it's `common/types` equivalent need to be kept in sync as the fields evolve and additional fields and substructs are added. Added test-data will continue to pass tests, but may silently omit new fields or not handle nil values safely as assumptions about what data must be present correctly. 

Adding a 'fuzz' or generator for data makes such tests relatively easy: A round-trip test for a serialiser with generated code will easily spot missed values and generated code with occasional nils will ensure safe nil handling before deep-reaching into subfields. 

#### Comparison vs alternatives

This is certainly not the only way to solve this problem: There are several obvious other ways to solve this: 

a) straightforward reflection on existing testdata to ensure all fields are present with some kind of linter (rejected: Requires inventing a kind of DSL to specify optional fields in lots of scenarios where structs only sometimes exist, significant maintenance overhead)
b)  The [native golang security fuzzer](https://go.dev/doc/security/fuzz/) available on `testing.F` (rejected because [more adverserial](https://news.ycombinator.com/item?id=30385558) than the problem I am trying to solve and computationally quite expensive to have it build a corpus of known inputs. 
c)  property based testing such as the stdlib `testing/quick` (rejected because the API is quite rigid and it's virtually impossible to use it to create nested substructures without adherence to the `Generate()` api and seems to completely not work with protobuf generated types (with their private fields)


service/history/execution/mutable_state_util.go
```diff
@@ -437,24 +441,30 @@ func CopyTimerInfo(sourceInfo *persistence.TimerInfo) *persistence.TimerInfo {
 // CopyCancellationInfo copies RequestCancelInfo
 func CopyCancellationInfo(sourceInfo *persistence.RequestCancelInfo) *persistence.RequestCancelInfo {
 	return &persistence.RequestCancelInfo{
-		Version:         sourceInfo.Version,
-		InitiatedID:     sourceInfo.InitiatedID,
-		CancelRequestID: sourceInfo.CancelRequestID,
+		Version:               sourceInfo.Version,
+		InitiatedID:           sourceInfo.InitiatedID,
+		InitiatedEventBatchID: sourceInfo.InitiatedEventBatchID,
```
davidporter-id-au, line 446: appears to have been missing. no idea waht the field means, probably minor

service/history/execution/mutable_state_util.go
```diff
@@ -437,24 +441,30 @@ func CopyTimerInfo(sourceInfo *persistence.TimerInfo) *persistence.TimerInfo {
...
-		InitiatedID:     sourceInfo.InitiatedID,
-		SignalRequestID: sourceInfo.SignalRequestID,
-		SignalName:      sourceInfo.SignalName,
-	}
-	result.Input = make([]byte, len(sourceInfo.Input))
-	copy(result.Input, sourceInfo.Input)
-	result.Control = make([]byte, len(sourceInfo.Control))
-	copy(result.Control, sourceInfo.Control)
+		Version:               sourceInfo.Version,
+		InitiatedEventBatchID: sourceInfo.InitiatedEventBatchID,
```
davidporter-id-au, line 455: appears to have been missing. no idea waht the field means, probably minor

service/history/execution/mutable_state_util.go
```diff
@@ -328,6 +327,7 @@ func CopyWorkflowExecutionInfo(sourceInfo *persistence.WorkflowExecutionInfo) *p
 		ParentDomainID:                     sourceInfo.ParentDomainID,
 		ParentWorkflowID:                   sourceInfo.ParentWorkflowID,
 		ParentRunID:                        sourceInfo.ParentRunID,
+		IsCron:                             sourceInfo.IsCron,
```
davidporter-id-au, line 330: missing I guess?

service/history/execution/mutable_state_util.go
```diff
@@ -354,6 +354,7 @@ func CopyWorkflowExecutionInfo(sourceInfo *persistence.WorkflowExecutionInfo) *p
 		DecisionRequestID:                  sourceInfo.DecisionRequestID,
 		DecisionTimeout:                    sourceInfo.DecisionTimeout,
 		DecisionAttempt:                    sourceInfo.DecisionAttempt,
+		DecisionScheduledTimestamp:         sourceInfo.DecisionScheduledTimestamp,
```
davidporter-id-au, line 357: seems like it's missing? I guess this is dead code mostly

service/history/historyEngine2_test.go
```diff
@@ -299,6 +300,7 @@ func (s *engine2Suite) TestRecordDecisionTaskStartedSuccessStickyEnabled() {
 	s.NotNil(response)
 	expectedResponse.StartedTimestamp = response.StartedTimestamp
 	expectedResponse.ScheduledTimestamp = common.Int64Ptr(0)
+	response.ScheduledTimestamp = common.Int64Ptr(0)
```
davidporter-id-au, line 303: I realise this has a new clock construct in it, I just didn't want to refactor a huge amount of this suite test

service/history/execution/mutable_state_util.go
```diff
@@ -381,8 +383,11 @@ func CopyWorkflowExecutionInfo(sourceInfo *persistence.WorkflowExecutionInfo) *p
 
 // CopyActivityInfo copies ActivityInfo
 func CopyActivityInfo(sourceInfo *persistence.ActivityInfo) *persistence.ActivityInfo {
-	details := make([]byte, len(sourceInfo.Details))
-	copy(details, sourceInfo.Details)
+	var details []byte
+	if sourceInfo.Details != nil {
+		details = make([]byte, len(sourceInfo.Details))
+		copy(details, sourceInfo.Details)
+	}
```
3vilhamster, line 390: nit: could be a one liner
```suggestion
details := append(make([]int, 0, len(sourceInfo.Details)), sourceInfo.Details...)
```

service/history/execution/mutable_state_util.go
```diff
@@ -437,24 +442,30 @@ func CopyTimerInfo(sourceInfo *persistence.TimerInfo) *persistence.TimerInfo {
...
-	copy(result.Input, sourceInfo.Input)
-	result.Control = make([]byte, len(sourceInfo.Control))
-	copy(result.Control, sourceInfo.Control)
+		Version:               sourceInfo.Version,
+		InitiatedEventBatchID: sourceInfo.InitiatedEventBatchID,
+		InitiatedID:           sourceInfo.InitiatedID,
+		SignalRequestID:       sourceInfo.SignalRequestID,
+		SignalName:            sourceInfo.SignalName,
+	}
+	if sourceInfo.Input != nil {
```
3vilhamster, line 461: nit: same suggestion as above

service/history/execution/mutable_state_util.go
```diff
@@ -437,24 +442,30 @@ func CopyTimerInfo(sourceInfo *persistence.TimerInfo) *persistence.TimerInfo {
...
+		result.Input = make([]byte, len(sourceInfo.Input))
+		copy(result.Input, sourceInfo.Input)
+	}
+	if sourceInfo.Control != nil {
```
taylanisikdemir, line 465: I checked the `Control` field a few weeks ago to see why/where it is used and couldn't find anything explanatory. Do you know why we carry over this Control field?

service/history/execution/mutable_state_util_test.go
```diff
@@ -24,13 +24,100 @@ import (
...
 	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/persistence"
+	"github.com/uber/cadence/common/testing/testdatagen"
 	"github.com/uber/cadence/common/types"
 )
 
+func TestCopyActivityInfo(t *testing.T) {
+	for i := 0; i < 100; i++ {
+		t.Run("test CopyActivityInfo Mapping", func(t *testing.T) {
+			f := fuzz.New().Funcs(testdatagen.GenHistoryEvent).NilChance(0.2)
```
taylanisikdemir, line 40: to preserve repeatability of tests let's use the fuzzer with fixed seed. `NewWithSeed`

common/testing/testdatagen/persistence.go
```diff
@@ -0,0 +1,46 @@
...
+func GenHistoryEvent(o *types.HistoryEvent, c fuzz.Continue) {
+	t1 := int64(1704127933)
+	e1 := types.EventType(1)
+	*o = types.HistoryEvent{
+		ID:        1,
+		Timestamp: &t1,
+		EventType: &e1,
+		Version:   2,
+		TaskID:    3,
+		// todo (david.porter) Find a better way to handle substructs
```
taylanisikdemir, line 43: the substructs are important to cover so until we find a better way maybe reuse from here https://github.com/uber/cadence/blob/db34d46ac892a0f8a79a5a5ebd57e599ae4b4350/common/types/testdata/history.go#L34

davidporter-id-au: Yeah, I was acutally going to do that in a follow up with a generator function and I think that's good feedback here. Let me do that here.

service/history/execution/mutable_state_util.go
davidporter-id-au: I have zero idea

service/history/execution/mutable_state_util_test.go
davidporter-id-au: 👍  will do

service/history/execution/mutable_state_util.go
Groxx: tbh we should probably just lean on x/exp/slices whenever possible: https://pkg.go.dev/golang.org/x/exp/slices#Clone

common/testing/testdatagen/persistence.go
```diff
@@ -0,0 +1,46 @@
...
+	fuzz "github.com/google/gofuzz"
+
+	"github.com/uber/cadence/common/types"
+)
+
+// GenHistoryEvent is a function to use with gofuzz which
+// skips the majority of difficult to generate values
+// for the sake of simplicity in testing. Use it with the fuzz.Funcs(...) generation function
+func GenHistoryEvent(o *types.HistoryEvent, c fuzz.Continue) {
+	t1 := int64(1704127933)
```
Groxx, line 35: maybe something more understandable at a glance? 
```suggestion
	t1 := MustParse(DateOnly, "2024-01-01").Unix()
```

service/history/execution/mutable_state_util.go
Groxx: I tried to figure it out a while ago too :|
I think the answer is somewhere in git history, but I didn't find anything understandable.

service/history/execution/mutable_state_util_test.go
Groxx: I may vote for:
- use `time.Now()` as a seed
- `t.Log` that seed so it can be recreated

fuzz tests kinda _should_ keep changing, to hit new edge cases you didn't expect.  otherwise you're just invisibly hardcoding an unknown set of assumptions, hidden behind `rand()`.

taylanisikdemir: I like that idea of changing seed but at least have it in the logs so we can reproduce

service/history/execution/mutable_state_util_test.go
```diff
@@ -24,13 +24,100 @@ import (
...
+			f.Fuzz(&d1)
+			d2 := CopyActivityInfo(&d1)
+
+			assert.Equal(t, &d1, d2)
+		})
+	}
+}
+
+func TestCopyWorkflowExecutionInfo(t *testing.T) {
+	for i := 0; i < 100; i++ {
```
Groxx, line 52: hmmm.  obviously this works (hopefully does different values each time too?), but I'm not sure we gain anything from N tests vs N loops in one test... neither is easy to short-circuit or narrow down when reproducing something.

service/history/execution/mutable_state_util.go
```diff
@@ -328,6 +328,7 @@ func CopyWorkflowExecutionInfo(sourceInfo *persistence.WorkflowExecutionInfo) *p
 		ParentDomainID:                     sourceInfo.ParentDomainID,
 		ParentWorkflowID:                   sourceInfo.ParentWorkflowID,
 		ParentRunID:                        sourceInfo.ParentRunID,
+		IsCron:                             sourceInfo.IsCron,
```
Groxx, line 331: bleh.  I really need to try to finish a "this struct must be completely filled out" linter...

taylanisikdemir: linter or unit test? 
one option could be to write tests using cmp.Diff and pass source object fully populated. to make it future proof, test can also validate all the public fields in the source object tree is non-nil.

davidporter-id-au: That's fundamentally what I'm looking to enable with the generator here - the filling out completely thing.

taylanisikdemir (approved):

PR #5576, opened 2024-01-04
Fix multiple lock acquire on membership update

3vilhamster:
**What changed?**
Membership ring on refresh need to update the ring members information. This was done one by one, even though it could be done in 1 operation.
I've switched to a single operation that does a single lock acquire on update.
No logical changes.

**Why?**
To decrease chances of seeing errors like "Failed to send listener notification, chanel is full".

**How did you test it?**
unit tests.

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/membership/hashring.go
+++ b/common/membership/hashring.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/dgryski/go-farm"
 	"github.com/uber/ringpop-go/hashring"
+	"github.com/uber/ringpop-go/membership"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/log"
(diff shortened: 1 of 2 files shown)


common/membership/hashring.go
```diff
@@ -292,3 +292,11 @@ func (r *ring) compareMembers(members []HostInfo) (map[string]HostInfo, bool) {
...
 	return newMembersMap, changed
 }
+
+func castToMembers[T membership.Member](members []T) []membership.Member {
+	result := make([]membership.Member, 0, len(members))
+	for _, h := range members {
+		result = append(result, h)
+	}
+	return result
+}
```
Groxx, line 302: go generics make this such a weird-looking operation :|

any idea how often we do this kind of "list of things compatible with T" -> "[]T list" conversion?  seems like it might be worth making a common helper for it...

Groxx (approved):
👍 yep, seems like a straightforward improvement, no underlying semantic changes or any other possible excitement.  Thanks!

taylanisikdemir (approved):

common/membership/hashring.go
Groxx: hmmm.  seems regrettably not possible with go's generics 🤔.  would work fine if we could have it compile-time checked only on types used, not on all possible types, but we don't have that option...

PR #5577, opened 2024-01-04
Add unit test for history/shard context

gazi-yestemirova:
**What changed?**
Add unit tests for Update & Get cluster replication level and Transfer & Timer Failover levels

**Why?**
To improve code coverage

**How did you test it?**
Unit tests 

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context_test.go
+++ b/service/history/shard/context_test.go
@@ -25,12 +25,12 @@ package shard
 import (
 	"context"
 	"errors"
-	"github.com/pborman/uuid"
 	"math"
 	"testing"
 	"time"
 
 	"github.com/golang/mock/gomock"
+	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


service/history/shard/context_test.go
```diff
@@ -117,6 +118,8 @@ func (s *contextTestSuite) newContext() *contextImpl {
 		timerMaxReadLevelMap:      make(map[string]time.Time),
 		remoteClusterCurrentTime:  make(map[string]time.Time),
 		eventsCache:               eventsCache,
+		transferFailoverLevels:    make(map[string]TransferFailoverLevel),
```
3vilhamster, line 121: I propose creating a constructor that fills all required fields and use it in the places of usages and here. It will be easier to keep track of all required fields.

gazi-yestemirova: Is `newContext()` not the constructor? 
Or you mean a separate method to create just `contextImpl` ?

davidporter-id-au (approved):

PR #5578, opened 2024-01-05
Add unit tests for workflow state non maps

Shaddoll:
**What changed?**
Add unit tests for workflow state non maps

**Why?**
to improve test coverage

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/sql/workflow_state_non_maps_test.go
```diff
@@ -0,0 +1,180 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package sql
+
+import (
+	"context"
+    "errors"
```
taylanisikdemir, line 26: indentation

taylanisikdemir (approved):

common/persistence/sql/workflow_state_non_maps_test.go
```diff
@@ -0,0 +1,180 @@
...
+                mockTx.EXPECT().IsNotFoundError(err).Return(true)
+            },
+            wantErr: true,
+        },
+    }
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
3vilhamster, line 167: Finish is no longer required. Controller will be cleaned up with t.Cleanup.
See https://pkg.go.dev/github.com/golang/mock/gomock#Controller.Finish

common/persistence/sql/workflow_state_non_maps_test.go
```diff
@@ -0,0 +1,180 @@
...
```
3vilhamster, line 109: Finish is no longer required. Controller will be cleaned up with t.Cleanup.
See https://pkg.go.dev/github.com/golang/mock/gomock#Controller.Finish

common/persistence/sql/workflow_state_non_maps_test.go
```diff
@@ -0,0 +1,180 @@
...
+        },
+        {
+            name: "Error case - failed to delete",
+            signalRequestedIDs: []string{"s1", "s2"},
+            deleteSignalRequestIDs: []string{"s3", "s4"},
+            mockSetup: func(mockTx *sqlplugin.MockTx) {
+                err := errors.New("some error")
+                mockTx.EXPECT().InsertIntoSignalsRequestedSets(gomock.Any(), gomock.Any()).Return(nil, nil)
+                mockTx.EXPECT().DeleteFromSignalsRequestedSets(gomock.Any(), gomock.Any()).Return(nil, err)
+                mockTx.EXPECT().IsNotFoundError(err).Return(true)
```
3vilhamster, line 100: For the NotFound error I think we can not only verify that error is returned, but also verify that error is of the expected type.

I suggest to either switch wantErr to a type check, or separate the error handling to a dedicated test that focuses on error handling.

Shaddoll: The check is done in a helper function which is called everywhere. I'll write a unit test for that helper function to cover the check.

PR #5580, opened 2024-01-05
Added deprecated domain check to the taskvalidator

agautam478:
**What changed?**
- Added a check if the workflow's domain exists or if it's not deprecated before sending it for the stale workflow check.
- Currently the check has no associated action for safety purposes. The plan is to call a delete on such workflows after each validation step.

**Why?**
- These changes have been made in order to add validation in the task processing workflows to ensure that they are not corrupted.

**How did you test it?**
Added a partner unit test.

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


common/taskvalidator/validateworkflow.go
```diff
@@ -49,14 +56,34 @@ func NewWfChecker(logger log.Logger, metrics metrics.Client) Checker {
...
 }
+
+func (w *checkerImpl) deprecatedDomainCheck(domainID string, domainName string) error {
+	domain, err := w.dc.GetDomainByID(domainID)
+	if err != nil {
+		w.logger.Error("Error getting domain by ID", tag.Error(err))
+		return err
+	}
+	if domain.GetInfo().Status == deprecatedDomainStatus || domain.GetInfo().Status == deletedDomainStatus {
+		w.logger.Error("The workflow domain doesn't exist", tag.WorkflowDomainID(domainID),
```
taylanisikdemir, line 83: this doesn't seem like an error scenario. did you mean to add a debug log here?

common/taskvalidator/validateworkflow.go
```diff
@@ -49,14 +56,34 @@ func NewWfChecker(logger log.Logger, metrics metrics.Client) Checker {
...
+	}
 	w.metricsClient.Scope(metrics.TaskValidatorScope, metrics.DomainTag(domainName)).IncCounter(metrics.ValidatedWorkflowCount)
 	return nil
 }
+
+func (w *checkerImpl) deprecatedDomainCheck(domainID string, domainName string) error {
+	domain, err := w.dc.GetDomainByID(domainID)
+	if err != nil {
+		w.logger.Error("Error getting domain by ID", tag.Error(err))
+		return err
```
taylanisikdemir, line 80: in general it is better to avoid logging the error and also returning it. ideally somewhere close to the top of the call chain does the error logging. e.g. a middleware in an rpc chain.
if you need additional information to be logged then you can wrap the error and return

common/taskvalidator/validateworkflow_test.go
```diff
@@ -25,34 +25,58 @@ package taskvalidator
...
+		name         string
+		domainID     string
+		domainStatus int
+		expectError  bool
+	}{
+		{
+			name:         "ActiveDomain",
+			domainID:     "activeDomainID",
+			domainStatus: 0, // Assuming 0 represents an active domain
+			expectError:  false,
```
taylanisikdemir, line 48: let's cover error case too

common/taskvalidator/validateworkflow.go
```diff
@@ -38,6 +44,7 @@ type Checker interface {
 type checkerImpl struct {
 	logger        log.Logger
 	metricsClient metrics.Client
+	dc            cache.DomainCache
```
taylanisikdemir, line 47: this needs to be initialized in constructor

common/cache/domainCache.go
```diff
@@ -127,7 +127,7 @@ type (
 	// DomainCacheEntry contains the info and config for a domain
 	DomainCacheEntry struct {
 		mu                          sync.RWMutex
-		info                        *persistence.DomainInfo
+		Info                        *persistence.DomainInfo
```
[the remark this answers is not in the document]
Groxx: this and: if it's just for testing purposes outside this package, there are a few constructor-funcs like `NewDomainCacheEntryForTest` already

allenchen2244 (approved):

common/taskvalidator/validateworkflow_test.go
```diff
@@ -25,34 +25,43 @@ package taskvalidator
...
-	// Create a mock logger and metrics client
-	logger := log.NewNoop()
-	metricsClient := metrics.NewNoopMetricsClient()
+	testCases := []struct {
+		name     string
+		domainID string
+		runID    string
+	}{
+		{"DomainFetchSuccess", "domain-id-success", "run-id-success"},
+		{"DomainFetchFailure", "domain-id-failure", "run-id-failure"},
```
taylanisikdemir, line 44: how are these 2 cases different? validations expect no error below

common/constants.go
```diff
@@ -264,3 +264,8 @@ const (
 	// DefaultHistoryMaxAutoResetPoints is the default maximum number for auto reset points
 	DefaultHistoryMaxAutoResetPoints = 20
 )
+
+const (
+	DeprecatedDomainStatus = 1
+	DeletedDomainStatus    = 2
+)
```
taylanisikdemir, line 271: these are still copies of original enum. see my prev comment https://github.com/uber/cadence/pull/5580#discussion_r1443319705

taylanisikdemir: @Shaddoll what is the pattern we follow for defining a corresponding type in common that is originally defined in persistence package only?

common/taskvalidator/validateworkflow_test.go
agautam478: The main code currently doesn't do anything except for logging the case. The test is for a supporting domain id retrieval check as there's not much to test really. Will be modifying it with something more substantial in the upcoming PR.

common/constants.go
agautam478: I tried importing the original enum but it didn't work out so ended up defining these in a common file.

taylanisikdemir: @agautam478 why didn't it work when imported from persistence? cyclic dependency?
We need to figure out a solution than re-defining these because there's no mechanism ensuring these don't go out of sync

PR #5588, opened 2024-01-12
Update access controlled frontend handlers to generated code

Shaddoll:
**What changed?**
- Introduced a template to generate access controlled handlers
Behavior changes:
- Add auth for all admin handlers, previously there are some methods not covered
- Remove domain tag from auth metrics of Register/Describe/Update/Deprecate APIs, because these APIs are not domain-level APIs. They are APIs for metadata.

**Why?**
To eliminate manual work.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


service/frontend/templates/accesscontrolled.tmpl
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/uber/cadence/common/config"
+	"github.com/uber/cadence/common/log/tag"
+    {{- if eq .Interface.Name "Handler"}}
+    "github.com/uber/cadence/common/metrics"
+    {{end}}
+	"github.com/uber/cadence/common/resource"
+	"github.com/uber/cadence/common/types"
+)
+
+{{$permissionMap := dict "CountWorkflowExecutions" "PermissionRead"}}
```
3vilhamster, line 14: Wow, I didn't know you can do that in template :)

service/frontend/adminHandler.go
```diff
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 //go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination adminHandler_mock.go -package frontend github.com/uber/cadence/service/frontend AdminHandler
+//go:generate gowrap gen -g -p . -i AdminHandler -t ./templates/accesscontrolled.tmpl -o ./access_controlled_admin_gen.go -v handler=AccessControlled
```
3vilhamster, line 22: I propose to separate code from generated code. 
Also, naming convention for generated files is _generated.go https://github.com/uber/cadence/pull/5574
I don't think it is set in stone, but coverage wise, I think we should be consistent

Shaddoll: Originally, I put them in a separate package but realized that introduced a cyclic dependency.

taylanisikdemir: It should be fine that generated files are in same folder but let's follow same suffix convention "_generated.go"

taylanisikdemir (approved):

service/frontend/adminHandler.go
Shaddoll: I need to do another refactoring to remove the cyclic dependency.

PR #5589, opened 2024-01-12
Switch to cassandra 4.1.1 image because 4.1.3 stopped working on some linux distros

taylanisikdemir:
**What changed?**
`cassandra:4.1.3` was recently updated and it doesn't work on linux devpods. 
```
% docker run cassandra:4.1.3
Cassandra 4.0 requires either Java 8 (update 151 or newer) or Java 11 (or newer).
```

**Why?**
Switching to 4.1.1 to keep the default experience working on all platforms.

**How did you test it?**
`make test` on mac and linux. CI checks.

### DIFF
--- a/docker/buildkite/docker-compose-es7.yml
+++ b/docker/buildkite/docker-compose-es7.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   cassandra:
-    image: cassandra:4.1.3
+    image: cassandra:4.1.1
     environment:
       - "MAX_HEAP_SIZE=256M"
       - "HEAP_NEWSIZE=128M"
--- a/docker/buildkite/docker-compose-local-es7.yml
+++ b/docker/buildkite/docker-compose-local-es7.yml
@@ -2,7 +2,7 @@ version: "3.5"
 
 services:
   cassandra:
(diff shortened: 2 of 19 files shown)


davidporter-id-au (approved):

Groxx (approved):
Yea - I've been trying to figure out if this is a "real" problem or if it's some host-OS/image combination issue unique to our infra... but no luck, and no apparent sign that others are having issues nor that a fix is coming.  So we probably do have to mitigate like this by hand.

Downgrade works 👍.
Do you know if Cassandra will choke on old data with this change and need erasing, or will it be able to handle 4.1.3-produced records?

taylanisikdemir:
> Yea - I've been trying to figure out if this is a "real" problem or if it's some host-OS/image combination issue unique to our infra... but no luck, and no apparent sign that others are having issues nor that a fix is coming. So we probably do have to mitigate like this by hand.
> 
> Downgrade works 👍. Do you know if Cassandra will choke on old data with this change and need erasing, or will it be able to handle 4.1.3-produced records?

Didn't try that scenario but I don't think we care. This docker compose setup should only be used for local ephemeral testing purposes.

PR #5590, opened 2024-01-13
Refactor frontend to split the code into multiple packages

Shaddoll:
**What changed?**
- Move admin handler, api handler and generated access controlled handler into separate packages
- Rename some interface for consistency
- Rename some file for consistency

**Why?**
One of the reason is to move the generated code to a separate package, but previously that introduced a circular dependency, because the generated code has a dependency on the interface defined in frontend package.
And the frontend package also depends on the generated code.
This also makes the code cleaner and make future refactoring easier.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/frontend/admin/adminGrpcHandler.go
```diff
@@ -29,174 +29,174 @@ import (
...
 }
 
-func newAdminGRPCHandler(h AdminHandler) adminGRPCHandler {
-	return adminGRPCHandler{h}
+func NewAdminGRPCHandler(h Handler) AdminGRPCHandler {
+	return AdminGRPCHandler{h}
 }
 
-func (g adminGRPCHandler) register(dispatcher *yarpc.Dispatcher) {
+func (g AdminGRPCHandler) Register(dispatcher *yarpc.Dispatcher) {
```
davidporter-id-au, line 40: I am guessing this needs to be publically exposed because it's moved into a subpackage?

davidporter-id-au (commented):
Lgtm. I think there's probably a good case for splitting the admin handler into machine APIs and human-admin api calls, but that's a much bigger split for another time.

davidporter-id-au (approved):

service/frontend/admin/adminGrpcHandler.go
Shaddoll: Yes.

PR #5592, opened 2024-01-15
Replace JWT validation library

mantas-sidlauskas:
**What changed?**
Replacing  https://github.com/cristalhq/jwt with https://github.com/golang-jwt/jwt.

**Why?**
To utilize "keyfunc" option for implementing [JWKS](https://datatracker.ietf.org/doc/html/rfc7517) functionality. It is planed to check token and pick correct public key for token validation. Having this, it is possible to use existing self-signed JWT for internal inter-node communication and 3rd party Identity Provider at the same time.

**How did you test it?**
Additional unit tests for validate TTL, local tests.

**Potential risks**

**Release notes**

**Documentation Changes**


go.mod
```diff
@@ -66,6 +66,7 @@ require (
 )
 
 require (
+	github.com/golang-jwt/jwt/v5 v5.2.0
```
taylanisikdemir, line 69: we have v5.0.0 of this package in monorepo. Since it doesn't have other dependencies it maybe OK but please confirm @Groxx.

common/authorization/oauthAuthorizer.go
```diff
@@ -101,49 +104,68 @@ func (a *oauthAuthority) Authorize(
...
 
 func (a *oauthAuthority) validateTTL(claims *JWTClaims) error {
-	if claims.TTL > a.authorizationCfg.MaxJwtTTL {
-		return fmt.Errorf("token TTL: %d is larger than MaxTTL allowed: %d", claims.TTL, a.authorizationCfg.MaxJwtTTL)
+
+	if claims.IssuedAt == nil {
+		return errors.New("IssuedAt is not set")
+	}
+
+	// Fill ExpiresAt when TTL is passed
```
davidporter-id-au, line 150: I'm a little surprised this is necessary, it's not a problem, but I wouldj have throught that the JWT lib would do this natively

go.mod
davidporter-id-au: D12725271

common/authorization/oauthAuthorizer.go
mantas-sidlauskas: Our clients (java and go) are not setting `ExpiresAt` claim. Instead, they are providing TTL claim. This is to support backwards compatibility. Later on, clients needs to be updated to fill in `ExpiresAt` and we can delegate this check to JWT lib.

go.mod
davidporter-id-au: I think it'll be ok

common/authorization/oauthAuthorizer.go
mantas-sidlauskas: After re-reading your comment, I realized that it was not about TTL check, but for `IssuedAt` check. And yes, you are right, this is not needed as library will do this validation. Thanks!

davidporter-id-au: ah, sorry, yeah, i wasn't being super clear and to be honest, i get confused between the different values. I'll don't think it's a blocker, so stamping. Thank you for looking into it.

davidporter-id-au (approved):

common/authorization/oauthAuthorizer.go
```diff
@@ -101,49 +104,63 @@ func (a *oauthAuthority) Authorize(
...
-	if claims.Iat+claims.TTL < time.Now().Unix() {
-		return errors.New("JWT has expired")
+
+	exp, err := claims.GetExpirationTime()
+
+	if err != nil || exp == nil {
+		return errors.New("ExpiresAt is not set")
 	}
+
+	timeLeft := exp.Unix() - time.Now().Unix()
```
davidporter-id-au, line 156: my comment wasn't great, I actually meant this: I thought this would be something that the library would do natively

PR #5594, opened 2024-01-16
Add cache for workflow specific in memory data

jakobht:
**What changed?**
Adding a struct for storing in memory workflow specific data. It will first be used for storing workflow specific rate limiters, so we can rate limit too many requests to the same workflowID.

In the future it could store e.g. the current state of the workflow (running, terminated), query answers etc. 

**Why?**
This is part of the general workflow specific rate limits project

**How did you test it?**
Unit tests

**Potential risks**
No risk, just adding the struct

**Release notes**

**Documentation Changes**


service/history/workflowcache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+// WFCache is a per workflow cache used for workflow specific in memory data
+type WFCache interface {
+	AllowExternal(domainID string, workflowID string) bool
+	AllowInternal(domainID string, workflowID string) bool
+}
+
+type wfCache struct {
+	muc                    sync.Mutex
```
taylanisikdemir, line 40: nit: we usually use `mu` for mutex variable names

service/history/workflowcache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+		}),
+		externalLimiterFactory: params.ExternalLimiterFactory,
+		internalLimiterFactory: params.InternalLimiterFactory,
+	}
+}
+
+// AllowExternal returns true if the rate limiter for this domain/workflow allows an external request
+func (c *wfCache) AllowExternal(domainID string, workflowID string) bool {
+	c.muc.Lock()
+	defer c.muc.Unlock()
```
taylanisikdemir, line 80: if the underlying rate limiters are thread safe we don't need to lock here. we can add a read/write lock inside `getCacheItem` to avoid duplicate cache entry creation

jakobht: I have removed the lock, the underlying LRU implementation is already thread safe, as well as the ratelimiters, so there should be no need to lock in this struct.

sankari165:
LGTM. do we plan to add observability with metrics/logging in a follow up diff?

jakobht:
> LGTM. do we plan to add observability with metrics/logging in a follow up diff?

I think that makes sense, I didn't want to add more things in this PR, just wanted to establish the foundation and get that reviewed, to make sure we have the same expectations

sankari165 (approved):

service/history/workflowcache/cache.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	value, ok := c.lru.Get(key).(*cacheValue)
+
+	if !ok {
+		value = &cacheValue{
+			externalRateLimiter: c.externalLimiterFactory.GetLimiter(domainID),
+			internalRateLimiter: c.internalLimiterFactory.GetLimiter(domainID),
+		}
+		c.lru.PutIfNotExist(key, value)
```
taylanisikdemir, line 103: I think you need lock here because if 2 concurrent threads reach here only one of them will add the `value` to the cache and the other one will return its local `value` instead of the cached instance

jakobht: Resolved here https://github.com/uber/cadence/pull/5605

PR #5596, opened 2024-01-17
Update grpc handler to use generated code

Shaddoll:
**What changed?**
- Update frontend, history and matching's grpc handlers to use generated code
- Rename some helper functions in types/mapper

**Why?**
To reduce manual work

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/interface.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2017-2020 Uber Technologies Inc.
```
taylanisikdemir, line 1: to be consistent you can move out the interface in matching/handler.go to separate file

taylanisikdemir (approved):

PR #5597, opened 2024-01-17
Submodules step 1/2: convert archivers to a plugin architecture

Groxx:
Breaking the gcloud plugin into a separate submodule (so the main module can remove these optional libraries) requires that the main module does not import it.

Since this was a hard-coded switch statement, that wasn't really possible.  So I've converted it to an at-`init`-time registration system, and I'll invert that import-flow in a later PR.  This one is intentionally trying to be a (relatively) straightforward refactor.

This is almost certainly better not using `init`, and the scheme/key separation seems strange... but that's a less-obviously-identical-behavior change, and we kinda need to split this apart ASAP.
Step 2 will introduce two submodules (gcloud plugin, and cmd/server which will import both the plugin and the "main" module), along with some makefile/automation changes to make things continue working like before.

### DIFF
--- a/common/archiver/provider/provider.go
+++ b/common/archiver/provider/provider.go
@@ -213,7 +213,7 @@ func (p *archiverProvider) GetVisibilityArchiver(scheme, serviceName string) (ar
 		return nil, fmt.Errorf("no visibility archiver constructor for scheme %q", scheme)
 	}
 
-	cfg, ok := p.historyArchiverConfigs[constructor.configKey]
+	cfg, ok := p.visibilityArchiverConfigs[constructor.configKey]
 	if !ok {
 		return nil, fmt.Errorf("no visibility archiver config for scheme %q, config key %q", scheme, constructor.configKey)
 	}
--- a/common/archiver/provider/provider.go
+++ b/common/archiver/provider/provider.go
@@ -22,12 +22,11 @@ package provider
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/uber/cadence/common/archiver"
-	"github.com/uber/cadence/common/archiver/filestore"
-	"github.com/uber/cadence/common/archiver/gcloud"
-	"github.com/uber/cadence/common/archiver/s3store"
+	"github.com/uber/cadence/common/archiver/provider/syncmap"
 	"github.com/uber/cadence/common/config"
 )
 
@@ -60,8 +59,8 @@ type (
 	archiverProvider struct {
 		sync.RWMutex
 
-		historyArchiverConfigs    *config.HistoryArchiverProvider
-		visibilityArchiverConfigs *config.VisibilityArchiverProvider
+		historyArchiverConfigs    config.HistoryArchiverProvider
+		visibilityArchiverConfigs config.VisibilityArchiverProvider
 
 		// Key for the container is just serviceName
 		historyContainers    map[string]*archiver.HistoryBootstrapContainer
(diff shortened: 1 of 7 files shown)


common/archiver/provider/provider.go
```diff
@@ -181,31 +208,19 @@ func (p *archiverProvider) GetVisibilityArchiver(scheme, serviceName string) (ar
...
-		visibilityArchiver, err = gcloud.NewVisibilityArchiver(container, p.visibilityArchiverConfigs.Gstorage)
-
-	default:
-		return nil, ErrUnknownScheme
+	constructor, ok := visibilityConstructors.Get(scheme)
+	if !ok {
+		return nil, fmt.Errorf("no visibility archiver constructor for scheme %q", scheme)
+	}
+
+	cfg, ok := p.historyArchiverConfigs[constructor.configKey]
```
Shaddoll, line 216: it should be visibility archiver

Groxx: ah, yea, I guess this got through builds because they have the same signature, and through tests because we rarely have different archivers... 🤔 

good catch, thanks!  will fix asap.

common/config/config.go
```diff
@@ -576,13 +591,51 @@ type (
 		// URI is the domain default URI for visibility archiver
 		URI string `yaml:"URI"`
 	}
+
+	// YamlNode is a lazy-unmarshaler, because *yaml.Node only exists in gopkg.in/yaml.v3, not v2,
+	// and go.uber.org/config currently uses only v2.
+	YamlNode struct {
```
Shaddoll, line 597: How does it work?

Shaddoll: Nevermind, I was reading doc of v3 package instead of v2.

Shaddoll (approved):

host/testcluster.go
```diff
@@ -335,13 +335,14 @@ func newArchiverBase(enabled bool, logger log.Logger) *ArchiverBase {
...
 		DirMode:  "0766",
 	}
-	provider := provider.NewArchiverProvider(
-		&config.HistoryArchiverProvider{
-			Filestore: cfg,
-		},
-		&config.VisibilityArchiverProvider{
-			Filestore: cfg,
-		},
+	node, err := config.ToYamlNode(cfg)
```
davidporter-id-au, line 338: nice

common/config/config.go
Groxx: yea.  I tried to use v3 first, and then the `Load` tests caught the issue :|

thankfully it's pretty simple to achieve in v2.

common/archiver/provider/syncmap/syncmap.go
```diff
@@ -0,0 +1,61 @@
...
+import "sync"
+
+// syncmap is a very simple type-safe locked map, with semantics similar to sync.Map,
+// but only supports inserting once and getting.
+type (
+	syncmap[K comparable, V any] struct {
+		mut  sync.Mutex
+		data map[K]V
+	}
+	SyncMap[K comparable, V any] interface {
```
Groxx, line 34: I didn't look too hard in our libraries for similar stuff... but this is simple enough that tbh I just kinda don't care.

host/testcluster.go
Groxx: not my proudest bit of shenaniganry, I'll admit, but it is at least pretty simple :)

common/archiver/provider/syncmap/syncmap.go
taylanisikdemir: can you add a unit test for this?

common/config/config.go
```diff
@@ -576,13 +591,51 @@ type (
...
 const (
 	// NonShardedStoreName is the shard name used for singular (non-sharded) stores
 	NonShardedStoreName = "NonShardedStore"
+
+	FilestoreConfig = "filestore"
+	S3storeConfig   = "s3store"
+	GCloudConfig    = "gstorage"
 )
 
+var _ yaml.Unmarshaler = (*YamlNode)(nil)
```
taylanisikdemir, line 611: can you add unit test for new functions here?

common/archiver/provider/syncmap/syncmap.go
Groxx: Added to #5609

common/config/config.go
Groxx: currently being tested via the config `Load`-using tests, which I think does cover this in this package... but lemme check

PR #5601, opened 2024-01-18
feat: pprof support config host

zedongh:
**What changed?**
Config on PProf support host

**Why?**
For container users, default `localhost` make outside access pprof api impossible, need to bind to `0.0.0.0`.

**How did you test it?**
Local

**Potential risks**
No. if not config `host` default `localhost` same to previous state.


common/config/pprof.go
```diff
@@ -65,10 +66,15 @@ func (initializer *PProfInitializerImpl) Start() error {
...
+	host := initializer.PProf.Host
+	if host == "" {
+		host = "localhost"
+	}
+
 	if atomic.CompareAndSwapInt32(&pprofStatus, pprofNotInitialized, pprofInitialized) {
 		go func() {
-			initializer.Logger.Info("PProf listen on ", tag.Port(port))
-			err := http.ListenAndServe(fmt.Sprintf("localhost:%d", port), nil)
+			initializer.Logger.Info("PProf listen on ", tag.Host(host), tag.Port(port))
```
taylanisikdemir, line 76: host tag might be confused with the host name of the service which is appended by logging agents usually. so let's just use `tag.Dynamic("pprof-host", host)` here

common/config/pprof.go
```diff
@@ -65,10 +66,15 @@ func (initializer *PProfInitializerImpl) Start() error {
 		return nil
 	}
 
+	host := initializer.PProf.Host
```
taylanisikdemir, line 69: thanks for adding this support

common/config/pprof.go
zedongh: Ok, changed to `tag.Dynamic("pprof-host", host)` and remove `Host` function from tag .

Shaddoll (approved):

PR #5604, opened 2024-01-18
Adding a stale workflow check to the taskvalidator and code cleanup.

agautam478:
**What changed?**
- Added a stale workflow check to the task validator.
- Added a delete workflow call if the workflow domain is deprecated or if the workflow is stale.
- Removed the call to task validator from the history layer as the plan is now to put it in the execution layer.

Upon detecting an error during the update process, the function call will be placed in the `execution` context to cover all workflow execution scenarios, ensuring a thorough and efficient approach to identifying and resolving corruption across the system. 

**Why?**
These changes were made to complete the taskvalidation service and call in the correct place in future.
With the addition of this final check the core service can be considered ready an usable. We can move forwarding with calling the `taskvalidation` step in the next PR.

**How did you test it?**
Uni tests.

**Potential risks**

**Release notes**

**Documentation Changes**


common/constants.go
```diff
@@ -264,8 +264,3 @@ const (
 	// DefaultHistoryMaxAutoResetPoints is the default maximum number for auto reset points
 	DefaultHistoryMaxAutoResetPoints = 20
 )
-
-const (
```
taylanisikdemir, line 268: thanks for eliminating these copies

common/taskvalidator/validateworkflow.go
```diff
@@ -24,62 +24,93 @@
...
 	metricsClient metrics.Client
 	dc            cache.DomainCache
+	pr            persistence.Retryer
+	staleCheck    staleChecker
 }
 
-// NewWfChecker creates a new instance of Checker.
-func NewWfChecker(logger log.Logger, metrics metrics.Client, domainCache cache.DomainCache) Checker {
+func NewWfChecker(logger log.Logger, metrics metrics.Client, domainCache cache.DomainCache, pr persistence.Retryer) Checker {
+	zapLogger, _ := zap.NewProduction()
```
taylanisikdemir, line 58: we shouldn't create side loggers like this. can `NewStaleWorkflow` accept `log.Logger` instead?

common/taskvalidator/validateworkflow.go
```diff
@@ -24,62 +24,93 @@
...
+	staleCheckInstance := invariant.NewStaleWorkflow(pr, domainCache, zapLogger)
+
+	staleCheck, _ := staleCheckInstance.(staleChecker)
```
taylanisikdemir, line 61: let's not ignore the cast result ok or not. if it is not ok this function should return an error

common/taskvalidator/validateworkflow.go
agautam478: it wasn't able to but maybe with the new stalechecker it will. I will check.

common/constants.go
agautam478: Thanks for the guidance.

common/taskvalidator/validateworkflow.go
```diff
@@ -24,62 +24,123 @@
...
+	pr            persistence.Retryer
+	staleCheck    staleChecker
 }
 
-// NewWfChecker creates a new instance of Checker.
-func NewWfChecker(logger log.Logger, metrics metrics.Client, domainCache cache.DomainCache) Checker {
+// NewWfChecker creates a new instance of a workflow validation checker.
+// It requires a logger, metrics client, domain cache, persistence retryer,
+// and a stale checker implementation to function.
+func NewWfChecker(logger *zap.Logger, metrics metrics.Client, domainCache cache.DomainCache, pr persistence.Retryer, staleCheck staleChecker) Checker {
```
taylanisikdemir, line 62: where is this going to be called?

common/taskvalidator/validateworkflow_test.go
```diff
@@ -23,45 +23,102 @@
...
+	"go.uber.org/zap"
 
 	"github.com/uber/cadence/common/cache"
-	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/metrics"
+	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/service/history/constants"
 )
 
+type MockStaleChecker struct {
```
taylanisikdemir, line 40: this doesn't need to be public

common/taskvalidator/validateworkflow_test.go
```diff
@@ -23,45 +23,102 @@
...
-
-			// Set up the mock behavior for GetDomainByID
+			mockPersistenceRetryer := persistence.NewMockRetryer(mockCtrl)
+			mockStaleChecker := &MockStaleChecker{
+				CheckAgeFunc: func(response *persistence.GetWorkflowExecutionResponse) (bool, error) {
+					return tc.isStale, nil
+				},
+			}
+			checker := NewWfChecker(mockLogger, mockMetricsClient, mockDomainCache, mockPersistenceRetryer, mockStaleChecker)
+			checker.(*checkerImpl).staleCheck = mockStaleChecker
```
taylanisikdemir, line 78: you already pass `mockStaleChecker` to constructor above. why is this line needed?

common/taskvalidator/validateworkflow.go
taylanisikdemir: I see you changed constructor to expect `*zap.Logger`. that doesn't address my concern. we should keep it consistent via `log.Logger`. let's look into changing the stale checker to accept `log.Logger` not the other way around

allenchen2244 (approved):

PR #5605, opened 2024-01-18
Addressed sync issue in workflow cache

jakobht:
**What changed?**
We now use the value returned from PutIfExists, this is thread safe, either the new value is inserted and returned, or the existing value is returned. 
We also do not need to pin values in the cache, if they get evicted while we use them that is fine. 
Added some more error handling, the error handling should never be triggered though. 

**Why?**
The old implementation had a race condition in creating the value and inserting it

**How did you test it?**

**Potential risks**
There could still be concurrency issues 

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/workflowcache/cache.go
+++ b/service/history/workflowcache/cache.go
@@ -43,6 +43,7 @@ type wfCache struct {
 	externalLimiterFactory quotas.LimiterFactory
 	internalLimiterFactory quotas.LimiterFactory
 	logger                 log.Logger
+	getCacheItemFn         func(domainID string, workflowID string) (*cacheValue, error)
 }
 
 type cacheKey struct {
@@ -66,7 +67,7 @@ type Params struct {
 
 // New creates a new WFCache
 func New(params Params) WFCache {
-	return &wfCache{
+	cache := &wfCache{
 		lru: cache.New(&cache.Options{
 			TTL:      params.TTL,
 			Pin:      false,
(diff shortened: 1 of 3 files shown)


service/history/workflowcache/cache.go
```diff
@@ -95,13 +111,33 @@ func (c *wfCache) getCacheItem(domainID string, workflowID string) *cacheValue {
...
-		c.lru.PutIfNotExist(key, value)
+		return nil, errors.New("Failed to insert new value into cache")
 	}
 
-	return value
+	return value, err
+}
+
+func (c *wfCache) logError(domainID string, workflowID string, err error) {
+	c.logger.Error("Unexpected error from workflow cache",
```
taylanisikdemir, line 137: not in this PR but we can add metrics to cover cache hit/miss/err counts per domain.

jakobht: Yes, that is the next step

service/history/workflowcache/cache.go
```diff
@@ -57,36 +61,48 @@ type Params struct {
...
 	}
 }
 
 // AllowExternal returns true if the rate limiter for this domain/workflow allows an external request
 func (c *wfCache) AllowExternal(domainID string, workflowID string) bool {
 	// Locking is not needed because both getCacheItem and the rate limiter are thread safe
-	value := c.getCacheItem(domainID, workflowID)
+	value, err := c.getCacheItem(domainID, workflowID)
+	if err != nil {
+		c.logError(domainID, workflowID, err)
```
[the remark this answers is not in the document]
jakobht: I have added a unit test for the error case, good suggestion

taylanisikdemir (approved):

PR #5606, opened 2024-01-19
add ContextKey Struct

bowenxia:
**What changed?**
Added a struct ContextKey

**Why?**
This is for Pinot/ES response comparator usage:
In mono repo (Cadence-Operation) our workflow needs to query both Pinot and ES in the runtime. So we need to pass in a Context Key-Value (visibility-override = Primary/Secondary) pair to reach this goal. Context key can only be a struct, so we need to create one here in OSS so that both OSS and Mono repo can use it. 

**How did you test it?**
It's just one struct, doesn't include any logics. 

**Potential risks**
none

**Release notes**

**Documentation Changes**


common/types/shared.go
```diff
@@ -8174,3 +8174,6 @@ func SerializeRequest(request interface{}) (string, error) {
 
 	return string(res), nil
 }
+
+// ContextKey is for Pinot/ES response comparator. This struct will be passed into ctx as a key.
+type ContextKey string
```
davidporter-id-au, line 8179: This name is unfortunately a bit too general for use for the Pinot / Comparitor. Not entirely sure what you're aiming for sorry. if we need to create a type, maybe we can either prefix it with the `VisibliityManager` and maybe keep it in Persistence? The types here are intended for IDL mapping only

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -46,8 +46,9 @@ type (
 )
 
 const (
-	Primary   = "Primary"
-	Secondary = "Secondary"
+	Primary    = "Primary"
+	Secondary  = "Secondary"
+	ContextKey = types.ContextKey("visibility-override")
```
davidporter-id-au, line 51: Mind if I ask what your thinking was with creating a type for this? This might be a pattern we're doing elsewhere, I may have just missed it

common/types/shared.go
bowenxia: moved it to a different folder

neil-xie (approved):
Good catch

PR #5612, opened 2024-01-22
StartWorkflowMetrics

jakobht:
**What changed?**
Made the cache aware if the cacheEnabled dynamic config

**Why?**
We need this before enabling anything as we need to be able to disable the cache if we start seeing issues

**How did you test it?**
Unit tests and local tests 

**Potential risks**
We could see a lot of error logs that cannot be disabled  without changing the code, if we cannot get the domains from the domainCache, I have tested locally and it looks to work

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/workflowcache/cache.go
+++ b/service/history/workflowcache/cache.go
@@ -125,6 +125,7 @@ func (c *wfCache) AllowExternal(domainID string, workflowID string) bool {
 // AllowInternal returns true if the rate limiter for this domain/workflow allows an internal request
 func (c *wfCache) AllowInternal(domainID string, workflowID string) bool {
 	if !c.workflowIdCacheEnabledCheck(domainID) {
+		// If we can't get the cache item, we should allow the request through
 		return true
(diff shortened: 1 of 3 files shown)


service/history/workflowcache/cache.go
```diff
@@ -97,6 +124,11 @@ func (c *wfCache) AllowExternal(domainID string, workflowID string) bool {
 
 // AllowInternal returns true if the rate limiter for this domain/workflow allows an internal request
 func (c *wfCache) AllowInternal(domainID string, workflowID string) bool {
+	if !c.workflowIdCacheEnabledCheck(domainID) {
+		// If we can't get the cache item, we should allow the request through
```
sankari165, line 128: nit: match the comment when we allow requests for internal and external requests because cache is not enabled

jakobht: Thanks, good catch!

sankari165 (approved):

PR #5616, opened 2024-01-23
Get/Update DomainAsyncWorkflowConfiguraton methods in admin API and CLI

taylanisikdemir:
**What changed?**
Continuing from where I left in #5608. Admin API methods `GetDomainAsyncWorkflowConfiguraton and `UpdateDomainAsyncWorkflowConfiguraton` are implemented and corresponding CLI tool commands are exposed.

Next steps:
- DB schema changes on DomainConfig type to make this work e2e
- Complete changes in config.go. 

**How did you test it?**
- Unit and integration tests to validate nothing is broken
- Tried new API via new CLI commands
Get example: 
```
go run cmd/tools/cli/main.go admin aq get \
  --domain samples-domain
```
Put example: 
```
go run cmd/tools/cli/main.go admin aq update \
  --domain cadence-samples \
  --json "{\"PredefinedQueueName\":\"yoyo\"}"`
```


common/config/config.go
```diff
@@ -597,6 +601,18 @@ type (
...
 	}
+
+	// AsyncWorkflowQueueProvider contains the config for an async workflow queue. Only one field must be set.
+	AsyncWorkflowQueueProvider struct {
+		Type                     string                    `yaml:"type"`
+		AsyncWorkflowQueueConfig *AsyncWorkflowQueueConfig `yaml:",inline"`
+	}
+
+	AsyncWorkflowQueueConfig struct {
+		Kafka *KafkaConfig `yaml:"kafka"`
```
Shaddoll, line 612: we should not use this config, because this config doesn't work for us. You can use it as a placeholder.
But we do need to define a new struct to store the config.

taylanisikdemir: this was leftover from prototyping. not used in this PR anyway so removing to avoid confusion

common/persistence/serializer_test.go
```diff
@@ -58,8 +58,8 @@ func (s *cadenceSerializerSuite) SetupTest() {
 	s.Assertions = require.New(s.T())
 }
 
+// TODO: update tests to cover asyncworkflowconfig
```
Shaddoll, line 61: Do you mind adding the tests?

Shaddoll (approved):

common/persistence/serializer_test.go
taylanisikdemir: [this test body](https://github.com/uber/cadence/blob/master/common/persistence/serializer_test.go#L210-L583) is not ideal. I will convert this to proper table test and simplify the body in a follow up PR

PR #5623, opened 2024-01-23
Set proper max reset points

neil-xie:
**What changed?**
Update to get max reset points from the config instead of math.MaxInt32

**Why?**
The number of binary checksums are growing unlimitedly, which will hammer the persistence
It also exceeds the limit of pinot json index length eventually

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -1982,7 +1982,7 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
 		}
 	}
 
-	if len(currResetPoints) == maxResetPoints {
+	if len(currResetPoints) >= maxResetPoints {
 		// If exceeding the max limit, do rotation by taking the oldest one out.
 		currResetPoints = currResetPoints[1:]
 		recentBinaryChecksums = recentBinaryChecksums[1:]
--- a/service/history/execution/mutable_state_builder.go
+++ b/service/history/execution/mutable_state_builder.go
@@ -1982,11 +1982,8 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
 		}
 	}
 
-	if len(currResetPoints) >= maxResetPoints {
-		// If exceeding the max limit, do rotation by taking the oldest one out.
-		currResetPoints = currResetPoints[1:]
-		recentBinaryChecksums = recentBinaryChecksums[1:]
-	}
+	recentBinaryChecksums, currResetPoints = trimBinaryChecksums(recentBinaryChecksums, currResetPoints, maxResetPoints)
+
 	// Adding current version of the binary checksum.
 	recentBinaryChecksums = append(recentBinaryChecksums, binChecksum)
 
--- a/service/history/execution/mutable_state_decision_task_manager.go
+++ b/service/history/execution/mutable_state_decision_task_manager.go
@@ -253,8 +253,12 @@ func (m *mutableStateDecisionTaskManagerImpl) ReplicateDecisionTaskCompletedEven
 	event *types.HistoryEvent,
 ) error {
 	m.beforeAddDecisionTaskCompletedEvent()
-	domainName := m.msb.GetDomainEntry().GetInfo().Name
-	return m.afterAddDecisionTaskCompletedEvent(event, m.msb.config.MaxAutoResetPoints(domainName))
+	maxResetPoints := common.DefaultHistoryMaxAutoResetPoints // use default when it is not set in the config
+	if m.msb.GetDomainEntry() != nil && m.msb.GetDomainEntry().GetInfo() != nil && m.msb.config != nil {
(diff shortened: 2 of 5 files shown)


service/history/execution/mutable_state_decision_task_manager.go
```diff
@@ -254,7 +253,8 @@ func (m *mutableStateDecisionTaskManagerImpl) ReplicateDecisionTaskCompletedEven
 	event *types.HistoryEvent,
 ) error {
 	m.beforeAddDecisionTaskCompletedEvent()
-	return m.afterAddDecisionTaskCompletedEvent(event, math.MaxInt32)
+	domainName := m.msb.GetDomainEntry().GetInfo().Name
+	return m.afterAddDecisionTaskCompletedEvent(event, m.msb.config.MaxAutoResetPoints(domainName))
```
davidporter-id-au, line 257: Does this need a nil check? I'm a bit worried that this config value isn't set in some circumstances?

neil-xie: It is used in multiple places, but yeah, seems we don't have nil check in these funcs. I can try writing unit test here, this file is not covered at all :(

service/history/execution/mutable_state_builder.go
```diff
@@ -1982,7 +1982,7 @@ func (e *mutableStateBuilder) addBinaryCheckSumIfNotExists(
 		}
 	}
 
-	if len(currResetPoints) == maxResetPoints {
+	if len(currResetPoints) >= maxResetPoints {
```
taylanisikdemir, line 1985: The 2 lines below assumes size of `currResetPoints` equals to `maxResetPoints` and discards first item. If it is possible that size is greater than max points then let's update trim offset accordingly

taylanisikdemir: and would be great to add a unit test for this function as a follow up

neil-xie: Sure, I will the test. As for trimming the size, there are some existing workflows with huge list, I think it is a good idea to remove them, just not sure the safety. @davidporter-id-au Any concerns here?

service/history/execution/mutable_state_decision_task_manager.go
neil-xie: Added  nil check here. Seems it is also safe without check, we build the domainEntry from the very base level

service/history/execution/mutable_state_builder.go
davidporter-id-au: I don't see any problems in trimming this. I think we should check with the team, but I'd suggest just keeping the last 5 seems reasonable

davidporter-id-au: sorry, reading this properly for the first time. 

Yeah++ to testing and maybe a small refactoring, 

I think actually we probably want to do it now, I'm sorry to say Neil.

This whole function is a bit of a mess. It should just be a simple pure function, rather than a method,

neil-xie: Sure, sounds like a good idea to refactor and move it to a func, also easier for test.

neil-xie: I moved the trim part to a func and added tests. Let me know if we need to refactor more there

service/history/execution/mutable_state_util_test.go
```diff
@@ -148,3 +148,62 @@ func TestFindAutoResetPoint(t *testing.T) {
...
+		},
+		{
+			name:           "reach at limit",
+			maxResetPoints: 5,
+			expected:       []string{"checksum2", "checksum3", "checksum4", "checksum5"},
+		},
+		{
+			name:           "exceeds limit",
+			maxResetPoints: 2,
+			expected:       []string{"checksum5"},
```
taylanisikdemir, line 171: shouldn't this be checksum4 and checksum5?

neil-xie: It needs an extra space for the current run's binChecksum after we trim https://github.com/uber/cadence/pull/5623/files#diff-734db27e2719815a56d959e5cc23ec535d603fdcd499ec5a9b18d0dbf8fa1a55R1988.
It is easier to think the reach at limit case, if we passed in 5 elements and max limit is 5, we will still need to get rid of oldest one.

taylanisikdemir: oh I see. just noticed the case above also trims checksum1 and leaves 4 elements. so basically we shrink the list to `maxResetPoints-1`.

davidporter-id-au: +1

taylanisikdemir (approved):

service/history/execution/mutable_state_util_test.go
davidporter-id-au: oh, yeah, sorry, was slow to catch this too

davidporter-id-au (approved):

PR #5626, opened 2024-01-24
Put a timeout for timer task deletion loop during shutdown

taylanisikdemir:
**What changed?**
Timer queue processor has a loop to delete tasks batch by batch periodically and also during shut down. It just uses background context so it can take a long time in theory.
For shut down scenario adding a timeout to make sure it doesn't hang the process when there's a lot of tasks to delete.

**Why?**
Respect shut down signals to make rollouts/restarts more smooth.

**How did you test it?**
Will test in a staging environment. Changes are behind a dynamic config.

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -347,7 +347,17 @@ func (t *timerQueueProcessor) UnlockTaskProcessing() {
 }
 
 func (t *timerQueueProcessor) drain() {
-	if err := t.completeTimer(); err != nil {
+	if !t.shard.GetConfig().QueueProcessorEnableGracefulSyncShutdown() {
+		if err := t.completeTimer(context.Background()); err != nil {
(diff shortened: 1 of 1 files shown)


service/history/queue/timer_queue_processor.go
```diff
@@ -365,7 +375,7 @@ func (t *timerQueueProcessor) completeTimerLoop() {
 			return
 		case <-completeTimer.C:
 			for attempt := 0; attempt < t.config.TimerProcessorCompleteTimerFailureRetryCount(); attempt++ {
-				err := t.completeTimer()
+				err := t.completeTimer(context.Background())
```
davidporter-id-au, line 378: I wonder if this should be unbounded, this is strictly better however.

davidporter-id-au (approved):

service/history/queue/timer_queue_processor.go
taylanisikdemir: I debated on this but leaving it as is for now to limit the changes to be tested in actual deployment

PR #5631, opened 2024-01-26
Add debug logs in PinotTripleVisibilityManager for response comparator testing

bowenxia:
**What changed?**
Added debugging logs to PinotTripleVisibilityManager, for testing pinot response comparator

**Why?**
Without the logs, we can't know which database Cadence chooses.

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


neil-xie:
This method can be moved into chooseVisibilityManagerForRead() so we don't need to do it twice. If it has override, we choose based on the override value, if not, use the original logic

neil-xie:
The title needs to be updated, we need it for release

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -344,6 +344,18 @@ func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userPara
 		tag.VisibilityQuery(userParam.customQuery))
 }
 
+func (v *pinotVisibilityTripleManager) chooseManagerByOverride(ctx context.Context, domain string) VisibilityManager {
+	if override := ctx.Value(ContextKey); override == Primary {
+		v.logger.Debug("primary was chosen")
+		return v.esVisibilityManager
```
neil-xie: It is better we also add a nil check, same as the chooseVisibilityManagerForRead()

bowenxia: Merged the 2 choose manager functions together. If there are no contextkey, it will fall through into the original logic.

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -541,16 +491,22 @@ func (v *pinotVisibilityTripleManager) CountWorkflowExecutions(
...
-	manager := v.chooseVisibilityManagerForRead(request.Domain)
+	manager := v.chooseVisibilityManagerForRead(ctx, request.Domain)
+	return manager.CountWorkflowExecutions(ctx, request)
+}
+
+func (v *pinotVisibilityTripleManager) chooseVisibilityManagerForRead(ctx context.Context, domain string) VisibilityManager {
+	// temporary code: for Pinot migration usage -->
 	if override := ctx.Value(ContextKey); override == Primary {
-		manager = v.esVisibilityManager
+		v.logger.Debug("primary was chosen")
```
neil-xie, line 501: let's change the message here to make it more reasonable to others, like primary visibility manager elasticsearch was chosen for read

common/persistence/pinotVisibilityTripleManager.go
neil-xie: Yeah, just in case, since we have nil check guard in the original logic, but no for our override. But this could be fine for now, since no customer will use

neil-xie (approved):

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -541,16 +491,22 @@ func (v *pinotVisibilityTripleManager) CountWorkflowExecutions(
...
+		return v.esVisibilityManager
 	} else if override == Secondary {
-		manager = v.pinotVisibilityManager
+		v.logger.Debug("Pinot Migration log: Secondary visibility manager was chosen for read.")
+		return v.pinotVisibilityManager
 	}
-	return manager.CountWorkflowExecutions(ctx, request)
-}
+	v.logger.Debug("Pinot Migration log: ContextKey was empty.")
+	// temporary code: ends here <--
```
davidporter-id-au, line 508: probably better to just link to a task ID or a todo

bowenxia: Created a Jira ticket to record this.

PR #5634, opened 2024-01-29
Adding a sample call to TaskValidator in update workflow cycle

agautam478:
**What changed?**
Made some dummy calls and updated the TaskValidation function to. make it compatible with the execution context.
The code is ready to be handed over to the future devs of this feature.

Added a dynamic config variable in order to control these calls on the basis of the domain ID.

**Why?**
The changes are made to test the TaskValidatiion functionality, see it's feasibility in the update flow and to prepare a dummy call instance so that it can be enabled right away.

**How did you test it?**
Yes. The feature comes with a unit test.

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**
NA


service/history/execution/context.go
```diff
@@ -1218,6 +1218,13 @@ func (c *contextImpl) updateWorkflowExecutionWithRetry(
 		resp, err = c.shard.UpdateWorkflowExecution(ctx, request)
 		return err
 	}
+	//Preparation for the task Validation.
+	//metricsClient := c.shard.GetMetricsClient()
+	//domainCache := c.shard.GetDomainCache()
+	//executionManager := c.shard.GetExecutionManager()
+	//historymanager := c.shard.GetHistoryManager()
+	//zapLogger, _ := zap.NewProduction()
+	//checker, _ := taskvalidator.NewWfChecker(zapLogger, metricsClient, domainCache, executionManager, historymanager)
```
neil-xie, line 1227: I think we could just guard here using the new flag you created

agautam478: yeah! that's the hope now. I didn't add it because I am not sure if you guys would want to call it here in this flow. currently this seems like a logical place but if you want to go more granular ie. cal the validation only on closed workflows then you might have to move this somewhere else.

We would require to do some extensive testing to finalise the most optimal place to call it.

neil-xie (approved):

PR #5635, opened 2024-01-29
Catch unit test failures in `make test`

Groxx:
Surprisingly, this has not worked since tests were introduced to the makefile _in 2017_.
We've apparently been relying on CI's grepping for go test failure lines or something.

Rather trivially, without `-o pipefail` this only returns error messages from `tee`:
```bash
> false | tee -a file.log
> echo $?
0
```
Which means this also does not return `exit 1`:
```bash
for ... do
  go test whatever | tee -a file.log;
done
```
Which means you can fail a test and the tests keep going, and it exits 0:
```bash
> make test
...
FAIL	github.com/uber/cadence/common/util	0.245s
FAIL
ok  	github.com/uber/cadence/common/locks	0.252s	coverage: 93.3% of statements
...

> echo $?
0
```
Which kinda stinks for local development.  Failing tests can be hard to spot in the massive output of `V=1`, or dozens of lines off-screen if a lot of successful packages ran afterward.

Now this will continue running tests, but will clearly fail with a list of all failed packages:
```bash
> make test
...
FAIL    github.com/uber/cadence/common/util     0.245s
FAIL
ok      github.com/uber/cadence/common/locks    0.252s  coverage: 93.3% of statements
...
Failed packages:  ./common/util
make: *** [test] Error 1

> echo $?
1
```
Which is roughly how `go test ./...` works: it fails but keeps going, and does an `exit 1` if there were failures along the way (though it does not show the list of packages).

---

This also removes `bins` as a prerequisite for tests, as:
- it's slow and usually not necessary
- we now have `make build` which is MUCH more complete
- CI does this explicitly anyway, no need to do it more than once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -575,7 +575,7 @@ GOCOVERPKG_ARG := -coverpkg="$(subst $(SPACE),$(COMMA),$(addprefix $(PROJECT_ROO
 # CAUTION: when changing to `go test ./...`, note that this DOES NOT test submodules.  Those must be run separately.
 define looptest
 $Q FAIL=""; for dir in $1; do \
-	go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log || FAIL="$$FAIL $$dir"; \
+	go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) 2>&1 | tee -a test.log || FAIL="$$FAIL $$dir"; \
 done; test -z "$$FAIL" || (echo "Failed packages; $$FAIL"; exit 1)
 endef
 
--- a/Makefile
+++ b/Makefile
@@ -608,7 +608,7 @@ cover_profile:
 	$Q for dir in $(PKG_TEST_DIRS); do \
 		mkdir -p $(BUILD)/"$$dir"; \
 		go test "$$dir" $(TEST_ARG) -coverprofile=$(BUILD)/"$$dir"/coverage.out || exit 1; \
-		cat $(BUILD)/"$$dir"/coverage.out | grep -v "^mode: \w\+" >> $(UNIT_COVER_FILE); \
+		(cat $(BUILD)/"$$dir"/coverage.out | grep -v "^mode: \w\+" >> $(UNIT_COVER_FILE)) || true; \
 	done;
 
 cover_integration_profile:
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,15 @@
 MAKEFLAGS += --no-builtin-rules
 .SUFFIXES:
 
+# pipefail is very easy to forget about, and sadly not the default.
+# this changes that, and makes our target scripts more strict.
+#
+# make sure to include all this if overriding on the CLI, e.g.:
+# 	DO:      make SHELL='bash -eoux pipefail'
+# 	DO NOT:  make SHELL='bash -x'
+# as otherwise you will ignore all errors.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -561,30 +570,32 @@ COVER_PKGS = client common host service tools
 # pkg -> pkg/... -> github.com/uber/cadence/pkg/... -> join with commas
 GOCOVERPKG_ARG := -coverpkg="$(subst $(SPACE),$(COMMA),$(addprefix $(PROJECT_ROOT)/,$(addsuffix /...,$(COVER_PKGS))))"
 
+# iterates over a list of dirs and runs go test on each one, collecting errors as it runs.
+# this is primarily written because it's a verbose bit of boilerplate, until we switch to `go test ./...` where possible.
+# CAUTION: when changing to `go test ./...`, note that this DOES NOT test submodules.  Those must be run separately.
+define looptest
+$Q FAIL=""; for dir in $1; do \
+	go test $(TEST_ARG) -coverprofile=$@ "$$dir" $(TEST_TAG) | tee -a test.log || FAIL="$$FAIL $$dir"; \
```
davidporter-id-au, line 578: I think the tee needs to grab stderr too?

Makefile
```diff
@@ -597,10 +608,10 @@ cover_profile:
 	$Q for dir in $(PKG_TEST_DIRS); do \
 		mkdir -p $(BUILD)/"$$dir"; \
 		go test "$$dir" $(TEST_ARG) -coverprofile=$(BUILD)/"$$dir"/coverage.out || exit 1; \
-		cat $(BUILD)/"$$dir"/coverage.out | grep -v "^mode: \w\+" >> $(UNIT_COVER_FILE); \
+		(cat $(BUILD)/"$$dir"/coverage.out | grep -v "^mode: \w\+" >> $(UNIT_COVER_FILE)) || true; \
```
davidporter-id-au, line 611: What is the grep filter for?  the fact that multiline comments are hard would make me lean to towards throwing into a bash script

davidporter-id-au (approved):
lgtm with nit about logs

Makefile
Groxx: Historically, it was because you can just concat multiple coverage files to have go  transparently compute the combined coverage... except for those mode lines, which would cause it to error.

Nowadays I know there are better options in a few ways, but I haven't dug into them in detail yet.

---

Re script file: yeah, this is borderline or slightly past I think.  But it is hopefully temporary and will also-hopefully be simpler after a modernization pass, so I'm kinda reluctant to do more than the minimum / putting it in here keeps the badness visible until it's replaced / etc :/

I could be convinced to do otherwise, just didn't feel like it was worth it at the moment.

Makefile
Groxx: That is probably true, yeah. Good catch.

Why it exists: outside human use (which I doubt anyone does on the team), it can in principle give you a simple upload to buildkite, like for coverage, and I think I saw that at one point (or maybe I'm getting it confused with internal systems).  Buildkite's "download the raw log file" format is pretty awkward comparatively.

I do mostly lean towards "not worth keeping" though. Feels like just unused legacy to me. I never did finish rewriting these makefiles, you can still see the separator where I stopped a bit above here...

Groxx: all I'm seeing it used for is build failures and probably race crashes... but yes, that seems unambiguously worth gathering into the same log file, otherwise it just says "fail: package".

added!

PR #5636, opened 2024-01-29
Generate rate limit frontend api handler

Shaddoll:
**What changed?**
Generate rate limit frontend api handler code
Some fix:
- Change rate limit type of the following APIs:
- ScanWorkflowExecution
- RefreshWorkflowTasks
- GetClusterInfo

**Why?**
To reduce manual work

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/frontend/templates/ratelimited.tmpl
```diff
@@ -0,0 +1,117 @@
...
+    {{- $domain := printf "%s.GetDomain()" (index $method.Params 1).Name}}
+    {{- $rateLimitHelper := "allowDomain"}}
+    {{- if has $method.Name $domainIDAPIs}}
+    {{- $domain = "domainName"}}
+    {{- $rateLimitHelper := "allowDomainID"}}
+    {{- if has $method.Name $queryTaskTokenAPIs}}
+    token, err := h.tokenSerializer.DeserializeQueryTaskToken({{(index $method.Params 1).Name}}.TaskToken)
+    {{- else}}
+    token, err := h.tokenSerializer.Deserialize({{(index $method.Params 1).Name}}.TaskToken)
+    {{- end}}
```
Groxx, line 95: fwiw: you can indent the code lines on all of these as deeply as you want, the formatting step will fix the indentation.

it won't remove empty lines, so the `{{- ` is still useful/desirable, but indents help a fair bit with where the if/ends are paired.

service/frontend/templates/ratelimited.tmpl
```diff
@@ -0,0 +1,117 @@
...
+{{$ratelimitTypeMap = set $ratelimitTypeMap "RequestCancelWorkflowExecution" "ratelimitTypeUser"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "RestartWorkflowExecution" "ratelimitTypeUser"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ListOpenWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ListArchivedWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ListClosedWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ListWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ScanWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "CountWorkflowExecutions" "ratelimitTypeVisibility"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "ResetStickyTaskList" "ratelimitTypeWorker"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "QueryWorkflow" "ratelimitTypeUser"}}
```
Groxx, line 38: I'll vote to group by type and alphabetize within that group fwiw.  makes it easier to spot associations (which are pretty relevant).

Groxx (commented):
gonna stick myself in here for "I think this needs a careful read before approval, but broadly I like it".

main concern is that this makes a significant change to when the call occurs, and that needs to be verified to be safe.

whether we return a 429 or "bad request" error on floods of bad requests: eh, I think both are defensible, changing it seems fine to me.  it's just worth checking if there are any particularly valuable ones that will be lost (I sincerely doubt it).

service/frontend/service.go
```diff
@@ -131,8 +133,36 @@ func (s *Service) Start() {
...
+			service.Frontend,
+			s.config.GlobalDomainVisibilityRPS,
+			s.config.MaxDomainVisibilityRPSPerInstance,
+			s.GetMembershipResolver(),
+		)),
+	)
 	// Additional decorations
 	var handler api.Handler = s.handler
+	handler = ratelimited.NewAPIHandler(handler, s.GetDomainCache(), userRateLimiter, workerRateLimiter, visibilityRateLimiter)
 	handler = metered.NewAPIHandler(handler, s.GetLogger(), s.GetMetricsClient(), s.GetDomainCache(), s.config)
```
Groxx, line 166: I think I approve of this change, just making it visible / as a reminder for myself for later: this will mean that ratelimited requests no longer contribute to response latency / success rate.

service/frontend/templates/ratelimited.tmpl
```diff
@@ -0,0 +1,119 @@
...
+            {{- else}}
+                token, err := h.tokenSerializer.Deserialize({{(index $method.Params 1).Name}}.TaskToken)
+            {{- end}}
+            if err != nil {
+                return
+            }
+            domainName, err := h.domainCache.GetDomainName(token.DomainID)
+            if err != nil {
+                return
+            }
```
Groxx, line 104:
```
            {{- $rateLimitHelper := "allowDomainID"}}
```
with `:=`, this looks unused in this block.  or does it still modify the lines below that have this?
```
            h.{{$rateLimitHelper}}({{$ratelimitType}}, {{$domain}})
```

(I'm not 100% sure what the scoping behavior is with templates, but I assume it's the same as go...?  hopefully?)

service/frontend/wrappers/ratelimited/ratelimit.go
```diff
@@ -0,0 +1,49 @@
...
+	case ratelimitTypeUser:
+		return h.userRateLimiter.Allow(quotas.Info{Domain: domain})
+	case ratelimitTypeWorker:
+		return h.workerRateLimiter.Allow(quotas.Info{Domain: domain})
+	case ratelimitTypeVisibility:
+		return h.visibilityRateLimiter.Allow(quotas.Info{Domain: domain})
+	default:
+		panic("coding error, unrecognized request ratelimit type value")
+	}
+}
```
Groxx, line 49: this is a fairly minimal change and if you prefer that then 👍 but:
we could take this opportunity to remove this method and directly call the appropriate limiter.

I didn't do that initially when adding these, to keep things minimal, but it's definitely due for a cleanup at some point.  Does not have to be now though.

service/frontend/templates/ratelimited.tmpl
Shaddoll: it's unused

service/frontend/api/handler.go
```diff
@@ -495,11 +455,6 @@ func (wh *WorkflowHandler) PollForActivityTask(
 		return nil, validate.ErrIdentityTooLong
 	}
 
-	if ok := wh.allow(ratelimitTypeWorker, pollRequest); !ok {
-		// pollers exponentially back off up to 10s
-		return nil, createServiceBusyError()
-	}
```
Groxx, line 501: so far I think the only concerning point is that this check is done earlier:
```go
	domainName := pollRequest.GetDomain()
	if domainName == "" {
		return nil, validate.ErrDomainNotSet
	}
```
which means that after this change, `wh.allow` may be called with an empty domain string.

I'm not 100% sure what happens in that case tbh.  I doubt it's _damaging_, so it might be fine to do that, but it's worth figuring out.

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+	return h.wrapped.QueryWorkflow(ctx, qp1)
+}
+
+func (h *apiHandler) RecordActivityTaskHeartbeat(ctx context.Context, rp1 *types.RecordActivityTaskHeartbeatRequest) (rp2 *types.RecordActivityTaskHeartbeatResponse, err error) {
+	token, err := h.tokenSerializer.Deserialize(rp1.TaskToken)
+	if err != nil {
+		return
+	}
+	domainName, err := h.domainCache.GetDomainName(token.DomainID)
```
Groxx, line 200: probably missing an empty-check / the `validate.ErrDomainNotSet` branch.  the domain cache doesn't seem to check for empty values early on at least, kinda seems like it might not expect emptys.

we might also want to change the deserialize-err to a bad-request too, but technically that'd be a semantic change / only if you feel like it IMO.

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,417 @@
...
+func (h *apiHandler) RestartWorkflowExecution(ctx context.Context, rp1 *types.RestartWorkflowExecutionRequest) (rp2 *types.RestartWorkflowExecutionResponse, err error) {
+	if ok := h.allowDomain(ratelimitTypeUser, rp1.GetDomain()); !ok {
+		err = &types.ServiceBusyError{Message: "Too many outstanding requests to the cadence service"}
+		return
+	}
+	return h.wrapped.RestartWorkflowExecution(ctx, rp1)
+}
+
+func (h *apiHandler) ScanWorkflowExecutions(ctx context.Context, lp1 *types.ListWorkflowExecutionsRequest) (lp2 *types.ListWorkflowExecutionsResponse, err error) {
+	if ok := h.allowDomain(ratelimitTypeVisibility, lp1.GetDomain()); !ok {
```
Groxx, line 372: surprisingly, ScanWorkflowExecutions is currently classified as a "user" request.

that does seem wrong, so 👍 to this change, just calling it out for visibility.

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,417 @@
...
+		tokenSerializer:       common.NewJSONTaskTokenSerializer(),
+		domainCache:           domainCache,
+		userRateLimiter:       userRateLimiter,
+		workerRateLimiter:     workerRateLimiter,
+		visibilityRateLimiter: visibilityRateLimiter,
+	}
+}
+
+func (h *apiHandler) CountWorkflowExecutions(ctx context.Context, cp1 *types.CountWorkflowExecutionsRequest) (cp2 *types.CountWorkflowExecutionsResponse, err error) {
+	if ok := h.allowDomain(ratelimitTypeVisibility, cp1.GetDomain()); !ok {
```
Groxx, line 68: same as scan, surprisingly this was user too, also 👍

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,417 @@
...
+func (h *apiHandler) DescribeWorkflowExecution(ctx context.Context, dp1 *types.DescribeWorkflowExecutionRequest) (dp2 *types.DescribeWorkflowExecutionResponse, err error) {
+	if ok := h.allowDomain(ratelimitTypeUser, dp1.GetDomain()); !ok {
+		err = &types.ServiceBusyError{Message: "Too many outstanding requests to the cadence service"}
+		return
+	}
+	return h.wrapped.DescribeWorkflowExecution(ctx, dp1)
+}
+
+func (h *apiHandler) GetClusterInfo(ctx context.Context) (cp1 *types.ClusterInfo, err error) {
+	return h.wrapped.GetClusterInfo(ctx)
```
Groxx, line 100: previously user-limited, but this seems fine.  it's a trivial endpoint / other cluster-wide things don't have limits and nobody really calls them anyway.

Groxx (commented):
All done with a round-1 careful read!

A couple smallish changes worth verifying, but other than those / check comments, this all looks good.  should be pretty easy to re-review after those things are checked/changed.

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,417 @@
...
+	return h.wrapped.RecordActivityTaskHeartbeatByID(ctx, rp1)
+}
+
+func (h *apiHandler) RefreshWorkflowTasks(ctx context.Context, rp1 *types.RefreshWorkflowTasksRequest) (err error) {
+	if ok := h.allowDomain(ratelimitTypeUser, rp1.GetDomain()); !ok {
+		err = &types.ServiceBusyError{Message: "Too many outstanding requests to the cadence service"}
+		return
+	}
+	return h.wrapped.RefreshWorkflowTasks(ctx, rp1)
+}
```
Groxx, line 223: ah, great, good catch.  yes, this should be user-limited.

service/frontend/wrappers/ratelimited/api_generated.go
Groxx: yea, domain cache will go this all the way down to the database without checking.  it's worth checking here.

service/frontend/wrappers/ratelimited/api_generated.go
```diff
@@ -0,0 +1,718 @@
...
+
+func (h *apiHandler) RespondActivityTaskCanceled(ctx context.Context, rp1 *types.RespondActivityTaskCanceledRequest) (err error) {
+	if rp1 == nil {
+		err = validate.ErrRequestNotSet
+		return
+	}
+	if rp1.TaskToken == nil {
+		err = validate.ErrTaskTokenNotSet
+		return
+	}
```
Groxx, line 417: at some point we should probably make some validation-wrappers that occur early and get rid of this need at random levels...
but 👍 yep, same as existing behavior, LGTM

Groxx (approved):
Yep, looks good!  Thanks for the template-ization, the more we can standardize and pull out of the handlers the better 👍

PR #5640, opened 2024-01-31
Added metrics for when we rate limit

jakobht:
**What changed?**
Add metrics and logs for when we rate limit workflowIDs 

**Why?**
We need to have observability on the workflow ids we rate limit

**How did you test it?**
Tested locally and in unit tests

**Potential risks**
It might cause excessive logging, but it is at most one log (extra) per GRPC call, so I expect it's ok

**Release notes**

**Documentation Changes**


service/history/workflowcache/cache.go
```diff
@@ -133,16 +133,29 @@ func (c *wfCache) allow(domainID string, workflowID string, rateLimitType rateLi
...
+		}
+		return true
 	default:
 		// This should never happen, and we fail open
 		c.logError(domainID, workflowID, errors.New("unknown rate limit type"))
 		return true
 	}
 }
 
+func (c *wfCache) emitMetrics(domainID string, workflowID string, domainName string) {
```
Groxx, line 154: tbh I think I'd name this more `emitLimited...` or similar.  "emit metrics" is completely unclear that it's only for one purpose, and elsewhere we use that kind of name for bulk-flushing aggregate metrics which isn't happening here.

service/history/workflowcache/cache.go
```diff
@@ -133,16 +133,29 @@ func (c *wfCache) allow(domainID string, workflowID string, rateLimitType rateLi
 
 	switch rateLimitType {
 	case external:
-		return value.externalRateLimiter.Allow()
+		if !value.externalRateLimiter.Allow() {
```
Groxx, line 136: since I'm not familiar with these two limits in history: what's the difference?  and is it worth separating them in metrics / likely worth adding to logs?

Groxx:
Log spam: eh, it's probably fine tbh.  Zap samples by default, per log message (the first string arg).  And if we decide we do need something fancier or more control beyond that, we have plenty of options.

Groxx (approved):
Some minor tweaks worth doing I think, but merge when ready :)

service/history/workflowcache/cache.go
jakobht: One is for ratelimiting external calls, one for internal (like the domain global limits) - I have seperated out the metrics and added a tag to the log - please have a look?

service/history/workflowcache/cache_test.go
```diff
@@ -248,3 +248,52 @@ func TestWfCache_CacheDisabled(t *testing.T) {
...
+	// Setup the mock logger
+	logger := new(log.MockLogger)
+
+	logger.On(
+		"Info",
+		"Rate limiting workflowID",
+		[]tag.Tag{
+			tag.WorkflowDomainName(testDomainName),
+			tag.WorkflowID(testWorkflowID),
+		},
```
Groxx, line 281: tests are failing because this is missing the new tags

Groxx (approved):
Tiny thing to fix in the tests, but LGTM once that unblocks :)

PR #5641, opened 2024-01-31
Enable the workflow ID cache in shadow mode for start workflow

jakobht:
**What changed?**
Enable the workflowID cache in shadow mode for startWorkflow only - still disabled by default by dynamic config

**Why?**
Enabling in shadow mode to see the behaviour before enabling fully 

**How did you test it?**
Tested locally 

**Potential risks**
The risk should be small as it is guarded by a dynamic config feature flag

**Release notes**

**Documentation Changes**


service/history/handler.go
```diff
@@ -681,6 +697,10 @@ func (h *handlerImpl) StartWorkflowExecution(
 
 	startRequest := wrappedRequest.StartRequest
 	workflowID := startRequest.GetWorkflowID()
+	if !h.workflowIDCache.AllowExternal(domainID, workflowID) {
```
sankari165, line 700: could you please add a unit test to see the method call. It is quite simple but the test would grow as we start to ratelimit .

sankari165: also a suggestion to skip the if like how we do in frontend.

https://github.com/uber/cadence/blob/master/service/frontend/api/handler.go#L873

service/history/handler.go
```diff
@@ -53,9 +53,14 @@ import (
...
 	"github.com/uber/cadence/service/history/shard"
 	"github.com/uber/cadence/service/history/task"
+	"github.com/uber/cadence/service/history/workflowcache"
 )
 
-const shardOwnershipTransferDelay = 5 * time.Second
+const (
+	shardOwnershipTransferDelay = 5 * time.Second
+	workflowIDCacheTTL          = 1 * time.Second
+	workflowIDCacheMaxCount     = 10_000
```
davidporter-id-au, line 62: [um what? Go supports underscores in numbers?](https://go.dev/ref/spec#:~:text=Integer%20literals,-An%20integer%20literal&text=In%20hexadecimal%20literals%2C%20letters%20a,not%20change%20the%20literal's%20value.) I never new that

davidporter-id-au (approved):

service/history/handler.go
jakobht: Yep, [introduced in 1.13](https://go.dev/doc/go1.13#language). Though honestly I was not consciously aware (just wrote it), if you had asked me yesterday, I'd not have known without trying it out 😅.

service/history/handler.go
jakobht: Added test and removed the if

service/history/handler.go
```diff
@@ -681,6 +697,10 @@ func (h *handlerImpl) StartWorkflowExecution(
 
 	startRequest := wrappedRequest.StartRequest
 	workflowID := startRequest.GetWorkflowID()
+
+	// TODO: actually rate limit
```
3vilhamster, line 701: I would avoid creating TODOs, since you will implement it. Not sure how this will help.

service/history/handler.go
```diff
@@ -681,6 +697,10 @@ func (h *handlerImpl) StartWorkflowExecution(
...
+	h.workflowIDCache.AllowExternal(domainID, workflowID)
```
3vilhamster, line 702: Is metric emission is somewhere inside? 
I expected seeing:
if  h.workflowIDCache.AllowExternal(domainID, workflowID) {
  metricTags = {"domain": domainName, "wftype": WFType}
  if enforced {
     scope.Tagged(metricTags).Counter("ratelimited_workflows").Inc()
     log.Info("rate limited by cache", wfID)
     return nil, Err429
  }
  scope.Tagged(metricTags).Counter("shadow_ratelimited_workflows").Inc()
  log.Info("rate limited but in shadow mode")
}

jakobht: Yeah, it's inside, so we dont have to have metric emission code everywhere

PR #5642, opened 2024-02-05
Implement StartWorkflowExecutionAsync API

Shaddoll:
**What changed?**
Implement StartWorkflowExecutionAsync API

**Why?**
To implement the new feature

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/rpc.go
+++ b/common/rpc.go
@@ -20,7 +20,11 @@
 
 package common
 
-import "go.uber.org/yarpc"
+import (
+	"context"
+
+	"go.uber.org/yarpc"
+)
 
 const (
 	// LibraryVersionHeaderName refers to the name of the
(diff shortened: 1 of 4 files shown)


service/frontend/api/handler.go
```diff
@@ -1737,42 +1739,100 @@ func (wh *WorkflowHandler) StartWorkflowExecutionAsync(
...
+	}
+	// serialize the message to be sent to the queue
+	payload, err := json.Marshal(startRequest)
+	if err != nil {
+		return nil, err
+	}
+	// propagate the headers from the context to the message
+	clientHeaders := common.GetClientHeaders(ctx)
+	header := &shared.Header{
+		Fields: map[string][]byte{},
```
taylanisikdemir, line 1764: size is known so you can do 
```
Fields: make(map[string][]byte, len(clientHeaders)
```

common/rpc.go
```diff
@@ -65,3 +69,31 @@ type (
...
+	if _, ok := headerExists[ClientImplHeaderName]; !ok {
+		headers[ClientImplHeaderName] = call.Header(ClientImplHeaderName)
+	}
+	if _, ok := headerExists[ClientFeatureFlagsHeaderName]; !ok {
+		headers[ClientFeatureFlagsHeaderName] = call.Header(ClientFeatureFlagsHeaderName)
+	}
+	if _, ok := headerExists[ClientIsolationGroupHeaderName]; !ok {
+		headers[ClientIsolationGroupHeaderName] = call.Header(ClientIsolationGroupHeaderName)
+	}
+	return headers
```
taylanisikdemir, line 98: this is an important part that I have some knowledge gap. 
1. do we forward headers to workflows in sync calls?  
2. by doing this allowlist based header selection would the user be disappointed when they don't see some of the custom headers in their workflow's headers?

taylanisikdemir (approved):

common/rpc.go
Shaddoll: These are the headers either forwarded in sync calls or consumed in sync calls.
For example, isolation group header is used in sync calls for zonal isolation.

davidporter-id-au: @Groxx  I believe you were looking at reducing the set of headers that were either persisted or forwarded, waht's the current status of that?

common/rpc.go
```diff
@@ -65,3 +69,31 @@ type (
...
 	}
 )
+
+// GetClientHeaders returns the headers that should be sent from client to server
+func GetClientHeaders(ctx context.Context) map[string]string {
+	call := yarpc.CallFromContext(ctx)
+	headerNames := call.HeaderNames()
+	headerExists := map[string]struct{}{}
+	for _, h := range headerNames {
+		headerExists[h] = struct{}{}
```
davidporter-id-au, line 79: I don't feel strongly about this, but I recall Steven making a point about headers not... technically being a map, since they support duplicates under some implementations.

PR #5643, opened 2024-02-05
Filtering the prefix in custom query log for pinot response comparator

bowenxia:
**What changed?**
Filtering the prefix in custom query log for pinot response comparator

**Why?**
The custom attributes in the log has "`Attr.`" prefix. In order for the Pinot response comparator to directly use the log, we need to get rid of it. 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/pinotVisibilityTripleManager.go
+++ b/common/persistence/pinotVisibilityTripleManager.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"strings"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/dynamicconfig"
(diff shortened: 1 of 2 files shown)


common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -341,7 +342,12 @@ func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userPara
 		tag.WorkflowType(userParam.workflowType),
 		tag.WorkflowID(userParam.workflowID),
 		tag.WorkflowCloseStatus(userParam.closeStatus),
-		tag.VisibilityQuery(userParam.customQuery))
+		tag.VisibilityQuery(filterAttrPrefix(userParam.customQuery)))
+}
+
+func filterAttrPrefix(str string) string {
+	str = strings.Replace(str, "`Attr.", "", -1)
```
taylanisikdemir, line 349: this can unintentionally replace value part in the query. correct way to do this would be to parse the whole query with regex capture groups or AST and then perform the sanitization. however this is probably a super edge case so don't block yourself on this but keep the current behavior caveat documented in the code and also add an example false positive test case to demonstrate it

taylanisikdemir (approved):

common/persistence/pinotVisibilityTripleManager.go
bowenxia: Comment and false positive case added.

PR #5644, opened 2024-02-06
The ratelimiter needs to be created with the domain name not the ID

jakobht:
**What changed?**
The ratelimiter is per domain _name_ not _id_, fixed this, so we can do the filtering correctly

**Why?**
Because we need to filter on the correct value

**How did you test it?**
Tested locally, and updated the unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/workflowcache/cache.go
```diff
@@ -54,7 +54,7 @@ type wfCache struct {
 	domainCache            cache.DomainCache
 	metricsClient          metrics.Client
 	logger                 log.Logger
-	getCacheItemFn         func(domainID string, workflowID string) (*cacheValue, error)
+	getCacheItemFn         func(domainName string, domainID string, workflowID string) (*cacheValue, error)
```
3vilhamster, line 57: Why do we need both domainName and domainID?

3vilhamster (changes requested):
The name is misleading, the change is related to domain name, not workflow name.

service/history/workflowcache/cache.go
jakobht: Good point, we can just use the name in the cache key, should be simpler, fixed

3vilhamster (approved):

PR #5646, opened 2024-02-06
Added 2 more tags in log for comparator to use.

bowenxia:
**What changed?**
Added 2 more tags in log for comparator to use.

**Why?**
We will need user's earliest time and latest time info in the response comparator.
Previously we didn't need it because we were testing the functionality of the comparator. In order to made it stable, we used our own hard coded earliest time and latest time. But this is not useful any more for the testing right now. 

**How did you test it?**
manual testing

**Potential risks**
nah

**Release notes**

**Documentation Changes**


common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -322,6 +323,8 @@ type userParameters struct {
 	workflowID   string
 	closeStatus  int // if it is -1, then will have --open flag in comparator workflow
 	customQuery  string
+	earlistTime  *int64
```
davidporter-id-au, line 326: small typo: `earliestTime`

common/log/tag/tags.go
```diff
@@ -51,6 +51,14 @@ func Timestamp(timestamp time.Time) Tag {
 	return newTimeTag("timestamp", timestamp)
 }
 
+func EarliestTime(time int64) Tag {
```
davidporter-id-au, line 54: if I may suggest making it `*int65`
```suggestion
func EarliestTime(time *int64) Tag {
  if time == nil {
    return newInt64("earliest-time", -1) 
  }
  return newInt64("earliest-time", time)
}
```

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -336,13 +339,24 @@ func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userPara
...
 
-	v.logger.Info("Logging user query parameters for Pinot/ES response comparator...",
-		tag.OperationName(userParam.operation),
-		tag.WorkflowDomainName(userParam.domainName),
-		tag.WorkflowType(userParam.workflowType),
-		tag.WorkflowID(userParam.workflowID),
-		tag.WorkflowCloseStatus(userParam.closeStatus),
-		tag.VisibilityQuery(filterAttrPrefix(userParam.customQuery)))
+	if userParam.earlistTime == nil && userParam.latestTime == nil {
+		v.logger.Info("Logging user query parameters for Pinot/ES response comparator with custom query...",
```
davidporter-id-au, line 343: I'm nitpicking, but I'd suggest pulling the responsibility for handling nil behaviour into the tag (as per the tag discussion) and avoiding branching an if here

common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -452,10 +472,15 @@ func (v *pinotVisibilityTripleManager) GetClosedWorkflowExecution(
 	ctx context.Context,
 	request *GetClosedWorkflowExecutionRequest,
 ) (*GetClosedWorkflowExecutionResponse, error) {
+	earlistTime := int64(0) // this is to get all closed workflow execution
```
davidporter-id-au, line 475: did you want to hardcode these values?

bowenxia: Yes. My understanding for GetClosedWorkflowExecutionRequest is to get all the closed workflow. Am I correct?

davidporter-id-au: I don't think so? Like this should be for a time range?

davidporter-id-au: Discussed offline, Bowen correctly points out that the request struct doesn't have a lower bound (for whatever reason), so this value is not used here.

davidporter-id-au (approved):

PR #5647, opened 2024-02-07
Splitting wfCacheEnabled config for internal and external requests

sankari165:
**What changed?**
WorkflowIDCacheEnabled config flag will be replaced by WorkflowIDCacheExternalEnabled and WorkflowIDCacheInternalEnabled

**Why?**
If the workflow cache is overwhelmed by one type of requests we can choose to enable/dsisable one of them without affecting the other

**How did you test it?**
unit tests

**Potential risks**
The cache is still not in use in production. The old flag WorkflowIDCacheEnabled is still kep to be backwards compatible with the usage

**Release notes**

**Documentation Changes**


sankari165:
@jakobht I did not want to delete it as I didnt want to interrupt the current flow of external requests.

service/history/workflowcache/cache.go
```diff
@@ -135,12 +136,20 @@ func (c *wfCache) allow(domainID string, workflowID string, rateLimitType rateLi
 
 	switch rateLimitType {
 	case external:
+		if !c.workflowIDCacheEnabled(domainName) && !c.workflowIDCacheExternalEnabled(domainName) {
```
taylanisikdemir, line 139: why do we need both `workflowIDCacheEnabled` and `workflowIDCacheExternalEnabled`?

sankari165: I had kept it because workflowIDCacheEnabled was already in use within Uber internal. I have removed it after checking with @jakobht that the usage is minimal and we can replace it

common/dynamicconfig/constants.go
```diff
@@ -1715,7 +1715,13 @@ const (
 	// Value type: Bool
```
jakobht, line 1715: Need to update the Keyname in line 1714 as well

service/history/workflowcache/cache_test.go
```diff
@@ -219,30 +223,86 @@ func TestWfCache_AllowDomainCacheError(t *testing.T) {
...
+	externalLimiterFactory := quotas.NewMockLimiterFactory(ctrl)
+	externalLimiter := quotas.NewMockLimiter(ctrl)
+	externalLimiterFactory.EXPECT().GetLimiter(testDomainName).Return(externalLimiter).Times(1)
+	externalLimiter.EXPECT().Allow().Return(false).Times(1)
+
+	internalLimiter := quotas.NewMockLimiter(ctrl)
+	internalLimiterFactory := quotas.NewMockLimiterFactory(ctrl)
+	internalLimiterFactory.EXPECT().GetLimiter(testDomainName).Return(internalLimiter).Times(1)
+
+	// Setup the cache, we do not need the factories, as we will mock the getCacheItemFn
```
jakobht, line 290: This comment does not apply here I think

service/history/workflowcache/cache_test.go
```diff
@@ -219,30 +223,86 @@ func TestWfCache_AllowDomainCacheError(t *testing.T) {
...
+	externalLimiter := quotas.NewMockLimiter(ctrl)
+	externalLimiterFactory.EXPECT().GetLimiter(testDomainName).Return(externalLimiter).Times(1)
+
+	internalLimiter := quotas.NewMockLimiter(ctrl)
+	internalLimiterFactory := quotas.NewMockLimiterFactory(ctrl)
+	internalLimiterFactory.EXPECT().GetLimiter(testDomainName).Return(internalLimiter).Times(1)
+
+	internalLimiter.EXPECT().Allow().Return(false).Times(1)
 
 	// Setup the cache, we do not need the factories, as we will mock the getCacheItemFn
```
jakobht, line 247: This comment does not apply any more, as we now add the factories

PR #5648, opened 2024-02-08
Rewrite async workflow queue provider component

Shaddoll:
**What changed?**
Rewrite async workflow queue provider component

**Why?**
The old design doesn't support our use case

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


common/asyncworkflow/queue/kafka/decoder.go
```diff
@@ -0,0 +1,50 @@
...
+
+func newDecoder(blob *types.DataBlob) provider.Decoder {
+	return &decoderImpl{
+		blob: blob,
+	}
+}
+
+func (d *decoderImpl) Decode(out any) error {
+	if d.blob.GetEncodingType() != types.EncodingTypeJSON {
+		return fmt.Errorf("unsupported encoding type %v", d.blob.GetEncodingType())
```
taylanisikdemir, line 47: let's mention supported encoding types in the error message

service/frontend/api/producer_manager.go
```diff
@@ -0,0 +1,116 @@
...
+	logger log.Logger,
+	metricsClient metrics.Client,
+) ProducerManager {
+	return &producerManagerImpl{
+		domainCache:   domainCache,
+		provider:      provider,
+		logger:        logger,
+		metricsClient: metricsClient,
+		producerCache: cache.New(&cache.Options{
+			TTL:             0,
```
taylanisikdemir, line 65: having a 5-10 minutes TTL would be need to be able to support queue config changes at runtime and also to avoid keeping these in memory forever

service/frontend/api/producer_manager.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	producer, err := queue.CreateProducer(&provider.Params{Logger: q.logger, MetricsClient: q.metricsClient})
+	if err != nil {
+		return nil, err
+	}
+	// PutIfNotExist is thread safe, and will either return the value that was already in the cache or the value we just created
+	// another thread might have inserted a value between the Get and PutIfNotExist, but that is ok
+	// it should never return an error as we do not use Pin
+	val, err = q.producerCache.PutIfNotExist(queueID, producer)
```
taylanisikdemir, line 111: side comment: a better semantic for this would be `PutIfNotExist(key string, valueCreator func() any)` so we could avoid the potential duplicate initialization problem

taylanisikdemir (approved):

service/frontend/api/producer_manager.go
Shaddoll: that requires rewriting cache implementation, will not do it in this pr

PR #5649, opened 2024-02-08
Store mutable state checksum in SQL storage

Shaddoll:
**What changed?**
Store mutable state checksum in SQL storage

**Why?**
Bug fix

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


3vilhamster:
Could you elaborate on the issue?
So we were not storing side effects data SQL persistence storage layer.
Should we have a integration test that ensures that we store everything for SQL and NonSQL storages?

common/persistence/dataStoreInterfaces.go
```diff
@@ -349,7 +349,8 @@ type (
 		SignalRequestedIDs  map[string]struct{}
 		BufferedEvents      []*DataBlob
 
-		Checksum checksum.Checksum
+		Checksum     checksum.Checksum
+		ChecksumData *DataBlob
```
taylanisikdemir, line 353: Can you please add comments on these fields and explain why we need these 2. At first look it's confusing

common/persistence/executionManager.go
```diff
@@ -112,6 +112,13 @@ func (m *executionManagerImpl) GetWorkflowExecution(
 	newResponse.State.VersionHistories = versionHistories
 	newResponse.MutableStateStats = m.statsComputer.computeMutableStateStats(response)
 
+	if len(newResponse.State.Checksum.Value) == 0 {
+		newResponse.State.Checksum, err = m.serializer.DeserializeChecksum(response.State.ChecksumData)
+		if err != nil {
+			return nil, err
```
taylanisikdemir, line 118: does the err indicate it's checksum deserialization that failed or will it be just "invalid character etc." kind of error? If so maybe wrap the err with more context

common/persistence/executionManager.go
```diff
@@ -635,6 +642,10 @@ func (m *executionManagerImpl) SerializeWorkflowMutation(
 	if err != nil {
 		return nil, err
 	}
+	checksumData, err := m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
+	if err != nil {
+		return nil, err
```
taylanisikdemir, line 647: same here about err message

taylanisikdemir (commented):
is there a mysql integration test that verifies checksum written/read properly?

Shaddoll:
> Could you elaborate on the issue? So we were not storing side effects data SQL persistence storage layer. Should we have a integration test that ensures that we store everything for SQL and NonSQL storages?

Yeah, the checksum is not stored in SQL databases. We have persistence test verifying the storage. But previously, it didn't compare checksum if the checksum read from database is empty.

common/persistence/executionManager.go
Shaddoll: The serializer provides the context.

taylanisikdemir (approved):

PR #5652, opened 2024-02-09
Emit metrics when transfer tasks could be ratelimited

sankari165:
**What changed?**
Adding call to workflowID cache based ratelimiter during transfer task execution

**Why?**
to get metrics on potential hot shards and investigate when requests will be ratelimited

**How did you test it?**
tested locally and added unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


jakobht (approved):

service/history/task/transfer_task_executor_base.go
```diff
@@ -99,6 +103,9 @@ func (t *transferTaskExecutorBase) pushActivity(
 		t.logger.Fatal("Cannot process non activity task", tag.TaskType(task.GetTaskType()))
 	}
 
+	// Ratelimiting is not done. This is only to count the number of requests via metrics
+	t.wfIDCache.AllowInternal(task.DomainID, task.WorkflowID)
```
taylanisikdemir, line 107: This function is called by transfer active task executor and transfer standby task executor. Are we going to count those together in same rate limit quota?

sankari165: Thats a good catch. I have considered to address this in a follow up PR

PR #5654, opened 2024-02-12
Make sure task processing rate limiter is only done in the active side

sankari165:
**What changed?**
Task processing rate limiting metrics is collected only in the active side. 

**Why?**
We do not want to ratelimit task processing in passive side

**How did you test it?**
local and unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


jakobht (approved):

service/history/task/transfer_active_task_executor.go
```diff
@@ -195,6 +196,10 @@ func (t *transferActiveTaskExecutor) processActivityTask(
 	// release the context lock since we no longer need mutable state builder and
 	// the rest of logic is making RPC call, which takes time.
 	release(nil)
+
+	// Ratelimiting is not done. This is only to count the number of requests via metrics
+	t.wfIDCache.AllowInternal(task.DomainID, task.WorkflowID)
```
taylanisikdemir, line 201: we need this in timer active task executor as well

sankari165: It is not needed for timer tasks because we dont plan to rate limit timer tasks

PR #5655, opened 2024-02-12
Async workflow request consumer manager in worker

taylanisikdemir:
**What changed?**
This PR adds implementation of async workflow request consumer manager which starts/stops consumers periodically based on domain configs.

**How did you test it?**
Added new unit tests. Integration tests will be added in follow up PRs.

### DIFF
--- a/common/metrics/defs.go
+++ b/common/metrics/defs.go
@@ -1341,6 +1341,8 @@ const (
 	CheckDataCorruptionWorkflowScope
 	// ESAnalyzerScope is scope used by ElasticSearch Analyzer (esanalyzer) workflow
 	ESAnalyzerScope
+	// AsyncWorkflowConsumerScope is scope used by async workflow consumer
+	AsyncWorkflowConsumerScope
 
 	NumWorkerScopes
 )
@@ -2565,6 +2567,7 @@ const (
 	ESAnalyzerNumStuckWorkflowsRefreshed
 	ESAnalyzerNumStuckWorkflowsFailedToRefresh
 	ESAnalyzerNumLongRunningWorkflows
+	AsyncWorkflowConsumerCount
 
 	NumWorkerMetrics
 )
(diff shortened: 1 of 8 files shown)


common/metrics/defs.go
```diff
@@ -1341,6 +1341,8 @@ const (
 	CheckDataCorruptionWorkflowScope
 	// ESAnalyzerScope is scope used by ElasticSearch Analyzer (esanalyzer) workflow
 	ESAnalyzerScope
+	// AsyncWorkflowConsumerScope is scope used by async workflow consumer
+	AsyncWorkflowConsumerScope
```
Shaddoll, line 1345: the definition is missing

service/worker/asyncworkflow/async_workflow_consumer_manager.go
```diff
@@ -0,0 +1,213 @@
...
+
+		cfg := domain.GetConfig().AsyncWorkflowConfig
+		queue, err := c.getQueue(cfg)
+		if err != nil {
+			c.logger.Error("Failed to get queue", tag.Error(err), tag.WorkflowDomainName(domain.GetInfo().Name))
+			continue
+		}
+
+		if !cfg.Enabled {
+			// lready running active consumers for such queues will be stopped in the next loop
```
Shaddoll, line 165: typo: already

common/metrics/defs.go
taylanisikdemir: thanks for the catch

service/worker/asyncworkflow/async_workflow_consumer_manager.go
Shaddoll, line 1: We're not processing the messages read from consumers. We only start and stop consumers in the consumer manager. Are you planning to introduce another component to processing the messages?

service/worker/service.go
Shaddoll, line 1: It seems that the stop processes for several components of worker are not triggered. But we can fix it in a separate PR.

service/worker/asyncworkflow/async_workflow_consumer_manager.go
taylanisikdemir: yes. left it out to keep PR small

service/worker/service.go
taylanisikdemir: I was going to make all of those components work like new asyn wf consumer but some of those components are not stoppable so require more work to get there. At least the new component shows how it should be done. It should be easy to follow same pattern when we decide to refactor rest.

Shaddoll (approved):

PR #5656, opened 2024-02-12
Update pinot query validator to handle raw time string

neil-xie:
**What changed?**
User can pass in raw date string when call ListWorkflowExecutions, ES can handle it when translate SQL into DSL, Pinot need to handle differently since it is using SQL query.

**Why?**
Handle uncovered case

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


bowenxia (approved):
LGTM

davidporter-id-au (approved):

PR #5658, opened 2024-02-13
Add async workflow request consumer for Start/SignalWithStart support

taylanisikdemir:
**What changed?**
Implemented Async workflow request consumer component which will consume requests from queue and call corresponding frontend API (sync version of the request). It currently support following async request types:
- [StartWorkflowExecutionAsyncRequest](https://github.com/uber/cadence/blob/a614c4c1fd1cf94b3722193cec919985304c9acb/.gen/go/sqlblobs/sqlblobs.go#L2823)
- [SignalWithStartWorkflowExecutionAsyncRequest](https://github.com/uber/cadence/blob/a614c4c1fd1cf94b3722193cec919985304c9acb/.gen/go/sqlblobs/sqlblobs.go#L2824)

**How did you test it?**
Added unit tests. Integration and bench tests are coming soon.


common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -0,0 +1,228 @@
...
+		frontendClient:  frontendClient,
+		ctx:             ctx,
+		cancelFn:        cancelFn,
+		shutdownTimeout: defaultShutdownTimeout,
+		msgDecoder:      codec.NewThriftRWEncoder(),
+		concurrency:     defaultConcurrency,
+	}
+
+	for _, opt := range options {
+		opt(c)
```
davidporter-id-au, line 94: this is an nice pattern

common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -0,0 +1,228 @@
...
+		return
+	}
+
+	c.innerConsumer.Stop()
+	c.logger.Info("Stopped consumer")
+}
+
+func (c *defaultConsumer) runProcessLoop() {
+	defer c.wg.Done()
+
```
davidporter-id-au, line 129: panic recover?

service/worker/asyncworkflow/async_workflow_consumer_manager.go
```diff
@@ -175,17 +178,23 @@ func (c *ConsumerManager) refreshConsumers() {
...
+			MetricsClient:  c.metricsClient,
+			FrontendClient: c.frontendClient,
 		})
 		if err != nil {
 			c.logger.Error("Failed to create consumer", tag.Error(err), tag.WorkflowDomainName(domain.GetInfo().Name), tag.Dynamic("queue-id", queue.ID()))
 			continue
 		}
 
+		if err := consumer.Start(); err != nil {
+			c.logger.Error("Failed to start consumer", tag.Error(err), tag.WorkflowDomainName(domain.GetInfo().Name), tag.Dynamic("queue-id", queue.ID()))
```
davidporter-id-au, line 191: nit: probably worth making a log tag for `queue-id`

common/asyncworkflow/queue/consumer/default_consumer.go
taylanisikdemir: yes I like this because it avoids infinitely growing constructor parameters which don't apply for some callers. also makes testing easier by providing overrides.

common/asyncworkflow/queue/consumer/default_consumer.go
taylanisikdemir: I'm not planning to handle panics here. there's no expected code path that should raise panic so an actual panic should cause crash. If we had a legacy underlying library/code that were raising panics instead of errors then it would make sense to handle it here.

service/worker/asyncworkflow/async_workflow_consumer_manager.go
taylanisikdemir: good point. I noticed I use it in multiple places now. let me turn that into a known tag

common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -0,0 +1,262 @@
...
+	if err := msg.Ack(); err != nil {
+		logger.Error("Failed to ack message", tag.Error(err))
+	}
+	logger.Debug("Processed message successfully")
+}
+
+func (c *DefaultConsumer) processRequest(logger log.Logger, request *sqlblobs.AsyncRequestMessage) error {
+	switch request.GetType() {
+	case sqlblobs.AsyncRequestTypeStartWorkflowExecutionAsyncRequest:
+		startWFReq, err := decodeStartWorkflowRequest(request.GetPayload(), request.GetEncoding())
```
Shaddoll, line 180: After from payload, we also need to parse the headers and set the headers when sending the sync requests to frontend

common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -0,0 +1,262 @@
...
+
+	return throttleRetry.Do(ctx, op)
+}
+
+func decodeStartWorkflowRequest(payload []byte, encoding string) (*types.StartWorkflowExecutionRequest, error) {
+	if encoding != string(common.EncodingTypeJSON) {
+		return nil, &UnsupportedEncoding{EncodingType: encoding}
+	}
+
+	var startRequest types.StartWorkflowExecutionRequest
```
Shaddoll, line 245: The payload is actually the async request. Although it only has one field which is the sync request today, I'm not sure if we'll add other fields to async requests such as retry policy.

common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -0,0 +1,262 @@
...
+			return
+		}
+	}
+}
+
+func (c *DefaultConsumer) processMessage(msg messaging.Message) {
+	logger := c.logger.WithTags(tag.Dynamic("partition", msg.Partition()), tag.Dynamic("offset", msg.Offset()))
+	logger.Debug("Received message")
+
+	sw := c.scope.StartTimer(metrics.ESProcessorProcessMsgLatency)
```
Shaddoll, line 150: should be `AsyncWorkflowProcessMsgLatency`

taylanisikdemir: good catch

Shaddoll (approved):

common/asyncworkflow/queue/consumer/default_consumer.go
davidporter-id-au: I guess crashing the worker's not the worst thing in the world. I get quite afraid of not handling panics with recovery in history, matching and frontend for their tendency to cause cascading failures (host crashes, the shard moves to a new host, repeat)

common/asyncworkflow/queue/consumer/default_consumer.go
taylanisikdemir: done

PR #5659, opened 2024-02-13
Set rate limit on Async APIs

Shaddoll:
**What changed?**
- Introduce new dynamic configurations for rate limit of Async APIs
- Set rate limit on Async APIs

**Why?**
To support new feature

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/frontend/templates/ratelimited.tmpl
+++ b/service/frontend/templates/ratelimited.tmpl
@@ -45,6 +45,18 @@ import (
 {{$ratelimitTypeMap = set $ratelimitTypeMap "ListWorkflowExecutions" "ratelimitTypeVisibility"}}
 {{$ratelimitTypeMap = set $ratelimitTypeMap "ScanWorkflowExecutions" "ratelimitTypeVisibility"}}
 
+{{$ratelimitTypeMap = set $ratelimitTypeMap "StartWorkflowExecutionAsync" "ratelimitTypeAsync"}}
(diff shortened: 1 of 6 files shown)


service/frontend/templates/ratelimited.tmpl
```diff
@@ -45,6 +45,9 @@ import (
 {{$ratelimitTypeMap = set $ratelimitTypeMap "ListWorkflowExecutions" "ratelimitTypeVisibility"}}
 {{$ratelimitTypeMap = set $ratelimitTypeMap "ScanWorkflowExecutions" "ratelimitTypeVisibility"}}
 
+{{$ratelimitTypeMap = set $ratelimitTypeMap "StartWorkflowExecutionAsync" "ratelimitTypeAsync"}}
+{{$ratelimitTypeMap = set $ratelimitTypeMap "SignalWithStartWorkflowExecutionAsync" "ratelimitTypeAsync"}}
```
taylanisikdemir, line 49: is the build or test going to fail when someone adds a new endpoint but misses to specify it here?

taylanisikdemir (approved):

service/frontend/templates/ratelimited.tmpl
Shaddoll: unfortunately, it doesn't. Because we have some endpoint with no limit.

taylanisikdemir: if we add a no-op ratelimiter and associate all those endpoints with it then we can add a check for existence of endpoint in this template. more future proof then current way. we don't have a runbook for "Add new endpoint" so it's easy to miss

PR #5662, opened 2024-02-14
Fix Pinot query validator bug when user pass in not equal query with value missing

neil-xie:
**What changed?**
Pinot can not handle CloseTime != missing case, only handled equal string

**Why?**

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


taylanisikdemir (approved):

bowenxia (approved):
LGTM

davidporter-id-au (approved):

davidporter-id-au:
Would it be useful building a bank of existing test cases from queries to test against?

PR #5663, opened 2024-02-15
Docker compose setup for async workflows with kafka queue

taylanisikdemir:
**What changed?**

- Added new docker compose file which runs all typical Cadence dependencies plus Kafka/Zookeeper. 
- Updated config_template.yaml with a new conditional `asyncWorkflowQueues` section. Only enabled by the new docker compose file via an env var.
- Updated Dockerfile to generate a new image that includes kafka binaries. Only need `kafka-topics.sh` to check during startup for now. The new image is tagged as `ubercadence/server:master-auto-setup-with-kafka`. 

**Why?**
This setup will be handy to try async workflows locally and also to setup CI integration test.

**How did you test it?**
Ran `docker-compose -f docker/docker-compose-async-wf-kafka.yml up` and validated all containers up and running. Verified the topic is created.
A follow up PR will validate async workflows working e2e.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,7 @@ ENTRYPOINT ["/docker-entrypoint.sh"]
 CMD /start-cadence.sh
 
 
-# All-in-one Cadence server
+# All-in-one Cadence server (~450mb)
 FROM cadence-server AS cadence-auto-setup
 
 RUN apk add --update --no-cache ca-certificates py3-pip mysql-client
@@ -95,6 +95,14 @@ COPY docker/start.sh /start.sh
 
 CMD /start.sh
 
+# All-in-one Cadence server with Kafka (~550mb)
+FROM cadence-auto-setup AS cadence-auto-setup-with-kafka
+
+RUN apk add openjdk11
+RUN wget https://archive.apache.org/dist/kafka/2.1.1/kafka_2.12-2.1.1.tgz -O kafka.tgz
+RUN mkdir -p kafka
+RUN tar -xvzf kafka.tgz --strip 1 -C kafka
+ENV KAFKA_HOME /etc/cadence/kafka
 
 # Cadence CLI
 FROM alpine AS cadence-cli
--- a/docker/start.sh
+++ b/docker/start.sh
@@ -155,9 +155,48 @@ wait_for_db() {
     fi
 }
 
+wait_for_async_wf_queue_kafka() {
+    ready="false"
+    while [ "$ready" != "true" ]; do
+        brokers=$(echo dump | nc "$ZOOKEEPER_SEEDS" "$ZOOKEEPER_PORT" | grep brokers | wc -l)
+        if [ "$brokers" -gt 0 ]; then
+            ready="true"
+        else
+            echo 'waiting for kafka broker to show up in zookeeper'
+            sleep 3
+        fi
+    done
+
+    echo 'kafka broker started'
+}
+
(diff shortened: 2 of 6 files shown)


docker/start.sh
```diff
@@ -155,9 +155,48 @@ wait_for_db() {
...
+
+    echo "Kafka topic $ASYNC_WF_KAFKA_QUEUE_TOPIC for async workflows created"
+}
+
 wait_for_db
 if [ "$SKIP_SCHEMA_SETUP" != true ]; then
     setup_schema
 fi
 
+setup_async_wf_queue
```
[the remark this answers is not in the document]
taylanisikdemir: didn't check that. maybe in the code. `wurstmeister/kafka` supports an env var to create the topic as well but didn't want to rely on that because we may switch to a more up to date kafka image. `wurstmeister/kafka` doesn't have kafka 3+. I kept it as is to not deal with kafka upgrade in first iteration

shijiesheng (approved):
tested 
```
root@74af085dad6c:/# $KAFKA_HOME/bin/kafka-topics.sh --describe --topic async-wf-topic1 --zookeeper zookeeper:2181
Topic:async-wf-topic1   PartitionCount:10       ReplicationFactor:1     Configs:
        Topic: async-wf-topic1  Partition: 0    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 1    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 2    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 3    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 4    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 5    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 6    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 7    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 8    Leader: 1001    Replicas: 1001  Isr: 1001
        Topic: async-wf-topic1  Partition: 9    Leader: 1001    Replicas: 1001  Isr: 1001
```
Also another possible optimization is to add another kafka setup image like this stack suggests to avoid a bigger additional cadence server image. 
https://stackoverflow.com/questions/47615751/docker-compose-run-a-script-after-container-has-started

shijiesheng (approved):

Dockerfile
```diff
@@ -95,6 +95,14 @@ COPY docker/start.sh /start.sh
 
 CMD /start.sh
 
+# All-in-one Cadence server with Kafka (~550mb)
+FROM cadence-auto-setup AS cadence-auto-setup-with-kafka
+
+RUN apk add openjdk11
+RUN wget https://archive.apache.org/dist/kafka/2.1.1/kafka_2.12-2.1.1.tgz -O kafka.tgz
```
shijiesheng, line 102: This can diverge from the kafka we use. We may think about using this pattern. https://stackoverflow.com/questions/47615751/docker-compose-run-a-script-after-container-has-started

taylanisikdemir: yeah this is not ideal. I may switch to running a separate kafka init container within docker compose

PR #5667, opened 2024-02-15
Update the build-changed message failure

Groxx:
`make tidy` isn't mentioned, and the command/checks weren't even in the right order anyway.

We should probably streamline pre-test PR checks anyway, but fixing the message is a clear improvement.

### DIFF
--- a/scripts/buildkite/golint.sh
+++ b/scripts/buildkite/golint.sh
@@ -4,16 +4,17 @@ set -ex
 
 make tidy
 make go-generate
+make copyright
 make fmt
 make lint
-make copyright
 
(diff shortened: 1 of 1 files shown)


scripts/buildkite/golint.sh
```diff
@@ -4,16 +4,17 @@ set -ex
 
 make tidy
 make go-generate
+make copyright
 make fmt
```
Groxx, line 8: fmt will fail if copyright is not up to date, so there's not really any point in fmt-then-copyright.

neil-xie (approved):

PR #5670, opened 2024-02-16
Add a `make pr` target for an easy "do automated checks for PR" command

Groxx:
We already have a fair number of quicker and more focused commands, but they can be hard to remember to run / remember when to run.
So now we have one command that should do everything that's reasonably quick: `make pr`
(it still takes a couple minutes)

I've avoided adding tests to this because tests take quite a lot longer and aren't totally isolated / setup-free,
and tbh I tend to just push tests to CI because it's more comprehensive anyway.
Arguably (quick, stable) unit tests should be included though, and if we can figure out a safe set to run it'd be easy to add.

---

This same command will also be added to the client repo, as it has similar needs, and if/when we add submodules
there we'll need a `make tidy` and possibly other things too.  Better to standardize on something simple.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -406,7 +406,7 @@ $Q rm -f $(addprefix $(BUILD)/,$(1))
 $Q +$(MAKE) --no-print-directory $(addprefix $(BUILD)/,$(1))
 endef
 
-.PHONY: lint fmt copyright
+.PHONY: lint fmt copyright pr
 
 # useful to actually re-run to get output again.
 # reuse the intermediates for simplicity and consistency.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -778,7 +792,9 @@ deps-all: ## Check for all dependency updates
 		| sort -n
 
 help:
-	$Q # print help first, so it's visible
+	$Q # print help and PR first, so they're highly visible
 	$Q printf "\033[36m%-20s\033[0m %s\n" 'help' 'Prints a help message showing any specially-commented targets'
+	$Q printf "\033[36m%-20s\033[0m %s\n" 'pr' 'Refresh all code, to ensure your PR can reach tests.  Recommended before opening a github PR.'
+	$Q echo '-----------------------------------'
```
Groxx, line 798: Resulting output:
```
❯ make
help                 Prints a help message showing any specially-commented targets
pr                   Refresh all code, to ensure your PR can reach tests.  Recommended before opening a github PR.
-----------------------------------
bins                 Make all binaries
build                Build all packages and all tests (ensures everything compiles)
clean                Clean build products
copyright            update copyright headers
deps                 Check for dependency updates, for things that are directly imported
deps-all             Check for all dependency updates
fmt                  run gofmt / organize imports / etc
go-generate          run go generate to regen mocks, enums, etc
lint                 (re)run the linter
release              Re-generate generated code and run tests
test                 Build and run all tests. This target is for local development. The pipeline is using cover_profile target
tidy                 go mod tidy all packages
```

davidporter-id-au (approved):
nice, appreciated

PR #5671, opened 2024-02-17
Fix SignalWithStartWorkflow Api

Shaddoll:
**What changed?**
Store request ID to mutable state when handling signalwithstart request

**Why?**
To make SignalWithStartWorkflow idempotent

**How did you test it?**
manual test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/engine/engineimpl/historyEngine.go
+++ b/service/history/engine/engineimpl/historyEngine.go
@@ -2584,6 +2584,10 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(
 				return nil, workflow.ErrSignalsLimitExceeded
 			}
 
+			if requestID := sRequest.GetRequestID(); requestID != "" {
+				mutableState.AddSignalRequested(requestID)
+			}
+
 			if _, err := mutableState.AddWorkflowExecutionSignaled(
 				sRequest.GetSignalName(),
 				sRequest.GetSignalInput(),


service/history/historyEngine.go
```diff
@@ -2584,6 +2584,10 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(
 				return nil, workflow.ErrSignalsLimitExceeded
 			}
 
+			if requestID := sRequest.GetRequestID(); requestID != "" {
```
taylanisikdemir, line 2587: How did you notice this was missing? It's a little bit unclear to me what are the side effects of doing this here. There's already `AddWorkflowExecutionSignaled` below, do these get deduped somehow? 
It would be great if we take this opportunity to add some extra documentation/diagram for SignalWithStart flow.

Shaddoll: I tested it locally and found it's not idempotent in this case.

service/history/engine/engineimpl/historyEngine.go
```diff
@@ -2584,6 +2584,10 @@ func (e *historyEngineImpl) SignalWithStartWorkflowExecution(
 				return nil, workflow.ErrSignalsLimitExceeded
 			}
 
+			if requestID := sRequest.GetRequestID(); requestID != "" {
```
davidporter-id-au, line 2587: I mean, pretty amazingly big fix, nice find. Is this flow covered with a unit-test? If they're not, I think we ought to probably cover all write (+query) APIs to ensure they're idempotent

PR #5672, opened 2024-02-18
Emits a counter value for every unique view of the hashring

davidporter-id-au:
### Context and problem 

The consistent hashring is an eventually consistent way to see the services' members for routing in amongst shards. It's an eventually-consistent system (since actual strong consistency for membership is controlled by the database shard locking). 

However, it's possible that it can be *very* eventually consistent, particularly if the membership resolving system is serving stale data for some reason. This can cause problems because particularly in the case of history, it breaks inter-service communication and routing. 

### Solution

A dumb way to determine if there's an inconsistency therefore, is to just hash the complete membership set and emit it for each host, like a fingerprint. In the healthy case, this random identifier value will quickly converge across hosts. . In the event their views are inconsistent, this will appear as a different guage values which remain persistently different, indicating that operationally some manual operation must be taken. 

### querying the data 

Assuming m3, the trick will be to look for differences between the identifier values. The value itself is just a hash value and arbitrary. Therefore, selecting differences between the upper bound thusly:

```
max = fetch service:cadence* operation:hashring region:region1 service:cadence-history | maxSeries;
min = fetch service:cadence* operation:hashring region:region1 service:cadence-history | minSeries;

-- select where they're different
max | != (min)

-- and emit -1 for the default case where the upper value and lower value are all the same
| transformNull -1
``` 

implies that extended periods of non -1 values are showing a split. 

### Callouts

This does emit metrics at container-level granularity, which might be somewhat higher cardinality for large users.

### DIFF
--- a/common/membership/hashring.go
+++ b/common/membership/hashring.go
@@ -282,7 +282,7 @@ func (r *ring) ring() *hashring.HashRing {
 	return r.value.Load().(*hashring.HashRing)
 }
 
-func (r *ring) emitHashIdentifier() int64 {
+func (r *ring) emitHashIdentifier() float64 {
 	members, err := r.peerProvider.GetMembers(r.service)
 	if err != nil {
 		r.logger.Error("Observed a problem getting peer members", tag.Error(err))
@@ -309,11 +309,12 @@ func (r *ring) emitHashIdentifier() int64 {
 	// the reason for this trimming is that collision is pretty unlikely, and this metric
 	// has only value in emitting that it is different from other hosts, so keeping the
 	// hash space small here
-	trimmedForMetric := int64(hashedView % 100)
+	trimmedForMetric := float64(hashedView % 100)
+	r.logger.Debug("Hashring view", tag.Dynamic("hashring-view", sb.String()), tag.Dynamic("trimmed-hash-id", trimmedForMetric), tag.Service(r.service))
 	r.scope.Tagged(
 		metrics.ServiceTag(r.service),
 		metrics.HostTag(self.identity),
-	).AddCounter(metrics.HashringViewIdentifier, trimmedForMetric)
+	).UpdateGauge(metrics.HashringViewIdentifier, trimmedForMetric)
 	return trimmedForMetric
 }
 
--- a/common/membership/hashring.go
+++ b/common/membership/hashring.go
@@ -22,6 +22,8 @@ package membership
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -33,6 +35,7 @@ import (
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/log"
 	"github.com/uber/cadence/common/log/tag"
+	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/types"
 )
 
@@ -64,6 +67,7 @@ type ring struct {
 	refreshChan  chan *ChangedEvent
 	shutdownCh   chan struct{}
 	shutdownWG   sync.WaitGroup
+	scope        metrics.Scope
 	logger       log.Logger
 
 	value atomic.Value // this stores the current hashring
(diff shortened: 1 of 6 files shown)


common/membership/hashring.go
```diff
@@ -84,21 +88,23 @@ func newHashring(
 	service string,
 	provider PeerProvider,
 	logger log.Logger,
+	scope metrics.Scope,
 ) *ring {
-	hashring := &ring{
+	ring := &ring{
```
davidporter-id-au, line 93: linter was complaiing about a naming collision with the package

common/membership/hashring.go
```diff
@@ -275,6 +282,42 @@ func (r *ring) ring() *hashring.HashRing {
...
 
+func (r *ring) emitHashIdentifier() float64 {
+	members, err := r.peerProvider.GetMembers(r.service)
+	if err != nil {
+		r.logger.Error("Observed a problem getting peer members", tag.Error(err))
+		return -1
+	}
+	self, err := r.peerProvider.WhoAmI()
+	if err != nil {
+		r.logger.Error("Observed a problem self", tag.Error(err))
```
jakobht, line 293: I suggest changing the message to be a bit more meaningful 
Maybe "Observed a problem getting self hash ID" ?

common/membership/hashring.go
```diff
@@ -275,6 +282,42 @@ func (r *ring) ring() *hashring.HashRing {
...
+	var sb strings.Builder
+	for i := range members {
+		sb.WriteString(members[i].addr)
+		sb.WriteString("\n")
+	}
+	hashedView := farm.Hash32([]byte(sb.String()))
+	// trimming here to 100 distinct values, since that should be enough
+	// the reason for this trimming is that collision is pretty unlikely, and this metric
+	// has only value in emitting that it is different from other hosts, so keeping the
+	// hash space small here
```
jakobht, line 311: I don't understand why we want t do this - I understand it's not an issue to do, but I do not understand what the advantage of doing it is?

common/membership/hashring.go
davidporter-id-au: 🤦  yes, I dunno what I was thinking

common/membership/hashring.go
davidporter-id-au: honestly, there's no really good reason, and I maybe should make it greater than 100 or remove it entirely. I did it initially because I was confused (I was thinking of creating series and therefore wanted to limit the cardinality). 

I realised afterwards that I didn't need to do that, but left it just to avoid any overflows in the metric system, since I wanted to keep negative integers separate.

jakobht: Maybe add something to the comment at least, so when I read it in a few months I'll not be confused again.

"Truncating the hash in case the metric system does not work well with too big numbers"?

I don't feel strongly either, but if there is no explanation we'll never be able to change this, because "they must have had a reason to do this".

I guess just having this thread helps 😄

jakobht (approved):

common/membership/hashring.go
davidporter-id-au: Let me fix that commit, it is confusing, you're right

common/membership/hashring.go
```diff
@@ -275,6 +282,41 @@ func (r *ring) ring() *hashring.HashRing {
 	return r.value.Load().(*hashring.HashRing)
 }
 
+func (r *ring) emitHashIdentifier() float64 {
+	members, err := r.peerProvider.GetMembers(r.service)
```
taylanisikdemir, line 286: does `GetMembers` return full list including self? If not we should add it

taylanisikdemir (approved):
since we don't have integration test to validate emitted metrics, did you test locally and checked the emitted metrics?

common/membership/hashring.go
davidporter-id-au: I depled and tested, yes

davidporter-id-au:
> since we don't have integration test to validate emitted metrics, did you test locally and checked the emitted metrics?

I deployed and tested it, yes

common/membership/hashring.go
davidporter-id-au: I don't know to be honest, but also I think a view of the hashring that excludes the current machine is a valid view (if it's drained, for example), and you still want to ensure it's consistent across both drained and undrained zones. 

I would not want to mutate the membership list to add this because it would make a drain case look like a split when it wasn't in this case

taylanisikdemir: If each host sees a list of members that doesn't include themselves then the hashed value will never match. Am I missing something?

davidporter-id-au: Maybe I'm missing something? 

If I'm on host `10.0.0.1` and I have a view of the hashrhing as equivalent to the hashed version of 
```
10.0.0.2
10.0.0.3
```
then, this should be the same as the view for `10.0.0.2` and 3 respectively? The fact that it's not actually on the active list is not a problem?

davidporter-id-au: the only reason I fetched self, was to add it as a dimension for the guage to prevent interleaving of the metric

PR #5678, opened 2024-02-21
Async workflows integration test with kafka

taylanisikdemir:
**What changed?**

Added an integration test to validate StartWorkflowExecutionAsync works. Due to the existing limitations and complicated nature of the integration test framework I had to make some enhancements.

Async WF Consumer side changes:
- Add timeout to context used in StartWorkflowExecution yarpc call because it's mandatory.
- Support for AsyncWorkflowConfig field in domain cache entry deep copy.

Test changes:
- Support per integration test dynamic config overrides. Also support passing different dynamic config overrides per service while creating in-memory cadence cluster.
- Create mock time source and pass it to domain cache and async wf consumer manager. We need this in order to update domain config during test and make sure the changes are reflected in the cache/consumer side.
- Get rid of cadence-auto-setup-with-kafka in favor of docker-compose based kafka topic creation. start.sh no longer needs kafka binaries.

**Why?**
Covering the new async workflows feature via an integration test to see start workflow path works e2e.

**How did you test it?**
1. Stop the previous run if any

`docker-compose -f docker/buildkite/docker-compose-local-async-wf.yml down`

2. Build the integration-test-async-wf image

`docker-compose -f docker/buildkite/docker-compose-local-async-wf.yml build integration-test-async-wf`

3. Run the test in the docker container

`docker-compose -f docker/buildkite/docker-compose-local-async-wf.yml run --rm integration-test-async-wf`

4. Full test run logs can be found at test.log file


common/cache/domainCache.go
```diff
@@ -98,7 +98,7 @@ type (
 		GetCacheSize() (sizeOfCacheByName int64, sizeOfCacheByID int64)
 	}
 
-	domainCache struct {
+	DefaultDomainCache struct {
```
Shaddoll, line 101: why do you export this struct?

host/integrationbase.go
```diff
@@ -267,6 +227,12 @@ func (s *IntegrationBase) registerDomain(
 	})
 }
 
+func (s *IntegrationBase) domainCacheRefresh() {
+	s.testClusterConfig.TimeSource.Advance(cache.DomainCacheRefreshInterval + time.Second)
+	// this sleep is necessary to yield execution to other goroutines. not 100% guaranteed to work
+	time.Sleep(2 * time.Second)
```
Shaddoll, line 233: https://pkg.go.dev/runtime#Gosched
followed by the time.Sleep might work

host/async_wf_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/dynamicconfig"
+	"github.com/uber/cadence/common/persistence"
+	pt "github.com/uber/cadence/common/persistence/persistence-tests"
+	"github.com/uber/cadence/common/types"
+
+	_ "github.com/uber/cadence/common/asyncworkflow/queue/kafka" // needed to load kafka asyncworkflow queue
+)
+
+type AsyncWFIntegrationSuite struct {
```
Shaddoll, line 63: Can you move the definition to `test_suites.go` file?
We need to reuse the definition in monorepo

host/async_wf_test.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+
+			// there's no worker or poller for async workflow, so we just validate whether it started.
+			// this is sufficient to verify the async workflow start path.
+			for i := 0; i < 1000; i++ {
+				timeFilter := &types.StartTimeFilter{
+					EarliestTime: common.Int64Ptr(time.Unix(0, 0).UnixNano()),
+					LatestTime:   common.Int64Ptr(s.testClusterConfig.TimeSource.Now().UnixNano()),
+				}
+				resp, err := s.engine.ListOpenWorkflowExecutions(ctx, &types.ListOpenWorkflowExecutionsRequest{
```
Shaddoll, line 218: Better to use `DescribeWorkflowExecution` API to verify, because it doesn't depend on visibility index.

common/cache/domainCache.go
taylanisikdemir: because that's the right way to do it :) Constructors should return structs not interfaces. Without this I wouldn't be able to expose `WithBlah` options

host/integrationbase.go
taylanisikdemir: Tried and it failed. Even though execution is yielded the test goroutine is still in the list of available goroutines to resume immediately after. Hence `Gosched` approach doesn't guarantee the test execution is not resumed until other pending goroutines have a chance to resume. `time.Sleep` also doesn't guarantee that but reduces the chances drastically.

host/async_wf_test.go
taylanisikdemir: why do we need to define this struct in `test_suites.go`? I was following pinot_test.go mainly

common/types/admin.go
```diff
@@ -497,6 +497,17 @@ type AsyncWorkflowConfiguration struct {
 	QueueConfig         *DataBlob
 }
 
+func (c AsyncWorkflowConfiguration) DeepCopy() AsyncWorkflowConfiguration {
```
Shaddoll, line 500: is it used?

common/cache/domainCache.go
```diff
@@ -718,12 +734,14 @@ func (entry *DomainCacheEntry) duplicate() *DomainCacheEntry {
 		VisibilityArchivalStatus: entry.config.VisibilityArchivalStatus,
 		VisibilityArchivalURI:    entry.config.VisibilityArchivalURI,
 		BadBinaries:              copyResetBinary(entry.config.BadBinaries),
+		AsyncWorkflowConfig:      entry.config.AsyncWorkflowConfig.DeepCopy(),
+		// Q: Should we set IsolationGroups as well? Othewise domaincache will not be able to detect changes in isolation groups
```
taylanisikdemir, line 738: @davidporter-id-au `IsolationGroups` is missed here. Was that intentional? If not I can add it

davidporter-id-au: Good spotting, I... think that's correct, we should add it

taylanisikdemir: adding it in #5679

host/async_wf_test.go
taylanisikdemir: synced offline.

Shaddoll (approved):

PR #5684, opened 2024-02-22
Add metrics for external calls for the workflow ID specific rate limits. 

jakobht:
**What changed?**
Added a wrapper around the history handler which logs rate limit metrics the external calls we want to rate limit with the workflow specific rate limiter.

**Why?**
We want to investigate the rate limiting for all external calls that can cause hot shards, before we deploy.

**How did you test it?**
Unit tests and local tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/wrappers/ratelimited/handler_generated.go
```diff
@@ -0,0 +1,314 @@
...
+	h.workflowIDCache.AllowExternal(hp1.SignalRequest.GetWorkflowExecution().GetWorkflowID(), hp1.GetDomainUUID())
+	return h.wrapped.SignalWorkflowExecution(ctx, hp1)
+}
+
+func (h *historyHandler) Start() {
+	h.wrapped.Start()
+	return
+}
+
+func (h *historyHandler) StartWorkflowExecution(ctx context.Context, hp1 *types.HistoryStartWorkflowExecutionRequest) (sp1 *types.StartWorkflowExecutionResponse, err error) {
```
taylanisikdemir, line 278: This approach would work fine but we will introduce other rate limiting types/signals in the near future. If each different rate limit type requires generating a new layer of handler it will be too much extra code/layers. Did you consider enhancing `rateLimiter` inside handler.go so it can support chained limiters including workflowid one?

jakobht: Hi, I did it like this because we have just introduced wrappers for rate limiters in the frontend. I think making a generalised chain-able rate limiting architecture, where we can target the different methods with different rate limiters, is probably too much scope for the current project. 

I'm not a huge fan of the wrappers TBH. An alternative would be to just write the straightforward code in the methods. IMO that would be much more readable, and easier to discover.

taylanisikdemir: Yeah there's no single best approach. I think sticking with the similar approach as other services (e.g. frontend) do this at least makes it consistent so that makes sense. However, the frontend handler object does not have the `rateLimiter` itself. It delegates that responsibility to [generated wrapper](https://github.com/uber/cadence/blob/master/service/frontend/wrappers/ratelimited/ratelimit.go#L38-L39). 
In this PR, we have workflowid rate limiter as a wrapper but there's still existing rate limiter in history handler. That's why I was thinking one wrapper per rate limiter type is too much code to generate/navigate. If you move the existing history handler's rate limiter to the new wrapper then it would be more consistent. 

Also @Groxx you have been more focused on rate limiters lately and probably have an opinion here. How were you planning to introduce the globabal rate limiter in service layer?

jakobht: Hi, I think we should absolutely move the history host handler to this wrapper, so all rate limiting is collected here, I'd suggest to do that in the next PR?

jakobht: I think we can pull it out in a function, like in front end. Could be easier to test.

Not a huge fan of passing the request type to another function, but it's OK, then we have alignment across the services

taylanisikdemir: Makes total sense to do that part in follow up PR.

taylanisikdemir (approved):

service/history/wrappers/ratelimited/handler_generated.go
```diff
@@ -0,0 +1,314 @@
...
+
+	if hp1 == nil {
+		err = validate.ErrRequestNotSet
```
davidporter-id-au, line 281: out of curiosity, why are these validations done in this layer?

service/history/wrappers/ratelimited/handler_generated.go
```diff
@@ -0,0 +1,314 @@
...
+		err = validate.ErrDomainNotSet
+		return
+	}
+
+	if hp1.StartRequest.GetWorkflowID() == "" {
+		err = validate.ErrWorkflowIDNotSet
+		return
+	}
+
+	h.workflowIDCache.AllowExternal(hp1.StartRequest.GetWorkflowID(), hp1.GetDomainUUID())
```
davidporter-id-au, line 295: I am guessing you're not actually acting on the result of the rate-limit but just emitting a side-effect for visibility, is that correct?

service/history/wrappers/ratelimited/handler_generated.go
jakobht: Mainly cargo culting the wrappers in fronded. So please correct me. 
But it makes sense to me, the wrapper is the first thing that runs, so we need to validate we have the data the wrapper needs, and the wrapper needs the request to get the domain name and the wfid.

We could make a validation wrapper maybe 🤔

I never thought I should say this, but I kinda miss Java Spring for this stuff 😩

service/history/templates/ratelimited.tmpl
```diff
@@ -0,0 +1,89 @@
...
+            {{ $workflowID := ( get ( get $ratelimitTypeMap $method.Name ) "workflowID" ) }}
+
+            if {{(index $method.Params 1).Name}} == nil {
+                err = validate.ErrRequestNotSet
+                return
+            }
+            {{- $domainID := printf "%s.GetDomainUUID()" (index $method.Params 1).Name }}
+            {{- $workflowID := printf "%s.%s" (index $method.Params 1).Name ( $workflowID ) }}
+
+            if {{$domainID}} == "" {
```
3vilhamster, line 74: I suggest moving validation to the workflowIDCache to not clutter it here.

service/history/templates/ratelimited.tmpl
```diff
@@ -0,0 +1,89 @@
...
+func (h *{{$decorator}}) {{$method.Declaration}} {
+    {{- if hasKey $ratelimitTypeMap $method.Name }}
+        {{- if eq
+            ( get ( get $ratelimitTypeMap $method.Name ) "ratelimit" )
+            "ratelimitTypeUserPerID"
+        }}
+
+            {{ $workflowID := ( get ( get $ratelimitTypeMap $method.Name ) "workflowID" ) }}
+
+            if {{(index $method.Params 1).Name}} == nil {
```
3vilhamster, line 67: Maybe, extract this into a var
```suggestion
            {{$req = index $method.Params 1).Name}}
            if {{$req}} == nil {
```
and then later
```
$workflowID := printf "%s.%s" $req $workflowID
```

service/history/templates/ratelimited.tmpl
jakobht: The cache always fail open (if the domainID is not found e.g., so it does not return `err`. I think this validation makes sense here?

davidporter-id-au (approved):
I would consider using a dynaimc-config switch in the rate-limit to rollout rather than just emitting metrics and then adding a second follow-up PR here, as having two PRs might be fairly slow to rollout, but no concerns

service/history/templates/ratelimited.tmpl
3vilhamster: Discussed on slack:
Outcome: Validation here is fine for now, but TODO:
- move metering of successes and errors into a separate wrapper.
- move all ratelimiting into a single wrapper.

3vilhamster:
+1 on David's point. Might be better to introduce a dry run mode and control it with dynamic config.

3vilhamster (approved):

PR #5686, opened 2024-02-22
Write tests for cdb (Cassandra DB wrapper) basic functions

taylanisikdemir:
**What changed?**

Adding tests for `common/persistence/nosql/nosqlplugin/cassandra/db.go`

**Why?**
Increase coverage

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/db.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/db.go
@@ -40,8 +40,11 @@ type cdb struct {
 
 var _ nosqlplugin.DB = (*cdb)(nil)
 
+// cassandraDBOption is used to provide optional settings for cdb object
 type cassandraDBOption func(*cdb)
 
+// dbWithClient returns a cdb option to set the gocql client.
+// If this is not used then the globally registered client is used.
 func dbWithClient(client gocql.Client) cassandraDBOption {
 	return func(db *cdb) {
 		db.client = client
--- a/common/persistence/nosql/nosqlplugin/cassandra/db.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/db.go
@@ -40,15 +40,38 @@ type cdb struct {
 
 var _ nosqlplugin.DB = (*cdb)(nil)
 
+type cassandraDBOption func(*cdb)
+
+func dbWithClient(client gocql.Client) cassandraDBOption {
+	return func(db *cdb) {
+		db.client = client
+	}
+}
+
 // newCassandraDBFromSession returns a DB from a session
-func newCassandraDBFromSession(cfg *config.NoSQL, session gocql.Session, logger log.Logger, dc *persistence.DynamicConfiguration) *cdb {
-	return &cdb{
-		client:  gocql.GetRegisteredClient(),
+func newCassandraDBFromSession(
+	cfg *config.NoSQL,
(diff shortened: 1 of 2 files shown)


common/persistence/nosql/nosqlplugin/cassandra/db.go
```diff
@@ -40,15 +40,38 @@ type cdb struct {
 
 var _ nosqlplugin.DB = (*cdb)(nil)
 
+type cassandraDBOption func(*cdb)
+
+func dbWithClient(client gocql.Client) cassandraDBOption {
```
timl3136, line 45: Seems like this function is only used in tests, I wonder is it good idea to put it in test file instead?

common/persistence/nosql/nosqlplugin/cassandra/db.go
```diff
@@ -40,15 +40,38 @@ type cdb struct {
 
 var _ nosqlplugin.DB = (*cdb)(nil)
 
+type cassandraDBOption func(*cdb)
```
timl3136, line 43: nit: cmiw, this is a future-proof type and not actively used right now. Maybe a comment explaining this type would be nice.

common/persistence/nosql/nosqlplugin/cassandra/db_test.go
```diff
@@ -0,0 +1,278 @@
...
+	logger := testlogger.New(t)
+	dc := &persistence.DynamicConfiguration{}
+
+	db := newCassandraDBFromSession(cfg, session, logger, dc, dbWithClient(client))
+
+	if db.PluginName() != PluginName {
+		t.Errorf("got plugin name: %v but want %v", db.PluginName(), PluginName)
+	}
+
+	client.EXPECT().IsNotFoundError(nil).Times(1)
```
Shaddoll, line 50: missing return value?

common/persistence/nosql/nosqlplugin/cassandra/db.go
```diff
@@ -40,15 +40,38 @@ type cdb struct {
...
 		cfg:     cfg,
 		dc:      dc,
 	}
+
+	for _, opt := range opts {
+		opt(res)
+	}
+
+	if res.client == nil {
+		res.client = gocql.GetRegisteredClient()
```
Shaddoll, line 71: why separate this field from others?

taylanisikdemir: Because RegisterClient is callable only once and this makes it not testable. GetRegisteredClient fails when something is not registered.

common/persistence/nosql/nosqlplugin/cassandra/db.go
taylanisikdemir: used right below as constructor parameter. It's a common pattern but let me add some comments here to make it obvious

common/persistence/nosql/nosqlplugin/cassandra/db_test.go
taylanisikdemir: let me check return values match

common/persistence/nosql/nosqlplugin/cassandra/db.go
taylanisikdemir: it's used in the constructor below. making the contract more extensible for testing

PR #5689, opened 2024-02-23
Global ratelimiter, part 1: core algorithm for computing weights

Groxx:
A high level overview is covered in `common/quotas/global/doc.go`, and this specific piece is further covered in the `./algorithm/requestweighted.go` file.

At a very high level though, this is an isolated piece of a "deployment-wide load-balance-aware ratelimiter", intended to solve problems with our internal clusters with our current high-level frontend ratelimiters around client-triggered domain actions (start workflow, poll, etc).

This is the "logical core" of the whole system, with as few dependencies and concerns as I can manage.  All frontend hosts that will be imposing limits eventually send the update arguments to an aggregating host, and the aggregating host largely just delegates to a single instance of this algorithm.  Essentially all the aggregating host needs to do to respond is multiply the weight by each configured ratelimit (from dynamic config).

I've left the concurrency fairly coarse, both for simplicity and for computational speed. I don't believe this will be a bottleneck in even our largest cluster, but even if it is it should be fairly trivial to split each batch's ratelimit keys into N different sharded instances, which can be completely independent.

# The problem

All of our current ratelimiters currently fall into two conceptual buckets:
1. they enforce N rps on that in-memory instance
2. they enforce N rps on that in-memory instance _divided by the number of hosts in a ring_

While these are very straightforward and have served us well, they're proving inadequate in our internal clusters due to imbalanced load on our frontend instances.  Sometimes _wildly_ imbalanced, e.g. 10% of hosts receiving _all_ traffic for a domain.

What this leads to is a lower _effective_ ratelimit than is intended, e.g.:
- 100 rps is allowed
- 10 hosts exist (so each host allows 10 rps)
- 1 host receives 100 rps, the others receive 0
- the requests are incorrectly limited to 10rps
- stuff slows down, oncalls get paged, nobody is happy

This is leading to a nasty scenario where we tell our users they get N rps, but they can't use it all, so we raise their limits to give them an _effective_ N rps (actually like 3N configured)... but if their requests shift in balance we might allow 3N requests through, possibly risking cluster health.  This is particularly risky for our largest users, as we lose our protection against significant un-planned load increases.

So we need to do something.

# The solution

We can rather clearly divide solutions into either "don't send imbalanced load" or "share information about load", because it's not possible to make correct purely-local decisions.  The first is not an option for us and the second has a variety of tradeoffs.

Within "share information" side we can also divide out two options:
1. Synchronously check on every request.  This allows "perfect" limiting and likely simpler algorithms, but has a non-trivial latency cost (unsuitable for high frequency) and risk (another point of failure, particularly in edge cases or sensitive locations).
2. Something asynchronous.  This is necessarily imprecise, but allows retries and puts no limits on limiter-usage frequency.

For _just_ our frontend requests, we might be willing to go synchronous because the rates are relatively low and the proportional added latency would be very small... but we have imbalanced load in many other places, e.g. due to hot shards causing more requests against the database in a History host, which can reach 100,000 rps.  Having a general tool seems useful, especially if it could cover all of these.

So this is part 1 of building option 2.  Other parts, like "actually share data" and "configure all this", are coming in later PRs.

# The end goal

Ultimately the goal is to have a system which will be ~90% accurate within ~10s regardless of changes, and "idle" at better than 99% for relatively stable / normal usage regardless of distribution.  For this implementation, that implies roughly a 3s update cycle at 0.5 weight (each frontend host pushes aggregate data every 3s, and each update moves data to the average of the previous data and the new data).

Brief, large spikes are essentially out of scope - they will quite possibly perform _worse_ with this system, particularly if they jump between hosts every couple seconds.  These have caused fairly major load spike related issues in the past though, so we have no real interest in allowing them... and we have some async APIs being built that will queue and handle this more smoothly, which will hopefully be a better solution anyway.

Whether this exact implementation survives our internal use remains to be seen.  The plan is to make different algorithms (and different data exchanged) relatively pluggable so we can experiment easily, once the framework is fully in place.

And last but not least, the existing N-rps and N-rps-divided-by-hosts ratelimiters are _not_ going away, and this async-load-balancing limiter will be fully disable-able to go back to using them (and may even forever remain disabled by default).  If your clusters do not have imbalanced request load, you might prefer to use the simpler ones, and they might even behave better.  They are also useful as fallbacks for the global system, in case communication breaks down for some reason.


common/quotas/global/algorithm/requestweighted_test.go
```diff
@@ -0,0 +1,746 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/maps"
+
+	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+// TODO: needs some smaller / simpler tests too
+
```
Groxx, line 42:
```suggestion
```

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+		HostWeights(host Identity, limits []Limit) (weights map[Limit]HostWeight, usedRPS map[Limit]PerSecond)
+
+		// GC can be called periodically to pre-emptively prune old ratelimits.
+		//
+		// Limit keys that are accessed normally will automatically garbage-collect themselves and old host data,
+		// and an unused limit only costs memory, so this may prove to be unnecessary.
+		GC() Metrics
+	}
+
+	Config struct {
```
Groxx, line 122: I'm not sure we need all these to be configurable, but it also seems relatively harmless... but I'm happy to drop any if someone thinks it's not worth the cost.

UpdateRate is I think _essentially_ required as it's the primary "limit check-in times to control CPU cost / contention" control.  I doubt we'll have to change it for that reason (IMO we're more likely to speed it up to make changes converge faster), but it seems like an important safety control to me, and it's sorta important that it matches the caller's rate (it'll be the same key).

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+func missedUpdateScalar(elapsed time.Duration, cfg configSnapshot) (scalar PerSecond, gc bool) {
+	reduce := PerSecond(1.0)
+
+	// fast path: check the bounds for "new enough to ignore"
+	if elapsed <= time.Duration(float64(cfg.rate)*cfg.maxMissed) {
+		return reduce, false // within SLA, no changes
+	}
+	// fast path: check the bounds for "old enough to prune"
+	if elapsed >= cfg.rate*time.Duration(cfg.gcAfter) {
+		return reduce, true
```
Shaddoll, line 325: why return 1.0 in this case? shouldn't it be 0?

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// calculate missed updates, and compute an exponential decay as if we had kept track along the way
+	missed := float64(elapsed) / float64(cfg.rate) // how many missed updates (fractional, 1 == one full duration passed, 2.5 == 2.5 durations passed, etc)
+	if missed >= cfg.maxMissed {
+		// missed at least one update period beyond decay-after, calculate an exponential decay to the old values.
+		// floor to an int when doing so because:
+		// - precision isn't important
+		// - tests are a bit easier (more stable / less crazy-looking values)
+		// - as a bonus freebie: integer exponents are typically faster to compute
+		reduce = PerSecond(math.Pow(cfg.weight, math.Floor(missed)))
```
Shaddoll, line 338: should the weight be `1.0 - cfg.weight`? These are the old data points

Groxx: Yeah, that sounds right.
Lemme double check the math (I'm basically just short-cutting N rounds of zeros) but I think you're correct

common/quotas/global/algorithm/requestweighted.go
Groxx: 0 would be safer, yeah, I'll do that.

I think I just didn't think about it because all gc cases are/should be ignoring the value entirely... but 0 makes that correct either way 👍

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+	// based on how many requests each Identity has received.
+	RequestWeighted interface {
+		// Update load-data for this host's requests, given a known elapsed time spent accumulating this load info.
+		//
+		// Elapsed time must be non-zero, but is not otherwise constrained.
+		Update(host Identity, load map[Limit]Requests, elapsed time.Duration)
+
+		// HostWeights returns the per-[Limit] weights for all requested + known keys for this Identity,
+		// as well as the Limit's overall used RPS (to decide RPS to allow for new hosts).
+		HostWeights(host Identity, limits []Limit) (weights map[Limit]HostWeight, usedRPS map[Limit]PerSecond)
```
Shaddoll, line 113: is host weight used to calculate the limit for that particular host? and is it calculated based on history requests?

common/quotas/global/doc.go
```diff
@@ -0,0 +1,45 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/*
+Package global will contain a global-load-balance-aware ratelimiter.
+
+At a very high level, this will:
+  - collect usage metrics in-memory in "limiting" hosts (which use an in-memory ratelimiter)
+  - asynchronously submit these usage metrics to "aggregating" hosts, which will return per-key RPS data
```
Shaddoll, line 28: do we need to define APIs for submitting the usage metrics?

Shaddoll (commented):
So the assumption of this algorithm is that the requests may not be evenly distributed among the hosts, but the distribution is somewhat stable and does not fluctuate. Is that correct?

davidporter-id-au:
I would add to your comment / update the implicit assumption we're making here is that it's history and the DB which is the thing which is vulnerable, and the frontend hosts themselves rarely need as much protection, so by sharing information around we are attempting to achieve fairness to the customer while protecting the vulnerable thing

common/quotas/global/doc.go
Groxx: at the RPC level, yes - that'll be in a later PR.

my current plan is to put these in History because (only) Frontend doesn't have "send to host in ring" set up.  in principle it doesn't matter where it is though, and I could add a Frontend ring if we think that's worth it, to keep data within the same kinds of services.

common/quotas/global/algorithm/requestweighted_test.go
```diff
@@ -0,0 +1,746 @@
...
+	}
+
+	// init with anything <~1000, too large and even a small fraction of the original value can be too big.
+	push(h1, rand.Intn(1000), rand.Intn(1000))
+	push(h2, rand.Intn(1000), rand.Intn(1000))
+	push(h3, rand.Intn(1000), rand.Intn(1000))
+
+	// now update multiple times and make sure it gets to 90% within 4 steps == 12s (normally).
+	//
+	// 4 steps with 0.5 weight should mean only 0.5^4 => 6.25% of the original influence remains,
```
taylanisikdemir, line 241: nit: picking 0.5 weight makes `1 - weight == weight` and can be misread here. for example 0.5^4 in the comment is actually referring to (1-0.5)^4. 
It makes more sense to give higher weight to new data as we don't want to reject calls as a punishment to high past rps count too much. 
So we can go with default weight=0.7 to address these.
Actual sane number will be determined after real world tests of course.

common/quotas/global/algorithm/requestweighted_test.go
```diff
@@ -0,0 +1,746 @@
...
+	b.Log("nonzero results:", sawnonzero)
+	assert.True(b, b.N < 500 || sawnonzero > 0, "no non-zero result found on a large enough benchmark, likely not benchmarking anything useful")
+}
+
+func TestHashmapIterationIsRandom(t *testing.T) {
+	// is partial hashmap iteration reliably random each time so we can statistically ensure coverage,
+	// or is it relatively fixed per map?
+	// language spec is somewhat vague on the details here, so let's check.
+	//
+	// verdict: yes!  looks random each time.
```
taylanisikdemir, line 701: nice find

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+
+	impl struct {
+		cfg Config
+
+		// mut protects usage, as it is the only mutable data
+		mut sync.Mutex
+		// usage data for ratelimits, per host.
+		//
+		// data is first keyed on the limit and then on the host, as it's assumed that there will be
+		// many times more limit keys than limiting hosts, and this will reduce cardinality more rapidly.
```
taylanisikdemir, line 75: > many times more limit keys than limiting hosts
Currently it's the opposite. Did you think of the workflowid based rate limits?

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+
+		// Expected rate of updates.  Should match the cluster's config for how often limiters check in.
+		UpdateRate dynamicconfig.DurationPropertyFn
+
+		// How many UpdateRate periods can pass without new information before considering a host-limit's
+		// RPS "probably inactive", rather than simply delayed.
+		//
+		// Should always be 1 or larger.  Unsure about a good default (try 2+?), but larger numbers mean smoother
+		// behavior but longer delays on adjusting to hosts that have disappeared or stopped receiving requests.
+		MaxMissedUpdates dynamicconfig.FloatPropertyFn
```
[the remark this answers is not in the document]
Groxx: Yea, I was considering this too.  I've gotta re-think it to make sure it handles everything I need, but it might / at worst I could probably just divide this by `UpdateRate` and infer the same thing (unsure if that'll be clearer or worse tho)

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+
+		// HostWeights returns the per-[Limit] weights for all requested + known keys for this Identity,
+		// as well as the Limit's overall used RPS (to decide RPS to allow for new hosts).
+		HostWeights(host Identity, limits []Limit) (weights map[Limit]HostWeight, usedRPS map[Limit]PerSecond)
+
+		// GC can be called periodically to pre-emptively prune old ratelimits.
+		//
+		// Limit keys that are accessed normally will automatically garbage-collect themselves and old host data,
+		// and an unused limit only costs memory, so this may prove to be unnecessary.
+		GC() Metrics
```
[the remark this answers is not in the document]
Groxx: Yea, I'm not actually planning on calling this at the moment.  I think the at-read-time pruning is likely to be good enough - memory use isn't huge and there's practically zero CPU cost to completely-unused limits because they're not looked at at all.  And if they're _not_ completely-unused, they'll clean up per-host data automatically, and that'll keep CPU use from growing.

I could remove it, it's more of a pre-emptive "I suspect someone might want this eventually, so it should be possible" -> yea it's easy, might as well just build it.

common/quotas/global/algorithm/requestweighted_test.go
Groxx: yep, I was thinking this last night too - definitely agreed, I should use non-0.5 pretty much universally in tests.

common/quotas/global/algorithm/requestweighted.go
Groxx: yea, I think I'll draw an ascii diagram of the whole thing for the docs.  it should help clarify stuff.

for the meantime tho:
- normal calls `Update` to push the data in, then get their `HostWeights` to get the info needed to compute RPS to respond
- yes, changing hosts will change identities, and they'll be empty until they update.  we currently have no way of knowing if they'll actually replace [host X] in a request-routing role so we can't really take advantage of previous data.  that's fine, it's just the empty state which we need to handle anyway.
  - we could return all known keys for an "at startup" request to mitigate this - I've got that built elsewhere actually.  I'm not sure if it's worth doing though, as we still have to handle delays (either in get-rps-response or in accept-requests)
  - alternatively, it can just fall back to the in-memory `RPS/hosts` count for one update cycle.  we need this anyway for safety.
  - this is all also true for aggregating-hosts disappearing or moving, we lose a shard of data like all our other caches.  limiting hosts will keep their last-received rps temporarily to cover these gaps.

common/quotas/global/algorithm/requestweighted.go
taylanisikdemir: I was asking because without workflowid based rate limits in the picture I thought we would have more hosts than limit keys. Limits in the order of endpoint groups (internal, external etc.) was the assumption but just realized it should be in the order of domains * endpoint_groups. So ignore this comment

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+		// - precision isn't important
+		// - tests are a bit easier (more stable / less crazy-looking values)
+		// - as a bonus freebie: integer exponents are typically faster to compute
+		reduce = PerSecond(math.Pow(cfg.weight, math.Floor(missed)))
+	}
+
+	return reduce, false
+}
+
+func weighted[T numeric](newer, older T, weight float64) T {
```
davidporter-id-au, line 344: is this used?

common/quotas/global/doc.go
Groxx: history hosts for persisting frontend ratelimit usage data.  (I just added a lot more to both package docs)

"history storing frontend's use" is legitimately a bit weird, but frontend does not currently have ringpop request-forwarding set up so I just grabbed something else because it doesn't actually matter what hosts store the data.

If we would prefer to add a ring for the frontends though, I think we might find other uses for it, and it would be _very_ simple to change from history to frontend.

common/quotas/global/doc.go
```diff
@@ -41,5 +43,116 @@ to allow shadowing and experimenting with different algorithms at runtime, or di
 the global logic altogether.
 
 See sub-packages for implementation details.
+
```
davidporter-id-au, line 46: this is *super* helpful to understand

common/quotas/global/doc.go
davidporter-id-au: I don't personally have any real concerns with using History as a an in-memory datastore for this, that seems fine. I think elsewhere the naming of one or two variables confused me

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+	// PerSecond represents already-scaled-to-per-second RPS values
+	PerSecond float64
+	// HostWeight is a float between 0 and 1, showing how much of a ratelimit should be allocated to a host.
+	HostWeight float64
+
+	impl struct {
+		cfg Config
+
+		// mut protects usage, as it is the only mutable data
+		mut sync.Mutex
```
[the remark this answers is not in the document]
Groxx: Yea, I even wrote a version around that originally.

I'm reasonably confident it was correct and would *work*, but it was also dramatically more complicated, and very early benchmarks implied it would be slower in synchronous use (which makes sense - it uses TONS of memory barriers).

Whether it would be faster for concurrent use or not: I'm not sure.  I think it would be a reasonable alternative to try (e.g. lost updates will probably be rare enough that we simply wouldn't care about the minor inaccuracies), but a single mutex makes things *super* simple and obvious and the benchmark makes me optimistic it'll be good enough.  Until it's clearly worth doing, I would much prefer to avoid complexity in concurrent code.

If it turns out to be, like, 10% of our CPU or something: yea, it's worth trying alternatives.  With my planned dynamicconfig-to-pick-algorithm, that should be relatively easy to add and switch back and forth in real-world conditions to validate benchmarks.

common/quotas/global/algorithm/requestweighted.go
Groxx: aah, gotcha.  yeah, this is literally a "per [finest-grained thing being limited]" key (1:1 with the `golang.org/x/time/rate.Limiter` instances), so we'll start with several hundred or thousands in basically all of our clusters, and might go a couple orders of magnitude larger eventually.

common/quotas/global/algorithm/requestweighted.go
Groxx: added quite a lot of global/doc.go and docs here - lemme know if that clears this up :)

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+
+	for id := range ih {
+		// scale by the total.
+		// this also ensures all values are between 0 and 1 (inclusive)
+		weights[id] = weights[id] / total
+	}
+	return weights, usedRPS
+}
+
+func (a *impl) HostWeights(host Identity, limits []Limit) (weights map[Limit]HostWeight, usedRPS map[Limit]PerSecond) {
```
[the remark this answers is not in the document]
Groxx: mostly "many", yes.  roughly: `num-domains * 3 / agg-hosts` (3 for user/worker/ES limit categories).

exactly how many will depend on use, but it seemed worth making it a batch-endpoint for reducing total time spent holding the lock, and keeping the calculations fast / avoiding memory barriers in hot paths.

common/quotas/global/algorithm/requestweighted.go
Groxx: it is, yes:
```go
next = history{
	lastUpdate: snap.now,
	accepted:   weighted(aps, prev.accepted*reduce, snap.weight),
	rejected:   weighted(rps, prev.rejected*reduce, snap.weight),
}
```
the extra types in `numeric` are largely to simplify tests, I only use `PerSecond` in this file.

common/quotas/global/algorithm/requestweighted_test.go
```diff
@@ -0,0 +1,746 @@
...
+			time.Sleep(time.Duration(rand.Intn(int(updateRate))))
+			// allow a random number of updates.
+			// should be non-blocking to further encourage racing, when possible
+			for i := rand.Intn(updatesPerBatch); i > 0; i-- {
+				do <- struct{}{}
+			}
+		}
+	}()
+
+	done.Wait()
```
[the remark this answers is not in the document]
Groxx: ah, TIL we have a thing for that.  yea, makes sense here since I don't actually need anything it's protecting, it's just waiting for the [hammer it concurrently] to complete.

I do hope we don't use it outside of tests though :\  using things like that often implies "intentionally allow race conditions"...

common/quotas/global/doc.go
Groxx: this is referring to the `history` struct, for tracking "history of request weight"?
you marked that as resolved so I'm assuming it's no longer a concern, but I do agree that "history" is a special term - I'm game to come up with something else if you think it'd help prevent misunderstandings.

common/quotas/global/doc.go
```diff
@@ -0,0 +1,158 @@
...
+intentionally flexible and not defined here - see individual implementations / sub-packages.
+These will likely evolve as needs change, and will ideally be pluggable and
+dynamicconfig-selectable and shadow-able at runtime without changing this overall structure.
+
+# Planned use
+
+The first use of this is intentionally focused on addressing imbalanced usage
+of per-domain user/worker/ES ratelimits across different frontend hosts, e.g.
+due to unfair load balancing or per-datacenter affinity or other
+occasionally-changing-else-mostly-stable routing decisions.
```
Groxx, line 100:
```suggestion
occasionally-changing-else-mostly-stable routing decisions.
This will essentially imply 3 keys per domain, and will look something like
`domain-user-request` / `domain-worker-request` / etc - they just have to be
unique per conceptual thing-with-a-configured-limit.
```

common/quotas/global/doc.go
```diff
@@ -0,0 +1,158 @@
...
+
+"Aggregating" hosts will internally:
+  - receive data for key X from all "limiting" hosts
+  - compute request-weight per host+key in the cluster
+  - return the "limiting"-host's weighted portion of each ratelimit to allow
+
+The exact logic that the "aggregating" hosts use is intentionally hidden from the
+"limiting" hosts, so it can be changed without changing how they enforce limits.
+
+The weight-calculation algorithm will (eventually) be controlled by a dynamic config value,
```
Groxx, line 41:
```suggestion
The weight-calculation algorithm will (eventually) be controlled by a dynamicconfig value,
```

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,431 @@
...
+	Identity string
+	// Limit is the key being used to Limit requests.
+	Limit string
+	// PerSecond represents already-scaled-to-per-second RPS values
+	PerSecond float64
+	// HostWeight is a float between 0 and 1, showing how much of a ratelimit should be allocated to a host.
+	HostWeight float64
+
+	impl struct {
+		cfg Config
```
Groxx, line 148:
```suggestion
		// intentionally value-typed so it copies fields from caller - manually copy if this changes.
		cfg Config
```

common/quotas/global/doc.go
```diff
@@ -0,0 +1,161 @@
...
+	└┬─────┬─────┬┘
+	 │     │     │
+	 a1    b1    c2     (each subset of keys is sent to its owning agg host)
+	 │     │     │
+	 │     │     │  ┌─────── other "c" usage from other frontends, etc
+	┌▽───┐┌▽───┐┌▽──▽┐
+	│agg1││agg2││agg3│  (aggs compare frontend's usage per key with others, decide on RPS to allow)
+	└┬───┘└┬───┘└┬───┘
+	 │     │     │
+	a/2   b/4   c3/4    (aggs return weighted RPS to allow)
```
davidporter-id-au, line 74: ensuring I'm interpreting this correctly, this is a push or pull model? Do the frontend hosts poll the history host aggregators or does the aggregator reach out with a notion of subscription?

common/quotas/global/doc.go
davidporter-id-au: I'd consider adding a prefix to the struct, but htat's not super consequential, the diagram has helped. 

Regarding request-forwarding: Using history as a means of in-memory aggregation seems perfectly reasonable, but I don't think you need request forwarding (also, I am not sure it's available in the uns implementation?)

Groxx: we need all hosts receiving requests for key X to meet in one place to exchange data, since there's no other way to figure out load distribution.  I'm not sure if that fits with "request forwarding" or not, but that's all it is.

hashing alone is fine to achieve that, but the consistent hash ring means less data loss on host-collection changes (i.e. not "almost everything disappears immediately").  and the UNS implementation supports this just fine, else e.g. matching would be unable to function for handling workflow-query requests.

common/quotas/global/doc.go
Groxx: frontends poll / repeatedly call history aggregators, those are the only requests.

I was trying to figure out how to make that clear from the diagram, but kinda gave up.  everything I can come up with needs more width to draw return-path lines, and that just gets worse and worse to render and display :\

common/quotas/global/algorithm/requestweighted_test.go
Groxx: I was honestly kinda surprised that it works even on the same instance.  lots of languages that have randomized maps do it with either an at-startup seed (that never changes) or a per-instance seed - Go apparently re-randomizes _every time_, which is almost certainly more costly.

I wholeheartedly approve, I like the implementation-specific-reliance-resistance this brings.  just mildly surprised they actually did it.

common/quotas/global/doc.go
Groxx: did a plain sequence diagram, and I think it might be better (updating in a moment)

common/quotas/global/doc.go
Groxx: re-did the diagram btw, how does this one look?

common/quotas/global/doc.go
```diff
@@ -0,0 +1,161 @@
...
+	  │      [b/4, c3/4] │    │ │   │ │
+	  │<────────────────────────────└┬┘
+	  │                  │    │ │    │
+	  │            [a/2] │    │ │    │
+	  │<──────────────────────└┬┘    │
+	  │                  │     │     │
+	┌─┴─┐              ┌─┴──┐┌─┴──┐┌─┴──┐
+	│lim│              │ring││agg1││agg2│
+	└───┘              └────┘└────┘└────┘
+
```
Groxx, line 79: diagram 2: sequence boogaloo.

renders very nicely in goland (yay correct monospace font usage):
Screenshot 2024-02-26 at 14 42 35

and godocs looks fine too (we'll have to see how the public site looks, as that's not open source and runnable AFAIK):
Screenshot 2024-02-26 at 14 42 17

and of course vscode is useful as ever:
Screenshot 2024-02-26 at 14 51 41

though if you choose a different-but-same-thing, you can get package docs from vscode:
Screenshot 2024-02-26 at 14 51 59

common/quotas/global/algorithm/requestweighted.go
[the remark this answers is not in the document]
Groxx: eh, I think I'll add a per-minute call just to emit metrics so we have some "how big is the thing" information.  should be trivial to add to later PRs.

3vilhamster:
Not sure if you've tried/looked, but I've recently found https://github.com/mennanov/limiters.
It supports multiple storage engines and has different implementations already.

That comes with some limitations of external db requirements, but it might be easier.
But that will support any kind of bursty traffic, given that we don't hit the limit of db itself.

davidporter-id-au (approved):

taylanisikdemir (approved):
Changes LGTM and we will probably iterate further as rest of the layers that depend on this get implemented. Only leftover comment from my side is `MaxDelayDuration` vs `UpdateRate` which would improve reasonability of the algo implementation.

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,442 @@
...
+triggered the request.
+
+If there are unused RPS remaining, the aggregating host *may* increase the RPS
+it returns by [some amount], to allow the limiting host to pre-emptively allow
+more requests than is "fair" before the next update.  Even if it does not, an
+increase in attempted usage will increase that limiter's weight on the next
+update cycle, so this is mostly intended as a tool for reducing incorrectly-rejected
+requests when a ratelimit's usage is well below its allowed limit.
+
+# Dealing with expired data
```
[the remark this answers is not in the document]
Groxx: yeah, the final goal for all this is to let us experiment internally while reusing the overall sharding / request flow / etc, since internally we have information that doesn't exist in other environments.

which is probably true of other environments too.  "customized" seems quite likely in general, though I do expect a majority will just have round-robin or random or something and won't need this.

common/quotas/global/algorithm/requestweighted.go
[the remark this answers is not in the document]
Groxx: yep, durations everywhere feel pretty good.  gonna stick with that.

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,351 @@
...
+func (a *impl) GC() Metrics {
+	a.mut.Lock()
+	defer a.mut.Unlock()
+
+	m := Metrics{}
+	snap := a.snapshot()
+	// TODO: too costly? can check the first-N% each time and it'll eventually visit all keys, demonstrated in tests.
+	for lim, dat := range a.usage {
+		for host, hist := range dat {
+			if _, gc := missedUpdateScalar(snap.now.Sub(hist.lastUpdate), snap); gc {
```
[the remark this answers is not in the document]
Groxx: results so far look like (while minimizing wasted calcs):
- sadly a bit larger and more "you have to know to check it" rather than having all things you need in one call
- reduces one unnecessary `math.Pow`

... not entirely sure how I feel about that tradeoff, but feels defensible anyway.
I'll push an update with that soon.

common/quotas/global/algorithm/requestweighted.go
```diff
@@ -0,0 +1,442 @@
...
+
+	"golang.org/x/exp/constraints"
+
+	"github.com/uber/cadence/common/clock"
+	"github.com/uber/cadence/common/dynamicconfig"
+)
+
+type (
+	// history holds the running per-second running average for a single key from a single host, and the last time it was updated.
+	history struct {
```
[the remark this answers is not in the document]
Groxx: yep, agreed!  changed it to `requests` :)

Groxx:
> Not sure if you've tried/looked, but I've recently found https://github.com/mennanov/limiters. It supports multiple storage engines and has different implementations already.
> 
> That comes with some limitations of external db requirements, but it might be easier. But that will support any kind of bursty traffic, given that we don't hit the limit of db itself.

I did finally check that + the code, and AFAICT they _all_ require a request per limit check / they're synchronous checks, not asynchronous.

Might be worth checking if we decide to toss this and use an external storage and hammer it super hard, but that's not really the goal at the moment.

PR #5691, opened 2024-02-23
Enable retry on mutable state checksum verification failure

Shaddoll:
**What changed?**
- Enable retry on mutable state checksum failure to reload the mutable state from database
- Improve logging on mutable state checksum failure

**Why?**
In the past 3 months, we're bothered by corrupted workflows from our internal sql database. And we suspect that it's because the mutable state we read from database under some edge case is from an inconsistent view of database. We enabled checksum and verified that there is some checksum failures in production, but we don't have the details in the logs and still don't know the root cause.
We add a retry mechanism and hope this will temporarily fix the issue.

**How did you test it?**

**Potential risks**
this is protected by feature flag, we can disable the feature flag if there is any issue

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/context.go
+++ b/service/history/execution/context.go
@@ -27,6 +27,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -45,6 +46,7 @@ import (
 
 const (
 	defaultRemoteCallTimeout = 30 * time.Second
+	checksumErrorRetryCount  = 3
 )
 
 type conflictError struct {
@@ -252,6 +254,10 @@ func (c *contextImpl) LoadExecutionStats(
 	return c.stats, nil
 }
 
+func isChecksumError(err error) bool {
+	return strings.Contains(err.Error(), "checksum mismatch error")
+}
+
 func (c *contextImpl) LoadWorkflowExecutionWithTaskVersion(
 	ctx context.Context,
 	incomingVersion int64,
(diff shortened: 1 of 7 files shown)


service/history/execution/context.go
```diff
@@ -263,22 +264,27 @@ func (c *contextImpl) LoadWorkflowExecutionWithTaskVersion(
...
-			domainEntry,
-		)
+			c.mutableState = NewMutableStateBuilder(
+				c.shard,
+				c.logger,
+				domainEntry,
+			)
 
-		c.mutableState.Load(response.State)
+			if err := c.mutableState.Load(response.State); err == nil {
```
taylanisikdemir, line 284: let's check err type and retry if it's checksum failure. I know that's only case `Load` will return err today but better to structure the code to be future proof.

service/history/execution/mutable_state_builder.go
```diff
@@ -357,10 +358,23 @@ func (e *mutableStateBuilder) Load(
 				// feature is tested and/or we have mechanisms in place to deal
 				// with these types of errors
 				e.metricsClient.IncCounter(metrics.WorkflowContextScope, metrics.MutableStateChecksumMismatch)
-				e.logError("mutable state checksum mismatch", tag.Error(err))
+				e.logError("mutable state checksum mismatch",
```
taylanisikdemir, line 361: I know this file doesn't have good enough test coverage already but these new lines will drop new line coverage metric significantly. Can you add some basic tests to cover these lines? we can expand on the coverage later on

taylanisikdemir (approved):

PR #5692, opened 2024-02-26
Added a test for the GetworkflowExecution in the nosql_execution_store.go file.

agautam478:
**What changed?**
To reduce duplication and improve the test file, I refactored the common setup into helper functions. This refactoring will focus on making the code more maintainable and concise.
Added two new test cases for the `GetworkflowExecution`.

I will add 5-6 more cases for this in the upcoming PR along with a test for UpdateWorkflowExecution.

**Why?**
To increase the code coverage.

**How did you test it?**
unit test

**Potential risks**
NA

**Release notes**
NA

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosql_execution_store_test.go
+++ b/common/persistence/nosql/nosql_execution_store_test.go
@@ -34,45 +34,37 @@ import (
 	"github.com/uber/cadence/service/history/constants"
 )
 
-func TestCreateWorkflowExecution(t *testing.T) {
+func TestNosqlExecutionStore(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	mockDB := nosqlplugin.NewMockDB(ctrl)
 	logger := log.NewNoop()
-
-	store := &nosqlExecutionStore{
-		shardID:    1,
-		nosqlStore: nosqlStore{logger: logger, db: mockDB},
-	}
+	store := newTestNosqlExecutionStore(mockDB, logger)
 
 	ctx := context.Background()
-	request := &persistence.InternalCreateWorkflowExecutionRequest{
-		RangeID:                  123,
-		Mode:                     persistence.CreateWorkflowModeBrandNew,
-		PreviousRunID:            "previous-run-id",
-		PreviousLastWriteVersion: 456,
-		NewWorkflowSnapshot:      getNewWorkflowSnapshot(),
-	}
 	testCases := []struct {
 		name          string
 		setupMock     func()
-		request       *persistence.InternalCreateWorkflowExecutionRequest
-		expectedError error
+		testFunc      func() error // Use a function to encapsulate test execution
+		expectedError interface{}
 	}{
 		{
-			name: "success",
+			name: "CreateWorkflowExecution success",
 			setupMock: func() {
 				mockDB.EXPECT().
 					InsertWorkflowExecutionWithTasks(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
 					Return(nil).Times(1)
 			},
-			request:       request,
+			testFunc: func() error {
+				_, err := store.CreateWorkflowExecution(ctx, newCreateWorkflowExecutionRequest())
+				return err
+			},
 			expectedError: nil,
 		},
 		{
-			name: "failure - workflow already exists",
-			setupMock: func() {
+			name: "CreateWorkflowExecution failure - workflow already exists",
+			testFunc: func() error {
 				mockDB.EXPECT().
 					InsertWorkflowExecutionWithTasks(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
 					Return(&persistence.WorkflowExecutionAlreadyStartedError{}).Times(1)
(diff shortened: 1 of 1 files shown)


common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -34,98 +34,114 @@ import (
...
 		name          string
 		setupMock     func()
-		request       *persistence.InternalCreateWorkflowExecutionRequest
-		expectedError error
+		testFunc      func() error // Use a function to encapsulate test execution
+		expectedError interface{}
 	}{
 		{
-			name: "success",
+			name: "CreateWorkflowExecution success",
```
taylanisikdemir, line 53: We should create `mockDB` and `store` per test case otherwise the proper isolation between test cases is not there. if one test fails for legitimate code change reasons and doesn't call the expected mock function then mock expectations will shift for subsequent tests. So let's make `mockDB` parameter to `testSetup` and create `store in the loop below not at the top.

agautam478: will do

common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -34,108 +34,179 @@ import (
...
 		expectedError error
 	}{
 		{
-			name: "success",
-			setupMock: func() {
+			name: "CreateWorkflowExecution success",
+			setupMock: func(ctrl *gomock.Controller) *nosqlExecutionStore {
+				mockDB := nosqlplugin.NewMockDB(ctrl)
 				mockDB.EXPECT().
 					InsertWorkflowExecutionWithTasks(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
```
davidporter-id-au, line 51: I don't think there's a huge amount of value in this kind of test unless we do some assertion on what's being passed to the DB (I don't know this function particularly well). I get that here though you're just adding to a prexxisting setup

common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -34,108 +34,179 @@ import (
...
 		{
-			name: "failure - workflow already exists",
-			setupMock: func() {
+			name: "CreateWorkflowExecution failure - workflow already exists",
+			setupMock: func(ctrl *gomock.Controller) *nosqlExecutionStore {
+				mockDB := nosqlplugin.NewMockDB(ctrl)
 				mockDB.EXPECT().
 					InsertWorkflowExecutionWithTasks(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
 					Return(&persistence.WorkflowExecutionAlreadyStartedError{}).Times(1)
 				mockDB.EXPECT().IsNotFoundError(gomock.Any()).Return(false).AnyTimes() // Assuming the error is not a "not found" error
```
davidporter-id-au, line 68: Are these ever called?

common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -34,108 +34,179 @@ import (
...
+			expectedError: nil,
+		},
+		{
+			name: "GetWorkflowExecution failure - not found",
+			setupMock: func(ctrl *gomock.Controller) *nosqlExecutionStore {
+				mockDB := nosqlplugin.NewMockDB(ctrl)
+				mockDB.EXPECT().
+					SelectWorkflowExecution(ctx, shardID, gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(nil, &types.EntityNotExistsError{}).Times(1)
+				mockDB.EXPECT().IsNotFoundError(gomock.Any()).Return(false).AnyTimes() // Assuming the error is not a "not found" error
```
davidporter-id-au, line 159: this is a not-found error?

agautam478: Removed the comment. I was fixing something and accidentally left it there.

common/persistence/nosql/nosql_execution_store_test.go
agautam478: There is some common call that is invoked somewhere in the code
`func convertCommonErrors(errChecker nosqlplugin.ClientErrorChecker, operation string, err error) error {
	if errChecker.IsNotFoundError(err) {
		return &types.EntityNotExistsError{
			Message: fmt.Sprintf("%v failed. Error: %v ", operation, err),
		}
	}

	if errChecker.IsTimeoutError(err) {
		return &persistence.TimeoutError{Msg: fmt.Sprintf("%v timed out. Error: %v", operation, err)}
	}

	if errChecker.IsThrottlingError(err) {
		return &types.ServiceBusyError{
			Message: fmt.Sprintf("%v operation failed. Error: %v", operation, err),
		}
	}

	if errChecker.IsDBUnavailableError(err) {
		return &persistence.DBUnavailableError{
			Msg: fmt.Sprintf("%v operation failed. Error: %v", operation, err),
		}
	}`
 that makes mocking these calls mandatory.
 
 I have tried moving these to pure packages but figured  I would do this in the next PR with extra cases for getworkflow.

common/persistence/nosql/nosql_execution_store_test.go
agautam478: yes! Adding to a preexisting setup just to increase coverage.

common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -34,108 +34,179 @@ import (
...
+			expectedError: nil,
+		},
+		{
+			name: "GetWorkflowExecution failure - not found",
+			setupMock: func(ctrl *gomock.Controller) *nosqlExecutionStore {
+				mockDB := nosqlplugin.NewMockDB(ctrl)
+				mockDB.EXPECT().
+					SelectWorkflowExecution(ctx, shardID, gomock.Any(), gomock.Any(), gomock.Any()).
+					Return(nil, &types.EntityNotExistsError{}).Times(1)
+				mockDB.EXPECT().IsNotFoundError(gomock.Any()).Return(false).AnyTimes()
```
davidporter-id-au, line 159: I mean this is going to pass because it's mocked, but this I think should return true?

agautam478: fixed!

davidporter-id-au (approved):

PR #5697, opened 2024-03-01
Do not get workflow execution from database when shard is closed

Shaddoll:
**What changed?**
Do not get workflow execution from database when shard is closed

**Why?**
When a shard is closed, we should stop all the operations to the database from that shard. Read operation was allowed because it is safe. But since we do a checksum validation after GetWorkflowExecution from database, a false negative result might be returned from the validation if we allow reading data after a shard is closed.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -107,6 +107,7 @@ type (
 		GetDomainNotificationVersion() int64
 		UpdateDomainNotificationVersion(domainNotificationVersion int64) error
 
+		GetWorkflowExecution(ctx context.Context, request *persistence.GetWorkflowExecutionRequest) (*persistence.GetWorkflowExecutionResponse, error)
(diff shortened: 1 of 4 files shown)


service/history/shard/context.go
```diff
@@ -578,6 +579,16 @@ func (s *contextImpl) UpdateTimerMaxReadLevel(cluster string) time.Time {
 	return s.timerMaxReadLevelMap[cluster]
 }
 
+func (s *contextImpl) GetWorkflowExecution(
+	ctx context.Context,
+	request *persistence.GetWorkflowExecutionRequest,
+) (*persistence.GetWorkflowExecutionResponse, error) {
+	if s.isClosed() {
+		return nil, ErrShardClosed
```
taylanisikdemir, line 587: great find

PR #5702, opened 2024-03-02
[code-coverage] update admin and frontend client to use generated code

ketsiambaku:
**What changed?**
In this PR admin & frontend client wrapper is generated using gowrap
history was generated in a previous PR #5728 however, I am adding to/modifiing the template so it can generate code for the other decorators as well.
Will have an follow up PR for matching client because of the multiple special cases :)

**Why?**
To improve code coverage
To reduce manual work when new methods are added in the future

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


client/admin/interface.go
```diff
@@ -34,6 +34,7 @@ import (
 //go:generate gowrap gen -g -p . -i Client -t ../templates/errorinjectors.tmpl -o ../wrappers/errorinjectors/admin_generated.go -v client=Admin
 //go:generate gowrap gen -g -p . -i Client -t ../templates/grpc.tmpl -o ../wrappers/grpc/admin_generated.go -v client=Admin -v package=adminv1 -v path=github.com/uber/cadence-idl/go/proto/admin/v1 -v prefix=Admin
 //go:generate gowrap gen -g -p . -i Client -t ../templates/thrift.tmpl -o ../wrappers/thrift/admin_generated.go -v client=Admin -v prefix=Admin
+//go:generate gowrap gen -g -p . -i Client -t ../templates/tchannel.tmpl -o ../wrappers/tchannel/admin_generated.go -v client=Admin
```
3vilhamster, line 37: Could you elaborate on why it is tchannel? :)
As far as I understand, it is an abstraction around thrift or grpc clients.
Not sure what would be the best name here.
timed? timed_out?

ketsiambaku: The naming was based off of this [comment](https://github.com/uber/cadence/blob/7fddc8984345048edb9f444411a9c0524ad4a4bf/client/frontend/client.go#L47)  under the constructors. However this is old code maybe previously used as a tchannel wrapper. I like timed better

3vilhamster: As far as I see @shijiesheng called it timeout wrapper in https://github.com/uber/cadence/pull/5728
Lets rename it to timeout then.

ketsiambaku: done

client/templates/timeout.tmpl
```diff
@@ -1,38 +1,72 @@
...
+    {{$fieldMap = merge $fieldMap (dict "timeout" "time.Duration" "client" .Interface.Type "longPollTimeout" "time.Duration") }}
+{{ else if eq $ClientName "Matching" }}
+    {{$fieldMap = merge $fieldMap (dict "timeout" "time.Duration" "client" .Interface.Type "longPollTimeout" "time.Duration" "peerResolver" "matching.PeerResolver" "loadBalancer" "matching.LoadBalancer") }}
+{{ end }}
+
 import (
-  "context"
-  "time"
+    "context"
+    "time"
```
Groxx, line 26: it's relatively minor but: see if you can switch your IDE to use tabs instead of spaces?  fewer changes.

client/wrappers/timeout/timeout.go
```diff
@@ -0,0 +1,49 @@
...
+	AdminDefaultTimeout = 10 * time.Second
+	// AdminDefaultLargeTimeout is the admin service large default timeout used to make calls
+	AdminDefaultLargeTimeout = time.Minute
+	// FrontendDefaultTimeout is the frontend service default timeout used to make calls
+	FrontendDefaultTimeout = 10 * time.Second
+	// FrontendDefaultLongPollTimeout is the frontend service long poll default timeout used to make calls
+	FrontendDefaultLongPollTimeout = time.Minute * 3
+)
+
+func createContext(parent context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
```
ketsiambaku, line 41: since all client (history/matching/frontend/admin) are using the same logic to create context, I seperated this in one function that they all share

client/wrappers/timeout/timeout.go
```diff
@@ -0,0 +1,49 @@
...
+	// AdminDefaultTimeout is the admin service default timeout used to make calls
+	AdminDefaultTimeout = 10 * time.Second
+	// AdminDefaultLargeTimeout is the admin service large default timeout used to make calls
+	AdminDefaultLargeTimeout = time.Minute
+	// FrontendDefaultTimeout is the frontend service default timeout used to make calls
+	FrontendDefaultTimeout = 10 * time.Second
+	// FrontendDefaultLongPollTimeout is the frontend service long poll default timeout used to make calls
+	FrontendDefaultLongPollTimeout = time.Minute * 3
+)
+
```
ketsiambaku, line 40: Since all client decorators are using the same logic to create context, I removed this from the template and into a function that they can all share

client/templates/timeout.tmpl
ketsiambaku: Noted

client/wrappers/timeout/timeout.go
ketsiambaku, line 41: Since all client decorators are using the same logic to create context, I removed this from the template and into a function that they can all share

3vilhamster (approved):
Looks nice :)

PR #5704, opened 2024-03-04
Implemented ratelimiting for external calls pr wfid (guarded by feature flag)

jakobht:
**What changed?**
Implemented the actual rate limiting on the pr workflowID project 

**Why?**
We can now roll it out and test it without needed to wait for rollouts

**How did you test it?**
Tested locally and with unit tests

**Potential risks**
Should be very low as it is guarded by a feature flag

**Release notes**

**Documentation Changes**


service/history/service.go
```diff
@@ -114,7 +114,13 @@ func (s *Service) Start() {
 	})
 
 	rawHandler := handler.NewHandler(s.Resource, s.config, wfIDCache)
```
taylanisikdemir, line 116: Will the rate limited handler contain other limit enforcements that are currently in handler? Just making sure we are on the same page because that was my understanding based on [this discussion](https://github.com/uber/cadence/pull/5684#discussion_r1500922598).

service/history/wrappers/ratelimited/ratelimit_test.go
```diff
@@ -0,0 +1,106 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+package ratelimited
```
taylanisikdemir, line 23: Writing a new integration test would be worth for rate limits, especially for the wfid based one. You can check with @Groxx about the plans.

taylanisikdemir (approved):

service/history/service.go
jakobht: Yes, I have tasks to move the ratelimiter to this wrapper, as well as the metrics to their own wrapper.

service/history/wrappers/ratelimited/ratelimit_test.go
jakobht: Yes, I think that is a good suggestion, will make a task for that as well.

PR #5727, opened 2024-03-05
Add test for ES v6 client Search method

neil-xie:
**What changed?**
Add unit test for Search method

**Why?**
Increase test coverage

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/elasticsearch/client/v6/client_test.go
+++ b/common/elasticsearch/client/v6/client_test.go
@@ -135,6 +135,104 @@ func TestCount(t *testing.T) {
 	assert.Equal(t, int64(42), count)
 }
 
+func TestSearch(t *testing.T) {
+	testServer := getTestServer(t)
+	defer testServer.Close()
+	// Create a new MockESClient
+	mockClient, err := elastic.NewClient(
+		elastic.SetURL(testServer.URL),
+		elastic.SetSniff(false),
+		elastic.SetHealthcheck(false),
+		elastic.SetHttpClient(testServer.Client()))
+	assert.NoError(t, err)
+	elasticV6 := ElasticV6{
+		client: mockClient,
+	}
+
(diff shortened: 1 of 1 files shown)


common/elasticsearch/client/v6/client_test.go
```diff
@@ -186,6 +284,74 @@ func getTestServer(t *testing.T) *httptest.Server {
 
 			w.WriteHeader(http.StatusOK)
 			w.Write([]byte(`{"count": 42}`))
+		case "/testIndex/_search":
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`{
+                "took": 5,
```
taylanisikdemir, line 290: instead of hardcoding different test case's expectations in the server handler you can pass these as a call back parameter to `getTestServer`. this way each test case's input and validations are self contained and test server can be more extensible for other scenarios

neil-xie: Refactored, also moved the mock client initialization inside it to make the code cleaner

common/elasticsearch/client/v6/client_test.go
```diff
@@ -71,39 +71,60 @@ func TestNewV6Client(t *testing.T) {
...
-		elastic.SetHealthcheck(false),
-		elastic.SetHttpClient(testServer.Client()),
-	)
-	assert.NoError(t, err)
-
-	elasticV6 := ElasticV6{
-		client: mockClient,
-	}
-	err = elasticV6.CreateIndex(context.Background(), "testIndex")
+	err := elasticV6.CreateIndex(context.Background(), "testIndex")
```
taylanisikdemir, line 84: I'd also validate the handler is called

taylanisikdemir (approved):

PR #5728, opened 2024-03-06
[history] refactor history client with timeout wrapper

shijiesheng:
**What changed?**

* added timeout template
* refactored history client to move out timeout logic
* added test for timeout client

**Why?**

make history client simple again

**How did you test it?**

unit test

**Potential risks**

**Release notes**

**Documentation Changes**


client/clientfactory.go
```diff
@@ -41,6 +41,7 @@ import (
 	"github.com/uber/cadence/client/wrappers/grpc"
 	"github.com/uber/cadence/client/wrappers/metered"
 	"github.com/uber/cadence/client/wrappers/thrift"
+	timoutwrapper "github.com/uber/cadence/client/wrappers/timeout"
```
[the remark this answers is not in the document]
shijiesheng: fixed

client/templates/timeout.tmpl
```diff
@@ -0,0 +1,38 @@
...
+   }
+}
+
+{{range $method := .Interface.Methods}}
+  {{if $method.AcceptsContext  }}
+    func (c *{{$decorator}}) {{$method.Declaration}} {
+      if ctx == nil {
+        ctx = context.Background()
+      }
+      var cancelFunc func()
```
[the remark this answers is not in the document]
shijiesheng: no this is to avoid ctx being scoped.

client/history/client.go
```diff
@@ -1134,8 +1061,6 @@ func (c *clientImpl) RespondCrossClusterTasksCompleted(
 	var response *types.RespondCrossClusterTasksCompletedResponse
 	op := func(ctx context.Context, peer string) error {
 		var err error
-		ctx, cancel := c.createContext(ctx)
-		defer cancel()
 		response, err = c.client.RespondCrossClusterTasksCompleted(ctx, request, append(opts, yarpc.WithShardKey(peer))...)
 		return err
 	}
```
Groxx, line 1066: unless I'm misreading things, this is _very_ different from the now-generated code.

it may be an improvement (createContext is rather terrifying iirc) but it's not clear this is correct.

taylanisikdemir: I didn't get what is the concern about correctness?

shijiesheng: I think for this method it's correct. What's incorrect is these methods below don't have timeout but now were generated incorrectly with timeout. I'll fix it in the next pr. 
```
GetReplicationMessages
GetDLQReplicationMessages
CountDLQMessages
ReadDLQMessages
PurgeDLQMessages
MergeDLQMessages
GetCrossClusterTasks
GetFailoverInfo
```

shijiesheng: fixed in https://github.com/uber/cadence/pull/5737

Groxx: previously this was potentially modifying the context on each `op` call during (possibly multiple) request redirection retries.

now it's only doing that once, before calling the retry impl.  (well before that, as it's the outer-most layer currently)

that's not necessarily a small change.

shijiesheng: timeout wrapper is still inside the "redirect client". So I think it's the same. 
Basically 
```
ctx, cancel := c.createContext(ctx)
defer cancel()
response, err = c.client.RespondCrossClusterTasksCompleted(ctx, request, append(opts, yarpc.WithShardKey(peer))...)
```
is modified to 
```
response, err = c.wrappedClient.RespondCrossClusterTasksCompleted(ctx, request, append(opts, yarpc.WithShardKey(peer))...)
```

PR #5737, opened 2024-03-06
[history] fix generated timeout wrapper

shijiesheng:
**What changed?**

https://github.com/uber/cadence/pull/5728 code generated history client should exclude some endpoints. 

**Why?**

Some replication endpoints don't have timeout before

**How did you test it?**

unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/client/templates/timeout.tmpl
+++ b/client/templates/timeout.tmpl
@@ -3,6 +3,7 @@ import (
   "time"
 )
 
+{{$exclude := splitList "|" (index .Vars "exclude")}}
 {{$clientName := (index .Vars "client")}}
 {{ $decorator := (printf "%s%s" (down $clientName) .Interface.Name) }}
 {{ $Decorator := (printf "%s%s" $clientName .Interface.Name) }}
@@ -22,8 +23,8 @@ func New{{$Decorator}}(client {{.Interface.Type}}, timeout time.Duration) {{.Int
 }
 
 {{range $method := .Interface.Methods}}
-  {{if $method.AcceptsContext  }}
-    func (c *{{$decorator}}) {{$method.Declaration}} {
+  func (c *{{$decorator}}) {{$method.Declaration}} {
+    {{- if and $method.AcceptsContext (not (has $method.Name $exclude)) }}
       if ctx == nil {
         ctx = context.Background()
       }
@@ -32,7 +33,7 @@ func New{{$Decorator}}(client {{.Interface.Type}}, timeout time.Duration) {{.Int
         ctx, cancelFunc = context.WithTimeout(ctx, c.timeout)
         defer cancelFunc()
       }
-      {{$method.Pass ("c.client.") }}
-    }
-  {{end}}
+    {{- end}}
+    {{$method.Pass ("c.client.") }}
+  }
 {{end}}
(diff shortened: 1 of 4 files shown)


client/templates/timeout.tmpl
```diff
@@ -3,6 +3,7 @@ import (
   "time"
 )
 
+{{$exclude := splitList "|" (index .Vars "exclude")}}
```
davidporter-id-au, line 6: I guess we're hitting the limitations of this templating

client/wrappers/timeout/history_generated_test.go
```diff
@@ -107,3 +108,25 @@ func Test_historyClient_CloseShard(t *testing.T) {
 		})
 	}
 }
+
+func Test_historyClient_GetReplicationMessages(t *testing.T) {
+	t.Run("no timeout", func(t *testing.T) {
+		m := history.NewMockClient(gomock.NewController(t))
+		m.EXPECT().GetReplicationMessages(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, r *types.GetReplicationMessagesRequest, opt ...yarpc.CallOption) (*types.GetReplicationMessagesResponse, error) {
+			for {
```
davidporter-id-au, line 116: sorry, maybe just me being dumb, but I am not sure I understand why this is a loop?

client/wrappers/timeout/history_generated_test.go
```diff
@@ -107,3 +108,25 @@ func Test_historyClient_CloseShard(t *testing.T) {
...
+				select {
+				case <-ctx.Done():
+					return nil, ctx.Err()
+				case <-time.After(time.Millisecond * 100):
```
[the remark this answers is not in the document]
shijiesheng: I tried. It's very complicated. The real issue here is incorrect test case with timeout set as 100. A quicker fix would be changing the timeout to 150ms and that should kill the flakiness

client/wrappers/timeout/history_generated_test.go
shijiesheng: It's for wait and handle the context timeout.

client/templates/timeout.tmpl
shijiesheng: Yes, but it's still useful to generate for most handlers.

davidporter-id-au: yeah, no concerns, just musing

client/wrappers/timeout/history_generated_test.go
davidporter-id-au: It doesn't matter very much, so this is more a nit comment, but wouldn't jsut the select statement work?

shijiesheng: I think you are right. We don't have default case so it's not really needed.

client/wrappers/timeout/history_generated_test.go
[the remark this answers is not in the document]
taylanisikdemir: yeah let's go with higher number then

taylanisikdemir (approved):

PR #5743, opened 2024-03-06
Coverage for dataStoreInterfaces

Groxx:
This is... much more questionable code than I anticipated going in.
Not gonna change it in this commit though.


common/persistence/data_store_interfaces.go
```diff
@@ -923,7 +923,8 @@ func NewDataBlob(data []byte, encodingType common.EncodingType) *DataBlob {
 		return nil
 	}
 	if encodingType != "thriftrw" && data[0] == 'Y' {
-		panic(fmt.Sprintf("Invalid incoding: \"%v\"", encodingType))
+		// original reason for this is not written down, but maybe for handling data prior to an encoding type?
+		panic(fmt.Sprintf("Invalid data blob encoding: \"%v\"", encodingType))
```
Groxx, line 927: no reason or comment about this in the original PR: #1176

common/persistence/data_store_interfaces_test.go
```diff
@@ -0,0 +1,177 @@
...
+			common.EncodingTypeUnknown,
+			common.EncodingTypeEmpty,
+			common.EncodingTypeProto,
+		}
+		const problematic = "Y..."
+
+		// I'm not entirely sure why this behavior exists, but to nail it down:
+		assert.NotPanics(t, func() {
+			NewDataBlob([]byte(problematic), common.EncodingTypeThriftRW)
+		}, "only thriftrw data can start with Y without panicking")
```
Groxx, line 58: no reason or comment about this in the original PR: #1176

common/persistence/data_store_interfaces_test.go
```diff
@@ -0,0 +1,177 @@
...
+			assert.Equal(t, common.EncodingTypeUnknown, (&DataBlob{Encoding: encoding}).GetEncoding())
+		}
+		// obvious
+		same(common.EncodingTypeGob)
+		same(common.EncodingTypeJSON)
+		same(common.EncodingTypeThriftRW)
+		same(common.EncodingTypeEmpty)
+
+		// highly suspicious
+		unknown(common.EncodingTypeProto)
```
Groxx, line 125: very strange.

also this entire func seems ridiculous to me.  it converts EncodingType -> string -> EncodingType and loses data in the process.  why in the world doesn't it just _return the EncodingType_

common/persistence/data_store_interfaces_test.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		assert.EqualValues(t, internal, blob.ToInternal(), "json should encode to internal type")
+		assert.Equal(t, blob, NewDataBlobFromInternal(internal), "json should decode from internal type")
+		// safe
+		internal.EncodingType = types.EncodingTypeThriftRW.Ptr()
+		blob.Encoding = common.EncodingTypeThriftRW
+		assert.EqualValues(t, internal, blob.ToInternal(), "thriftrw should encode to internal type")
+		assert.Equal(t, blob, NewDataBlobFromInternal(internal), "thriftrw should decode from internal type")
+
+		// all these panic
```
Groxx, line 150: we don't have enums for this in thrift/proto... beyond that I'm not sure why this is like this.  not sure where it's serialized / where it would matter.

taylanisikdemir (approved):

common/persistence/data_store_interfaces_test.go
```diff
@@ -0,0 +1,177 @@
...
+		same(common.EncodingTypeEmpty)
+
+		// highly suspicious
+		unknown(common.EncodingTypeProto)
+
+		// should be unknown
+		unknown(common.EncodingTypeUnknown)
+		unknown("any other value")
+	})
+	t.Run("To and FromInternal", func(t *testing.T) {
```
davidporter-id-au, line 131: @Groxx  What do you think about doing some round-trip testing against the various encoding methods - if they're not there already (I haven't looked at the existing coverage). 

For all bidirectional encodings, we should be able to round-trip any generated that:
```
for each encoding type:
    blob := generate(type)
    roundtripped := blob.FromInternal(blob.ToInternal())
    assert.Equal(t, blob, roundtripped)
```

davidporter-id-au: means to create data is possible with `testing/quick` for limited stuff and `common/testing/testdatagen/idlfuzzedtestdata` might be useful if you need either inspiration or concrete types

davidporter-id-au: if we're covering this already, feel free to ignore

common/persistence/data_store_interfaces.go
```diff
@@ -922,8 +922,9 @@ func NewDataBlob(data []byte, encodingType common.EncodingType) *DataBlob {
 	if len(data) == 0 {
 		return nil
 	}
-	if encodingType != "thriftrw" && data[0] == 'Y' {
-		panic(fmt.Sprintf("Invalid incoding: \"%v\"", encodingType))
+	if encodingType != common.EncodingTypeThriftRW && data[0] == 'Y' {
+		// original reason for this is not written down, but maybe for handling data prior to an encoding type?
+		panic(fmt.Sprintf("Invalid data blob encoding: \"%v\"", encodingType))
```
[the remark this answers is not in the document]
Shaddoll: There seems no reason. Temporal removed that logic: https://github.com/temporalio/temporal/pull/638/files#diff-f4778e6c9f74affb5767ad2ce6672a85ac6ba16eb64b42057e7dcef273ed5ac0L641

Shaddoll (approved):

common/persistence/data_store_interfaces_test.go
Groxx: Yeah, an explicit "round trips with X" seems semantically-meaningful, as it's not just "they look similar".  I'll add that 👍

common/persistence/data_store_interfaces.go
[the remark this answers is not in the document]
Groxx: we should probably remove it at some point then too.
I'll leave that for another diff / let's make a task or something around it, and see if thrift/proto ever start with Y because what even

PR #5747, opened 2024-03-07
Enforce leading space on comments

Groxx:
# What

Changes all comments like this:
```go
//stuff
type thing struct {
  field string //comment
}
//go:embed directives are left alone
```
To this:
```go
// stuff
type thing struct {
  field string // comment
}
//go:embed directives are left alone
```
And added a fairly basic lint check to prevent them from coming back.
The lint check is imprecise and doesn't catch comments at the ends of non-blank lines of code,
but fixing that seemed like much more work than it was worth.

And, since this will likely neutralize a `//nolint` comment that wasn't doing anything anyway,
I've removed that one comment entirely to avoid being misleading.

# How

"Starts with X but not followed by Y" is famously hard for regex, so I only tried briefly with sed / perl before giving up.
So instead I just had Goland format the entire repository, and undid anything that wasn't a Go file.

Somewhat surprisingly, all it did was add these spaces.  Which is perfect.
Oddly it missed a single obvious one, but I'm willing to forgive it for that.

# Why

IDEs often do this when formatting files, which is thrashing a bit.

gofmt (and therefore goimports) in theory doesn't actually care... but despite their
protestations about not caring about comment formatting in github issues, Go's
source-printer and therefore gofmt *does* treat some comments specially, most notably
"directives" like `go:build`: https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/go/printer/comment.go;l=15
and anything that might be a "doc comment" may get restructured too:
```diff
diff --git cmd/server/main.go cmd/server/main.go
index 6af56db58..b857f9211 100644
--- cmd/server/main.go
+++ cmd/server/main.go
@@ -35,6 +35,14 @@ import (
 )

 // main entry point for the cadence server
+//go:build
+// did you know
+//go:link
+// that go special
+//cases:things
+// that use
+//go:embed
+// specific comment structures?
 func main() {
 	app := cadence.BuildCLI(metrics.ReleaseVersion, metrics.Revision)
 	app.Run(os.Args)
```
which `gofmt -w` turns into:
```diff
diff --git cmd/server/main.go cmd/server/main.go
index 6af56db58..2511eb414 100644
--- cmd/server/main.go
+++ cmd/server/main.go
@@ -18,6 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.

+//go:build
+
 package main

 import (
@@ -35,6 +37,14 @@ import (
 )

 // main entry point for the cadence server
+// did you know
+// that go special
+// that use
+// specific comment structures?
+//
+//go:link
+//cases:things
+//go:embed
 func main() {
 	app := cadence.BuildCLI(metrics.ReleaseVersion, metrics.Revision)
 	app.Run(os.Args)
```
which IMO means we should explicitly avoid colliding with those kinds of comment patterns.
Linters like staticcheck generally use suppressing-comment structures like `//nolint:sa123`,
I suspect in part to fit with / avoid running afoul of this kind of reformatting.

Go has only claimed `x:y` as far as I can tell, so specifically `x:` is not a "directive"...
but that seems unnecessarily near-colliding to me, and it would allow `//TODO: asdf` which
are definitely not "tool directives" in any form, so I haven't allowed those.


Makefile
```diff
@@ -361,6 +361,12 @@ $(BUILD)/gomod-lint: go.mod internal/tools/go.mod common/archiver/gcloud/go.mod
 $(BUILD)/code-lint: $(LINT_SRC) $(BIN)/revive | $(BUILD)
 	$Q echo "lint..."
 	$Q $(BIN)/revive -config revive.toml -exclude './vendor/...' -exclude './.gen/...' -formatter stylish ./...
+	$Q bad="$$(find . -type f -name '*.go' -not -path './idls/*' | xargs grep -n -E '^\s*//\S' | grep -E -v '^[^:]+:[^:]+:\s*//[a-z]+:[a-z]+' || true)"; \
```
davidporter-id-au, line 364: I'd add a comment explaining this

I'm interpreting it to be selecting all relevant Go files, and then for each greppign them for comments in the pattern with whitespace followed by non-whitespace (\S), and then filtering out... I'm not sure what the intent is with that last regex

davidporter-id-au (approved):

Makefile
davidporter-id-au: ignore, I didn't read your PR docs

Groxx: Nah, it's legitimately weird looking unless you know these kinda esoteric details. I can comment the line above to keep intent clear without needing to read history.

taylanisikdemir (approved):

common/metrics/defs.go
```diff
@@ -36,7 +36,6 @@ type (
 
 	// metricDefinition contains the definition for a metric
 	metricDefinition struct {
-		//nolint
```
Groxx, line 39: this is the one being removed.
removing it doesn't cause any lints to fail, and it doesn't really make sense here anyway, so I'm just getting rid of it.

PR #5766, opened 2024-03-08
Try new codecov layout

Groxx:
Codecov's current recommendation is these "condensed_" versions.
Let's see if we like that better.

Before:
Screenshot 2024-03-08 at 12 29 59

After:
Screenshot 2024-03-15 at 14 50 01


codecov.yml
```diff
@@ -21,7 +21,7 @@ coverage:
       default:
         informational: true
 comment:
-  layout: "header, files, footer"
+  layout: "condensed_header, condensed_files, condensed_footer"
```
taylanisikdemir, line 24: before seems to have percentage diff which is nice to have on the comment.

Groxx:
Overall vote is "no", there's useful stuff in the larger view and it doesn't seem confusing or anything.

PR #5768, opened 2024-03-09
adding mutable state builder tests - adding continue-as-new events

davidporter-id-au:
**What changed?**

Adds some unit tests to the mutable state builder around the Continue-as-new event. 

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/mutable_state_builder_add_continue_as_new_event_test.go
```diff
@@ -0,0 +1,344 @@
...
+			// todo(david): make this mockable
+			td.expectedReturnedState.RunID = "a run id"
+			resultExecutionInfo.RunID = "a run id"
+			td.expectedReturnedState.CreateRequestID = "a request id"
+			resultExecutionInfo.CreateRequestID = "a request id"
+			for _, historyEvent := range returnedBuilder.GetHistoryBuilder().history {
+				if historyEvent.WorkflowExecutionStartedEventAttributes != nil {
+					historyEvent.WorkflowExecutionStartedEventAttributes.OriginalExecutionRunID = "a run id"
+				}
+			}
```
taylanisikdemir, line 338: another option for such fields that are not predictable to validate in tests is to use `cmpopts.IgnoreFields` and `cmp.Diff` instead of `assert.Equal`. Example https://github.com/uber/cadence/blob/a614c4c1fd1cf94b3722193cec919985304c9acb/common/types/mapper/thrift/shared_test.go#L1407

common/convert.go
```diff
@@ -28,6 +28,12 @@ import (
 	s "github.com/uber/cadence/.gen/go/shared"
 )
 
+// Returns a pointer to the given value.
+// todo (david.porter) Remove the remaining helpers here which are now obsolete
```
taylanisikdemir, line 32: let's add a test for this too

service/history/execution/mutable_state_builder_add_continue_as_new_event_test.go
```diff
@@ -0,0 +1,344 @@
...
+		historyManagerAffordance func(historyManager *persistence.MockHistoryManager)
+		shardManagerAffordance   func(shardContext *shardCtx.MockContext, msb *mutableStateBuilder, domainCache cache.DomainCache)
+		domainCacheAffordance    func(domainCache *cache.MockDomainCache)
+		taskgeneratorAffordance  func(taskGenerator *MockMutableStateTaskGenerator, msb *mutableStateBuilder)
+
+		expectedReturnedState   *persistence.WorkflowExecutionInfo // this is returned
+		expectedReturnedHistory []*types.HistoryEvent
+		expectedErr             error
+	}{
+		"a continue-as-new event with no errors": {
```
taylanisikdemir, line 232: There are a few err returning if blocks in `AddContinueAsNewEvent`. Are you planning to add more cases in follow up PRs?

taylanisikdemir (approved):

service/history/execution/mutable_state_builder_add_continue_as_new_event_test.go
davidporter-id-au: Yes, that was the intent, there's quite a few non-error cases not covered as well (with child workflow data), but the data will have to be hand-created rather than snapshotted with https://github.com/hexops/valast so I was deferring it a bit because it's slow going

common/convert.go
davidporter-id-au: good call

service/history/execution/mutable_state_builder_add_continue_as_new_event_test.go
davidporter-id-au: Fair point, I didn't know it had that param.

I would like to refactor the code so that this param can be passed in (I hate when functions use time.Now, uuid generation or math.rand in the middle of their logic) but that can wait.

PR #5769, opened 2024-03-11
Refactor/adding mutable state builder tests iv

davidporter-id-au:
**What changed?**
- Grinds out some tests for a fiddly function `assignEventIDToBufferedEvents` who's function it is is to increment the various counters and ensure that all buffered decision tasks have assigned IDs. 
- Adds a few minor guards for the logging functions

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/execution/mutable_state_builder.go
```diff
@@ -4811,24 +4812,30 @@ func (e *mutableStateBuilder) unixNanoToTime(
 }
 
 func (e *mutableStateBuilder) logInfo(msg string, tags ...tag.Tag) {
-	tags = append(tags, tag.WorkflowID(e.executionInfo.WorkflowID))
-	tags = append(tags, tag.WorkflowRunID(e.executionInfo.RunID))
-	tags = append(tags, tag.WorkflowDomainID(e.executionInfo.DomainID))
-	e.logger.Info(msg, tags...)
+	if e != nil && e.executionInfo != nil {
```
taylanisikdemir, line 4815: do we need these nil checks? If so let's check for nilness and terminate early.
```
if e == nil || e.executionInfo == nil {
  return
}

// log with tags
```

taylanisikdemir (approved):

service/history/execution/mutable_state_builder.go
davidporter-id-au: ah, sorry about that, let me address in a follow up

davidporter-id-au: https://github.com/uber/cadence/pull/5776

PR #5771, opened 2024-03-11
[code-coverage] Generate code for matching client timeout wrapper

ketsiambaku:
**What changed?**
Update timeout template to Generate code in client/matching/client.go with gowrap

**Why?**
To reduce manual work when adding APIs
Improve code coverage

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


client/templates/timeout.tmpl
```diff
@@ -70,3 +118,54 @@ func (c * {{$decorator}}) {{$method.Declaration}} {
     {{$method.Pass ("c.client.") }}
 }
 {{end}}
+{{end}}
+
+{{/* ↓↓↓ Special Case APIs ↓↓↓ */}}
```
Shaddoll, line 123: I suggest moving special case APIs out of generated code.

ketsiambaku: Done. I left it in the current file

Shaddoll (approved):

PR #5772, opened 2024-03-11
Added new error type that will not be retried by frontend

jakobht:
Note this PR needs to be updated after this IDL update is landed: https://github.com/uber/cadence-idl/pull/165, so we can point the submodule to the correct commit. 

**What changed?**
The workflowID ratelimiting now returns `ServiceBusyError` with a special `Reason`, frontend will _not_ retry the requests to history, but will instead immediately return the error to the caller. 

**Why?**
We do not want frontend to retry the requests to history when the ratelimit hit is the per workflow one. We just want it to return to the user.

**How did you test it?**
Tested via integration test and unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/types/mapper/thrift/errors.go
```diff
@@ -67,6 +67,8 @@ func FromError(err error) error {
 		return FromRetryTaskV2Error(e)
 	case *types.ServiceBusyError:
 		return FromServiceBusyError(e)
+	case *types.UserServiceBusyError:
```
taylanisikdemir, line 70: this file is missing unit tests. since you are touching it would be great to add some basic tests. maybe in a follow up PR. CDNC-8058

service/history/wrappers/ratelimited/handler_generated.go
```diff
@@ -109,7 +109,7 @@ func (h *historyHandler) DescribeWorkflowExecution(ctx context.Context, hp1 *typ
 	}
 
 	if !h.allowFunc(hp1.GetDomainUUID(), hp1.Request.GetExecution().GetWorkflowID()) {
-		err = &types.ServiceBusyError{"Too many requests for the workflow ID"}
+		err = &types.UserServiceBusyError{"Too many requests for the workflow ID"}
```
taylanisikdemir, line 112: Also commented on the idl PR about this. Maybe we should keep one type `ServiceBusyError` and add a `Reason` and `RateLimiterType` fields there. Or something along those lines to indicate error returned due to a specific type of rate limiter. 
In the near future we may introduce more types of rate limits and how/if frontend retries for them is subject to change.

common/util.go
```diff
@@ -257,7 +257,8 @@ func IsServiceTransientError(err error) bool {
 	case errors.As(err, &typesInternalServiceError):
 		return true
 	case errors.As(err, &typesServiceBusyError):
-		return true
+		// If the service busy error is due to workflow id rate limiting, then it is not transient
+		return typesServiceBusyError.Reason != WorkflowIDRateLimitReason
```
taylanisikdemir, line 261: Please add tests to cover new lines in mappers and this file

davidporter-id-au: I think *is a transient* error? it's just requiring backoff, though in the case of frontend it doesn't make sense to retry. 

If I'm not mistaken, what's missing from this classifying function is context as to where it's being used - error handling for frontend and history will probably need different definitions of what is retriable. I worry that with this rate-limit being potentially added by various non-frontend places with @sankari165's work it will possibly lead to a lot of unnecessarily failed tasks, DLQ'd messages etc. 

May I suggest either we split this  `IsServiceTransientError` into a function specific to frontend vs background processing either here or when Sankari's implementation lands

common/types/mapper/thrift/errors.go
jakobht: I have taken this task

common/util.go
jakobht: Coverage of mappers added in this PR: https://github.com/uber/cadence/pull/5795

taylanisikdemir (approved):

PR #5773, opened 2024-03-11
Ratelimit integration test

jakobht:
Note, this will fail until this PR is merged: https://github.com/uber/cadence/pull/5772

**What changed?**
Adds integration test for the workflowID specific ratelimits 

**Why?**
We need to make sure the requests and errors are passed correctly all the way to and from the client.

**How did you test it?**
N/A

**Potential risks**

**Release notes**

**Documentation Changes**


host/workflowidratelimit_test.go
```diff
@@ -0,0 +1,50 @@
...
+		ctx, cancel := createContext()
+		defer cancel()
+
+		_, err := s.engine.StartWorkflowExecution(ctx, request)
+
+		var busyErr *types.ServiceBusyError
+		if errors.As(err, &busyErr) {
+			busyErrCount += 1
+		} else {
+			s.NoError(err)
```
taylanisikdemir, line 44: we expect first 5 requests to succeed and last 5 to fail with busy err so you can do assertions in the loop based on `i` instead of final check at the end. makes the intent more obvious that way

davidporter-id-au (approved):
nice!

host/workflowidratelimit_test.go
jakobht: I have split it into two loops. I think that's more straightforward, please take a look

taylanisikdemir (approved):

host/workflowidratelimit_test.go
```diff
@@ -23,14 +23,86 @@
...
+	"github.com/uber/cadence/common/persistence"
+	pt "github.com/uber/cadence/common/persistence/persistence-tests"
 
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/types"
 )
 
-func (s *IntegrationSuite) TestWorkflowIDSpecificRateLimits() {
+func TestWorkflowIDRateLimitIntegrationSuite(t *testing.T) {
+	flag.Parse()
```
dkrotx, line 43: why do you need to parse command-line?

host/workflowidratelimit_test.go
```diff
@@ -23,14 +23,86 @@
...
+
+	clusterConfig, err := GetTestClusterConfig("integration_wfidratelimit_cluster.yaml")
+	if err != nil {
```
dkrotx, line 46: consider using `require.NoError` instead

host/workflowidratelimit_test.go
dkrotx: If it is required, it deserves a comment since it is very unexpected for test to parse cmd-line.

host/workflowidratelimit_test.go
```diff
@@ -57,12 +129,14 @@ func (s *IntegrationSuite) TestWorkflowIDSpecificRateLimits() {
 
 	// The ratelimit is 5 per second, so we should be able to start 5 workflows without any error
 	for i := 0; i < 5; i++ {
+		s.Require().NotNil(s.engine)
```
dkrotx: Having the check in single every loop make me think `s.engine` is assigned during the execution.
I think it should be enough to check outside as a pre-requisite. Maybe even in the `SetupSuite` where it is assigned?

jakobht: Yeah good catch, will delete, it's a debug leftover

host/workflowidratelimit_test.go
jakobht: Added a comment

host/workflowidratelimit_test.go
jakobht: Changed to require.NoError

host/workflowidratelimit_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	// Now we should get a rate limit error
+	for i := 0; i < 5; i++ {
+		_, err := s.engine.StartWorkflowExecution(ctx, request)
+		var busyErr *types.ServiceBusyError
+		assert.ErrorAs(s.T(), err, &busyErr)
+		assert.Equal(s.T(), common.WorkflowIDRateLimitReason, busyErr.Reason)
+	}
+}
```
dkrotx, line 142: I wonder you do not test for sending the requests again after a second.
This would check for the whole cycle - accepting, ratelimiting, accepting again.

dkrotx (commented):
nit: I recommend adding

host/workflowidratelimit_test.go
jakobht: Added a check in the end

dkrotx (approved):

PR #5775, opened 2024-03-11
Addition of tests for HandleDomainCreationReplicationTask in replicationTaskExecutor

abhishekj720:
**What changed?**
Added tests in common/domain folder for HandleDomainCreationReplicationTask in replicationTaskExecutor go file

**Why?**
to increase code coverage

**How did you test it?**
Unit Tests

**Potential risks**
No risks, as it only involves tests with mocking behaviors.

### DIFF
--- a/common/domain/replicationTaskExecutor_test.go
+++ b/common/domain/replicationTaskExecutor_test.go
@@ -22,6 +22,7 @@ package domain
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/golang/mock/gomock"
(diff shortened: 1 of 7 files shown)


common/domain/replicationTaskExecutor_test.go
```diff
@@ -248,3 +249,195 @@ func TestDomainReplicationTaskExecutor_Execute(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleDomainCreationReplicationTask(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockDomainManager := persistence.NewMockDomainManager(ctrl)
```
davidporter-id-au, line 257: so, the problem with putting it up here is that you're sharing it between table-rows, so it's a little tricky to get the assertions

common/domain/replicationTaskExecutor_test.go
```diff
@@ -248,3 +249,195 @@ func TestDomainReplicationTaskExecutor_Execute(t *testing.T) {
...
+					}, nil).AnyTimes() // Successfully finds a domain but with a different name
+			},
+			task:      validTask,
+			wantError: true, // Expect an error because of the name/UUID collision
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.setup()
```
davidporter-id-au, line 434: Why not create the domain manager and pass it in to the setup function here? That way you'll be creating a new mock every table test iteration and there's no risk of two rows setting up different expectations

```
t.Run(tt.name, func(t *testing.T) {
  ctrl := gomock.NewController(t)
  defer ctrl.Finish()
  
  mockDomainManager := persistence.NewMockDomainManager(ctrl)
  tt.setup(mockDomainManager)
```

abhishekj720: oh right, i think i missed it since just pushed it in the morning, thanks for the catch.

abhishekj720: but what about validTask? 
then i will be defining it in each case?

common/domain/replicationTaskExecutor_test.go
```diff
@@ -248,3 +250,341 @@ func TestDomainReplicationTaskExecutor_Execute(t *testing.T) {
...
+				PreviousFailoverVersion: 0,
+			},
+			wantError: true, // Expect an error because of the name/UUID collision
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			mockDomainManager := persistence.NewMockDomainManager(ctrl)
+			mockLogger := log.NewNoop()
```
3vilhamster, line 573: nit: testlogger.New

common/domain/replicationTaskExecutor_test.go
```diff
@@ -248,3 +250,341 @@ func TestDomainReplicationTaskExecutor_Execute(t *testing.T) {
...
+			mockTimeSource := clock.NewRealTimeSource()
```
3vilhamster, line 574: nit: clock.NewMockedTimeSourceAt with a fixed time.

common/domain/replicationTaskExecutor_test.go
```diff
@@ -248,3 +250,341 @@ func TestDomainReplicationTaskExecutor_Execute(t *testing.T) {
...
+				mockDomainManager.EXPECT().
+					CreateDomain(gomock.Any(), gomock.Any()).
+					Return(nil, ErrInvalidDomainStatus).Times(1)
+
+				mockDomainManager.EXPECT().
+					GetDomain(gomock.Any(), gomock.Any()).
+					Return(nil, errors.New("unexpected error")).Times(1) // Simulate an unexpected error from GetDomain
+			},
+			task: &types.DomainTaskAttributes{
+				DomainOperation: types.DomainOperationCreate.Ptr(),
```
3vilhamster, line 486: I suggest refactoring out test data from here. It is very hard to read the test cases that are over 200 lines with intendation.

3vilhamster (commented):
A bit hard to read the test. :(

abhishekj720:
Already done the changes in another PR.

PR #5782, opened 2024-03-13
Retryable error for workflow rate limits in task processing

sankari165:
**What changed?**
New error that will be introduced when task processing requests are ratelimited

**Why?**
Task processing limits will be ratelimited based on wf-id based rate limits and needs to be retried 

**How did you test it?**
Unit tests

**Potential risks**
error is still not in use

**Release notes**

**Documentation Changes**


jakobht (approved):

service/history/task/task_test.go
```diff
@@ -280,6 +290,18 @@ func (s *taskSuite) TestHandleErr_ErrMaxAttempts() {
...
 
+func (s *taskSuite) TestRetryErr() {
+	taskBase := s.newTestTask(func(task Info) (bool, error) {
+		return true, nil
+	}, nil)
+
+	s.Equal(false, taskBase.RetryErr(errWorkflowBusy))
+	s.Equal(false, taskBase.RetryErr(ErrTaskPendingActive))
+	s.Equal(false, taskBase.RetryErr(context.DeadlineExceeded))
+	s.Equal(false, taskBase.RetryErr(&redispatchError{Reason: "random-reason"}))
```
dkrotx, line 301: this is a bit unexpected to see. What's the point of these NotEquals if there is the next line?

dkrotx (approved):

service/history/task/task_test.go
sankari165: added a comment

PR #5788, opened 2024-03-14
Ignore requests send from pinot response comparator

bowenxia:
**What changed?**
Ignored the requests sent from Pinot Response Comparator

**Why?**
If we don't ignore, we will have duplicated data. 

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/pinotVisibilityTripleManager.go
+++ b/common/persistence/pinotVisibilityTripleManager.go
@@ -331,11 +331,8 @@ type userParameters struct {
 // logUserQueryParameters will log user queries' parameters so that a comparator workflow can consume
 func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userParameters, domain string, override bool) {
 	// Don't log if it is not enabled
-	if !v.logCustomerQueryParameter(domain) {
-		return
-	}
-	// Don't log if it is a call from Pinot Response Comparator workflow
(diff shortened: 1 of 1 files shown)


common/persistence/pinotVisibilityTripleManager.go
```diff
@@ -329,11 +329,16 @@ type userParameters struct {
 
 // For Pinot Migration uses. It will be a temporary usage
 // logUserQueryParameters will log user queries' parameters so that a comparator workflow can consume
-func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userParameters, domain string) {
+func (v *pinotVisibilityTripleManager) logUserQueryParameters(userParam userParameters, domain string, override bool) {
 	// Don't log if it is not enabled
 	if !v.logCustomerQueryParameter(domain) {
```
neil-xie, line 334: we can combine this with override check `if !v.logCustomerQueryParameter(domain) || override {}`

neil-xie (approved):
Left one nit comment, otherwise lgtm

common/persistence/pinotVisibilityTripleManager.go
bowenxia: Sure. I was thinking to make it more readable

PR #5790, opened 2024-03-15
Fix checksum validation for SQL implementation

Shaddoll:
**What changed?**
Add a check for the SQL implementation of GetWorkflowExecution operation to exclude false positive checksum validation failure cases.

**Why?**
To make sure the checksum validation result is true, the data we read from GetWorkflowExecution operation are from a consistent view. In the NoSQL implementation, the operation is a single read, so the data is from a consistent view. However, in the SQL implementation, the operation is multiple reads from different table. If there is a concurrent update, the data we read isn't from a consistent view and the checksum validation could fail. Normally, we don't have concurrent updates with reads. But when the shard ownership changed, it might not be the case. 

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/executionManager.go
+++ b/common/persistence/executionManager.go
@@ -73,6 +73,7 @@ func (m *executionManagerImpl) GetWorkflowExecution(
 	internalRequest := &InternalGetWorkflowExecutionRequest{
 		DomainID:  request.DomainID,
 		Execution: request.Execution,
+		RangeID:   request.RangeID,
 	}
 	response, err := m.persistence.GetWorkflowExecution(ctx, internalRequest)
 	if err != nil {
@@ -642,12 +643,9 @@ func (m *executionManagerImpl) SerializeWorkflowMutation(
 	if err != nil {
 		return nil, err
 	}
-	var checksumData *DataBlob
-	if len(input.Checksum.Value) > 0 {
-		checksumData, err = m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
-		if err != nil {
-			return nil, err
-		}
+	checksumData, err := m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
+	if err != nil {
+		return nil, err
 	}
 
 	return &InternalWorkflowMutation{
--- a/common/persistence/executionManager.go
+++ b/common/persistence/executionManager.go
@@ -642,9 +642,12 @@ func (m *executionManagerImpl) SerializeWorkflowMutation(
 	if err != nil {
 		return nil, err
 	}
-	checksumData, err := m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
-	if err != nil {
-		return nil, err
+	var checksumData *DataBlob
+	if len(input.Checksum.Value) > 0 {
+		checksumData, err = m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	return &InternalWorkflowMutation{
--- a/common/persistence/serializer.go
+++ b/common/persistence/serializer.go
@@ -304,6 +304,9 @@ func (t *serializerImpl) DeserializeAsyncWorkflowsConfig(data *DataBlob) (*types
 }
 
 func (t *serializerImpl) SerializeChecksum(sum checksum.Checksum, encodingType common.EncodingType) (*DataBlob, error) {
+	if len(sum.Value) == 0 {
+		return nil, nil
+	}
 	return t.serialize(sum, encodingType)
 }
 
--- a/common/persistence/sql/sql_execution_store.go
+++ b/common/persistence/sql/sql_execution_store.go
@@ -35,6 +35,7 @@ import (
 	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/collection"
 	"github.com/uber/cadence/common/log"
+	"github.com/uber/cadence/common/persistence"
 	p "github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/persistence/serialization"
 	"github.com/uber/cadence/common/persistence/sql/sqlplugin"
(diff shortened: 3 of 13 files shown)


service/history/execution/context.go
```diff
@@ -1209,6 +1209,9 @@ func (c *contextImpl) getWorkflowExecutionWithRetry(
 	case *types.EntityNotExistsError:
 		// it is possible that workflow does not exists
 		return nil, err
+	case *persistence.ShardOwnershipLostError:
```
davidporter-id-au, line 1212: Little surprised about this, @Shaddoll  do you think this being missing could have caused problems?

common/persistence/executionManager.go
```diff
@@ -642,9 +642,12 @@ func (m *executionManagerImpl) SerializeWorkflowMutation(
 	if err != nil {
 		return nil, err
 	}
-	checksumData, err := m.serializer.SerializeChecksum(input.Checksum, common.EncodingTypeJSON)
-	if err != nil {
-		return nil, err
+	var checksumData *DataBlob
+	if len(input.Checksum.Value) > 0 {
```
taylanisikdemir, line 646: `serializer.SerializerChecksum` does nil check but we are losing that with this change. Can `input.Checksum` be nil? Either way I'd move this length check to `SerializeChecksum`

common/persistence/sql/sql_execution_store_test.go
```diff
@@ -3342,6 +3347,110 @@ func TestGetWorkflowExecution(t *testing.T) {
...
+				db.EXPECT().SelectFromSignalsRequestedSets(gomock.Any(), gomock.Any()).Return(nil, sql.ErrNoRows)
+				db.EXPECT().SelectFromBufferedEvents(gomock.Any(), gomock.Any()).Return(nil, sql.ErrNoRows)
+				parser.EXPECT().WorkflowExecutionInfoFromBlob(gomock.Any(), gomock.Any()).Return(&serialization.WorkflowExecutionInfo{
+					Checksum:         []byte("test-checksum"),
+					ChecksumEncoding: "test-checksum-encoding",
+				}, nil)
+				db.EXPECT().SelectFromShards(gomock.Any(), gomock.Any()).Return(nil, sql.ErrNoRows)
+				db.EXPECT().IsNotFoundError(gomock.Any()).Return(true).AnyTimes()
+			},
+			wantErr: true,
```
taylanisikdemir, line 3428: also assertErr for this one?

service/history/execution/context.go
taylanisikdemir: Without this specific line change it would fall to default case below. Returned err type is not impacted. This switch case seems mostly about keeping error log noise under control

common/persistence/sql/sql_execution_store.go
```diff
@@ -375,6 +376,24 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
 			Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
 		}
 	}
+	// if we have checksum, we need to make sure the rangeID did not change
+	// if the rangeID changed, it means the shard ownership might have changed
+	// and the workflow might have been updated when we read the data, so the data
+	// we read might not from a consistent view, the checksum validation might fail
```
taylanisikdemir, line 382: nit: "..might not BE from.."

common/persistence/sql/sql_execution_store.go
```diff
@@ -375,6 +376,24 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
+	// in that case, we need to return an error
+	if state.ChecksumData != nil {
+		row, err := m.db.SelectFromShards(ctx, &sqlplugin.ShardsFilter{ShardID: int64(m.shardID)})
+		if err != nil {
+			return nil, convertCommonErrors(m.db, "GetWorkflowExecution", "", err)
+		}
+		if row.RangeID != request.RangeID {
+			return nil, &persistence.ShardOwnershipLostError{
```
taylanisikdemir, line 390: `TestCreateWorkflowExecutionConcurrentCreate` is timing out. Probably retrying for this case forever.

service/history/execution/context.go
davidporter-id-au: true, I didn't read this carefully enough.

common/persistence/sql/sql_execution_store.go
Shaddoll: Fixed, it's because the original context was overwritten by a child context.

service/history/shard/context.go
```diff
@@ -585,6 +585,10 @@ func (s *contextImpl) GetWorkflowExecution(
 	ctx context.Context,
 	request *persistence.GetWorkflowExecutionRequest,
 ) (*persistence.GetWorkflowExecutionResponse, error) {
+	s.RLock()
+	currentRangeID := s.getRangeID()
+	s.RUnlock()
+	request.RangeID = currentRangeID
```
taylanisikdemir, line 591: can be done above in one line

common/persistence/serializer.go
```diff
@@ -304,6 +304,9 @@ func (t *serializerImpl) DeserializeAsyncWorkflowsConfig(data *DataBlob) (*types
 }
 
 func (t *serializerImpl) SerializeChecksum(sum checksum.Checksum, encodingType common.EncodingType) (*DataBlob, error) {
+	if len(sum.Value) == 0 {
```
taylanisikdemir, line 307: also check `sum == nil`

common/persistence/sql/sql_execution_store.go
```diff
@@ -375,6 +375,21 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
...
+	// and the workflow might have been updated when we read the data, so the data
+	// we read might not be from a consistent view, the checksum validation might fail
+	// in that case, we clear the checksum data so that we will not perform the validation
+	if state.ChecksumData != nil {
+		row, err := m.db.SelectFromShards(ctx, &sqlplugin.ShardsFilter{ShardID: int64(m.shardID)})
+		if err != nil {
+			return nil, convertCommonErrors(m.db, "GetWorkflowExecution", "", err)
+		}
+		if row.RangeID != request.RangeID {
+			state.ChecksumData = nil
```
taylanisikdemir, line 389: let's add a debug log here

common/persistence/sql/sql_execution_store.go
```diff
@@ -375,6 +375,21 @@ func (m *sqlExecutionStore) GetWorkflowExecution(
 			Message: fmt.Sprintf("GetWorkflowExecution: failed. Error: %v", err),
 		}
 	}
+	// if we have checksum, we need to make sure the rangeID did not change
+	// if the rangeID changed, it means the shard ownership might have changed
+	// and the workflow might have been updated when we read the data, so the data
+	// we read might not be from a consistent view, the checksum validation might fail
+	// in that case, we clear the checksum data so that we will not perform the validation
```
taylanisikdemir, line 382: let's mention "The GetWorkflowExecution operation will not be impacted by this. `ChecksumData` is purely for validation purposes."

taylanisikdemir (approved):

PR #5793, opened 2024-03-18
Fix workflow deletion

Shaddoll:
**What changed?**
Fix workflow deletion

**Why?**
To prevent garbage data from workflows which are supposed to be deleted.
Workflow execution should exist in order to execute the deletion task, so it must be the last one to be deleted.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/sql/sql_execution_store.go
+++ b/common/persistence/sql/sql_execution_store.go
@@ -625,104 +625,77 @@ func (m *sqlExecutionStore) DeleteWorkflowExecution(
 	ctx context.Context,
 	request *p.DeleteWorkflowExecutionRequest,
 ) error {
-	recoverPanic := func(recovered interface{}, err *error) {
-		if recovered != nil {
-			*err = fmt.Errorf("DB operation panicked: %v %s", recovered, debug.Stack())
-		}
-	}
+	dbShardID := sqlplugin.GetDBShardIDFromHistoryShardID(m.shardID, m.db.GetTotalNumDBShards())
 	domainID := serialization.MustParseUUID(request.DomainID)
 	runID := serialization.MustParseUUID(request.RunID)
 	wfID := request.WorkflowID
-	g, ctx := errgroup.WithContext(ctx)
-
-	g.Go(func() (e error) {
-		defer func() { recoverPanic(recover(), &e) }()
-		_, e = m.db.DeleteFromExecutions(ctx, &sqlplugin.ExecutionsFilter{
+	return m.txExecute(ctx, dbShardID, "DeleteWorkflowExecution", func(tx sqlplugin.Tx) error {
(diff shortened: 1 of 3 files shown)


common/persistence/sql/sql_execution_store.go
```diff
@@ -625,104 +625,77 @@ func (m *sqlExecutionStore) DeleteWorkflowExecution(
...
+	dbShardID := sqlplugin.GetDBShardIDFromHistoryShardID(m.shardID, m.db.GetTotalNumDBShards())
 	domainID := serialization.MustParseUUID(request.DomainID)
 	runID := serialization.MustParseUUID(request.RunID)
 	wfID := request.WorkflowID
-	g, ctx := errgroup.WithContext(ctx)
-
-	g.Go(func() (e error) {
-		defer func() { recoverPanic(recover(), &e) }()
-		_, e = m.db.DeleteFromExecutions(ctx, &sqlplugin.ExecutionsFilter{
+	return m.txExecute(ctx, dbShardID, "DeleteWorkflowExecution", func(tx sqlplugin.Tx) error {
```
taylanisikdemir, line 632: You are not only changing the order but also making this operation a transaction. 
- Other transactional operations in this file are using `txExecuteShardLockedFn`. Should this also grab shard lock? 
- This transaction will perform deletions on multiple tables an will have to acquire corresponding locks at DB level. Compared to previous implementation the overall latency might be high which is OK for deletion case but do you think this has a chance to impact performance of other queries on those tables?

Shaddoll: - It doesn't have to grab a shard lock, we're trying to delete a workflow passing the retention period. Whether it is deleted by the owner of the shard isn't important.
- For innodb with repeatable read isolation level, the delete acquires a next-key lock on every record it encounters. The deletion are either point deletion or range deletion that deleting all records within a range. So the impact performance of other queries on those tables should be low. (Only 1 workflow next to the deleted workflow could be impacted.) For innodb with read-committed isolation level and myrocksdb, there is no next-key lock or gap lock, only the records being deleted are locked. So there is no impact to other workflows.

service/history/task/timer_task_executor_base.go
```diff
@@ -197,12 +199,6 @@ func (t *timerTaskExecutorBase) archiveWorkflow(
 		return err
 	}
 
-	if err := t.deleteCurrentWorkflowExecution(ctx, task); err != nil {
-		return err
-	}
-	if err := t.deleteWorkflowExecution(ctx, task); err != nil {
-		return err
-	}
```
taylanisikdemir, line 205: please double check `!HistoryArchivedInline` case doesn't leave history behind

taylanisikdemir (approved):

PR #5796, opened 2024-03-19
Move error_test.go to a separate test package

jakobht:
**What changed?**
Moved error_test.go to a separate test package

**Why?**
The folder testdata is ignored when running tests, so moving this test to a test package that isn't ignored

**How did you test it?**
Made sure the test runs when doing `go test` and `make test`

**Potential risks**

**Release notes**

**Documentation Changes**


common/types/testdata_test/error_test.go
```diff
@@ -20,17 +20,18 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-package testdata
+package testdata_test
```
taylanisikdemir, line 23: FYI currently the coverage numbers we collect only include the tests targeting the code in the same package. So a separate package with unit tests don't add to coverage numbers. There are workarounds for this setup but it increases total test time. 
Looks like this file is only covering functionality of some stuff in testdata which is not part of actual code so maybe we can leave this there

jakobht: Oh, that's sad, I was not aware. 
Go has some magic folder names (e.g. starting with _ and `testdata`) for which it does not run tests at all, so that's why I moved it out. So now the test is acctually run by the CI (before it was not).

davidporter-id-au: @taylanisikdemir  I'd broadcast this to the team, I wasn't aware either and this convention is pretty common

jakobht: I agree, just still note that in _this_ PR, we _have to_ move it out as tests in folders called `testdata` is never picked up by `go test`

taylanisikdemir: Stuff in `testdata` should only be used by other tests and shouldn't require to be tested. If this functionality deserves its own tests then let's move it to some other folder or add those tests to one of the relevant _test.go file. No need to track coverage for this.

jakobht: Moved to the testing package

common/testing/allIsSet_test.go
```diff
@@ -0,0 +1,40 @@
+// The MIT License (MIT)
```
taylanisikdemir, line 1: We have been renaming files to snake_case as we touch them

PR #5800, opened 2024-03-19
Write tests for cassandra TestSelectAll*Workflows

taylanisikdemir:
**What changed?**
Added unit tests for `SelectAllCurrentWorkflows` and `SelectAllWorkflowExecutions`

**Why?**
Improve coverage

**How did you test it?**
unit tests

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/workflow_test.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/workflow_test.go
@@ -670,6 +670,252 @@ func TestDeleteWorkflowExecution(t *testing.T) {
 	}
 }
 
+func TestSelectAllCurrentWorkflows(t *testing.T) {
+	tests := []struct {
+		name           string
+		shardID        int
+		pageToken      []byte
+		pageSize       int
+		iter           *fakeIter
+		wantExecutions []*persistence.CurrentWorkflowExecution
+		wantErr        bool
+	}{
+		{
+			name:      "nil iter returned",
+			shardID:   1,
+			pageToken: []byte("test-page-token"),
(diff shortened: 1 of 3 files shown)


common/persistence/nosql/nosqlplugin/cassandra/workflow_test.go
```diff
@@ -670,6 +670,252 @@ func TestDeleteWorkflowExecution(t *testing.T) {
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+
+			query := gocql.NewMockQuery(ctrl)
+			query.EXPECT().PageSize(tc.pageSize).Return(query).Times(1)
+			query.EXPECT().PageState(tc.pageToken).Return(query).Times(1)
+			query.EXPECT().WithContext(gomock.Any()).Return(query).Times(1)
+			if tc.iter != nil {
```
davidporter-id-au, line 756: nit:  I would consider setting up a per-test affordance for these if they're going to expand at all; lots of branching assertions can get pretty hard to read imho. 

ie
```
tests := []struct {
		name           string
		...
		querymock              func(m *mockQuery) // pass in what you expect the mock to return per row
	}{
...

```

davidporter-id-au (approved):

common/persistence/nosql/nosqlplugin/cassandra/workflow_test.go
taylanisikdemir: In general I agree but these are all shared mocks and each test case provides inputs to the function which are used to setup the mock here.

PR #5805, opened 2024-03-21
Enforce new line coverage 85%

taylanisikdemir:
**What changed?**
Trying codecov's `patch` coverage setting to enforce new line coverage per diff to be > 85%.

**How did you test it?**
I added a new dummy function without tests in the first commit on this PR. Codecov patch check failed with below message:

Screenshot 2024-03-21 at 10 37 13 AM

Then I added tests to cover half of the lines:
Screenshot 2024-03-21 at 11 31 27 AM

Then I removed the dummy function so PR only has yml changes:

Screenshot 2024-03-21 at 12 10 14 PM

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -19,7 +19,11 @@ coverage:
         if_ci_failed: ignore   # require the CI to pass before setting the status
     patch:
       default:
-        informational: true
+        target: 85%            # specify the target coverage for each commit status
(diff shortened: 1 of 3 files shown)


abhishekj720 (approved):

davidporter-id-au (approved):
discussed offline that we don't expect this to hit a problem with the nondeterministic coverage problem due to it being patch related only

PR #5809, opened 2024-03-21
Add additional tests for ExecutionManager

3vilhamster:
**What changed?**
Adding more tests to ExecutionManager.
Now covering GetExecutionInfo and UpdateExecution.
That includes a lot of deserialization and data conversion. I've added some sane checks, if we'll spot any missing fields, it should be easy enough extend the test with extra fields.

**Why?**
Improving code coverage.

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/execution_manager_test.go
```diff
@@ -24,8 +24,13 @@ package persistence
 
 import (
 	"context"
+	"fmt"
+	"github.com/uber/cadence/common"
+	"github.com/uber/cadence/common/checksum"
+	"github.com/uber/cadence/common/types"
```
taylanisikdemir, line 30: make fmt

common/persistence/execution_manager_test.go
```diff
@@ -263,3 +268,277 @@ func TestGetReplicationTasks(t *testing.T) {
 		})
 	}
 }
+
```
taylanisikdemir, line 271: tests have failed on CI https://github.com/uber/cadence/actions/runs/8382150830/job/22955320462?pr=5809

common/persistence/execution_manager_test.go
3vilhamster: ohwell. I've tried adding gci as file watcher, but it makes code writing very weird. I'll try to figure out how to set a helper for myself. Or maybe a recommit hook.

common/persistence/execution_manager_test.go
3vilhamster: Good catch.
The issue was caused by not setting UTC in time.Unix. Fixed.

taylanisikdemir (approved):

PR #5814, opened 2024-03-25
Upgrade apache thrift to v.0.17.0

3vilhamster:
**What changed?**
I'm upgrading Apache Thrift to a recent version. The old replace was caused by monorepo/ringpop lock on the ancient version. It was removed from Monorepo, so we no longer require a lock.

**Why?**
Remove unsupported ancient version lock.

**How did you test it?**
Verified local with xdc server setup 
```
./cadence-server --config config --zone xdc --env development_xdc_cluster0 start
./cadence-server --config config --zone xdc --env development_xdc_cluster1 start
./cadence-server --config config --zone xdc --env development_xdc_cluster2 start
```

Verified inside Uber without issues on staging environments.

**Potential risks**
None expected. Thrift is binary backward compatible, but they updated the interface with context-aware calls, so the service should fail at compile time.

**Release notes**
Update apache/thrift to 0.17.0

**Documentation Changes**


davidporter-id-au (approved):
good luck :salut:

PR #5817, opened 2024-03-26
Global ratelimiter, part 2: Any-typed RPCs, mappers, and stub handler

Groxx:
Docs are broadly the same as https://github.com/uber/cadence-idl/pull/168, but are repeated here for visibility with some additions at the end for stuff unique to this repo:

# Any type:
This API is purely for inter-Cadence-service use, and the Thrift equivalent is in the github.com/uber/cadence-idl repository: https://github.com/uber/cadence-idl/pull/168

Design-wise: this has gone through several adjustment and experimenting rounds, and nearly all top-level keys I can come up with are either totally insufficient or totally unnecessary for one ratelimit design or another, so I've leaned more and more towards "define nothing, allow passing anything".

In the extreme, this means: just one `Any` field.  We can add more later if needed / if we discover some kind of truly universal data that is always worth including.
Obviously this is not super ergonomic, but as the ratelimiter is intended to be pluggable by outside implementations it cannot *wholly* be defined in internal protocol definitions.  There needs to be an extendable type of some kind, and arbitrarily requiring e.g. a `list<?>` or `map<string,?>` doesn't actually lend any beneficial semantics to the system, nor reduce space on the wire.

Implementers will need to maintain in-process definitions on both ends of the protocol, and check type IDs before decoding.  Generally speaking this is probably best done with either a shared protobuf-or-something definition (anywhere) or something schema-free like JSON, so cross-version communication can be done safely e.g. during server upgrades.

# Intended use
Eventually this will be part of a pluggable global-ratelimit-data-exchange system, for anything that can make use of the high level "periodically check in to sharded hosts, share data, update limits" pattern that github.com/uber/cadence/common/quotas/global is building up.

The first version will end up using inside-`Any` structures like this:
```go
type Request struct {
  Caller  string             `json:"caller"`
  Elapsed time.Duration      `json:"elapsed"`
  // a compact map[ratelimit key][allowed, rejected] structure.
  // this could be a regular proto/thrift map<string, struct> instead
  // but that'd have lots of repeated field names in JSON
  Data    map[string][2]int  `json:"data"`
}

type Response struct {
  RPS map[ratelimit key]float64  `json:"rps"`
}
```
which will probably just use JSON for simplicity for starters.
Or I'll make a new proto struct and shove that in there instead, using proto-encoded data for both thrift and proto transports.
It doesn't particularly matter.

These request/response structures match the API defined in github.com/uber/cadence/common/quotas/global/algorithm/requestweighted.go , and there will be a fairly simple mapper to convert to/from the `Any` types defined here to make things work.

# Why `Any` instead of strict types?
Because:
- Types we define in these internally-controlled APIs cannot be modified by anyone running Cadence without maintaining a full fork of multiple repos.
- Rate-limiting details are assumed to be relatively unique to internal networking setups, and is therefore not something we intend to fix centrally in every possible variation that server-hosts encounter.

So instead this is loosely typed, and the API will allow Cadence-wrapping developers to add wholly externally defined types as needed, and use DynamicConfig to shadow/enable/disable/etc between any registered algorithms.  If your company needs to e.g. share custom info from Kubernetes to decide how many requests host X can allow (say it's based on number of cores allocated), you will have a relatively simple way to build that and safely try it out.

# Why not `google.protobuf.Any`?
Covered in code comments, but mostly: because that would _require_ using both protobuf _and_ the same protobuf code-generator we use, so the types can be registered and decoded automatically.
That auto-decoding isn't necessary (nor even particularly useful) though, it can be deferred to the ratelimiter-plugin, and then it can be entirely encoding-agnostic.

So that's what this does.  It's not a `google.protobuf.Any` because it doesn't _have_ to be, and you're free to put a `google.protobuf.Any` inside your `types.Any` if ~~yo dawg I heard you like loose typing~~ you want to do so.

# Mappers
As the RPC specs and the internal type do essentially no interpreting of their contents, the mappers are both trivial and complete.

To actually _use_ an `Any` requires custom code per `ValueType`, which will come in a later PR.

I've written far more mapping-test-code than they really deserve, to try to figure out what some fuzzed round-trip tests might look like for this and other types.
As part of that, I discovered that https://github.com/google/gofuzz has some rather severe issues when generating protobuf types (e.g. it unavoidably panics if it finds any interface fields, which `oneof` use), and was recently archived, so I think we might need to come up with something else before going too much farther.  Maybe just a fork?

# Handler
Fully wiring this up will come later, and at a very high level will require:
- a `common/quota` wrapper to collect usage metrics in the "limiting" hosts (frontend), which will:
  - periodically, send bulk `RatelimitUpdateRequest`s to "aggregating" hosts (history)
  - use the responses to update quotas as needed
- dynamicconfig to enable/disable/shadow/roll-back all this
- relatively simple handler/encoding/decoding code to convert between the "real" types

All of which is essentially out of scope for this PR.


service/history/handler/interface.go
```diff
@@ -86,4 +86,5 @@ type Handler interface {
 	SyncShardStatus(context.Context, *types.SyncShardStatusRequest) error
 	TerminateWorkflowExecution(context.Context, *types.HistoryTerminateWorkflowExecutionRequest) error
 	GetFailoverInfo(context.Context, *types.GetFailoverInfoRequest) (*types.GetFailoverInfoResponse, error)
+	RatelimitUpdate(ctx context.Context, Request *types.RatelimitUpdateRequest) (*types.RatelimitUpdateResponse, error)
```
taylanisikdemir, line 89: do we have a way to avoid non-cadence services calling this?

Groxx: No more than any other history-API method

common/types/mapper/testutils/testutils.go
```diff
@@ -0,0 +1,82 @@
...
+		if t.Failed() { // else a bit noisy
+			t.Logf("expected to see:  %#v", expected)
+			t.Logf("observed results: %#v", results)
+			t.Logf("detailed results: %#v", details)
+			t.Logf("fuzz seed: %v", seed)
+		}
+	}()
+
+	for tries := 0; tries < 100; tries++ { // retry a few times if needed
+		for i := 0; i < 100; i++ { // always fuzz a moderate amount, don't stop immediately
```
Groxx, line 53: when I was covering a few more branches (e.g. 4-5) I did in fact see 100 fail more than once.

I've used this "try, check, retry if necessary" a few times in personal stuff to check fairness / branch coverage / etc to be both fair and stable, and it works quite well.

common/types/mapper/proto/shared_test.go
```diff
@@ -296,3 +300,60 @@ func TestCrossClusterApplyParentClosePolicyResponse(t *testing.T) {
...
+			// report what branch of behavior was fuzzed occurred
+			if len(orig.Value) == 0 {
+				return "empty data" // ignoring nil vs empty difference
+			}
+			return "filled data"
+		})
+	})
+	t.Run("round trip from proto", func(t *testing.T) {
+		testutils.EnsureFuzzCoverage(t, []string{
+			"empty data", "filled data",
```
Groxx, line 341: part of the reason I like having this "make sure you did actually do N branches" check:
it ensures that `f.Fuzz(&orig.ValueType)` actually does do different things, rather than "fuzzing" the same value over and over.  e.g. if it didn't mutate the data for some reason.

common/types/mapper/proto/history_test.go
```diff
@@ -363,3 +365,22 @@ func TestHistoryGetTaskInfoResponse(t *testing.T) {
 		assert.Equal(t, item, ToHistoryGetFailoverInfoResponse(FromHistoryGetFailoverInfoResponse(item)))
 	}
 }
+
+func TestRatelimitUpdate(t *testing.T) {
```
taylanisikdemir, line 369: Any specific reason not to follow table based approach that is used in other tests in this file? They are much shorter and easier to add more cases

service/history/handler/handler.go
```diff
@@ -2055,6 +2055,14 @@ func (h *handlerImpl) GetFailoverInfo(
 	return resp, nil
 }
 
+func (h *handlerImpl) RatelimitUpdate(
+	ctx context.Context,
+	request *types.RatelimitUpdateRequest,
+) (*types.RatelimitUpdateResponse, error) {
+	// TODO: wire up to real global-ratelimit aggregator
+	return nil, nil
```
taylanisikdemir, line 2063: nit: return unimplemented err

taylanisikdemir (approved):

proto/internal/uber/cadence/shared/v1/any.proto
```diff
@@ -0,0 +1,52 @@
...
+// of known encodings so it can be interpreted everywhere.  Any supports literally
+// any contents, and needs to be considered opaque until it is given to something
+// that is expecting it.
+//
+// See value_type to interpret the contents.
+message Any {
+  // Type-string describing value's contents, and intentionally avoiding the
+  // name "type" as it is often a special term.
+  // This should usually be a hard-coded string of some kind.
+  string value_type = 1;
```
davidporter-id-au, line 48: so, iirc proto.Any does some magic to dynamically download the encoding automatically (which, I don't love to be honest). Are you intending to use that functionality?

davidporter-id-au: As per offline discussion, suggest adding a unit test proving this out. My (bad) memory is just that this had a few dragons in it, including GoGo proto's odd choice to require a valid and addressable encoding http address. My memory's hopefully wrong

go.work.sum
```diff
@@ -262,6 +262,7 @@ github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802 h1:1BDTz0u9nC3//pOC
 github.com/DataDog/datadog-go v3.2.0+incompatible h1:qSG2N4FghB1He/r2mFrWKCaL7dXCilEuNEeAn20fdD4=
 github.com/DataDog/zstd v1.4.5/go.mod h1:1jcaCB/ufaK+sKp1NBhlGmpz41jOoPQ35bpF36t7BBo=
 github.com/GoogleCloudPlatform/cloudsql-proxy v1.22.0 h1:aTDBS16pX1X4ZR/GFsC2NcOCYJ1hDJwJm3WmKRA905Q=
+github.com/Groxx/cadence-idl v0.0.0-20240325232920-032063621ac4/go.mod h1:oyUK7GCNCRHCCyWyzifSzXpVrRYVBbAMHAzF5dXiKws=
```
davidporter-id-au, line 265: I assume you're going to update this?

proto/internal/uber/cadence/shared/v1/any.proto
Groxx: (also from IRL) There are some annoying bits with proto's Any, but I don't recall anything like that.  Though I'm definitely not certain it doesn't exist.

And if it does exist, we should probably run screaming for the hills and never ever touch anything that Even approaches it 😱

Groxx: And also from IRL: yes, very good point. I should add round-trip examples with thrift and proto objects inside the Any, because the odds of one or both of those eventually is pretty high imo.

Which shouldn't take too long. And if it does, that's probably a sign of problems.

go.work.sum
Groxx: Thought I had, but apparently not - will fix

common/types/mapper/proto/history_test.go
Groxx: it'd be a moderate bit larger after accounting for the data in the other files.
but that's not really relevant here, this type probably shouldn't ever gain other fields.

those tests are also simpler because they're _only_ round-trip checks, not "input X produces output Y".  bug-for-bug compatibility between mappers won't be caught, nor missed fields (which is why I spent the time to figure out a fuzzing pattern too).

common/types/mapper/thrift/shared_test.go
```diff
@@ -3310,3 +3317,157 @@ func TestStickyWorkerUnavailableErrorConversion(t *testing.T) {
...
+			ValueType: typeName,
+			Value:     internalBytes, // store thrift bytes in the Any
+		}
+
+		// --- convert the whole container to thrift (mimics making a call via yarpc)
+		thriftAny := FromAny(anyVal)      // we map to the rpc type
+		networkBytes := encode(thriftAny) // yarpc does this
+		// ^ this is what's sent over the network.
+
+		// the final data is not double-encoded, so this "encode -> wrap -> encode" process is reasonably efficient.
```
davidporter-id-au, line 3439: why is it not double-encoded? I may have missed that part

Groxx: (also via chat but for public purposes...)

semantically: it doesn't matter, no behavior depends on it.

practically: because thrift (and I'm pretty sure proto) store binary blobs as `size, data...` so there's no need to e.g. escape `\0` as `\\0` or base64-encode it or other costly stuff.  it's just a memcpy.

my main goal here is to point out that anything you'd do with normal thrift is reasonable to do when wrapped by this `Any`-container because it doesn't meaningfully increase any of the runtime costs.  push gigabytes / million-field things through it, it's entirely fine and no different than using the non-wrapped type.

go.work.sum
Shaddoll: still not fixed

common/types/mapper/thrift/shared_test.go
Groxx: I explained this more in the test comments now btw

go.work.sum
Groxx: yep, just cleaned it up now.
`go mod tidy` and `go work sync` are for some reason not removing it :\  no idea why.

go.work
```diff
@@ -19,6 +19,6 @@ use (
 	./cmd/server
 	./common/archiver/gcloud
 
-	// DO NOT include, tools dependencies are intentionally separate.
-	// ./internal/tools
+// DO NOT include, tools dependencies are intentionally separate.
+// ./internal/tools
```
Groxx, line 23: `go work sync` on go 1.20 did this for some reason.

no idea why, but it wants it, and it seems harmless.

davidporter-id-au (approved):

davidporter-id-au (approved):
lgtm, can't fault anything there

PR #5819, opened 2024-03-26
Fix log for archival

davidporter-id-au:
**What changed?**
The log to work out what's nonretriable is not logged

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/worker/archiver/activities.go
```diff
@@ -89,11 +89,17 @@ func uploadHistoryActivity(ctx context.Context, request ArchiveRequest) (err err
 	if err == nil {
 		return nil
 	}
-	if err.Error() == errUploadNonRetriable.Error() {
-		logger.Error(carchiver.ArchiveNonRetriableErrorMsg, tag.ArchivalArchiveFailReason("got non-retryable error from history archiver"))
+	if errors.Is(err, errUploadNonRetriable) {
```
Groxx, line 92: Superfically seems fine, but to double check: are we returning that exact error value?
Comparing strings is a looser check in many common ways, so generally this would be _less_ likely to be true than it was before.

davidporter-id-au: I don't know what error we're returning because the log isn't showing anything

davidporter-id-au: I did check this is coverd by a unit test before changing the errors.Is, I could remove it if it's a concern, but I don't really see a lot of values in jsut string comparing like this

3vilhamster (approved):
Since it is affecting only staging, maybe you could merge this from your custom branch to verify if it resolves the issues/helps identifying the root cause?

PR #5822, opened 2024-03-27
check time type in custom search attribute

bowenxia:
**What changed?**
Add a logic to check time type in custom search attribute.

**Why?**
Previously, if there's a time.Time() type passed in, like CustomDateTimeField = '2024-02-07T15:32:30Z', there'll be an error. 
In order to handle this, I converted it before forming the pinot query. 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/pinot/pinotQueryValidator.go
+++ b/common/pinot/pinotQueryValidator.go
@@ -328,14 +328,16 @@ func (qv *VisibilityQueryValidator) processCustomKey(expr sqlparser.Expr) (strin
 	if !ok {
 		return "", fmt.Errorf("invalid search attribute")
 	}
-	// get the value type
-	indexValType := common.ConvertIndexedValueTypeToInternalType(valType, log.NewNoop())
-
+	
 	// get the column value
 	colVal, ok := comparisonExpr.Right.(*sqlparser.SQLVal)
 	if !ok {
 		return "", errors.New("invalid comparison expression, right")
 	}
+
+	// get the value type
+	indexValType := common.ConvertIndexedValueTypeToInternalType(valType, log.NewNoop())
+
 	// if it is dataTime, then check if it is time.Time() type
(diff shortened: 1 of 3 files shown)


common/pinot/pinotQueryValidator.go
```diff
@@ -328,17 +328,24 @@ func (qv *VisibilityQueryValidator) processCustomKey(expr sqlparser.Expr) (strin
 	if !ok {
 		return "", fmt.Errorf("invalid search attribute")
 	}
+	// get the value type
+	indexValType := common.ConvertIndexedValueTypeToInternalType(valType, log.NewNoop())
```
neil-xie, line 332: let's move this line after get the colVal, so when it is invalid comparison expression, we don't need to do this conversion.

common/pinot/pinotQueryValidator.go
```diff
@@ -334,11 +334,20 @@ func (qv *VisibilityQueryValidator) processCustomKey(expr sqlparser.Expr) (strin
 	if !ok {
 		return "", errors.New("invalid comparison expression, right")
 	}
-	colValStr := string(colVal.Val)
 
 	// get the value type
 	indexValType := common.ConvertIndexedValueTypeToInternalType(valType, log.NewNoop())
 
+	// if it is dataTime, then check if it is time.Time() type
+	if indexValType == types.IndexedValueTypeDatetime {
```
neil-xie, line 342: can we move this to L358, so if it is type IndexedValueTypeDatetime, we trim then return processCustomNum

bowenxia: There'll be an extra step to convert the result to a string then. `colValStr := string(colVal.Val)`

neil-xie (approved):
LGTM

PR #5823, opened 2024-03-29
Notify named subscribers for membership changes

mantas-sidlauskas:
**What changed?**
Notify "named" membership change subscribers

**Why?**
There are two levels of membership change notification mechanisms. One is per-service (frontend, history, etc), another one is for services to use in their internal components (like shard controller in history service).

For service-level subscribers, there was a missing piece of code to send notification on registered channel. As this is a bug, automatic 10s interval refresh helped to mitigate this bug. Now service-level subscribers will be informed immediately. 

**How did you test it?**
Added unit test to check if notifications are sent when receiving message from "peer provider" (ringpop for example)

**Potential risks**

**Release notes**

**Documentation Changes**


common/membership/hashring.go
```diff
@@ -265,7 +273,8 @@ func (r *ring) refreshRingWorker() {
 		select {
 		case <-r.shutdownCh:
 			return
-		case <-r.refreshChan: // local signal or signal from provider
+		case event := <-r.refreshChan: // local signal or signal from provider
+			r.notifySubscribers(event)
```
Groxx, line 277: why notify about "a refresh has been requested"?  doesn't seem related to what subscribers are observing

mantas-sidlauskas: hmm, this case is for notification from peer provider (ringpop by default). When ringpop sends a an event to refreshChan, we notify "service-level" subscribers.

taylanisikdemir (approved):

common/membership/hashring.go
Groxx: that's... rather trivially incorrect about what's happening in this code

Groxx (changes requested):
blocking until core issues are cleared up

common/membership/hashring.go
Groxx: yea... the whole setup of writing to the notification channel seems to assume asynchronous work (non-blocking writes in a tight loop), so any notified subscriber would get a blend of old and new data if they call `Lookup(...)`.  that seems very unlikely to be desired.

e.g. the acquireShards subscriber does exactly that, and calls `Lookup(..)` for all shards, racing with the actual new-data update.

taylanisikdemir: Good point about the order of operations here. It looks like if we do `r.notifySubscribers(event)` after `r.refresh()` then the acquireShards will do `Lookup()` on new data.

I also just checked the shard controller's [membershipUpdateCh](https://github.com/uber/cadence/blob/5003ffbff3d7ee7ba430b8bc601865448650c265/service/history/shard/controller.go#L128) which is initialized with 10 capacity. So notifications will most likely be unblocked but acquiring shards for each change might cause too many redundant work. Since r.refresh() already has [4s cooldown](https://github.com/uber/cadence/blob/ff95842fb10d2d9630bbd1665d3d02430b447d9c/common/membership/hashring.go#L231), it should return a bool to indicate whether a refresh happened and then only notify subscribers if refresh occurred

common/membership/hashring_test.go
```diff
@@ -128,15 +128,40 @@ func TestRefreshUpdatesRingOnlyWhenRingHasChanged(t *testing.T) {
...
+		HostsAdded:   []string{"a"},
+		HostsUpdated: []string{"b"},
+		HostsRemoved: []string{"c"},
+	}
+
+	hr := newHashring("test-service", pp, log.NewNoop(), metrics.NoopScope(0))
+	hr.Start()
+	var changeCh = make(chan *ChangedEvent)
+
+	err := hr.Subscribe("notifyMe", changeCh)
```
taylanisikdemir, line 147: can you add multiple subscribers just to make sure iteration works

mantas-sidlauskas: yep, it's a good idea. thanks

common/membership/hashring.go
```diff
@@ -265,10 +273,11 @@ func (r *ring) refreshRingWorker() {
 		select {
 		case <-r.shutdownCh:
 			return
-		case <-r.refreshChan: // local signal or signal from provider
+		case event := <-r.refreshChan: // local signal or signal from provider
 			if err := r.refresh(); err != nil {
 				r.logger.Error("refreshing ring", tag.Error(err))
 			}
+			r.notifySubscribers(event)
```
taylanisikdemir, line 280: As mentioned in the other comment, let's change refresh to return a bool to indicate refresh occurred.

```
refreshed, err := r.refresh()
if err != nil {
  r.logger.Error("failed to refresh ring", tag.Error(err))
}
if refreshed && err == nil {
  r.notifySubscribers(event)
}
```

common/membership/hashring_test.go
mantas-sidlauskas: tests are updated

common/membership/hashring.go
mantas-sidlauskas: done

common/membership/hashring.go
mantas-sidlauskas: now we check if ring has actually changed and only after that subscribers are notified

taylanisikdemir (approved):

common/membership/hashring_test.go
```diff
@@ -116,27 +116,85 @@ func TestRefreshUpdatesRingOnlyWhenRingHasChanged(t *testing.T) {
 	pp := NewMockPeerProvider(ctrl)
 
 	pp.EXPECT().Subscribe(gomock.Any(), gomock.Any()).Times(1)
-	pp.EXPECT().GetMembers("test-service").Times(3)
+	pp.EXPECT().GetMembers("test-service").Times(1).Return(randomHostInfo(3), nil)
```
Groxx, line 119: hmm.  any idea why this changed?

`Times(1)` seems correct since these start/refresh/refresh should mean one call during start's refresh, and then the other two are "too soon" and are skipped...:
```go
func (r *ring) refresh() error {
	if r.members.refreshed.After(time.Now().Add(-minRefreshInternal)) {
		// refreshed too frequently
		return nil
	}
```
but this was true before as well, and I don't see any time-manipulating in this test that would have made `Times(3)` correct previously.

common/membership/hashring_test.go
```diff
@@ -116,27 +116,85 @@ func TestRefreshUpdatesRingOnlyWhenRingHasChanged(t *testing.T) {
...
+	assert.NoError(t, hr.Subscribe("notifyMe2", changeCh2))
+
+	wg := sync.WaitGroup{}
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		changedEvent := <-changeCh1
+		close(changeCh1)
+		assert.Equal(t, changed, changedEvent)
+	}()
```
Groxx, line 170: nothing guarantees these goroutines reach the `<-changeCh1` read before the the `hr.refreshChan <- changed` occurs, which would lead to missing the notification.

probably just buffer these chans?

common/membership/hashring_test.go
```diff
@@ -116,27 +116,85 @@ func TestRefreshUpdatesRingOnlyWhenRingHasChanged(t *testing.T) {
...
+		assert.Equal(t, changed, changedEvent2)
+	}()
+	// to bypass internal check
+	hr.members.refreshed = time.Now().AddDate(0, 0, -1)
+	hr.refreshChan <- changed
+	// Test if internal members are updated
+	hostLock.RLock()
+	assert.ElementsMatch(t, hr.Members(), hostsToReturn, "members should contain just-added nodes")
+	hostLock.RUnlock()
+	wg.Wait() // wait until both subscribers will get notification
```
Groxx, line 185: similarly, nothing ensures the `refresh()`->DoAndReturn write has occurred by this point.

waiting before comparing solves that, and removes the need to lock in the first place.

Groxx (changes requested):
- Q on a surprising change (just not sure how the changes lead to it, if there's a cause/effect I'm missing then it's totally fine and it looks correct to me)
- tests are rather highly racy and not reliable

Groxx (approved):
Yea, looks correct and I also can't get any test failures by stressing it*.  Will merge as soon as the merge-and-retest finishes 👍 

It'd be nice to shift all this to the mock clock so we can reduce some sleeps / _actually_ ensure that only the test-triggered refresh runs (and not the ticker, though that'd take absurd machine load to hit), but that's well outside the scope of this PR.

\* unless it runs too long / `-count 1000` but that appears to be due to not shutting down hashrings in _any_ of these tests.

PR #5826, opened 2024-03-29
Persist workflow request ids into Cassandra

Shaddoll:
# WorkflowRequest:
WorkflowRequest is a new type of data introduced in this PR which will be stored in database to allow detecting duplicate requests. By detecting duplicate requests, we can guarantee idempotency of the following APIs: 
- StartWorkflowExecution
- SignalWorkflowExecution
- CancelWorkflowExecution
- ResetWorkflowExecution
- SignalWithStartWorkflowExecution

A workflow request is identified by (`shard_id`, `domain_id`, `workflow_id`, `request_id`, `request_type`, `version`), and `run_id` is associated indicating which workflow the request has been applied to. Among them, `shard_id` is derived from `workflow_id`, and `version` is the failover version of the domain when the request is applied. `request_id` is a UUID set in Cadence API requests.

## request_type:
Normally, `request_id` is not reused by clients because it's a UUID, but it may be reused incorrectly by different APIs. For example, a SignalWorkflow request with `request_id: 7cfc8e22-b252-447d-ab03-9e5eda68dc35` has been applied to workflow `test-wf` and later a CancelWorkflow request with the same request_id is sent to the same workflow. In such case, we need to handle the CancelWorkflow request properly. There are 3 options to handle the second request:
1. Make it a no-op
2. Apply the request
3. Return an error telling the customer that the request_id shouldn't be reused

Our implementation chooses the 2nd option, because it's more reasonable compared to option 1 and requires no change from client side compared to option 3. `request_type` is used for implementation of 2nd option, and without `request_type` we can only go with option 1, because we cannot differentiate different API requests.

We currently support 4 request types:
- WorkflowRequestTypeStart
- WorkflowRequestTypeSignal
- WorkflowRequestTypeCancel
- WorkflowRequestTypeReset

And they are mapped to the APIs mentioned above. However, `SignalWithStartWorkflowExecution` is treated differently. This single API request is mapped to 2 workflow requests stored in database, 1 with type `WorkflowRequestTypeStart` and 1 with type `WorkflowRequestTypeSignal`. And if a workflow is signaled and started by this API with `request_id: 7cfc8e22-b252-447d-ab03-9e5eda68dc35`. Not only SignalWithStartWorkflowExecution with the same `request_id` will be no-op, but also `StartWorkflowExecution` and `SignalWorkflowExecution` with the same `request_id` will also be no-op. This is to simplify the implementation and keeps the behavior the same with the current idempotency implementation.

# Replication of WorkflowRequest:
WorkflowRequests need to be replicated to guarantee idempotency of APIs for global domains after a domain failover. But replication is done asynchronously, and by the time a workflow request is replicated, there might be already an existing request with the same `request_id` in the database, because the request is retried and applied after failover before getting replicated from the old cluster. `version` is used for such conflict resolution. And we also introduce `CreateWorkflowRequestMode` when writing workflow_requests to database.
- `CreateWorkflowRequestModeNew` is used for workflow_requests generated from Cadence API requests. And we should not apply the request if there is already a workflow_request with the same `request_id` in the database, even if the `version` is different.
- `CreateWorkflowRequestModeReplicated` is used for workflow_requests generated from replication. And we should always apply the request to achieve eventual consistency.

# Cassandra Implementation:
Because Cassandra doesn't support cross-table LWT, the workflow_request has to be stored in the existing `execution` table so that the data can be inserted as a part of the transaction which stores the change of mutable state. We uses the following columns of execution table to store workflow_request data:
- shard_id -> shard_id
- type -> request_type
- domain_id -> domain_id
- workflow_id -> workflow_id
- run_id -> request_id
- task_id -> version * -1 (this is because we want to fetch the data in descending order)
- current_run_id -> run_id

# Risk:
This PR has no risk, because it only defines the struct and interface needed for storing workflow_requests. A following PR will use the change and stores data in database.

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
@@ -22,6 +22,7 @@
 package cassandra
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"strings"
@@ -35,6 +36,7 @@ import (
 )
 
 func executeCreateWorkflowBatchTransaction(
+	ctx context.Context,
 	session gocql.Session,
 	batch gocql.Batch,
 	currentWorkflowRequest *nosqlplugin.CurrentWorkflowWriteRequest,
@@ -57,12 +59,29 @@ func executeCreateWorkflowBatchTransaction(
 		return nil
 	}
 
+	requestRangeID := shardCondition.RangeID
 	requestConditionalRunID := ""
 	if currentWorkflowRequest.Condition != nil {
 		requestConditionalRunID = currentWorkflowRequest.Condition.GetCurrentRunID()
 	}
 	// There can be two reasons why the query does not get applied. Either the RangeID has changed, or
 	// the workflow is already started. Check the row info returned by Cassandra to figure out which one it is.
+	rangeIDMismatch := false
+	actualRangeID := int64(0)
+	currentExecutionAlreadyExists := false
+	var actualExecution map[string]interface{}
+	runIDMismatch := false
+	actualCurrRunID := ""
+	lastWriteVersionMismatch := false
+	actualLastWriteVersion := int64(common.EmptyVersion)
+	stateMismatch := false
+	actualState := int(0)
+	concreteExecutionAlreadyExists := false
+	workflowRequestAlreadyExists := false
+	workflowRequestID := ""
+	requestRowType := int(0)
+	var allPrevious []map[string]interface{}
+
 	for {
 		rowType, ok := previous["type"].(int)
 		if !ok {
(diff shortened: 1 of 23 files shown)


common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -57,95 +59,79 @@ func executeCreateWorkflowBatchTransaction(
...
-
 		if rowType == rowTypeShard {
-			if rangeID, ok := previous["range_id"].(int64); ok && rangeID != shardCondition.RangeID {
-				// CreateWorkflowExecution failed because rangeID was modified
-				return &nosqlplugin.WorkflowOperationConditionFailure{
-					ShardRangeIDNotMatch: common.Int64Ptr(rangeID),
-				}
+			if actualRangeID, ok = previous["range_id"].(int64); ok && actualRangeID != requestRangeID {
+				// UpdateWorkflowExecution failed because rangeID was modified
+				rangeIDMismatch = true
```
taylanisikdemir, line 95: it's much easier to trace the code with early termination. let's just return WorkflowOperationConditionFailure here as before

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -57,95 +59,79 @@ func executeCreateWorkflowBatchTransaction(
...
-			if prevRunID := previous["current_run_id"].(gocql.UUID).String(); requestConditionalRunID != "" && prevRunID != requestConditionalRunID {
-				// currentRunID on previous run has been changed, return to caller to handle
-				msg := fmt.Sprintf("Workflow execution creation condition failed by mismatch runID. WorkflowId: %v, Expected Current RunID: %v, Actual Current RunID: %v",
-					execution.WorkflowID, currentWorkflowRequest.Condition.GetCurrentRunID(), prevRunID)
-				return &nosqlplugin.WorkflowOperationConditionFailure{
-					CurrentWorkflowConditionFailInfo: &msg,
+				if currentWorkflowRequest.Condition != nil && currentWorkflowRequest.Condition.LastWriteVersion != nil {
+					ok := false
+					if actualLastWriteVersion, ok = previous["workflow_last_write_version"].(int64); ok && *currentWorkflowRequest.Condition.LastWriteVersion != actualLastWriteVersion {
+						lastWriteVersionMismatch = true
```
taylanisikdemir, line 113: same here. just return the err here instead of keeping a bool flag for it which is handled below. the more local the decision and action is the more readable/maintainable the code is

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -57,95 +59,79 @@ func executeCreateWorkflowBatchTransaction(
...
+					}
+				}
+				if currentWorkflowRequest.Condition != nil && currentWorkflowRequest.Condition.State != nil {
+					ok := false
+					if actualState, ok = previous["workflow_state"].(int); ok && *currentWorkflowRequest.Condition.State != actualState {
+						stateMismatch = true
```
taylanisikdemir, line 119: same here

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -57,95 +59,79 @@ func executeCreateWorkflowBatchTransaction(
...
-			msg := fmt.Sprintf("Workflow execution creation condition failed. WorkflowId: %v, CurrentRunID: %v, columns: (%v)",
-				execution.WorkflowID, execution.RunID, strings.Join(columns, ","))
-			return &nosqlplugin.WorkflowOperationConditionFailure{
-				CurrentWorkflowConditionFailInfo: &msg,
 			}
 		} else if rowType == rowTypeExecution && execution.RunID == runID {
-			msg := fmt.Sprintf("Workflow execution already running. WorkflowId: %v, RunId: %v, rangeID: %v",
-				execution.WorkflowID, execution.RunID, shardCondition.RangeID)
-			lastWriteVersion := common.EmptyVersion
+			concreteExecutionAlreadyExists = true
```
taylanisikdemir, line 124: same here

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go
```diff
@@ -259,35 +320,70 @@ func TestExecuteCreateWorkflowBatchTransaction(t *testing.T) {
...
+					DomainID:   "fd88863f-bb32-4daa-8878-49e08b91545e",
+					WorkflowID: "wfid",
+				},
+			},
 			shardCond: &nosqlplugin.ShardCondition{
+				ShardID: 1,
 				RangeID: 100,
 			},
 			wantErr: &nosqlplugin.WorkflowOperationConditionFailure{
-				CurrentWorkflowConditionFailInfo: common.StringPtr(
```
taylanisikdemir, line 284: why removing this case? is it covered by something else? If possible let's leave existing test cases intact to avoid changing behavior

taylanisikdemir (commented):
overall LGTM but due to a lot of conditional changes in workflow_utils.go it's hard to review. I'd avoid changing existing branches there and focus on addition of new functionality in this PR. In a follow up PR we can improve those if needed

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -154,10 +140,101 @@ func executeCreateWorkflowBatchTransaction(
...
+			shardCondition.ShardID,
+			requestRowType,
+			execution.DomainID,
+			execution.WorkflowID,
+			workflowRequestID,
+			defaultVisibilityTimestamp,
+		).WithContext(ctx).MapScan(result); err != nil {
+			return err
+		}
+		runID := result["current_run_id"].(gocql.UUID).String()
```
taylanisikdemir, line 161: check if result has `current_run_id` before casting

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
```diff
@@ -1357,6 +1454,75 @@ func createWorkflowExecution(
 	return nil
 }
 
+func isRequestRowType(rowType int) bool {
```
taylanisikdemir, line 1457: add a specific unit test for this and also leave a comment on top of `rowTypeWorkflowRequestStart` as a reminder to update this when new request types are introduced.

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
Shaddoll: we need to check the errors in a specific order. This early return doesn't allow us to check the errors in the order we want.

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils_test.go
Shaddoll: it's renamed to `current execution already exists and write mode is CurrentWorkflowWriteModeInsert`

common/persistence/nosql/nosqlplugin/cassandra/workflow_utils.go
taylanisikdemir: ok makes sense.

common/persistence/persistence-tests/executionManagerTest.go
```diff
@@ -408,6 +477,105 @@ func (s *ExecutionManagerSuite) TestCreateWorkflowExecutionWithZombieState() {
...
+	updateReq := &p.UpdateWorkflowExecutionRequest{
+		UpdateWorkflowMutation: p.WorkflowMutation{
+			ExecutionInfo:    updatedInfo,
+			ExecutionStats:   updatedStats,
+			Condition:        nextEventID,
+			Checksum:         csum,
+			VersionHistories: versionHistories,
+			WorkflowRequests: []*p.WorkflowRequest{
+				{
+					RequestID:   requestID,
```
taylanisikdemir, line 561: another test case similar to this function but with different requestID

taylanisikdemir (approved):

PR #5831, opened 2024-04-01
Add unit tests for execution context

Shaddoll:
**What changed?**
Add unit tests for execution context

**Why?**
Improve coverage

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/context.go
+++ b/service/history/execution/context.go
@@ -40,6 +40,7 @@ import (
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/types"
 	hcommon "github.com/uber/cadence/service/history/common"
+	"github.com/uber/cadence/service/history/engine"
 	"github.com/uber/cadence/service/history/events"
 	"github.com/uber/cadence/service/history/shard"
 )
@@ -174,6 +175,11 @@ type (
 		mutableState    MutableState
 		stats           *persistence.ExecutionStats
 		updateCondition int64
+
+		appendHistoryNodesFn              func(context.Context, string, types.WorkflowExecution, *persistence.AppendHistoryNodesRequest) (*persistence.AppendHistoryNodesResponse, error)
+		createWorkflowExecutionFn         func(context.Context, *persistence.CreateWorkflowExecutionRequest) (*persistence.CreateWorkflowExecutionResponse, error)
+		notifyTasksFromWorkflowSnapshotFn func(*persistence.WorkflowSnapshot, events.PersistedBlobs, bool)
+		emitSessionUpdateStatsFn          func(string, *persistence.MutableStateUpdateSessionStats)
 	}
 )
 
@@ -187,6 +193,7 @@ func NewContext(
 	executionManager persistence.ExecutionManager,
 	logger log.Logger,
 ) Context {
+	logger = logger.WithTags(tag.WorkflowDomainID(domainID), tag.WorkflowID(execution.GetWorkflowID()), tag.WorkflowRunID(execution.GetRunID()))
 	return &contextImpl{
 		domainID:          domainID,
 		workflowExecution: execution,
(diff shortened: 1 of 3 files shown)


service/history/execution/context.go
```diff
@@ -1182,11 +1197,8 @@ func (c *contextImpl) createWorkflowExecutionWithRetry(
 		// workflow ID reuse policy
 		return nil, err
 	default:
-		c.logger.Error(
+		logger.Error(
 			"Persistent store operation failure",
-			tag.WorkflowID(c.workflowExecution.GetWorkflowID()),
-			tag.WorkflowRunID(c.workflowExecution.GetRunID()),
-			tag.WorkflowDomainID(c.domainID),
```
taylanisikdemir, line 1189: why removing these log tags? It's fine if this function needs to live detached from `contextImpl` but let's not lose these useful tags. Pass them as parameter maybe?

service/history/execution/context_test.go
```diff
@@ -0,0 +1,939 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			mockShard := shard.NewMockContext(mockCtrl)
+			if tc.mockSetup != nil {
+				tc.mockSetup(mockShard)
+			}
+			resp, err := createWorkflowExecutionWithRetry(context.Background(), mockShard, testlogger.New(t), tc.request)
```
taylanisikdemir, line 587: I'd make the retry policy overridable so test can pass low value for max retry attempt/duration. 
You can then add a case which keeps failing even after all attempts

service/history/execution/context_test.go
```diff
@@ -0,0 +1,939 @@
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			mockShard := shard.NewMockContext(mockCtrl)
+			if tc.mockSetup != nil {
+				tc.mockSetup(mockShard)
+			}
+			resp, err := appendHistoryV2EventsWithRetry(context.Background(), mockShard, tc.domainID, tc.execution, tc.request)
```
taylanisikdemir, line 684: same retry policy overridability applies here too

service/history/execution/context.go
Shaddoll: it's not removed. We added these tags using `WithTags` method during the initialization step of context

taylanisikdemir: ok perfect. missed that

service/history/execution/context.go
```diff
@@ -198,6 +205,37 @@ func NewContext(
...
+					// TODO: is timeout error retryable for create workflow?
+					// if we treat it as retryable, user may receive workflowAlreadyRunning error
+					// on the first start workflow execution request.
+					return false
+				}
+				return persistence.IsTransientError(err)
+			}
+
+			throttleRetry := backoff.NewThrottleRetry(
+				backoff.WithRetryPolicy(common.CreatePersistenceRetryPolicy()),
```
taylanisikdemir, line 228: you could just add override for the policy instead of making the surrounding code an inline function. this inline function is not covered now

service/history/execution/context.go
```diff
@@ -704,8 +720,8 @@ func (c *contextImpl) UpdateWorkflowExecutionTasks(
 		DomainName: domainName,
 	})
 	if err != nil {
-		if c.isPersistenceTimeoutError(err) {
-			c.notifyTasksFromWorkflowMutation(currentWorkflow, nil, true)
+		if isOperationPossiblySuccessfulError(err) {
+			notifyTasksFromWorkflowMutation(c.shard.GetEngine(), currentWorkflow, nil, true)
```
taylanisikdemir, line 724: this is not related to this PR but what happens if operation actually timed out and engine is still notified about those tasks that were not persisted?

taylanisikdemir (approved):

service/history/execution/context.go
Shaddoll: the tasks are possibly persisted, so we still notify the engine to try fetching those tasks

PR #5853, opened 2024-04-03
Added some validations in the nosql store and new tests

agautam478:
**What changed?**
Added tests for the noswl_execution_store.go.

I noticed that some methods are lacking common validations so added those as well.

**Why?**
increase code coveration

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/nosql/nosql_execution_store_test.go
```diff
@@ -860,6 +860,211 @@ func TestNosqlExecutionStore(t *testing.T) {
...
+			name: "GetReplicationDLQSize success",
+			setupMock: func(ctrl *gomock.Controller) *nosqlExecutionStore {
+				mockDB := nosqlplugin.NewMockDB(ctrl)
+				mockDB.EXPECT().
+					SelectReplicationDLQTasksCount(ctx, shardID, "sourceCluster").
+					Return(int64(42), nil)
+				return newTestNosqlExecutionStore(mockDB, log.NewNoop())
+			},
+			testFunc: func(store *nosqlExecutionStore) error {
+				resp, err := store.GetReplicationDLQSize(ctx, &persistence.GetReplicationDLQSizeRequest{
```
taylanisikdemir, line 1038: It's fine here since you already wrote this but going forward I recommend having separate top level TestXyz function for each store function and avoid a `testFunc` in the table. Table tests should provide inputs, mock setup and expected results. Glueing it all together and doing validations should be left to test body inside t.Run. All table tests share same test body function.

taylanisikdemir (approved):

common/persistence/nosql/nosql_execution_store_test.go
agautam478: got it!

PR #5854, opened 2024-04-03
Added test cases for more util methods

agautam478:
**What changed?**
And more test cases for nosqlutil.

**Why?**
to increase code coverage.

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


common/persistence/nosql/nosql_execution_store_util.go
```diff
@@ -568,6 +568,10 @@ func (d *nosqlExecutionStore) prepareRequestCancelsForWorkflowTxn(requestCancels
 func (d *nosqlExecutionStore) prepareSignalInfosForWorkflowTxn(signalInfos []*persistence.SignalInfo) (map[int64]*persistence.SignalInfo, error) {
 	m := map[int64]*persistence.SignalInfo{}
 	for _, c := range signalInfos {
+		if _, exists := m[c.InitiatedID]; exists {
+			// Duplicate InitiatedID found, return an error
+			return nil, fmt.Errorf("duplicate signal ID detected: %d", c.InitiatedID)
```
taylanisikdemir, line 573: will this cause retries?

taylanisikdemir: I think it will be retried https://github.com/uber/cadence/blob/fcbc00ec11ed0af72fae91ac1922ca9b2e87594b/service/history/execution/context.go#L1275. You might want to make this change separately by carefully picking a custom error type to return and then handling it in places determining retryability

agautam478: This is used in create, update and and Reset without retry executions so I believe no. I have noticed that a lot of these utils have years old TODO validations that haven't been implemented.

agautam478: > I think it will be retried
> 
> https://github.com/uber/cadence/blob/fcbc00ec11ed0af72fae91ac1922ca9b2e87594b/service/history/execution/context.go#L1275
> 
> . You might want to make this change separately by carefully picking a custom error type to return and then handling it in places determining retryability

Alright! Will remove this test for now.

taylanisikdemir: thanks for improving the underlying as you go. we should look at call chain and implications carefully while introducing new error cases. better to handle in a follow up PR in this case

taylanisikdemir (approved):

PR #5857, opened 2024-04-04
fixed timer queue processor goroutine leak in test

shijiesheng:
**What changed?**

* ~~Correctly stop timer queue processor on InitializedState~~ Instead, stop timergate in unit test only and added a comment for improvement
* added basic start stop test on transfer queue processor

**Why?**

Previously in initialized state, stopping timer queue processor will not stop the timer gate which has a daemon loop; although this shall never happen in production, it's causing go routine leak in unit tests

**How did you test it?**

unit test

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/queue/timer_queue_processor_base.go
```diff
@@ -176,7 +176,10 @@ func (t *timerQueueProcessorBase) Start() {
 
 func (t *timerQueueProcessorBase) Stop() {
 	if !atomic.CompareAndSwapInt32(&t.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
-		return
+		// initialized should also be stopped to stop TimerGate
+		if !atomic.CompareAndSwapInt32(&t.status, common.DaemonStatusInitialized, common.DaemonStatusStopped) {
```
taylanisikdemir, line 180: This is a good catch but instead of solving problem here we can fix TimerGate itself. It should have Start/Stop functions and shouldn't create any goroutine in constructor. That would be cleaner solution I think.

service/history/queue/transfer_queue_processor_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/uber/cadence/service/history/reset"
+	"github.com/uber/cadence/service/history/shard"
+	"github.com/uber/cadence/service/history/task"
+	"github.com/uber/cadence/service/history/workflowcache"
+	"github.com/uber/cadence/service/worker/archiver"
+)
+
+func TestTransferQueueProcessor_RequireStartStop(t *testing.T) {
+	// some goroutine leak not from this test
+	defer goleak.VerifyNone(t)
```
taylanisikdemir, line 44: great idea to verify leaks in these critical components with bunch of underlying goroutines!

service/history/queue/timer_queue_processor_base.go
shijiesheng: I still think the code is buggy since lifecycle of timer gate is not handled correctly here.
* if you just initialize the timer queue processor without starting it, timer gate needs to be stopped separately
* if you start timer queue processor and stop it, timer gate is closed. 

 But to make minimum change. I'll handle this in the test case. The remaining concern is whether we do create timer gate without close it in the first scenario. But it should be addressed separately. I'll just add a comment.

taylanisikdemir (approved):

PR #5859, opened 2024-04-04
persistence/client/factory.go test coverage

Groxx:
Our ability to cover this file is rather harshly limited due to it directly calling funcs in other packages that try to make network connections... but thankfully the DataStore-based stuff can be intercepted.  So I've at least tested those branches.

More generally this package is yet another strong argument in favor of leaning more on go.uber.org/fx for dependency injection.  Very little of this has to be (or should be) hard-coded in this file.  It feels like Java's "Service Locator" pattern, but even worse because the language doesn't allow test controls of globals like Java does... so I doubt it's a coincidence that there's a `bean.go` file in here.


common/persistence/client/factory.go
```diff
@@ -18,6 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination factory_mock.go
```
Groxx, line 21: unfortunately this doesn't allow choosing the interfaces.
I only need the `DataStoreFactory`.

common/log/testlogger/testlogger.go
```diff
@@ -34,11 +31,6 @@ import (
 
 // New is a helper to create new development logger in unit test
 func New(t zaptest.TestingT) log.Logger {
-	if testing.Verbose() {
-		logger, err := loggerimpl.NewDevelopment()
-		require.NoError(t, err)
-		return logger
-	}
```
Groxx, line 41: I have no idea why anyone ever thought this was a good idea, so I removed it.

Since CI runs tests verbosely this _might_ lead to more race failures in tests, as it touches `*testing.T` more than it did before, and there are quite a lot of async logging goroutines that are not cleanly shut down across this repo.

**If** that happens and proves onerous to fix or work around, we could undo this.  So far I haven't seen other tests fail due to this change though.

taylanisikdemir (approved):

PR #5863, opened 2024-04-04
stop redirection for deprecated domains

3vilhamster:
Handle deprecated and deleted domains on frontend redirection and add context to panics inside redirection.

**What changed?**
If a domain is deleted or deprecated, redirecting to the other cluster is pointless. So, it should be safe to return an error.
Also, redirection can panic if the domain has an active cluster that is not currently connected to the cluster. I've added domain information to the panic message so it will be easier to track offending callers.

**Why?**
In Uber, we have a lot of clusters, and domains are migrated across them. If a customer is calling to the wrong environment, it causes panic during redirection. We want to:
a. Improve visibility of these failures to identify problematic domains.
b. Deprecate domains that were migrated out to stop it altogether.

**How did you test it?**
I added a unit test for a deprecated domain case.

**Potential risks**
The outgoing error for deprecated domain calls can change because the error will now be returned on the frontend without propagating the request to underlying layers.

**Release notes**
cadence-frontend no longer redirects calls to deprecated domains.

**Documentation Changes**


service/frontend/wrappers/clusterredirection/callwrappers.go
```diff
@@ -0,0 +1,62 @@
...
+	domainName string,
+	domainID string,
+	cluster string,
+	retError *error,
+) {
+	var domainTag tag.Tag
+	if domainName != "" {
+		domainTag = tag.WorkflowDomainName(domainName)
+	} else if domainID != "" {
+		domainTag = tag.WorkflowDomainID(domainID)
```
Shaddoll, line 52: can we get the domainName from domainID and only use domainName in metric tags?

Shaddoll (approved):

service/frontend/wrappers/clusterredirection/api_generated.go
```diff
@@ -736,17 +743,18 @@ func (handler *clusterRedirectionHandler) RespondDecisionTaskFailed(ctx context.
 	var apiName = "RespondDecisionTaskFailed"
 	var cluster string
 
+	var token domainIDGetter
 	scope, startTime := handler.beforeCall(metrics.DCRedirectionRespondDecisionTaskFailedScope)
 	defer func() {
-		handler.afterCall(recover(), scope, startTime, cluster, &err)
+		handler.afterCall(recover(), scope, startTime, "", token.GetDomainID(), cluster, &err)
```
taylanisikdemir, line 749: `token` will be nil if it's failed to be created and returned err at 754.

3vilhamster: Good catch, updated the code with noopdomainIDGetter

service/frontend/wrappers/clusterredirection/callwrappers.go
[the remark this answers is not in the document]
3vilhamster: It is never the case since we use only them. But I don't mind, so I added both tags if they are provided.

PR #5876, opened 2024-04-05
Tests: For handler: Register Domain in common/domain folder

abhishekj720:
**What changed?**
Added tests for registerDomain for handler in common/domain

**Why?**
As part of test coverage week.

**How did you test it?**
Unit Test

**Potential risks**
Tests are written to avoid risks

### DIFF
--- a/common/domain/handler_test.go
+++ b/common/domain/handler_test.go
@@ -1,0 +1,160 @@
+// Copyright (c) 2024 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


common/domain/handler_test.go
```diff
@@ -0,0 +1,160 @@
...
+	mockArchiverProvider provider.ArchiverProvider
+	handler              Handler
+	context              context.Context
+}
+
+func TestHandlerSuite(t *testing.T) {
+	suite.Run(t, new(HandlerSuite))
+}
+
+func (s *HandlerSuite) SetupTest() {
```
taylanisikdemir, line 55: each table test should start fresh instead of reusing same `handler`. So I'd ditch the suite and make this a helper function to be called inside t.Run

agautam478 (approved):

PR #5882, opened 2024-04-05
Added mocks for interfaces and also unit tests for functions used in history handler

timl3136:
**What changed?**
Added mocks for interface and also unit tests for functions used in history handler

**Why?**
Improve test coverage and stability

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


.gen/go/history/history.go
bowenxia, line 1: Are those .gen files commits intentional?

service/history/handler/handler_test.go
```diff
@@ -33,34 +33,44 @@ import (
...
 
 const (
 	testWorkflowID    = "test-workflow-id"
 	testWorkflowRunID = "test-workflow-run-id"
 	testDomainID      = "BF80C53A-ED56-4DD9-84EB-BE9AD4E45867"
+	testValidUUID     = "FCD00931-EBD4-4028-B67E-4DE624641255"
 )
 
 type (
 	handlerSuite struct {
```
bowenxia, line 61: Also, @taylanisikdemir mentioned that a better practice is _not_ to use test suites because it is better to initiate things like mocks within the smallest range (in each test case).

timl3136: We can refractor the code later. For now, I am just writing tests based on the existing infrastructure.

agautam478 (approved):

PR #5886, opened 2024-04-07
Added tests for RegisterDomain, ListDomain and DescribeDomain

abhishekj720:
**What changed?**
Tests:
- Register domain
- List domain
- describe domain

**Why?**
As part of test coverage week

**How did you test it?**
Unit test

**Potential risks**
Tests to avoid risk

**Release notes**
N.A.

**Documentation Changes**
Not required

### DIFF
--- a/common/domain/handler_test.go
+++ b/common/domain/handler_test.go
@@ -23,11 +23,14 @@ package domain
 import (
 	"context"
 	"errors"
+	"net/url"
 	"testing"
+	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
 
+	"github.com/uber/cadence/common"
 	"github.com/uber/cadence/common/archiver"
 	"github.com/uber/cadence/common/archiver/provider"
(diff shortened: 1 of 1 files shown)


common/domain/handler_test.go
```diff
@@ -244,3 +342,233 @@ func TestRegisterDomain(t *testing.T) {
...
+				{ClusterName: "standby"},
+			},
+		},
+	}
+
+	resp, err := handler.DescribeDomain(context.Background(), describeRequest)
+	assert.NoError(t, err)
+	assert.EqualValues(t, expectedResponse, resp)
+
+	mockDomainManager.EXPECT().
```
ketsiambaku, line 566: nit: my understanding is that this is testing a different scenario. Maybe organize the test with have test names so it's easier to understand

abhishekj720: It is just a err!= nil  1 case. that is why did add table test etc. But you are right, i ll add the comment whuich case it demonstrates in next diff.

PR #5888, opened 2024-04-08
Increased timeout value for delete history event task

timl3136:
**What changed?**
We increased the timeout value for delete history event task to dynamic config with default at 30s.

**Why?**
The timeout value for delete history event is too low causing large history deletion to constantly timeout

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/task/timer_standby_task_executor.go
```diff
@@ -110,7 +110,10 @@ func (t *timerStandbyTaskExecutor) Execute(
 	case persistence.TaskTypeWorkflowBackoffTimer:
 		return t.executeWorkflowBackoffTimerTask(ctx, timerTask)
 	case persistence.TaskTypeDeleteHistoryEvent:
-		return t.executeDeleteHistoryEventTask(ctx, timerTask)
+		// special timeout for delete history event
+		deleteHistoryEventContext, deleteHistoryEventCancel := context.WithTimeout(context.Background(), taskDeleteHistoryEventContextTimeout)
```
taylanisikdemir, line 114: I know the existing `ctx` has 3 sec timeout and doesn't listen for queue shut down signals. 
With the new 60s timeout for history deletions it is going to be more likely to end up hanging a shutdown. We should consider making these operations responsive to shut down signals.
@Shaddoll what do you think?

taylanisikdemir: Discussed offline. We will hook up Stop signals to task executor in a follow up PR. 
In this PR, let's make this 60s dynamically configurable

davidporter-id-au (approved):

common/dynamicconfig/constants.go
```diff
@@ -1440,6 +1440,13 @@ const (
 	IsolationGroupStateUpdateRetryAttempts
 
 	LargeShardHistoryBlobMetricThreshold
+
+	// DeleteHistoryEventContextTimeout
```
taylanisikdemir, line 1444: let's mention the unit of time in comment or variable name.

taylanisikdemir (approved):

common/dynamicconfig/constants.go
timl3136: Thanks, I added that in L3803 but forgot to add it here.

PR #5892, opened 2024-04-08
Addition of test for TestHandler_UpdateAsyncWorkflowConfiguration

abhishekj720:
**What changed?**
Test for UpdateAsyncWorkflowConfiguraton

**Why?**
As part of code coverage week

**How did you test it?**
Unit test

**Potential risks**
No risk


common/domain/handler_test.go
```diff
@@ -939,3 +939,350 @@ func TestHandler_UpdateIsolationGroups(t *testing.T) {
...
+
+	tests := []struct {
+		name             string
+		setupMocks       func(mockDomainManager *persistence.MockDomainManager, mockReplicator *MockReplicator)
+		isPrimaryCluster bool
+		request          *types.UpdateDomainAsyncWorkflowConfiguratonRequest
+		expectedErr      error
+	}{
+		{
+			name: "fail to get metadata",
```
davidporter-id-au, line 955: not a real problem, but just a tip: Genuinely starting with the success case is a bit easier because you typically have to construct the *most* mocks and return results. And thereafter all the error cases are typically just subsets of the first entry. It just makes copying + pasting easier

common/domain/handler_test.go
```diff
@@ -939,3 +939,350 @@ func TestHandler_UpdateIsolationGroups(t *testing.T) {
...
+						ActiveClusterName: "active",
+						Clusters:          []*persistence.ClusterReplicationConfig{{ClusterName: "active"}, {ClusterName: "standby"}},
+					},
+					ConfigVersion:   1,
+					FailoverVersion: 123,
+					IsGlobalDomain:  true,
+					LastUpdatedTime: time.Now().Add(-24 * time.Hour).UnixNano(),
+				}, nil)
+
+				mockDomainManager.EXPECT().UpdateDomain(gomock.Any(), gomock.Any()).Return(nil)
```
davidporter-id-au, line 1197: Sorry to be annoying, but if the test is to really have any serious value at all the updateDomain request should be specified here. Just doing gomock.Any() doesn't really tell you much if anything's broken

agautam478 (approved):

PR #5896, opened 2024-04-09
Unit tests for AppendHistoryNodes

jakobht:
**What changed?**
Added unit tests for AppendHistoryNodes 

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosql_history_store_test.go
+++ b/common/persistence/nosql/nosql_history_store_test.go
@@ -1,0 +1,173 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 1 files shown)


common/persistence/nosql/nosql_history_store_test.go
```diff
@@ -0,0 +1,172 @@
...
+	mockPlugin.EXPECT().CreateDB(gomock.Any(), gomock.Any(), gomock.Any()).Return(mockDB, nil).AnyTimes()
+	RegisterPlugin("cassandra", mockPlugin)
+}
+
+func TestNewNoSQLHistoryStore(t *testing.T) {
+	registerCassandraMock(t)
+	cfg := getValidShardedNoSQLConfig()
+
+	_, err := newNoSQLHistoryStore(cfg, log.NewNoop(), nil)
+	assert.NoError(t, err)
```
dkrotx, line 98: nit: I would also check for not nil result-pointer. just in case

common/persistence/nosql/nosql_history_store_test.go
```diff
@@ -0,0 +1,172 @@
...
+	store, dbMock := setUpMocks(t)
+
+	// Expect to insert the node into the history tree and node, as this is not a new branch, expect treeRow to be nil
+	dbMock.EXPECT().InsertIntoHistoryTreeAndNode(gomock.Any(), nil, validHistoryNodeRow()).Return(nil).Times(1)
+
+	request := validInternalAppendHistoryNodesRequest()
+	err := store.AppendHistoryNodes(ctx.Background(), request)
+
+	assert.NoError(t, err)
+}
```
dkrotx, line 147: Is there something else we could check in `store`? I wonder if we can check the amount of nodes is now == 1

common/persistence/nosql/nosql_history_store_test.go
```diff
@@ -0,0 +1,172 @@
...
+			assert.Equal(t, request.Info, treeRow.Info)
+
+			assert.WithinDuration(t, time.Now(), treeRow.CreateTimestamp, time.Second)
+			return nil
+		})
+
+	err := store.AppendHistoryNodes(ctx.Background(), request)
+
+	assert.NoError(t, err)
+}
```
dkrotx, line 172: same as above

common/persistence/nosql/nosql_history_store_test.go
jakobht: No, the main check is that we send the correct thing to the database. The historyStore struct does not store anything itself.

common/persistence/nosql/nosql_history_store_test.go
jakobht: No again :)

dkrotx (approved):

PR #5902, opened 2024-04-11
Add memo in pinot

bowenxia:
**What changed?**
Add memo in pinot:
When write, add memo into search attribute
When read, get memo from search attribute and cleanup that field

NOTE: don't merge it until our customer test it in staging

**Why?**
memo was stored in ES but can't be indexed(not in ES schema). We made the pinot schemas according to ES's schema, so we thought that we did't need memo in Pinot, but one customer was complaining about that. 

**How did you test it?**
unit test

**Potential risks**
not risk since Memo is not a functional attribute. 

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/data_store_interfaces.go
+++ b/common/persistence/data_store_interfaces.go
@@ -23,6 +23,7 @@ package persistence
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -957,6 +958,10 @@ func (d *DataBlob) GetEncodingString() string {
 		return ""
 	}
 	return string(d.Encoding)
+}
+
+func (d *DataBlob) GetVisibilityStoreInfo() ([]byte, error) {
+	return json.Marshal(d)
 }
 
 // GetData is a safe way to get the byte array or nil
--- a/common/persistence/pinot/pinot_visibility_store.go
+++ b/common/persistence/pinot/pinot_visibility_store.go
@@ -61,6 +61,7 @@ const (
 	ExecutionTime        = "ExecutionTime"
 	IsDeleted            = "IsDeleted"   // used for Pinot deletion/rolling upsert only, not visible to user
 	EventTimeMs          = "EventTimeMs" // used for Pinot deletion/rolling upsert only, not visible to user
+	Memo                 = "Memo"
 
 	// used to be micro second
 	oneMicroSecondInNano = int64(time.Microsecond / time.Nanosecond)
@@ -119,8 +120,6 @@ func (v *pinotVisibilityStore) RecordWorkflowExecutionStarted(
 		request.StartTimestamp.UnixMilli(),
 		request.ExecutionTimestamp.UnixMilli(),
 		request.TaskID,
-		request.Memo.Data,
-		request.Memo.GetEncoding(),
 		request.IsCron,
 		request.NumClusters,
 		-1, // represent invalid close time, means open workflow execution
@@ -130,6 +129,7 @@ func (v *pinotVisibilityStore) RecordWorkflowExecutionStarted(
 		int64(request.ShardID),
 		request.SearchAttributes,
 		false,
+		request.Memo,
 	)
 
 	if err != nil {
@@ -150,8 +150,6 @@ func (v *pinotVisibilityStore) RecordWorkflowExecutionClosed(ctx context.Context
 		request.StartTimestamp.UnixMilli(),
 		request.ExecutionTimestamp.UnixMilli(),
 		request.TaskID,
-		request.Memo.Data,
-		request.Memo.GetEncoding(),
 		request.IsCron,
 		request.NumClusters,
 		request.CloseTimestamp.UnixMilli(),
@@ -161,6 +159,7 @@ func (v *pinotVisibilityStore) RecordWorkflowExecutionClosed(ctx context.Context
 		int64(request.ShardID),
 		request.SearchAttributes,
 		false,
+		request.Memo,
 	)
 
 	if err != nil {
@@ -181,8 +180,6 @@ func (v *pinotVisibilityStore) RecordWorkflowExecutionUninitialized(ctx context.
 		-1,
 		-1,
 		0,
-		nil,
-		"",
 		false,
 		0,
 		-1, // represent invalid close time, means open workflow execution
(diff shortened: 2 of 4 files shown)


common/pinot/response_utility.go
```diff
@@ -83,12 +83,23 @@ func ConvertSearchResultToVisibilityRecord(hit []interface{}, columnNames []stri
 	if err != nil {
 		logger.Error("Unable to Unmarshal searchAttribute map", tag.Error(err))
 	}
+	var memo *p.DataBlob
+	if attributeMap["Memo"] != nil {
```
davidporter-id-au, line 87: Nit: do you mind making this key a const somewhere? Perhaps on this package as a publically accessible string?

common/pinot/response_utility.go
```diff
@@ -83,12 +83,23 @@ func ConvertSearchResultToVisibilityRecord(hit []interface{}, columnNames []stri
...
+		err = json.Unmarshal([]byte(fmt.Sprintf("%s", attributeMap["Memo"])), &memo)
```
davidporter-id-au, line 88: no need for sprintf, just 

```suggestion
		err = json.Unmarshal([]byte(attributeMap[MemoKey]), &memo)
```

common/pinot/response_utility_test.go
```diff
@@ -36,6 +38,9 @@ import (
 func TestConvertSearchResultToVisibilityRecord(t *testing.T) {
 	columnName := []string{"WorkflowID", "RunID", "WorkflowType", "DomainID", "StartTime", "ExecutionTime", "CloseTime", "CloseStatus", "HistoryLength", "TaskList", "IsCron", "NumClusters", "UpdateTime", "Attr"}
 	closeStatus := types.WorkflowExecutionCloseStatusFailed
+	testMemo := p.NewDataBlob([]byte{}, p.VisibilityEncoding)
```
davidporter-id-au, line 41: maybe use the method?

common/pinot/response_utility_test.go
```diff
@@ -36,6 +38,9 @@ import (
...
+	marshalMemo, err := json.Marshal(testMemo)
```
davidporter-id-au, line 42: maybe use the method. Consider (the unlikely) case of the encoding change

common/persistence/pinot/pinot_visibility_store.go
```diff
@@ -614,6 +610,14 @@ func createVisibilityMessage(
 		}
 		SearchAttributes[key] = val
 	}
+
+	// add memo into search attr
+	searchAttributesMemo, err := memo.GetVisibilityStoreInfo()
+	if err != nil {
+		return nil, err
+	}
+	SearchAttributes[Memo] = searchAttributesMemo
```
neil-xie, line 619: I think we only need to set Memo when memo is used

common/persistence/pinot/pinot_visibility_store.go
```diff
@@ -61,6 +61,8 @@ const (
 	ExecutionTime        = "ExecutionTime"
 	IsDeleted            = "IsDeleted"   // used for Pinot deletion/rolling upsert only, not visible to user
 	EventTimeMs          = "EventTimeMs" // used for Pinot deletion/rolling upsert only, not visible to user
+	Memo                 = "Memo"
+	MemoEncoding         = "Memo_Encoding"
```
neil-xie, line 65: This will be a new search attribute, we need to update flipr for all clusters

common/persistence/data_store_interfaces.go
```diff
@@ -969,6 +970,15 @@ func (d *DataBlob) GetEncodingString() string {
 	return string(d.Encoding)
 }
 
+func (d *DataBlob) GetVisibilityStoreInfo() ([]byte, common.EncodingType, error) {
+	data, err := json.Marshal(d)
+	if err != nil {
+		return nil, common.EncodingTypeEmpty, err
+	}
+
+	return data, common.EncodingTypeJSON, nil // for visibility, we use Json encode to keep it safe
```
neil-xie, line 979: Are we always using encoding type json? From this method seems so. If it is true, we don't need to save it in search attribute

bowenxia: Yes. David told me to add this because we are wrapping memo was another layer. And this encoding is for future reference for other people.

common/pinot/response_utility.go
```diff
@@ -64,33 +67,44 @@ type VisibilityRecord struct {
 	ShardID       int16
 }
 
-func ConvertSearchResultToVisibilityRecord(hit []interface{}, columnNames []string, logger log.Logger) *p.InternalVisibilityWorkflowExecutionInfo {
+func ConvertSearchResultToVisibilityRecord(hit []interface{}, columnNames []string) (*p.InternalVisibilityWorkflowExecutionInfo, error) {
 	if len(hit) != len(columnNames) {
-		return nil
+		return nil, fmt.Errorf("length of hit (%v) is not equal with length of columnNames(%v)", len(hit), len(columnNames))
```
neil-xie, line 72: What if we have upgraded the pinot schema but haven't deployed code changes, will it fail here?

neil-xie: Probably not in this diff, but should have a version to guard here, also backward compatible cc: @davidporter-id-au

bowenxia: Ideally those will always be equal since both things are from the same piece pinot response. If those are not equal, then it should be an error from the pinot side not us.

neil-xie: Oh, this schema also comes from the Pinot response, not reading from config, correct?

bowenxia: correct

common/pinot/response_utility.go
```diff
@@ -64,33 +67,44 @@ type VisibilityRecord struct {
...
 
 	attributeMap := make(map[string]interface{})
-	err = json.Unmarshal([]byte(fmt.Sprintf("%s", systemKeyMap["Attr"])), &attributeMap)
-	if err != nil {
-		logger.Error("Unable to Unmarshal searchAttribute map", tag.Error(err))
+	var memo *p.DataBlob
+	if systemKeyMap[Attr] != nil {
+		attrMapStr, ok := systemKeyMap[Attr].(string)
+		if !ok {
+			return nil, fmt.Errorf(`assertion error. Can't convert systemKeyMap["Attr"] to string. Found %T`, systemKeyMap["Attr"])
```
davidporter-id-au, line 87: nit: `systemKeyMap[Attr]`

davidporter-id-au (approved):

PR #5904, opened 2024-04-11
Add missing mappers for AsyncWorkflowConfiguration

taylanisikdemir:
**What changed?**
There are many places in the codebase that maps one type to another. While adding `AsyncWorkflowConfiguration` field to domain config a few places was missed. Updating them along with corresponding test data.

**How did you test it?**
unit test

### DIFF
--- a/common/types/testdata/domain.go
+++ b/common/types/testdata/domain.go
@@ -88,6 +88,14 @@ var (
 				State: types.IsolationGroupStateDrained,
 			},
 		},
+		AsyncWorkflowConfig: &types.AsyncWorkflowConfiguration{
+			Enabled:   true,
+			QueueType: "custom",
+			QueueConfig: &types.DataBlob{
+				EncodingType: types.EncodingTypeThriftRW.Ptr(),
+				Data:         []byte("custom queue config"),
+			},
+		},
 	}
 	DomainReplicationConfiguration = types.DomainReplicationConfiguration{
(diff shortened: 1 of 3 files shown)


Shaddoll (approved):

common/types/testdata/domain.go
```diff
@@ -88,6 +88,14 @@ var (
 				State: types.IsolationGroupStateDrained,
 			},
 		},
+		AsyncWorkflowConfig: &types.AsyncWorkflowConfiguration{
```
Groxx, line 91: hmm.  so there wasn't previously data, and adding it + mapping it -> no tests need to change?
that seems to imply the tests are seeing it _not_ propagate correctly, as they had to have empty values before, yea?

or do we _only_ have round-trip tests, not `B==map(A)` checks with pre-calculated values?

Groxx:
small test-surprise question, but I don't see how these changes would make anything _less_ correct, so 👍

common/types/testdata/domain.go
taylanisikdemir: testdata had to be updated to make the round trip tests pass.

Groxx: double checked, and yea this is fine:
- with this, tests fail before the fix because the value was not retained (yay better coverage)
- both with and without this, tests succeed after the fix, because the value is retained

Groxx (approved):

PR #5907, opened 2024-04-11
Switch async workflow request encoding from json to thrift

taylanisikdemir:
**What changed?**
While testing async workflows in dev environment I noticed the workflow input bytes are not carried over to the final sync call. It's because request object is json marshaled by async api handlers before sent to the queue. Certain fields in our request payloads such as [this one](https://github.com/uber/cadence/blob/5d3c740e4c37aefe5e6420a36eb12eb8104a9ed5/common/types/shared.go#L6277) are excluded in json output because they may contain PII and can get logged. 

Request objects implement `FilteredRequestBody` interface which exposes `SerializeForLogging()` function to  authorizers ([example](https://github.com/uber/cadence/blob/a614c4c1fd1cf94b3722193cec919985304c9acb/service/frontend/wrappers/accesscontrolled/api_generated.go#L542)) and authorizer implementation can choose to log response of `SerializeForLogging()`. This is used in uber internally for audit purposes. 

To avoid breaking that existing "PII hiding from logs of jsonified requests" flow, I am switching encoding of async workflow requests to thrift. 

Also updating the cherry pick logic of headers by carrying over all original headers as is to avoid follow up work to support context/trace propagation work for async APIs.

**Why?**
Carry over request and headers of async call to sync call as is.

**How did you test it?**
Unit tests


Groxx:
Ah.  Yes, we should not be relying on json except for explicit opt-in types for that exact purpose :|  It's WILDLY error-prone for reasons like this.

That said, why not use thrift or proto?  We already have these types, and they have well-understood behavior and known versioning/etc behavior.

Shaddoll (approved):

common/asyncworkflow/queue/consumer/default_consumer.go
```diff
@@ -270,26 +270,30 @@ func getYARPCOptions(header *shared.Header) []yarpc.CallOption {
...
-	if encoding != string(common.EncodingTypeJSON) {
+func (c *DefaultConsumer) decodeStartWorkflowRequest(payload []byte, encoding string) (*types.StartWorkflowExecutionRequest, error) {
+	if encoding != string(common.EncodingTypeThriftRW) {
 		return nil, &UnsupportedEncoding{EncodingType: encoding}
 	}
 
-	var startRequest types.StartWorkflowExecutionAsyncRequest
-	if err := json.Unmarshal(payload, &startRequest); err != nil {
+	var thriftObj shared.StartWorkflowExecutionAsyncRequest
+	if err := c.msgDecoder.Decode(payload, &thriftObj); err != nil {
```
Groxx, line 279: why pass encoding as a string if the decoder doesn't even look at it?
seems like this should be either "ask the decoder what the type is" or "just always do thrift, no need to accept it as an argument"

common/codec/version0_thriftrw_test.go
```diff
@@ -84,6 +88,35 @@ func (s *thriftRWEncoderSuite) TestEncode() {
...
+			if diff := cmp.Diff(thriftEncodedBinary, binary); diff != "" {
+				mu.Lock()
+				finalErr = multierr.Append(finalErr, fmt.Errorf("Mismatch (-want +got):\n%s", diff))
+				mu.Unlock()
+				return
+			}
+		}()
+	}
+	wg.Wait()
+	s.NoError(finalErr)
```
Groxx, line 117:
```suggestion
	var mu sync.Mutex
	count := 200
	errs := make([]error, count)
	wg.Add(count)
	for i := 0; i < count; i++ {
		i := i // until loop-var experiment stabilizes
		go func() {
			defer wg.Done()
			binary, err := s.encoder.Encode(thriftObject)
			if err != nil {
				errs[i] = err
				return
			}

			if diff := cmp.Diff(thriftEncodedBinary, binary); diff != "" {
				errs[i] = fmt.Errorf("Mismatch (-want +got):\n%s", diff))
				return
			}
		}()
	}
	wg.Wait()
	s.NoError(multierr.Append(errs...))
```
^ possibly a simpler pattern for future stuff

taylanisikdemir: correct one is `multierr.Combine(errs...)`

Groxx: yea, I'm not usually excited about nil-coalescing APIs.  there's no real naming pattern that identifies them :\

this one I've just seen a fair number of places because it's handy for reasons like this, where errors are just collected and not individually fatal.

PR #5909, opened 2024-04-15
[code-coverage] Add more tests for service/history/decision package 

ketsiambaku:
**What changed?**
- Generated mock for service/history/query/registry interface
- Cover additional scenarios in HandleBufferedQueries API
- Added tests for CreateRecordDecisionTaskStartedResponse API
- Added more tests for HandleDecisionTaskStarted API

**Why?**
improve code coverage in service/history/decision

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/decision/handler_test.go
```diff
@@ -462,11 +472,43 @@ func (s *DecisionHandlerSuite) TestHandleDecisionTaskStarted() {
...
+				engine.EXPECT().NotifyNewReplicationTasks(gomock.Any())
+			},
+			expectErr: nil,
+			mutablestate: &persistence.WorkflowMutableState{
+				ExecutionInfo: &persistence.WorkflowExecutionInfo{
+					DecisionStartedID: -23,
+					NextEventID:       2,
+				},
+			},
+			assertCalls: func(resp *types.RecordDecisionTaskStartedResponse) {},
```
shijiesheng, line 511: Should we assert the response body like the previous success?

shijiesheng (approved):

service/history/decision/handler_test.go
```diff
@@ -547,6 +653,84 @@ func (s *DecisionHandlerSuite) TestHandleBufferedQueries_QueryTooLarge() {
...
+			queryResults:         map[string]*types.WorkflowQueryResult{"some-id": &types.WorkflowQueryResult{}},
+			assertCalls: func(logs *observer.ObservedLogs) {
+				s.Equal(1, logs.FilterMessage("failed to set query termination state to completed").Len())
+				s.Equal(1, logs.FilterMessage("failed to set query termination state to unblocked").Len())
+			},
+		},
+	}
+	for _, test := range tests {
+		s.Run(test.name, func() {
+			core, observedLogs := observer.New(zap.ErrorLevel)
```
davidporter-id-au, line 723: oh nice! I didn't know about this

service/history/decision/handler_test.go
```diff
@@ -547,6 +653,84 @@ func (s *DecisionHandlerSuite) TestHandleBufferedQueries_QueryTooLarge() {
...
+			clientFeatureVersion: client.GoWorkerConsistentQueryVersion,
+			queryResults:         map[string]*types.WorkflowQueryResult{"some-id": &types.WorkflowQueryResult{}},
+			assertCalls: func(logs *observer.ObservedLogs) {
+				s.Equal(1, logs.FilterMessage("failed to set query termination state to completed").Len())
+				s.Equal(1, logs.FilterMessage("failed to set query termination state to unblocked").Len())
+			},
+		},
+	}
+	for _, test := range tests {
+		s.Run(test.name, func() {
```
davidporter-id-au, line 722: you're kinda using a table test and a suite test at the same time which - normally I'd say is probably a bit confusing, but I take it that there's existing tests in here that you're re-using. One thing that *might* cause problems is that the `s.` may maintain state between tests, (like on line 725) so I'd suggest if possible moving to just use plain `t.Run()` and creating a new decisionHandler every time in future.

davidporter-id-au (approved):

service/history/decision/handler_test.go
ketsiambaku: changed in #5932

service/history/decision/handler_test.go
ketsiambaku: Yes, thank you! Added in #5932

PR #5917, opened 2024-04-17
Get rid of mutex in matching/liveness and reduce test duration

taylanisikdemir:
**What changed?**
- Replaced mutex with atomic
- Updated `Stop()` to wait for background goroutine completion
- Added a new log to distinguish between regular tasklist stop vs liveness originated tasklist stop
- Removed time argument from `markAlive` because only `time.Now` is passed
- Reduced test duration from 8 sec to 2s

**Why?**
Just applying some best practices to improve things step by step.

**How did you test it?**
```
go test -timeout 60s -run ^TestLivenessSuite$ github.com/uber/cadence/service/matching -count=100 -v -race
```

### DIFF
--- a/service/matching/liveness.go
+++ b/service/matching/liveness.go
@@ -102,7 +102,7 @@ func (l *liveness) eventLoop() {
 func (l *liveness) isAlive() bool {
 	now := l.timeSource.Now().UnixNano()
 	lastUpdate := atomic.LoadInt64(&l.lastEventTimeNano)
-	return now-lastUpdate < int64(l.ttl)
+	return now-lastUpdate < l.ttl.Nanoseconds()
 }
 
 func (l *liveness) markAlive() {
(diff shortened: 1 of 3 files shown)


service/matching/liveness.go
```diff
@@ -36,90 +36,76 @@ type (
...
-		common.DaemonStatusStopped,
-	) {
+	if !atomic.CompareAndSwapInt32(&l.status, common.DaemonStatusStarted, common.DaemonStatusStopped) {
 		return
 	}
 
-	close(l.shutdownChan)
+	close(l.stopCh)
 	l.broadcastShutdownFn()
+	l.wg.Wait()
```
davidporter-id-au, line 81: I guess this will help make this block upstream callers know when this is done?

davidporter-id-au (approved):

davidporter-id-au (approved):

service/matching/liveness.go
taylanisikdemir: Desirable pattern is to finish all cleanup and termination of goroutines before returning from `Stop()`. Also stop child components. If each component does this we will be in very good shape.

PR #5919, opened 2024-04-18
Fix failover error causing child workflows to get stuck

davidporter-id-au:
**What changed?**

This hopefully fixes some child workflows erroring-out during a failover. I was doing some testing and got reports of stuck child workflows. The child workflows, as far as I can tell, were being created at high throughput during a failover, and the failover caused a race between presumably the various history shards in having up-to-date information about the state of the domain. 

Since a domain update is a fairly async process, and there's no clear guantee that a transfer task to create a child workflow will not get there before the other history node has updated it's domain cache, it seems likely to me that we shouldn't have ever treated this as a non-retriable error (I wrote this when I first join the team, but it was probably a bad idea). 

My guess is that we probably will back out most of  [these](https://github.com/uber/cadence/pull/4599/files#diff-4dda032d386904720fb9ac56ba521d4fd469f66c66884f0b717df2c3ff3a2c91) changes to handle domain-not-active errors for cross-domain code, since the feature-itself is likely to be deprecated. For now though, I think just guarding it more safely should avoid the problem (hopefully). 

#### Reasoning: 
Screenshot 2024-04-17 at 8 35 03 PM

I did a failover *to* this region at precisely the second that this error is omitted: 

Screenshot 2024-04-17 at 7 49 21 PM


service/history/task/task.go
```diff
@@ -288,20 +288,14 @@ func (t *taskImpl) HandleErr(err error) (retErr error) {
 		err = nil
 	}
 
-	// target domain not active error, we should retry the task
-	// so that a cross-cluster task can be created.
-	if err == errTargetDomainNotActive {
-		t.scope.IncCounter(metrics.TaskTargetNotActiveCounterPerDomain)
-		t.logger.Error("Dropping 'domain-not-active' error as non-retriable", tag.Error(err))
-		return nil
```
davidporter-id-au, line 296: this is the problem I think, it's just dropping a task on the floor which is a problem in a race

Shaddoll: if the domain is failed over to the other region, the tasks are not able to start child workflow in this region because this region is active. I think even if you retry the task, you'll still get the same error

taylanisikdemir: that makes sense @Shaddoll. in that case should we also remove the time elapsed check all together?

davidporter-id-au: The logs show that the error was emitted from the *newly* active region, in which case I expect it to converge towards being correct on retry. 

(for context in the failover, I was failing up-cm-dca from PHX *back* to its original location in DCA at 12:43PST, the errors were emitted from *dca*)

Shaddoll: synced offline and now I understand that there might be a race condition with failover and domain cache update in different history hosts.
Approved, since the risk of the change is very low.

Shaddoll (approved):

service/history/task/task.go
taylanisikdemir: Got it. this is to avoid dropping the task on the newly active side 👍

taylanisikdemir (approved):

PR #5920, opened 2024-04-18
Added Executor Interface and TimerTaskExecutorBase with stop() Method and improve context management in TimerQueueProcessor

timl3136:
**What changed?**
Added stop() to the Executor interface and created an empty stop() function for future implementations in Executors.
Implemented stop() in TimerTaskExecutorBase.
Added context and cancelFn in TimerTaskExecutorBase struct and replaced context.Background() with internal context to prevent memory leaks.
Called executor.stop() in TimerQueueProcessor to prevent context leaks.

**Why?**
Currently we are creating new context and pass that to downstream without proper stop signal propagation

**How did you test it?**
Unit tests

**Potential risks**
We are calling the stop method at the end of the lifecycle so there won't be any added risks.

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -196,7 +196,7 @@ func (t *timerQueueProcessor) Stop() {
 		t.logger.Warn("transferQueueProcessor timed out on shut down", tag.LifeCycleStopTimedout)
 	}
 	t.activeQueueProcessor.Stop()
-	t.activeQueueProcessor.Stop()
+	t.activeTaskExecutor.Stop()
 	for _, standbyQueueProcessor := range t.standbyQueueProcessors {
 		standbyQueueProcessor.Stop()
 	}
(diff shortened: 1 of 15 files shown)


service/history/task/timer_active_task_executor.go
```diff
@@ -87,7 +87,7 @@ func (t *timerActiveTaskExecutor) Execute(
 		return nil
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), taskDefaultTimeout)
+	ctx, cancel := context.WithTimeout(t.ctx, taskDefaultTimeout)
```
taylanisikdemir, line 90: this line is unnecessary for `TaskTypeDeleteHistoryEvent` case. so maybe create the context within each case below with corresponding timeout value.

service/history/queue/timer_queue_processor.go
```diff
@@ -178,6 +178,8 @@ func (t *timerQueueProcessor) Stop() {
 
 	if !t.shard.GetConfig().QueueProcessorEnableGracefulSyncShutdown() {
 		t.activeQueueProcessor.Stop()
+		// stop active executor after queue processor
+		t.activeTaskExecutor.Stop()
```
taylanisikdemir, line 182: There are also `standbyTaskExecutor` objects created and passed to `newTimerQueueStandbyProcessor` in the constructor. We need to stop them as well.  Timer queue processor is creating those so it should be responsible of stopping the executors

timl3136: Thanks, I have added a field accounting for the standbyTaskExecutors and stop then in the end.

PR #5926, opened 2024-04-22
Add peer provider plugin registration

mantas-sidlauskas:
**What changed?**
Adding peer provider plugin registration code.

**Why?**
This code is going to be used to load external peer provider plugin. Peer provider plugin will provide an easy way to replace ringpop with another technology.

**How did you test it?**

**Potential risks**
None, this code is noop right now. 

Related https://github.com/uber/cadence/issues/5593

### DIFF
--- a/common/peerprovider/plugin.go
+++ b/common/peerprovider/plugin.go
@@ -1,0 +1,87 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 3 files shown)


common/peerprovider/plugin.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if !inserted {
+		return fmt.Errorf("peer provider %q is already registered", configKey)
+	}
+
+	return nil
+}
+
+func (p *Provider) Provider() (membership.PeerProvider, error) {
+	for configKey, cfg := range p.config {
```
taylanisikdemir, line 80: do you need multiple provider support? If so the config should have a place to define these list of providers and another place to pick which one is active. Returning the first registered one might cause unintended provider to be picked

mantas-sidlauskas: adding more thoughts here: at least in theory, you might use multiple peer providers at the same time, for migration needs. Will revisit this method to check what it takes to support this scenario.

mantas-sidlauskas: Multiple provider support will increase complexity in membership package highly, what we can do instead, limit provider number to 1, if anyone needs multi-source peer provider solution, you can replace plugin with  multi-source plugin. This would keep internal membership logic unchanged and there will be no requirement to consult multiple peer providers.

taylanisikdemir (approved):

PR #5928, opened 2024-04-23
Workaround for query-consistency-strong which is presently partially broken

davidporter-id-au:
**What changed?**
As per comment discription, this is a workaround for an apparently outstanding bug where it looks like headers are not being passed along on all query paths. The problem remains, but to prevent a customer requiring this feature for now this removes the check on this one specific part of flow. 

Obviously the correct and more ideal fix is to fix the missing headers, however, that'll take a bit of time and time-being what it is I'm trying to unblock Strong queries for this user.

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/decision/handler.go
```diff
@@ -725,7 +725,15 @@ func (handler *handlerImpl) handleBufferedQueries(
...
-	if versionErr := handler.versionChecker.SupportsConsistentQuery(clientImpl, clientFeatureVersion); versionErr != nil {
+	versionErr := handler.versionChecker.SupportsConsistentQuery(clientImpl, clientFeatureVersion)
+	// todo (David.Porter) remove the skip on version check for
+	// clientImpl and clientFeatureVersion where they're nil
+	// There's a bug, probably in matching somewhere which isn't
+	// forwarding the client headers for version
+	// info correctly making this call erroneously fail sometimes.
+	// https://t3.uberinternal.com/browse/CDNC-8641
+	// So defaulting just this flow to fail-open in the absence of headers.
+	if versionErr != nil && clientImpl != "" && clientFeatureVersion != "" {
```
ketsiambaku, line 736: Hmm sure, this is fine for now

ketsiambaku (approved):

service/history/decision/handler.go
taylanisikdemir: there are code paths within vesionChecker that handles `clientImpl == nil` gracefully. so expecting both `clientImpl` and `clientFeatureVersion` to be empty might exclude some genuine cases.

davidporter-id-au: discussed offline, I believe we concluded that the clients that didn't support this and were using QueryConsistencyStrong to be a small set.

PR #5929, opened 2024-04-23
Changed the error to DomainNotActive for Deprecated domains

abhishekj720:
**What changed?**
As part of oncall fix causing muttley service side error, we think the code is breaking because of misclassified error as it see sprintf error as cadence causing issue, but in real it is not.

I think we have to return errors from below for poll for DecisionTask.

Pasted Graphic

**Why?**
To fix the misclassified error.

**How did you test it?**
Uni test

**Potential risks**
Added change to avoid muttley service SLA mess

**Release notes**

**Documentation Changes**


davidporter-id-au:
this isn't for crashes, it's for a misclassified error making a bit of a mess of the SLA

davidporter-id-au (approved):

davidporter-id-au:
Before merging do you mind correcting the description: Nothing's crashing here

PR #5935, opened 2024-04-23
Fix bug when pass close status as an integar string

neil-xie:
**What changed?**
Currently it will directly pass the string to Pinot which can cause query failure

**Why?**
Bug fix

**How did you test it?**
Unit test

**Potential risks**

**Release notes**

**Documentation Changes**


ibarrajo (approved):
LGTM :shipit:

PR #5937, opened 2024-04-24
Fix slice reuse in cassandra/domain.go

natemort:
**What changed?**
Reset the fields for isolationGroups and asyncWFConfigData after each row.

While reading domains from Cassandra the isolationGroups and asncWFConfigData are not properly reset for each row. This can result in the same backing array being reused across multiple rows, resulting in an incorrect or corrupted value being read.

**Why?**
Fixes a bug where the domain cache contains the wrong async wf config data and is unable to trigger an async workflow despite it being enabled for that ddomain.

**How did you test it?**
- Updated the persistence tests to include isolationGroups and asyncWFConfigData.
- Added a secondary domain to the async wf integration test
- Ran locally and started an async workflow via the java client

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/persistence/nosql/nosqlplugin/cassandra/domain.go
+++ b/common/persistence/nosql/nosqlplugin/cassandra/domain.go
@@ -410,6 +410,10 @@ func (db *cdb) SelectAllDomains(
 		replicationClusters = []map[string]interface{}{}
 		badBinariesData = []byte("")
 		badBinariesDataEncoding = ""
+		isolationGroups = []byte("")
+		isolationGroupsEncoding = ""
+		asyncWFConfigData = []byte("")
+		asyncWFConfigEncoding = ""
 		failoverEndTime = 0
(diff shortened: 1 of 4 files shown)


common/persistence/nosql/nosqlplugin/cassandra/domain.go
```diff
@@ -410,6 +410,10 @@ func (db *cdb) SelectAllDomains(
 		replicationClusters = []map[string]interface{}{}
 		badBinariesData = []byte("")
 		badBinariesDataEncoding = ""
+		isolationGroups = []byte("")
+		isolationGroupsEncoding = ""
+		asyncWFConfigData = []byte("")
```
taylanisikdemir, line 415: `asyncWFConfigData = nil` would also do the job. not sure why the byte arrays were reset like this before.

taylanisikdemir (approved):

PR #5938, opened 2024-04-25
We now wait 10 seconds before we start returning shard closed errors, also stop retrying on shard closed errors 

jakobht:
**What changed?**

- Introduced a shardRecentlyClosed error to signal from the shard that it was recently closed. This error will not cause the task handler to emit error logs and metrics. 
- Add a test that hits this guard in all the places it exists
- ~~Deleted unreachable instances of the guard. These were introduced in https://github.com/uber/cadence/pull/4547 when the checks were in a loop that was accessing the DB, which was able to change the shard's closed state in each iteration~~
- Removed redundant tests that checked the same as the new guard test now checks  
- Stop retrying on shard closed errors 

**Why?**
Shard closing is _not_ an unexpected state, so we should not emit error logs and metrics for this.

If we stay in a state where a closed shard keeps getting requests, then we should start emitting error logs and metrics, so we wait 10 seconds, and if we still see requests then we start emitting the metrics.

All the guard testing and deleting is necessary to make the new line coverage check happy. 

**How did you test it?**
Tested with unit tests and by deploying to staging. The deployment shows we can now do restarts without seeing these errors.

**Potential risks**
This does change some relatively core task processing logic, however the main change is on how the error states are communicated. The main flow is not touched. 

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -173,24 +173,29 @@ type (
 
 var _ Context = (*contextImpl)(nil)
 
-var (
-	// ErrShardClosed is returned when shard is closed and a req cannot be processed
-	ErrShardClosed = errors.New("shard closed")
+type ErrShardClosed struct {
+	Msg      string
+	ClosedAt time.Time
+}
 
-	// ErrShardRecentlyClosed is returned when a shard has recently been closed, this
-	// error is transient, and should not cause error logs and error metrics to be emitted
-	ErrShardRecentlyClosed = errors.New("shard recently closed")
+var _ error = (*ErrShardClosed)(nil)
+
+func (e *ErrShardClosed) Error() string {
+	return e.Msg
+}
+
+const (
+	TimeBeforeShardClosedIsError = 10 * time.Second
 )
 
 const (
 	// transfer/cross cluster diff/lag is in terms of taskID, which is calculated based on shard rangeID
 	// on shard movement, taskID will increase by around 1 million
-	logWarnTransferLevelDiff     = 3000000 // 3 million
-	logWarnCrossClusterLevelLag  = 3000000 // 3 million
-	logWarnTimerLevelDiff        = time.Duration(30 * time.Minute)
-	historySizeLogThreshold      = 10 * 1024 * 1024
-	minContextTimeout            = 1 * time.Second
-	timeBeforeShardClosedIsError = 10 * time.Second
+	logWarnTransferLevelDiff    = 3000000 // 3 million
+	logWarnCrossClusterLevelLag = 3000000 // 3 million
+	logWarnTimerLevelDiff       = time.Duration(30 * time.Minute)
+	historySizeLogThreshold     = 10 * 1024 * 1024
+	minContextTimeout           = 1 * time.Second
 )
 
 func (s *contextImpl) GetShardID() int {
@@ -1007,16 +1012,14 @@ func (s *contextImpl) getRangeID() int64 {
 
 func (s *contextImpl) closedError() error {
 	closedAt := s.closedAt.Load()
-
 	if closedAt == nil {
 		return nil
 	}
 
-	if time.Since(*closedAt) < timeBeforeShardClosedIsError {
-		return ErrShardRecentlyClosed
+	return &ErrShardClosed{
+		Msg:      "shard closed",
+		ClosedAt: *closedAt,
 	}
-
-	return ErrShardClosed
 }
 
 func (s *contextImpl) closeShard() {
--- a/service/history/shard/context.go
+++ b/service/history/shard/context.go
@@ -638,6 +638,9 @@ func (s *contextImpl) CreateWorkflowExecution(
 		return nil, err
 	}
 
+	if err := s.closedError(); err != nil {
+		return nil, err
+	}
 	currentRangeID := s.getRangeID()
 	request.RangeID = currentRangeID
 
@@ -745,6 +748,9 @@ func (s *contextImpl) UpdateWorkflowExecution(
 		}
 	}
 
+	if err := s.closedError(); err != nil {
+		return nil, err
+	}
 	currentRangeID := s.getRangeID()
 	request.RangeID = currentRangeID
 
@@ -860,6 +866,9 @@ func (s *contextImpl) ConflictResolveWorkflowExecution(
 		}
 	}
 
+	if err := s.closedError(); err != nil {
+		return nil, err
+	}
 	currentRangeID := s.getRangeID()
 	request.RangeID = currentRangeID
 	resp, err := s.executionManager.ConflictResolveWorkflowExecution(ctx, request)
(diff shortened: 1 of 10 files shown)


service/history/shard/context.go
```diff
@@ -633,9 +638,6 @@ func (s *contextImpl) CreateWorkflowExecution(
 		return nil, err
 	}
 
-	if s.isClosed() {
```
Shaddoll, line 636: shard can be closed by another thread before this thread reaching line 620, so it's still reachable

service/history/shard/context_test.go
```diff
@@ -280,17 +279,6 @@ func TestGetWorkflowExecution(t *testing.T) {
 			expectedResult: nil,
 			expectedError:  errors.New("some random error"),
 		},
-		{
```
Shaddoll, line 283: why do you remove this test case?

service/history/shard/context.go
jakobht: That's true! The description I made is wrong. Do you think we should not delete it? I only deleted it because of the new line check — it would be quite hard to write a test hitting this. That's of cause a terrible reason to make the code worse, just to be clear. 

I suppose it is better to have both? What do you think?

Shaddoll (approved):

service/history/shard/context_test.go
```diff
@@ -280,17 +279,6 @@ func TestGetWorkflowExecution(t *testing.T) {
...
-			name:     "Shard closed",
-			isClosed: true,
-			request: &persistence.GetWorkflowExecutionRequest{
-				DomainID:  "testDomain",
-				Execution: types.WorkflowExecution{WorkflowID: "testWorkflowID", RunID: "testRunID"},
-			},
-			mockSetup:      func(mgr *mocks.ExecutionManager) {},
-			expectedResult: nil,
-			expectedError:  ErrShardClosed,
```
taylanisikdemir, line 292: why remove this case?

service/history/shard/context.go
```diff
@@ -176,16 +176,21 @@ var _ Context = (*contextImpl)(nil)
...
-	logWarnCrossClusterLevelLag = 3000000 // 3 million
-	logWarnTimerLevelDiff       = time.Duration(30 * time.Minute)
-	historySizeLogThreshold     = 10 * 1024 * 1024
-	minContextTimeout           = 1 * time.Second
+	logWarnTransferLevelDiff     = 3000000 // 3 million
+	logWarnCrossClusterLevelLag  = 3000000 // 3 million
+	logWarnTimerLevelDiff        = time.Duration(30 * time.Minute)
+	historySizeLogThreshold      = 10 * 1024 * 1024
+	minContextTimeout            = 1 * time.Second
+	timeBeforeShardClosedIsError = 10 * time.Second
```
taylanisikdemir, line 193: did you found cases shard is closed momentarily and and re-opened? In which cases this can happen?

service/history/shard/context.go
```diff
@@ -176,16 +176,21 @@ var _ Context = (*contextImpl)(nil)
 var (
 	// ErrShardClosed is returned when shard is closed and a req cannot be processed
 	ErrShardClosed = errors.New("shard closed")
+
+	// ErrShardRecentlyClosed is returned when a shard has recently been closed, this
+	// error is transient, and should not cause error logs and error metrics to be emitted
+	ErrShardRecentlyClosed = errors.New("shard recently closed")
```
taylanisikdemir, line 182: There are other places (history queues) that stop themselves when they see shard closed error. This change will delay that process and might cause more attempts on recently closed shards. 
See usages of `shard.ErrShardClosed`

service/history/shard/context.go
jakobht: We have seen cases where the same host keeps emitting the `shard closed` error for good. We are not sure what causes this, but we have seen it happen. We then have had to restart the history host. 

This is not root caused, so we should still have these errors emitted if we keep seeing this a long time after the shard was closed.

service/history/shard/context.go
jakobht: Yeah, I see, I should have found that. 

Maybe it's better to add a timestamp to the shardClosed error, then the places that care about it can act accordingly?

jakobht: So I'd suggest creating a shardClosed error type, this would have the same message, but include the timestamp of when the shard was closed. 
Then I will change the places that use the error to check the type with errors.As/errors.Is, and then I can extract the timestamp in the handler where I need it. 

Do you think that would work? (just want to confirm before I start working on it :slightly_smiling_face:  )

taylanisikdemir: Not a bad idea to include shard closed timestamp for informational purposes.
However I don't see a reason to check the time elapsed since shard closed. Just change the task.go with
```
if err == shard.ErrShardClosed {
	return err
}
```

We should not keep retry on it. Doesn't matter when it's closed. it will not re-open :) 
cc @Shaddoll

jakobht: So I added the context in the alert. 

We have seen the bad state persist for hours, so I'm worried about just removing the metrics completely. 

I'm not sure what causes it, but we just keep seeing these errors emitted for some shard/host combination, until the host is restarted. 

So we should handle that as a "real" error state and emit the logs, etc. 

If we can resolve this, then we should probably not need this timeout any more.

service/history/shard/context.go
taylanisikdemir: I investigated that nonstop `shard closed` errors issue a few weeks ago and root cause is the [failover queues that are created but never stopped](https://github.com/uber/cadence/blob/2f67371c07ee68d157ffd1afb1a072f9c93ea18e/service/history/queue/transfer_queue_processor.go#L290). See CDNC-8135

service/history/task/task.go
```diff
@@ -265,6 +265,12 @@ func (t *taskImpl) HandleErr(err error) (retErr error) {
 		return err
 	}
 
+	// If the shard were recently closed we just return an error, so we retry in a bit.
+	var errShardClosed *shard.ErrShardClosed
+	if errors.As(err, &errShardClosed) && time.Since(errShardClosed.ClosedAt) < shard.TimeBeforeShardClosedIsError {
```
taylanisikdemir, line 270: Rest of the changes look good in this iteration. I don't think we need this check here because once shard is closed it's closed. The queue/task components will not be re-used even if that same shard is re-assigned to this host

jakobht: This is the check that ensures that we do not emit error logs and metrics immediately when the shard is closed, causing all the noise. 
So we need to return early here, to avoid the noise.

taylanisikdemir: let's also consider making shard closed error non-retryable here https://github.com/uber/cadence/blob/b005d32d12c1c58a74acc46a40e84e567b7e7795/service/history/task/task.go#L323

PR #5941, opened 2024-04-26
[Pinot] fix bug when querying a string field in attr with an empty value

bowenxia:
**What changed?**
Add an if statement in when the query has an empty value, query validator will return a difference query.

**Why?**
One customer that was querying with a empty value had an error in mono repo. It is an edge case that the previous query couldn't handle. 

**How did you test it?**
unit test + manual test with running Pinot locally

**Potential risks**

**Release notes**

**Documentation Changes**


neil-xie (approved):
LGTM

PR #5942, opened 2024-04-26
fix: close schema file

testwill:
**What changed?**

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/tools/common/schema/setuptask.go
+++ b/tools/common/schema/setuptask.go
@@ -65,6 +65,7 @@ func (task *SetupTask) Run() error {
 			return err
 		}
 		stmts, err := ParseFile(file)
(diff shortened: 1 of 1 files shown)


tools/common/schema/setuptask.go
```diff
@@ -64,6 +64,7 @@ func (task *SetupTask) Run() error {
 		if err != nil {
 			return err
 		}
+		defer file.Close()
```
dkrotx, line 67: good catch!
I'd recommend using defer only in smaller scope.
I suggest extracting the parsing to a separate function, then defer is OK.
Alternatively close right after ParseFile.

PR #5952, opened 2024-04-29
Bugfix: we address hosts using string(rune(shardID)), not by itoa(shardD)

dkrotx:
I wish it be the other way, but we can't change how it works now since
in all the other places we already use rune(shardID).
This was not a critical issue - we just returned ShardOwnershipLostError
with bad target host in very specific cases. For instance, at the
persistence engine error - when we got an error from transaction.

Previously we returned wrong host from ShardOwnershipLostError which led cadence-frontend to hit a wrong 
host when cadence-history instance lost ownership of the shard during persistance operation.

Everywhere else we use hashring.Lookup(string(rune(shardID))), except for two cases.

Did not check it, looks very much like an error + hard to reproduce (we should be in a middle of persistance Op. for this)

I think it is a clear error now, so let's just fix.

Fixed a bug which sometimes caused a wrong host in ShardOwnershipLostError.

### DIFF
--- a/service/history/handler/handler.go
+++ b/service/history/handler/handler.go
@@ -42,7 +42,6 @@ import (
 	"github.com/uber/cadence/common/metrics"
 	"github.com/uber/cadence/common/persistence"
 	"github.com/uber/cadence/common/quotas"
-	"github.com/uber/cadence/common/service"
 	"github.com/uber/cadence/common/types"
 	"github.com/uber/cadence/common/types/mapper/proto"
 	"github.com/uber/cadence/service/history/config"
@@ -51,6 +50,7 @@ import (
 	"github.com/uber/cadence/service/history/engine/engineimpl"
 	"github.com/uber/cadence/service/history/events"
 	"github.com/uber/cadence/service/history/failover"
+	"github.com/uber/cadence/service/history/lookup"
 	"github.com/uber/cadence/service/history/replication"
 	"github.com/uber/cadence/service/history/resource"
 	"github.com/uber/cadence/service/history/shard"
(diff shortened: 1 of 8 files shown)


service/history/handler/handler.go
```diff
@@ -2069,7 +2068,7 @@ func (h *handlerImpl) RatelimitUpdate(
 func (h *handlerImpl) convertError(err error) error {
 	switch err := err.(type) {
 	case *persistence.ShardOwnershipLostError:
-		info, err2 := h.GetMembershipResolver().Lookup(service.History, strconv.Itoa(err.ShardID))
+		info, err2 := h.GetMembershipResolver().Lookup(service.History, string(rune(err.ShardID)))
```
3vilhamster, line 2071: maybe as a followup PR, but could we have a function that accepts shardID and then hides this logic in 1 place? :)

jakobht (approved):

service/history/handler/handler.go
dkrotx: That's a good idea!
I'm afraid codecov for new lines won't pass this though :-(
Let me check what's with the coverage of the other multiple lines where `Lookup()` is used.

taylanisikdemir (approved):
Good catch and thanks for addressing this. +1 for @3vilhamster 's idea of moving the shardid casting to a common place

service/history/handler/handler.go
dkrotx: Turns out, not that bad! Had to introduce a new directory to avoid cycle dependencies on `go-generate` though.

PR #5959, opened 2024-04-30
Adding unit tests for client/matching/client.go

sankari165:
**What changed?**
Unit tests for client/matching/client.go
Renamed peerResolver file 
Added mocks for peer resolver and load balancer

**Why?**
Improve unit tests for the code

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


client/matching/client_test.go
```diff
@@ -0,0 +1,521 @@
...
+			client := NewMockClient(ctrl)
+			peerResolverMock := NewMockPeerResolver(ctrl)
+			loadbalancerMock := NewMockLoadBalancer(ctrl)
+			tt.mock(peerResolverMock, loadbalancerMock, client)
+			c := NewClient(client, peerResolverMock, loadbalancerMock)
+
+			res, err := tt.op(c)
+			if tt.wantError {
+				assert.Error(t, err)
+				assert.Nil(t, res)
```
jakobht, line 444: Nit: I think we should not care about the result when we return an error

jakobht (approved):
LGTM - added a nit comment

client/matching/client_test.go
sankari165: I agree to some extent but also think its good to keep this check to make sure all results are included

PR #5967, opened 2024-05-03
Enforcing go vet -copylocks and fixing current violations

Groxx:
Thankfully this all appears to be test-related and not any real risks,
but I realized while writing some other code that `go test` doesn't check this.

I'm really not sure what scenarios it might be wrong in, but it catches loads
of incorrect concurrent code in my experience.  So now it's required.


Groxx:
Currently blocked on https://github.com/uber/cadence/pull/5967 due to test flakiness.

PR #5969, opened 2024-05-06
Addition of tests for dlqMessageHandler in common domain

abhishekj720:
**What changed?**
Added test for start, stop, merge and emitDLQSize in dlqMessageHandler

**Why?**
As part of test coverage week

**How did you test it?**
Unit tests

**Potential risks**
No risks, as these are unit tests

**Release notes**

**Documentation Changes**


common/domain/dlqMessageHandler_test.go
```diff
@@ -22,14 +22,19 @@ package domain
 
 import (
 	"context"
+	"errors"
```
taylanisikdemir, line 25: let's change the file names (test file and corresponding code file) to snake_case as we go.

common/domain/dlqMessageHandler_test.go
```diff
@@ -102,6 +107,124 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+			initialStatus:  common.DaemonStatusStarted,
+			expectedStatus: common.DaemonStatusStarted,
+			shouldStart:    false,
+		},
+	}
+
+	for _, test := range tests {
+		s.Run(test.name, func() {
+			atomic.StoreInt32(&s.dlqMessageHandler.status, test.initialStatus)
+			s.dlqMessageHandler.Start()
```
taylanisikdemir, line 134: always stop what you start in tests to avoid leaks

common/domain/dlqMessageHandler_test.go
```diff
@@ -359,3 +482,49 @@ func (s *dlqMessageHandlerSuite) TestMergeMessages_IgnoreErrorOnUpdateDLQAckLeve
...
+	go s.dlqMessageHandler.emitDLQSizeMetricsLoop()
+
+	// Allow some time for the goroutine to run and tick at least once
+	time.Sleep(100 * time.Millisecond)
+
+	// Close the done channel to signal the loop to stop
+	close(s.dlqMessageHandler.done)
+
+	// Wait a bit to ensure the loop exits
+	time.Sleep(100 * time.Millisecond)
```
taylanisikdemir, line 526: can we lower the timer interval to avoid wasting 200ms here

common/domain/dlq_message_handler_test.go
```diff
@@ -102,6 +107,125 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+		s.Run(test.name, func() {
+			atomic.StoreInt32(&s.dlqMessageHandler.status, test.initialStatus)
+			s.dlqMessageHandler.Start()
+			if test.shouldStart {
+				s.dlqMessageHandler.logger.Info("Domain DLQ handler started.")
+			}
+			s.Equal(test.expectedStatus, atomic.LoadInt32(&s.dlqMessageHandler.status))
+		})
+	}
+	s.dlqMessageHandler.Stop()
```
taylanisikdemir, line 141: this doesn't look right. it should be stopped within the context it is started. 
```
s.dlqMessageHandler.Start()
defer s.dlqMessageHandler.Stop()
```

abhishekj720: We can put it in shouldStart. But not defer for both cases, as in 2nd case (Should not start when already started) Daemon is already started. 
``	s.dlqMessageHandler.Start()
			s.Equal(test.expectedStatus, atomic.LoadInt32(&s.dlqMessageHandler.status))
			if test.shouldStart {
				s.dlqMessageHandler.logger.Info("Domain DLQ handler started.")
				s.dlqMessageHandler.Stop()
			}
			``

abhishekj720: or can do something like:
s.NotPanics(func() {
                s.dlqMessageHandler.Stop()
            })
for multiple stop calls irrespective of less number of start

taylanisikdemir: what is down side of `defer stop()`? it should work whether it's started or not

abhishekj720: No, it does not. stop is not able to complete because of panic:
panic: close of closed channel

abhishekj720: so, if you see the first statement in Start there it is checking whether it is already started or not, and if it was already started Daemon will be not be started again.

DaemonStart.

taylanisikdemir: It's because you are initializing `dlqMessageHandler` inside `SetupTest` and it gets reused by all test cases inside `TestStart`. I recommend not using the suite at all here. We want full isolation and control over object lifecycle in these tests.

common/domain/dlq_message_handler_test.go
```diff
@@ -102,6 +110,146 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+			},
+			initialStatus:  common.DaemonStatusStarted,
+			expectedStatus: common.DaemonStatusStarted,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
```
Groxx, line 141: no longer necessary

Groxx, line 141:
```suggestion
```
gomock has done this automatically via `t.Cleanup` for 4 years now: https://github.com/golang/mock/issues/407

common/domain/dlq_message_handler_test.go
```diff
@@ -102,6 +110,146 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+		},
+	}
+
+	for _, test := range tests {
+		s.Run(test.name, func() {
+			atomic.StoreInt32(&s.dlqMessageHandler.status, test.initialStatus)
+			s.dlqMessageHandler.Stop()
+			if test.shouldStop {
+				s.dlqMessageHandler.logger.Info("Domain DLQ handler shutting down.")
+			}
```
Groxx, line 195: why log here?

common/domain/dlq_message_handler_test.go
```diff
@@ -102,6 +110,146 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+				logger:             mockLogger,
+				done:               doneChan,
+				status:             tc.initialStatus,
+				timeSrc:            mockedTime,
+			}
+
+			handler.Start()
+			mockedTime.Advance(5 * time.Minute)
+			defer handler.Stop()
+			assert.Equal(t, tc.expectedStatus, atomic.LoadInt32(&handler.status), assert.Equal(t, tc.expectedStatus, atomic.LoadInt32(&handler.status), "Handler did not reach the expected status for test case: %s", tc.name))
```
Groxx, line 163: er.  an assert inside an assert used as the message for an assert?

abhishekj720: oh shit :( silly mistake

common/domain/dlq_message_handler_test.go
```diff
@@ -102,6 +110,141 @@ func (s *dlqMessageHandlerSuite) TestReadMessages() {
...
+				replicationQueue:   mockQueue,
+				metricsClient:      metrics.NewNoopMetricsClient(),
+				logger:             mockLogger,
+				done:               doneChan,
+				status:             tc.initialStatus,
+				timeSrc:            mockedTime,
+			}
+
+			handler.Start()
+			mockedTime.Advance(5 * time.Minute)
```
taylanisikdemir, line 159: nice use of mock clock!

common/domain/dlq_message_handler_test.go
```diff
@@ -359,3 +502,49 @@ func (s *dlqMessageHandlerSuite) TestMergeMessages_IgnoreErrorOnUpdateDLQAckLeve
...
+
+	s.mockReplicationQueue.EXPECT().GetDLQAckLevel(gomock.Any()).Return(ackLevel, nil).Times(1)
+	s.mockReplicationQueue.EXPECT().GetMessagesFromDLQ(gomock.Any(), ackLevel, lastMessageID, pageSize, pageToken).
+		Return(tasks, nil, nil).Times(1)
+
+	token, err := s.dlqMessageHandler.Merge(context.Background(), lastMessageID, pageSize, pageToken)
+
+	s.NotNil(err)
+	s.IsType(&types.InternalServiceError{}, err)
+	s.Equal("Encounter non domain replication task in domain replication queue.", err.Error())
```
taylanisikdemir, line 528: instead of doing all these 3 checks, could we have used `s.EqualError`?

common/domain/dlq_message_handler_test.go
```diff
@@ -359,3 +502,49 @@ func (s *dlqMessageHandlerSuite) TestMergeMessages_IgnoreErrorOnUpdateDLQAckLeve
...
+	// Allow some time for the goroutine to run and tick at least once
+	time.Sleep(1 * time.Nanosecond)
+
+	// Close the done channel to signal the loop to stop
+	close(s.dlqMessageHandler.done)
+
+	// Wait a bit to ensure the loop exits
+	time.Sleep(1 * time.Nanosecond)
+
+	s.dlqMessageHandler.logger.Warn("Failed to get DLQ size.", tag.Error(errors.New("DomainReplicationQueueSizeLimit")))
```
taylanisikdemir, line 548: did you mean to add an expected warning call here instead of actually calling warn function?

common/domain/dlq_message_handler_test.go
```diff
@@ -359,3 +502,49 @@ func (s *dlqMessageHandlerSuite) TestMergeMessages_IgnoreErrorOnUpdateDLQAckLeve
...
+	go s.dlqMessageHandler.emitDLQSizeMetricsLoop()
+
+	// Allow some time for the goroutine to run and tick at least once
+	time.Sleep(1 * time.Nanosecond)
+
+	// Close the done channel to signal the loop to stop
+	close(s.dlqMessageHandler.done)
+
+	// Wait a bit to ensure the loop exits
+	time.Sleep(1 * time.Nanosecond)
```
taylanisikdemir, line 546: I don't think this setup ensures [the desired err handling branch](https://github.com/uber/cadence/blob/807a0e2899b4203dbf09cc0c2ae78bf2d75f10ec/common/domain/dlqMessageHandler.go#L226) to be hit. It's a tricky one and I think we can ignore that

Groxx: yea.  if we wanted to test that we'd probably need to:
- start the goroutine
- advance the clock
- wait for the mock call to occur (use a `.DoAndReturn` to monitor this, e.g. close a channel in there)
- then close the done chan

common/domain/dlq_message_handler_test.go
abhishekj720: right, i was just doing additional check for type of error. Maybe not needed. will remove it.

common/domain/dlq_message_handler_test.go
abhishekj720: yeah, maybe will add some time later

common/domain/dlq_message_handler_test.go
taylanisikdemir: yes. you are just adding another log. no validation here. 
as mentioned in other comment let's ignore this edge case. not much value in validating warning log is there

common/domain/dlq_message_handler_test.go
davidporter-id-au: I think the suggestion would be to skip the test for now then, time.Sleep tends to cause test flakiness in CD pipelines and make everyone sad

agautam478 (approved):

PR #5970, opened 2024-05-06
Added unit tests for service/history/handler

timl3136:
**What changed?**
Added more unit tests for methods in history/hander.go

**Why?**
Increase test coverage and stability

**How did you test it?**
Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/handler/handler_test.go
```diff
@@ -764,6 +764,464 @@ func (s *handlerSuite) TestRespondActivityTaskFailed() {
...
+		s.Run(name, func() {
+			switch input.caseName {
+			case "valid input":
+				s.mockTokenSerializer.EXPECT().Deserialize(gomock.Any()).Return(&common.TaskToken{
+					WorkflowID: testWorkflowID,
+					RunID:      testValidUUID,
+				}, nil).Times(1)
+				s.mockShardController.EXPECT().GetEngine(testWorkflowID).Return(s.mockEngine, nil).Times(1)
+				s.mockEngine.EXPECT().RespondActivityTaskCanceled(gomock.Any(), input.input).Return(nil).Times(1)
+				s.mockRatelimiter.EXPECT().Allow().Return(true).Times(1)
```
taylanisikdemir, line 864: Having such dependency on test case name is not a pattern we would like to follow. Ideally each test case definition is self contained and doesn't require test body to do special things for it. 
You can expose a `mockFn(<mock objects as parameter>` on each test case and just call it here

timl3136: Thanks, I have applied those changes accordingly.

taylanisikdemir (approved):

PR #5972, opened 2024-05-06
Split historyEngine.go into small files

taylanisikdemir:
**What changed?**
[historyEngine.go](https://github.com/uber/cadence/blob/b5674321fd0cf097e63ff70c87b9a21e369ad512/service/history/engine/engineimpl/historyEngine.go) is a giant file with ~4k lines which makes it hard to maintain/review and add tests for. 
Splitting to small files (mostly one file per public method) to make it more testable.

There are 3 separate test suites with some overlap of what they are testing
- `engineSuite`
- `engine2Suite`
- `engine3Suite`

In follow up PRs we are going to add test files for each file and split (get rid of) those suites as well.

**Why?**
Make it more readable and testable

**How did you test it?**
unit tests


natemort (approved):

service/history/engine/engineimpl/rerespond_decision_task_completed.go
Shaddoll, line 1: nit: filename

taylanisikdemir: thanks fixed

PR #5973, opened 2024-05-06
Added tests for starting, stopping and purging daemon for replicationQueue

abhishekj720:
**What changed?**
Added tests for replicationQueue present in common domain

**Why?**
As part of test coverage week

**How did you test it?**
Unit tests

**Potential risks**
No risks. Added to avoid risks

**Release notes**

**Documentation Changes**


common/domain/replication_queue_test.go
```diff
@@ -619,3 +710,28 @@ func TestPurgeAckedMessages(t *testing.T) {
...
+		rq.purgeProcessor()
+		close(done)
+	}()
+
+	time.Sleep(1 * time.Nanosecond)
+	rq.Stop()
+	select {
+	case <-done:
+		// Pass if the goroutine exits
+	case <-time.After(1 * time.Millisecond):
```
taylanisikdemir, line 734: 1ms might be too aggressive of deadline. On CI machines with limited CPU resources this may cause flakyness. I recommend 10ms. If `done` is closed earlier we will not wait for 10ms anyway

common/domain/replication_queue_test.go
```diff
@@ -619,3 +710,28 @@ func TestPurgeAckedMessages(t *testing.T) {
...
+	mockQueue := persistence.NewMockQueueManager(ctrl)
+	rq := NewReplicationQueue(mockQueue, "testCluster", nil, nil).(*replicationQueueImpl)
+	atomic.StoreInt32(&rq.status, common.DaemonStatusStarted)
+
+	done := make(chan bool)
+	mockQueue.EXPECT().GetAckLevels(gomock.Any()).Return(map[string]int64{}, nil).AnyTimes()
+	mockQueue.EXPECT().DeleteMessagesBefore(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+	go func() {
+		rq.purgeProcessor()
```
taylanisikdemir, line 725: instead of calling `purgeProcessor` directly you can start the replication queue. it will give us better coverage by validating that start eventually calls purgeProcessor

common/domain/replication_queue_test.go
```diff
@@ -619,3 +710,28 @@ func TestPurgeAckedMessages(t *testing.T) {
...
+		close(done)
+	}()
+
+	time.Sleep(1 * time.Nanosecond)
+	rq.Stop()
+	select {
+	case <-done:
+		// Pass if the goroutine exitsit
```
Groxx, line 733:
```suggestion
		// Pass if the goroutine exits
```

common/domain/replication_queue_test.go
```diff
@@ -619,3 +710,28 @@ func TestPurgeAckedMessages(t *testing.T) {
...
+	done := make(chan bool)
+	mockQueue.EXPECT().GetAckLevels(gomock.Any()).Return(map[string]int64{}, nil).AnyTimes()
+	mockQueue.EXPECT().DeleteMessagesBefore(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+	go func() {
+		rq.purgeProcessor()
+		close(done)
+	}()
+
+	time.Sleep(1 * time.Nanosecond)
```
Groxx, line 729: sleeping should be unnecessary here, and it doesn't really do anything for the purposes of this test - it'll effectively always take the `<-done` route because the ticker won't fire before `rq.Stop()` is called.

common/domain/replication_queue_test.go
abhishekj720: that is already done in testing for Start. This is specically for purge, do you want me to remove it then?

common/domain/replication_queue_test.go
```diff
@@ -619,3 +709,27 @@ func TestPurgeAckedMessages(t *testing.T) {
...
+	done := make(chan bool)
+	mockQueue.EXPECT().GetAckLevels(gomock.Any()).Return(map[string]int64{}, nil).AnyTimes()
+	mockQueue.EXPECT().DeleteMessagesBefore(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+	go func() {
+		rq.Start()
+		close(done)
+	}()
+
+	rq.Stop()
```
taylanisikdemir, line 728: `rq.Stop()` would be called before `rq.Start()` inside goroutine. 

What exactly are you trying to validate in this function that is not covered by Start/Stop? Expecting `purgeAckedMessages()`  is called once would be nice

agautam478 (approved):

PR #5975, opened 2024-05-07
Replace wurstmeister kafka/zookeeper images with bitnami kafka image

taylanisikdemir:
**What changed?**
Integration tests depend on following images that seems to be [gone](https://github.com/wurstmeister/zookeeper-docker/issues/31). 
- `wurstmeister/kafka:2.12-2.1.1`
- `wurstmeister/zookeeper:3.4.6`

These kafka/zookeeper images are 3rd party built and that project doesn't seem to be actively maintained. Not having these images is blocking [CI checks](https://buildkite.com/uberopensource/cadence-server/builds/19209#_) for all active PRs at the moment. 

To fix this issue I am switching to `bitnami/kafka:3.7` image which is newer version of kafka and doesn't have zookeeper dependency anymore. Updated all the docker compose files but didn't test the ones that are not covered by CI. There might be some follow up fixes.

**Why?**
Unblock CI.

**How did you test it?**
CI checks in this PR.

### DIFF
--- a/docker/buildkite/docker-compose-local-async-wf.yml
+++ b/docker/buildkite/docker-compose-local-async-wf.yml
@@ -40,7 +40,7 @@ services:
     networks:
       services-network:
         aliases:
-          - integration-test
+          - kafka
 
   integration-test-async-wf:
     build:
--- a/docker/buildkite/docker-compose-local-async-wf.yml
+++ b/docker/buildkite/docker-compose-local-async-wf.yml
@@ -1,4 +1,4 @@
-# version: "3.5"
+version: "3.5"
 
 services:
   cassandra:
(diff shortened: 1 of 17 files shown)


docker/buildkite/docker-compose-local-async-wf.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.5"
+# version: "3.5"
```
Shaddoll, line 1: why is this line commented out?

Shaddoll (approved):

docker/buildkite/docker-compose-local-async-wf.yml
taylanisikdemir: locally `docker-compose` was complaining about obsolete version. I'll revert. if new version of docker-compose doesn't need this we can remove from all files at bulk

PR #5977, opened 2024-05-07
Add tests for history/execution/history_builder.go

natemort:
**What changed?**
- Add additional test cases to history_builder_test.go

**Why?**
- Increase test coverage

**How did you test it?**
- Unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/execution/history_builder_test.go
+++ b/service/history/execution/history_builder_test.go
@@ -1276,6 +1276,253 @@ func (s *historyBuilderSuite) TestHistoryBuilderWorkflowExternalSignaled() {
 	}]
 	s.True(exists)
 }
+
+func (s *historyBuilderSuite) TestHistoryBuilderActivityTaskTimedOut() {
+	id := "historybuilder-workflow-id"
+	rid := "historybuilder-test-run-id"
+	wt := "historybuilder-type"
+	tl := "historybuilder-tasklist"
+	identity := "historybuilder-worker"
+	input := []byte("historybuilder-input")
+	execTimeout := int32(60)
+	taskTimeout := int32(10)
+	we := types.WorkflowExecution{
+		WorkflowID: id,
+		RunID:      rid,
+	}
+	partitionConfig := map[string]string{
+		"zone": "dca1",
+	}
+
+	// 1
+	s.addWorkflowExecutionStartedEvent(we, wt, tl, input, execTimeout, taskTimeout, identity, partitionConfig)
+	// 2
+	s.addDecisionTaskScheduledEvent()
+	// 3
+	s.addDecisionTaskStartedEvent(2, tl, identity)
+	decisionContext := []byte("historybuilder-context")
+	// 4
+	s.addDecisionTaskCompletedEvent(2, 3, decisionContext, identity)
+	activityID := "activity1"
+	activityType := "activity1-type"
+	activityDomain := ""
+	activityInput := []byte("activity1-input")
+	activityTimeout := int32(60)
+	queueTimeout := int32(20)
+	hearbeatTimeout := int32(10)
+	// 5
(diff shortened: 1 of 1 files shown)


service/history/execution/history_builder_test.go
```diff
@@ -1276,6 +1276,253 @@ func (s *historyBuilderSuite) TestHistoryBuilderWorkflowExternalSignaled() {
 	}]
 	s.True(exists)
 }
+
+func (s *historyBuilderSuite) TestHistoryBuilderActivityTaskTimedOut() {
+	id := "historybuilder-workflow-id"
```
timl3136, line 1281: Since you are reusing those constants for different tests, you can use const at the top of the file to declare those fields and avoiding repetitive declaration in each test functions.

service/history/execution/history_builder_test.go
```diff
@@ -1276,6 +1276,253 @@ func (s *historyBuilderSuite) TestHistoryBuilderWorkflowExternalSignaled() {
...
+	partitionConfig := map[string]string{
+		"zone": "dca1",
+	}
+
+	// 1
+	s.addWorkflowExecutionStartedEvent(we, wt, tl, input, execTimeout, taskTimeout, identity, partitionConfig)
+	req := &types.HistoryRequestCancelWorkflowExecutionRequest{
+		CancelRequest: &types.RequestCancelWorkflowExecutionRequest{
+			Identity:  "identity",
+			RequestID: "b071cbe8-3a95-4223-a8ac-f308a42db383",
```
timl3136, line 1443: Again, this uuid seems like a generic uuid and we can put them at top so that it can be reused.

service/history/execution/history_builder_test.go
timl3136: Nit: i'd prefer const instead of variable declaration since you are not changing it.

taylanisikdemir (approved):

service/history/execution/history_builder_test.go
```diff
@@ -1276,6 +1276,253 @@ func (s *historyBuilderSuite) TestHistoryBuilderWorkflowExternalSignaled() {
...
+	taskTimeout := int32(10)
+	we := types.WorkflowExecution{
+		WorkflowID: id,
+		RunID:      rid,
+	}
+	partitionConfig := map[string]string{
+		"zone": "dca1",
+	}
+
+	// 1
```
timl3136, line 1478: Nit: mind expanding on this comment? just a number in comments does not offer much value to readers.

service/history/execution/history_builder_test.go
davidporter-id-au: imho both ways are fine, the counterargument to putting them up the top of the file is for very big packages, then they're sorta hard to track down as they're package-scoped

davidporter-id-au (approved):

service/history/execution/history_builder_test.go
natemort: This is consistent with the rest of the file so I'm going to leave it as is. I think I generally would agree with you though. This file is massive and I don't think it's worth changing it or introducing a new standard.

service/history/execution/history_builder_test.go
natemort: This is another oddity of the file that I'm sticking with. These are the event ids, which are allocated to each event in order. Many of the events reference other events, [like on this line](https://github.com/uber/cadence/pull/5977/files/5801149c3bcd91f015a40ec7f2f2a30729d598bd#diff-0fbf138df4faf3cfe428e66630a1b71231e579fdeb388c5157636806b6b059a0R1485), so it's easier to tell what `2` and `3` are referring to.

service/history/execution/history_builder_test.go
timl3136: Got it, thanks for explaining.

PR #5985, opened 2024-05-08
History engine start/stop unit tests

taylanisikdemir:
**What changed?**
It is now possible to write small scoped unit tests for history engine after it got splitted into small files in #5972.
This PR adds a basic Start/Stop test. 
Changes:
- Generated some mocks that were missing
- Introduced a QueueProcessorFactory to easily mock queue processors. Previously they were created inside constructor. This will also be handy when we try different queue implementations.

**Why?**
Improve coverage and make code testable.

**How did you test it?**
Unit tests


service/history/engine/testdata/engine_for_tests.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/uber/cadence/service/history/replication"
+	"github.com/uber/cadence/service/history/shard"
+	"github.com/uber/cadence/service/history/task"
+	"github.com/uber/cadence/service/history/workflowcache"
+)
+
+type EngineForTest struct {
+	Engine  engine.Engine
+	Cleanup func()
+	// Add mocks or other fields here
```
taylanisikdemir, line 50: Planning to attach mock objects created inline inside `NewEngineForTest` function here so different unit test scenarios can adjust the behavior/expectations

service/history/engine/testdata/engine_for_tests.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type NewEngineFn func(
```
taylanisikdemir, line 53: this type is defined as an alias to `engineimpl.NewEngineWithShardContext` to avoid circular dependency

service/history/queue/factory.go
```diff
@@ -0,0 +1,128 @@
...
+		executionCache *execution.Cache,
+		taskProcessor task.Processor,
+	) Processor
+}
+
+func NewProcessorFactory() ProcessorFactory {
+	return &factoryImpl{}
+}
+
+type factoryImpl struct {
```
taylanisikdemir, line 71: unit tests for this file will be added later

service/history/engine/testdata/engine_for_tests.go
Groxx: seems like that may be worth adding a a comment tbh.  it's a good reason, it just doesn't seem all that likely to be clear at a glance months later and there's no reason to need to figure it out again.

service/history/queue/factory.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/uber/cadence/service/history/reset"
+	"github.com/uber/cadence/service/history/shard"
+	"github.com/uber/cadence/service/history/task"
+	"github.com/uber/cadence/service/history/workflowcache"
+	"github.com/uber/cadence/service/worker/archiver"
+)
+
+//go:generate mockgen -package $GOPACKAGE -source $GOFILE -destination factory_mock.go -self_package github.com/uber/cadence/service/history/queue
+
+type ProcessorFactory interface {
```
Groxx, line 37: tbh it feels a little odd that these are bundled together.  is there any reason other than convenience, since all three are used by the same caller(s?)?

taylanisikdemir: I haven't made my mind how the final objects/interfaces will look like after refactoring but I am also considering the option where there's just single "Queue" object that history interacts with. Based on the task type, domain and other attributes that top level object decides where to route the request.

service/history/queue/factory.go
taylanisikdemir: this is blocking codecov patch coverage checks so addressing in this PR.

PR #5989, opened 2024-05-08
testing slack notifications

Groxx:


Groxx:
switched to buildkite config as this wasn't working. maybe we'll try again some other time.

PR #5996, opened 2024-05-09
Replication task processor shutdown improvements and start/stop unit tests

taylanisikdemir:
**What changed?**
Replication task processor was not waiting for underlying goroutines when stopped. Added a waitgroup to fix that and validated with a unit test using `goleak.VerifyNone`

Also added similar leak check in history engine's start/stop test.

**Why?**
Improve shutdowns and code coverage

**How did you test it?**
unit test

### DIFF
--- a/service/history/replication/task_processor.go
+++ b/service/history/replication/task_processor.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"math"
 	"strconv"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


timl3136 (approved):

service/history/replication/task_processor.go
```diff
@@ -172,25 +175,39 @@ func (p *taskProcessorImpl) Stop() {
...
 			return
-		default:
+		case p.requestChan <- &request{
+			token: &types.ReplicationToken{
+				ShardID:                int32(p.shard.GetShardID()),
+				LastRetrievedMessageID: p.lastRetrievedMessageID,
+				LastProcessedMessageID: p.lastProcessedMessageID,
+			},
+			respChan: respChan,
+		}:
```
taylanisikdemir, line 207: previously this was done inside `sendFetchMessageRequest` which was not checking `p.done` and blocking shutdowns

PR #5998, opened 2024-05-09
mutable-state: copy to persistence round-trip test

davidporter-id-au:
**What changed?**
Fixes minor bugs in copy method and adds a small round-trip test

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


Shaddoll (approved):

service/history/execution/mutable_state_builder_test.go
```diff
@@ -1892,3 +1894,43 @@ func TestLog(t *testing.T) {
 	assert.NotPanics(t, func() { e.logWarn("a") })
 	assert.NotPanics(t, func() { e.logError("a") })
 }
+
+func TestMutableStateBuilder_CopyToPersistence_roundtrip(t *testing.T) {
+
+	for i := 0; i <= 100; i++ {
```
natemort, line 1900: How long does this take to run? Since we're using unix seconds as the seed and creating a new fuzz instance each time isn't it possible that a majority (or all of these) are running with the same seed?

If we used the iteration count as the seed it would be deterministic (so no chance of ever flaking) and guarantee that each run is different.

davidporter-id-au: good spotting, yeah, let me use more granular seeds

service/history/execution/mutable_state_builder.go
```diff
@@ -316,6 +316,8 @@ func (e *mutableStateBuilder) CopyToPersistence() *persistence.WorkflowMutableSt
 	state.BufferedEvents = e.bufferedEvents
 	state.VersionHistories = e.versionHistories
 	state.Checksum = e.checksum
+	state.ReplicationState = e.replicationState
+	state.ExecutionStats = e.executionStats
```
taylanisikdemir, line 320: did the roundtrip test uncover these missed fields?

taylanisikdemir (approved):

service/history/execution/mutable_state_builder.go
davidporter-id-au: yes

PR #6000, opened 2024-05-09
Refactor matching and split the files into subpackages

Shaddoll:
**What changed?**
Refactor matching and split the files into subpackages

**Why?**
To improve the quality of the codebase

**How did you test it?**
existing unit tests

**Potential risks**
The risk is a bit high because it changes almost everything inside matching packages by splitting the files into subpackages. We need at 2 pairs of eyes for reviewing this PR.

**Release notes**

**Documentation Changes**


service/matching/poller/history.go
```diff
@@ -35,15 +35,15 @@ const (
...
 
 type (
-	pollerIdentity string
+	Identity string
 
-	pollerInfo struct {
-		ratePerSecond  *float64
-		isolationGroup string
+	Info struct {
+		RatePerSecond  float64
```
davidporter-id-au, line 41: Might suggest we don't change this as part of a refactor unless we have to? It'll be tricky to track down if it causes problems.

Shaddoll: We have to change it, because otherwise we need to duplicate the definition of defaultTaskDispatchRPS in both tasklist package and poller package or export that constant, which is not ideal.

codecov.yml
```diff
@@ -63,6 +63,7 @@ ignore:
   - "service/history/workflow/errors.go"
   - "service/history/service.go"
   - "service/matching/service.go"
+  - "service/matching/tasklist/testing.go"
```
taylanisikdemir, line 66: FYI I'm adding this to internal sonarqube exclusion file

service/matching/poller/history.go
```diff
@@ -55,65 +55,61 @@ type pollerHistory struct {
...
 		if earliestAccessTime.Before(lastAccessTime) {
-			rps := _defaultTaskDispatchRPS
-			if value.ratePerSecond != nil {
-				rps = *value.ratePerSecond
-			}
 			result = append(result, &types.PollerInfo{
 				Identity:       string(key),
 				LastAccessTime: common.Int64Ptr(lastAccessTime.UnixNano()),
-				RatePerSecond:  rps,
+				RatePerSecond:  value.RatePerSecond,
```
taylanisikdemir, line 94: Previously this value was either `_defaultTaskDispatchRPS` or `value.ratePerSecond` if it's not nil.
Is there any place using this value that will be impacted by probably seeing more zeros?

service/matching/tasklist/task_list_manager.go
```diff
@@ -138,19 +153,26 @@ const (
 	maxSyncMatchWaitTime = 200 * time.Millisecond
 )
 
-var _ taskListManager = (*taskListManagerImpl)(nil)
+var _ Manager = (*taskListManagerImpl)(nil)
```
taylanisikdemir, line 156: nit: this is not needed

service/matching/tasklist/testing.go
```diff
@@ -0,0 +1,375 @@
+// The MIT License (MIT)
```
taylanisikdemir, line 1: this file path is not in the exclusions. can you move it to testing/ or testdata/ folder?

service/matching/poller/history.go
Shaddoll: I updated the caller in this PR to let the caller pass the default value explicitly.
The reason is that I want the default value to be defined in one place and don't want to export that value to other packages. And now the default value is defined in `config/config.go`.

taylanisikdemir (approved):

service/matching/tasklist/testing.go
```diff
@@ -0,0 +1,375 @@
...
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/emirpasic/gods/maps/treemap"
```
davidporter-id-au, line 33: was wondering about why this was introduced as part of this refactor?

davidporter-id-au: discussed offline, was in matching_test.go

cmd/server/go.sum
```diff
@@ -97,6 +97,7 @@ github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21/go.mod h1
 github.com/eapache/queue v1.1.0 h1:YOEu7KNc61ntiQlcEeUIoDTJ2o8mQznoNvUhiigpIqc=
 github.com/eapache/queue v1.1.0/go.mod h1:6eCeP0CKFpHLu8blIFXhExK/dRa7WDZfr6jVFPTqq+I=
 github.com/emirpasic/gods v0.0.0-20190624094223-e689965507ab h1:eTc1vwMHNg4WtS95PtYi3FFCKwlPjtN/Lw9IALTRtd8=
+github.com/emirpasic/gods v0.0.0-20190624094223-e689965507ab/go.mod h1:YfzfFFoVP/catgzJb4IKIqXjX78Ha8FMSDh3ymbK86o=
```
davidporter-id-au, line 100: discussed, not introduced in this change, but it was pulled into the production compilation path from testing and I am guessing this triggered gomod to do something

codecov.yml
davidporter-id-au: @shaddoll what was the rationale for pulling this into a non `_test.go` file? Nit comment, but this means it's compiled into prod binaries

service/matching/handler/engine.go
```diff
@@ -240,11 +249,12 @@ func (e *matchingEngineImpl) getTaskListByDomainLocked(domainID string) *types.G
...
 	for tl, tlm := range e.taskLists {
-		if tlm.GetTaskListKind() == types.TaskListKindNormal && tl.domainID == domainID {
-			if types.TaskListType(tl.taskType) == types.TaskListTypeDecision {
-				decisionTaskListMap[tl.baseName] = tlm.DescribeTaskList(false)
+		if tlm.GetTaskListKind() == types.TaskListKindNormal && tl.GetDomainID() == domainID {
+			if types.TaskListType(tl.GetType()) == types.TaskListTypeDecision {
+				decisionTaskListMap[tl.GetRoot()] = tlm.DescribeTaskList(false)
 			}
-			activityTaskListMap[tl.baseName] = tlm.DescribeTaskList(false)
+			// TODO: review this logic
```
davidporter-id-au, line 256: mind putting in the comment what seems off?

service/matching/tasklist/matcher.go
```diff
@@ -60,20 +61,15 @@ type TaskMatcher struct {
...
 var ErrTasklistThrottled = errors.New("tasklist limit exceeded")
 
 // newTaskMatcher returns an task matcher instance. The returned instance can be
 // used by task producers and consumers to find a match. Both sync matches and non-sync
 // matches should use this implementation
-func newTaskMatcher(config *taskListConfig, fwdr *Forwarder, scope metrics.Scope, isolationGroups []string, log log.Logger) *TaskMatcher {
-	dPtr := _defaultTaskDispatchRPS
-	limiter := quotas.NewRateLimiter(&dPtr, _defaultTaskDispatchRPSTTL, config.MinTaskThrottlingBurstSize())
+func newTaskMatcher(config *config.TaskListConfig, fwdr *Forwarder, scope metrics.Scope, isolationGroups []string, log log.Logger) *TaskMatcher {
+	dPtr := config.TaskDispatchRPS
```
davidporter-id-au, line 71: Is this what you were referring to as having to be pulled into a central location?

davidporter-id-au (approved):
lgtm, actually looks a lot better

codecov.yml
Shaddoll: Synced offline, we need to export the type, because it's used by other packages.

service/matching/tasklist/matcher.go
Shaddoll: the central location of the default values is config/config.go

PR #6003, opened 2024-05-09
Add unit test for frontend/admin/handler - part 2

neil-xie:
**What changed?**
Add more unit test for frontend/admin/handler

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/frontend/admin/handler_test.go
+++ b/service/frontend/admin/handler_test.go
@@ -400,54 +400,6 @@ func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2_SameStartIDAnd
 	s.NoError(err)
 }
 
-func (s *adminHandlerSuite) Test_GetWorkflowExecutionRawHistoryV2_NextPageToken() {
-	ctx := context.Background()
-	s.mockDomainCache.EXPECT().GetDomainID(s.domainName).Return(s.domainID, nil).AnyTimes()
-	branchToken := []byte{1}
-	versionHistory := persistence.NewVersionHistory(branchToken, []*persistence.VersionHistoryItem{
-		persistence.NewVersionHistoryItem(int64(10), int64(100)),
-	})
-	rawVersionHistories := persistence.NewVersionHistories(versionHistory)
-	versionHistories := rawVersionHistories.ToInternalType()
-	mState := &types.GetMutableStateResponse{
-		NextEventID:        11,
-		CurrentBranchToken: branchToken,
-		VersionHistories:   versionHistories,
-	}
-	s.mockHistoryClient.EXPECT().GetMutableState(gomock.Any(), gomock.Any()).Return(mState, nil).AnyTimes()
-	token := &getWorkflowRawHistoryV2Token{
-		VersionHistories: &types.VersionHistories{
-			CurrentVersionHistoryIndex: 1,
-			Histories: []*types.VersionHistory{
-				{
-					BranchToken: branchToken,
-					Items: []*types.VersionHistoryItem{
-						{
-							EventID: 10,
-							Version: 100,
-						},
-					},
-				},
-			},
-		},
-	}
-	serializedToken, err := serializeRawHistoryToken(token)
-	s.NoError(err)
-	_, err = s.handler.GetWorkflowExecutionRawHistoryV2(ctx,
-		&types.GetWorkflowExecutionRawHistoryV2Request{
-			Domain: s.domainName,
-			Execution: &types.WorkflowExecution{
-				WorkflowID: "workflowID",
-				RunID:      uuid.New(),
-			},
-			StartEventID:      common.Int64Ptr(10),
-			StartEventVersion: common.Int64Ptr(100),
-			MaximumPageSize:   1,
-			NextPageToken:     serializedToken,
-		})
-	s.NoError(err)
-}
-
 func (s *adminHandlerSuite) Test_SetRequestDefaultValueAndGetTargetVersionHistory_DefinedStartAndEnd() {
 	inputStartEventID := int64(1)
 	inputStartVersion := int64(10)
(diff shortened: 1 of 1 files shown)


service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,950 @@ func TestGetReplicationMessages(t *testing.T) {
...
+				ClusterName:            "test-cluster",
+			},
+			hcHandlerFunc: func(mock *history.MockClient) {
+			},
+			drqHandlerFunc: func(mock *domain.MockReplicationQueue) {
+				mock.EXPECT().GetAckLevels(gomock.Any()).Return(map[string]int64{
+					"test-cluster": 1,
+				}, nil)
+				mock.EXPECT().GetReplicationMessages(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, int64(2), nil)
+				mock.EXPECT().UpdateAckLevel(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
davidporter-id-au, line 1650: I would suggest not using goock.Any() on write operations like this where you want to ensure that if you were to do a big refactor after this test was landed, that if they were to get zero'd out or were wrong or something that the test would still pass. 

Asserting - particularly in write cases - at least the important values, at both ends is probably worthwhile

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,950 @@ func TestGetReplicationMessages(t *testing.T) {
...
+		"normal request": {
+			input: &types.ReapplyEventsRequest{
+				DomainName: "test-domain",
+				WorkflowExecution: &types.WorkflowExecution{
+					WorkflowID: "test-workflow-id",
+				},
+				Events: &types.DataBlob{},
+			},
+			hcHandlerFunc: func(mock *history.MockClient) {
+				mock.EXPECT().ReapplyEvents(gomock.Any(), gomock.Any()).Return(nil)
```
davidporter-id-au, line 1849: same as above, you may want to be a little more specific about what's accepted

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,950 @@ func TestGetReplicationMessages(t *testing.T) {
...
+			if td.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func Test_SerializeRawHistoryToken(t *testing.T) {
```
davidporter-id-au, line 2505: consider doing a round-trip test on serializations. It is usually easier too. If you can serialize a struct, then unserialise it and then assert that they're equal you'll catch a lot of bugs. Look at `common/testdatagen` if you want an easy way to make data for round-trip tests

neil-xie: Right, it is much better than testing separately

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,1120 @@ func TestGetReplicationMessages(t *testing.T) {
...
+						FirstEventID: 1,
+					},
+				},
+			},
+			hcHandlerFunc: func(mock *history.MockClient) {
+				mock.EXPECT().GetDLQReplicationMessages(gomock.Any(), gomock.Any()).Return(nil, nil)
+			},
+			wantErr: false,
+		},
+		"get dql message error": {
```
taylanisikdemir, line 1749: nit typo: dlq

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,1120 @@ func TestGetReplicationMessages(t *testing.T) {
...
+						RunID:        uuid.New(),
+						TaskID:       int64(1),
+						TaskType:     1,
+						Version:      0,
+						FirstEventID: 1,
+					},
+				},
+			},
+			hcHandlerFunc: func(mock *history.MockClient) {
+				mock.EXPECT().GetDLQReplicationMessages(gomock.Any(), gomock.Any()).Return(nil, nil)
```
taylanisikdemir, line 1745: let's return some non-nil object here and validate it's received in test code

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,1120 @@ func TestGetReplicationMessages(t *testing.T) {
...
+				handler.throttleRetry = backoff.NewThrottleRetry(
+					backoff.WithRetryPolicy(policy),
+					backoff.WithRetryableError(func(_ error) bool { return true }),
+				)
+				mockDlqHandler := domain.NewMockDLQMessageHandler(goMock)
+				mockDlqHandler.EXPECT().Read(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil, nil)
+				handler.domainDLQHandler = mockDlqHandler
+			}
+
+			_, err := handler.ReadDLQMessages(context.Background(), td.input)
```
taylanisikdemir, line 2000: same here. missing validation of returned messages for success case

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,1120 @@ func TestGetReplicationMessages(t *testing.T) {
...
+			handler := adminHandlerImpl{
+				Resource: &resource.Test{
+					Logger:        testlogger.New(t),
+					MetricsClient: metrics.NewNoopMetricsClient(),
+					HistoryClient: hcMock,
+				},
+				domainDLQHandler: dlqMock,
+			}
+
+			_, err := handler.CountDLQMessages(context.Background(), td.input)
```
taylanisikdemir, line 2156: same here

service/frontend/admin/handler_test.go
```diff
@@ -1619,3 +1621,1120 @@ func TestGetReplicationMessages(t *testing.T) {
...
+			if td.wantErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+func Test_GetCrossClusterTasks(t *testing.T) {
```
taylanisikdemir, line 2398: I think @davidporter-id-au will cleanup cross cluster related code but doesn't hurt to keep these new tests until that happens

taylanisikdemir (approved):
LGTM with comments

PR #6004, opened 2024-05-09
Add test for history_replicator, ApplyEvent function

bowenxia:
**What changed?**
Add test for ApplyEvent function in history_replicator.go, and NewReplicationTask in replication_task.go.
Refactored code to have newReplicationTask function as a parameter of historyReplicatorImpl.

**Why?**
Code coverage week.
Use newReplicationTaskFn as a parameter is to increase the testability. Doing it this way, we can pass in mock newReplicationTaskFn functions so that we can test ApplyEvent function as a whole, and test the functions it used in different unit tests. 

**How did you test it?**
unit test

**Potential risks**

**Release notes**

**Documentation Changes**


service/history/ndc/history_replicator_test.go
```diff
@@ -357,3 +360,196 @@ func TestNewHistoryReplicator_newMutableState(t *testing.T) {
...
+	historyEvents4 := []*types.HistoryEvent{historyEvent4}
+	serializedEvents4, err := serializer.SerializeBatchEvents(historyEvents4, common.EncodingTypeThriftRW)
+	assert.NoError(t, err)
+
+	tests := map[string]struct {
+		request          *types.ReplicateEventsV2Request
+		eventsAffordance func() []*types.HistoryEvent
+		expectedErrorMsg string
+	}{
+		"Case1: empty case": {
```
davidporter-id-au, line 406: remark, no action required: I generally suggest starting from the full case first since it's the most laborious, and then working through the exception cases, it is just easier to delete stuff than to add it

davidporter-id-au: actually, are all of these jsut error cases? Is there a test somewhere where we verify a successful application?

bowenxia: Not yet. There are so many branches in one function, I was trying to build a full case first, but looks like it is more efficient to build the successful case step by step from different fail cases.

service/history/ndc/history_replicator_test.go
```diff
@@ -357,3 +360,196 @@ func TestNewHistoryReplicator_newMutableState(t *testing.T) {
 	)
 	assert.NotNil(t, testReplicatorImpl.newMutableState(mockDomainCacheEntry, log.NewNoop()))
 }
+
+func TestApplyEvents_newReplicationTask_validateReplicateEventsRequest(t *testing.T) {
```
Shaddoll, line 364: It looks like all the errors are from `validateReplicateEventsRequest`. 
Might consider creating some function parameters and overwriting them in unit tests to mock these functions.
And you can write tests for individual functions separately to test those functions referenced by ApplyEvents.
Here is an example you can follow: https://github.com/uber/cadence/pull/5971

bowenxia: Amazing idea! I've made refactors according to your advice. Please take another look when you got some time. 🙏

service/history/ndc/replication_task_test.go
```diff
@@ -79,3 +79,201 @@ func TestReplicationTaskResetEvent(t *testing.T) {
...
+					WorkflowID: "test-workflow-id",
+					RunID:      "test-run-id",
+				},
+				VersionHistoryItems: nil,
+				Events:              nil,
+				NewRunEvents:        nil,
+			},
+			expectedErrorMsg: "invalid run ID",
+		},
+		"Case3-3: fail case in validateReplicateEventsRequest with invalid run id": {
```
Shaddoll, line 159: this case is the same as 3-2.

Shaddoll (approved):
Are you planning to add tests for success case in another PR?

bowenxia:
@Shaddoll Yes. I'll add success case later in another PR since it will need more refactor in applyEvent function.

service/history/ndc/replication_task_test.go
bowenxia: was deleted

davidporter-id-au (approved):

PR #6010, opened 2024-05-10
Write tests for replication task processor main loop

taylanisikdemir:
**What changed?**
Added more test cases to cover replication/task_processor.go's main processing loop logic.

**Why?**
Improve coverage

**How did you test it?**
```
go test -timeout 60s -run ^TestTaskProcessorSuite$ github.com/uber/cadence/service/history/replication -race -v -count=3
```


service/history/replication/task_processor_test.go
```diff
@@ -167,6 +172,167 @@ func (s *taskProcessorSuite) TestSendFetchMessageRequest() {
...
+func (s *taskProcessorSuite) TestProcessorLoop_RespChanClosed() {
+	// start the process loop
+	s.taskProcessor.wg.Add(1)
+	go s.taskProcessor.processorLoop()
+	defer close(s.taskProcessor.done)
+
+	// act like taskFetcher here and populate respChan of the request
+	requestMessage := <-s.requestChan
+	close(requestMessage.respChan)
+	time.Sleep(50 * time.Millisecond)
```
Shaddoll, line 184: since we have time.After(50 * time.Millisecond), can we remove this time.Sleep?

service/history/replication/task_processor_test.go
```diff
@@ -167,6 +172,167 @@ func (s *taskProcessorSuite) TestSendFetchMessageRequest() {
...
+	// loop should have continued by now. validate by checking the new request
+	select {
+	case <-s.requestChan:
+	// expected
+	case <-time.After(50 * time.Millisecond):
+		s.Fail("new request not sent to requestChan")
+	}
+}
+
+func (s *taskProcessorSuite) TestProcessorLoop_TaskExecuteSuccess() {
```
Shaddoll, line 195: These tests are more like component tests. Why not use fake functions so that we don't need to worry about the logic in `processResponse` function?

taylanisikdemir: I could have gone that route but there's still value in such multi-function coverage. if the test logic became too complicated to mock all downstream calls then using function overrides would make more sense

service/history/replication/task_processor_test.go
taylanisikdemir: makes sense. this is not needed.

Shaddoll (approved):

PR #6012, opened 2024-05-10
Add unit test for replication task executor

Shaddoll:
**What changed?**
Add unit test for replication task executor

**Why?**
Improve coverage

**How did you test it?**
unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/service/history/replication/task_executor.go
+++ b/service/history/replication/task_executor.go
@@ -147,43 +147,45 @@ func (e *taskExecutorImpl) handleActivityTask(
 	}
 
 	err = syncActivityAction()
+	retryErr, ok := toRetryTaskV2Error(err)
+	if !ok {
+		return err
+	}
 	// Handle resend error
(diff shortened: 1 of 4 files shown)


service/history/replication/task_executor.go
```diff
@@ -147,43 +147,45 @@ func (e *taskExecutorImpl) handleActivityTask(
 	}
 
 	err = syncActivityAction()
+	retryErr, ok := toRetryTaskV2Error(err)
+	if !ok {
+		return err
```
taylanisikdemir, line 152: in existing implementation `syncActivityAction()` is still called for not `ok` cases at line 189

Shaddoll: I think that's a bug.

Shaddoll: Synced offline, we are in an agreement that this is a bug.

taylanisikdemir (approved):

PR #6013, opened 2024-05-10
Stop orphan failover queue processors when its parent stops

taylanisikdemir:
**What changed?**
Failover queue processors are started on the fly when domains are failed over. They are supposed to process the tasks until domain failover is completed. These failover queue processors are started and never stopped. 
So there's orphaned goroutines for sure.
Another problem is that the failover queue processor might have work to do but the shard ownership might have changed. In that case the task executor etc. is already stopped by the parent and failover queue processor keeps attempting/failing forever. Causing error/alert noise.

With the changes in this PR, failover queue processors will be stopped when its creator is stopped (due to restart or shard movement). No leftover goroutines and forever retryed tasks. Changes are hidden behind existing/relevant `history.replicationTaskFetcherEnableGracefulSyncShutdown` dynamic config. 

Another problem that's not address in this PR: There is no mechanism to re-start ongoing failover queue processors in the new shard owner.

**Why?**
- Eliminate source of goroutine leaks
- Reduce alert noise

**How did you test it?**
Will be tested in staging environments first.

### DIFF
--- a/service/history/queue/timer_queue_processor.go
+++ b/service/history/queue/timer_queue_processor.go
@@ -310,7 +310,6 @@ func (t *timerQueueProcessor) FailoverDomain(domainIDs map[string]struct{}) {
 	updateClusterAckLevelFn, failoverQueueProcessor := newTimerQueueFailoverProcessor(
 		standbyClusterName,
 		t.shard,
-		t.historyEngine,
 		t.taskProcessor,
 		t.taskAllocator,
 		t.activeTaskExecutor,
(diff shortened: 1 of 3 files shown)


davidporter-id-au (approved):

service/history/queue/transfer_queue_processor.go
```diff
@@ -517,7 +528,6 @@ func newTransferQueueActiveProcessor(
 func newTransferQueueStandbyProcessor(
 	clusterName string,
 	shard shard.Context,
-	historyEngine engine.Engine,
```
Shaddoll, line 520: historyEngine parameter can also be removed for failover queue processor

service/history/queue/timer_queue_processor.go
```diff
@@ -215,6 +216,15 @@ func (t *timerQueueProcessor) Stop() {
...
 	}
+
+	if len(t.failoverQueueProcessors) > 0 {
+		t.logger.Info("Shutting down failover timer queues", tag.Counter(len(t.failoverQueueProcessors)))
+		for _, failoverQueueProcessor := range t.failoverQueueProcessors {
+			failoverQueueProcessor.Stop()
+		}
+	}
+
+	t.activeTaskExecutor.Stop()
```
Shaddoll, line 227: why is the order changed?

taylanisikdemir: `t.activeTaskExecutor` is used by failover queues so stopping it after failover queues

service/history/queue/transfer_queue_processor.go
taylanisikdemir: yes done

Shaddoll (approved):

PR #6025, opened 2024-05-16
add Copy() method to configstore structs

mantas-sidlauskas:
**What changed?**
Helper functions `copy*` are replaced with struct methods

**Why?**
Better reusability 

**How did you test it?**
Added unit tests

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/common/types/configStore_test.go
+++ b/common/types/configStore_test.go
@@ -24,7 +24,7 @@ package types
 
 import (
 	"testing"
-	
+
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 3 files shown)


common/types/configStore_test.go
Shaddoll, line 1: Do you mind adding tests verifying that changing the copy doesn't change the original?

mantas-sidlauskas: good idea! Will do

taylanisikdemir (approved):

PR #6027, opened 2024-05-18
Global ratelimiter helper: a small atomic-like map

Groxx:
This commit largely exists to break this out for separate review, as it's relatively simple and isolated.

In a later PR, this will be used by the "limiter"-side logic to hold known ratelimiters, and simplify its logic.  So it does just two core things, and exposes a *very* small API that's intended to be difficult or impossible to misuse:

1. Tracks length, so a "iterate over everything and collect data for sending to aggregators" loop can pre-allocate a collection that is likely large enough to store everything.
2. Implicitly initializes missing values, so the using logic does not ever need to check for existence or explicitly handle fallbacks or decide how to order atomic operations.

It could have other methods, but so far they do not seem necessary.  Delete is not currently planned to be used, but it seems important to reserve / guarantee it's possible to build, so I've included that as well.

---

This is not intended to be broadly reusable as we do not currently have any other locations that will use it, and the auto-initializing behavior is potentially unique.  It's mostly just isolating and simplifying some logic to prevent accidental misuse, because early iterations had some easily-missed flaws.

We very likely *should* have a "just a type-safe `sync.Map`" wrapper, which this could easily be changed to use, but the vast majority of that API would be unused here so I haven't built that.

### DIFF
--- a/common/quotas/global/collection/internal/atomicmap.go
+++ b/common/quotas/global/collection/internal/atomicmap.go
@@ -1,0 +1,104 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


common/quotas/global/collection/internal/atomicmap.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"sync"
+	"sync/atomic"
+)
+
+// AtomicMap adds type safety around a sync.Map (which has atomic-like behavior), and:
+//   - implicitly constructs values as needed, not relying on zero values
+//   - simplifies the API quite a bit because very few methods are in use.
+//     in particular there is no "Store" currently because it is not needed.
+//   - tracks length (atomically, so values are only an estimate)
```
Shaddoll, line 34: is the length only an estimate number?

common/quotas/global/collection/internal/atomicmap_external_test.go
```diff
@@ -0,0 +1,185 @@
...
+	m := internal.NewAtomicMap(func(key string) *string {
+		s := key
+		s += "-"
+		s += strconv.Itoa(int(creates.Inc())) // just to be recognizable
+		return &s
+	})
+
+	// call ALL the methods concurrently
+	var g errgroup.Group
+	const loops = 1000 // 100 has had some coverage flapping, raise further if needed
```
Shaddoll, line 104: nit: mismatch

Shaddoll (approved):

common/quotas/global/collection/internal/atomicmap.go
Groxx: it's mostly a "by the time the func returns it may be incorrect, so you can never trust it except as an estimate".

the count should be accurate / it won't drift or be a `return 100` false number of some kind.  but it's inherently unreliable.

common/quotas/global/collection/internal/atomicmap_external_test.go
Groxx: I can reword that, it was meant as a "using 1000 because 100 was too low"

Groxx, line 104:
```suggestion
	const loops = 1000 // using 1,000 because 100 had some coverage flapping
```

PR #6028, opened 2024-05-18
Global ratelimiter helper: usage-tracking fallback-capable rate.Limiter

Groxx:
This commit largely exists to break this out for separate review, as it's relatively simple and isolated.

In a later PR, this will be used by the "limiter"-side logic to track and enforce usage of a ratelimit.  At a high level it does only a couple things:

1. Tracks calls to `Allow() bool` so we can find out how much a limit is actually used (so we can distribute the allowances fairly)
2. Updates ratelimit rates when requested
3. Automatically falls back to different ratelimit if the system seems unhealthy

^ all of which is done in an atomic-like way so concurrent operations do not block each other, as true precision is not necessary.  It only has to be "good enough", and performance is more important than precision.

### DIFF
--- a/common/quotas/global/collection/internal/limiter.go
+++ b/common/quotas/global/collection/internal/limiter.go
@@ -45,12 +45,13 @@ type (
 	// to use after a configurable number of failed updates.
 	//
 	// Intended use is:
-	//   - collect allowed vs rejected metrics
-	//   - periodically, the limiting host gathers all FallbackLimiter metrics and zeros them
+	//   - collect allowed vs rejected metrics (implicitly tracked by calling Allow())
+	//   - periodically, the limiting host gathers all FallbackLimiter metrics and zeros them (with Collect())
 	//   - this info is submitted to aggregating hosts, who compute new target RPS values
-	//   - these new target values are used to adjust this ratelimiter
+	//   - these new target values are used to adjust this ratelimiter (with Update(...))
 	//
-	// During this sequence, a requested limit may not be returned by an aggregator for two major reasons:
+	// During this sequence, a requested limit may not be returned by an aggregator for two major reasons,
+	// and will result in a FailedUpdate() call to shorten a "use fallback logic" fuse:
 	//   - this limit is legitimately unused and no data exists for it
 	//   - ring re-sharding led to losing track of the limit, and it is now owned by a host with insufficient data
 	//
--- a/common/quotas/global/collection/internal/limiter.go
+++ b/common/quotas/global/collection/internal/limiter.go
@@ -1,0 +1,201 @@
+// The MIT License (MIT)
+
+// Copyright (c) 2017-2020 Uber Technologies Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


Groxx:
this might change if we go with #6030, e.g. it could be implemented as a wrapper and then it could collect accurate reserve->cancel metrics too.

common/quotas/global/collection/internal/limiter.go
```diff
@@ -0,0 +1,201 @@
...
+	// Wait and Reserve are intentionally excluded here.  They are not currently needed,
+	// and it seems likely to be much more difficult to correctly track their usage.
+	Limiter interface {
+		Allow() bool
+	}
+
+	// FallbackLimiter wraps a [rate.Limiter] with a fallback Limiter (i.e. a [github.com/uber/cadence/common/quotas.Limiter])
+	// to use after a configurable number of failed updates.
+	//
+	// Intended use is:
```
Shaddoll, line 47: Could you mention which method is called for each use case?

common/quotas/global/collection/internal/limiter.go
```diff
@@ -0,0 +1,201 @@
...
+	//   - collect allowed vs rejected metrics
+	//   - periodically, the limiting host gathers all FallbackLimiter metrics and zeros them
+	//   - this info is submitted to aggregating hosts, who compute new target RPS values
+	//   - these new target values are used to adjust this ratelimiter
+	//
+	// During this sequence, a requested limit may not be returned by an aggregator for two major reasons:
+	//   - this limit is legitimately unused and no data exists for it
+	//   - ring re-sharding led to losing track of the limit, and it is now owned by a host with insufficient data
+	//
+	// To mitigate the impact of the second case, "insufficient data" responses from aggregating hosts
```
Shaddoll, line 57: perhaps, it's worth adding some metrics tracking when we fallback

Groxx: yeah, I'm planning on metrics.  that'll be outside these objects though, and it's why `Collect()` returns the "use fallback" bool.

common/quotas/global/collection/internal/limiter.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// FallbackLimiter wraps a [rate.Limiter] with a fallback Limiter (i.e. a [github.com/uber/cadence/common/quotas.Limiter])
+	// to use after a configurable number of failed updates.
+	//
+	// Intended use is:
+	//   - collect allowed vs rejected metrics
+	//   - periodically, the limiting host gathers all FallbackLimiter metrics and zeros them
+	//   - this info is submitted to aggregating hosts, who compute new target RPS values
+	//   - these new target values are used to adjust this ratelimiter
```
Groxx, line 51:
```suggestion
	//   - collect allowed vs rejected metrics (implicitly tracked by calling Allow())
	//   - periodically, the limiting host gathers all FallbackLimiter metrics and zeros them (with Collect())
	//   - this info is submitted to aggregating hosts, who compute new target RPS values
	//   - these new target values are used to adjust this ratelimiter (with Update(...))
```

common/quotas/global/collection/internal/limiter.go
```diff
@@ -0,0 +1,201 @@
...
+	//
+	// During this sequence, a requested limit may not be returned by an aggregator for two major reasons:
```
Groxx, line 53:
```suggestion
	// During this sequence, a requested limit may not be returned by an aggregator for two major reasons,
	// and will result in a FailedUpdate() call to shorten a "use fallback logic" fuse:
```

common/quotas/global/collection/internal/limiter.go
Groxx: added - how does that look?

common/quotas/global/collection/internal/limiter.go
Shaddoll: If there is a transition from non-fallback to fallback or vice versa, the counter will be inaccurate.

Groxx: if you mean the limiting behavior will be inaccurate, e.g. it'll allow / deny more requests than it should:

yes, but we don't have any great way to resolve that.  `*rate.Limiter` doesn't allow directly manipulating its tokens, and there isn't even a true "correct" behavior if this happens because it implies we _don't know_ the correct behavior - we haven't been able to sync with other instances.

Shaddoll (approved):

common/quotas/global/collection/internal/limiter.go
Groxx: tl;dr from IRL, from clarifying some of the surroundings here:

a layer above this, which holds all these individual tracking-and-fallback-able limiters, will be periodically calling `Collect()` and it'll emit a counter for "using fallback" so we can monitor how often it uses the fallback.

but "uses fallback" should only occur in practice:
- temporarily after startup / when a key is first used (no data, must use fallback)
- when the global limiter is disabled and rejects all requests (falling back is desired here)
- when something cataclysmic is happening and global limit data is being lost rapidly, e.g. many host crashes -> ring changes rapidly -> does not return weights before crashing too.  (falling back is desirable here too, as the old values are old enough to not trust)

so the sustained existence of fallback-using things is a concern, but not individual "failed update" events or temporary use.

the exact metrics emitted are going to be in a later diff, that all happens outside of this object.

PR #6029, opened 2024-05-19
Mock-clock-capable ratelimiter wrapper

Groxx:


Groxx:
abandoning, #6030 has some subtle fixes and quite a large benchmark ensuring it's usable.

